PR #25, opened 2013-11-26
Navitiaii 364

antoine-de:
reopening of the connections modifications.

The patch with the search from additional stop points (by reaching the connections) was very slow.

It was due to several problems but mostly it was because some points were not reached by the dijkstra (they were not connected to the graph)

the changes are:
- all projections are done on the right graph (one graph by transportation mode)
- cache of the connections by stop points
- storage of the stop point projections for each graph
- misc corrections in the get distances
- small refactoring of the offset (they are now stored in a flat_enum_array, to be accessible with an Mode_e enum value)

### DIFF
--- a/source/georef/georef.h
+++ b/source/georef/georef.h
@@ -14,6 +14,7 @@
 #include <boost/graph/dijkstra_shortest_paths.hpp>
 #include <map>
 #include "adminref.h"
+#include "utils/logger.h"
 
 
 namespace nt = navitia::type;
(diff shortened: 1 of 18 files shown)


source/georef/georef.h
```diff
@@ -324,6 +331,7 @@ struct GeoRef {
 // Visiteur qui lève une exception dès qu'une des cibles souhaitées est atteinte
 struct target_visitor : public boost::dijkstra_visitor<> {
     const std::vector<vertex_t> & destinations;
+    size_t cpt;
```
l-vincent-l: I think this was for debuging purpose.

source/georef/georef.h
```diff
@@ -380,11 +376,15 @@ struct ProjectionData {
     ProjectionData() : found(false), source_distance(-1), target_distance(-1){}
     /// Initialise la structure à partir d'une coordonnée et d'un graphe sur lequel on projette
     ProjectionData(const type::GeographicalCoord & coord, const GeoRef &sn, const proximitylist::ProximityList<vertex_t> &prox);
+    /// Initialise la structure à partir d'une coordonnée, d'un graphe sur lequel on projette et d'un offset qui correspond au mode de transport
```
l-vincent-l: English please !

PR #28, opened 2013-11-28
Jormungandr: Small changes in the api

antoine-de:
Correction of the api entry points
Correction of the coord href


Tristramg:
Yihaaa ! At last :dancer: :dancers:

PR #36, opened 2013-12-04
All: a few fix for cppcheck

kinnou02:
It's Mostly variables that were not initialized in the constructor


kinnou02:
dans le cas des types primitif c'est pas faux

PR #47, opened 2013-12-12
Kraken: signal handling

antoine-de:
add as the first method called in main the initialization of loggers and
the handling the signal.
With that if the program is about to end it prints the stack before core
dumping
- correction of a mistake in previous merge

### DIFF
--- a/source/ed/gtfs2ed.cpp
+++ b/source/ed/gtfs2ed.cpp
@@ -114,7 +114,7 @@ int main(int argc, char * argv[])
     LOG4CPLUS_INFO(logger, "alias : " <<data.alias.size());
     LOG4CPLUS_INFO(logger, "synonyms : " <<data.synonymes.size());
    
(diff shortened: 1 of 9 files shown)


source/ed/gtfs2ed.cpp
```diff
@@ -111,8 +113,8 @@ int main(int argc, char * argv[])
     LOG4CPLUS_INFO(logger, "journey_pattern point connections : " << data.journey_pattern_point_connections.size());
     LOG4CPLUS_INFO(logger, "alias : " <<data.alias.size());
     LOG4CPLUS_INFO(logger, "synonyms : " <<data.synonymes.size());
-
-    start = pt::microsec_clock::local_time();
+   
+	start = pt::microsec_clock::local_time();
```
l-vincent-l: Du coup, le but de ce commit c'est d'enlever une indentation correcte ?

PR #100, opened 2014-01-24
Navitiaii 546

antoine-de:
several fixes:
- path item length computation. 
  the length has to be computed since we only store the edges duration, and the computation was buggy
- projections are now done with the same mean of transport as the rest, so there will be no separated section
- there is however still a problem with the length of the projection.
  the projection are correctly taken into account during computation but their durations are now displayed in the path.
  so the path.duration is not the sum of the path_item durations.
  we need to fixe that

### DIFF
--- a/source/routing/tests/routing_api_test.cpp
+++ b/source/routing/tests/routing_api_test.cpp
@@ -813,7 +813,7 @@ BOOST_FIXTURE_TEST_CASE(biking_length_test, streetnetworkmode_fixture<normal_spe
     BOOST_CHECK_EQUAL(pathitem.name(), "rue bs");
     BOOST_CHECK_EQUAL(pathitem.duration(), 0);
     BOOST_CHECK_CLOSE(pathitem.length(), 0, 1);
-    //BUG: for the moment the projections are now taken into account, we have to work on that
+    //BUG: for the moment the projections are not taken into account, we have to work on that
 //    BOOST_CHECK_EQUAL(pathitem.duration(), to_duration(B.distance_to(S), type::Mode_e::Bike).total_seconds());
 //    BOOST_CHECK_CLOSE(pathitem.length(), B.distance_to(S), 1);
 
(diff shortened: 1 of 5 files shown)


l-vincent-l:
T'as fait un ticket pour le truc  à fixer ?

source/routing/tests/routing_api_test.cpp
```diff
@@ -745,6 +773,84 @@ struct streetnetworkmode_fixture {
...
+    pathitem = section.street_network().path_items(cpt++);
+    BOOST_CHECK_EQUAL(pathitem.name(), "rue gh");
+    BOOST_CHECK_EQUAL(pathitem.duration(), to_duration(G.distance_to(H), type::Mode_e::Bike).total_seconds());
+    BOOST_CHECK_CLOSE(pathitem.length(), G.distance_to(H), 1);
+
+    pathitem = section.street_network().path_items(cpt++);
+    BOOST_CHECK_EQUAL(pathitem.name(), "rue ag");
+    BOOST_CHECK_EQUAL(pathitem.duration(), 0);
+    BOOST_CHECK_CLOSE(pathitem.length(), 0, 1);
+    //same than below, it's a bug
```
l-vincent-l: Je pense que tu voulais dire above

source/routing/tests/routing_api_test.cpp
```diff
@@ -745,6 +773,84 @@ struct streetnetworkmode_fixture {
...
+        }
+    }
+
+    BOOST_REQUIRE_EQUAL(path_items.size(), 7);
+    int cpt(0);
+    auto pathitem = path_items[cpt++];
+    BOOST_CHECK_EQUAL(pathitem.name(), "rue bs");
+    BOOST_CHECK_EQUAL(pathitem.duration(), 0);
+    BOOST_CHECK_CLOSE(pathitem.length(), 0, 1);
+    //BUG: for the moment the projections are now taken into account, we have to work on that
```
l-vincent-l: now = not ?

PR #143, opened 2014-02-18
Cheap journeys

antoine-de:
modification to give alternatives if only train solution (to show cheaper journeys with bus for example).

If there is only train journeys, call again kraken forbidding the use of the french railway network.

this can be activated or desctivated with the parameter:
[functional]
cheap_journey={True|False}

### DIFF
--- a/source/jormungandr/jormungandr/scripts/default.py
+++ b/source/jormungandr/jormungandr/scripts/default.py
@@ -199,7 +199,7 @@ class Script(object):
                     uris.physical_mode = pt_infos.uris.physical_mode
                     uris.network = pt_infos.uris.network
 
-    def get_journey(self, req, instance, trip_type, debug):
+    def get_journey(self, req, instance):
         resp = None
 
         for o_mode, d_mode in itertools.product(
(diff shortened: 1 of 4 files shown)


source/jormungandr/jormungandr/scripts/default.py
```diff
@@ -168,6 +170,84 @@ def departure_boards(self, request, instance):
...
+
+        if req.journeys.streetnetwork_params.origin_mode == "bike_rental":
+            req.journeys.streetnetwork_params.origin_mode = "bss"
+        if req.journeys.streetnetwork_params.destination_mode == "bike_rental":
+            req.journeys.streetnetwork_params.destination_mode = "bss"
+        if "forbidden_uris[]" in request and request["forbidden_uris[]"]:
+            for forbidden_uri in request["forbidden_uris[]"]:
+                req.journeys.forbidden_uris.append(forbidden_uri)
+        if not "type" in request:
+            request["type"] = "all" #why ?
```
l-vincent-l, line 219: ouais, vu que normalement dans les paramètres il y a un défaut ça peut s'enlever.

PR #145, opened 2014-02-19
hotfix 0.82.4

antoine-de:
fare bug correciton


antoine-de:
integrated from release

PR #156, opened 2014-03-03
Tyr improvement

kinnou02:
Mostly related to the log, by default in dev the log are on stdout. The configuration format is now the same as in jormungandr.

it's also possible to launch tyr with a Procfile:
Procfile describe the processes needed for run the application.
All components of tyr can no be start with the command:

```
honcho start
```

honcho must be installed with pip.
you can create a file named ".env" for automatically export environnement
variable. Example:

```
PYTHONPATH=../navitiacommon
TYR_CONFIG_FILE=default_settings.py
```

it's also possible to use foreman in place of honcho

### DIFF
--- a/source/tyr/PaysDeLaLoire.ini
+++ b/source/tyr/PaysDeLaLoire.ini
@@ -1,5 +1,5 @@
 [instance]
-name = nantes
+name = picardie
 source-directory = /tmp/ed/
 backup-directory = /tmp/edbackup/
 tmp-file = /tmp/ed/tmpdata.nav.lz4
--- a/source/tyr/PaysDeLaLoire.ini
+++ b/source/tyr/PaysDeLaLoire.ini
@@ -1,5 +1,5 @@
 [instance]
-name = picardie
+name = PaysDeLaLoire
 source-directory = /tmp/ed/
 backup-directory = /tmp/edbackup/
 tmp-file = /tmp/ed/tmpdata.nav.lz4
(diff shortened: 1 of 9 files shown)


source/tyr/PaysDeLaLoire.ini
```diff
@@ -1,5 +1,5 @@
 [instance]
-name = nantes
+name = picardie
```
antoine-de: on peut pas mentir sur ses origines.
par contre je suis pas certain que ce soit le lieu pour l'assumer.

source/tyr/tyr/aggregate_places.py
```diff
@@ -22,12 +23,14 @@ def aggregate_places(instance_config, job_id):
...
 
+    logger = logging.getLogger(__name__)
+
     instance_ = models.Instance.query.\
         filter(models.Instance.name == instance_config.name)
     if instance_.first():
         instance = instance_.first()
     else:
-        app.logger.error("Instance %s can not be found"%instance_config.name)
+        logger.error("Instance %s can not be found"%instance_config.name)
```
antoine-de: bon je pinaille mais c'est mieux de mettre des espaces autour du % non ?

source/tyr/tyr/helper.py
```diff
@@ -131,20 +127,22 @@ def build_error(config, validate_result):
...
-    if not logger.handlers:
-        log_dir = os.path.dirname(current_app.config['LOG_FILENAME'])
+    root_logger = logging.root
+    logger.setLevel(root_logger.level)
+    file_hanlders = [h for h in root_logger.handlers
+                    if isinstance(h, logging.FileHandler)]
+    #does not add the handler at each time or if we do not log to a file
+    if not logger.handlers and file_hanlders:
+        log_dir = os.path.dirname(file_hanlders[0].stream.name)
         log_filename = log_dir + '/{0}.log'.format(instance.name)
```
antoine-de: autant faire un "{}/{}.log".format(log_dir, instance.name) non ?

PR #165, opened 2014-03-04
Kraken: speed of a bike is now 15km/h (i.e 4.1m/s)

l-vincent-l:


Tristramg:
I think it is a bit fast in a urban context. 15km/h seems more reasonable

l-vincent-l:
Took in account the suggestion of @Tristramg

PR #186, opened 2014-03-10
Jormungandr: fix resource uri

l-vincent-l:
Check the presence of region for api that needs it.


source/jormungandr/jormungandr/interfaces/v1/ResourceUri.py
```diff
@@ -205,54 +206,6 @@ def add_note(data):
 
         return wrapper
 
-class add_exception_dates(object):
```
antoine-de, line 208: je pense que tu as effacé ca par erreur

PR #210, opened 2014-03-18
Models: We can now request models by external_code and type

l-vincent-l:
We were sending the first one before.

### DIFF
--- a/source/jormungandr/jormungandr/interfaces/v1/Uri.py
+++ b/source/jormungandr/jormungandr/interfaces/v1/Uri.py
@@ -48,12 +48,18 @@ class Uri(ResourceUri):
         args = self.parsers["get"].parse_args()
         if region is None and lat is None and lon is None:
             if "external_code" in args and args["external_code"]:
-                res = PtObject.get_from_external_code(args["external_code"])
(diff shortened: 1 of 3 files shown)


source/jormungandr/jormungandr/interfaces/v1/Uri.py
```diff
@@ -48,12 +48,21 @@ def get(self, region=None, lon=None, lat=None, uri=None, id=None):
         args = self.parsers["get"].parse_args()
         if region is None and lat is None and lon is None:
             if "external_code" in args and args["external_code"]:
-                res = PtObject.get_from_external_code(args["external_code"])
+                if not collection:
+                    abort(400)
```
antoine-de: en fait le test ne sert a rien

source/jormungandr/jormungandr/interfaces/v1/Uri.py
```diff
@@ -48,12 +48,21 @@ def get(self, region=None, lon=None, lat=None, uri=None, id=None):
...
+                if collection not in collections_to_resource_type:
+                    abort(400)
```
antoine-de: mais la il faudrait peut etre des messages d'erreur

PR #232, opened 2014-04-02
Jormungandr: fix next link for journey pagination

l-vincent-l:
There was a mess between arrival and departure.

Plus we know check only journeys with public transport.
And we were asking for the one leaving one minute after, now we take the one leaving one second after.

### DIFF
--- a/source/jormungandr/jormungandr/interfaces/v1/Journeys.py
+++ b/source/jormungandr/jormungandr/interfaces/v1/Journeys.py
@@ -320,16 +320,14 @@ class add_journey_pagination(object):
     def extremes(self, resp):
         datetime_before = None
         datetime_after = None
+        section_is_pt = lambda section: section['type'] == "public_transport"\
+                           or section['type'] == "on_demand_transport"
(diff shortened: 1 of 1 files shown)


source/jormungandr/jormungandr/interfaces/v1/Journeys.py
```diff
@@ -321,23 +321,28 @@ def extremes(self, resp):
         datetime_before = None
         datetime_after = None
         try:
+            section_is_pt = lambda section : section['type'] == "public_transport"\
+                               or section['type'] == "on_demand_transport"
+            or_func = lambda a,b : a or b
+            journey_as_pt = lambda journey : reduce(or_func, journey['sections'])
```
antoine-de: ca peut etre sorti du try non ?

antoine-de: je suis pas certain de bien comprendre, mais en fait tu veux pas faire un any(section_is_pt(s) for s in journey['sections']) ?

l-vincent-l:
test this please

l-vincent-l:
test this please

PR #277, opened 2014-04-22
Add commands tyr

l-vincent-l:
- Add command reload_kraken, you can use it with manage_tyr.py reload_kraken <instance_name>
- Add command build_datas that build datas of all instances, you can use it with manage_tyr.py build_datas


source/tyr/manage_tyr.py
```diff
@@ -15,6 +16,8 @@
 manager.add_command('aggregate_places', AggregatePlacesCommand())
 manager.add_command('reload_at', ReloadAtCommand())
 manager.add_command('at_reloader', AtReloader())
+manager.add_command('reload_kraken', ReloadKrakenCommand())
+manager.add_command('build_datas', BuildDatasCommand())
```
antoine-de: je pinaille mais data c'est indénombrable :)

PR #296, opened 2014-04-25
Fix #292 part3

### LINKED ISSUE #292: Doesn't compile on Ubuntu 14.04 with listed dependencies
With all the dependencies that are listed installed, getting this linker error at 48% of the compile:

/usr/bin/ld: CMakeFiles/osm2ed.dir/osm2ed.cpp.o: undefined reference to symbol 'inflate'
//lib/x86_64-linux-gnu/libz.so.1: error adding symbols: DSO missing from command line

koch-t:
These were also necessary for GCC 4.8
the package visu also had to be disabled, but i'm not sure if that's compiler related.

### DIFF
--- a/source/calendar/tests/CMakeLists.txt
+++ b/source/calendar/tests/CMakeLists.txt
@@ -5,6 +5,6 @@ SET(BOOST_LIBS ${BOOST_LIB} ${Boost_UNIT_TEST_FRAMEWORK_LIBRARY} ${Boost_PROGRAM
 
 add_executable(calendar_api_test calendar_api_test.cpp)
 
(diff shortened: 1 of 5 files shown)


buildbotctp:

antoine-de:
Strange, worked well for us without that.

I'll try again as soon as possible.

Thanks for the patch again!

Jenkins, test this please.

kinnou02:
For visu it's probably related to the dependence to marble. Anyway visu is not realy used.

ok to test.

kinnou02:
ok to test.

kinnou02:
Sorry for the spam, jenkins don't like commands with other messages.
Thanks for the patch.

PR #300, opened 2014-04-27
Replace zmq_device with zmq_proxy.

skinkie:
In ZeroMQ3 zmq_device has been removed. We should use zmq_proxy instead.

### DIFF
--- a/source/kraken/kraken_zmq.cpp
+++ b/source/kraken/kraken_zmq.cpp
@@ -79,7 +79,7 @@ int main(int, char** argv){
     // Connect work threads to client threads via a queue
     do{
         try{
-            zmq::device(ZMQ_QUEUE, clients, workers);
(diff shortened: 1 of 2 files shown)


buildbotctp:

kinnou02:
I don't think it'll work with zmq2 who is used in debian stable

Jenkins, test this please.

skinkie:
If it doesn't work I'll add:

``` C
#if ZMQ_VERSION_MAJOR == 2
...
#elif ZMQ_VERSION_MAJOR >= 3
...
#endif
```

kinnou02:
jenkins: ok to test.

kinnou02:
zmq 3 compatibility has been integrated in https://github.com/CanalTP/navitia/pull/307
So I close this one.

PR #301, opened 2014-04-27
In recent versions of libpqxx the pqxx::prepare::treat_string has been removed.

skinkie:
Please validate this code works as expected.


buildbotctp:

kinnou02:
ok to test.

kinnou02:
Same problem as with zmq, it's builds with pqxx 3 but doesn't works:

```
Too many arguments for prepared statement messages: expected 0, received 2
```

antoine-de:
This PR cannot be merged 'as is' for the moment because we need a more complex patch to handle the 2 different versions of pqxx (the are unfortunatly not compatible...).

We'll work on that a soon as possible, thanks again @skinkie !

PR #304, opened 2014-04-28
Set GTFS bikes allowed property in VehicleJourney

koch-t:
bikes_allowed has been a official GTFS parameter since a couple of months.


buildbotctp:

kinnou02:
ok to test.

antoine-de:
Thanks for the patch @skywave.

You have to take note however that the BIKE_ACCEPTED property is not used for the moment in the algorithm.
It is however still useful information to have and to be displayed in the ptref queries.

PR #308, opened 2014-04-29
kraken: Correction for API place with uri

kadhikari:
For an appel place with uri we can use the function fill_pb_placemark
where attribut Name is formatted properly.


buildbotctp:

kinnou02:
ok to test.

PR #318, opened 2014-05-03
Use the departure validity pattern for pickup

koch-t:
Reasoning for this patch: st->departure(_adapted)_validity_pattern is set in data.cpp but never used.

Note:
I'm looking over the midnight switchover code due to some segfaults we're experiencing. This didn't fix it however.


buildbotctp:

l-vincent-l:
ok to test

PR #321, opened 2014-05-05
Fix one cause for segfault for trains w/ midnight departure

koch-t:
It fixes some segfault's but not all.

### DIFF
--- a/source/type/data.cpp
+++ b/source/type/data.cpp
@@ -188,7 +188,7 @@ ValidityPattern* Data::get_similar_validity_pattern(ValidityPattern* vp) const{
 
 ValidityPattern* Data::get_or_create_validity_pattern(ValidityPattern* ref_validity_pattern, const uint32_t time){
 
-    if(time > 24*3600) {
+    if(time >= 24*3600) { //Seconds in a day
         ValidityPattern vp(ref_validity_pattern->beginning_date,ref_validity_pattern->str());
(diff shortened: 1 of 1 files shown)


buildbotctp:

koch-t:
Without this patch Navitia is unable to find stop_time's with a departure_time of 24:00:00 and trigger segfaults in some scenario's.

But with this patch Navitia triggers another bug causing segfaults which would previously result in just a "No solutions found" error.

antoine-de:
We are refactoring this part, so I think we'll integrate your patch in a bigger pull request.

antoine-de:
ok to test

l-vincent-l:
Can you send the request causing the segfault with this PR?

koch-t:
stop_area:alm to stop_area:asd departure after 20140506T0000 

http://api.navitia.io/v1/journeys?from=stop_area:OAM:SA:17827&to=stop_area:OAM:SA:18188&datetime=20140506T000000
http://localhost:5000/v1/journeys?from=stop_area:alm&to=stop_area:asd&datetime=20140506T000000

Should give the train
(stop_area:alm, arrives at 23:59 departs at 24:01) -> (stop_area:asd, arrives at 24:24:00)

The biggest problem is that it does work on the smaller train-dataset.

l-vincent-l:
Fixed by #331

PR #299, opened 2014-04-26
Refactor double negation away

koch-t:

Issue #297, opened 2014-04-25
GTFS reader fails to read stopname with backslash


koch-t:
```
WARN - Impossible de parser la ligne :  141581,16058601,"Siebengewald, Gochsedijk\Centrum",51.648438,6.108465,0,stoparea:42204,,,,
WARN - Impossible de parser la ligne :  141582,16058602,"Siebengewald, Gochsedijk\Centrum",51.647821,6.106519,0,stoparea:42204,,,,
WARN - Impossible de parser la ligne :  stoparea:42204,,"Siebengewald, Gochsedijk\Centrum",51.6481295,6.107492,1,,,0,,
WARN - Impossible de parser la ligne :  stoparea:11488,,"Maastricht, Meerssenenerweg\Miradorplein",50.865959,5.714405,1,,,0,,
```

koch-t:
This appears to be a issue with Boost's escaped_list_separator:
http://stackoverflow.com/posts/1595366/revisions

PR #322, opened 2014-05-07
Fix gtfs

### LINKED ISSUE #298: GTFS parser has difficulties with stoppoints without stoparea.
It deletes all the stoppoints with stoparea, resulting in not being able to find stops for a lot of records in stop_times.txt. 
Reading at the code, i think this is only a issue with feeds that do not contain a stoparea for all stops.

antoine-de:
Fix #298 and close #299

still need to work on the handling of '\' in csv files (for #297), but since it does not seems possible to handle it with boost tokenizer, it may be quite a serious refactoring (and with impact on all csv).

I don't know yet what is the best course of action:
- ask to double the slashes in the data
- change the string before hand to double the slashes (seems quite uggly)
- change to boost::spirit to parse the file
- another way :)

Any suggestion ?


antoine-de:
close the issue, we found some more stuff to be fixed

PR #329, opened 2014-05-09
Parse OSM platform ways

koch-t:
Many micromapped stations are tagged with either railway=platform / public_transport=platform.
TODO: Check if liker prefers linking with these public_transport platforms.


buildbotctp:

PR #333, opened 2014-05-10
Don't expand house numbers to 1/2 the world

koch-t:
st_expand with geometry is in the units of the given geometry, thus for WGS84, we were expanding the housenumbers to 100 degrees, this negated any efficiency of the index. 
Reduce to 0.0025 which is 195 meters at 45 degrees of latitude, 107m at 67 degrees, 277m at the equator


buildbotctp:

l-vincent-l:
ok to test

PR #338, opened 2014-05-14
Kraken: Fill stop_area instead of stop_point

l-vincent-l:
For the from of a first section if the origin is a stop_area
The same thing is done for the to of a last section

### DIFF
--- a/source/routing/raptor_api.cpp
+++ b/source/routing/raptor_api.cpp
@@ -128,7 +128,8 @@ pbnavitia::Response make_pathes(const std::vector<navitia::routing::Path>& paths
         const type::VehicleJourney* vj = nullptr;
         size_t item_idx(0);
         // La partie TC et correspondances
-        for(PathItem & item : path.items) {
+        for(auto path_i = path.items.begin(); path_i < path.items.end(); ++path_i) {
+            const auto item = *path_i;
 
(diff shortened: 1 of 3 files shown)


source/routing/raptor_api.cpp
```diff
@@ -139,43 +140,62 @@ pbnavitia::Response make_pathes(const std::vector<navitia::routing::Path>& paths
...
-                            departure_ptime = p_deptime;
-                        // L'heure d'arrivée au dernier stop point
-                        arrival_ptime = p_arrtime;
-                        if(i>0) {
-                            const auto & previous_coord = item.stop_points[i-1]->coord;
-                            const auto & current_coord = item.stop_points[i]->coord;
-                            length += previous_coord.distance_to(current_coord);
-                        }
+                    if (vj->has_boarding() || vj->has_landing()) {
+                        continue;
```
antoine-de, line 144: \o/

l-vincent-l:
Is that OK with you if i nest a stop point in the stop area?

koch-t:
I could work with that but i do dislike the inconsistency of it.

l-vincent-l:
It was made for consistency purpose, we want you to have in the form of the first section (and in the "to" of the last section) the object you asked in the the from (respectively the "to") parameter of your request.

PR #354, opened 2014-05-21
Tyr load data

antoine-de:
Add a load_data target for tyr

2 params are needed: the name of the kraken instance and the data dir.

the target will launch the whole data import process:
- data import in bdd (fusio2ed, gtfs2ed, poi2ed, ...)
- export bdd to nav file
- update the jormungandr db with the new data for the instance
- reload the krakens

the original data will be neither moved nor modified.

This target will be used mainly in Artemis (the business test framework)

example call:
`TYR_CONFIG_FILE=/srv/tyr/settings.py python /srv/tyr/manage.py load_data airport-01 /srv/ed/airport-01/data/`

### DIFF
--- a/source/tyr/tyr/tasks.py
+++ b/source/tyr/tyr/tasks.py
@@ -80,76 +80,70 @@ def finish_job(job_id):
     models.db.session.commit()
 
 
+def import_data(files, instance, backup_file):
+    actions = []
+    job = models.Job()
+    instance_config = load_instance_config(instance.name)
+    job.instance = instance
+    job.state = 'pending'
+    task = {
+        'gtfs': gtfs2ed,
+        'fusio': fusio2ed,
+        'osm': osm2ed,
+        'geopal': geopal2ed,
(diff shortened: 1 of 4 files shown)


source/tyr/tyr/tasks.py
```diff
@@ -80,77 +80,71 @@ def finish_job(job_id):
...
-            elif dataset.type == 'fare':
-                filename = move_to_backupdirectory(_file,
-                        instance_config.backup_directory)
-                actions.append(fare2ed.si(instance_config, filename))
-            elif dataset.type == 'poi':
-                filename = move_to_backupdirectory(_file,
-                        instance_config.backup_directory)
-                actions.append(poi2ed.si(instance_config, filename))
-            elif dataset.type == 'synonym':
+def import_data(files, instance, backup_file):
```
kinnou02, line 83: Il manque la doc string
surtout parce qu'en premiére lecture j'avais pas compris le backup_file

PR #362, opened 2014-05-27
kraken : not display "pick_up only" if first stoptime

azime:
not display "pick_up only" if first stoptime , "drop_off only" if last stoptime in vehiclejourney


source/type/pb_converter.cpp
```diff
@@ -43,6 +43,25 @@ namespace pt = boost::posix_time;
 
 namespace navitia{
 
+void fill_pb_object(const navitia::type::StopTime* stop_time, const type::Data&,
+                    pbnavitia::Properties* properties, int,
+                    const boost::posix_time::ptime&, const boost::posix_time::time_period&){
```
antoine-de, line 48: @azime could you add a comment here the explain why we do that ?

azime, line 48: refactoring add properties,
not display : 
1) "pick_up only" if first stoptime in vehiclejourney
2) "drop_off only" if last stoptime in vehiclejourney

azime:
add comment

PR #372, opened 2014-06-06
Autocomplete: Sort StopArea by StopPoint count

kadhikari:
For each StopArea calculate the number of StopPoints.
Using maximun number of StopPoints, calculates score
of each StopPoints from 0 to 100.

### DIFF
--- a/source/autocomplete/autocomplete.cpp
+++ b/source/autocomplete/autocomplete.cpp
@@ -55,17 +55,6 @@ void compute_score_way(type::PT_Data&, georef::GeoRef &georef) {
 }
 
 
-void compute_score_stop_area(type::PT_Data &pt_data, georef::GeoRef &georef) {
-    for(auto it = pt_data.stop_area_autocomplete.word_quality_list.begin(); it != pt_data.stop_area_autocomplete.word_quality_list.end(); ++it){
-        for (navitia::georef::Admin* admin : pt_data.stop_areas[it->first]->admin_list){
-            if (admin->level == 8){
-                (it->second).score = georef.fl_admin.word_quality_list.at(admin->idx).score;
-            }
-        }
-    }
-}
-
-
 void compute_score_stop_point(type::PT_Data &pt_data, georef::GeoRef &georef) {
     //Récupérer le score de son admin du niveau 8 dans le autocomplete: georef.fl_admin
     for (auto it = pt_data.stop_point_autocomplete.word_quality_list.begin(); it != pt_data.stop_point_autocomplete.word_quality_list.end(); ++it){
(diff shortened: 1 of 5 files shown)


source/autocomplete/autocomplete.cpp
```diff
@@ -77,6 +66,31 @@ void compute_score_stop_point(type::PT_Data &pt_data, georef::GeoRef &georef) {
     }
 }
 
+void compute_score_stop_area_by_spcount(type::PT_Data & pt_data){
+    //count de number of stop_points in each stop_area:
+    std::map<type::idx_t, std::set<type::idx_t>> map_sa_sp;
+    for(navitia::type::StopPoint* sp : pt_data.stop_points){
+        if (sp->stop_area != NULL){
```
antoine-de: it is better to use nullptr instead of NULL

source/autocomplete/autocomplete.cpp
```diff
@@ -77,6 +66,31 @@ void compute_score_stop_point(type::PT_Data &pt_data, georef::GeoRef &georef) {
...
+            map_sa_sp[sp->stop_area->idx].insert(sp->idx);
+        }
+    }
+
+    //Calculate de maximum stop-point count;
+    int max_score = 0;
+    for (auto it = map_sa_sp.begin(); it != map_sa_sp.end(); ++it){
```
antoine-de: it is easier to use normalized for loop:

```
for (const auto& pair: map_sa_sp) {
max_score = pair.second.size()...
}
```

source/autocomplete/autocomplete.cpp
```diff
@@ -77,6 +66,31 @@ void compute_score_stop_point(type::PT_Data &pt_data, georef::GeoRef &georef) {
...
+        max_score = it->second.size() > max_score ? it->second.size() : max_score;
+    }
+
+
+    //Ajust the score of each stop_area from 0 to 100 using maximum score (max_score)
+    for (auto it = pt_data.stop_area_autocomplete.word_quality_list.begin(); it != pt_data.stop_area_autocomplete.word_quality_list.end(); ++it){
+        auto sa_sp = map_sa_sp.find(it->first);
+        if (sa_sp != map_sa_sp.end()){
+            it->second.score = max_score == 0 ? 0 : (sa_sp->second.size() * 100)/max_score;
```
antoine-de: I think there is a problem with this line. don't you want to only check that max_score is not null ?

source/autocomplete/autocomplete.cpp
```diff
@@ -77,6 +66,31 @@ void compute_score_stop_point(type::PT_Data &pt_data, georef::GeoRef &georef) {
     }
 }
 
+void compute_score_stop_area_by_spcount(type::PT_Data & pt_data){
```
antoine-de: why rename the function ? compute_score_stop_area is still fine no ? (since there are not other method for stop areas)

source/autocomplete/autocomplete.cpp
```diff
@@ -77,6 +66,31 @@ void compute_score_stop_point(type::PT_Data &pt_data, georef::GeoRef &georef) {
...
+    for(navitia::type::StopPoint* sp : pt_data.stop_points){
+        if (sp->stop_area != NULL){
+            map_sa_sp[sp->stop_area->idx].insert(sp->idx);
+        }
+    }
+
+    //Calculate de maximum stop-point count;
+    int max_score = 0;
+    for (auto it = map_sa_sp.begin(); it != map_sa_sp.end(); ++it){
+        max_score = it->second.size() > max_score ? it->second.size() : max_score;
```
antoine-de: for this you can just say max_score = std::max(max_score, it->second.size());

source/autocomplete/autocomplete.cpp
```diff
@@ -77,6 +66,20 @@ void compute_score_stop_point(type::PT_Data &pt_data, georef::GeoRef &georef) {
...
+void compute_score_stop_area(type::PT_Data & pt_data){
+    //Calculate de maximum stop-point count;
+    unsigned int max_score = 0;
+    for (navitia::type::StopArea* sa : pt_data.stop_areas){
+        max_score = std::max(max_score, sa->stop_point_list.size());
+    }
+
+    //Ajust the score of each stop_area from 0 to 100 using maximum score (max_score)
+    if (max_score > 0){
+        for (auto it = pt_data.stop_area_autocomplete.word_quality_list.begin(); it != pt_data.stop_area_autocomplete.word_quality_list.end(); ++it){
```
antoine-de: there you can also put the standard for loop, it will be easier to read

PR #382, opened 2014-06-09
ed: better error handling

antoine-de:
Add more precise errors for date parsing and calendars

### DIFF
--- a/source/ed/connectors/fusio_parser.cpp
+++ b/source/ed/connectors/fusio_parser.cpp
@@ -769,7 +769,7 @@ void GridCalendarFusioHandler::handle_line(Data& data, const csv_row& row, bool 
     }
     if (has_col(id_c, row) && row[id_c].empty()) {
         //we don't want empty named calendar as they often are empty lines in reality
-        LOG4CPLUS_ERROR(logger, "calendar name is empty, we skip it");
+        LOG4CPLUS_INFO(logger, "calendar name is empty, we skip it");
         return;
     }
     ed::types::Calendar* calendar = new ed::types::Calendar();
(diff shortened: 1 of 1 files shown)


source/ed/connectors/fusio_parser.cpp
```diff
@@ -693,18 +693,22 @@ void TripPropertiesFusioHandler::handle_line(Data&, const csv_row& row, bool is_
...
+    }
     try {
         return boost::gregorian::from_undelimited_string(str);
     } catch(const boost::bad_lexical_cast& ) {
-        LOG4CPLUS_ERROR(logger, "Impossible to parse the begin date for " << str);
+        LOG4CPLUS_ERROR(logger, "Impossible to parse the date for " << str);
     } catch(const boost::gregorian::bad_day_of_month&) {
-        LOG4CPLUS_ERROR(logger, "bad_day_of_month : Impossible to parse the begin date for " << str);
+        LOG4CPLUS_ERROR(logger, "bad_day_of_month : Impossible to parse the date for " << str);
     } catch(const boost::gregorian::bad_day_of_year&) {
```
kinnou02, line 706: It seem that all exceptions throw by `from_undelimited_string` inherit from `boost::lexical_cast` or `std::out_of_range`, it would be easier I think to only catch these two.

source/ed/connectors/fusio_parser.cpp
```diff
@@ -763,6 +767,11 @@ void GridCalendarFusioHandler::handle_line(Data& data, const csv_row& row, bool
                         "  file has more than one calendar and no id column");
         throw InvalidHeaders(csv.filename);
     }
+    if (has_col(id_c, row) && row[id_c].empty()) {
+        //we don't want empty named calendar as they often are empty lines in reality
+        LOG4CPLUS_ERROR(logger, "calendar name is empty, we skip it");
```
kinnou02: it's more like a warning if we only skip it.

source/ed/connectors/fusio_parser.cpp
antoine-de, line 706: sure but it add more precise logs. you think it useless ?

source/ed/connectors/fusio_parser.cpp
antoine-de: you're right, even INFO for this kind of message.

changed.

PR #385, opened 2014-06-09
Jormungandr: add param to force https in link

antoine-de:
for navitia.io to work behind reverse proxy we need to be able to force
https scheme in flask link (url_for).

To force this add the param

```
# Behind reverse proxy we might want to force https for the jormungandr
link
# default is False
FORCE_HTTPS = True
```

in the settings file.

### DIFF
--- a/source/jormungandr/jormungandr/__init__.py
+++ b/source/jormungandr/jormungandr/__init__.py
@@ -70,6 +70,10 @@ i_manager.initialisation(start_ping=app.config['START_MONITORING_THREAD'])
 from jormungandr.stat_manager import StatManager
 stat_manager = StatManager()
 
+from jormungandr.helpers import UrlHandler
+url_handler = UrlHandler(app.config.get('FORCE_HTTPS', False))
+
+
 from jormungandr import api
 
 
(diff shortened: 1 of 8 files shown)


source/jormungandr/jormungandr/index.py
```diff
@@ -28,7 +28,7 @@
 # www.navitia.io
 
 from flask.ext.restful import Resource
-from flask import url_for
+from jormungandr import url_handler
```
kinnou02, line 31: It would be easier to name it `url_for` so  we just have to change the import and not all the call.

kinnou02:
it's good for @guikcd ?

guikcd:
I need to test this in pre (patch seems to sucessfully apply to 0.97.0)

guikcd:
Ok to me (thanks nginx + snakeoil certificate + patch), tested on -pre.

l-vincent-l:
I think you can do it in a more simple way by copying: http://flask.pocoo.org/snippets/35/

antoine-de:
not sure to understand this but seems way better especially since it seems
to allows us to be able to return http or https link depending of the
caller (we don't want https for internal call).

antoine-de:
close, done better with #387

PR #387, opened 2014-06-11
Add a wsgi middleware for handling https behind a reverse proxy.

kinnou02:
The HTTP header `X-Scheme` can be set to override the scheme used by
jormungandr

Commits are pretty clean, you can review each one of them individually if you want :p


guikcd:
Looks good, maybe a test on scheme variable content more safe (?).

PR #388, opened 2014-06-11
jormungandr: Error handling in stat_manger for RabbitMQ

kadhikari:
When rabbitMQ is re-initialized and jormungandr is running,
jormungandr cannot push stat masage to rabbitMQ even if
the connection exists.

If it happens the connection will have to be re-initialized.

### DIFF
--- a/source/jormungandr/jormungandr/stat_manager.py
+++ b/source/jormungandr/jormungandr/stat_manager.py
@@ -339,15 +339,13 @@ class StatManager(object):
     def publish_request(self, stat_request):
         try:
             self.producer.publish(stat_request.SerializeToString())
-        except Exception as e:
+        except self.connection.connection_errors + self.connection.channel_errors:
             logging.getLogger(__name__).info('Server went away, will be reconnected..')
-
-            #Release the existing connection and re-initialize rabbitMQ
-            if self.connection.connect():
-                self.connection.release()
             #Initialize a new connection to RabbitMQ
             self._init_rabbitmq()
-            self.producer.publish(stat_request.SerializeToString())
+            #If connection is established publish the stat message.
+            if self.save_stat:
+                self.producer.publish(stat_request.SerializeToString())
 
     def fill_admin_from(self,stat_section, admins):
         for admin in admins:
--- a/source/jormungandr/jormungandr/stat_manager.py
+++ b/source/jormungandr/jormungandr/stat_manager.py
@@ -97,6 +97,7 @@ class StatManager(object):
             self.connection = kombu.Connection(self.broker_url)
             exchange = kombu.Exchange(self.exchange_name, type="direct")
             self.producer = self.connection.Producer(exchange=exchange)
+            self.save_stat = True
         except:
             self.save_stat = False
             logging.getLogger(__name__).warn('Unable to activate the producer of stat')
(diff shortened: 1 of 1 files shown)


source/jormungandr/jormungandr/stat_manager.py
```diff
@@ -336,8 +337,17 @@ def fill_sections(self, stat_journey, resp_journey):
 
 
     def publish_request(self, stat_request):
-        self.producer.publish(stat_request.SerializeToString())
+        try:
+            self.producer.publish(stat_request.SerializeToString())
+        except Exception as e:
```
kinnou02: too general

source/jormungandr/jormungandr/stat_manager.py
```diff
@@ -336,8 +337,17 @@ def fill_sections(self, stat_journey, resp_journey):
...
+            logging.getLogger(__name__).info('Server went away, will be reconnected..')
 
+            #Release the existing connection and re-initialize rabbitMQ
+            if self.connection.connect():
+                self.connection.release()
+            #Initialize a new connection to RabbitMQ
+            self._init_rabbitmq()
```
kinnou02, line 348: we can use pool: http://ask.github.io/kombu/userguide/pools.html

source/jormungandr/jormungandr/stat_manager.py
```diff
@@ -336,8 +337,17 @@ def fill_sections(self, stat_journey, resp_journey):
...
+            self.producer.publish(stat_request.SerializeToString())
```
kinnou02: if the connection fail before, this call raise a exception again

source/jormungandr/jormungandr/stat_manager.py
```diff
@@ -336,8 +337,17 @@ def fill_sections(self, stat_journey, resp_journey):
...
 
     def publish_request(self, stat_request):
-        self.producer.publish(stat_request.SerializeToString())
+        try:
+            self.producer.publish(stat_request.SerializeToString())
+        except Exception as e:
+            logging.getLogger(__name__).info('Server went away, will be reconnected..')
 
+            #Release the existing connection and re-initialize rabbitMQ
+            if self.connection.connect():
```
kinnou02: we connect so we can close the connection? why?

source/jormungandr/jormungandr/stat_manager.py
```diff
@@ -97,6 +97,7 @@ def _init_rabbitmq(self):
             self.connection = kombu.Connection(self.broker_url)
             exchange = kombu.Exchange(self.exchange_name, type="direct")
             self.producer = self.connection.Producer(exchange=exchange)
+            self.save_stat = True
```
kinnou02, line 100: `_init_rabbitmq` seem to be call only if save_stat is true

kadhikari, line 100: But if re-initialized, the variable save_stat has to be affected with "True" as false is affected by exception.

source/jormungandr/jormungandr/stat_manager.py
kadhikari: Infact, the connection still exists mais stat_mpanager is not able to push message to rabbitMQ because of error "Broken pipe()??"

kadhikari:
Alex,
i modified the code after your comments. I couldn't find answer to some of your comments.
Could you verify my answers and corrections  please?

source/jormungandr/jormungandr/stat_manager.py
```diff
@@ -336,8 +337,15 @@ def fill_sections(self, stat_journey, resp_journey):
 
 
     def publish_request(self, stat_request):
-        self.producer.publish(stat_request.SerializeToString())
-
+        try:
+            self.producer.publish(stat_request.SerializeToString())
+        except self.connection.connection_errors + self.connection.channel_errors:
+            logging.getLogger(__name__).info('Server went away, will be reconnected..')
```
kinnou02:
```
 logging.getLogger(__name__).exception('Server went away, will be reconnected..')
```

this way we have the message of the exception and the stacktrace

source/jormungandr/jormungandr/stat_manager.py
```diff
@@ -336,8 +337,15 @@ def fill_sections(self, stat_journey, resp_journey):
...
+            #Initialize a new connection to RabbitMQ
+            self._init_rabbitmq()
```
kinnou02, line 348: we have to release/close the previous connection, if not we may leak handles

kadhikari:
Alex,
i corrected for your last two comments.

PR #391, opened 2014-06-12
Manage odt

azime:
- Ajout d'un nouveau type enum pour gérer les niveaux ODT
- Remplacement de la variable is_odt dans journeypattern par OdtLevel
- Refactorin dans ptref
- Gestion des niveaux ODT de la ligne dans ptref


source/jormungandr/jormungandr/interfaces/v1/Uri.py
```diff
@@ -77,7 +83,12 @@ def __init__(self, is_collection, collection, *args, **kwargs):
 
     def get(self, region=None, lon=None, lat=None, uri=None, id=None):
         collection = self.collection
+
         args = self.parsers["get"].parse_args()
+
+        if "odt_level" in args and args["odt_level"] != "all" and "lines" not in collection:
+            abort(404, message="bad request")
```
antoine-de: You should maybe put a more precise error message

source/ptreferential/ptreferential.h
```diff
@@ -100,6 +100,12 @@ struct parsing_error : public ptref_error{
 std::vector<type::idx_t> make_query(type::Type_e requested_type,
                                     std::string request,
                                     const std::vector<std::string>& forbidden_uris,
+                                    const type::OdtLevel_e odt_level,
+                                    const type::Data &data);
+
+std::vector<type::idx_t> make_query(type::Type_e requested_type,
+                                    std::string request,
```
antoine-de, line 107: you can put a default odl_level value can't you ?

source/ptreferential/ptreferential.cpp
```diff
@@ -195,8 +195,70 @@ std::vector<Filter> parse(std::string request){
...
+                case navitia::type::OdtLevel_e::zonal:
+                    if (line->get_odt_level() == navitia::type::OdtLevel_e::zonal){
+                        forbidden_idx.push_back(idx);
+                    };
+                    break;
+                case navitia::type::OdtLevel_e::all:
+                    break;
+            }
+        }
+        final_indexes = get_intersection(final_indexes, forbidden_idx);
```
antoine-de: I don't understand, what do you want to do ? you want only line that match the filter ? why not only return a list so ? I don't think the intersection is very usefull. (and the name forbidden_idx does not seems the fit)

azime:
corrections: 
- manage_odt function : return list of idx
- error message in ptref

azime:
refactoring ptref : modification of sort_and_unique function

source/jormungandr/jormungandr/interfaces/v1/Uri.py
```diff
@@ -77,7 +83,12 @@ def __init__(self, is_collection, collection, *args, **kwargs):
 
     def get(self, region=None, lon=None, lat=None, uri=None, id=None):
         collection = self.collection
+
         args = self.parsers["get"].parse_args()
+
+        if "odt_level" in args and args["odt_level"] != "all" and "lines" not in collection:
+            abort(404, message="bad request: odt_level is not valid argument")
```
antoine-de: sorry to be picky, but could you change that to:
"bad request: odt_level filter can only be applied to lines"
?

azime:
jormungandr : message odt level

PR #402, opened 2014-06-19
Catch zmq socket_t::bind() failures on startup

guikcd:
Without this zmq socket_t::bind() failure is printed on stdout (and
difficult to track when started in background and with initscript)

Feel free to manage differently, just want to have the error message in logs in case of startup failure (EADDRINUSE).


buildbotctp:

kinnou02:
ok to test.

source/kraken/kraken_zmq.cpp
```diff
@@ -60,29 +60,35 @@ int main(int, char** argv){
...
     zmq::socket_t clients(context, ZMQ_ROUTER);
     std::string zmq_socket = conf->get_as<std::string>("GENERAL", "zmq_socket", "ipc:///tmp/default_navitia");
-    clients.bind(zmq_socket.c_str());
-    zmq::socket_t workers(context, ZMQ_DEALER);
-    workers.bind("inproc://workers");
+    // Catch startup exceptions; without this, startup errors are on stdout
+    try{
+        clients.bind(zmq_socket.c_str());
+        zmq::socket_t workers(context, ZMQ_DEALER);
+        workers.bind("inproc://workers");
```
kinnou02, line 73: Only this 3 lines need to be in the try/catch, others don't throw `zmq::error_t` since they don't deal with zmq

source/kraken/kraken_zmq.cpp
```diff
@@ -60,29 +60,35 @@ int main(int, char** argv){
...
-        }catch(zmq::error_t){}//lors d'un SIGHUP on restore la queue
-    }while(true);
+        // Connect work threads to client threads via a queue
+        do{
+            try{
+                zmq::device(ZMQ_QUEUE, clients, workers);
+            }catch(zmq::error_t){}//lors d'un SIGHUP on restore la queue
+        }while(true);
+    }catch(zmq::error_t& e){
+        LOG4CPLUS_ERROR(logger, "Fail to start: " << e.what());
```
antoine-de: for the moment we do not use that, but wouldn't that be better to return an error code != 0 ?

source/kraken/kraken_zmq.cpp
```diff
@@ -60,29 +60,35 @@ int main(int, char** argv){
...
-    do{
-        try{
-            zmq::device(ZMQ_QUEUE, clients, workers);
-        }catch(zmq::error_t){}//lors d'un SIGHUP on restore la queue
-    }while(true);
+        // Connect work threads to client threads via a queue
+        do{
+            try{
+                zmq::device(ZMQ_QUEUE, clients, workers);
+            }catch(zmq::error_t){}//lors d'un SIGHUP on restore la queue
```
antoine-de: shouldn't we log those ?

guikcd:
I've updated the PR with your advices.

antoine-de:
great thanks !

PR #420, opened 2014-07-01
Add stop_area, stop_point to forbidden_uri

l-vincent-l:
Wait before merging this PR, it needs some benchmark.

There's no incidence on performances, so it can be merged.

### DIFF
--- a/source/routing/raptor.cpp
+++ b/source/routing/raptor.cpp
@@ -182,18 +182,18 @@ void RAPTOR::init(Solutions departs,
         if(sp->accessible(required_properties)) {
             for(auto journey_pattern_point : sp->journey_pattern_point_list) {
                 type::idx_t jpp_idx = journey_pattern_point->idx;
-                if(journey_patterns_valides.test(journey_pattern_point->journey_pattern->idx)) {
-                        b_dest.add_destination(jpp_idx, item.second, clockwise);
-                        best_labels[jpp_idx] = clockwise ?
+                if(valid_journey_patterns.test(journey_pattern_point->journey_pattern->idx)) {
+                    b_dest.add_destination(jpp_idx, item.second, clockwise);
+                    best_labels[jpp_idx] = clockwise ?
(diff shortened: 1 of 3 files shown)


source/routing/tests/raptor_test.cpp
```diff
@@ -446,6 +446,25 @@ BOOST_AUTO_TEST_CASE(validity_pattern){
 }
 */
 
+
+BOOST_AUTO_TEST_CASE(forbidden_uri){
```
antoine-de, line 450: un p'tit commentaire siouplait

source/routing/raptor.cpp
```diff
@@ -288,7 +288,21 @@ void RAPTOR::set_journey_patterns_valides(uint32_t date, const std::vector<std::
             }
         }
     }
-    journey_patterns_valides &= ~forbidden_journey_patterns;
+    valid_journey_patterns &= ~forbidden_journey_patterns;
+
+    for(const type::JourneyPatternPoint* jpp: data.pt_data->journey_pattern_points) {
```
antoine-de: wouldn't it be better to loop first on the forbidden (then on all jpp), and hence not to have any overhead if no forbidden is given?

l-vincent-l:
@antoine-de I have taken in account your remarks

source/routing/raptor.cpp
```diff
@@ -182,7 +182,7 @@ void RAPTOR::init(Solutions departs,
         if(sp->accessible(required_properties)) {
             for(auto journey_pattern_point : sp->journey_pattern_point_list) {
                 type::idx_t jppidx = journey_pattern_point->idx;
-                if(journey_patterns_valides.test(journey_pattern_point->journey_pattern->idx)) {
+                if(valid_journey_patterns.test(journey_pattern_point->journey_pattern->idx)) {
```
l-vincent-l: I need to add valid_journey_pattern_points

PR #427, opened 2014-07-03
Stay in fixes

koch-t:
Fixes SQL query error.
Fixes stay_in's where vehicle has more than 120 seconds between two interline sections.


buildbotctp:

antoine-de:
Jenkins, test this please

koch-t:
My feeling is that the change in raptor_path is not the fix but covering the real bug. Why is the vehicle extension overruled by a non-direct Journey of equal lenght?

koch-t:
ok to test

antoine-de:
ok to test

koch-t:
Now the reversal errors are taken care of in #431, i feel a little bit more comfortable about this PR

l-vincent-l:
@skywave do you think we can merge this branch?
We know the problem is not on your branch but on #431 (and this branch is not mergeable yet, I haven't find why it's buggy).

koch-t:
This branch is mergeable.

PR #429, opened 2014-07-13
Use constants instead of double negation

koch-t:
Due to the if(clockwise) !clockwise and clockwise were already constants.


buildbotctp:

antoine-de:
ok to test

PR #430, opened 2014-07-13
Assert on reversal errors

koch-t:
If there are no solutions found at this point, a bug is present causing unsymetric searches in clockwise vs counterclockwise. If no paths exists, this assert will never be reached anyway after the invalid_idx check

--Do not merge as such a bug is present --


buildbotctp:

antoine-de:
ok to test

PR #431, opened 2014-07-15
Fix 2nd pass bug

l-vincent-l:
There was a bug in the second pass with this.
Some journeys were not able to be compute. This solves it, but I can't really understand why it works.


koch-t:
I think the common pattern between journeys that failed: journeys with the same departuretime (same trip in fact) but differing amount of transfers/travel-length.

koch-t:
Should be closed after #478 ?

l-vincent-l:
It might, i still have to do tests, i noticed strange behaviours

antoine-de:
can you rebase it ?

Is it still up to date ?

l-vincent-l:
it isn't I close it

PR #432, opened 2014-07-15
Add debug tools

koch-t:
This PR adds a tool to query RAPTOR standalone from the shell, allowing for easier debugging without having to start kraken and jormungandr
It also adds functionality to benchmark to check the last request made, useful if it crashes somewhere.


buildbotctp:

antoine-de:
ok to test

PR #433, opened 2014-07-16
Tyr: do not run load_data async

patochectp:
since artemis is the only one the call that target, it's easier to wait for the answer.


buildbotctp:

source/tyr/tyr/command/load_data.py
```diff
@@ -31,6 +31,7 @@
 from navitiacommon import models
 from tyr.tasks import load_data
 import logging
+import sys
```
kinnou02: not needed

PR #435, opened 2014-07-17
Doc install

antoine-de:
add an install script that:
- build navitia
- import data (by default it download a Paris dataset)
- run a front end and a kraken core

it is made more as an installation guide, but can be used as an out of the box build script

Note: I don't understand why the kraken is not killed at the end of the run.
Anybody has an idea ?

### DIFF
--- a/build_navitia.sh
+++ b/build_navitia.sh
@@ -15,14 +15,55 @@
 
 navitia_du_user_password='navitia' #TODO recuperer un param
 navitia_dir=`pwd $0`
-gtfs_data_dir=$1  #TODO recup un param
-osm_file=$2  #TODO recup un param
+gtfs_data_dir=
+osm_file=
 
 install_dependencies=1
 
+usage()
+{
+cat << EOF
+usage: $0 options
+
+This script setup a running navitia
+No options are mandatory:
+ - if no dataset are given a default Paris one will be used
+ - by default all dependencies are installed
+
+OPTIONS:
+   -h                  Show this message
+   -g                  gtfs directory 
+   -o                  osm directory
+   -n                  don't install dependencies
+EOF
+}
+
+while getopts “hg:o:n” OPTION
+do
+     case $OPTION in
+         h)
+             usage
+             exit 1
+             ;;
+         g)
+             gtfs_data_dir=$OPTARG
+             ;;
+         o)
+             osm_file=$OPTARG
+             ;;
+         n)
+             install_dependencies=0
+             ;;
+         ?)
+             usage
+             exit
+             ;;
+     esac
+done
+
 if [ -z "$gtfs_data_dir" ] || [ -z "$osm_file" ]
 then
-    echo "no gtfs or osm file given, we'll take a test data set"
+    echo "no gtfs or osm file given, we'll take a default data set, Paris"
 
     echo "getting gtfs paris data from data.navitia.io"
     wget -P /tmp http://data.navitia.io/gtfs_paris_20140502.zip
--- a/build_navitia.sh
+++ b/build_navitia.sh
@@ -11,10 +11,17 @@
 #
 # It's more an install guide but it can help as an out-of-a-box build script
 # the prerequisite the run that script is to have cloned the sources repository
+#
 # git clone https://github.com/CanalTP/navitia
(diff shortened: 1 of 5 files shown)


antoine-de:
Jenkins, test this please

build_navitia.sh
```diff
@@ -0,0 +1,255 @@
...
+sed -i 's,git\@github.com:\([^/]*\)/\(.*\).git,https://github.com/\1/\2,' .gitmodules
+
+#we need to get the submodules
+git submodule update --init
+
+#========
+#Building
+#========
+#
+#First you need to install all dependecies.
```
kinnou02: typo

build_navitia.sh
```diff
@@ -0,0 +1,255 @@
...
+#
+#Each kraken is backed by a postgres database
+#
+#You need to set up the database
+kraken_db_name='navitia'
+db_owner='navitia'
+
+kraken_db_name='navitia'
+# for the default build we give ownership of the base to a 'navitia' user, but you can do whatever you want here
+sudo su postgres -c "psql postgres -tAc \"SELECT 1 FROM pg_roles WHERE rolname='$db_owner'\""  # we check if there is already a user
```
kinnou02, line 169: i don't really like the sudo in a script, because I always need to read it before entering my password.
yes, I am a little paranoid

build_navitia.sh
```diff
@@ -0,0 +1,255 @@
...
+#You need to set up the database
+kraken_db_name='navitia'
+db_owner='navitia'
+
+kraken_db_name='navitia'
+# for the default build we give ownership of the base to a 'navitia' user, but you can do whatever you want here
+sudo su postgres -c "psql postgres -tAc \"SELECT 1 FROM pg_roles WHERE rolname='$db_owner'\""  # we check if there is already a user 
+if [ ! $? ]
+then
+sudo su postgres -c "psql -c \"create user $db_owner;alter user $db_owner password '$kraken_db_user_password';\""
```
kinnou02, line 172: the `createuser` shell command would probably be easier. there is a `createdb` too

build_navitia.sh
```diff
@@ -0,0 +1,255 @@
...
+#ipc socket in default.ini file in the jormungandr instances dir
+zmq_socket = ipc:///tmp/default_kraken
+#number of threads
+nb_threads = 1
+#name of the instance
+instance_name=default
+[LOG]
+log4cplus.rootLogger=DEBUG, ALL_MSGS
+
+log4cplus.appender.ALL_MSGS=log4cplus::FileAppender
```
kinnou02, line 230: for testing purpose it can be easier to log on stdout

build_navitia.sh
```diff
@@ -0,0 +1,255 @@
...
+then
+    echo "** instaling all dependencies"
+    sudo apt-get install -y git g++ cmake liblog4cplus-dev libzmq-dev libosmpbf-dev libboost-all-dev libpqxx3-dev libgoogle-perftools-dev libprotobuf-dev python-pip libproj-dev protobuf-compiler libgeos-c1 
+    
+    sudo apt-get install -y postgresql-9.3 postgresql-9.3-postgis-2.1 #Note: postgres 9.1 and postgis 2.0 would be enough, be postgis 2.1 is easier ton setup 
+    
+    # then you need to install all python dependencies: ::
+    
+    sudo pip install -r $navitia_dir/source/jormungandr/requirements.txt
+    sudo pip install -r $navitia_dir/source/tyr/requirements.txt
```
kinnou02, line 145: Tyr is not used

build_navitia.sh
antoine-de, line 230: sure, but since Jormun and kraken are lanched in the same shell the output can become quite messy so I thought It would be better to output Jormungandr and kraken in files

build_navitia.sh
antoine-de, line 169: I do agree, but then how can you manage the postgres stuff ?

build_navitia.sh
antoine-de, line 172: you cant give easily a password with createuser, so the psql  stuff is easier.

Ok for the createdb

PR #437, opened 2014-07-18
Autocomplete test functional init

kadhikari:
Test functional for places initialized

### DIFF
--- a/source/autocomplete/autocomplete_api.cpp
+++ b/source/autocomplete/autocomplete_api.cpp
@@ -32,6 +32,7 @@ www.navitia.io
 #include "type/pb_converter.h"
 #include "autocomplete/autocomplete.h"
 #include "type/pt_data.h"
+#include "string.h"
 
 namespace navitia { namespace autocomplete {
 /**
@@ -351,7 +352,7 @@ pbnavitia::Response autocomplete(const std::string &q,
             const auto b_order = get_embedded_type_order(b.embedded_type());
             return  a_order< b_order;
         } else if(a.quality() == b.quality()) {
-            return a.name() < b.name();
+                return boost::algorithm::lexicographical_compare(a.name(), b.name(), boost::is_iless());
         }
         else {
             return a.quality() > b.quality();
--- a/source/autocomplete/autocomplete_api.cpp
+++ b/source/autocomplete/autocomplete_api.cpp
@@ -32,7 +32,6 @@ www.navitia.io
 #include "type/pb_converter.h"
 #include "autocomplete/autocomplete.h"
 #include "type/pt_data.h"
-#include "string.h"
 
 namespace navitia { namespace autocomplete {
 /**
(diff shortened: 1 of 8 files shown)


source/autocomplete/autocomplete_api.cpp
```diff
@@ -32,6 +32,7 @@ www.navitia.io
 #include "type/pb_converter.h"
 #include "autocomplete/autocomplete.h"
 #include "type/pt_data.h"
+#include "string.h"
```
antoine-de: the convention is to include outside stuff with '<>'

you are sure you need this include ?

source/autocomplete/autocomplete_api.cpp
```diff
@@ -351,7 +352,7 @@ pbnavitia::Response autocomplete(const std::string &q,
             const auto b_order = get_embedded_type_order(b.embedded_type());
             return  a_order< b_order;
         } else if(a.quality() == b.quality()) {
-            return a.name() < b.name();
+                return boost::algorithm::lexicographical_compare(a.name(), b.name(), boost::is_iless());
```
antoine-de: indentation problem

source/ed/build_helper.cpp
```diff
@@ -387,4 +387,24 @@ void builder::connection(const std::string & name1, const std::string & name2, f
          }
      }
  }
+
+ void builder::manage_admin() {
```
antoine-de, line 395: put a comment here to explain what you are doing.
(eg that only the first admin is used)

source/georef/georef.cpp
```diff
@@ -241,6 +241,16 @@ void GeoRef::add_way(const Way& w){
     ways.push_back(to_add);
 }
 
+void GeoRef::add_admin(const Admin& a){
```
antoine-de: If it is only used for test purposes, it's better to move this function in the tests (a free function for example).

That being said, I'm not sure this function is really useful. If you want it to be syntaxic sugar, I think I would be more convenient to put the parameter.

ex:
in the test:

void add_admin(const std::string& name, ...level, .. post_code) {
Admin\* to_add = new Admin;
 to_add->name = name;
 to_add->uri = a.name;
 to_add->post_code = post_code;
 to_add->level = level;
to_add->idx = admins.size();
 admins.push_back(to_add);
}

antoine-de:
Thanks for adding lots of tests!

source/autocomplete/autocomplete_api.cpp
kadhikari: deleted

source/autocomplete/autocomplete_api.cpp
kadhikari: ok

source/ed/build_helper.cpp
kadhikari, line 395: comments added

source/georef/georef.cpp
kadhikari: Like add_way, add_housenumber we can certainly use add_admin in other functions.

kadhikari:
Antoine, i posted some corrections

kadhikari:
I deleted add_admin in georef

PR #438, opened 2014-07-18
Cli: factorize loading

l-vincent-l:


source/routing/routing_cli_utils.cpp
```diff
@@ -42,20 +42,28 @@ namespace pb = pbnavitia;
...
 
-        bool compute_options::compute(const boost::program_options::variables_map& vm,
-                nr::RAPTOR& raptor) {
+        void compute_options::load(const std::string& file) {
+            {
+                Timer t("Loading datafile : " + file);
+                data.load(file);
+            }
+            data.build_raptor();
+            raptor = new nr::RAPTOR(data);
```
antoine-de: why do you use a pointer here ?

antoine-de: std::unique_ptr whould be better

Issue #425, opened 2014-07-03
Use stop_area on first/last section of street_network


koch-t:
Currently a geocoded name is used as the starting/end point of any streetnetwork section. 
If for the request from stop_area:orig to stop_area_dest the street_network is used from stop_area:orig to stop_area:boarding then it would be desirable to also use the names of those two stop_area's for the from/to in the section 
Same applies for the last section

koch-t:
Fix with #441

PR #441, opened 2014-07-18
Add coherence to sections

l-vincent-l:
**Add coherence in journey' sections**

We want to have coherence between the from (to) of a section and the from (to) of the request.
It also adds coherence between the to (from) of the streetnetwork section and the  from(to) of the first (last) public transport section. Therefore the to(from) of the streetnetwork section is now a stop_point, you still have the adress of the stop point in the place.

We also don't really know from where we start when we start from a area (a stop_area or an admin), thus we added a crow_fly section which starts from the centroid of this area and goes to the first stop_point of the first pt_section.

### DIFF
--- a/source/routing/raptor_api.cpp
+++ b/source/routing/raptor_api.cpp
@@ -130,7 +130,7 @@ pbnavitia::Response make_pathes(const std::vector<navitia::routing::Path>& paths
             // we indicate a time.
             if (origin.type == type::Type_e::StopArea && origin.uri != sp_dest->stop_area->uri) {
                 const auto duration = worker.get_distance(sp_dest->idx);
-                if (!duration.is_pos_infinity() && !duration.is_neg_infinity() && !duration.is_not_a_date_time()) {
+                if (!duration.is_special()) {
                     auto* first_section = pb_journey->mutable_sections(0);
                     first_section->set_duration(duration.total_seconds());
                     first_section->mutable_street_network()->set_mode(convert(origin.streetnetwork_params.mode));
@@ -286,7 +286,7 @@ pbnavitia::Response make_pathes(const std::vector<navitia::routing::Path>& paths
             if (destination.type == type::Type_e::StopArea &&
                     destination.uri != sp_orig->stop_area->uri && pb_journey->sections_size() > 0) {
                 auto duration = worker.get_distance(sp_orig->idx, true);
-                if (!duration.is_pos_infinity() && !duration.is_neg_infinity() && !duration.is_not_a_date_time()) {
+                if (!duration.is_special()) {
                     auto* last_section = pb_journey->mutable_sections(pb_journey->sections_size()-1);
                     last_section->set_duration(duration.total_seconds());
                     last_section->mutable_street_network()->set_mode(convert(destination.streetnetwork_params.mode));
--- a/source/routing/raptor_api.cpp
+++ b/source/routing/raptor_api.cpp
@@ -111,7 +111,7 @@ pbnavitia::Response make_pathes(const std::vector<navitia::routing::Path>& paths
          * 1) We start from an area, we will add a crow fly section from the centroid of the area
          *    to the origin stop point of the first section
          * 2) We start from a ponctual place (everything but stop_area or admin)
-         * We had a street network section from this place to the departure of the first pt_section
+         * We add a street network section from this place to the departure of the first pt_section
          * If the uri of the origin point and the uri of the departure of the first section are the
          * same we do nothing
          * */
@@ -126,7 +126,7 @@ pbnavitia::Response make_pathes(const std::vector<navitia::routing::Path>& paths
                                           path.items.front().departures.front()+bt::minutes(1));
             fill_crowfly_section(origin, destination_tmp, path.items.front().departures.front(),
                                  d, enhanced_response, pb_journey, now, action_period);
-            // If we start from a stop_area, and the last stop point is not from that stop_area,
+            // If we start from a stop_area, and the first stop point is not from that stop_area,
             // we indicate a time.
             if (origin.type == type::Type_e::StopArea && origin.uri != sp_dest->stop_area->uri) {
                 const auto duration = worker.get_distance(sp_dest->idx);
@@ -157,14 +157,13 @@ pbnavitia::Response make_pathes(const std::vector<navitia::routing::Path>& paths
                     auto section = pb_journey->mutable_sections(pb_journey->mutable_sections()->size()-1);
                     bt::time_period action_period(boost::posix_time::from_iso_string(section->begin_date_time()),
                                                   boost::posix_time::from_iso_string(section->end_date_time()));
-                    // We had coherence with the origin of the request
+                    // We add coherence with the origin of the request
                     fill_pb_placemark(origin, d, section->mutable_origin(), 2, now, action_period, show_codes);
-                    // We had coherence with the first pt section
+                    // We add coherence with the first pt section
                     fill_pb_placemark(departure_stop_point, d, section->mutable_destination(), 2, now, action_period, show_codes);
                 }
             }
         }
-        // Here we make origin and departure consistent with the parameters and the first section
 
         const type::VehicleJourney* vj(nullptr);
         size_t item_idx(0);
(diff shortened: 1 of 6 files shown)


antoine-de:
@skywave, It should handle #425 

Is it ok with you ?

source/routing/raptor_api.cpp
```diff
@@ -103,12 +103,33 @@ pbnavitia::Response make_pathes(const std::vector<navitia::routing::Path>& paths
...
-            const auto& departure_stop_point = path.items.front().stop_points.front();
-            // for stop areas, we don't want to display the fallback section if start
-            // from one of the stop area's stop point
-            if (origin.uri != departure_stop_point->stop_area->uri) {
+        if (path.items.empty()) {
+            continue;
+        }
+        /*
+         * For the first section, we can distinguish 2 cases
+         * 1) We start from an area, we will add a crow fly section from the centroid of the area
```
antoine-de, line 111: It should be a bit more precise than 'area', say it is ok for stop_area and admin

source/routing/raptor_api.cpp
```diff
@@ -103,12 +103,33 @@ pbnavitia::Response make_pathes(const std::vector<navitia::routing::Path>& paths
...
+         * */
+
+        if (origin.type == type::Type_e::Admin || origin.type == type::Type_e::StopArea) {
+            if (path.items.front().stop_times.empty()) {
+                continue;
+            }
+            const auto sp_dest = path.items.front().stop_times.front()->journey_pattern_point->stop_point;
+            type::EntryPoint destination(type::Type_e::StopPoint, sp_dest->uri);
+            bt::time_period action_period(path.items.front().departures.front(),
+                                          path.items.front().departures.front()+bt::minutes(1));
```
antoine-de, line 126: why add 1mn ?

koch-t:
Two things, with crow_fly you can't distinguish  walking from bicycle, car, etc. Durations are 0 seconds also?

koch-t:
You now also get crow_fly's if you request from/to the stop_area you are also actually departing from. Something previously turned off explicitly

l-vincent-l:
Crowfly sections are informative, and not precise, you don't really know from where you start, so you don't have a walking or bicycle (actually you're traveling on a crow, how cool is it?!?), so can't really put a duration, maybe we could put an indicative one.

I'm not sure on what to do with your second comment, the thing is it would be easier for web integrator to always have stop_point in public transport section. Also if I put a stop_area in the pt_section and no crow_fly before, you won't have the information about the stop_point you're starting by, and I think you want that information.

koch-t:
There are also situations where you start from Stop A, but it's much faster to start from Station C which is 10 minutes of walking from A. 
With the current code Navitia makes a crow_fly of 0 seconds for that, also there is no longer a instructions/geometry for the walking/biking/car path

l-vincent-l:
Yep, I need to put a duration for this case.

l-vincent-l:
@skywave There's now a duration when we start/go to a stop_area and the first/last stop point of the next/previous pt section is not one of the one of the stop_area.

koch-t:
I still think the mode (walking/bike/car) is missing depending on the first/last section_mode. And i think it would still be a good idea to include geometry/instructions for this section. Except when it is the same stop_area ofcourse.

l-vincent-l:
Jenkins, test this please

source/routing/raptor_api.cpp
```diff
@@ -103,15 +103,47 @@ pbnavitia::Response make_pathes(const std::vector<navitia::routing::Path>& paths
...
+            if (path.items.front().stop_times.empty()) {
+                continue;
+            }
+            const auto sp_dest = path.items.front().stop_times.front()->journey_pattern_point->stop_point;
+            type::EntryPoint destination_tmp(type::Type_e::StopPoint, sp_dest->uri);
+            bt::time_period action_period(path.items.front().departures.front(),
+                                          path.items.front().departures.front()+bt::minutes(1));
+            fill_crowfly_section(origin, destination_tmp, path.items.front().departures.front(),
+                                 d, enhanced_response, pb_journey, now, action_period);
+            // If we start from a stop_area, and the last stop point is not from that stop_area,
```
antoine-de: first stop point

source/routing/raptor_api.cpp
```diff
@@ -103,15 +103,47 @@ pbnavitia::Response make_pathes(const std::vector<navitia::routing::Path>& paths
...
+            // we indicate a time.
+            if (origin.type == type::Type_e::StopArea && origin.uri != sp_dest->stop_area->uri) {
+                const auto duration = worker.get_distance(sp_dest->idx);
+                if (!duration.is_pos_infinity() && !duration.is_neg_infinity() && !duration.is_not_a_date_time()) {
```
antoine-de: ya une fonction boost pour ca: is_special()

source/routing/raptor_api.cpp
```diff
@@ -120,10 +152,19 @@ pbnavitia::Response make_pathes(const std::vector<navitia::routing::Path>& paths
 
                     const auto walking_time = temp.duration;
                     departure_time = path.items.front().departure - walking_time.to_posix();
-                    fill_street_sections(enhanced_response, origin, temp, d, pb_journey, departure_time);
+                    fill_street_sections(enhanced_response, origin, temp, d, pb_journey,
+                        departure_time);
+                    auto section = pb_journey->mutable_sections(pb_journey->mutable_sections()->size()-1);
+                    bt::time_period action_period(boost::posix_time::from_iso_string(section->begin_date_time()),
+                                                  boost::posix_time::from_iso_string(section->end_date_time()));
+                    // We had coherence with the origin of the request
```
antoine-de: we add ?

source/routing/raptor_api.cpp
```diff
@@ -120,10 +152,19 @@ pbnavitia::Response make_pathes(const std::vector<navitia::routing::Path>& paths
...
+                    fill_pb_placemark(origin, d, section->mutable_origin(), 2, now, action_period, show_codes);
+                    // We had coherence with the first pt section
```
antoine-de: same

source/routing/raptor_api.cpp
```diff
@@ -120,10 +152,19 @@ pbnavitia::Response make_pathes(const std::vector<navitia::routing::Path>& paths
...
+                    fill_pb_placemark(departure_stop_point, d, section->mutable_destination(), 2, now, action_period, show_codes);
                 }
             }
         }
+        // Here we make origin and departure consistent with the parameters and the first section
```
antoine-de: remove comment ?

source/routing/raptor_api.cpp
```diff
@@ -230,24 +271,54 @@ pbnavitia::Response make_pathes(const std::vector<navitia::routing::Path>& paths
...
+            bt::time_period action_period(path.items.back().departures.back(),
+                                          path.items.back().departures.back()+bt::minutes(1));
+            fill_crowfly_section(origin_tmp, destination,path.items.back().departures.back(),
+                                 d, enhanced_response, pb_journey, now, action_period);
+            // If we go to a stop_area, and the last stop point is not from that stop_area,
+            // we indicate a time.
+            if (destination.type == type::Type_e::StopArea &&
+                    destination.uri != sp_orig->stop_area->uri && pb_journey->sections_size() > 0) {
+                auto duration = worker.get_distance(sp_orig->idx, true);
+                if (!duration.is_pos_infinity() && !duration.is_neg_infinity() && !duration.is_not_a_date_time()) {
```
antoine-de: same is special

source/type/pb_converter.cpp
```diff
@@ -1257,4 +1304,15 @@ void fill_pb_object(const std::string comment, const nt::Data&,
...
 }
+
+pbnavitia::StreetNetworkMode convert(const navitia::type::Mode_e& mode) {
+    switch (mode) {
+        case navitia::type::Mode_e::Walking : return pbnavitia::Walking;
+        case navitia::type::Mode_e::Bike : return pbnavitia::Bike;
+        case navitia::type::Mode_e::Car : return pbnavitia::Car;
+        case navitia::type::Mode_e::Bss : return pbnavitia::Bss;
+    }
+    return pbnavitia::Walking;
```
antoine-de: throw no ?

PR #443, opened 2014-07-19
Move frequency paramaters to VJ

koch-t:
-Move frequency parameters from stop_time to vehicle_journey, save at least 3\* uint32_t per vehicle_journey
-Enforce arrival/departure time in stop_time to be relative from 0
-Calculate low/upper threshold using relative times and VJ frequency paramaters
-Reduces size of stop_time with 8 bytes, increases size of VJ by 16 byes.

IMPORTANT: fails best_stop_time test, but i have a feeling the test is wrong there, please verify.

### DIFF
--- a/source/ed/build_helper.cpp
+++ b/source/ed/build_helper.cpp
@@ -36,17 +36,24 @@ namespace pt = boost::posix_time;
 namespace ed {
 
 VJ & VJ::frequency(uint32_t start_time, uint32_t end_time, uint32_t headway_secs) {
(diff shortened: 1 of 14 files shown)


buildbotctp:

antoine-de:
ok to test

koch-t:
The last commit reduces the size of VJ to 304bytes instead of 312 bytes.
Without the frequency paramters the size of VJ was 296 bytes.

PR #445, opened 2014-07-20
Use std::fill instead of assign

koch-t:
See http://stackoverflow.com/posts/8849789/revisions, with -03 on std::assign appears to be slower than std:fill
This patch results in unscientifically tested 1-2% speedup on a dataset with 306776 journey_pattern_point's


buildbotctp:

kinnou02:
ok to test

l-vincent-l:
Thank you for you hard work!

PR #446, opened 2014-07-22
Remove unused members of JourneyPatternPoint

koch-t:
Removed seemingly unused members in JourneyPatternPoint struct
Reorder members
Reduce order from int to uint16_t (65k stops in VJ should be more than enough)
Reduce sizeof from 48 to 40 bytes


buildbotctp:

l-vincent-l:
ok to test

PR #421, opened 2014-07-02
Time zone

antoine-de:
## PR for time zone handling

The main idea is to convert ALL times to UTC  in the import process, handle only UTC in kraken (the computation part) and change it to local time only at the end of the process in Jörmungandr (the middleware) using the added time zone info in each stop areas.

The UTC conversion difficult part is day saving time (DST) handling.  
For example a stoptime at 7am valid over the year in a paris (Europe/Paris timezone) will be 6AM UTC in winter but 5AM UTC in summer.

Thus all validity_pattern (gtfs calendar), vehicule_journey (gtfs trips) and stoptime have to be split over the DST period.
## Note

despite all its flaws we use boost::date_time for timzeone handling.
for the moment the boost csv is hardlinked in the project, but it will change (maybe by patching boost for it to read the  tzdb).

PR #456, opened 2014-07-25
Stop headsigns

koch-t:
This PR establishes three things
- Move all unnecessary attributes of VJ to dynamically allocated struct,  reduces size of VJ from 304 bytes to 176 bytes. A further reduce to 128 bytes is possible when the std::map (48 bytes) for calendar_id is removed.
- Explicitly make a field for headsign instead of using the name field. Name field is now used for the name of a vj, for example a train-number
  -Use a dynamically allocated struct with headsign and comment to allow a StopTime to override the VJ-headsign


source/type/type.h
```diff
@@ -691,28 +704,40 @@ struct VehicleJourney: public Header, Nameable, hasVehicleProperties, HasMessage
     // They have the same block id
     VehicleJourney* next_vj = nullptr;
     VehicleJourney* prev_vj = nullptr;
-    std::string odt_message;
+
     ///map of the calendars that nearly match the validity pattern of the vj, key is the calendar name
     std::map<std::string, AssociatedCalendar*> associated_calendars;
 
+    std::unique_ptr<VehicleJourneyInformation> information;
```
antoine-de, line 711: I don't see the point of adding an additional layer of indirection for this since we use the VJ only as dynamically allocated objects.

What are the rationals behind this ?

koch-t, line 711: The idea was that the for the router less memory has to be pulled to check validity_patterns and vehicle_properties. But now i have to admit i kinda looked over the fact that all VJ's are dynamically allocated.

antoine-de, line 711: You're right about the stop times though, because we'll have to change the dynamic allocation for them soon.

That being said, for the stop times I think we have to move the `comment` into a separate map (since the comment are not on all stop times and are used only for output.
only the head sign remain thus, so I'm not sure the unique_ptr indirection for a unique string is useful.

koch-t:
It is a headsign and comment per stop_time. std:map could be a solution but i am not sure how that will play out with adapted VJ's

antoine-de:
we're refactoring the adapted stuff.

we added a meta_vehicle_journey shell as a wrapper around:
- a list of theoric vehicle journey (the vj split because of dst)
- a list of adapted vehicle journey 
- a list of realtime vehicle journey 

we could move some information to the meta vj if relevant.

the refactoring stuff is part of our UTC refactoring (we sadly haven't managed to split this refactoring into smaller PR :cry: ), you can see the changes in the PR #421

PR #457, opened 2014-07-26
No longer use stop_area as first/last item

koch-t:
We now have the crow_fly sections to indicate we start/end from/to a stop_area.
With this patch we get for the request from stop_area:lls to stop_area:asdz the following response

crow_fly stop_area:lls to platform 4 of stop_area:lls
train stop_area:lls platform 4 to stop_area:asdz platform 3
crow_fly stop_area:asdz_platform4 to stop_area:asdz

This gives the most consistent response


koch-t:
I'll make a new PR, i used the Github editor for this

PR #464, opened 2014-07-29
Kraken: Empty fare when not all sections are filled

l-vincent-l:


source/routing/raptor_api.cpp
```diff
@@ -334,7 +335,12 @@ pbnavitia::Response make_pathes(const std::vector<navitia::routing::Path>& paths
 
         //fare computation, done at the end for the journey to be complete
         auto fare = d.fare->compute_fare(path);
-        fill_fare_section(enhanced_response, pb_journey, fare);
+        try {
+            fill_fare_section(enhanced_response, pb_journey, fare);
+        } catch(const navitia::exception& e) {
+            pb_journey->clear_fare();
+            LOG4CPLUS_TRACE(logger, "Unable to compute fare, error : " << e.what());
```
kinnou02, line 342: A trace? We don't find a fare because the data are not available?
it seems related to the currency, so this is not meant to happen and if it  happens we will have a problem

PR #473, opened 2014-07-31
Connector at: Disable debug log of sqlalchemy

kinnou02:
and add the config property "sqlalchemy-log-level" for setting it. The
default value is WARN


kinnou02:
jenkins, test this please

PR #483, opened 2014-08-12
Cli: add last, first section_mode to cli

l-vincent-l:


source/type/type.h
```diff
@@ -1036,6 +1036,20 @@ struct EntryPoint {
     EntryPoint(const Type_e type, const std::string & uri);
 
     EntryPoint() : type(Type_e::Unknown), house_number(-1) {}
+    bool set_mode(const std::string& mode) {
```
kinnou02, line 1079: why return an int32? :100: :3
and the defintion of the function must be in the cpp file, we will save some recompilation this way

antoine-de:
can you rebase it ? (and update with @kinnou02 's comment)

source/type/type.h
l-vincent-l, line 1079: I don't understand what you're talking about. NOOB

PR #489, opened 2014-08-14
Ed: Rewriting of osm2ed

l-vincent-l:
We are now able to compute netherlands with 1,2 GB of RAM

I think we can make some time and RAM improvement by using int32 in polygons

### DIFF
--- a/source/ed/osm2ed.cpp
+++ b/source/ed/osm2ed.cpp
@@ -1,4 +1,34 @@
-#include "osm2ed2.h"
+/* Copyright © 2001-2014, Canal TP and/or its affiliates. All rights reserved.
+
+This file is part of Navitia,
+    the software to build cool stuff with public transport.
+
+Hope you'll enjoy and contribute to this project,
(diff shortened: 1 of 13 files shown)


source/ed/osm2ed.cpp
```diff
@@ -486,45 +785,29 @@ int main(int argc, char** argv) {
...
-
-    v.persistor.clean_georef();
-    v.persistor.clean_poi();
-    v.fill_PoiTypes();
-    v.fill_properties_to_ignore();
-    v.persistor.lotus.prepare_bulk_insert("georef.way", {"id", "name", "uri"});
+    ed::EdPersistor persistor(connection_string);
+    persistor.clean_georef();
+    persistor.clean_poi();
+    po::notify(vm);
```
kinnou02: it's already done line 787

kinnou02:
J'ai une branche avec un fix et du lint a rebaser correctement: https://github.com/kinnou02/navitia/tree/bigger_osm

PR #493, opened 2014-08-14
Timezone

l-vincent-l:
Same than antoine, but with rebase of dev

AD note: for consistency, I copy and adapt the readme of the cloned PR #421 
## PR for time zone handling

The main idea is to convert ALL times to UTC  in the import process, handle only UTC in kraken (the computation part) and change it to local time only at the end of the process in Jörmungandr (the middleware) using the added time zone info in each stop areas.

The UTC conversion difficult part is day saving time (DST) handling.  
For example a stoptime at 7am valid over the year in a paris (Europe/Paris timezone) will be 6AM UTC in winter but 5AM UTC in summer.

Thus all validity_pattern (gtfs calendar), vehicule_journey (gtfs trips) and stoptime have to be split over the DST period.
## Note

despite all its flaws we use boost::date_time for timzeone handling.
for the moment the boost csv is hardlinked in the project, but it will change (maybe by patching boost for it to read the  tzdb).


koch-t:
Several lines are marked by REMOVE 
here https://github.com/CanalTP/navitia/pull/493/files#diff-42ea72ca7fdc62e6ed05061c2e1a7000R711

source/ed/connectors/gtfs_parser.cpp
```diff
@@ -581,47 +805,63 @@ void StopTimeGtfsHandler::init(Data&) {
...
+        stop_time->arrival_time = to_utc(row[arrival_c], utc_offset);
+        stop_time->departure_time = to_utc(row[departure_c], utc_offset);
+
+        stop_time->tmp_stop_point = stop_it->second;
+        //stop_time->journey_pattern_point = journey_pattern_point;
+        stop_time->order = boost::lexical_cast<int>(row[stop_seq_c]);
+        stop_time->vehicle_journey = vj_it->second;
+
+        if(has_col(pickup_c, row) && has_col(drop_off_c, row))
+            stop_time->ODT = (row[pickup_c] == "2" && row[drop_off_c] == "2");
```
koch-t, line 846: I would use || instead of &&. A stop_time is still on_demand if you're not allowed to alight/board (which results in a pickup or drop_off with 1) and not 2 at the stop.

source/type/type.h
```diff
@@ -677,42 +679,48 @@ struct AssociatedCalendar {
...
+    // all times are stored in UTC
+    // however, sometime we do not have a date to convert the time to a local value (in jormungandr)
+    // For example for departure board over a period (calendar)
+    // thus we store the shit needed to convert all stop times of the vehicle journey to local
+    int16_t utc_to_local_offset = 0; //in minutes
+
+    bool is_adapted = false; //REMOVE (change to enum ?)
+    ValidityPattern* adapted_validity_pattern = nullptr; //REMOVE
+    std::vector<VehicleJourney*> adapted_vehicle_journey_list; //REMOVE
+    VehicleJourney* theoric_vehicle_journey = nullptr; //REMOVE
```
koch-t, line 714: TODO markings?

fixtures/date_time_zonespec.csv
```diff
@@ -0,0 +1,383 @@
+"ID","STD ABBR","STD NAME","DST ABBR","DST NAME","GMT offset","DST adjustment","DST Start Date rule","Start time","DST End date rule","End time"
```
koch-t, line 1: Why roll your own timezone database? It will get outdated pretty quickly.

source/type/datetime.h
```diff
@@ -119,6 +119,23 @@ inline std::string to_iso_string_no_fractional(Time t) {
...
+}
+
+inline uint32_t to_posix_timestamp(DateTime datetime, const type::Data &d) {
+    return to_posix_timestamp(to_posix_time(datetime, d));
+}
+
+boost::posix_time::ptime from_posix_timestamp(uint32_t val);
+
+// date are represented as time stamp to midnight
+inline int32_t to_int_date(boost::posix_time::ptime ptime) {
```
koch-t, line 135: y2038 problem?

source/sql/ed/02-migration.sql
```diff
@@ -204,6 +204,17 @@ DO $$
 $$;
 
 DO $$
+    BEGIN
+        BEGIN
+            ALTER TABLE navitia.vehicle_journey ADD
+                COLUMN utc_to_local_offset INT REFERENCES navitia.vehicle_journey;
```
koch-t, line 210: The modifications to vehicle_journey, stop_area, parameters are only added in the migration script, not in the SQL tables.

source/type/type.h
antoine-de, line 714: yes those fields needs to be removed, but the PR is already huge and the deletion of the fields is not trivial (we need to update ed/adapted module).

So this refactoring will be done in another PR

fixtures/date_time_zonespec.csv
antoine-de, line 1: yes it is a very ugly and dumb solution but unfortunately the tz db is not supported by boost::date_time (it is done with boost::locale but we don't want to pay the cost of this module).

boost::date_time can only be fed with a csv or with code.
We don't want to bother the delivery of the file with the binary so we are stuck with the hardcoding of the timezone data (we'll add it to the binary...)

I'm not proud at all of this solution, but I don't see any other means apart from changing boost to read the tz db (we'll see if we can manage some time to do that in the future).

Do you see an alternative ?

source/sql/ed/02-migration.sql
antoine-de, line 210: the migration script is run in the same time as the table script.

you are right that is would be cleaner to put it in both file, but those scripts will be refactored in the short term (replaced by cleaner sql alchemy migration scripts) so I don't think it is worth to add them in the other file.

source/ed/connectors/gtfs_parser.cpp
antoine-de, line 846: you're right, I'll change that

source/type/datetime.h
antoine-de, line 135: yup changed int32 to int64 in protobuf but forgot this method.

Thanks !

antoine-de:
closed because it's a dupplicate from #421

PR #497, opened 2014-08-22
No car direct path

antoine-de:
linked to http://jira.canaltp.fr/browse/NAVITIAII-1032 and http://jira.canaltp.fr/browse/NAVITIAII-1065

We now don't want to show car-only solution (we don't want to promote car), so we filter them in jormungandr.

add a new error too when solutions were found but all has been filtered


antoine-de:
jenkins, test this please

PR #506, opened 2014-08-26
Fix UTC conversion in GTFS

koch-t:
- UTC offset was in minutes, departure-times in seconds
- UTC offset should have been subtracted


antoine-de:
thx!

PR #513, opened 2014-08-29
Jormungandr: modification of sort order for journeys

kinnou02:
we stop to force the "best" journey to be the first result.

http://jira.canaltp.fr/browse/NAVITIAII-1104


antoine-de:
hum not sure it's a good idea to do that, there is some implications.

For example we advise the customer wanting only one journey not to filter on a certain type (because we don't handle those filters well) and to add a 'count=1' filter to get the best one/

kinnou02:
good to go!

PR #514, opened 2014-08-29
ed: support pqxx v3 and v4

TeXitoi:
As pqxx::prepare changed in pqxx v4 and that we need to support pqxx v3 because v4
is not yet in Debian stable, we check the pqxx version with the preprocessor.

fix #301.


buildbotctp:

antoine-de:
jenkins, test this please

PR #515, opened 2014-08-29
Only do crowfly if we leave the requested area

kinnou02:
- If the stop_point is part of the stop_area, then : crow_fly section, no path and no geoJson
- If the stop_point is not part of the stop_area, then : street_network section, with path and geoJson to the centroid of the stop_area

Same principle for admin.

refer to: http://jira.canaltp.fr/browse/NAVITIAII-1093

### DIFF
--- a/source/routing/raptor_api.cpp
+++ b/source/routing/raptor_api.cpp
@@ -57,6 +57,26 @@ void fill_section(pbnavitia::Section *pb_section, const type::VehicleJourney* vj
     fill_pb_object(vj, d, add_info_vehicle_journey, 0, now, action_period);
 }
 
+/**
+ * Choose if we must use a crowfly or a streetnework for a section. This function is call for the first and last section of a journey
+ *
+ * @param point: the object where we are going/coming: the requested origin for the first section or the destination for the last section
+ * @param stop_point: for the first section, the stop point where we are going, or for the last section the stop point from where we come
+ */
+bool use_crow_fly(const type::EntryPoint& point, const type::StopPoint* stop_point){
+    if(point.type == type::Type_e::StopArea){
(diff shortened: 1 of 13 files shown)


source/routing/raptor_api.cpp
```diff
@@ -83,16 +103,22 @@ pbnavitia::Response make_pathes(const std::vector<navitia::routing::Path>& paths
...
+         * 3) We start from a ponctual place (everything but stop_area or admin)
          * We add a street network section from this place to the departure of the first pt_section
+         *
          * If the uri of the origin point and the uri of the departure of the first section are the
          * same we do nothing
-         * */
+         **/
 
-        if (origin.type == type::Type_e::Admin || origin.type == type::Type_e::StopArea) {
+        if (!path.items.front().stop_points.empty() && use_crow_fly(origin, path.items.front().stop_points.front())){
```
antoine-de, line 103: you can delete the test below, useless now

source/routing/raptor_api.cpp
```diff
@@ -235,7 +249,7 @@ pbnavitia::Response make_pathes(const std::vector<navitia::routing::Path>& paths
             pb_section->set_duration((item.arrival - item.departure).total_seconds());
         }
 
-        if (destination.type == type::Type_e::Admin || destination.type == type::Type_e::StopArea) {
+        if (!path.items.back().stop_points.empty() && use_crow_fly(destination, path.items.back().stop_points.back())){
```
antoine-de, line 231: same here

source/routing/raptor_api.cpp
kinnou02, line 103: right, I have deleted them

PR #519, opened 2014-09-01
all: remove warnings

TeXitoi:
Nothing very interesting, except a pretty sum between a const char\* and a
size_t (instead of concatenation).  It also removes some useless assert on
collection sizes.


source/autocomplete/autocomplete.h
```diff
@@ -312,7 +312,7 @@ struct Autocomplete
     /** On passe une chaîne de charactère contenant des mots et on trouve toutes les positions contenant au moins un des mots*/
     std::vector<fl_quality> find_complete(const std::string & str,
                                           const autocomplete_map& synonyms,
-                                          const int wordweight,
```
antoine-de, line 315: I think you should remove the parameter completely

PR #520, opened 2014-09-01
improvement of stay in connection

kinnou02:
- We flag the section before the "stay in", the one where we actually stay.
- Even if we don't move in the vj, we display a origin and destination point (the same), because if we board the vj on the last stop_point it's the connection than move us. Before that the from and to was empty.

refet to: http://jira.canaltp.fr/browse/NAVITIAII-1081

### DIFF
--- a/source/routing/raptor_api.cpp
+++ b/source/routing/raptor_api.cpp
@@ -175,13 +175,16 @@ pbnavitia::Response make_pathes(const std::vector<navitia::routing::Path>& paths
                         departure_ptime = p_deptime;
                     // L'heure d'arrivée au dernier stop point
                     arrival_ptime = p_arrtime;
-                    if(i>0) {
+                    if(i > 0) {
                         const auto & previous_coord = item.stop_points[i-1]->coord;
                         const auto & current_coord = item.stop_points[i]->coord;
                         length += previous_coord.distance_to(current_coord);
                     }
                 }
-                if (item.stop_points.size() > 1) {
+                if (! item.stop_points.empty()) {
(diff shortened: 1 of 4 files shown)


source/routing/raptor_api.cpp
```diff
@@ -175,13 +175,16 @@ pbnavitia::Response make_pathes(const std::vector<navitia::routing::Path>& paths
...
                     arrival_ptime = p_arrtime;
-                    if(i>0) {
+                    if(i > 0) {
                         const auto & previous_coord = item.stop_points[i-1]->coord;
                         const auto & current_coord = item.stop_points[i]->coord;
                         length += previous_coord.distance_to(current_coord);
                     }
                 }
-                if (item.stop_points.size() > 1) {
+                if (item.stop_points.size() > 0) {
```
antoine-de: "! empty()" no ?

source/routing/raptor_api.cpp
```diff
@@ -202,7 +205,13 @@ pbnavitia::Response make_pathes(const std::vector<navitia::routing::Path>& paths
             } else {
                 pb_section->set_type(pbnavitia::TRANSFER);
                 switch(item.type) {
-                    case stay_in : pb_section->set_transfer_type(pbnavitia::stay_in); break;
+                    case stay_in :{
+                        pb_section->set_transfer_type(pbnavitia::stay_in);
+                        //We "stay in" the precedent section, this one is only a transfer
+                        auto* prec_section = pb_journey->mutable_sections(pb_journey->sections_size() - 2);
```
antoine-de: shouldn't you check the sections size ?

PR #524, opened 2014-09-02
Jormungandr: take crow fly mode a a fall back mode in tags

antoine-de:
it's mode coherent that way


antoine-de:
I wanted to split that in 2 pull request, but I failed...

Second commit is to remove walking if bss is enabled.

if bss and walking are enable as fallback mode, we need only to call for
the bss fallback mode since it will enable walking (if it's better than
bss)

that way we can save some kraken calls

PR #529, opened 2014-09-08
Ed: sort stop_time of a vehicle journey in the gtfs_parser

kinnou02:
In some gtfs the stoptime are not sorted

### DIFF
--- a/source/ed/connectors/gtfs_parser.cpp
+++ b/source/ed/connectors/gtfs_parser.cpp
@@ -37,6 +37,7 @@ www.navitia.io
 #include "utils/encoding_converter.h"
 #include "utils/csv.h"
 #include "utils/logger.h"
+#include <boost/range/algorithm/sort.hpp>
 
 namespace nm = ed::types;
 typedef boost::tokenizer< boost::escaped_list_separator<char> > Tokenizer;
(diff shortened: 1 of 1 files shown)


source/ed/connectors/gtfs_parser.cpp
```diff
@@ -886,6 +886,16 @@ void StopTimeGtfsHandler::init(Data&) {
 }
 
 void StopTimeGtfsHandler::finish(Data& data) {
+    LOG4CPLUS_TRACE(logger, "sorting stoptimes of vehicle_journeys");
+    for(auto& vj: data.vehicle_journeys){
+        std::sort(std::begin(vj->stop_time_list), std::end(vj->stop_time_list),
+                [](const nm::StopTime* st1, const nm::StopTime* st2)->bool{
+                    if(st1->order == st2->order){
+                        throw navitia::exception("two stoptime with the same order!!");
```
antoine-de: don't we need some more log for that problem ? (vj name and stoptime id ?)

source/ed/connectors/gtfs_parser.cpp
```diff
@@ -886,6 +886,16 @@ void StopTimeGtfsHandler::init(Data&) {
 }
 
 void StopTimeGtfsHandler::finish(Data& data) {
+    LOG4CPLUS_TRACE(logger, "sorting stoptimes of vehicle_journeys");
+    for(auto& vj: data.vehicle_journeys){
+        std::sort(std::begin(vj->stop_time_list), std::end(vj->stop_time_list),
```
TeXitoi: you can use boost::algorithm:

``` c++
      boost::sort(vj->stop_time_list,
```

Simpler, but I don't block the PR because of that.

TeXitoi: Also, if you want to be a c++11 nazi, you should use `begin()` and `end()` without namespace quantifiers.

source/ed/connectors/gtfs_parser.cpp
kinnou02: I don't have the stop_time id here, but I log the vehicle journey uri and the order of the stop time

source/ed/connectors/gtfs_parser.cpp
kinnou02: I added the `std` namespace after some hesitation, but `boost::sort` is prettier :)

PR #528, opened 2014-09-05
ed & georef: force using a parking when doing a car to stop area

TeXitoi:

PR #530, opened 2014-09-08
georef: nearest_edge returns really the nearest edge now

TeXitoi:
fix http://jira.canaltp.fr/browse/NAVITIAII-1134


antoine-de:
You have to remember to change the #528 after

PR #531, opened 2014-09-09
Convex shape

antoine-de:
Compute convex bounding shape for the regions

we cannot compute concave shape (better than the convex one) because some servers do not have the correct postgis version (postgis >= 2).

We also added a mechanism in jormungandr to query overlapped region if the first one does not return results.

### DIFF
--- a/source/ed/geopal2ed.cpp
+++ b/source/ed/geopal2ed.cpp
@@ -64,7 +64,7 @@ int main(int argc, char * argv[])
     po::store(po::parse_command_line(argc, argv, desc), vm);
 
     if(vm.count("version")){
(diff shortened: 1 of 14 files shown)


source/ed/geopal2ed.cpp
```diff
@@ -71,15 +71,15 @@ int main(int argc, char * argv[])
...
             throw navitia::exception("loading config file failed");
-        }else{
+        } else {
             po::store(po::parse_config_file(stream, desc), vm);
         }
     }
 
     if(vm.count("help") || !vm.count("input")) {
-        std::cout << desc <<  "\n";
+        LOG4CPLUS_INFO(logger, desc);
```
kinnou02: not really

PR #546, opened 2014-09-11
Connector cities

l-vincent-l:
Reads and insert in database all cities of a given osm file

### DIFF
--- a/source/cities/cities.cpp
+++ b/source/cities/cities.cpp
@@ -65,12 +65,12 @@ void ReadRelationsVisitor::relation_callback(uint64_t osm_id, const CanalTP::Tag
             switch(ref.member_type) {
             case OSMPBF::Relation_MemberType::Relation_MemberType_WAY:
                 if (ref.role == "outer" || ref.role == "" || ref.role == "exclave") {
-                    cache.ways.insert(OSMWay(ref.member_id));
+                    cache.ways.insert(std::make_pair(ref.member_id, OSMWay()));
                 }
                 break;
             case OSMPBF::Relation_MemberType::Relation_MemberType_NODE:
                 if (ref.role == "admin_centre" || ref.role == "admin_center") {
-                    cache.nodes.insert(OSMNode(ref.member_id));
+                    cache.nodes.insert(std::make_pair(ref.member_id, OSMNode()));
                 }
                 break;
             case OSMPBF::Relation_MemberType::Relation_MemberType_RELATION:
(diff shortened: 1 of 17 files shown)


source/cities/cities.h
```diff
@@ -0,0 +1,230 @@
...
+
+struct OSMRelation;
+struct OSMCache;
+
+struct OSMNode {
+    static const uint USED_MORE_THAN_ONCE = 0,
+                      FIRST_OR_LAST = 1;
+    uint64_t osm_id = std::numeric_limits<uint64_t>::max();
+    // these attributes are mutable because this object would be use in a set, and
+    // all object in a set are const, since these attributes are not used in the key we can modify them
```
TeXitoi: Why don't you remove osm_id from these struct, remove these mutable things, and build std::map<osm_id, OSMStruct>?

source/cities/cities.h
```diff
@@ -0,0 +1,230 @@
...
+        osm_id(osm_id), references(refs), insee(insee),
+        postal_code(postal_code), name(name), level(level) {}
+
+    bool operator <(const OSMRelation& other) const{
+        return this->osm_id < other.osm_id;
+    }
+
+    bool operator <(const u_int64_t other) const {
+        return osm_id < other;
+    }
```
TeXitoi, line 109: `osm_relation < id` will work, but not `id < osm_relation`. Is it really usefull?

source/cities/cities.h
l-vincent-l: Because I'm a flemmard

source/cities/cities.cpp
```diff
@@ -0,0 +1,253 @@
...
+                    cache.nodes.insert(OSMNode(ref.member_id));
+                }
+                break;
+            case OSMPBF::Relation_MemberType::Relation_MemberType_RELATION:
+                continue;
+            }
+        }
+        if(tags.find("ref:INSEE") != tags.end()){
+            insee = tags.at("ref:INSEE");
+        }
```
TeXitoi: It can be simplified with `const std::string insee = find_or_default("ref:INSEE", tags);`

find_or_default is in `source/ed/data.cpp:123`, to import in utils.

TeXitoi:
Jenkins, test this please.

source/tyr/tyr/tasks.py
```diff
@@ -268,6 +269,22 @@ def load_data(instance_id, data_path):
     import_data(files, instance, backup_file=False, async=False)
 
 
+@celery.task()
+def cities(osm_path):
```
antoine-de, line 273: shouldn't you add a tyr command too ?

l-vincent-l, line 273: It's in tyr/tyr/command/cities.py NOOB

PR #547, opened 2014-09-12
Tyr: add a command for deleting old backup directories 

kinnou02:
For one instance:

```
./manage.py purge_instance fr-idf
```

For purging all the instances:

```
./manage.py purge_instances
```

The option "-b" can be use for run the task in background (with a
celery worker)


antoine-de:
I do aggry with @guikcd I think it would be a good thing to have to possibility to keep more than one data set (but the sql query would be becoming even more complicated :smile: )

kinnou02:
I can do it, but it will be a lot less pretty :)

PR #551, opened 2014-09-15
Jormungandr: fiw the error message when one instance is dead

kinnou02:
refer to: http://jira.canaltp.fr/browse/NAVITIAII-896


antoine-de:
thx!

PR #554, opened 2014-09-15
kraken & tyr: add a fake heartbeat amqp for all kraken from tyr

kinnou02:


source/kraken/maintenance_worker.cpp
```diff
@@ -86,8 +86,11 @@ void MaintenanceWorker::listen_rabbitmq(){
             LOG4CPLUS_WARN(logger, "protobuf not valid!");
             continue;
         }
-        if(task.action() == pbnavitia::RELOAD){
-            load();
+        switch(task.action()){
+            case pbnavitia::RELOAD:
+                load(); break;
+            default:
+                LOG4CPLUS_TRACE(logger, "Message ignored");
```
TeXitoi, line 93: If I understand well, this is a keep alive message, not a ping (which expect a pong response).

kinnou02, line 93: It's true

PR #566, opened 2014-09-17
kraken: usage of boost program_option for the configuration

kinnou02:


source/kraken/configuration.cpp
```diff
@@ -0,0 +1,72 @@
...
+}
+
+void Configuration::load(const std::string& filename){
+    po::options_description desc = get_options_description();
+    std::ifstream stream(filename);
+
+    if(!stream.is_open() || !stream.good()){
+        throw navitia::exception("impossible to open: " + filename);
+    }
+    auto tmp = po::parse_config_file(stream, desc, true);
```
antoine-de, line 66: you should add a smal comment to say we ignore unkown param for log4cplus

source/kraken/maintenance_worker.cpp
```diff
@@ -96,20 +93,16 @@ void MaintenanceWorker::listen_rabbitmq(){
...
-    std::string host = conf->get_as<std::string>("BROKER", "host", "localhost");
-    int port = conf->get_as<int>("BROKER", "port", 5672);
-    std::string username = conf->get_as<std::string>("BROKER", "username", "guest");
-    std::string password = conf->get_as<std::string>("BROKER", "password", "guest");
-    std::string vhost = conf->get_as<std::string>("BROKER", "vhost", "/");
+    std::string instance_name = conf.vm["GENERAL.instance_name"].as<std::string>();
+    std::string exchange_name = conf.vm["BROKER.exchange"].as<std::string>();
+    std::string host = conf.vm["BROKER.host"].as<std::string>();
+    int port = conf.vm["BROKER.port"].as<int>();
+    std::string username = conf.vm["BROKER.username"].as<std::string>();
```
antoine-de: I think it's better to put all conf in a custom struct, I don't like the string key stuff

source/kraken/maintenance_worker.h
```diff
@@ -43,6 +44,7 @@ class MaintenanceWorker{
     private:
         DataManager<type::Data>& data_manager;
         log4cplus::Logger logger;
+        kraken::Configuration conf;
```
antoine-de: you can const it

source/kraken/worker.h
```diff
@@ -55,12 +56,12 @@ class Worker {
 
         // we keep a reference to data_manager in each thread
         DataManager<navitia::type::Data>& data_manager;
-
+        kraken::Configuration conf;
```
antoine-de: you can const it

PR #594, opened 2014-09-29
Disruptions: add new object model

antoine-de:
add chaos object model to navitia.

Still keeping the old one before approval of this model.

### DIFF
--- a/source/type/message.h
+++ b/source/type/message.h
@@ -92,13 +92,20 @@ struct PtObject {
 
 struct Disruption;
 
+struct Message {
+    std::string text;
+
+    boost::posix_time::ptime created_at;
+    boost::posix_time::ptime updated_at;
+};
+
 struct Impact {
     std::string uri;
     boost::posix_time::ptime created_at;
     boost::posix_time::ptime updated_at;
 
     // the application period define when the impact happen
-    std::vector<boost::gregorian::date_period> application_periods;
+    std::vector<boost::posix_time::time_period> application_periods;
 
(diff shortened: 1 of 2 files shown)


source/type/message.h
```diff
@@ -51,6 +51,119 @@ www.navitia.io
...
+    std::vector<boost::gregorian::date_period> application_periods;
+
+    Severity severity;
+
+    std::vector<PtObject> informed_entities;
+
+    std::vector<Message> messages;
+
+    //link to the parent disruption
+    Disruption* disruption;
```
kinnou02, line 119: so we keep a raw pointer?

source/type/message.h
```diff
@@ -51,6 +51,126 @@ www.navitia.io
...
+    boost::posix_time::ptime created_at;
+    boost::posix_time::ptime updated_at;
+};
+
+struct Severity {
+    std::string id;
+    std::string wording;
+    boost::posix_time::ptime created_at;
+    boost::posix_time::ptime updated_at;
+    std::string color;
```
l-vincent-l, line 81: color?

source/type/message.h
```diff
@@ -51,6 +51,126 @@ www.navitia.io
...
+struct Message {
+    std::string text;
+
+    boost::posix_time::ptime created_at;
+    boost::posix_time::ptime updated_at;
+};
+
+
+struct Disruption {
+    std::string id;
```
l-vincent-l: uri non ?

PR #598, opened 2014-09-30
Protobuf: rename place to pbobject

l-vincent-l:


TeXitoi:
in navitia/source/type/CMakeLists.txt, add something like:

``` diff
-target_link_libraries(types utils)
+target_link_libraries(types utils pb_lib)
+target_link_libraries(data types)
```

It may fix these protobuf problems on jenkins

PR #604, opened 2014-10-02
ed: use geometries.txt in fusio to export shapes

TeXitoi:


koch-t:
More general toward shapes, it might be worth to use st_simplify somewhere during storing or loading. Nobody cares about <1 meter accuracy

TeXitoi:
@skywave thanks for the idea of simplifying!

PR #610, opened 2014-10-06
Jormungandr drop db

l-vincent-l:
Drop usage of database in Jormungandr:
We check in every instance if the id is in it or not, then we cache it.

To cache we use: https://pythonhosted.org/Flask-Cache/

Thus we no longer need aggregate_places \o/.

Some improvements has to be made: 
- we need to paralellize calls to kraken.
- Maybe we can cache per request the region.
- It would be great if id request to kraken were computed before others


source/jormungandr/requirements.txt
```diff
@@ -22,3 +22,4 @@ six==1.6.1
 wsgiref==0.1.2
 kombu==3.0.8
 python-dateutil==1.5
+Flask-Dogpile-Cache==0.2
```
kinnou02: you don't use this one

source/navitiacommon/navitiacommon/models.py
```diff
@@ -107,10 +96,7 @@ def _has_access(self, instance_name, api_name):
 
     def has_access(self, instance_name, api_name):
```
kinnou02: you can remove this one, and rename `_has_access` for using it directly.

source/jormungandr/jormungandr/__init__.py
```diff
@@ -57,17 +59,13 @@
...
-               password=app.config['REDIS_PASSWORD'],
-               default_ttl=app.config['AUTH_CACHE_TTL'])
-
+cache = Cache(app, config=app.config['CACHE_CONFIGURATION'])
 
 from jormungandr.instance_manager import InstanceManager
-i_manager = InstanceManager()
-i_manager.initialisation(start_ping=app.config['START_MONITORING_THREAD'])
+ini_files = None if 'INI_FILE' not in app.config else app.config['INI_FILE']
+instances_dir = None if 'INSTANCES_DIR' not in app.config else app.config['INSTANCES_DIR']
```
antoine-de: it's clearer with app.config.get('INSTANCE_DIR', None) no ? (same for ini_files)

source/jormungandr/jormungandr/__init__.py
```diff
@@ -57,17 +59,13 @@
...
+i_manager = InstanceManager(ini_files, instances_dir, start_ping=app.config['START_MONITORING_THREAD'])
```
antoine-de: should'nt we protect the 'START_MONITORING_THREAD' the same way ?

source/jormungandr/jormungandr/authentication.py
```diff
@@ -38,33 +38,33 @@
...
 
 def authentication_required(func):
     """
     decorateur chargé de l'authentification des requetes
     fonctionne pour chaque API prenant un paramétre la région
-    si la region est absente de la requéte la requete est automatique autorisé
     """
     @wraps(func)
     def wrapper(*args, **kwargs):
+        logging.info('authentication_required')
```
antoine-de, line 51: you think we should keep this log ?

source/jormungandr/jormungandr/authentication.py
```diff
@@ -38,33 +38,33 @@
...
         if 'region' in kwargs:
             region = kwargs['region']
             #TODO revoir comment on gere le lon/lat
         elif 'lon' in kwargs and 'lat' in kwargs:
-            try:
-                from jormungandr import i_manager  # quick fix to avoid circular dependencies
-                region = i_manager.key_of_coord(lon=kwargs['lon'],
+            try:  # quick fix to avoid circular dependencies
+                from jormungandr import i_manager
+                region = i_manager.get_region(lon=kwargs['lon'],
```
antoine-de, line 59: So actually if the lon/lat is not valid we'll return a 401/403 instead of a RegionNotFound exception.

Is it what we want ?

source/jormungandr/jormungandr/instance.py
```diff
@@ -81,5 +94,31 @@ def send_and_receive(self, request, timeout=10000):
                 socket.setsockopt(zmq.LINGER, 0)
                 socket.close()
                 logging.error("La requete : " + request.SerializeToString()
-                              + " a echoue " + self.socket_path)
+                        + " a echoue " + self.socket_path)
                 raise DeadSocketException(self.name, self.socket_path)
+
+
+    def get_id(self, id_):
```
antoine-de, line 114: even if the methods are small you should put a comment explaining the purpose (at least for get_id and get_external_codes)

source/jormungandr/jormungandr/instance_manager.py
```diff
@@ -91,28 +91,31 @@ class InstanceManager(object):
...
 
-    def __init__(self):
-        pass
+    def __init__(self, ini_files=None, instances_dir=None, start_ping=False):
+        # if a .ini file is defined in the settings we take it
+        # else we load all .ini file found in the INSTANCES_DIR
+        if ini_files is None and instances_dir is None:
+            raise ValueError("ini_files or instance_dir has to be set")
+        if ini_files:
+            self.ini_files = app.config['INI_FILES']
```
antoine-de: self.ini_files = ini_files no ?

source/jormungandr/jormungandr/instance_manager.py
```diff
@@ -91,28 +91,31 @@ class InstanceManager(object):
...
+        else:
+            self.ini_files = glob.glob(instances_dir + '/*.ini')
+        self.start_ping = start_ping
 
-    def initialisation(self, start_ping=True):
+    def __repr__(self):
+        return "<instance singleton"
```
antoine-de: is this **repr** mandatory ?

source/jormungandr/jormungandr/instance_manager.py
```diff
@@ -152,7 +155,12 @@ def dispatch(self, arguments, api, instance_obj=None, instance_name=None,
             instance = self.instances[instance_name]
             if api in instance.script.apis:
                 api_func = getattr(instance.script, api)
-                return api_func(arguments, instance)
+                resp = api_func(arguments, instance)
+                if resp.HasField("publication_date") and\
```
antoine-de, line 145: can some response have no publication date ? is so, is it a problem if we don't delete the cache ?

source/jormungandr/jormungandr/instance_manager.py
```diff
@@ -192,12 +200,20 @@ def stop(self):
...
-    def key_of_id(self, object_id, only_one=True):
-        """ Retrieves the key of the region of a given id
-            if it's a coord calls key_of_coord
-            Return one region key if only_one param is true, or None if it doesn't exists
-            and the list of possible regions if only_one is set to False
-        """
+
+    def _filter_authorized_instances(self, instances, api):
+        user = authentication.get_user(token=authentication.get_token())
+        if not instances:
```
antoine-de, line 191: you should put that the line before no ? (and return None would be clearer afaik)

source/jormungandr/jormungandr/instance_manager.py
```diff
@@ -208,75 +224,57 @@ def key_of_id(self, object_id, only_one=True):
                 flat = float(lat)
             except:
                 raise RegionNotFound(object_id=object_id)
```
antoine-de: instead of a region not found there, shouldn't it be better to say that the argument are not well formated ?

source/jormungandr/jormungandr/instance_manager.py
```diff
@@ -192,12 +200,20 @@ def stop(self):
...
+        if not instances:
+            return instances
+        valid_regions = [i for i in instances if authentication.has_access(i,
+            abort=False, user=user, api=api)]
+        if not valid_regions:
+            authentication.abort_request(user)
+        return valid_regions
+
+
+    @cache.memoize()
```
antoine-de: no timeout for this cache ? since it is wiped out with the data reloading ?
if so a comment would ease the understanding :)

from a general point of view shouldn't we think about the configuration of the cache (for the production plaftorm for example) ?

source/jormungandr/jormungandr/authentication.py
l-vincent-l, line 51: You never know

source/jormungandr/jormungandr/interfaces/v1/Uri.py
```diff
@@ -94,23 +93,18 @@ def get(self, region=None, lon=None, lat=None, uri=None, id=None):
...
-                    for instance in res.instances:
-                        if authentication.has_access(instance, abort=False, user=user):
-                            region = instance.name
-                    if not region:
-                        authentication.abort_request(user=user)
-                else:
+                for instance in i_manager.get_regions():
+                    if i_manager.instances[instance].has_external_code(type_, args["external_code"]):
+                        region = instance
+                        break
```
antoine-de, line 99: so we stop at the first one ?

source/navitiacommon/models.py
```diff
@@ -0,0 +1,629 @@
+# encoding: utf-8
```
antoine-de: I don't understand why you copy all this file

source/tyr/manage_tyr.py
```diff
@@ -33,15 +33,12 @@
...
-from tyr.command import AggregatePlacesCommand, ReloadAtCommand, AtReloader,\
+from tyr.command import ReloadAtCommand, AtReloader,\
     ReloadKrakenCommand, BuildDataCommand, LoadDataCommand
 
 
 migrate = Migrate(app, db)
 manager.add_command('db', MigrateCommand)
-#A command used in development environment to run aggregate places without
-#having to run tyr
-manager.add_command('aggregate_places', AggregatePlacesCommand())
```
antoine-de, line 44: \o/

source/jormungandr/jormungandr/instance_manager.py
l-vincent-l: I think singleton need this.

source/jormungandr/jormungandr/instance_manager.py
l-vincent-l, line 145: In normal mode nop, but when we will upgrade the version this could happen

source/jormungandr/jormungandr/interfaces/v1/Uri.py
l-vincent-l, line 99: yup, that's strange to me too, but that's out of the scope of this PR.
This should be log as a bug.

antoine-de:
Jenkins, test this please

source/jormungandr/jormungandr/__init__.py
TeXitoi: And shorter with app.config.get('INSTANCE_DIR') ;)

l-vincent-l:
Jenkins, test this please

source/jormungandr/jormungandr/__init__.py
```diff
@@ -34,12 +34,14 @@
...
 from flask import Flask, got_request_exception
 from flask.ext.restful import Api
+from flask.ext.cache import Cache
 import sys
 from jormungandr.exceptions import log_exception
 from jormungandr.helper import ReverseProxied
 
 app = Flask(__name__)
 app.config.from_object('jormungandr.default_settings')
+app.bob = 'bob'
```
kinnou02: ?

l-vincent-l:
Jenkins, test this please

kinnou02:
I don't see the migration file for dropping all the tables filled by aggregate_places

PR #616, opened 2014-10-07
Osm2ed: relation with way and admin were not inserted in database...

l-vincent-l:


source/ed/osm2ed.cpp
```diff
@@ -367,10 +372,37 @@ void OSMCache::insert_relations() {
...
+    LOG4CPLUS_INFO(logger, "Ignored " << std::to_string(nb_empty_polygons) 
+            << " admins because their polygons were empty");
 }
 
 
+void OSMCache::insert_rel_way_admins() {
+    lotus.prepare_bulk_insert("georef.rel_way_admin", {"admin_id", "way_id"});
+    size_t n_inserted = 0 ;
+    const size_t max_n_inserted = 20000;
+    for (auto map_ways : this->way_admin_map) {
```
antoine-de: you might want to add const & to avoid useless copy

antoine-de:
maybe not now, but we really need to put a small pbf in fixture, and test the load in a 'unit' test

PR #619, opened 2014-10-08
Scenario

kinnou02:
refactoring of scenario before implementing the next step of traveler profile


source/jormungandr/jormungandr/scenarios/CMakeLists.txt
```diff
@@ -0,0 +1,3 @@
+INSTALL_FILES(/srv/jormungandr/scenarios __init__.py)
```
antoine-de: you need to add the glob no ?

PR #630, opened 2014-10-13
Jormungandr: add possibility to sort by departure hour on journeys

kinnou02:
It's another parameter in the jormungandr database, so you need to
upgrade the database.
I will add cache after the PR on "drop db" is merged

refer to: http://jira.canaltp.fr/browse/NAVITIAII-1218?filter=12022


source/jormungandr/jormungandr/scenarios/utils.py
```diff
@@ -99,19 +99,19 @@ class JourneySorter:
...
         self.clockwise = clockwise
+        self.sort_order = sort_order
 
     def __call__(self, j1, j2):
-
-        if self.clockwise:
-            #for clockwise query, we want to sort first on the arrival time
-            if j1.arrival_date_time != j2.arrival_date_time:
-                return -1 if j1.arrival_date_time < j2.arrival_date_time else 1
+        if self.sort_order == 'arrival_time':
```
antoine-de: I think it's better to have several JourneySorter

source/tyr/tyr/resources.py
```diff
@@ -126,9 +132,14 @@ def put(self, id=None, name=None):
         parser.add_argument('scenario', type=str, required=True,
                 case_sensitive=False, help='the name of the scenario used by jormungandr', choices=['default', 'keolis'],
                 location=('json', 'values'))
+        parser.add_argument('journey_order', type=str, required=True,
+                case_sensitive=False, help='the name of the scenario used by jormungandr',
+                choices=['arrival_time', 'departure_time'], location=('json', 'values'))
```
antoine-de, line 137: When you'll split the journeySorter, you might be able to do choices=possible_sorts.key()

PR #631, opened 2014-10-13
Jormungandr: retrieve only regions that you have access to

l-vincent-l:


source/jormungandr/tests/authentication_tests.py
```diff
@@ -81,6 +82,17 @@ def setUp(self):
...
+
+    def test_coverage(self):
+        """
+        User only has access to the first region
+        """
+        with user_set(app, 1):
+            response_obj = self.app.get('/v1/coverage')
+            response = json.loads(response_obj.data)
+            assert('regions' in response)
+            assert(len(response['regions']) == 1)
```
antoine-de, line 94: you should also check that it's 'main_routing_test' no ?

l-vincent-l, line 94: Yeah, I could have

PR #635, opened 2014-10-14
Hide null regions

l-vincent-l:


source/jormungandr/tests/coverage_tests.py
```diff
@@ -0,0 +1,45 @@
...
+import logging
+
+@dataset(["main_routing_test", "null_status_test"])
+class TestNullStatus(AbstractTestFixture):
+    """
+    Test with an empty coverage
+    """
+
+    def null_status_test(self):
+        response = self.query("/v1/coverage", display=False)
```
antoine-de, line 44: could you also add a test to check that we can anyway do a "v1/coverage/null_status_test" to get the status of the region ?

source/jormungandr/tests/coverage_tests.py
```diff
@@ -0,0 +1,52 @@
...
+    def test_null_status_is_hidden(self):
+        response = self.query("/v1/coverage", display=False)
+        logging.info(response['regions'])
+        assert('regions' in response)
+        assert(len(response['regions']) == 1)
+        assert(response['regions'][0]['id'] == 'main_routing_test')
+
+
+    def test_null_status(self):
+        response = self.query("/v1/coverage/null_status_test", display=False)
```
antoine-de, line 56: can you put a comment there ? to explain the intent of the test (same above)

PR #638, opened 2014-10-15
journeys: add the posibility to have different fallback duration for each mode

kinnou02:
Split the "max_duration_to_pt" for each fallback mode. We keep the
"max_duration_to_pt" in the interface for not breaking the compatibily.

The the max duration for car is now set to 30 min.

The "max_duration_to_pt" field in the protobuf is keeped for compatibily
during the migration to the next version. It will be removed in another
pull request.

### DIFF
--- a/source/kraken/worker.cpp
+++ b/source/kraken/worker.cpp
@@ -322,25 +322,49 @@ type::StreetNetworkParams Worker::streetnetwork_params_of_entry_point(const pbna
         result.mode = type::static_data::get()->modeByCaption(request.destination_mode());
         result.set_filter(request.destination_filter());
     }
+    int max_non_pt;
     switch(result.mode){
         case type::Mode_e::Bike:
             result.offset = data->geo_ref->offsets[type::Mode_e::Bike];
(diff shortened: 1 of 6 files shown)


source/kraken/worker.cpp
```diff
@@ -322,25 +322,45 @@ type::StreetNetworkParams Worker::streetnetwork_params_of_entry_point(const pbna
...
     }
+    int max_non_pt;
     switch(result.mode){
         case type::Mode_e::Bike:
             result.offset = data->geo_ref->offsets[type::Mode_e::Bike];
             result.speed_factor = request.bike_speed() / georef::default_speed[type::Mode_e::Bike];
+            if(request.has_max_bike_duration_to_pt()){
+                max_non_pt = request.max_bike_duration_to_pt();
+            }else{
+                max_non_pt = request.max_duration_to_pt();
```
antoine-de, line 334: shouldn't you also put your "we keep this field for compatibily with kraken 1.2, to be removed after the release of the 1.3" comment ?

PR #642, opened 2014-10-15
Change dominance rule

l-vincent-l:
We accept journeys in public transport arriving later than journeys without.


source/routing/tests/routing_api_test.cpp
```diff
@@ -1187,7 +1187,12 @@ BOOST_FIXTURE_TEST_CASE(biking_length_test, streetnetworkmode_fixture<normal_spe
     pbnavitia::Response resp = make_response();
 
     BOOST_REQUIRE(resp.journeys_size() != 0);
-    pbnavitia::Journey journey = resp.journeys(0);
+    pbnavitia::Journey journey;
+    if (resp.journeys_size() == 1) {
```
antoine-de: why do you have a choice there ? (it's always 2 journeys no?)

l-vincent-l: You're right

PR #646, opened 2014-10-17
Nm planner - fix this api

djludo:


buildbotctp:

antoine-de:
@djludo you need to rebase your pullrequest before pushing it

source/routing/raptor.cpp
```diff
@@ -276,10 +276,31 @@ RAPTOR::compute_nm_all(const std::vector<std::pair<type::EntryPoint, std::vector
...
+        type::EntryPoint m_entry_point = m_point.first;
+
+        auto calc_dest = m_point.second;
 		auto tmp = makePathes(calc_dep, calc_dest, accessibilite_params, *this, clockwise, disruption_active);
-		result.insert(result.end(), tmp.begin(), tmp.end());
+
+        for(Path& path : tmp){
+            path.origin.type = (nt::Type_e)-1;
+
+            if (path.items.size() == 0)
```
antoine-de: path.items.empty() is better

TeXitoi:
jenkins, test this please

source/jormungandr/jormungandr/scenarios/default.py
```diff
@@ -349,6 +349,16 @@ def delete_journeys(self, resp, request, final_filter):
...
+	if "clockwise" not in request or request['clockwise']:
+		if "destination" in request and isinstance(request['destination'], list) \
+					    and len(request['destination']) > 1:
+			return #for n-m calculation we don't want to filter
+
+	if "clockwise" in request and not request['clockwise']:
+		if "origin" in request and isinstance(request['origin'], list) \
+				       and len(request['origin']) > 1:
+			return #for n-m calculation we don't want to filter
+
```
TeXitoi, line 361: You have problem with indentation here: you use TAB to indent, but we require 4 spaces. It is particulary critical for python code as the indentation is meaningful.

source/routing/raptor.h
```diff
@@ -116,11 +116,11 @@ struct RAPTOR
...
 
-    /** Calcul d'itinéraires multiples dans le sens horaire à partir de plusieurs
-     * stop points de départs, vers plusieurs stoppoints d'arrivée,
-     * à une heure donnée.
-     */
-    std::vector<Path>
+	/** Calcul d'itinéraires multiples dans le sens horaire à partir de plusieurs
+	 * stop points de départs, vers plusieurs stoppoints d'arrivée,
+	 * à une heure donnée.
+	 */
```
TeXitoi: you replaced correct 4 spaces indentation by tabs

source/routing/raptor.cpp
```diff
@@ -272,10 +272,31 @@ RAPTOR::compute_nm_all(const std::vector<std::pair<type::EntryPoint, std::vector
 
 	auto m_points = clockwise ? arrivals : departures;
 
-	for(auto item : m_points) {
-		auto calc_dest = item.second;
+    for(auto m_point : m_points) {
+        type::EntryPoint m_entry_point = m_point.first;
```
TeXitoi: better to use `const type::EntyPoint& m_entry_point` to avoid useless copy

source/routing/raptor.cpp
```diff
@@ -272,10 +272,31 @@ RAPTOR::compute_nm_all(const std::vector<std::pair<type::EntryPoint, std::vector
...
+
+        auto calc_dest = m_point.second;
```
TeXitoi: better to use `const auto& calc_dest` to avoid useless copy, as auto will expand to a type without qualifiers (const, volatile, &...)

source/routing/raptor.cpp
```diff
@@ -272,10 +272,31 @@ RAPTOR::compute_nm_all(const std::vector<std::pair<type::EntryPoint, std::vector
...
 		auto tmp = makePathes(calc_dep, calc_dest, accessibilite_params, *this, clockwise, disruption_active);
-		result.insert(result.end(), tmp.begin(), tmp.end());
+
+        for(Path& path : tmp){
+            path.origin.type = (nt::Type_e)-1;
```
TeXitoi: Why do you need such a guard? `nt::Type_e::Unknown` don't do the job? Maybe you should check at the end of the block that the type is correct?

source/routing/raptor.cpp
```diff
@@ -272,10 +272,31 @@ RAPTOR::compute_nm_all(const std::vector<std::pair<type::EntryPoint, std::vector
...
+
+            if (path.items.empty())
+                continue;
+
+            PathItem path_item = clockwise ? path.items.front() : path.items.back();
+            if (path_item.stop_points.size() == 0)
```
TeXitoi: Another `size()` -> `empty()`

source/routing/raptor.cpp
```diff
@@ -272,10 +272,31 @@ RAPTOR::compute_nm_all(const std::vector<std::pair<type::EntryPoint, std::vector
...
+                continue;
+
+            // must find which item of calc_dep has been computed
+            const nt::StopPoint* stop_point = clockwise ? path_item.stop_points.front() : path_item.stop_points.back();
+            for(auto n_point : n_calc_dep)
+                for(auto n_stop_point : n_point.second)
+                    if (stop_point->idx == n_stop_point.first)
+                        path.origin = n_point.first;
```
TeXitoi, line 296: Prefere `const auto&` when possible to avoid useless copy.

Can't we stop this double loop when this test is true? Musn't this test be true at least one time?

If yes to these, what do you think of a function (about) like that:

``` c++
static const nt::EntryPoint*
get_n_item(const nt::StopPoint* stop_point,
           const std::vector<std::pair<type::EntryPoint, std::vector<std::pair<type::idx_t, navitia::time_duration>>>>& n_calc_dep) {
    for (const auto& n_point: n_calc_dep)
        for (const auto& n_stop_point: n_point.second)
            if (stop_point->idx == n_stop_point.first)
                return &n_point.first;
    return nullptr;
}
```

like that, we stop the loop earlier, if we don't find, we know it (and we don't need the guard above).

Edit: allow not found by returning a pointer

source/routing/raptor_api.cpp
```diff
@@ -556,43 +564,54 @@ make_nm_response(RAPTOR &raptor, const std::vector<type::EntryPoint> &origins,
...
     }
     
     std::vector<std::pair<type::EntryPoint, std::vector<std::pair<type::idx_t, navitia::time_duration> > > > departures;
     std::vector<std::pair<type::EntryPoint, std::vector<std::pair<type::idx_t, navitia::time_duration> > > > arrivals;
     
-    worker.init(origins[0]);
-    
-    for(auto org : origins) {
-		departures.push_back(std::make_pair(org, get_stop_points(org, raptor.data, worker)));
+    for(type::EntryPoint org : origins) {
```
TeXitoi: `const&`

source/routing/raptor_api.cpp
```diff
@@ -556,43 +564,54 @@ make_nm_response(RAPTOR &raptor, const std::vector<type::EntryPoint> &origins,
...
+        worker.init(org);
+        std::vector<std::pair<type::idx_t, navitia::time_duration> > org_stop_points = get_stop_points(org, raptor.data, worker);
```
TeXitoi: `auto org_stop_points` to avoid the long line?

source/routing/raptor_api.cpp
```diff
@@ -556,43 +564,54 @@ make_nm_response(RAPTOR &raptor, const std::vector<type::EntryPoint> &origins,
...
+        for (std::pair<type::idx_t, navitia::time_duration>& org_stop_point : org_stop_points) {
+            org_stop_point.second += navitia::seconds(org.access_duration);
+        }
+        departures.push_back(std::make_pair(org, org_stop_points));
 	}
 	
 	for(auto dst : destinations) {
+        worker.init(dst);
+        std::vector<std::pair<type::idx_t, navitia::time_duration> > dst_stop_points = get_stop_points(dst, raptor.data, worker);
```
TeXitoi: auto to avoid the long line?

source/routing/raptor_api.cpp
TeXitoi: And org is quite a strange abreviation for origin.

source/routing/raptor_api.cpp
```diff
@@ -608,26 +627,46 @@ make_nm_response(RAPTOR &raptor, const std::vector<type::EntryPoint> &origins,
...
-        // Lorsqu'on demande qu'un seul horaire, on garde tous les résultas
-        if(datetimes.size() == 1) {
-            result = tmp;
-            for(auto & path : result) {
-                path.request_time = datetime;
+        for(std::pair<type::EntryPoint, std::vector<Path>> pathes_for_m_point : pathes_by_entrypoint){
+            std::vector<Path> tmp = pathes_for_m_point.second;
+
+            type::EntryPoint origin;
+            origin.type = (nt::Type_e)-1;
```
TeXitoi: OK, here is the guard checking. Then I think Unknown will do the job.

source/routing/raptor.cpp
```diff
@@ -272,10 +272,31 @@ RAPTOR::compute_nm_all(const std::vector<std::pair<type::EntryPoint, std::vector
...
+        }
+
+        result.push_back(std::make_pair(m_entry_point, tmp));
```
TeXitoi: If I understand well, here, elemets of `tmp` that have type == -1 can be removed. You can filter `tmp` before push_backing it, to avoid checking the guard in raptor_api.cpp

source/routing/raptor_api.cpp
djludo: "org" is a common abreviation for origin (http://www.allacronyms.com/origin/abbreviated) and an assembly instruction that means the origin address where instructions should be loaded :)

TeXitoi:
jenkins, test this please

PR #649, opened 2014-10-20
Fix crash stay in

l-vincent-l:
It will have to be cherry-pick in release


source/routing/tests/raptor_test.cpp
```diff
@@ -1194,3 +1194,19 @@ BOOST_AUTO_TEST_CASE(over_midnight_special) {
     BOOST_CHECK_EQUAL(res.items[2].arrival.date().day(), 14);
 
 }
+
+
+BOOST_AUTO_TEST_CASE(invalid_stay_in_overmidnight) {
```
antoine-de, line 1199: can you add comments ?

and wouldn't it better to have the same test the second day ? it should work no ?

PR #651, opened 2014-10-20
ed: importing shapes (much) faster

TeXitoi:


l-vincent-l:
Can you add tests?

source/ed/connectors/gtfs_parser.cpp
```diff
@@ -911,6 +895,7 @@ void TripsGtfsHandler::handle_line(Data& data, const csv_row& row, bool) {
         //we add them on our meta vj
         meta_vj.theoric_vj.push_back(vj);
         vj->meta_vj_name = row[trip_c];
+        vj->shape = shape;
```
antoine-de: don't you think it would be clearer to call that shape_id ?

TeXitoi:
@l-vincent-l testing ed seems difficult, but if you have a good idea to ĥow test that, I listen.

PR #654, opened 2014-10-21
more postal code for cities

kinnou02:
For cities with "Municipal Arrondissement" (admin level 9) we take the postal code of
them and affect a range to the corresponding city.

And we now get the postcode of the "admin_centre" too.

This changes will have to be ported in osm2ed...

### DIFF
--- a/source/cities/cities.cpp
+++ b/source/cities/cities.cpp
@@ -60,12 +60,11 @@ namespace navitia { namespace cities {
 void ReadRelationsVisitor::relation_callback(uint64_t osm_id, const CanalTP::Tags &tags, const CanalTP::References &refs) {
     auto logger = log4cplus::Logger::getInstance("log");
     const auto boundary = tags.find("boundary");
-    if (boundary == tags.end() || (boundary->second != "administrative" && 
-                boundary->second != "multipolygon")) {
+    if (boundary == tags.end() || (boundary->second != "administrative" && boundary->second != "multipolygon")) {
         return;
     }
     const auto tmp_admin_level = tags.find("admin_level");
-    if(tmp_admin_level != tags.end() && tmp_admin_level->second == "8") {
+    if(tmp_admin_level != tags.end() && (tmp_admin_level->second == "8" || tmp_admin_level->second == "9")) {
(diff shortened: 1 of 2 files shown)


source/cities/cities.cpp
```diff
@@ -122,8 +122,49 @@ void ReadNodesVisitor::node_callback(uint64_t osm_id, double lon, double lat,
...
+    std::vector<OSMRelation*> result;
+    auto callback = [](OSMRelation* rel, void* c)->bool{
+        std::pair<uint32_t, std::vector<OSMRelation*>*>* context;
+        context = reinterpret_cast<std::pair<uint32_t, std::vector<OSMRelation*>*>*>(c);
+        if(rel->level == context->first){
+            context->second->push_back(rel);
+        }
+        return true;
+    };
+    std::pair<uint32_t, std::vector<OSMRelation*>*> context = std::make_pair(level, &result);
```
TeXitoi: `typedef`ing that (as level_relations for example) will help reading.

source/cities/cities.h
```diff
@@ -134,15 +158,17 @@ struct OSMCache {
     std::unordered_map<uint64_t, OSMRelation> relations;
     std::unordered_map<uint64_t, OSMNode> nodes;
     std::unordered_map<uint64_t, OSMWay> ways;
+    RTree<OSMRelation*, double, 2> admin_tree;
```
TeXitoi, line 161: You have a prettier rtree in boost: http://www.boost.org/doc/libs/1_56_0/libs/geometry/doc/html/geometry/spatial_indexes.html (this RTree interface is just sooo C89 ;))

source/cities/cities.cpp
```diff
@@ -251,8 +321,9 @@ void OSMRelation::build_geometry(OSMCache& cache) {
             if (!node_it->second.is_defined()) {
                 continue;
             }
-            if (ref.role == "admin_centre") {
+            if (ref.role == "admin_centre" || ref.role == "admin_center") {
```
TeXitoi:
``` c++
in(ref.role, {"admin_centre", "admin_center"})
```

PR #655, opened 2014-10-21
proximity_list: correct the name of the place object by using fill_pb_placemark

TeXitoi:


antoine-de:
can you add tests ? :smile: 

jormun integration test could be nice on places/id

TeXitoi:
JIRA: http://jira.canaltp.fr/browse/NAVITIAII-1112

source/jormungandr/tests/places_test.py
```diff
@@ -0,0 +1,84 @@
...
+@dataset(["main_routing_test"])
+class TestPlaces(AbstractTestFixture):
+    """
+    Test places responses
+    """
+
+    def test_places_by_id(self):
+        """can we get the complete address from coords"""
+
+        lon = 10. / 111319.9
```
antoine-de, line 45: can you put a comment about this magic number ?

source/jormungandr/tests/check_utils.py
```diff
@@ -547,12 +547,40 @@ def is_valid_line(line, depth_check=1):
...
+    else:
+        assert(False, "invalid type")
+
+
+def is_valid_address(address, depth_check=1):
+    get_not_null(address, "id")
+    get_not_null(address, "house_number")
+    get_not_null(address, "name")
+    get_not_null(address, "administrative_regions") # TODO test
+    get_not_null(address, "coord") # TODO test
```
antoine-de: you can use is_valid_{lon,lat} for the coords (maybe it's better to embed that in a is_valid_coord)

PR #660, opened 2014-10-24
Install update

antoine-de:
correction of the install script

migration to alembic

fix alembic bug regarding schema handling

### DIFF
--- a/build_navitia.sh
+++ b/build_navitia.sh
@@ -287,6 +287,8 @@ EOFJ
 # the Jormungnandr configuration is in the source/jormungandr/jormungandr/default_settings.py file
 # should be almost enough for the moment, we just need to change the location of the krakens configuration
 sed "s,^INSTANCES_DIR.*,INSTANCES_DIR = '$run_dir/jormungandr'," "$navitia_dir"/source/jormungandr/jormungandr/default_settings.py > "$run_dir"/jormungandr_settings.py
+#we also don't want to depend on the jormungandr database for this test
+sed -i 's/DISABLE_DATABASE.*/DISABLE_DATABASE=False/' "$run_dir"/jormungandr_settings.py
 
 JORMUNGANDR_CONFIG_FILE="$run_dir"/jormungandr_settings.py PYTHONPATH="$navitia_dir/source/navitiacommon:$navitia_dir/source/jormungandr" python "$navitia_dir"/source/jormungandr/jormungandr/manage.py runserver -d -r & 
 
--- a/build_navitia.sh
+++ b/build_navitia.sh
@@ -162,10 +162,10 @@ then
 
     # then you need to install all python dependencies: ::
 
-    #we want a custom protobuff version
-    sudo pip install protobuf==2.5.0
     sudo pip install -r $navitia_dir/source/jormungandr/requirements.txt
     sudo pip install -r $navitia_dir/source/tyr/requirements.txt
+    #we want a custom protobuff version
+    sudo pip install -U protobuf==2.5.0
 fi
 
 #the build procedure is explained is the install documentation
--- a/build_navitia.sh
+++ b/build_navitia.sh
@@ -43,7 +43,7 @@ clean_exit()
 }
 
 kraken_db_user_password=
-navitia_dir=`dirname $(readlink -f $0)`
+navitia_dir="$(dirname $(readlink -f $0))"
 gtfs_data_dir=
 osm_file=
 
@@ -76,19 +76,19 @@ EOF
 
 while getopts “hp:g:o:nc” OPTION
 do
-     case $OPTION in
+     case "$OPTION" in
          h)
              usage
              exit 1
              ;;
          p)
-             kraken_db_user_password=$OPTARG
+             kraken_db_user_password="$OPTARG"
              ;;
          g)
-             gtfs_data_dir=$OPTARG
+             gtfs_data_dir="$OPTARG"
              ;;
          o)
-             osm_file=$OPTARG
+             osm_file="$OPTARG"
              ;;
          n)
              install_dependencies=
@@ -123,8 +123,8 @@ then
     osm_file=/tmp/paris.osm.pbf
 fi
 
-run_dir=$navitia_dir/run
-mkdir -p $run_dir
+run_dir="$navitia_dir"/run
+mkdir -p "$run_dir"
 
 #Hack
 #for convenience reason, some submodule links are in ssh (easier to push)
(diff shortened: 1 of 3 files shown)


build_navitia.sh
```diff
@@ -134,15 +144,32 @@ git submodule update --init
...
     sudo apt-get install -y git g++ cmake liblog4cplus-dev libzmq-dev libosmpbf-dev libboost-all-dev libpqxx3-dev libgoogle-perftools-dev libprotobuf-dev python-pip libproj-dev protobuf-compiler libgeos-c1 
-    
-    sudo apt-get install -y postgresql-9.3 postgresql-9.3-postgis-2.1 #Note: postgres 9.1 and postgis 2.0 would be enough, be postgis 2.1 is easier ton setup 
-    
+
+    postgresql_package='postgresql-9.3'
+    postgresql_postgis_package='postgis postgresql-9.3-postgis-2.1 postgresql-9.3-postgis-scripts'
+    distrib=`lsb_release -si`
+
+    if [ "$distrib" = "Debian" ]; then
```
TeXitoi: must check for wheezy with, for example, `grep -q '^7\.' /etc/debian_version`

``` sh
if [ "$distrib" = "Debian" ] && grep -q '^7\.' /etc/debian_version; then
```

build_navitia.sh
```diff
@@ -134,15 +144,32 @@ git submodule update --init
...
+            # on Debian, we must add the APT repository of PostgreSQL project
+            # to have the right version of postgis
+            # no magic stuff : https://wiki.postgresql.org/wiki/Apt#PostgreSQL_packages_for_Debian_and_Ubuntu
+            apt_file='/etc/apt/sources.list.d/postgresql.list'
+            echo "deb http://apt.postgresql.org/pub/repos/apt/ wheezy-pgdg main" >> $apt_file
```
TeXitoi: need sudo

``` sh
echo "deb http://apt.postgresql.org/pub/repos/apt/ wheezy-pgdg main" | sudo tee "$apt_file"
```

build_navitia.sh
```diff
@@ -166,27 +193,30 @@ db_owner='navitia'
 
 kraken_db_name='navitia'
 # for the default build we give ownership of the base to a 'navitia' user, but you can do whatever you want here
-sudo su postgres -c "psql postgres -tAc \"SELECT 1 FROM pg_roles WHERE rolname='$db_owner'\""  # we check if there is already a user 
-if [ ! $? ]
+encap=`sudo su - postgres -c "psql postgres -tAc \"SELECT 1 FROM pg_roles WHERE rolname='$db_owner'\""`  # we check if there is already a user 
+if [ -z $encap ]
 then
-sudo su postgres -c "psql -c \"create user $db_owner;alter user $db_owner password '$kraken_db_user_password';\""
+sudo su - postgres -c "psql -c \"create user $db_owner;alter user $db_owner password '$kraken_db_user_password';\""
```
TeXitoi: Why all these `-` ? We don't need a login shell, don't we? and `sudo -u user` is much better: no double escaping.

build_navitia.sh
```diff
@@ -166,27 +193,30 @@ db_owner='navitia'
...
 else
 echo "user $db_owner already exists"
 fi
 
-if ! sudo su postgres -c "psql -l" | grep "^ $kraken_db_name "
+if ! sudo su - postgres -c "bash -c 'psql -l | grep \"^ ${kraken_db_name}\"'"
```
TeXitoi:
``` sh
if ! sudo -u postgres psql -l | grep -q "^ ${kraken_db_name}"
```

build_navitia.sh
```diff
@@ -273,3 +303,6 @@ echo "curl localhost:5000/v1/coverage/default/stop_areas"
 #we block the script for the user to test the api
 read -p "when you are finished, hit  a key to close kraken and jormungandr" n
 clean_exit
+
+# cleaning APT repository if -c option was specified
+test -n $clean_apt && rm -f $apt_file && printf "Option -c was specified, removing %s" "$apt_file"
```
TeXitoi: lack a `\n` in the last `printf`

build_navitia.sh
fdegomme-canaltp: Yes, better condition. If you tried it (shell syntax…), it's ok :)

build_navitia.sh
fdegomme-canaltp: You can use : 
sudo /bin/sh -c "echo 'deb http://apt.postgresql.org/pub/repos/apt/ wheezy-pgdg main' > $apt_file"

More dynamic :
sudo /bin/sh -c "echo \"deb http://apt.postgresql.org/pub/repos/apt/ $(lsb_release -sc)-pgdg main\" > $apt_file"

build_navitia.sh
fdegomme-canaltp: The removed method in this commit did not work. IIRC, it was about the return code sent by psql. This one does not send the return code we need so i preferred to encapsulate the command output (empty or not) in a variable "encap" that i check just after.

About the "-", it's not mandatory but i prefer login shells (no psql/shell history issue about permissions, …). For example :

<pre>
if ! sudo su postgres -c "psql -l" | grep "^ $kraken_db_name"; then echo "db exists"; fi
could not change directory to "/root": Permission denied
 navitia   | navitia  | UTF8     | en_US.UTF-8 | en_US.UTF-8
</pre>


With login shell :

<pre>
# if ! sudo su - postgres -c "psql -l" | grep "^ $kraken_db_name"; then echo "db exists"; fi
 navitia   | navitia  | UTF8     | en_US.UTF-8 | en_US.UTF-8 | 
</pre>


 About escaping, it's not bound to "-". i could use simple quote for psql command but we need variables.

build_navitia.sh
fdegomme-canaltp: Does not work.

build_navitia.sh
fdegomme-canaltp: You're right. I noticed that but the commit was already done :)

build_navitia.sh
TeXitoi: My point about excaping is that using sudo to call su, your command will be parsed 2 times by bash. Using `sudo -u user command arg` bash will only parse your command once, witch is much more readable.

`sudo su - user "foo \"escaped command\""` <=> `sudo -i -u user foo "excaped command"`

Then, if you prefere a login shell, OK for me.

build_navitia.sh
TeXitoi:
```
gupinot@par-lap04:~/dev/navitia$ if ! sudo -u postgres psql -l | grep -q "^ toto"; then echo true; else echo false; fi
true
gupinot@par-lap04:~/dev/navitia$ if ! sudo -u postgres psql -l | grep -q "^ bretagne"; then echo true; else echo false; fi
false
gupinot@par-lap04:~/dev/navitia$
```

Works here, what's your problem with it?

fdegomme-canaltp: It was a mistake. You're right it works but we should add "-i" to sudo command :

<pre>if ! sudo -i -u postgres psql -l | grep -q "^ ${kraken_db_name}"; then echo true; else echo false; fi</pre>


Without "-i" option i get an error/warning : <pre>could not change directory to "/root": Permission denied</pre> but i have a result even.

build_navitia.sh
fdegomme-canaltp: Sorry i just forgot the "-i" option in sudo. I agree, it's better than 'sudo su -'. However, be sure it works.

PR #663, opened 2014-10-27
Fix less fallback

l-vincent-l:
There were several mistakes made.
We were comparing arrival time with departure_time in computation of lost_time.
Also, we were computing departure walking time with the destination journey pattern point...

### DIFF
--- a/source/routing/raptor_solutions.cpp
+++ b/source/routing/raptor_solutions.cpp
@@ -285,12 +285,18 @@ get_final_jppidx_and_date(int count, type::idx_t jpp_idx, bool clockwise, bool d
 
 struct VisitorWalkingTime : public BasePathVisitor {
     navitia::time_duration walking_time = {};
+    type::idx_t departure_jpp_idx = type::invalid_idx;
     void connection(type::StopPoint* , type::StopPoint* ,
                 boost::posix_time::ptime dep_time, boost::posix_time::ptime arr_time,
                 type::StopPointConnection*) {
 
         walking_time += navitia::seconds((arr_time - dep_time).total_seconds());
     }
+
+    void final_step(type::idx_t current_jpp, size_t , const std::vector<std::vector<Label>> &){
+        departure_jpp_idx = current_jpp;
+    }
+
 };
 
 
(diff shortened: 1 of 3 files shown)


source/routing/raptor_solutions.cpp
```diff
@@ -151,9 +151,11 @@ get_pareto_front(bool clockwise, const std::vector<std::pair<type::idx_t, naviti
             for(auto spid_dep : departs) {
                 if(raptor.data.pt_data->journey_pattern_points[final_jpp_idx]->stop_point->idx == spid_dep.first) {
                     if(clockwise) {
-                        s.upper_bound = s.upper_bound + (spid_dep.second.total_seconds());
+                        s.upper_bound = s.upper_bound + spid_dep.second.total_seconds();
+                        s.total_arrival = raptor.labels[round][best_jpp].dt_pt - spid_dep.second.total_seconds();
                     }else {
                         s.upper_bound = s.upper_bound - (spid_dep.second.total_seconds());
```
TeXitoi: you can also remote parenthesis here.

TeXitoi:
Is https://github.com/CanalTP/navitia/blob/dev/source/routing/raptor.cpp#L247 still needed with this patch? (just return true in the lambda and check the unit tests to check it)

l-vincent-l:
why is it related?

source/routing/raptor_solutions.cpp
```diff
@@ -296,7 +305,7 @@ navitia::time_duration getWalkingTime(int count, type::idx_t jpp_idx, const std:
                      bool clockwise, bool disruption_active, const type::AccessibiliteParams & accessibilite_params,
                      const RAPTOR& raptor) {
 
-    const type::JourneyPatternPoint* current_jpp = raptor.data.pt_data->journey_pattern_points[jpp_idx];
+    const auto* current_jpp = raptor.data.pt_data->journey_pattern_points[jpp_idx];
```
antoine-de, line 308: the '*' is not mandatory

TeXitoi, line 308: it's not true: `const auto` wll give `type::JourneyPatternPoint * const`, and `const auto *` will give `type::JourneyPatternPoint const *`

l-vincent-l:
So this https://github.com/CanalTP/navitia/blob/dev/source/routing/raptor.cpp#L247 is still needed.

PR #680, opened 2014-10-30
ed: shift frequencies start_time in [0, 24:00[

TeXitoi:


antoine-de:
fix http://jira.canaltp.fr/browse/NAVITIAII-1247

PR #683, opened 2014-10-31
jormungandr: fix rounding problems with coords as id

TeXitoi:
fix http://jira.canaltp.fr/browse/NAVITIAII-847


source/jormungandr/jormungandr/interfaces/v1/Places.py
```diff
@@ -94,17 +95,7 @@ class PlaceUri(ResourceUri):
...
-            try:
-                args["uri"] = "coord:" + str(float(lon)) + ":"
-                args["uri"] += str(float(lat))
-            except ValueError:
-                pass
-        elif id[:5] == "admin":
-            args["uri"] = "admin:" + id.split(":")[-1]
-        if not "uri" in args.keys():
-            args["uri"] = id
+        args["uri"] = transform_id(id)
```
antoine-de: would'nt just `args = {'uri': transform_id(id)}` be simpler ?

PR #685, opened 2014-10-31
update config.h to limit compilation time

TeXitoi:


source/conf.h
```diff
@@ -0,0 +1,9 @@
+#pragma once
+
+namespace navitia { namespace config {
+
+extern const char * fixtures_dir;
```
antoine-de: aaaaaaaaaaaaaarg 2 spaces ?

PR #693, opened 2014-11-04
Destineo sort

kinnou02:
add a custom sort for destineo: http://jira.canaltp.fr/browse/NAVITIAII-1271


source/jormungandr/jormungandr/scenarios/destineo.py
```diff
@@ -29,6 +29,164 @@
...
+from operator import itemgetter
+from datetime import datetime, timedelta
+
+def is_pure_tc(journey):
+    has_tc = False
+    has_non_walking = False
+    for section in journey.sections:
+        if section.type in (response_pb2.STREET_NETWORK, response_pb2.CROW_FLY) \
+                and section.street_network.mode in (response_pb2.Bike, response_pb2.Car, response_pb2.Bss):
+            has_non_walking = True
```
antoine-de: you can return early there no ?

source/jormungandr/jormungandr/scenarios/destineo.py
```diff
@@ -29,6 +29,164 @@
...
+        elif is_non_pt_walk(j2):
+            return 1
+
+        #the bike only trip
+        if is_non_pt_bike(j1) and is_non_pt_bike(j2):
+            return self.sort_by_time(j1, j2)
+        elif is_non_pt_bike(j1):
+            return -1
+        elif is_non_pt_bike(j2):
+            return 1
```
antoine-de, line 188: don't you need a default return 0 ?

source/jormungandr/jormungandr/scenarios/destineo.py
```diff
@@ -68,3 +234,30 @@ def _remove_car_if_possible(self, response):
...
+        datetime_before = None
+        datetime_after = None
+        prev_journey = None
+        next_journey = None
+        filter_journey = lambda journey: 'is_pure_tc' in journey['tags']
+        try:
+            list_journeys = filter(filter_journey, resp['journeys'])
+            prev_journey = min(list_journeys, key=itemgetter('arrival_date_time'))
+            next_journey = max(list_journeys, key=itemgetter('departure_date_time'))
+        except:
```
antoine-de: why this wide catch block ? don't you just have to check for the non emptiness of list_journeys ?

source/jormungandr/jormungandr/scenarios/destineo.py
```diff
@@ -68,3 +234,30 @@ def _remove_car_if_possible(self, response):
...
+            response.journeys.sort(DestineoJourneySorter(clockwise=clockwise))
+
+    def extremes(self, resp):
+        logger = logging.getLogger(__name__)
+        logger.debug('calling extremes for destineo')
+        datetime_before = None
+        datetime_after = None
+        prev_journey = None
+        next_journey = None
+        filter_journey = lambda journey: 'is_pure_tc' in journey['tags']
```
antoine-de: so for destineo if you have a query with only bss (or bike, car) you won't get prev/next ?

source/jormungandr/tests/routing_tests_destineo.py
```diff
@@ -0,0 +1,344 @@
+# Copyright (c) 2001-2014, Canal TP and/or its affiliates. All rights reserved.
```
antoine-de, line 1: lot's of duplicate here, don't you think we can factorize ?

like call directly the JourneyTest methods (or find a way to copy it, but then I don't know how nose will react)

source/jormungandr/jormungandr/scenarios/destineo.py
kinnou02: good point, I need to check the specification for that point

source/jormungandr/tests/routing_tests_destineo.py
```diff
@@ -0,0 +1,66 @@
...
+import logging
+
+from tests_mechanism import AbstractTestFixture, dataset
+from routing_tests import TestJourneys
+from check_utils import *
+from nose.tools import eq_
+import jormungandr.scenarios.destineo
+from jormungandr.instance import Instance
+
+def check_links(response):
```
antoine-de: useless now

source/jormungandr/tests/routing_tests_destineo.py
```diff
@@ -0,0 +1,87 @@
...
+        eq_(len(response['journeys']), 2)
+        eq_(response['journeys'][0]['type'], 'best')
+        eq_(response['journeys'][1]['type'], 'non_pt_walk')
+
+    def test_journeys_destineo_with_bss(self):
+        #NOTE: we query /v1/coverage/main_routing_test/journeys and not directly /v1/journeys
+        #not to use the jormungandr database
+        query = journey_basic_query + "&first_section_mode=bss&first_section_mode=bike&first_section_mode=car" \
+                "&last_section_mode=bss&last_section_mode=bike&last_section_mode=car"
+        print query
```
antoine-de: old log

PR #697, opened 2014-11-04
Fix pareto front

l-vincent-l:
We were missing some solution of the pareto front


source/routing/tests/raptor_test.cpp
```diff
@@ -1305,3 +1305,57 @@ BOOST_AUTO_TEST_CASE(less_fallback) {
...
+ *      Line 1:
+ *      stop1   stop2
+ *      9h      9h50
+ *
+ *      Line 2:
+ *      stop2   stop3
+ *      9h55    10h00
+ *
+ * walking times:
+ *  B->I: 15 minutes
```
antoine-de: what's B, C and I ?

PR #701, opened 2014-11-06
Tyr: only keep data_set correctly integrated

kinnou02:
last_dataset now return only data_set associated to a job with a state
at "done"


guikcd:
Thx !

PR #706, opened 2014-11-13
Kraken: remove warning

antoine-de:


source/time_tables/departure_boards.cpp
```diff
@@ -185,10 +185,6 @@ departure_board(const std::string& request,
     if(handler.journey_pattern_points.size() == 0)
         return handler.pb_response;
 
-    if (max_date_times < 0) {
-        fill_pb_error(pbnavitia::Error::bad_filter, "stop_schedules : value of max_date_times invalid", handler.pb_response.mutable_error());
-        return handler.pb_response;
-    }
```
TeXitoi, line 191: We've discused of this one: it should be checked in jormun before removing it...

antoine-de, line 191: yep it's done in jormun because the type of the param in the api is a `natural`, so you got a nice flask error with negative values

PR #707, opened 2014-11-13
remove journey without enough tc or alternative fallback

kinnou02:
This change is only for the scenario destineo.

We remove journey with less than 5 min (default value) of bike or car,
and for journey with a fallback in bike or car we require a minimum 5
min of public transport.
http://jira.canaltp.fr/browse/NAVITIAII-1288


kinnou02:
I will add a few test tomorow

kinnou02:
tests added

source/jormungandr/jormungandr/scenarios/destineo.py
```diff
@@ -106,6 +108,37 @@ def is_non_pt_bike(journey):
...
+    in_bss = False
+    for section in journey.sections:
+        if section.type == response_pb2.BSS_RENT:
+            in_bss = True
+        if section.type == response_pb2.BSS_PUT_BACK:
+            in_bss = False
+        if section.type in (response_pb2.STREET_NETWORK, response_pb2.CROW_FLY) \
+                and section.street_network.mode == response_pb2.Bike \
+                and not in_bss:
+            duration = duration + section.duration
```
antoine-de, line 122: += ?

source/jormungandr/jormungandr/scenarios/destineo.py
```diff
@@ -227,8 +262,47 @@ def journeys(self, request, instance):
...
 
 
+    def _remove_not_long_enough_fallback(self, journeys, instance):
+        to_delete = []
+        for idx, journey in enumerate(journeys):
+            if journey.type in non_pt_types:
+                continue
+            bike_dur = bike_duration(journey)
+            car_dur = car_duration(journey)
+            if bike_dur and bike_dur < instance.destineo_min_bike:
```
antoine-de, line 272: so you don't remove them only if you have better journeys ?

PR #711, opened 2014-11-14
add timeout on tyr jobs

kinnou02:
add a timeout on the lock

### DIFF
--- a/source/tyr/tyr/binarisation.py
+++ b/source/tyr/tyr/binarisation.py
@@ -42,6 +42,7 @@ import kombu
 from navitiacommon import models
 import shutil
 from tyr.helper import get_instance_logger
+from functools import wraps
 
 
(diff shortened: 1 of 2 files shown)


source/tyr/tyr/binarisation.py
```diff
@@ -327,7 +327,7 @@ def nav2rt(instance_config, job_id):
     """ Launch nav2rt"""
     job = models.Job.query.get(job_id)
     instance = job.instance
-    lock = redis.lock('tyr.lock|' + instance.name)
+    lock = redis.lock('tyr.lock|' + instance.name, timeout=10*60*60)
```
antoine-de: isn't it a bit short ?

TeXitoi:
all these hardcoded timeouts make me think that a proper abstraction is lacking...

PR #719, opened 2014-11-18
Type: StopTime is in VehicleJourney without indirection

TeXitoi:
This is the most instanciated object, thus we optimize it.

A Journey computation is 5% faster with this modification, and memory
almost the same (a little bit less).

### DIFF
--- a/source/ed/tests/adapted_test.cpp
+++ b/source/ed/tests/adapted_test.cpp
@@ -1597,12 +1597,12 @@ BOOST_AUTO_TEST_CASE(impact_stoppoint_0){
 
     messages.push_back(m);
 
-    BOOST_CHECK_EQUAL(b.data->pt_data->stop_times.size(), 3);
+    BOOST_CHECK_EQUAL(b.data->pt_data->nb_stop_times(), 3);
 
     AtAdaptedLoader loader;
     loader.apply(messages, *b.data->pt_data);
 
-    BOOST_CHECK_EQUAL(b.data->pt_data->stop_times.size(), 5);
+    BOOST_CHECK_EQUAL(b.data->pt_data->nb_stop_times(), 5);
 
     vj = b.data->pt_data->vehicle_journeys[0];
     BOOST_CHECK_EQUAL(vj->uri,  "vehicle_journey:vj1");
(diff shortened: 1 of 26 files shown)


source/ed/tests/adapted_test.cpp
```diff
@@ -2242,10 +2242,10 @@ BOOST_AUTO_TEST_CASE(impact_stoppoint_4){
     m.active_days = nt::Ven | nt::Lun;
     messages.push_back(m);
 
-    BOOST_CHECK_EQUAL(b.data->pt_data->stop_times.size(), 2);
+    //BOOST_CHECK_EQUAL(b.data->pt_data->stop_times.size(), 2);
     AtAdaptedLoader loader;
     loader.apply(messages, *b.data->pt_data);
-    BOOST_CHECK_EQUAL(b.data->pt_data->stop_times.size(), 4);
+    //BOOST_CHECK_EQUAL(b.data->pt_data->stop_times.size(), 4);
```
l-vincent-l: can you make a stop_times_size() function?

source/routing/dataraptor.h
```diff
@@ -49,8 +49,8 @@ struct dataRAPTOR {
     std::vector<pair_int> footpath_index_backward;
     std::vector<uint32_t> arrival_times;
     std::vector<uint32_t> departure_times;
-    std::vector<type::StopTime*> st_idx_forward;// Nom a changer ce ne sont plus des idx mais des pointeurs
-    std::vector<type::StopTime*> st_idx_backward;
+    std::vector<const type::StopTime*> st_idx_forward;// Nom a changer ce ne sont plus des idx mais des pointeurs
```
antoine-de: I think it's time to change the comment :wink:

antoine-de: Ok, it is not nice to ask you that, but I think some more comment about st_idx_forward and st_idx_backward would be nice 

(and maybe some explanation about why the pointer to a vector contained object is not that bad in this case)

antoine-de:
ok for me

source/routing/dataraptor.h
```diff
@@ -47,10 +47,24 @@ struct dataRAPTOR {
     std::vector<pair_int> footpath_index_forward;
     std::vector<const navitia::type::StopPointConnection*> foot_path_backward;
     std::vector<pair_int> footpath_index_backward;
-    std::vector<uint32_t> arrival_times;
-    std::vector<uint32_t> departure_times;
-    std::vector<type::StopTime*> st_idx_forward;// Nom a changer ce ne sont plus des idx mais des pointeurs
-    std::vector<type::StopTime*> st_idx_backward;
+
+    // arrival_times (resp. departure_times) are the different arrival
+    // (resp. departure) times of each stop points sorted by
```
l-vincent-l: stop time

PR #720, opened 2014-11-18
Streetnetwork: overflow on time_duration

antoine-de:
fix http://jira.canaltp.fr/browse/NAVITIAII-1334

there was bad data leading to huge edge in geopal.

reduce time_duration precision and most of all add a check on overflow not to import huge edge.

We could have split the too-big edge but since this part is meant to be refactored in the short term it is not worth it

### DIFF
--- a/source/ed/ed_reader.cpp
+++ b/source/ed/ed_reader.cpp
@@ -1059,14 +1059,13 @@ void EdReader::fill_graph(navitia::type::Data& data, pqxx::work& work) {
         navitia::georef::Edge e;
         float len = const_it["leng"].as<float>();
         e.way_idx = way->idx;
-        way->edges.push_back(std::make_pair(source, target));
 
-        //TODO et les pietons ??!
+        if (const_it["pede"].as<bool>()) {
+            e.duration = navitia::seconds(len / ng::default_speed[nt::Mode_e::Walking]);
 
-        e.duration = navitia::seconds(len / ng::default_speed[nt::Mode_e::Walking]);
-        boost::add_edge(source, target, e, data.geo_ref->graph);
-        nb_walking_edges++;
-
+            way->edges.push_back(std::make_pair(source, target));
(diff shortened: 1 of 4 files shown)


source/ed/ed_reader.cpp
```diff
@@ -1059,14 +1059,19 @@ void EdReader::fill_graph(navitia::type::Data& data, pqxx::work& work){
...
-        //TODO et les pietons ??!
 
-        e.duration = navitia::seconds(len / ng::default_speed[nt::Mode_e::Walking]);
-        boost::add_edge(source, target, e, data.geo_ref->graph);
-        nb_walking_edges++;
+        if (const_it["pede"].as<bool>()) {
+            e.duration = navitia::seconds(len / ng::default_speed[nt::Mode_e::Walking]);
 
+            // overflow check since we want to store that on a int32
+            if (e.duration.total_seconds() + 1 < len / ng::default_speed[nt::Mode_e::Walking]) {
```
l-vincent-l: You may do e.duration.total_seconds() < std::floor(len / ng::default_speed[nt::Mode_e::Walking])

source/ed/ed_reader.cpp
```diff
@@ -1059,14 +1059,19 @@ void EdReader::fill_graph(navitia::type::Data& data, pqxx::work& work){
...
         float len = const_it["leng"].as<float>();
         e.way_idx = way->idx;
-        way->edges.push_back(std::make_pair(source, target));
-
-        //TODO et les pietons ??!
 
-        e.duration = navitia::seconds(len / ng::default_speed[nt::Mode_e::Walking]);
-        boost::add_edge(source, target, e, data.geo_ref->graph);
-        nb_walking_edges++;
+        if (const_it["pede"].as<bool>()) {
```
TeXitoi, line 1063: We need to check it for each mode.

PR #725, opened 2014-11-19
Routing modularization proposal

xavierraffin:
We aim to make Jormungandr more flexible.
This is a functional (almost) non intrusive modification which add a module system to Jormungandr.

All modules are now stored in modules directory.

A module have a namespace which is an url prefix.
That avoids route collision between modules

Navitia.io API is now a module prefixed by v1

The Jormungandr behaviour is unchanged

Future improvements
- generate dynamicaly the Jormungandr entry point http://api.navitia.io with modules information
- add some other example modules ?


buildbotctp:

antoine-de:
jenkins, test this please

antoine-de:
thanks @xavierraffin we're looking into it

source/jormungandr/jormungandr/modules/__init__.py
```diff
@@ -0,0 +1,32 @@
+#!/usr/bin/env python
```
antoine-de: you shouldn't put those 2 lines.

the file should not be in utf-8 and is not executable as is.

source/jormungandr/jormungandr/modules/v1_routing.py
```diff
@@ -0,0 +1,222 @@
...
+from jormungandr.interfaces.v1 import Ptobjects
+from jormungandr.interfaces.v1 import Coord
+from jormungandr.interfaces.v1 import Disruptions
+from jormungandr.interfaces.v1 import Calendars
+from jormungandr.interfaces.v1 import converters_collection_type
+from jormungandr.interfaces.v1 import Status
+from werkzeug.routing import BaseConverter, FloatConverter, PathConverter
+from jormungandr.modules_loader import AModule
+
+__author__ = 'Nicolas Garcia <nicolas.garcia@tisseo.fr>'
```
antoine-de: we usually don't put a **author** tag in the file, I don't think they're relevant.

source/jormungandr/jormungandr/modules/v2_routing.py
```diff
@@ -0,0 +1,42 @@
...
+# twitter @navitia
+# IRC #navitia on freenode
+# https://groups.google.com/d/forum/navitia
+# www.navitia.io
+from jormungandr.modules_loader import AModule
+
+__author__ = 'Nicolas Garcia <nicolas.garcia@tisseo.fr>'
+
+
+class V2Routing(AModule):
```
antoine-de: I think v2 should be for the next 'classic' api version.

what about a 'tisseo' end point ?

`like https://api.navitia.io/v1/...`
`like https://api.navitia.io/v2/...`
`like https://api.navitia.io/tisseo/...` 
? (it is open to debate, not sure about it)

nicolas-garcia:
Hi,
About the modularity, I've two questions :
First, in end_point_tests, you check if the number of versions in the endpoint is equal to 1.
But this test became deprecated due to the modularization, right ?

Do I have to remove this assertion ?

Finally, still in the same test, there is an assertion about endpoints status which can only be 'current' or 'deprecated', but this assertion isn't relevant in the case where we have the modularization.

What do I have to do with those tests ?

antoine-de:
Hi @nicolas-garcia,

you're right, those tests have to be updated following the modifications (and we'll have to find a new status for your api)

nicolas-garcia:
Thanks for your quick answer.

Due to the modularity of endpoints, the solution might be to remove the assertion 
`assert len(versions) == 1`, 
or do you have an other solution ?

antoine-de:
you can maybe update it with 
`assert len(versions) >= 1`

and check that at least one is the 'current'

xavierraffin:
The pull request has been updated

TeXitoi:
jenkins, test this please

kinnou02:
jenkins, test this please

xavierraffin:
So, do you will merge this pull-request ?

source/jormungandr/jormungandr/instance.py
```diff
@@ -81,12 +82,16 @@ def scenario(self):
             logger = logging.getLogger(__name__)
             instance_db = self._get_models()
             if instance_db:
-                logger.info('loading of scenario %s for instance %s', instance_db.scenario, self.name)
```
kinnou02, line 86: we use a line length at  something like 110 characters, it must readable in github without scrolling

source/jormungandr/jormungandr/api.py
```diff
@@ -46,5 +48,16 @@ def output_jsonp(data, code, headers=None):
...
     return resp
 
+
+rest_api.module_loader = ModulesLoader(rest_api)
+
+rest_api.module_loader.load(V1Routing(rest_api, 'v1',
+                                      'Current version of navitia API',
+                                      status='current'))
+# Uncomment the following lines if you want to activate the test module
+#rest_api.module_loader.load(TisseoRouting(rest_api, 'tisseo/test',
```
kinnou02: Do we really need to set the description and the status at the instantiation? I think it can be statically set in the constructor since it's a property of the module, we don't want to change it without a code modification

kinnou02: The list of enabled modules must be configurable with the settings.py file, something like this:

```
from jormungandr.modules.tisseo_routing.TisseoRouting import TisseoRouting
from jormungandr.modules.v1_routing.v1_routing import V1Routing
modules = {'v1': V1Routing, 'tisseo/test':  TisseoRouting}
```

then

```
for prefix, module in modules.iteritems():
    rest_api.module_loader.load(module(rest_api, prefix))
```

source/jormungandr/jormungandr/instance.py
nicolas-garcia, line 86: I note , I just applied pep8 standard strictly .
I'll fix that. Anyway, that file should not have been modified by the commit.

nicolas-garcia:
The PR has been updated according to previous comments of @kinnou02

source/jormungandr/jormungandr/api.py
```diff
@@ -46,5 +46,15 @@ def output_jsonp(data, code, headers=None):
...
+
+# If modules are configured, then load and run them
+if 'MODULES' in rest_api.app.config:
+    rest_api.module_loader = ModulesLoader(rest_api)
+    for prefix, module_info in rest_api.app.config['MODULES'].iteritems():
+        module_file = importlib.import_module(module_info['import_path'])
+        module = getattr(module_file, module_info['class_name'])
+        rest_api.module_loader.load(module(rest_api, prefix))
+
+    rest_api.module_loader.run()
```
antoine-de, line 56: since nothing will be done if empty, it might be good to log in that case

source/jormungandr/jormungandr/module_resource.py
```diff
@@ -0,0 +1,65 @@
...
+            Register the resource into the ModuleResourcesManager
+            Inject the module name into the resource
+        :param resource:
+        :type resource: ModuleResource
+        :raise TypeError: Raise TypeError if resource is not an instance
+                of ModuleResource
+        """
+        if not isinstance(resource, ModuleResource):
+            raise TypeError('Expected type {0}, got {1} instead'
+                            .format('ModuleResource',
```
antoine-de: if ModuleResource is hard coded, you can put it directly in the string no ?

kinnou02:
jenkins, test this please

buildbotctp:
thanks @xavierraffin  and @nicolas-garcia !

PR #727, opened 2014-11-20
Ed: fix huge bug on walking edges not added

antoine-de:


source/ed/ed_reader.cpp
```diff
@@ -1058,6 +1058,7 @@ void EdReader::fill_graph(navitia::type::Data& data, pqxx::work& work) {
 
             // overflow check since we want to store that on a int32
             if (e.duration.total_seconds() == std::floor(len / ng::default_speed[nt::Mode_e::Walking])) {
+                boost::add_edge(bike_source, bike_target, e, data.geo_ref->graph);
```
koch-t: copy/pase error. You meant source and target. bike_source and bike_target are not defined in this block

PR #731, opened 2014-11-21
Type: stop time comments were not saved

antoine-de:


TeXitoi:
This change is present in #730.

PR #734, opened 2014-11-21
Include line and routeref in jp equal check

koch-t:
Without this check, ed would cluster vehicle_journeys with the same sequence of stop_point's into the same journey_pattern regardless of route/line.
This resulted in wrongly clustered journey-patterns where the wrong line number was presented


antoine-de:
thanks again @skywave !

PR #735, opened 2014-11-24
Fix sorting of vj in route schedules

l-vincent-l:
We were only comparing vj with their first stop_time defined. So with this case:

```
A    7h00
B    8h00   7h30      
C           8h00
```

And we actually want: 

```
A         7h00
B   7h30  8h00
C   8h00
```

### DIFF
--- a/source/ed/build_helper.cpp
+++ b/source/ed/build_helper.cpp
@@ -70,7 +70,7 @@ nt::MetaVehicleJourney* get_or_create_metavj(builder& b, const std::string name)
     return it->second;
 }
 
-nt::JourneyPattern* get_or_create_journey_pattern(builder&b, std::string uri) {
+nt::JourneyPattern* get_or_create_journey_pattern(builder& b, const std::string& uri) {
     auto it = std::find_if(b.data->pt_data->journey_patterns.begin(),
                         b.data->pt_data->journey_patterns.end(),
                         [uri](nt::JourneyPattern* jp) {
--- a/source/ed/build_helper.cpp
+++ b/source/ed/build_helper.cpp
@@ -70,7 +70,26 @@ nt::MetaVehicleJourney* get_or_create_metavj(builder& b, const std::string name)
     return it->second;
 }
 
-VJ::VJ(builder & b, const std::string &line_name, const std::string &validity_pattern, const std::string &/*block_id*/, bool wheelchair_boarding, const std::string& uri, std::string meta_vj_name) : b(b){
+nt::JourneyPattern* get_or_create_journey_pattern(builder&b, std::string uri) {
+    auto it = std::find_if(b.data->pt_data->journey_patterns.begin(),
+                        b.data->pt_data->journey_patterns.end(),
+                        [uri](nt::JourneyPattern* jp) {
+                            return jp->uri == uri;
+                        });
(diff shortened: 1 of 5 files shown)


source/ed/build_helper.cpp
```diff
@@ -70,7 +70,26 @@ nt::MetaVehicleJourney* get_or_create_metavj(builder& b, const std::string name)
     return it->second;
 }
 
-VJ::VJ(builder & b, const std::string &line_name, const std::string &validity_pattern, const std::string &/*block_id*/, bool wheelchair_boarding, const std::string& uri, std::string meta_vj_name) : b(b){
+nt::JourneyPattern* get_or_create_journey_pattern(builder&b, std::string uri) {
```
antoine-de: there is a missing space there (and while you're on it, you can const& the string (and use boost::find_if :) )

source/time_tables/route_schedules.cpp
```diff
@@ -77,51 +78,106 @@ get_all_stop_times(const vector_idx &journey_patterns,
...
+    std::vector<std::vector<datetime_stop_time> > result, tmp;
+    const size_t thermometer_size = thermometer.get_thermometer().size();
+    for (size_t i=0; i < stop_times.size(); ++i) {
+        tmp.push_back(std::vector<datetime_stop_time>(thermometer_size));
     }
 
-    //On remplit le tableau
+    // We match every stop_time with the journey pattern
     int y=0;
     for(std::vector<datetime_stop_time> vec : stop_times) {
```
antoine-de: const & would be better

source/time_tables/route_schedules.cpp
```diff
@@ -77,51 +78,106 @@ get_all_stop_times(const vector_idx &journey_patterns,
...
+    for (size_t i=0; i < stop_times.size(); ++i) {
+        tmp.push_back(std::vector<datetime_stop_time>(thermometer_size));
     }
 
-    //On remplit le tableau
+    // We match every stop_time with the journey pattern
     int y=0;
-    for(std::vector<datetime_stop_time> vec : stop_times) {
+    for(const std::vector<datetime_stop_time>& vec : stop_times) {
         auto jpp = *vec.front().second->vehicle_journey->journey_pattern;
```
antoine-de: it's a jp there no ?

antoine-de: not very important but the journey pattern is copied there.

source/time_tables/route_schedules.cpp
```diff
@@ -77,51 +78,106 @@ get_all_stop_times(const vector_idx &journey_patterns,
...
-    for(const std::vector<datetime_stop_time> vec : stop_times){
-        type::VehicleJourney* vj = vec.front().second->vehicle_journey;
-        auto it = std::find_if(result.begin(), result.end(),
-                               [vj](type::VehicleJourney* vj1){
-                                return (vj->idx == vj1->idx);
-                        });
-        if(it == result.end()){
-            result.push_back(vj);
+    for(unsigned int i=0; i<thermometer_size; ++i) {
+        result.push_back(std::vector<datetime_stop_time>());
```
antoine-de: isn't it easier to allocate directly the vector ?

PR #736, opened 2014-11-24
 Remove journey with more fallback than the matching non_pt 

kinnou02:
For exemple with the walking, if the non_pt_walk has a duration of 12
minutes, the following journey will be masked:
    - walking 14min + TC
    - walking 8min + TC + walking 6min

refer to: http://jira.canaltp.fr/browse/NAVITIAII-1304


kinnou02:
if you have a better idea of name for `_delete_non_optimal_journey` I'm interested

kinnou02:
jenkins, test this please

source/jormungandr/jormungandr/scenarios/default.py
```diff
@@ -287,6 +290,27 @@ def fill_journeys(self, pb_req, request, instance):
...
+        type_func = {'non_pt_bss': bss_duration, 'non_pt_bike': bike_duration, 'non_pt_walk': walking_duration}
+        to_delete = []
+        for journey in journeys:
+            if journey.type in type_journeys:
+                type_journeys[journey.type] = journey
+        for idx, journey in enumerate(journeys):
+            for type, func in type_func.iteritems():
+                if not type_journeys[type] or type_journeys[type] == journey:
+                    continue
+                if func(journey) > type_journeys[type].duration:
```
antoine-de: isn't it clearer to compare it with func(type_journeys[type]) ?

I know it is the same thing but it would be easier to read AFAIK.

And a bit of comment on this function wouldn't hurt.

source/jormungandr/jormungandr/scenarios/helpers.py
```diff
@@ -131,6 +131,23 @@ def bike_duration(journey):
 
     return duration
 
+def bss_duration(journey):
```
antoine-de, line 134: don't we want to take the walking duration into account for BSS ?

kinnou02, line 134: I don't think so

antoine-de:
OK for me

PR #737, opened 2014-11-24
jormungandr: add an id on each request and log this id for traceability

kinnou02:
Each request as now an id that can be displayed in the log


source/jormungandr/jormungandr/helper.py
```diff
@@ -53,3 +55,11 @@ def __call__(self, environ, start_response):
             environ['wsgi.url_scheme'] = scheme
         return self.app(environ, start_response)
 
+
+class Request(Request):
```
l-vincent-l: Maybe you can call it NavitiaRequest so we're sure we're using it.

PR #738, opened 2014-11-24
Disruption: fix some bugs on the whole at chain

antoine-de:


source/jormungandr/jormungandr/interfaces/v1/fields.py
```diff
@@ -380,6 +379,7 @@ def output(self, key, obj):
     "messages": NonNullList(NonNullNested(disruption_message)),
 }
 
+
```
l-vincent-l, line 382: 2 new lines???

source/type/task.proto
```diff
@@ -11,5 +11,8 @@ enum Action{
 
 message Task{
     required Action action = 1;
+
+    optional pbnavitia.realtime.Message message = 3;
```
l-vincent-l, line 15: Why don't you use 2?

source/jormungandr/jormungandr/interfaces/v1/fields.py
antoine-de, line 382: yes it is pep8 compliant

source/jormungandr/jormungandr/interfaces/v1/fields.py
```diff
@@ -367,7 +367,6 @@ def output(self, key, obj):
     "text": fields.String(),
     "content_type": fields.String(),
 }
-
```
l-vincent-l, line 370: it's not pep8 compliant

PR #743, opened 2014-11-25
Jormungandr: fix error for wrongly formated tokens

antoine-de:
wrongly formated token were resulting with internal error, now we get a nicer error

### DIFF
--- a/source/jormungandr/jormungandr/authentication.py
+++ b/source/jormungandr/jormungandr/authentication.py
@@ -29,6 +29,7 @@
 # https://groups.google.com/d/forum/navitia
 # www.navitia.io
 import logging
+import binascii
 
 from flask_restful import reqparse, abort
 import flask_restful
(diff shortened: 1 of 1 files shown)


source/jormungandr/jormungandr/authentication.py
```diff
@@ -82,17 +82,20 @@ def get_token():
...
         return None
-
-    args = request.headers['Authorization'].split(' ')
+    auth = request.headers['Authorization']
+    args = auth.split(' ')
     if len(args) == 2:
+        b64 = args[1]
         try:
-            b64 = args[1]
             decoded = base64.decodestring(b64)
```
l-vincent-l, line 91: it can throw a binascii.Error

source/jormungandr/jormungandr/authentication.py
```diff
@@ -82,17 +82,20 @@ def get_token():
...
             return decoded.split(':')[0]
```
l-vincent-l, line 92: it doesn't throw error

antoine-de, line 92: you're right, done

PR #744, opened 2014-11-25
Alembic: change the bounding shape computation

antoine-de:
Remove the upsert for the bounding shape computation, since the
computation was done even if the shape_computed was false


TeXitoi:
OK for me

PR #746, opened 2014-11-26
 Kraken: We want to favoritize normal vj against stay_in vj 

l-vincent-l:
So we apply vj stay_in after applying normal vjs

### DIFF
--- a/source/routing/raptor.cpp
+++ b/source/routing/raptor.cpp
@@ -486,7 +486,7 @@ void RAPTOR::raptor_loop(Visitor visitor, const type::AccessibiliteParams & acce
         const auto & prec_labels=labels[count -1];
         auto& working_labels = labels[this->count];
         this->make_queue();
-
+        std::vector<std::tuple<const type::VehicleJourney*, type::idx_t, uint16_t, DateTime>> stay_in_vjs;
         for(const auto & journey_pattern : data.pt_data->journey_patterns) {
             if(Q[journey_pattern->idx] != std::numeric_limits<int>::max()
                     && Q[journey_pattern->idx] != -1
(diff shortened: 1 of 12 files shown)


source/routing/raptor.cpp
```diff
@@ -486,7 +485,7 @@ void RAPTOR::raptor_loop(Visitor visitor, const type::AccessibiliteParams & acce
         const auto & prec_labels=labels[count -1];
         auto& working_labels = labels[this->count];
         this->make_queue();
-
+        std::vector<RoutingState> states_stay_in;
```
TeXitoi, line 492: add a comment like "we stock stay_in extentions to apply them at the end to deprioritize stay_in" somewere

source/routing/raptor.cpp
```diff
@@ -558,12 +557,17 @@ void RAPTOR::raptor_loop(Visitor visitor, const type::AccessibiliteParams & acce
                     }
                 }
                 if(boarding) {
-                    end_algorithm &= !apply_vj_extension(visitor, global_pruning, it_st->vehicle_journey, boarding->idx,
-                        workingDt, l_zone, disruption_active);
+                    const type::VehicleJourney* vj_stay_in = visitor.get_extension_vj(it_st->vehicle_journey);
+                    if (vj_stay_in) {
+                        states_stay_in.emplace_back(vj_stay_in, boarding->idx, l_zone, workingDt);
+                    }
```
TeXitoi, line 567: don't we should 

``` c++
} else { end_algorithm &= !apply_vj_extension(visitor, global_pruning, disruption_active, state); }
```

or something like that? Or I didn't understand something?

l-vincent-l, line 567: get_extension_vj returns nullptr if there's no vj-extension, so if there's no vj extension we don't want to apply it.

PR #747, opened 2014-11-26
Do not index (for autocomplete) way without city associated

TeXitoi:
fix http://jira.canaltp.fr/browse/NAVITIAII-1359

### DIFF
--- a/source/georef/georef.cpp
+++ b/source/georef/georef.cpp
@@ -378,40 +378,37 @@ void GeoRef::build_proximity_list(){
     poi_proximity_list.build();
 }
 
+static const Admin* find_city_admin(const std::vector<Admin*>& admins) {
+    for(Admin* admin : admins){
+        //Level 8: City
+        if (admin->level == 8) { return admin; }
+    }
+    return nullptr;
+}
+
 void GeoRef::build_autocomplete_list(){
-    int pos = 0;
+    int pos = -1;
     fl_way.clear();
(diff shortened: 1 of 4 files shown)


source/georef/georef.cpp
```diff
@@ -378,22 +378,25 @@ void GeoRef::build_proximity_list(){
...
-            std::string key="";
-            for(Admin* admin : way->admin_list){
-                //Level Admin 8  : City
-                if (admin->level == 8) {
-                    key+= " " + admin->name;
-                }
-                if ((!admin->post_code.empty()) && (admin->level == 8)) {
-                    key += " "+ admin->post_code;
-                }
+            std::string key = "";
```
l-vincent-l: You can move the declaration in the if, and gain a level of indentation by using a && !! :+1:

source/georef/georef.cpp
```diff
@@ -404,11 +407,8 @@ void GeoRef::build_autocomplete_list(){
     for(const POI* poi : pois){
         if ((!poi->name.empty()) && (poi->visible)) {
             std::string key="";
-            for(Admin* admin : poi->admin_list) {
-                //Level Admin 8  : City
-                if (admin->level == 8) {
-                    key += " " + admin->name;
-                }
+            if (auto admin = find_city_admin(poi->admin_list)) {
+                key += " " + admin->name;
```
l-vincent-l: Same thing

PR #748, opened 2014-11-26
Jormungandr: add CORS support

l-vincent-l:
So it can call directly from a browser application.
But the headers are override by apache...


l-vincent-l:
It appears that apache actually doesn't override headers, BUT we have some extra-work to do because we are using http auth as said in : http://mortoray.com/2014/04/09/allowing-unlimited-access-with-cors/

PR #750, opened 2014-11-26
ed: fix bad admin with cities

TeXitoi:
fix http://jira.canaltp.fr/browse/NAVITIAII-1363


TeXitoi:
jenkins, test this please

PR #751, opened 2014-11-27
minimum duration of bss for destineo

kinnou02:
Remove journey with less bss than `destineo_min_bss`, this way we don't show a journey with only a few meters done in bss.
We also remove journey with only a few minutes of public transport after (or before) the bss.


kinnou02:
jenkins, test this please

PR #754, opened 2014-11-27
Kraken: Waiting time can be estimated

l-vincent-l:
waiting time for or after a vehicle with an estimated datetime, is also estimated.


source/routing/raptor_api.cpp
```diff
@@ -256,6 +267,19 @@ void add_pathes(EnhancedResponse &enhanced_response, const std::vector<navitia::
                     case waiting : pb_section->set_type(pbnavitia::WAITING); break;
                     default : pb_section->set_transfer_type(pbnavitia::walking); break;
                 }
+                // For a waiting section, if the previous public transport section,
+                // has estimated datetime we need to set it has estimated too.
+                if (pb_journey->sections_size() > 1) {
+                    for (int i=pb_journey->sections_size()-1; i>=0; --i) {
```
antoine-de, line 273: why the reverse loop ?

source/routing/raptor_api.cpp
```diff
@@ -256,6 +267,19 @@ void add_pathes(EnhancedResponse &enhanced_response, const std::vector<navitia::
...
+                        auto section = pb_journey->mutable_sections(i);
+                        if (section->type() == pbnavitia::PUBLIC_TRANSPORT) {
+                            if (section->add_info_vehicle_journey().has_date_time_estimated()) {
+                                pb_section->mutable_add_info_vehicle_journey()->set_has_date_time_estimated(true);
```
antoine-de, line 277: you can break no ?

and don't you want to look only for the previous PUBLIC_TRANSPORT section ?

source/routing/raptor_api.cpp
```diff
@@ -256,6 +267,19 @@ void add_pathes(EnhancedResponse &enhanced_response, const std::vector<navitia::
                     case waiting : pb_section->set_type(pbnavitia::WAITING); break;
                     default : pb_section->set_transfer_type(pbnavitia::walking); break;
                 }
+                // For a waiting section, if the previous public transport section,
+                // has estimated datetime we need to set it has estimated too.
+                if (pb_journey->sections_size() > 1) {
+                    for (int i=pb_journey->sections_size()-1; i>=0; --i) {
+                        auto section = pb_journey->mutable_sections(i);
```
antoine-de: not sure, but isn't there a not mutable section ? (I think it's just `pb_journey->mutable_sections(i)`)

source/routing/raptor_api.cpp
l-vincent-l, line 273: because we want to check the previous public transport section.
And before there could be a transfer or a public transport

PR #759, opened 2014-11-28
Fix arrival at isochron

djludo:


source/routing/raptor_api.cpp
```diff
@@ -750,7 +751,8 @@ make_nm_response(RAPTOR &raptor, const std::vector<type::EntryPoint> &origins,
 
                 std::vector<type::StopPoint*> stop_points;
                 for(auto& path : paths) {
-                    const type::StopPoint* sp = path.items[0].stop_points[0];
+                    const type::StopPoint* sp = clockwise ? path.items.front().stop_points.front() :
+                                                             path.items.back().stop_points.back();
                     stop_points.push_back(const_cast<type::StopPoint*>(sp));
```
TeXitoi, line 756: Please remove this `const_cast`, change `stop_points` to `std::vector<const type::StopPoint*>` and `add_isochrone_response` using `std::vector<const type::StopPoint*>`.

buildbotctp:

source/routing/raptor_api.cpp
djludo, line 756: the "make_isochron" method calls "add_isochron" with raptor.data.pt_data->stop_points as parameter, that's the reason of the const_cast

TeXitoi:
Jenkins, test this please

PR #760, opened 2014-11-28
Tyr: add param to give a file name for the data.nav.lz4

antoine-de:
This param can be given in the delivery process  to reduce downtime


source/tyr/tyr/binarisation.py
```diff
@@ -323,7 +323,7 @@ def ed2nav(instance_config, job_id):
 
 
 @celery.task()
-def nav2rt(instance_config, job_id):
+def nav2rt(instance_config, job_id, custom_output):
```
kinnou02: and ed2nav?

PR #761, opened 2014-11-28
Waiting time estimated

l-vincent-l:
waiting time for or after a vehicle with an estimated datetime, is also estimated.


TeXitoi:
OK for me

PR #762, opened 2014-11-28
Kraken: We want to favoritize normal vj against stay_in vj

l-vincent-l:
So we apply vj stay_in after applying normal vjs


TeXitoi:
OK for me

PR #768, opened 2014-12-02
build: add requirement.txt in deb

kinnou02:


antoine-de:
nice!

do you want to hotfix the 1.6.x to include that in the next release (this there are pip requirements changes) ?

PR #770, opened 2014-12-04
add metadatas on each kraken response, and update jormun instance each time

kinnou02:
At startup jormungandr don't load the configuration of a new instance
fast enougth, it  wait 10 second before each scan, and if there is
some dead instances we add another second for them.

with this patch at most one request will be done without the timezone set

### DIFF
--- a/source/jormungandr/jormungandr/instance.py
+++ b/source/jormungandr/jormungandr/instance.py
@@ -40,6 +40,8 @@ from .exceptions import DeadSocketException
 from navitiacommon import models
 from importlib import import_module
 from jormungandr import cache, app
+from shapely import wkt
+from shapely.geos import ReadingError
 
 type_to_pttype = {
       "stop_area" : request_pb2.PlaceCodeRequest.StopArea,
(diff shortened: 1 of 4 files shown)


source/jormungandr/jormungandr/instance.py
```diff
@@ -196,6 +198,7 @@ def send_and_receive(self, request, timeout=10000, quiet=False):
                 pb = socket.recv()
                 resp = response_pb2.Response()
                 resp.ParseFromString(pb)
+                self.update_property(resp)#we update the timezone and geom of the instances at eeach request
```
antoine-de: typo on eeach

source/jormungandr/jormungandr/instance_manager.py
```diff
@@ -161,18 +160,11 @@ def init_kraken_instances(self):
         for key, instance in self.instances.iteritems():
             try:
                 resp = instance.send_and_receive(req, timeout=1000, quiet=True)
+                #the instance is automatcaly updated on a call
```
antoine-de: typo too

PR #773, opened 2014-12-05
Disruption: Add param disruption

antoine-de:
add a 2 ways of filtering some disruptions:
- 1 on application period (with `datetime` and `duration` api params)
- 1 on publication period (with `publication_datetime`) (this one is more for debug and should not be used much)

to do more tests we need to add the publication_datetime param in ptref too

linked to http://jira.canaltp.fr/browse/NAVITIAII-1369

### DIFF
--- a/source/jormungandr/jormungandr/interfaces/v1/Disruptions.py
+++ b/source/jormungandr/jormungandr/interfaces/v1/Disruptions.py
@@ -35,10 +35,11 @@ from fields import PbField, error, network, line,\
     NonNullList, NonNullNested, pagination, stop_area
 from ResourceUri import ResourceUri
 from jormungandr.interfaces.argument import ArgumentDoc
-from jormungandr.interfaces.common import odt_levels
-from jormungandr.interfaces.parsers import option_value
+from jormungandr.interfaces.parsers import date_time_format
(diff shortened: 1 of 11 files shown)


source/jormungandr/jormungandr/interfaces/v1/Disruptions.py
```diff
@@ -65,11 +73,19 @@ def __init__(self):
...
-                                description="The datetime from which you want\
-                                the disruption")
+        parser_get.add_argument("period", type=int,
+                                description="Period in days from datetime. DEPRECATED use duration parameter")
+        parser_get.add_argument("duration", type=duration, default=timedelta(days=365),
+                                description="Duration from the period we want to display the disruption on")
+        parser_get.add_argument("datetime", type=date_time_format, default=datetime.now(),
+                                description="The datetime from which you want the disruption "
+                                            "(filter on the disruption application periods)")
+        parser_get.add_argument("publication_datetime", type=date_time_format, default=datetime.now(),
```
kinnou02: I don't really like the name, something like `current_datetime` would be better.
And it would be nice to explain that this parameter is mostly use for debug.

Finally, `now` or `utcnow`?

PR #786, opened 2014-12-11
kraken: correction odt zonal

azime:
corrections method of zonal odt for journeypettern


l-vincent-l:
I think you have to close this one

PR #788, opened 2014-12-12
Jormungandr: add the display_duration in the connection api

kinnou02:


source/type/type.proto
```diff
@@ -368,7 +368,9 @@ message CommercialMode {
 message Connection {
     optional StopPoint origin = 1;
     optional StopPoint destination = 2;
-    optional int32 seconds = 3;
+    optional int32 duration = 3;
```
antoine-de, line 371: you should not put it under the same index as seconds no ?

kinnou02, line 371: since it's the same thing, it's good

PR #791, opened 2014-12-15
Adapted: fix retrieval of validity pattern

l-vincent-l:
We were using already existing validity pattern. We don't want that
becauase we will delete some days on it.

### DIFF
--- a/source/ed/adapted.cpp
+++ b/source/ed/adapted.cpp
@@ -39,22 +39,8 @@ namespace pt = boost::posix_time;
 
 namespace ed{
 
-nt::ValidityPattern* get_or_create_validity_pattern(nt::PT_Data& data, nt::ValidityPattern* validity_pattern){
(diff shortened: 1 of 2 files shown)


source/ed/adapted.cpp
```diff
@@ -225,6 +210,9 @@ void duplicate_vj(nt::VehicleJourney* vehicle_journey,
     //ceci permet de ne pas recréer un vj adapté à chaque fois que l'on change de vj de référence
     std::map<nt::VehicleJourney*, nt::VehicleJourney*> prec_vjs;
     //on teste le validitypattern adapté car si le VJ est déjà supprimé le traitement n'est pas nécessaire
+    if (vehicle_journey->uri.find("SAR:2607-17364") != std::string::npos) {
+        std::cout << "pouet "<< std::endl;
```
kinnou02: debug

PR #798, opened 2014-12-22
Add notes on lines and routes for journeys and schedules API

kinnou02:
A field "notes" has been added to the PtDisplayInfo protobuf message, it's
contain notes for the route, the line and the vehicle journey.

In the journey API the notes are referenced in the item "links" of a
section.

In the route schedules API the notes are in `route_schedules.table.headers.links`
and in `route_schedules.links`. Notes on lines and routes are duplicated, one time on the global links
and on each headers.

And finally for Stop Schedules the notes are in the links of each items: `stop_chedules[].links`

### DIFF
--- a/source/jormungandr/jormungandr/interfaces/v1/Journeys.py
+++ b/source/jormungandr/jormungandr/interfaces/v1/Journeys.py
@@ -80,6 +80,10 @@ class SectionLinks(fields.Raw):
         if links:
             for type_, value in links:
                 response.append({"type": type_.name, "id": value})
+
+        if obj.HasField('pt_display_informations'):
+            for value in obj.pt_display_informations.notes:
+                response.append({"type": 'notes', "id": value.uri, 'value': value.note})
         return response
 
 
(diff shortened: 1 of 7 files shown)


source/jormungandr/jormungandr/interfaces/v1/Journeys.py
```diff
@@ -80,6 +80,12 @@ def output(self, key, obj):
         if links:
             for type_, value in links:
                 response.append({"type": type_.name, "id": value})
+        try:
```
antoine-de: the try + check existence might be a bit overkill no ?

source/type/type.h
```diff
@@ -575,7 +575,7 @@ struct Line : public Header, Nameable, HasMessages, Codes{
         ar & idx & name & uri & code & forward_name & backward_name
                 & additional_data & color & sort & commercial_mode
                 & company_list & network & route_list & physical_mode_list
-                & impacts & calendar_list & codes & shape;
+                & impacts & calendar_list & codes & shape & comment;
```
antoine-de, line 578: you need to update the data version number too no ?

PR #804, opened 2014-12-23
Raptor: Freq vj

antoine-de:
Split vehicle journeys in 2 different subclasses:
- DiscreteVehicleJourney
  
  the 'classic' vj
- FrequencyVehicleJourney
  
  the frequency vj with start, end and headway as additional informations

This has been done:
- to reduce the size of the VJ (we removed 3 useless fields for DiscreteVJ, the most used kind of VJ)
- to reduce the size of the departure/arrival raptor structures (we now only store discreteVJ stoptime, no frequencyVJ stoptime)
- to better handle frequencies (there was some bug in the way frequencies were handled)

The 2 kind of VJ are handled 2 different way in raptor (in best_stop_time to be precise):
- DiscreteVJ are handled as usual with a departure/arrival sorted vector on which we loop through to find the best stop time (seenext_valid_discrete_pick_up /previous_valid_discrete_drop_off)
- FrequencyVJ are handled for the moment in a naive way
  
  We loop through all the jpp's vj to find the next one.
  We can use more clever data structure if needed

### DIFF
--- a/source/routing/best_stoptime.cpp
+++ b/source/routing/best_stoptime.cpp
@@ -46,7 +46,7 @@ best_stop_time(const type::JourneyPatternPoint* jpp,
 /** Which is the first valid stop_time in this range ?
  *  Returns invalid_idx is none is
  */
-static std::pair<const type::StopTime*, DateTime>
+std::pair<const type::StopTime*, DateTime>
 next_valid_discrete_pick_up(const type::JourneyPatternPoint* jpp, const DateTime dt,
         const type::Data &data, bool reconstructing_path,
         const type::VehicleProperties &required_vehicle_properties,
@@ -242,7 +242,7 @@ previous_valid_discrete_drop_off(const type::JourneyPatternPoint* jpp, const Dat
  *
  *      Note: if nothing found, we do the same lookup the day after
  */
-static std::pair<const type::StopTime*, uint32_t>
+std::pair<const type::StopTime*, uint32_t>
 earliest_stop_time(const type::JourneyPatternPoint* jpp,
                    const DateTime dt, const type::Data &data,
                    bool disruption_active,
--- a/source/routing/best_stoptime.cpp
+++ b/source/routing/best_stoptime.cpp
@@ -142,7 +142,7 @@ previous_valid_frequency_drop_off(const type::JourneyPatternPoint* jpp, const Da
                              bool reconstructing_path,
                              const type::VehicleProperties &vehicle_properties,
                              bool disruption_active) {
-    std::pair<const type::StopTime*, DateTime> best = {nullptr, DateTimeUtils::inf};
+    std::pair<const type::StopTime*, DateTime> best = {nullptr, DateTimeUtils::not_valid};
     for (const auto& freq_vj: jpp->journey_pattern->frequency_vehicle_journey_list) {
         //we get stop time corresponding to the jpp
         const auto& st = freq_vj->stop_time_list[jpp->order];
--- a/source/routing/best_stoptime.cpp
+++ b/source/routing/best_stoptime.cpp
@@ -70,6 +70,7 @@ next_valid_discrete_pick_up(const type::JourneyPatternPoint* jpp, const DateTime
     auto date = DateTimeUtils::date(dt);
     for(; idx < end; ++idx) {
         const type::StopTime* st = data.dataRaptor->st_forward[idx];
+        BOOST_ASSERT(st->journey_pattern_point == jpp);
         if (st->valid_end(reconstructing_path) &&
             st->is_valid_day(date, false, disruption_active)
             && st->vehicle_journey->accessible(required_vehicle_properties) ){
--- a/source/routing/best_stoptime.cpp
+++ b/source/routing/best_stoptime.cpp
@@ -46,55 +46,116 @@ best_stop_time(const type::JourneyPatternPoint* jpp,
 /** Which is the first valid stop_time in this range ?
  *  Returns invalid_idx is none is
  */
-static const type::StopTime*
-next_valid_pick_up(type::idx_t idx,
-                   const type::idx_t end,
-                   const DateTime dt,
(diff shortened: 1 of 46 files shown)


source/routing/best_stoptime.cpp
```diff
@@ -46,79 +46,202 @@ best_stop_time(const type::JourneyPatternPoint* jpp,
...
+        }
+    }
+
+    //if non was found, we try again the next day
+    idx = begin - data.dataRaptor->departure_times.begin();
+    date++;
     for(; idx < end; ++idx) {
         const type::StopTime* st = data.dataRaptor->st_forward[idx];
-        if (st->valid_end(reconstructing_path) && st->valid_hour(hour, true) &&
+        if (st->valid_end(reconstructing_path) &&
```
kinnou02: I prefer having the `&&` at the start of the line

source/type/type.cpp
```diff
@@ -379,6 +376,19 @@ void Calendar::build_validity_pattern(boost::gregorian::date_period production_p
...
 
+bool VehicleJourney::operator<(const VehicleJourney& other) const {
+    if (this->journey_pattern->uri != other.journey_pattern->uri)
+        return this->journey_pattern->uri < other.journey_pattern->uri;
+    return this->uri < other.uri;
+
+    if (this == &other) return false;
+    if(this->journey_pattern == other.journey_pattern){
+        // On compare les pointeurs pour avoir un ordre total (fonctionnellement osef du tri, mais techniquement c'est important)
+        return this->stop_time_list.front() < other.stop_time_list.front();
```
TeXitoi, line 386: Je crois que le commentaire n'est plus d'actualité depuis que les stop_time sont des objects dans des vecteurs

source/type/type.cpp
```diff
@@ -552,6 +577,26 @@ std::vector<idx_t> VehicleJourney::get(Type_e type, const PT_Data &) const {
...
 }
 
+VehicleJourney::~VehicleJourney() {}
+FrequencyVehicleJourney::~FrequencyVehicleJourney() {}
+DiscreteVehicleJourney::~DiscreteVehicleJourney() {}
+
+JourneyPattern::~JourneyPattern() {}
+
+//we need to define the copy constructor because we don't want to copy the vjs
+//(idx can be spiped too)
```
TeXitoi: I don't know the work spiped ;)

source/routing/best_stoptime.cpp
```diff
@@ -46,79 +46,194 @@ best_stop_time(const type::JourneyPatternPoint* jpp,
...
-               const DateTime dt,
-               const type::Data& data,
-               bool reconstructing_path,
-               const type::VehicleProperties& required_vehicle_properties,
+std::pair<const type::StopTime*, DateTime>
+previous_valid_frequency_drop_off(const type::JourneyPatternPoint* jpp, const DateTime dt,
+                             bool reconstructing_path,
+                             const type::VehicleProperties &vehicle_properties,
+                             bool disruption_active) {
+    std::pair<const type::StopTime*, DateTime> best = {nullptr, DateTimeUtils::inf};
```
TeXitoi: A previous  with a default to inf, that's strange, non?

source/routing/best_stoptime.cpp
```diff
@@ -126,53 +241,74 @@ earliest_stop_time(const type::JourneyPatternPoint* jpp,
...
-            DateTimeUtils::update(first_st.second, DateTimeUtils::hour(tmp_dt));
+    const auto first_discrete_st_pair =
+            next_valid_discrete_pick_up(jpp, dt, data, reconstructing_path, vehicle_properties, disruption_active);
+
+    //TODO use first_discrete as a LB ?
+    const auto first_frequency_st_pair =
+            next_valid_frequency_pick_up(jpp, dt, reconstructing_path, vehicle_properties, disruption_active);
+
+    // If no trip was found, we look for one the day after
+    if (first_discrete_st_pair.first || first_frequency_st_pair.first) {
```
TeXitoi: That's quite complicated, if next_valid returns the good thing, just a "max" is enough.

source/routing/best_stoptime.cpp
```diff
@@ -126,53 +241,74 @@ earliest_stop_time(const type::JourneyPatternPoint* jpp,
...
                    const std::string calendar_id,
                    const type::VehicleProperties& vehicle_properties) {
 
     std::set<const type::MetaVehicleJourney*> meta_vjs;
-    for (auto vj: jpp->journey_pattern->vehicle_journey_list) {
-        if (! vj->meta_vj) {
-            throw navitia::exception("vj " + vj->uri + " has been ill constructed, it has no meta vj");
+    jpp->journey_pattern->for_each_vehicle_journey([&](const nt::VehicleJourney& vj ) {
+        if (! vj.meta_vj) {
+            throw navitia::exception("vj " + vj.uri + " has been ill constructed, it has no meta vj");
```
TeXitoi: an assert is enough, no?

source/routing/best_stoptime.cpp
```diff
@@ -221,50 +358,143 @@ get_all_stop_times(const type::JourneyPatternPoint* jpp,
...
+    if (! freq_vj.is_valid(date - 1, adapted)) {
+        //the vj was not valid, next departure is lower
+        return DateTimeUtils::set(date, lower_bound);
+    }
+
+    double diff = hour - lower_bound + DateTimeUtils::SECONDS_PER_DAY;
+    const int32_t x = std::ceil(diff / double(freq_vj.headway_secs));
+
+    std::cout << " next : " << str(lower_bound) << "," << str(upper_bound) << " diff= " << str(diff)
+              << " x = " << x << " res == " << DateTimeUtils::set(date, (lower_bound + x * freq_vj.headway_secs) % DateTimeUtils::SECONDS_PER_DAY) << std::endl;
```
TeXitoi: trace to remove

PR #807, opened 2015-01-07
fix disruption integration

kinnou02:
- Some fields nullable were defined as required
- we use a left join for the tag since it's possible for a disruption to have no tags
- when a disruption is received by rabbitmq we check if we already have it for doing an update of this one (more a delete and create but who care...)

### DIFF
--- a/source/kraken/fill_disruption_from_database.h
+++ b/source/kraken/fill_disruption_from_database.h
@@ -80,7 +80,7 @@ namespace navitia {
                 impact = nullptr;
             }
 
-            if (disruption && (!tag ||
+            if (disruption && !const_it["tag_id"].is_null() && (!tag ||
                     tag->id() != const_it["tag_id"].template as<std::string>())) {
                 fill_tag(const_it);
             }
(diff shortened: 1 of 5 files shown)


source/kraken/fill_disruption_from_database.h
```diff
@@ -173,9 +173,9 @@ namespace navitia {
         template<typename T>
         void fill_pt_object(T const_it) {
             auto ptobject = impact->add_informed_entities();
-            FILL_REQUIRED(ptobject, updated_at, uint64_t)
-            FILL_NULLABLE(ptobject, created_at, uint64_t)
-            FILL_REQUIRED(ptobject, uri, std::string)
+            FILL_NULLABLE(ptobject, updated_at, uint64_t)
+            FILL_REQUIRED(ptobject, created_at, uint64_t)
+            FILL_NULLABLE(ptobject, uri, std::string)
```
TeXitoi, line 178: Why NULLABLE? PtObject::uri is required in chaos.proto

kinnou02, line 178: but it is nullable in the database...
I have opened an issue in chaos: http://jira.canaltp.fr/browse/IS-1031

PR #825, opened 2015-01-15
Raptor: Fix overflow

antoine-de:
fix overflow in raptor when asking previous arrival the first day


source/type/datetime.cpp
```diff
@@ -94,8 +94,9 @@ std::vector<pt::time_period>
 expand_calendar(pt::ptime start, pt::ptime end,
```
TeXitoi, line 94: move that in nav2rt

antoine-de, line 94: it's difficult because it has to be shared with disruption_tests

PR #830, opened 2015-01-19
ed2nav: better projection for finding the admin of a coord

kinnou02:
The previous method was excluding way without name.

### DIFF
--- a/source/georef/georef.cpp
+++ b/source/georef/georef.cpp
@@ -570,14 +570,42 @@ void GeoRef::project_stop_points(const std::vector<type::StopPoint*> &stop_point
    }
 }
 
-const std::vector<Admin*> &GeoRef::find_admins(const type::GeographicalCoord& coord) const {
-    try {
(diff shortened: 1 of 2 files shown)


source/georef/georef.cpp
```diff
@@ -570,10 +570,44 @@ void GeoRef::project_stop_points(const std::vector<type::StopPoint*> &stop_point
...
+    for (const auto& pair_coord: pl.find_within(coord)) {
+        BOOST_FOREACH (edge_t e, boost::out_edges(pair_coord.first, graph)) {
+            const Way* w = ways[graph[e].way_idx];
+            if (w->admin_list.empty()) { continue; }
+            if (way_dist.count(w) == 0) {
+                way_dist[w] = coord.distance_to(w->projected_centroid(graph));
+            }
+        }
+    }
+    if (way_dist.empty()) { throw proximitylist::NotFound(); }
```
antoine-de: I don't like this throw stuff

PR #837, opened 2015-01-26
routing: refactor best_stoptime in next_stop_time

TeXitoi:
This refactor abstract all the next_stop_time mechanic
and allow futur optimizations as LRU for the stop time
validities


TeXitoi:
Note: as of 2015-01-26T1852, it merges without conflict on the command line.

TeXitoi:
And tests are passing

source/routing/next_stop_time.h
```diff
@@ -0,0 +1,167 @@
...
+    stop_time_range_before(const JppIdx jpp_idx, const DateTime dt) const {
+        const auto& tst = backward[jpp_idx];
+        const auto it = boost::lower_bound(tst.times, DateTimeUtils::hour(dt),
+                                           std::greater<DateTime>());
+        const auto idx = it - tst.times.begin();
+        return boost::make_iterator_range(tst.stop_times.begin() + idx, tst.stop_times.end());
+    }
+
+private:
+    struct TimesStopTimes {
```
antoine-de: lash1kom

antoine-de: wouldn't it be better for the comparator  (from init) to be part on this structure ?

source/routing/idx_map.h
```diff
@@ -33,16 +33,18 @@ www.navitia.io
 #include <vector>
 #include <iterator>
 #include <boost/range/iterator_range_core.hpp>
-#include "type/type.h" // for type::idx_t
 
 namespace navitia { namespace routing {
 
+typedef uint32_t idx_t;
+const idx_t invalid_idx = std::numeric_limits<idx_t>::max();
```
l-vincent-l: Isn't it a bit dangerous?
the typedef of idx_t should be global no?

TeXitoi: The problem is that idx_map.h depends on type.h if we don't do that. I can move idx_map.h in utils, and define the idx_t of type.h as a typedef of the one of idx_map.h, if it's ok for you.

l-vincent-l:
OK for me

PR #843, opened 2015-01-28
Time representation in navitia.io documentation

vinzentt:
Using YYYYMMDDTHHMMSS  in doc may be confusing (both months and minutes = M), ISO 8601 format is YYYYMMDDThhmmss


antoine-de:
Thanks for the update!

PR #854, opened 2015-02-02
stat: interpreted parameters

kinnou02:
In stat we keep the parameter set by the client and log in another table the parameters used by jormungandr with default values and eventually modification done by the internal logic of each api.

Currently ptref filter are stored as is, the parsing of the query is not done, so for a departure query on one stop_area for a specific line, we will store something like this:

```
line.uri=line:NAN:NA-0 and stop_point.uri=stop_point:NAN:SP:COMD3
```

### DIFF
--- a/source/jormungandr/jormungandr/interfaces/v1/Coord.py
+++ b/source/jormungandr/jormungandr/interfaces/v1/Coord.py
@@ -64,6 +64,7 @@ class Coord(ResourceUri):
                 "start_page": 0,
                 "filter": ""
             }
+            self._register_interpreted_parameters(args)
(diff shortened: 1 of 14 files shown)


source/jormungandr/jormungandr/interfaces/v1/Coord.py
```diff
@@ -64,6 +64,7 @@ def get(self, region=None, lon=None, lat=None, id=None, *args, **kwargs):
                 "start_page": 0,
                 "filter": ""
             }
+            #self._register_interpreted_parameters(args)
```
antoine-de: what is this ?

PR #866, opened 2015-02-05
Jormungandr: stat the region for each call, even if the region isn't specified

kinnou02:


source/jormungandr/jormungandr/interfaces/v1/StatedResource.py
```diff
@@ -39,7 +40,17 @@ def __init__(self, *args, **kwargs):
...
             self.method_decorators.append(manage_stat_caller(stat_manager))
 
     def _register_interpreted_parameters(self, args):
         stat_manager.register_interpreted_parameters(args)
+
+    def _stat_regions(self, f):
+        @wraps(f)
+        def wrapper(*args, **kwargs):
+            result = f(*args, **kwargs)
+            if hasattr(self, 'region') and self.region:
```
l-vincent-l: You could do : 
region = getattr(self, 'region', None)
if region:

Also, I think you want to do stat_manager.register_regions([region])

l-vincent-l: or add a function register_region

source/jormungandr/jormungandr/stat_manager.py
```diff
@@ -183,10 +186,14 @@ def fill_coverages(self, stat_request):
         if 'region' in request.view_args:
             stat_coverage = stat_request.coverages.add()
             stat_coverage.region_id = request.view_args['region']
-        elif hasattr(g, 'regions_called'):
-            for region_id in g.regions_called:
+        elif hasattr(g, 'stat_regions'):
+            if isinstance(g.stat_regions, list):
+                for region_id in g.stat_regions:
```
l-vincent-l: I think it would be better if regions is always a list

PR #867, opened 2015-02-05
Ed: check range of validity_patterns

l-vincent-l:


source/ed/connectors/fusio_parser.cpp
```diff
@@ -325,8 +325,8 @@ std::vector<ed::types::VehicleJourney*> TripsFusioHandler::get_split_vj(Data& da
     }
     ed::types::Route* route = it->second;
 
-    auto vp_it = gtfs_data.tz.vp_by_name.lower_bound(row[service_c]);
-    if(vp_it == gtfs_data.tz.vp_by_name.end()) {
+    auto vp_range = gtfs_data.tz.vp_by_name.equal_range(row[service_c]);
+    if(vp_range.first == gtfs_data.tz.vp_by_name.end() || vp_range.first->first != row[service_c]) {
```
TeXitoi: no, you want to know if the range is empty, ie `vp_range.first == vp_range.second`.

better to add a function like that in gtfs_parser.h:

``` c++
template<typename T> bool
empty(const std::pair<T, T>& r) {
    return r.first == r.second;
}
```

and then use it.

source/ed/connectors/fusio_parser.cpp
```diff
@@ -346,14 +346,12 @@ std::vector<ed::types::VehicleJourney*> TripsFusioHandler::get_split_vj(Data& da
     // get shape if possible
     const std::string &shape_id = has_col(geometry_id_c, row) ? row.at(geometry_id_c) : "";
 
-    const auto vp_end_it = gtfs_data.tz.vp_by_name.upper_bound(row[service_c]);
-
-    auto second_elt = vp_it;
-    bool has_been_split = (++second_elt != vp_end_it); //check if the trip has been split over dst
+    auto second_elt = vp_range.second;
+    bool has_been_split = (++second_elt != vp_range.second); //check if the trip has been split over dst
```
TeXitoi: Same here, abstract that, you want more_than_one_elt:

``` c++
template<typename T> bool more_than_one_elt(const std::pair<T, T>& r) {
    return ! empty(r) && boost::next(r.first) != r.second;
}
```

See http://www.boost.org/doc/libs/1_57_0/libs/utility/utility.htm#functions_next_prior for `boost::next`

l-vincent-l:
Jenkins, test this please

PR #869, opened 2015-02-05
Remove sindri

kinnou02:


antoine-de:
don't you need to remove the tmp_database_file in tyr too ?

antoine-de:
we also need to look into those:

``` cpp

    bool is_adapted = false; //REMOVE (change to enum ?)
    ValidityPattern* adapted_validity_pattern = nullptr; //REMOVE
    std::vector<VehicleJourney*> adapted_vehicle_journey_list; //REMOVE
    VehicleJourney* theoric_vehicle_journey = nullptr; //REMOVE
```

I think the adapted_validity_pattern needs to be kept but for the rest, that should be useless with the meta vj no ? (we need to check that the meta vj is well created)

PR #876, opened 2015-02-06
Ed: don't import empty connection

l-vincent-l:


source/ed/connectors/gtfs_parser.cpp
```diff
@@ -578,7 +578,7 @@ void TransfersGtfsHandler::handle_line(Data& data, const csv_row& row, bool) {
     GtfsData::vector_sp departures, arrivals;
     // We don't want to have these connections, because it could lead to weird things
     // And also we don't want to show that to our users, this must be a data problem
-    if (boost::lexical_cast<int>(row[time_c]) > 24*3600) {
```
antoine-de, line 581: that will go to the same log even if that is not the same thing.

can you split it ?

PR #877, opened 2015-02-10
save parsed ptref filter in stat

kinnou02:
parsing and saving of ptref filter.

The main drawback of the solution is that if the parser is modified in kraken we will have to modify jormungandr too.

### DIFF
--- a/source/jormungandr/jormungandr/stat_manager.py
+++ b/source/jormungandr/jormungandr/stat_manager.py
@@ -226,7 +226,7 @@ class StatManager(object):
         """
         if not value:
             return
-        for elem in value.split('and'):
+        for elem in re.split('\s+and\s+', r, flags=re.IGNORECASE):
             match = re.match('^\s*(\w+).(\w+)\s*([=!><]{1,2}|DWITHIN)\s*(.*)$', elem, re.IGNORECASE)
             if match:
                 filter = stat_parameter.filters.add()
--- a/source/jormungandr/jormungandr/stat_manager.py
+++ b/source/jormungandr/jormungandr/stat_manager.py
@@ -37,6 +37,7 @@ import logging
 from jormungandr import app
 from jormungandr.authentication import get_user, get_token, get_app_name
 from jormungandr import utils
+import re
 
 import time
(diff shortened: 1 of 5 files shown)


source/stat_persistor/migrations/alembic/versions/2c90cc5083e6_add_filter.py
```diff
@@ -0,0 +1,37 @@
...
+# revision identifiers, used by Alembic.
+revision = '2c90cc5083e6'
+down_revision = 'd8cff35d69'
+
+from alembic import op
+import sqlalchemy as sa
+
+
+def upgrade():
+    op.execute('''CREATE SEQUENCE interpred_parameter_seq;''')
```
antoine-de: I think  you should add the namespace

source/jormungandr/jormungandr/stat_manager.py
```diff
@@ -214,6 +215,28 @@ def fill_parameters(self, stat_request):
...
+        if not value:
+            return
+        for elem in value.split('and'):
+            match = re.match('^\s*(\w+).(\w+)\s*([=!><]{1,2}|DWITHIN)\s*(.*)$', elem, re.IGNORECASE)
+            if match:
+                filter = stat_parameter.filters.add()
+                filter.object = match.group(1)
+                filter.attribute = match.group(2)
+                filter.operator = match.group(3)
+                filter.value = match.group(4)
```
l-vincent-l: You cna have multiple fitlers per request

source/jormungandr/jormungandr/stat_manager.py
```diff
@@ -214,6 +215,28 @@ def fill_parameters(self, stat_request):
...
+                        self.fill_filters(item[1], stat_parameter)
+
+
+    def fill_filters(self, value, stat_parameter):
+        """
+        basic parsing of a ptref filter
+        """
+        if not value:
+            return
+        for elem in value.split('and'):
```
l-vincent-l: you can have AND too.
You need to replace AND by and before.

PR #879, opened 2015-02-10
Ed: read and store line opening/closing times

l-vincent-l:


source/ed/types.h
```diff
@@ -195,23 +195,25 @@ struct PhysicalMode : public Header, Nameable{
 
 struct Line : public Header, Nameable {
     const static nt::Type_e type = nt::Type_e::Line;
-    std::string external_code;
-    std::string code;
-    std::string forward_name;
-    std::string backward_name;
+    std::string external_code = "";
```
antoine-de: it is not needed, it is the default behavior

source/ed/types.h
```diff
@@ -195,23 +195,25 @@ struct PhysicalMode : public Header, Nameable{
...
-    Network* network;
-    Company* company;
+    CommercialMode* commercial_mode = nullptr;
+    Network* network = nullptr;
+    Company* company = nullptr;
     nt::MultiLineString shape;
+    std::string opening_time = "", closing_time = "";
 
-    Line(): color(""), commercial_mode(NULL), network(NULL), company(NULL){}
+    Line() {}
```
antoine-de: thus you can remove the constructor no ?

source/ed/types.h
```diff
@@ -195,23 +195,25 @@ struct PhysicalMode : public Header, Nameable{
...
     int sort = std::numeric_limits<int>::max();
 
-    CommercialMode* commercial_mode;
-    Network* network;
-    Company* company;
+    CommercialMode* commercial_mode = nullptr;
+    Network* network = nullptr;
+    Company* company = nullptr;
     nt::MultiLineString shape;
+    std::string opening_time = "", closing_time = "";
```
antoine-de: why do you store that as string ?
It is a time no ?

l-vincent-l: Because this is easier, but yep, I should do it.

source/type/type.proto
```diff
@@ -213,6 +213,8 @@ message Line {
     optional string comment = 15;
     repeated Code codes = 16;
     optional MultiLineString geojson = 17;
+    optional string opening_time = 18;
+    optional string closing_time = 19;
```
TeXitoi: passing as second til midnight?

l-vincent-l:
Finally fixed

PR #885, opened 2015-02-12
ww autocompletion: first draft

TeXitoi:
Add a http://localhost:5000/v1/places?q=8%20mart%20ber%20par you need an elasticsearch server with data inside to test.

### DIFF
--- a/source/jormungandr/jormungandr/interfaces/v1/Places.py
+++ b/source/jormungandr/jormungandr/interfaces/v1/Places.py
@@ -171,19 +171,21 @@ class Places(ResourceUri):
         if len(args['q']) == 0:
             abort(400, message="Search word absent")
 
-        if not any([region, lon, lat]):
-            return self.get_ww(args)
+        # If a region or coords are asked, we do the search according
+        # to the region, else, we do a word wide search
+        if any([region, lon, lat]):
+            return self._search_region(args, region, lon, lat)
         else:
-            return self.get_region(args, region, lon, lat)
+            return self._search_world_wide(args)
 
     @marshal_with(places)
-    def get_region(self, args, region=None, lon=None, lat=None):
+    def _search_region(self, args, region=None, lon=None, lat=None):
         self.region = i_manager.get_region(region, lon, lat)
         response = i_manager.dispatch(args, "places", instance_name=self.region)
         return response, 200
 
     @marshal_with(ww_places)
-    def get_ww(self, args):
+    def _search_world_wide(self, args):
         q = args['q']
         query = {
             "query": {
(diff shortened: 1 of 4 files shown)


source/jormungandr/jormungandr/interfaces/v1/Places.py
```diff
@@ -51,11 +55,78 @@
...
+            "name": source["name"],
+            "administrative_region": marshal_es_admin(source)
+        }
+    return place
+
+class marshal_es:
+    def __call__(self, f):
+        @wraps(f)
+        def wrapper(*args, **kwargs):
+            r = f(*args, **kwargs)
```
antoine-de: I think it's better to destructure the response like:

```
response, status_code = f(*args, **kwargs)
```

source/jormungandr/jormungandr/interfaces/v1/Places.py
```diff
@@ -86,19 +164,107 @@ def __init__(self, *args, **kwargs):
...
+                            ]
+                        }
+                    }
+                }
+            }
+        }
+        try:
+            # TODO: get params?
+            es = Elasticsearch()
+            res = es.search(index="munin", size=args['count'], body=query)
```
l-vincent-l, line 265: I don't understand where is the configuration.
How do we know which ES is called?

TeXitoi, line 265: By default, it'll use localhost:9200. The TODO is about being able to specify hosts in configuration.

source/jormungandr/jormungandr/interfaces/v1/Places.py
```diff
@@ -86,19 +164,107 @@ def __init__(self, *args, **kwargs):
...
-        response = i_manager.dispatch(args, "places",
-                                      instance_name=self.region)
+
+        if not any([region, lon, lat]):
+            return self.get_ww(args)
+        else:
+            return self.get_region(args, region, lon, lat)
+
+    @marshal_with(places)
+    def get_region(self, args, region=None, lon=None, lat=None):
```
antoine-de: I don't think this method is correctly named (and there is another get_region method which get the region :p)

source/jormungandr/jormungandr/interfaces/v1/Places.py
```diff
@@ -86,19 +164,107 @@ def __init__(self, *args, **kwargs):
...
-        g.use_old_disruptions = args['_use_old_disruptions']
         self._register_interpreted_parameters(args)
-
+        g.use_old_disruptions = args['_use_old_disruptions']
         if len(args['q']) == 0:
             abort(400, message="Search word absent")
-        response = i_manager.dispatch(args, "places",
-                                      instance_name=self.region)
+
+        if not any([region, lon, lat]):
```
antoine-de: maybe put a comment here

antoine-de:
you should add authentication test (because it's handled manually in 'places' now)

and maybe one access test on the global place API (and check the 'no elastic search available' error)

PR #888, opened 2015-02-13
Kraken: fix segfault in route_schedule if a route has no vehiclejourney

kinnou02:
jira.canaltp.fr/browse/NAVITIAII-1596


TeXitoi:
Please add a test

kinnou02:
test added

PR #889, opened 2015-02-16
 kraken: recast of disruption api 

kinnou02:
All publishable disruption are returned, the status is dependant of the
application period.

The Api don't take date anymore, only the current date is used

### DIFF
--- a/source/disruption/tests/disruption_test.cpp
+++ b/source/disruption/tests/disruption_test.cpp
@@ -211,6 +211,7 @@ BOOST_FIXTURE_TEST_CASE(network_filter1, Params) {
     BOOST_REQUIRE_EQUAL(line.disruptions_size(), 1);
     auto disruption = line.disruptions(0);
     BOOST_CHECK_EQUAL(disruption.uri(), "mess1");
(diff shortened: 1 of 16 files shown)


source/type/message.cpp
```diff
@@ -62,6 +62,19 @@ bool Impact::is_valid(const boost::posix_time::ptime& publication_date, const bo
         }
     }
     return false;
+
```
TeXitoi: Why an empty line here? same at the end of the next method.

source/disruption/tests/disruption_test.cpp
```diff
@@ -234,19 +211,10 @@ BOOST_FIXTURE_TEST_CASE(network_filter1, Params) {
     BOOST_REQUIRE_EQUAL(line.disruptions_size(), 1);
     auto disruption = line.disruptions(0);
     BOOST_CHECK_EQUAL(disruption.uri(), "mess1");
+    BOOST_CHECK_EQUAL(disruption.status(), pbnavitia::ActiveStatus::future);
```
antoine-de, line 215: could you put a comment to explain why the status is 'future' ?

PR #894, opened 2015-02-18
Kraken: remove duplicate impact on vehicle journey

kinnou02:
For blocking impact we have to update the vehicle_journeys attached at the impacted object, typically a line or a route. For being able to update a impact we have to store each impact impacting the vj. But this impacts must not be displayed for this vj else we will have duplicate items on the response.

So impacts not directly on a vj are stored in a new attributes: `impacted_by`.

### DIFF
--- a/source/kraken/fill_disruption_from_chaos.cpp
+++ b/source/kraken/fill_disruption_from_chaos.cpp
@@ -37,6 +37,7 @@ www.navitia.io
 #include <boost/variant/apply_visitor.hpp>
 #include <boost/range/algorithm/for_each.hpp>
 #include <boost/date_time/posix_time/posix_time.hpp>
+#include <boost/range/algorithm_ext/erase.hpp>
 
 namespace navitia {
 
@@ -341,7 +342,7 @@ struct functor_add_vj {
             ++ order;
         }
         // We need to link the newly created vj with this impact
-        vj->add_impact(impact);
+        vj->impacted_by.push_back(impact);
     }
 
     bool operator()(nt::DiscreteVehicleJourney& vj_ref) const {
@@ -388,7 +389,7 @@ struct add_impacts_visitor : public apply_impacts_visitor {
             }
         }
         vj.adapted_validity_pattern = pt_data.get_or_create_validity_pattern(tmp_vp);
-        vj.add_impact(impact);
+        vj.impacted_by.push_back(impact);
         return true;
     }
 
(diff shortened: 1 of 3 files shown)


source/kraken/fill_disruption_from_chaos.cpp
```diff
@@ -474,10 +474,21 @@ struct delete_impacts_visitor : public apply_impacts_visitor {
...
         ++ nb_vj_reassigned;
-        vj.remove_impact(impact);
+        //it don't seem to be possible to capture an attribute of the object
+        const auto& impact = this->impact;
+        auto it = std::find_if(vj.impacted_by.begin(), vj.impacted_by.end(),
+                [&impact](const boost::weak_ptr<nt::new_disruption::Impact>& i) {
+            return i.lock() == impact;
+        });
+        if (it != vj.impacted_by.end()) {
+            vj.impacted_by.erase(it);
```
l-vincent-l: maybe you can use http://en.cppreference.com/w/cpp/algorithm/remove

source/kraken/fill_disruption_from_chaos.cpp
```diff
@@ -474,10 +474,21 @@ struct delete_impacts_visitor : public apply_impacts_visitor {
...
+        }
         for (auto i: vj.get_impacts()) {
             apply_impact(i.lock(), pt_data, meta);
         }
```
l-vincent-l: You don't need it anymore, it will be done by impacts of impacted_by

source/kraken/fill_disruption_from_chaos.cpp
```diff
@@ -474,8 +474,13 @@ struct delete_impacts_visitor : public apply_impacts_visitor {
...
         vj.adapted_validity_pattern = vj.validity_pattern;
         ++ nb_vj_reassigned;
-        vj.remove_impact(impact);
-        for (auto i: vj.get_impacts()) {
+        //it don't seem to be possible to capture an attribute of the object
+        const auto& impact = this->impact;
+        vj.impacted_by.erase(std::remove_if(vj.impacted_by.begin(), vj.impacted_by.end(),
+                [&impact](const boost::weak_ptr<nt::new_disruption::Impact>& i) {
+            return i.lock() == impact;
+        }), vj.impacted_by.end());
```
TeXitoi: you can use `boost::remove_erase_if` http://www.boost.org/doc/libs/1_57_0/libs/range/doc/html/range/reference/algorithms/new/erase.html

source/kraken/fill_disruption_from_chaos.cpp
```diff
@@ -474,8 +474,13 @@ struct delete_impacts_visitor : public apply_impacts_visitor {
     bool func_on_vj(nt::VehicleJourney& vj) {
         vj.adapted_validity_pattern = vj.validity_pattern;
         ++ nb_vj_reassigned;
-        vj.remove_impact(impact);
-        for (auto i: vj.get_impacts()) {
+        //it don't seem to be possible to capture an attribute of the object
```
TeXitoi: True, you cannot do `[&member]() {}`. Please remove this useless bad english comment ;)

TeXitoi:
No tests?

PR #895, opened 2015-02-18
jormungandr: fix /v1/coord/lon;lat when region shapes overlap

TeXitoi:
fix http://jira.canaltp.fr/browse/NAVITIAII-1599

### DIFF
--- a/source/jormungandr/jormungandr/interfaces/v1/Coord.py
+++ b/source/jormungandr/jormungandr/interfaces/v1/Coord.py
@@ -47,32 +47,34 @@ class Coord(ResourceUri):
             lon, lat = splitted
 
         if region is None:
-            self.region = i_manager.get_region("", lon, lat)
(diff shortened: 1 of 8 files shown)


source/jormungandr/jormungandr/interfaces/v1/Coord.py
```diff
@@ -47,32 +47,34 @@ def get(self, region=None, lon=None, lat=None, id=None, *args, **kwargs):
...
-                "start_page": 0,
-                "filter": ""
-            }
-            self._register_interpreted_parameters(args)
-            pb_result = i_manager.dispatch(args, "places_nearby",
-                                           instance_name=self.region)
+            regions = [region]
+
+        if lon is None or lat is None:
+            return abort(404, message="No address for these coords")
```
antoine-de: I think we should have a custom message for the lack of parameters

PR #903, opened 2015-02-23
Jormungandr: add log on error and add retry

kinnou02:
retry to establish the rabbitmq connection each time


source/jormungandr/jormungandr/stat_manager.py
```diff
@@ -129,8 +129,7 @@ def _init_rabbitmq(self):
             self.producer = self.connection.Producer(exchange=exchange)
             self.save_stat = True
```
antoine-de: what is the use of this save_stat ?

I think save_stat should be a parameter only set by the user

PR #904, opened 2015-02-23
Raptor solution reader: first draft of the reader

TeXitoi:
This pull request is not on dev, but on a feature branch that we will use until its mergeable.


source/routing/raptor.cpp
```diff
@@ -602,7 +607,9 @@ void RAPTOR::raptor_loop(Visitor visitor, const type::AccessibiliteParams & acce
                             // We want to update the labels, if it's better than the one computed before
                             // Or if it's an destination point if it's equal and not unitialized before
                             const bool best_add_result = this->b_dest.add_best(visitor, jpp.sp_idx, workingDt, this->count);
-                            if (visitor.comp(workingDt, bound) || best_add_result) {
+                            if ((!visitor.comp(bound, workingDt)
+                                 && visitor.comp(workingDt, working_labels.mut_dt_pt(jpp.sp_idx)))
+                                || best_add_result) {
```
antoine-de, line 612: shouldn't we swap the tests ?

source/routing/raptor.cpp
```diff
@@ -81,9 +81,9 @@ bool RAPTOR::apply_vj_extension(const Visitor& v, const bool global_pruning,
             const auto sp_idx = SpIdx(*sp);
             const DateTime bound = (v.comp(best_labels_pts[sp_idx], b_dest.best_now) || !global_pruning) ?
                                         best_labels_pts[sp_idx] : b_dest.best_now;
-            if(!v.comp(workingDt, bound)) {
-                continue;
-            }
+            if (v.comp(bound, workingDt)) { continue; }
+            if (!v.comp(workingDt, working_labels.dt_pt(sp_idx))) { continue; }
```
l-vincent-l, line 85: Can you add a comment on that?

source/routing/raptor.cpp
```diff
@@ -286,6 +288,9 @@ RAPTOR::compute_all(const vec_stop_point_duration& departures_,
     }
     BOOST_ASSERT( departures.size() > 0 );    //Assert that reversal search was symetric
 
+    if (nb_sol_reader < result.size()) {
+        std::cout << "***** error reader fail" << nb_sol_reader << " " << result.size() << std::endl;
```
l-vincent-l, line 292: It should be a log isn't it?

source/routing/raptor.cpp
TeXitoi, line 612: We can, but the goal is to remove b_dest

source/routing/raptor.cpp
TeXitoi, line 85: I don't really know how to do that cleanly as it is done in several different location, here and in apply_vj_extention, and the equivalent is done in foot_path. Adding comments at all these locations seems dangerous, but I do not have any better idea. Do you have?

source/routing/raptor.cpp
TeXitoi, line 292: That's pure debugging for the moment as the reader just do nothing more than printing things.

PR #906, opened 2015-02-24
Jormungandr: save forbidden uris for every calls

l-vincent-l:


source/jormungandr/jormungandr/scenarios/default.py
```diff
@@ -200,7 +200,7 @@ def call_kraken(self, req, instance, tag=None):
         self.__fill_uris(resp)
         return resp
 
-    def change_request(self, pb_req, resp):
+    def change_request(self, pb_req, resp, forbidden_uris = []):
```
kinnou02: no need of space before or after a equals for a default parameter

PR #908, opened 2015-02-24
Raptor solution reader: filter on objectives

TeXitoi:


source/routing/raptor_solution_reader.cpp
```diff
@@ -70,6 +75,120 @@ struct Journey {
...
+        }
+    }
+    uint8_t count_vj_extentions() const {
+        uint8_t nb_ext = 0;
+        for (const auto& s: sections) {
+            for (auto* vj = s.get_in_st->vehicle_journey; vj != nullptr; vj = vj->next_vj) {
+                const auto search = boost::find_if(vj->stop_time_list, [&](const type::StopTime& st) {
+                        return &st == s.get_out_st;
+                    });
+                if (search != vj->stop_time_list.end()) { break; }
```
antoine-de, line 126: shouldn't you check (at least in debug) that the break is used ?
vj extension has been quite troublesome up to now

source/routing/raptor_solution_reader.cpp
```diff
@@ -85,43 +204,52 @@ struct Journey {
...
+    DateTime end_dt = 0;// the end of the journey, the main objective (depend of the clockwise)
+    DateTime begin_dt = 0;// the begin of the journey (depend of the clockwise)
+    uint8_t nb_vj_extentions = 0;
+};
+
+struct Dominates {
+    bool request_clockwise;
+    Dominates(bool rc): request_clockwise(rc) {}
+    bool operator()(const Journey& lhs, const Journey& rhs) const {
+        return lhs.cmp_dt(rhs, request_clockwise)
```
antoine-de: I think a comment would be nice here

TeXitoi: OK, I'll comment the struct to explain what it does.

l-vincent-l:
jenkins, test this please

PR #914, opened 2015-02-27
Timetables: take in account max_date_times in route_schedules

l-vincent-l:


kinnou02:
A few tests would be great don't you think?

l-vincent-l:
I added a test

PR #915, opened 2015-03-02
Kraken: handle shapes of odt sections

l-vincent-l:


source/routing/raptor_api.cpp
```diff
@@ -72,6 +72,12 @@ static void fill_shape(pbnavitia::Section* pb_section,
                        const std::vector<const type::StopTime*>& stop_times)
 {
     if (stop_times.empty()) return;
+    if (stop_times.front()->vehicle_journey->is_odt()) {
```
kinnou02: We want to do that only for zonal odt no?

PR #917, opened 2015-03-02
Jormungandr: ask for more journeys if you only have street_network journeys

l-vincent-l:


source/jormungandr/tests/routing_tests.py
```diff
@@ -362,4 +364,39 @@ def test_one_dead_region(self):
...
+        from jormungandr import i_manager
+        self.instance_map = {
+            'main_routing_without_pt_test': MockKraken(i_manager.instances['main_routing_without_pt_test'], True),
+            'main_routing_test': MockKraken(i_manager.instances['main_routing_test'], True),
+        }
+        self.real_method = models.Instance.get_by_name
+        self.real_comparator = instance_manager.instances_comparator
+
+        models.Instance.get_by_name = self.mock_get_by_name
+        instance_manager.instances_comparator = lambda a, b: a.name == "main_routing_without_pt_test"
```
kinnou02: Maybe you can add a test for this, we want the test to fail if for some reason we don't return "main_routing_without_pt_test" first

source/jormungandr/tests/routing_tests.py
```diff
@@ -362,4 +364,39 @@ def test_one_dead_region(self):
         eq_(len(response['debug']['regions_called']), 1)
         eq_(response['debug']['regions_called'][0], "main_routing_test")
 
+@dataset(["main_routing_without_pt_test", "main_routing_test"])
+class TestWithoutPt(AbstractTestFixture):
+    """
+    Test if we still responds when one kraken is dead
```
antoine-de, line 370: the comment is not up to date

source/jormungandr/tests/routing_tests.py
```diff
@@ -362,4 +364,39 @@ def test_one_dead_region(self):
...
+
+    def mock_get_by_name(self, name):
+        return self.instance_map[name]
+
+    def teardown(self):
+        models.Instance.get_by_name = self.real_method
+        instance_manager.instances_comparator = self.real_comparator
+
+    def test_one_region_wihout_pt(self):
+        response = self.query("v1/"+journey_basic_query+"&debug=true",
```
antoine-de, line 393: we should also test without debug no ? (since the behavior is a bit different)

PR #919, opened 2015-03-02
raptor solution reader: 3rd pass

TeXitoi:


source/routing/raptor_solution_reader.cpp
```diff
@@ -74,7 +74,13 @@ struct Journey {
                 sections.emplace_back(elt->end_st, elt->end_dt, elt->begin_st, elt->begin_dt);
             }
         }
-        if (reader.v.clockwise()) { boost::reverse(sections); }
+        if (reader.v.clockwise()) {
+            boost::reverse(sections);
+        } else {
+            // the request is clockwise, thus we need to left align
```
antoine-de: I don't think the comment 'the request is clockwise' is very clear since it is in the 'else' of a clockwise test :smile: 

It might be better to remind here that the reader's visitor clockwise is the opposite of the request's clockwise

source/routing/raptor_solution_reader.cpp
```diff
@@ -115,6 +121,60 @@ struct Journey {
...
+                    // check if the get_out is valid?
+                    return {&st, cur_dt};
+                }
+            }
+        }
+        assert(false);
+        return {nullptr, 0};
+    }
+    template<typename Visitor>
+    void align_left(const RaptorSolutionReader<Visitor>& reader) {
```
l-vincent-l, line 156: A comment would be great.

source/routing/raptor_solution_reader.cpp
```diff
@@ -115,6 +121,60 @@ struct Journey {
...
+        if (sections.size() < 3) { return; }
```
l-vincent-l: Here too.

source/routing/raptor_solution_reader.cpp
```diff
@@ -115,6 +121,60 @@ struct Journey {
...
+        const auto* prev_s = &sections.at(0);
+        for (auto& cur_s: boost::make_iterator_range(sections.begin() + 1, sections.end() - 1)) {
+            const auto* cur_jpp = cur_s.get_in_st->journey_pattern_point;
+            const auto* conn = reader.raptor.data.pt_data->get_stop_point_connection(
+                *prev_s->get_out_st->journey_pattern_point->stop_point,
+                *cur_jpp->stop_point);
```
l-vincent-l, line 167: Do we have something to check this isn't possible to have 2 connections linking the same two stop_points?

PR #922, opened 2015-03-03
Ed: do not read pois from OSM if for one instance we red it from poi2ed

l-vincent-l:


source/ed/ed_persistor.h
```diff
@@ -42,9 +43,31 @@ namespace ed{
...
-
-    EdPersistor(const std::string& connection_string) : lotus(connection_string),
-                        logger(log4cplus::Logger::getInstance("log")){}
+    bool parse_pois = true, is_osm_reader;
+
+
+    EdPersistor(const std::string& connection_string,
+            const bool is_osm_reader = true) : lotus(connection_string),
+                        logger(log4cplus::Logger::getInstance("log")),
+                        is_osm_reader(is_osm_reader) {
```
kinnou02: move the definition in the cpp file

PR #930, opened 2015-03-04
An API key could be provided by GET parameter instead of  'Authorization' HTTP header.

xavierraffin:


antoine-de:
Hi @xavierraffin , 
why do you need to be able to give a token with a 'key' param ?

If you want to give a token directly in the url you can use:
`http://{token}@{url}`

xavierraffin:
Because I would like to use Kraken and Jormungandr to serve an API with a ?key=... parameter.
This was our choice when implementing our opendata API : 
https://data.toulouse-metropole.fr/les-donnees/-/opendata/card/14505-api-temps-reel-tisseo

Some request are of the following form : 
http://api.tisseo.fr/v1/stops_schedules.json?key=XXXXXXXXXXXXXXXXXXX&...

We need that Jormungandr allow that.
We can't change because we have already a lot of existing OpenData users.

We will very apreciate if it could be possible to not patch Jormungandr to do this.

xavierraffin:
More explaination about our usecase.

We currently implement a Jormungandr module that provide our OpenData API with a kraken backend.

This is work in progress but very encouraging.

If you are agree with that we will contribute a lot more features in Kraken in the future (3rd & 4rth 2015 quarters).

xavierraffin:
When do you think merge this PR ?

Are you waiting for something from me ?

antoine-de:
yes sorry we were discussing other ways to do this, but we were not able to find better solutions, so we can merge it is if even if we don't find that solution perfect.

PR #936, opened 2015-03-05
Disruptions: load only disruptions active in the data period

l-vincent-l:


source/kraken/fill_disruption_from_database.cpp
```diff
@@ -132,13 +132,13 @@ void fill_disruption_from_database(const std::string& connection_string,
                "     LEFT JOIN pt_object AS ls_end ON line_section.end_object_id = ls_end.id"
                "     LEFT JOIN message AS m ON m.impact_id = i.id"
                "     LEFT JOIN channel AS ch ON m.channel_id = ch.id"
-               "     WHERE co.contributor_code = ANY('{%s}')" // it's like a "IN" but won't crash if empty
+               "     WHERE d.start_publication_date >= '%s' AND d.end_publication_date <= '%s'"
```
kinnou02: The end publication date of a disruption can be null, if it's the case we consider it valid

PR #950, opened 2015-03-11
Jormungandr: use simplejson, it's supposed to be faster

l-vincent-l:


TeXitoi:
We wait our benchmark platform to merge that.

antoine-de:
seems to improve by 1-2% the overall bench

l-vincent-l:
http://permalink.gmane.org/gmane.comp.db.sqlite.general/90549

The 50% faster number above is not about better query plans.  This is 50%
faster at the low-level grunt work of moving bits on and off disk and
search b-trees.  We have achieved this by incorporating hundreds of
micro-optimizations.  Each micro-optimization might improve the performance
by as little as 0.05%.

antoine-de:
:)

I need to test ultra json too, to see if it changes something

PR #952, opened 2015-03-11
jormun: remove walking solution for alternatives journeys in destineo

kinnou02:


kinnou02:
jenkins, test this please

PR #954, opened 2015-03-12
Pt object

azime:
Commercial modes added in StopArea Object

### DIFF
--- a/source/type/type.cpp
+++ b/source/type/type.cpp
@@ -450,10 +450,11 @@ std::vector<idx_t> StopArea::get(Type_e type, const PT_Data & data) const {
     std::vector<idx_t> result;
     switch(type) {
     case Type_e::StopPoint: return indexes(this->stop_point_list); break;
-    case Type_e::CommercialMode:{
+    case Type_e::CommercialMode:
+    case Type_e::PhysicalMode:{
          std::set<idx_t> tmp_result;
         for(const auto& stp: stop_point_list){
-            std::vector<idx_t> tmp = stp->get(Type_e::CommercialMode, data);
+            std::vector<idx_t> tmp = stp->get(type, data);
             for(const idx_t idx: tmp){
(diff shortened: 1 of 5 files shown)


TeXitoi:
jenkins, test this please.

source/type/type.cpp
```diff
@@ -446,10 +446,26 @@ std::vector<idx_t> Calendar::get(Type_e type, const PT_Data & data) const{
     }
     return result;
 }
-std::vector<idx_t> StopArea::get(Type_e type, const PT_Data &) const {
+std::vector<idx_t> StopArea::get(Type_e type, const PT_Data & data) const {
     std::vector<idx_t> result;
     switch(type) {
-    case Type_e::StopPoint: return indexes(this->stop_point_list);
+    case Type_e::StopPoint: return indexes(this->stop_point_list); break;
```
TeXitoi: break here is useless as you return

source/type/type.cpp
```diff
@@ -655,6 +671,20 @@ std::vector<idx_t> StopPoint::get(Type_e type, const PT_Data &) const {
...
+            }
+        }
+        break;
+    case Type_e::PhysicalMode:
+        for(const auto& jpp: journey_pattern_point_list){
+            if(jpp->journey_pattern && jpp->journey_pattern->physical_mode){
+                result.push_back(jpp->journey_pattern->physical_mode->idx);
+            }
+        }
+        break;
```
TeXitoi, line 701: You should unique the types as in StopArea::get using a std::set, no?

source/type/type.cpp
azime: OK

azime:
corrections

TeXitoi:
@azime forgot to  push?

kinnou02:
jenkins, test this please.

PR #958, opened 2015-03-13
Raptor: fix second pass

TeXitoi:
Should fix (at least some of) artemis and http://jira.canaltp.fr/browse/NAVITIAII-1644


kinnou02:
jenkins, test this please.

PR #960, opened 2015-03-16
Places fix tokenizer navitiaii 1158

kadhikari:
This pull request concerns : http://jira.canaltp.fr/browse/NAVITIAII-1158
Modifications are as followings:
1. While creation dictionnary, for each key of synonym existing in name, words in values are also added instead of replacing.
2. While creation dictionnary, for each value of synonym existing in name, words in key are also added instead of replacing.
3. While searching ("find_complete" and "find_partial_with_pattern"), synonym is not used.

For more examples please see http://jira.canaltp.fr/browse/NAVITIAII-1158

### DIFF
--- a/source/autocomplete/autocomplete.h
+++ b/source/autocomplete/autocomplete.h
@@ -143,7 +143,6 @@ struct Autocomplete
         int distance = 0;
 
         //Appeler la méthode pour traiter les synonymes avant de les ajouter dans le dictionaire:
-        //auto vec_word = tokenize_without_replace(str, synonyms);
         auto vec_word = tokenize(str, synonyms);
         //créer des patterns pour chaque mot et les ajouter dans temp_pattern_map:
         add_vec_pattern(vec_word, position);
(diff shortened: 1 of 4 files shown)


TeXitoi:
If I have `"mairie de vannes"` in our base, we search for `"hotel de vannes", does is work with:
-  `synonyms["hotel de ville"] = "mairie"`, or
-  `synonyms["mairie"] = "hotel de ville"`.

And can we test that? Else, the PR is OK to me, only I can't see where we manage the now symetry of the synonyms.

kadhikari:
1. If we have synonys["hotel de ville"]="mairie", we don't need another synonyms["mairie"]="hotel de ville" because for each synonyms.key found in database we replace by synonyms.value and for each synonyms.value found in database we replace by synonyms.key.
2. I will add some tests.
3. I don't get what do you mean by "Else, the PR is OK to me, only I can't see where we manage the now symetry of the synonyms."

kadhikari:
Guillaume, i added some tests as you asked.

TeXitoi:
3 -> I mean that this PR is OK for me except for the remarks.

PR #962, opened 2015-03-18
Improve raptor solution reader

TeXitoi:


source/routing/raptor.cpp
```diff
@@ -193,6 +193,50 @@ void RAPTOR::first_raptor_loop(const vec_stop_point_duration& dep,
     boucleRAPTOR(accessibilite_params, clockwise, disruption_active, max_transfers);
 }
 
+namespace {
+struct StartingPointSndPhase {
+    SpIdx sp_idx;
+    unsigned count;
+    DateTime end_dt;
+    unsigned walking_t;
```
kinnou02: the fallback duration I think?

source/routing/raptor.cpp
```diff
@@ -193,6 +193,50 @@ void RAPTOR::first_raptor_loop(const vec_stop_point_duration& dep,
     boucleRAPTOR(accessibilite_params, clockwise, disruption_active, max_transfers);
 }
 
+namespace {
+struct StartingPointSndPhase {
+    SpIdx sp_idx;
+    unsigned count;
```
kinnou02, line 199: Count of what? the same as used by raptor?

TeXitoi, line 199: Exactly

PR #964, opened 2015-03-19
Places fix quality 100 navitiaii 1496

kadhikari:
Modification:
places in the result sorted as follows
1. by place type
2. by score (except quality = 100 on the top of each type)
3. by quality
4. then by name

A complete test with descriptions is also added.

### DIFF
--- a/source/autocomplete/autocomplete_api.cpp
+++ b/source/autocomplete/autocomplete_api.cpp
@@ -353,28 +353,17 @@ pbnavitia::Response autocomplete(const std::string &q,
             const auto a_order = get_embedded_type_order(a.embedded_type());
             const auto b_order = get_embedded_type_order(b.embedded_type());
             return  a_order < b_order;
-        } else if (a.quality() == 100  || b.quality() == 100){
-            //Sort by quality, score and then by name
-            if (a.quality() == b.quality()) {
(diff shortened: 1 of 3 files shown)


kadhikari:
Concerns ticket : http://jira.canaltp.fr/browse/NAVITIAII-1496

source/autocomplete/autocomplete_api.cpp
```diff
@@ -345,22 +345,34 @@ pbnavitia::Response autocomplete(const std::string &q,
...
+            if (a.quality() == b.quality()) {
+                if (a.score() == b.score()) {
+                    return boost::algorithm::lexicographical_compare(a.name(), b.name(), boost::is_iless());
+                } else {
+                    return a.score() > b.score();
+                }
+
+            } else {
+                return a.quality() > b.quality();
+            }
```
TeXitoi: I prefere reading comparators with the following pattern:

``` c++
if (a.a != b.a) { return a.a < b.a; }
if (a.b != b.b) { return a.b < b.b; }
...
```

It's easier to follow the order. If you can rewrite this comparator using this kind of style, it may increase its readability.

kadhikari:
Guillaume,
After our discussion I simplified the sort method.

PR #965, opened 2015-03-20
fare: add some reference for less copy

kinnou02:
40% of the time passed in compute_fare is dedicated as copying Label


source/fare/fare.cpp
```diff
@@ -153,9 +153,9 @@ results Fare::compute_fare(const routing::Path& path) const {
                 if (! valid(g[v], section_key))
                     continue;
 
-                for (Label label: labels[u]) {
+                for (Label& label: labels[u]) {
```
TeXitoi: const Label&

PR #968, opened 2015-03-24
jormungandr: filter on too long duration can select standard by duration

kinnou02:
it's now possible to choose the criteria used for select the journey
used a standard for calculate the max_duration, before it was only on
departure_time/arrival_time, the select on duration has been added.

Before journey with other fallback mode than walking were ignored, we
can now choose the modes allowed.

Refer to: http://jira.canaltp.fr/browse/NAVITIAII-1651


source/jormungandr/jormungandr/scenarios/default.py
```diff
@@ -324,19 +324,11 @@ def _find_max_duration(self, journeys, instance, clockwise):
         calculate the max duration of a journey from a pool of journey
         we search the earliest one who doesn't use car, bike or bss
         """
```
TeXitoi, line 327: I think the comment should be updated.

kinnou02:
Jenkins, test this please

PR #971, opened 2015-03-25
Optimisation

kinnou02:
Don't calculate fare if we only have a default ticket

It's taking a lot of time for doing nothing.
If there is fare data, the response now contains no tickets

Also remove a useless copy of section in raptor_api

### DIFF
--- a/source/routing/raptor_api.cpp
+++ b/source/routing/raptor_api.cpp
@@ -330,8 +330,7 @@ static void add_pathes(EnhancedResponse& enhanced_response,
                 // For a waiting section, if the previous public transport section,
                 // has estimated datetime we need to set it has estimated too.
                 if (pb_journey->sections_size() > 1) {
-                    for (int i=pb_journey->sections_size()-1; i>=0; --i) {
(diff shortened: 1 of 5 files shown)


source/routing/raptor_api.cpp
```diff
@@ -331,9 +331,9 @@ static void add_pathes(EnhancedResponse& enhanced_response,
                 // has estimated datetime we need to set it has estimated too.
                 if (pb_journey->sections_size() > 1) {
                     for (int i=pb_journey->sections_size()-1; i>=0; --i) {
-                        auto section = pb_journey->sections(i);
-                        if (section.type() == pbnavitia::PUBLIC_TRANSPORT) {
-                            if (section.add_info_vehicle_journey().has_date_time_estimated()) {
+                        auto section = pb_journey->mutable_sections(i);
```
TeXitoi:
``` c++
const auto& section = pb_journey->sections(i)
```

please, or better, if it works:

``` c++
for (const auto& section: pb_journey->sections()) {
```

kinnou02:
done

PR #972, opened 2015-03-25
do not remove mostly equal vjs

TeXitoi:


kinnou02:
it's related to the dst bug we were speaking about the other day?

PR #982, opened 2015-04-02
Disruption: compute a disruption status on journey

antoine-de:
it was done in jormungandr, but was way too slow, moved it in kraken


source/routing/raptor_api.cpp
```diff
@@ -90,6 +90,44 @@ static void fill_shape(pbnavitia::Section* pb_section,
     }
 }
 
+static void
+_update_max_severity(boost::optional<type::new_disruption::Effect>& worst_disruption,
+                     type::new_disruption::Effect new_val) {
+
+    //the severity are sorted, the first one is the worst one
```
kinnou02: The effect no?

source/routing/raptor_api.cpp
```diff
@@ -240,11 +278,12 @@ static void add_pathes(EnhancedResponse& enhanced_response,
             }
         }
 
+        boost::optional<type::new_disruption::Severity> most_serius_disruption_severity;
```
kinnou02: it's never used

PR #983, opened 2015-04-02
Jormun: Remove old disruptions

antoine-de:
remove old disruptions


source/jormungandr/jormungandr/interfaces/v1/fields.py
```diff
@@ -421,13 +407,10 @@ def output(self, key, obj):
     "cause": fields.String(),
     "severity": NonNullNested(disruption_severity),
     "messages": NonNullList(NonNullNested(disruption_message)),
+    "uri": fields.String(),
```
kinnou02: why?

source/type/realtime.proto
```diff
@@ -6,30 +6,6 @@ message TCObject{
...
-    //number of seconds since midnigth
-    optional uint32 start_application_daily_hour = 7;
-    optional uint32 end_application_daily_hour = 8;
-    optional string active_days = 9;
-
-    repeated LocalizedMessage localized_messages = 10;
-    required MessageStatus message_status = 11;
-}
-
 message AtPerturbation{
```
kinnou02: It can be removed too I think?

PR #985, opened 2015-04-02
Jormungandr: those functions were called twice

antoine-de:


source/jormungandr/jormungandr/interfaces/v1/Journeys.py
```diff
@@ -580,8 +580,6 @@ def __init__(self):
             parser_post.args[index].location = "json"
 
     @add_debug_info()
-    @clean_links()
```
kinnou02, line 583: You can do the same for the post method I think

l-vincent-l, line 583: Btw, it should be an OPTIONS method.

antoine-de, line 583: why ?

antoine-de, line 583: remove in post done

l-vincent-l, line 583: If you wanna be restful, post should create a resource.
Actually, OPTIONS is not in restful defintions...
So...

The good restful way to do it is to create requests resources, and then call it

PR #987, opened 2015-04-03
jormungandr: create address id un kraken

TeXitoi:
Performances improvements must be between 3 and 10%

### DIFF
--- a/source/jormungandr/jormungandr/interfaces/v1/ResourceUri.py
+++ b/source/jormungandr/jormungandr/interfaces/v1/ResourceUri.py
@@ -52,7 +52,6 @@ class ResourceUri(StatedResource):
             self.method_decorators.append(add_computed_resources(self))
             self.method_decorators.append(add_pagination_links())
             self.method_decorators.append(clean_links())
-        self.method_decorators.append(add_address_poi_id(self))
 
         if authentication:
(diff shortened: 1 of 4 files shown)


source/jormungandr/jormungandr/interfaces/v1/ResourceUri.py
```diff
@@ -95,39 +94,6 @@ def get_filter(self, items):
...
-                    if 'address' == type_ and 'coord' in objects:
-                        lon = objects['coord']['lon']
-                        lat = objects['coord']['lat']
-                        objects['id'] = lon + ';' + lat
-                    if 'embedded_type' in objects.keys() and\
-                        (objects['embedded_type'] == 'address' or
-                         objects['embedded_type'] == 'poi' or
-                         objects['embedded_type'] == 'administrative_region') and\
-                         objects['embedded_type'] in objects:
-                        objects["id"] = objects[objects['embedded_type']]["id"]
```
kinnou02, line 124: I don't see where do you set this id, it was already in the response?

TeXitoi, line 124: In practice, that's a modification, not a creation. I think it's created in fields, generic_type_admin or the like.

PR #994, opened 2015-04-09
Min fallback in kraken

antoine-de:


source/routing/raptor_api.cpp
```diff
@@ -693,9 +692,15 @@ get_stop_points( const type::EntryPoint &ep, const type::Data& data,
...
                     data.pt_data->stop_point_proximity_list,
                     use_second);
-        LOG4CPLUS_TRACE(logger, "find " << tmp_sn.size() << " stop_points");
+        LOG4CPLUS_TRACE(logger, tmp_sn.size() << " stop_points found");
         for(auto idx_duration : tmp_sn) {
             auto sp_idx = SpIdx(idx_duration.first);
+
+            if (idx_duration.second < ep.streetnetwork_params.min_duration) {
+                // we sometimes want a minimum fallback duration (not to take the car for 1mn for example)
+                LOG4CPLUS_INFO(logger, idx_duration.second << " is not enough, we skip this stop");
```
TeXitoi: INFO is a bit high, I thing it should be DEBUG or TRACE

antoine-de: oups i think it should not even pe printed, was for debug.

kinnou02: A log in trace can be useful I think

source/type/request.proto
```diff
@@ -63,6 +63,10 @@ message StreetNetworkParams{
     optional int32 max_bike_duration_to_pt      = 14;
     optional int32 max_bss_duration_to_pt       = 15;
     optional int32 max_car_duration_to_pt       = 16;
+
+    optional int32 min_bss_duration             = 17;
```
kinnou02: you can put some default value, this way during the upgrade we will not break all the platform

source/routing/tests/routing_api_test.cpp
```diff
@@ -1360,6 +1359,27 @@ BOOST_FIXTURE_TEST_CASE(bss_test, streetnetworkmode_fixture<test_speed_provider>
...
+    origin.streetnetwork_params.min_duration = 20_s;
+    origin.streetnetwork_params.max_duration = bt::pos_infin;
+
+    ng::StreetNetwork sn_worker(*b.data->geo_ref);
+    sn_worker.init(origin);
+    auto sp = nr::get_stop_points(origin, *b.data, sn_worker, false);
+    BOOST_CHECK_EQUAL(sp.size(), 2);
+
+    origin.streetnetwork_params.min_duration = 21_s;
+    //it takes 20s to go to B, so if the min is 30s, we cannot touch the stop point B,
```
TeXitoi: unsyncronised comment

PR #1000, opened 2015-04-16
ed2nav: fix bug with lacking edges on street network

kinnou02:
When choosing nodes to be ignored in the graph we were using the
vertex_t of the node and not it's id, so we were deleting the wrong
node, that caused some edge to be deleted too.

Ironically when removing small connected component we were in fact
randomly created some of them.

Refer to: http://jira.canaltp.fr/browse/NAVITIAII-1691

### DIFF
--- a/source/ed/ed_reader.cpp
+++ b/source/ed/ed_reader.cpp
@@ -1007,6 +1007,7 @@ void EdReader::fill_vector_to_ignore(navitia::type::Data& , pqxx::work& work,
     std::unordered_map<uint64_t, idx_t> node_map_temp;
     std::unordered_map<idx_t, size_t> way_nb_edges;
 
+
     // chargement des vertex
     std::string request = "select id, ST_X(coord::geometry) as lon, ST_Y(coord::geometry) as lat from georef.node;";
     pqxx::result result = work.exec(request);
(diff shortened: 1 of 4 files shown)


antoine-de:
rah you got the thousandth ! :sob:

source/georef/georef.h
```diff
@@ -151,6 +151,16 @@ struct Way :public nt::Nameable, nt::Header{
     }
     std::string get_label() const;
 
+#ifdef _DEBUG_DIJKSTRA_QUANTUM_
+    template <typename Stream, typename G>
+    void print(Stream& stream, const G& g){
+        for(auto& edge: this->edges){
+            stream << "LINESTRING(" << g[edge.first].coord.lon() << " " << g[edge.first].coord.lat()
```
TeXitoi: you can add `stream << std::setprecision(16) << "LINESTRING...` to have a correct precision.

source/ed/ed_reader.cpp
```diff
@@ -1078,8 +1079,9 @@ void EdReader::fill_vector_to_ignore(navitia::type::Data& , pqxx::work& work,
         BOOST_FOREACH(navitia::georef::edge_t e, boost::out_edges(vertex_idx, geo_ref_temp.graph)) {
             uint64_t target = boost::target(e, geo_ref_temp.graph);
             edge_to_ignore.insert({source, target});
```
TeXitoi: source is a "osm_idex" and target is a "node_descriptor", I can't understand how it can work. And osm_idex is a very strange name.

kinnou02: it won't work :) or more precisely, it work because it don't do anything...

source/ed/ed_reader.cpp
```diff
@@ -1072,13 +1072,12 @@ void EdReader::fill_vector_to_ignore(navitia::type::Data& , pqxx::work& work,
         if (nb_elt_in_component / principal_component->second >= min_non_connected_graph_ratio)
             continue; //big enough, we skip
 
-        uint64_t source = osmid_idex[vertex_idx];
+        uint64_t source = map_idx_to_id[vertex_idx];
```
antoine-de, line 1075: why don't you use the edge_to_ignore anymore ? 
If it is not useful anymore, remove it

PR #1001, opened 2015-04-17
Stat: add some basic indexes

kinnou02:


TeXitoi:
There is no model (as source/sql/models/*) in stat_persistor?

kinnou02:
nop, we don't use models in stat persistor

PR #1003, opened 2015-04-21
Streetnetwork: add bss virtual cost

antoine-de:
add a virtual additional cost to bss : "les 2 minutes du peuple" for BSS/walking "connection".

It can be viewed as "the bss has to at least improve the walking
solution by that time to be taken".

It does not change the real fallback duration, only the computation in
the shortest path.

On Paris instance, add 10ms on the street network computation time per query.

[Review on Reviewable](https://reviewable.io/reviews/canaltp/navitia/1003)

### DIFF
--- a/source/georef/street_network.cpp
+++ b/source/georef/street_network.cpp
@@ -60,8 +60,8 @@ void StreetNetwork::init(const type::EntryPoint& start, boost::optional<const ty
                                start.streetnetwork_params.speed_factor, start.streetnetwork_params.bss_additional_cost);
 
     if (end) {
-        arrival_path_finder.init((*end).coordinates, (*end).streetnetwork_params.mode,
-                                 (*end).streetnetwork_params.speed_factor, (*end).streetnetwork_params.bss_additional_cost);
+        arrival_path_finder.init(end->coordinates, end->streetnetwork_params.mode,
+                                 end->streetnetwork_params.speed_factor, end->streetnetwork_params.bss_additional_cost);
     }
 }
 
@@ -177,7 +177,7 @@ Path StreetNetwork::get_direct_path(const type::EntryPoint& origin,
 PathFinder::PathFinder(const GeoRef& gref):
     geo_ref(gref) {}
 
-void PathFinder::init(const type::GeographicalCoord& start_coord, nt::Mode_e mode,
+void PathFinder::init(const type::GeographicalCoord& start_coord, const nt::Mode_e mode,
                       const float speed_factor, const navitia::time_duration bss_cost) {
     computation_launch = false;
     bss_additional_cost = bss_cost;
@@ -190,7 +190,7 @@ void PathFinder::init(const type::GeographicalCoord& start_coord, nt::Mode_e mod
 
     //we initialize the distances to the maximum value
     size_t n = boost::num_vertices(geo_ref.graph);
-    distances.assign(n, {bt::pos_infin});
+    distances.assign(n, cost());
     //for the predecessors no need to clean the values, the important one will be updated during search
     predecessors.resize(n);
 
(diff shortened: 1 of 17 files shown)


source/georef/street_network.cpp
```diff
@@ -56,10 +56,12 @@ StreetNetwork::StreetNetwork(const GeoRef &geo_ref) :
...
 
 void StreetNetwork::init(const type::EntryPoint& start, boost::optional<const type::EntryPoint&> end) {
-    departure_path_finder.init(start.coordinates, start.streetnetwork_params.mode, start.streetnetwork_params.speed_factor);
+    departure_path_finder.init(start.coordinates, start.streetnetwork_params.mode,
+                               start.streetnetwork_params.speed_factor, start.streetnetwork_params.bss_additional_cost);
 
     if (end) {
-        arrival_path_finder.init((*end).coordinates, (*end).streetnetwork_params.mode, (*end).streetnetwork_params.speed_factor);
+        arrival_path_finder.init((*end).coordinates, (*end).streetnetwork_params.mode,
+                                 (*end).streetnetwork_params.speed_factor, (*end).streetnetwork_params.bss_additional_cost);
```
TeXitoi: you can use `end->field`

source/georef/street_network.h
```diff
@@ -35,20 +35,65 @@ www.navitia.io
...
 namespace navitia { namespace georef {
 
-struct SpeedDistanceCombiner : public std::binary_function<navitia::time_duration, navitia::time_duration, navitia::time_duration> {
+struct cost {
+    cost(navitia::time_duration d, bool b): duration(d), bss_taken(b) {}
+    cost(navitia::time_duration d): duration(d) {}
+    cost() {}
+    bool operator<(const cost& c) const { return duration < c.duration; }
+    bool operator==(const cost& c) const { return duration == c.duration && bss_taken == c.bss_taken; }
+    navitia::time_duration duration = {};
```
TeXitoi: better to have default to pos_inf, no? And then you can init with cost()

source/georef/street_network.h
```diff
@@ -109,18 +152,24 @@ struct PathFinder {
...
 
     /**
      *  Update the structure for a given starting point and transportation mode
      *  The init HAS to be called before any other methods
      */
-    void init(const type::GeographicalCoord& start_coord, nt::Mode_e mode, const float speed_factor);
+    void init(const type::GeographicalCoord& start_coord,
+              nt::Mode_e mode,
+              const float speed_factor,
+              const navitia::time_duration);
```
TeXitoi, line 171: you can constify mode

source/georef/street_network.h
```diff
@@ -148,13 +197,16 @@ struct PathFinder {
...
-                                               boost::get(&Edge::duration, geo_ref.graph), // weigth map
-                                               boost::identity_property_map(),
-                                               std::less<navitia::time_duration>(),
-                                               SpeedDistanceCombiner(speed_factor), //we multiply the edge duration by a speed factor
-                                               navitia::seconds(0),
+                                               boost::typed_identity_property_map<edge_t>(), // weigth map
+                                               boost::typed_identity_property_map<vertex_t>(),
+                                               std::less<cost>(),
+                                               speed_combiner,
+                                               cost(navitia::seconds(0), false),
```
TeXitoi: Can't you say `cost(0_s, false)` ?

antoine-de:
Jenkins, test this please

antoine-de:
Do not compile with boost 1.49 due to a bug in boost::graph (https://svn.boost.org/trac/boost/ticket/8398)

we wait to see if the new debian stable is released at the due date to see if we can upgrade our boost version or if we need to backport / dupplicate boost...

pbougue:
The target is Valhalla on that need.
Been there for too long and not easily mergeable anymore.

PR #1004, opened 2015-04-21
Implement a way to have the same user multiple time in the database

kinnou02:
it's possible to have multiple users with the same login/email for
different endpoint.

So it's possible to handle multiple users bases, typically if we want to
sell the api with different name.


kinnou02:
:heart: it's more sexy now?

source/tyr/tyr/resources.py
```diff
@@ -464,3 +484,32 @@ def post(self, user_id):
...
+            end_point = models.EndPoint()
+            end_point.name = args['name']
+            if 'hostnames' in request.json:
+                for host in request.json['hostnames']:
+                    end_point.hosts.append(models.Host(host))
+
+            db.session.add(end_point)
+            db.session.commit()
+        except (sqlalchemy.exc.IntegrityError, sqlalchemy.orm.exc.FlushError):
+            return ({'error': 'duplicate entry'}, 409)
```
antoine-de: can you log it and return the real error ?

antoine-de:
:dancers: yes!

PR #1005, opened 2015-04-21
add documentation sncf

kinnou02:


antoine-de:
not sure about the use of this doc, this should not be in the repository, but it's ok with me if it's a customer demand.

kinnou02:
Currently Coline wants it here, so we can start to work on it.
We will maybe move it afterwards.

If we let this doc we can hope that it will be updated at the same time of main documentation...

PR #1007, opened 2015-04-22
improvement of the logging

kinnou02:


source/jormungandr/jormungandr/instance.py
```diff
@@ -250,7 +251,8 @@ def send_and_receive(self,
                 socket.setsockopt(zmq.LINGER, 0)
                 socket.close()
                 if not quiet:
-                    logging.error("request: " + str(request) + " failed " + self.socket_path)
+                    logger = logging.getLogger(__name__)
+                    logger.error("request: " + str(request) + " failed " + self.socket_path)
```
antoine-de: use %s no ?

source/jormungandr/jormungandr/exceptions.py
```diff
@@ -110,15 +111,13 @@ def __init__(self, msg):
...
     error = exception.__class__.__name__ + " " + message + " " + request.url
 
-    if hasattr(exception, "data"):
-        # if an exception is thrown with 'brief_log', we just log it in one line in debug
-        if 'brief_log' in exception.data:
-            logging.debug(error)
-            return
-
-    logging.exception(error)
+    if isinstance(exception, (HTTPException, RegionNotFound)):
```
antoine-de, line 120: I think we should keep the brief log facility

kinnou02, line 120: And I don't think so,  we will need a little help here

TeXitoi, line 120: I vote "he's blu^W^Wremove brief log"

PR #1008, opened 2015-04-23
Jormungandr: Add distance param

antoine-de:
This way you can do:

http://api.navitia.io/v1/coverage/default/physical_modes/physical_mode:Bus/coords/2.272498;48.836897/stop_areas?distance=700

to get all stop areas with bus within 700m from the coord

linked to http://jira.canaltp.fr/browse/NAVITIAII-1659

### DIFF
--- a/source/jormungandr/tests/disruptions_tests.py
+++ b/source/jormungandr/tests/disruptions_tests.py
@@ -181,7 +181,7 @@ class TestDisruptions(AbstractTestFixture):
         response = self.query_region('networks/base_network/arrivals?from_datetime=20120614T070000&_current_datetime=20120614T080000')
 
         arrivals = get_not_null(response, 'arrivals')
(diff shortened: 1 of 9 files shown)


source/jormungandr/tests/disruptions_tests.py
```diff
@@ -181,7 +181,7 @@ def test_disruption_with_arrival(self):
         response = self.query_region('networks/base_network/arrivals?from_datetime=20120614T070000&_current_datetime=20120614T080000')
 
         arrivals = get_not_null(response, 'arrivals')
-        eq_(len(arrivals), 1)
+        eq_(len(arrivals), 2)
```
kinnou02, line 184: ?

antoine-de, line 184: I added a stop in the test referential

PR #1018, opened 2015-04-28
Jormungandr: New scenario

antoine-de:
Add a new scenario to jormungandr.

This scenario aims to take more benefits from all the solutions from RAPTOR (with our new reader).

For the moment we do not handle well the max_nb_journeys parameter as we only filter duplicate journeys (and soon incoherent journeys). 

I think the max_nb_journey is quite complicated to code right and we should take a look to multiobjectif solution filtering algorithms like NSGAII.

### DIFF
--- a/source/jormungandr/jormungandr/scenarios/journey_filter.py
+++ b/source/jormungandr/jormungandr/scenarios/journey_filter.py
@@ -30,8 +30,26 @@ from collections import namedtuple
 import logging
 import itertools
 import datetime
+from operator import attrgetter
 from jormungandr.scenarios.utils import compare
 from navitiacommon import response_pb2
+
+
+def _delete_journeys(responses, request):
+
+    if request.get('debug', False):
+        return
+
+    nb_deleted = 0
+    for r in responses:
+        for idx, j in enumerate(reversed(r.journeys)):
+            if not _to_be_deleted(j):
+                continue
+            del r.journeys[idx]
+            nb_deleted += 1
+
+    if nb_deleted:
+        logging.getLogger(__name__).info('filtering {} journeys'.format(nb_deleted))
 
 
 def filter_journeys(response_list, request):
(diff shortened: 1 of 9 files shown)


source/jormungandr/jormungandr/scenarios/tests/journey_compare_tests.py
```diff
@@ -6,7 +6,7 @@
 # Hope you'll enjoy and contribute to this project,
 #     powered by Canal TP (www.canaltp.fr).
 # Help us simplify mobility and open public transport:
-#     a non ending quest to the responsive locomotion way of traveling!
+#     a non ending quest to the res locomotion way of traveling!
```
kinnou02: ?

source/jormungandr/jormungandr/scenarios/utils.py
```diff
@@ -207,3 +207,78 @@ def build_pagination(request, resp):
...
 }
+
+
+def get_or_default(request, val, default):
+    """
+    Flask add all API param in the request dict with None by default,
+    so here is a simple helper to get the default value is the key
+    is not in the dict or if the value is None
+    """
+    return request.get(val, default) or default
```
kinnou02: You have to check the return directly, in this case 0 or '' would be considered as None

source/jormungandr/jormungandr/scenarios/journey_filter.py
```diff
@@ -0,0 +1,257 @@
...
+    -------------------------------
+             asap pseudo duration
+
+    ------------------------------------------------------
+                       journey pseudo duration
+
+    """
+    requested_dt = original_request['datetime']
+    if original_request.get('clockwise', True):
+        speudo_asap_duration = asap_journey.arrival_date_time - requested_dt
```
kinnou02: typo

PR #1020, opened 2015-05-04
Ed: first step to read the new comment_link file

antoine-de:
for the moment we only store the last comment for each object, the PR
with the reading of multiple elements will land soon

### DIFF
--- a/source/ed/connectors/fusio_parser.cpp
+++ b/source/ed/connectors/fusio_parser.cpp
@@ -258,9 +258,10 @@ void TransfersFusioHandler::fill_stop_point_connection(ed::types::StopPointConne
 
 void StopTimeFusioHandler::init(Data& data) {
     StopTimeGtfsHandler::init(data);
-    itl_c = csv.get_pos_col("stop_times_itl"),
-               desc_c = csv.get_pos_col("stop_desc"),
-            date_time_estimated_c = csv.get_pos_col("date_time_estimated");
+    itl_c = csv.get_pos_col("stop_times_itl");
+    desc_c = csv.get_pos_col("stop_desc");
+    date_time_estimated_c = csv.get_pos_col("date_time_estimated");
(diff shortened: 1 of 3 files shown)


source/ed/connectors/fusio_parser.cpp
```diff
@@ -1127,6 +1132,85 @@ void AdminStopAreaFusioHandler::handle_line(Data& data, const csv_row& row, bool
...
+    if (it == map.end()) {
+        LOG4CPLUS_INFO(logger, "impossible to find " << obj_id << ", cannot add the comment, skipping line");
+        return false;
+    }
+    it->second->comment = comment;
+    return true;
+}
+
+template <typename T>
+bool add_comment(const std::unordered_map<std::string, std::vector<T>>& list_map, const std::string& obj_id, const std::string& comment) {
```
TeXitoi, line 1155: long line

source/ed/connectors/fusio_parser.cpp
```diff
@@ -1127,6 +1132,85 @@ void AdminStopAreaFusioHandler::handle_line(Data& data, const csv_row& row, bool
...
 }
 
+void CommentLinksFusioHandler::init(Data&){
+    object_id_c = csv.get_pos_col("object_id");
+    object_type_c = csv.get_pos_col("object_type");
+    comment_id_c = csv.get_pos_col("comment_id");
+}
+
+template <typename T>
+bool add_comment(const std::unordered_map<std::string, T>& map, const std::string& obj_id, const std::string& comment) {
```
TeXitoi, line 1142: long line

source/ed/connectors/fusio_parser.cpp
antoine-de, line 1155: this code will soon disappear so I don't think it's worth bothering with lint.

PR #1021, opened 2015-05-04
TMA 1706

TeXitoi:
fix http://jira.canaltp.fr/browse/NAVITIAII-1706


source/type/geographical_coord.cpp
```diff
@@ -86,4 +86,40 @@ std::ostream & operator<<(std::ostream & os, const GeographicalCoord & coord){
     return os;
 }
 
+GeographicalCoord project(const LineString& line, const GeographicalCoord& p) {
```
kinnou02, line 89: With this method we can only project on a node of the graph right?
In simple case the result will be worse than before :worried:

PR #1022, opened 2015-05-04
Compilation fix on ubuntu 12.04 desktop

xavierraffin:


antoine-de:
you still use Ubuntu 12.04 ? I'm surprised that there is only this fix to make it work on this platform :smile: 

By the way we plan to move the requirement on kraken to a newer boost version because we need some more features in boost::graph. Do you think it would be a problem on your part ?

PR #1024, opened 2015-05-05
jormungandr: fix a 500 if a type was misspelled

kinnou02:


source/jormungandr/jormungandr/interfaces/v1/ResourceUri.py
```diff
@@ -70,6 +70,8 @@ def get_filter(self, items, args):
                     if item == "calendars":
                         type_ = 'calendar'
                     else:
+                        if item not in collections_to_resource_type:
+                            abort(400, message="unknown type: {}".format(item))
```
antoine-de, line 74: sayyyy could you add a jormungandr test ?

PR #1036, opened 2015-05-12
ed: correction in cities

azime:
correction in cities


kinnou02:
We will have to do the job correctly in cities during the next sprint

PR #1038, opened 2015-05-13
routing: fix duration and date_times for direct path when using speed_factor

TeXitoi:
fix http://jira.canaltp.fr/browse/NAVITIAII-1612

### DIFF
--- a/source/jormungandr/tests/routing_tests.py
+++ b/source/jormungandr/tests/routing_tests.py
@@ -94,11 +94,12 @@ class TestJourneys(AbstractTestFixture):
         assert len(response['journeys']) == 1
         assert response['journeys'][0]["type"] == "non_pt_walk"
         assert len(response['journeys'][0]['sections']) == 1
-        print response['journeys'][0]['duration']
-        print response['journeys'][0]['sections'][0]['duration']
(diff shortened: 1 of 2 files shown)


source/jormungandr/tests/routing_tests.py
```diff
@@ -85,6 +85,21 @@ def test_other_filtering(self):
...
+                                     format(query=journey_basic_query))
+
+        assert len(response['journeys']) == 1
+        assert response['journeys'][0]["type"] == "non_pt_walk"
+        assert len(response['journeys'][0]['sections']) == 1
+        print response['journeys'][0]['duration']
+        print response['journeys'][0]['sections'][0]['duration']
+        assert response['journeys'][0]['duration'] == response['journeys'][0]['sections'][0]['duration']
+        assert response['journeys'][0]['departure_date_time'] == response['journeys'][0]['sections'][0]['departure_date_time']
+        assert response['journeys'][0]['arrival_date_time'] == response['journeys'][0]['sections'][0]['arrival_date_time']
```
antoine-de, line 101: could you add a test checking that the duration is 50% greater than with a speed factor=1 ?

PR #1039, opened 2015-05-13
autocomplete: Correction for ghostword

kadhikari:
test added
concerns http://jira.canaltp.fr/browse/NAVITIAII-1724

### DIFF
--- a/source/autocomplete/autocomplete.h
+++ b/source/autocomplete/autocomplete.h
@@ -505,6 +505,18 @@ struct Autocomplete
             }
         }
 
+        //We discard the GhostWords like de, le, la configured in synonyms as below
+        //de,
+        //le,
+        //la,
+        //For each synonyms.key found in strTemp if synonyms.value is empty delete the substring
+        for(const auto & it : synonyms){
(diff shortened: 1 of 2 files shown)


TeXitoi:
It could be great to have a complete test with the search if possible.

Else, OK for me.

source/autocomplete/autocomplete.h
```diff
@@ -505,6 +505,18 @@ struct Autocomplete
             }
         }
 
+        //We discard the GhostWords like de, le, la configured in synonyms as below
+        //de,
+        //le,
+        //la,
+        //For each synonyms.key found in strTemp if synonyms.value is empty delete the substring
+        for(const auto & it : synonyms){
```
antoine-de, line 513: I think you should store that in a separate container with only ghost word.

source/autocomplete/autocomplete.h
```diff
@@ -505,6 +505,18 @@ struct Autocomplete
...
+            if  ((boost::regex_search(strTemp,boost::regex("\\<" + it.first + "\\>")))
+                && (it.second.empty())){
```
antoine-de, line 515: you should test first the emptiness before checking the regexp (but it is useless if you use a specific container ;) )

PR #1042, opened 2015-05-18
jormungandr: in places return an empty list when there is no result

kinnou02:


kinnou02:
In some API we return a 404 if there is no result, in this one we return a 200 no matters what.
What do you think, do we return a 404 here too?

PR #1047, opened 2015-05-21
Add distance param to documentation

vinzentt:


documentation/navitia.io/source/integration.rst
```diff
@@ -273,7 +273,7 @@ Once you have selected a region, you can explore the public transportation objec
 easily with these apis. You just need to add at the end of your url 
 a collection name to see all the objects of a particular collection.
 To see an object add the id of this object at the end of the collection's url.
-The only arguments are the ones of `paging`_.
+The arguments are the ones of `paging`_.
```
stifoon: I would have written "You can add arguments from `paging`_."

vinzentt: You mean you would not have written what you wrote? =)
Personally, I would prefer sthg like "`paging`_ arguments may be used to paginate results."

stifoon: na na na na na...
Yes, your proposition is fine. Nay better than mine :)

antoine-de:
thanks for the PR @vinzentt !

PR #1054, opened 2015-05-27
osm2ed: fix import of blind alley

kinnou02:
We were only importing edge with nodes used at least by two edges, but
in case of blind alley we need this edges too, so, we now import all
edges from a street even is they are used only one time!


kinnou02:
With this PR the memory consumption increase a lot,  we multiply by two the size of the streetnetwork on fr-idf.

PR #1057, opened 2015-05-28
All: rename kraken_version in project_version

antoine-de:
with the hugin split, we need to better separate utils and navitia


kinnou02:
jenkins, test this please.

PR #1058, opened 2015-05-29
Isochron missing results

djludo:


source/jormungandr/tests/isochrone_tests.py
```diff
@@ -32,37 +32,52 @@
 from check_utils import *
 
 
-@dataset(["main_routing_test"])
+@dataset(["basic_routing_test"])
```
antoine-de, line 35: Nice some tests :v: 

can you rename the class ? it should be called TestIsochrone no ?

source/jormungandr/tests/isochrone_tests.py
```diff
@@ -32,37 +32,52 @@
...
     """
     Test the structure of the journeys response
     """
 
     def test_from_isochrone_coord(self):
-        #NOTE: we query /v1/coverage/main_routing_test/journeys and not directly /v1/journeys
+        #NOTE: we query /v1/coverage/basic_routing_test/journeys and not directly /v1/journeys
         #not to use the jormungandr database
-        query = "v1/coverage/main_routing_test/journeys?from={}&datetime={}"
+        query = "v1/coverage/basic_routing_test/journeys?from={}&datetime={}"
```
antoine-de, line 44: not important but if you don't want to bother with the region name (and if you have only one region in you dataset), you can just do `self.query_region('journeys?from={}&datetime={}'.format(s_coord, "20120614T080000"))`

PR #1059, opened 2015-06-01
fix a segfault when updating a disruption on two stop_areas

kinnou02:
We were deleting the journey_pattern multiple times, as you can guess it didn't go well..

### DIFF
--- a/source/kraken/tests/fill_disruption_from_chaos_tests.cpp
+++ b/source/kraken/tests/fill_disruption_from_chaos_tests.cpp
@@ -113,8 +113,10 @@ BOOST_AUTO_TEST_CASE(add_impact_on_stop_area) {
 
 BOOST_AUTO_TEST_CASE(add_impact_and_update_on_stop_area) {
     ed::builder b("20120614");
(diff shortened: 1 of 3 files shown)


source/kraken/tests/fill_disruption_from_chaos_tests.cpp
```diff
@@ -0,0 +1,127 @@
...
+    severity->set_effect(transit_realtime::Alert_Effect_NO_SERVICE);
+    auto* object = impact->add_informed_entities();
+    object->set_pt_object_type(chaos::PtObject_Type_stop_area);
+    object->set_uri("stop_area:stop1");
+    auto* app_period = impact->add_application_periods();
+    app_period->set_start(ntest::to_posix_timestamp("20120614T123200"));
+    app_period->set_end(ntest::to_posix_timestamp("20120616T123200"));
+
+
+    navitia::add_disruption(disruption, *b.data->pt_data, *b.data->meta);
```
antoine-de, line 95: some more checks (and comments) would be useful.

you can check that the disruption is there, that the adapted objects are correctly created, that the theoric one are not changed, ...

PR #1061, opened 2015-06-02
Osm: filter osm admin on boundary=administrative

antoine-de:
this was done on cities but not on osm2ed...


antoine-de:
Jenkins, test this please

PR #1064, opened 2015-06-03
kraken: improvement of disruptions application

kinnou02:
The circulation date of each circulation wasn't check, so it was
possible to disable a vj that wasn't impacted

### DIFF
--- a/source/kraken/fill_disruption_from_chaos.cpp
+++ b/source/kraken/fill_disruption_from_chaos.cpp
@@ -326,8 +326,9 @@ struct functor_add_vj {
         vj->is_adapted = true;
         vj->name = vj_ref.name;
         vj->company = vj_ref.company;
-        vj->next_vj = vj_ref.next_vj;
-        vj->prev_vj = vj_ref.prev_vj;
+        //TODO: we loose stay_in on impacted vj, we will need to work on this.
+        vj->next_vj = nullptr;
+        vj->prev_vj = nullptr;
         vj->meta_vj = vj_ref.meta_vj;
         vj->utc_to_local_offset = vj_ref.utc_to_local_offset;
         // The validity_pattern is only active on the period of the impact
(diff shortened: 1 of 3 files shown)


kinnou02:
over midnight case... yeah....

antoine-de:
we (huuuum you :stuck_out_tongue_closed_eyes: ) should also test some stay_in case.

There are no reason for those to go well :smile:

kinnou02:
kill me, kill me with fire :fire:

source/kraken/fill_disruption_from_chaos.cpp
```diff
@@ -79,6 +81,21 @@ make_cause(const chaos::Cause& chaos_cause, nt::new_disruption::DisruptionHolder
...
+    auto day_period = bt::time_period(bt::ptime(date, bt::hours(0)), bt::hours(24));
+    //we get the validity period of the imapct for this day
+    return day_period.intersection(period);
+}
+
+//return the time period of circulation of a vj for one day
+static boost::posix_time::time_period execution_period(const boost::gregorian::date& date,
+                                                          const nt::VehicleJourney& vj){
+    return bt::time_period(bt::ptime(date, bt::seconds(vj.stop_time_list.front().departure_time)),
+            bt::ptime(date, bt::seconds(vj.stop_time_list.back().arrival_time)));
```
antoine-de: I think you should add an epsilon to the arrival date, the end in not in the time_period

source/kraken/fill_disruption_from_chaos.cpp
```diff
@@ -381,13 +414,17 @@ struct add_impacts_visitor : public apply_impacts_visitor {
             tmp_vp.days.set(i, vj.adapted_validity_pattern->days[i]);
```
antoine-de: It really seems like a classic copy

PR #1071, opened 2015-06-10
Add line groups to navitia

eturck:
This Pull Request add the concept of line_groups in navitia based on "Group of lines" in transmodel.
Things added : 
- 2 new tables in ed : 
  - line_group (id, uri, name),
  - line_group_link (group_id, line_id, is_main_line).
- 2 more files imported in fusio2ed : 
  - line_groups.txt (line_group_id, line_group_name, main_line_id),
  - line_group_links.txt (line_group_id, line_id).
- Necessary code to write to ed, read from ed and load in kraken,
- Necessary code to write to protobuf in kraken and read protobuf in jormungandr,
- New URL in jormungandr : http://<ip>/coverage/<instance>/v1/line_groups
  line_groups
- Output group in v1/lines if the line is part of one or more 
  line_with_group
- Unit test for fusio2ed, jormungandr integration and ptref.

Some .proto files have been modified but they are stored in another repository. It's necessary to merge the pull request from this repository (navitia-proto) first, since in this one, source/navitia-proto refers to a new commit. ( see https://github.com/CanalTP/navitia-proto/pull/1 )

Waiting for your feedback, let me know if there is anything to change / improve :) !

### DIFF
--- a/fixtures/ed/ntfs/line_groups.txt
+++ b/fixtures/ed/ntfs/line_groups.txt
@@ -1,0 +1,4 @@
+line_group_id,line_group_name,main_line_id
+lg1,"Groupe lignes B",l2
+lg2,"Groupe ligne Bs seule",l3
+lg3,"Groupe inconnu",iddeligneprincipalequinexistepas
--- a/fixtures/ed/ntfs/line_groups.txt
+++ b/fixtures/ed/ntfs/line_groups.txt
@@ -1,4 +1,4 @@
 line_group_id,line_group_name,main_line_id
-lg1,"Groupe lignes B",l2
-lg2,"Groupe ligne Bs seule",l3
-lg3,"Groupe inconnu",iddeligneprincipalequinexistepas
+lg1,"B lines group",l2
+lg2,"Bn line only group",l3
+lg3,"Unknown group",unknownlineidtotestifitiscorrectlyignored
--- a/source/ed/connectors/fusio_parser.cpp
+++ b/source/ed/connectors/fusio_parser.cpp
@@ -667,6 +667,7 @@ void LineGroupFusioHandler::handle_line(Data& data, const csv_row& row, bool) {
     if(!(has_col(id_c, row) && has_col(name_c, row) && has_col(main_line_id_c, row))) {
         LOG4CPLUS_WARN(logger, "LineGroupFusioHandler: Line ignored in " << csv.filename <<
             " missing line_group_id, line_group_name or main_line_id column");
+        throw InvalidHeaders(csv.filename);
     }
     auto line = gtfs_data.line_map.find(row[main_line_id_c]);
     if (line == gtfs_data.line_map.end()) {
@@ -682,9 +683,8 @@ void LineGroupFusioHandler::handle_line(Data& data, const csv_row& row, bool) {
     ed::types::LineGroupLink line_group_link;
     line_group_link.line_group = line_group;
     line_group_link.line = line->second;
-    line_group_link.is_main_line = true;
 
-    gtfs_data.linked_lines_by_line_group_uri[line_group->uri].push_back(line->second->uri);
+    gtfs_data.linked_lines_by_line_group_uri[line_group->uri].insert(line->second->uri);
 
     data.line_groups.push_back(line_group);
     data.line_group_links.push_back(line_group_link);
@@ -700,6 +700,7 @@ void LineGroupLinksFusioHandler::handle_line(Data& data, const csv_row& row, boo
     if(!(has_col(line_group_id_c, row) && has_col(line_id_c, row))) {
         LOG4CPLUS_WARN(logger, "LineGroupLinksFusioHandler: Line ignored in " << csv.filename <<
         " missing line_group_id or line_id column");
+        throw InvalidHeaders(csv.filename);
     }
     auto line_group = gtfs_data.line_group_map.find(row[line_group_id_c]);
     if (line_group == gtfs_data.line_group_map.end()) {
@@ -713,8 +714,8 @@ void LineGroupLinksFusioHandler::handle_line(Data& data, const csv_row& row, boo
         return;
     }
 
-    std::vector<std::string> linked_lines = gtfs_data.linked_lines_by_line_group_uri[line_group->second->uri];
-    if(std::find(linked_lines.begin(), linked_lines.end(), line->second->uri) != linked_lines.end()) {
+    auto linked_lines = gtfs_data.linked_lines_by_line_group_uri[line_group->second->uri];
+    if(linked_lines.find(line->second->uri) != linked_lines.end()) {
         // Don't insert duplicates
         return;
     }
@@ -722,9 +723,7 @@ void LineGroupLinksFusioHandler::handle_line(Data& data, const csv_row& row, boo
     ed::types::LineGroupLink line_group_link;
     line_group_link.line_group = line_group->second;
     line_group_link.line = line->second;
-    // To know if the line is the main line of the group, we compare URI
-    line_group_link.is_main_line = (line_group->second->main_line->uri == line->second->uri);
-    gtfs_data.linked_lines_by_line_group_uri[line_group->second->uri].push_back(line->second->uri);
+    gtfs_data.linked_lines_by_line_group_uri[line_group->second->uri].insert(line->second->uri);
     data.line_group_links.push_back(line_group_link);
 }
 
--- a/source/ed/connectors/fusio_parser.cpp
+++ b/source/ed/connectors/fusio_parser.cpp
@@ -657,6 +657,76 @@ void LineFusioHandler::handle_line(Data& data, const csv_row& row, bool is_first
     gtfs_data.line_map[line->uri] = line;
 }
 
+void LineGroupFusioHandler::init(Data&) {
+    id_c = csv.get_pos_col("line_group_id");
+    name_c = csv.get_pos_col("line_group_name");
+    main_line_id_c = csv.get_pos_col("main_line_id");
+}
+
+void LineGroupFusioHandler::handle_line(Data& data, const csv_row& row, bool) {
+    if(!(has_col(id_c, row) && has_col(name_c, row) && has_col(main_line_id_c, row))) {
+        LOG4CPLUS_WARN(logger, "LineGroupFusioHandler: Line ignored in " << csv.filename <<
+            " missing line_group_id, line_group_name or main_line_id column");
+    }
+    auto line = gtfs_data.line_map.find(row[main_line_id_c]);
+    if (line == gtfs_data.line_map.end()) {
+        LOG4CPLUS_WARN(logger, "LineGroupFusioHandler: Impossible to find the line " << row[main_line_id_c]);
+        return;
+    }
+    ed::types::LineGroup* line_group = new ed::types::LineGroup();
+    line_group->uri = row[id_c];
+    line_group->name = row[name_c];
+    line_group->main_line = line->second;
+
+    // Link main_line to line_group
+    ed::types::LineGroupLink line_group_link;
+    line_group_link.line_group = line_group;
+    line_group_link.line = line->second;
+    line_group_link.is_main_line = true;
+
+    line_group->linkedLinesURI.insert(line->second->uri);
+
+    data.line_groups.push_back(line_group);
+    data.line_group_links.push_back(line_group_link);
+    gtfs_data.line_group_map[line_group->uri] = line_group;
+}
+
+void LineGroupLinksFusioHandler::init(Data&) {
+    line_group_id_c = csv.get_pos_col("line_group_id");
+    line_id_c = csv.get_pos_col("line_id");
+}
+
+void LineGroupLinksFusioHandler::handle_line(Data& data, const csv_row& row, bool) {
+    if(!(has_col(line_group_id_c, row) && has_col(line_id_c, row))) {
+        LOG4CPLUS_WARN(logger, "LineGroupLinksFusioHandler: Line ignored in " << csv.filename <<
+        " missing line_group_id or line_id column");
+    }
+    auto line_group = gtfs_data.line_group_map.find(row[line_group_id_c]);
+    if (line_group == gtfs_data.line_group_map.end()) {
+        LOG4CPLUS_WARN(logger, "LineGroupLinksFusioHandler: Impossible to find the line_group " << row[line_group_id_c]);
+        return;
+    }
+
+    auto line = gtfs_data.line_map.find(row[line_id_c]);
+    if (line == gtfs_data.line_map.end()) {
+        LOG4CPLUS_WARN(logger, "LineGroupLinksFusioHandler: Impossible to find the line " << row[line_id_c]);
+        return;
+    }
+
+    if(line_group->second->linkedLinesURI.find(line->second->uri) != line_group->second->linkedLinesURI.end()) {
+        // Don't insert duplicates
+        return;
+    }
+
+    ed::types::LineGroupLink line_group_link;
+    line_group_link.line_group = line_group->second;
+    line_group_link.line = line->second;
+    // To know if the line is the main line of the group, we compare URI
+    line_group_link.is_main_line = (line_group->second->main_line->uri == line->second->uri);
+    line_group->second->linkedLinesURI.insert(line->second->uri);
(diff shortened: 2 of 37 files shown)


buildbotctp:

eturck:
It's failing because of navitia-proto.
I made the PR : https://github.com/CanalTP/navitia-proto/pull/1

source/ed/connectors/fusio_parser.cpp
```diff
@@ -657,6 +657,76 @@ void LineFusioHandler::handle_line(Data& data, const csv_row& row, bool is_first
...
+        return;
+    }
+
+    auto line = gtfs_data.line_map.find(row[line_id_c]);
+    if (line == gtfs_data.line_map.end()) {
+        LOG4CPLUS_WARN(logger, "LineGroupLinksFusioHandler: Impossible to find the line " << row[line_id_c]);
+        return;
+    }
+
+    if(line_group->second->linkedLinesURI.find(line->second->uri) != line_group->second->linkedLinesURI.end()) {
```
antoine-de: not sure, but it seems that the linkedLinesURI are only used to make sure no duplicates are added.

For this I prefer to add a std::map<line, std::set<>> in the class GtfsData. It's clearer that way, what is only need during data import and what is part on the object model.

antoine-de:
Nice PR lot's of added test :+1: !

Sorry, I haven't the time to finish the reading, I'll continue on monday.

Thanks!

eturck:
Thank you! It was indeed there only to prevent duplicate entries for line_group_links.

I have a generic question on the pull request workflow. Since my original commits 2 days ago, a few commits have been made on the dev branch, causing conflicts with my code.
What should I do : 
- Rebase + push force on my branch to keep the code up to date ?
- Merge dev into my branch (history will not be super clean I guess if I do that) ?
- Wait until the end of the PR to do one those ?
- Something else I didn't think of / don't know about :) ?

Have a nice day / week end !

antoine-de:
yes you should rebase + force push on your fork, this way the git history will be cleaner and since the force push is only on you repository, it's not possible to break the central repository.

You can either to this now or at the end of the PR, it's up to you, but I think it's easier not to wait too long not to have too many conflicts.

stifoon:
i would like to interact with the json interface of a "http://coverage/v1/line_groups" request.
The "lines" property does not contain lines, but a couple [line, "is_main_line"]. So it is not consistent with the other services.
Perharps a json like this one should be more consistent:

```
{
  "line_groups": [
    {
      "id": "line_group:bob",
      "name": "Direction Bob",
      "lines": [
        {
          "code": "B",
          "name": "Blue Bob",
          "color": "Blue",
          "geojson": "",
          "id": "line:bob:blue"
        },
        {
          "code": "R",
          "name": "Red Bob",
          "color": "Red",
          "geojson": "",
          "id": "line:bob:red"
        }
      ],
      "main_line": {
        "type": "line",
        "id": "line:bob:blue"
      }
    },
    {
      "id": "line_group:bobette",
      "name": "Direction Bobette",
      "lines": [
        {
          "code": "Y",
          "name": "Yellow Bobette",
          "color": "AAFFBB",
          "geojson": "",
          "id": "line:bobette:yellow"
        },
        {
          "code": "G",
          "name": "Green Bobette",
          "color": "AAFFBB",
          "geojson": "",
          "id": "line:bobette:green"
        }
      ],
      "main_line": {
        "type": "line",
        "id": "line:bobette:yellow"
      }
    }
  ]
}
```

eturck:
Thank you for the feedback.
It seems to be a better way to display line_groups, I'm going to make the necessary changes.

When I display groups in a "v1/lines" request, with a depth of 1, I'm going to change the current output to : 

``` json
lines: [
   {
       "id": "line:1",
        "groups": [
            {
               "id": "group:1",
               "name": "1st group",
               "main_line": {
                  "id": "line:1"
               }
            }
        ]
   }
]
```

Is this ok ? Or would you prefer to have the "main_line" only if we have a depth > 1 ?

eturck:
I've updated the code here and in the navitia-proto PR.

Here is the new output for line_groups : 
line_groups

And for lines (I decided to not display the main_line here, it seemed better that way and the depth parameter is here if we want it) : 
line_with_group
With depth=2 groups are displayed with its lines and main_line.

I'm working on providing you a link to test the json output live, but as you may know we had a few problem building on debian 7.8 the past few days... I have fallen a bit behind on that.

fixtures/ed/ntfs/line_groups.txt
```diff
@@ -0,0 +1,4 @@
+line_group_id,line_group_name,main_line_id
+lg1,"Groupe lignes B",l2
+lg2,"Groupe ligne Bs seule",l3
+lg3,"Groupe inconnu",iddeligneprincipalequinexistepas
```
TeXitoi: Maybe in english?

source/ed/connectors/fusio_parser.cpp
```diff
@@ -657,6 +657,77 @@ void LineFusioHandler::handle_line(Data& data, const csv_row& row, bool is_first
...
+void LineGroupFusioHandler::init(Data&) {
+    id_c = csv.get_pos_col("line_group_id");
+    name_c = csv.get_pos_col("line_group_name");
+    main_line_id_c = csv.get_pos_col("main_line_id");
+}
+
+void LineGroupFusioHandler::handle_line(Data& data, const csv_row& row, bool) {
+    if(!(has_col(id_c, row) && has_col(name_c, row) && has_col(main_line_id_c, row))) {
+        LOG4CPLUS_WARN(logger, "LineGroupFusioHandler: Line ignored in " << csv.filename <<
+            " missing line_group_id, line_group_name or main_line_id column");
```
TeXitoi: you want to return here, no?

antoine-de:
your new json seems nice!

maybe you should call the line groups of the line 'line_groups' and not 'groups' (we try to name each collection with the type of the object).

And in the line_groups api we could:
- display only id and name for depth=0 as you say
- display the lines and main lines on depth=1
- force the main_line depth to be 0 not to duplicate too many information ?

would that be ok with you @stifoon  ?

source/ed/connectors/gtfs_parser.h
```diff
@@ -106,6 +107,9 @@ struct GtfsData {
     //for gtfs we group the comments together, so the key is the comment and the value is the id of the comment
     std::unordered_map<std::string, std::string> comments_id_map;
 
+    // Store lines linked for each group to avoid duplicates
+    std::unordered_map<std::string, std::vector<std::string>> linked_lines_by_line_group_uri;
```
antoine-de: since it's to avoid duplicate, isn't a set better than the vector ?

eturck: It's only to store uris of lines of each group. I don't really mind duplicate inside the vector, I just use "find" before creating a link between the group and the line.
And since I read http://lafstern.org/matt/col1.pdf I try to not use them too much =P.

source/ed/connectors/fusio_parser.cpp
eturck: Indeed. Thanks !

eturck:
@antoine-de on the last part, on depth=1, lines and main_line are displayed, but already with a depth of 0 (in pb_converter.cpp I use depth - 1).
If we want to reduce the output size I can maybe have for the main_line : 
- If depth=0, no output,
- If depth=1, just output id / name (it should be possible since all the field in the proto file are optional I guess ?),
- If depth > 1 use fill_pb_object with depth - 1.

antoine-de:
IMHO, we should display the main line for depth > 1, but not with depth=(depth -1), with depth=0, since all the information is redundant.

source/ed/connectors/fusio_parser.cpp
```diff
@@ -657,6 +657,77 @@ void LineFusioHandler::handle_line(Data& data, const csv_row& row, bool is_first
...
+
+void LineGroupLinksFusioHandler::init(Data&) {
+    line_group_id_c = csv.get_pos_col("line_group_id");
+    line_id_c = csv.get_pos_col("line_id");
+}
+
+void LineGroupLinksFusioHandler::handle_line(Data& data, const csv_row& row, bool) {
+    if(!(has_col(line_group_id_c, row) && has_col(line_id_c, row))) {
+        LOG4CPLUS_WARN(logger, "LineGroupLinksFusioHandler: Line ignored in " << csv.filename <<
+        " missing line_group_id or line_id column");
```
TeXitoi: same here

source/ed/connectors/fusio_parser.cpp
```diff
@@ -657,6 +657,77 @@ void LineFusioHandler::handle_line(Data& data, const csv_row& row, bool is_first
...
+    }
+
+    auto line = gtfs_data.line_map.find(row[line_id_c]);
+    if (line == gtfs_data.line_map.end()) {
+        LOG4CPLUS_WARN(logger, "LineGroupLinksFusioHandler: Impossible to find the line " << row[line_id_c]);
+        return;
+    }
+
+    std::vector<std::string> linked_lines = gtfs_data.linked_lines_by_line_group_uri[line_group->second->uri];
+    if(std::find(linked_lines.begin(), linked_lines.end(), line->second->uri) != linked_lines.end()) {
```
TeXitoi: Why linked line is not a set?

antoine-de:
Jenkins, test this please

source/ed/ed_persistor.cpp
```diff
@@ -726,6 +730,35 @@ void EdPersistor::insert_lines(const std::vector<types::Line*>& lines){
...
+    for(const auto& line_group: groups) {
+        this->lotus.insert({
+            std::to_string(line_group->idx),
+            navitia::encode_uri(line_group->uri),
+            line_group->name
+        });
+    }
+    this->lotus.finish_bulk_insert();
+
+    this->lotus.prepare_bulk_insert("navitia.line_group_link", {"group_id","line_id","is_main_line"});
```
TeXitoi: Why is there a `is_main_line` in this relation table, and not an id in the `line_group` table with the id of the main line, as it is in the NTFS? It seems cleaner and simpler to me. (Same for `types::LineGroupLinks`)

source/ed/connectors/gtfs_parser.h
TeXitoi: In this case, you should use a `boost::flat_set`, a `std::set` interface with a sorted vector behind.

http://www.boost.org/doc/libs/1_58_0/doc/html/container/non_standard_containers.html#container.non_standard_containers.flat_xxx

source/ed/ed_reader.cpp
```diff
@@ -550,6 +551,35 @@ void EdReader::fill_lines(nt::Data& data, pqxx::work& work){
...
 
+void EdReader::fill_line_groups(nt::Data& data, pqxx::work& work){
+    std::string request = "SELECT id, uri, name FROM navitia.line_group";
+
+    pqxx::result result = work.exec(request);
+    for(auto const_it = result.begin(); const_it != result.end(); ++const_it){
+        nt::LineGroup* line_group = new nt::LineGroup();
+        const_it["uri"].to(line_group->uri);
+        const_it["name"].to(line_group->name);
+        this->line_group_map[const_it["id"].as<idx_t>()] = line_group;
```
TeXitoi, line 563: you can also use .to() here no?

source/ed/types.cpp
```diff
@@ -114,6 +114,10 @@ bool Line::operator<(const Line& other) const {
     }
 }
 
+bool LineGroup::operator<(const LineGroup& other) const {
+    return this->name < other.name || (this->name == other.name && this < &other);
+}
```
TeXitoi, line 122: I don't like comparing pointers. As LineGroup is a Header, is has an idx on which we can compare instead of the address (which is not the same accross different run).

I prefere writing my lexical comparator more closely to the corresponding order:

``` c++
bool LineGroup::operator<(const LineGroup& other) const {
    if (this->name != other.name) { return this->name < other.name; }
    return this->idx < other.idx;
}
```

eturck:
I think I have made all the changes : 
- Line_group table has a main_line_id column,
- Line_group_link doesn't have a is_main_line column anymore,
- Is_main_line is removed everywhere in the code since we don't use it anymore (neither in the database nor the json output),
- Using a flat_set instead of a vector for linked_lines_by_line_group_uri,
- Rename a few "groups" to "line_groups" (especially in the lines API),
- Force a depth of 0 for the main_line in line_groups API to reduce duplicated informations,
- Add 2 throw in NTFS file parsing when the header is not correct,
- Fix comparator to not compare pointer and compare idx instead,
- Translate line_groups.txt in ntfs fixture.

antoine-de:
Jenkins, test this please

source/ed/types.h
```diff
@@ -224,6 +224,19 @@ struct Line : public Header, Nameable {
 
 };
 
+struct LineGroup : public Header, Nameable {
+    const static nt::Type_e type = nt::Type_e::LineGroup;
+    std::string name;
```
antoine-de, line 229: you already have a name with the Nameable struct.

I'm not sure you want to inherit from this one, it's up to you.

source/jormungandr/tests/check_utils.py
```diff
@@ -598,6 +598,14 @@ def is_valid_line(line, depth_check=1):
     g = line.get('geojson')
     g is None or shape(g) #TODO check length
 
+def is_valid_line_group(line_group, depth_check=1):
+    get_not_null(line_group, "name")
+    get_not_null(line_group, "id")
+
+    if depth_check > 0:
+        is_valid_line(get_not_null(line_group, "main_line"), depth_check - 1)
```
antoine-de: the main_line is with depth_check = 0 no ?

source/jormungandr/tests/ptref_tests.py
```diff
@@ -119,6 +119,28 @@ def test_line(self):
         assert physical_modes[0]['id'] == 'physical_mode:Car'
         assert physical_modes[0]['name'] == 'name physical_mode:Car'
 
+        group = get_not_null(l, 'line_groups')
```
antoine-de: shouldn't you call is_valid_line_group there too?

source/jormungandr/tests/ptref_tests.py
```diff
@@ -119,6 +119,28 @@ def test_line(self):
...
+        assert len(line_groups) == 1
+
+        lg = line_groups[0]
+
+        is_valid_line_group(lg, depth_check=1)
+
+        com = get_not_null(lg, 'comments')
+        assert len(com) == 1
+        assert com[0]['type'] == 'standard'
+        assert com[0]['value'] == "I'm a happy comment"
```
antoine-de: it's really great that you added some jormungandr tests! Since we decided to do some non conventional stuffs with the depth I think it would be great to add some more tests:
- one test per depth there just to be sure (after the call, a call to is_valid_line_group should be enough)
- some tests to check that the ptref stuff work well with the line_group: 
  - `self.query_region("v1/line_groups/{line group id}/lines")`
  - and maybe `self.query_region("v1/routes/{one route id}/line_groups")` to check if you can get all line_groups for a route

source/type/pb_converter.cpp
```diff
@@ -409,6 +413,30 @@ void fill_pb_object(nt::Line const* l, const nt::Data& data,
...
 }
 
+void fill_pb_object(const nt::LineGroup* lg, const nt::Data& data,
+        pbnavitia::LineGroup* line_group, int max_depth,
+        const pt::ptime& now, const pt::time_period& action_period, const bool show_codes){
+    if(lg == nullptr)
+        return ;
+    int depth = (max_depth <= 3) ? max_depth : 3;
+
+    for (const auto& comment: data.pt_data->comments.get(lg)) {
```
antoine-de: comments for depth = 0 ? not sure it's really useful but you know better than me :smile:

source/type/type.h
```diff
@@ -616,12 +620,15 @@ struct Line : public Header, Nameable, HasMessages, Codes{
 
     std::map<std::string,std::string> properties;
 
+    //The bool part of the pair stand for "is_main_line"
```
antoine-de: it seems there is no bool anymore :wink:

eturck:
New update!
- Add some tests recommanded by @antoine-de,
- Add a case in type.cpp in order to have line_groups accessible through the graph (and move a data->complete in main_ptref_test.cpp),
- Remove old / useless code / comment

antoine-de:
Nice, thanks @eturck !

PR #1079, opened 2015-07-01
Jormungandr: add fordidden_uris[] for coherence

antoine-de:
in journeys we use fordidden_uris[] and in ptref fordidden_id[]

for coherence purpose fordidden_uris[] has been added, and
fordidden_id[] kept for retrocompatibility purpose

### DIFF
--- a/source/jormungandr/jormungandr/interfaces/v1/Uri.py
+++ b/source/jormungandr/jormungandr/interfaces/v1/Uri.py
@@ -70,8 +70,14 @@ class Uri(ResourceUri):
                             default=1,
                             description="The depth of your object")
         parser.add_argument("forbidden_id[]", type=unicode,
+                            description="DEPRECATED, replaced by forbidden_uris[]",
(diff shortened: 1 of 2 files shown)


kinnou02:
Jenkins, test this please

source/jormungandr/jormungandr/interfaces/v1/Uri.py
```diff
@@ -100,6 +106,10 @@ def get(self, region=None, lon=None, lat=None, uri=None, id=None):
 
         args = self.parsers["get"].parse_args()
 
+        # for retrocompatibility purpose
+        for forbid_id in args.get('forbidden_ids[]', []):
```
kinnou02, line 110: changing the `dest` won't work?

antoine-de, line 110: nop seems that only one param can write in a `dest`, so the old param was not working

PR #1081, opened 2015-07-01
Jormungandr: tag TAD StopDateTime on /departures

azime:
- fill properties in StopDateTime object in departures API
  Jira:
  http://jira.canaltp.fr/browse/NAVITIAII-1775

### DIFF
--- a/source/jormungandr/tests/check_utils.py
+++ b/source/jormungandr/tests/check_utils.py
@@ -827,5 +827,7 @@ def get_all_disruptions(elem, response):
 
 def is_valid_stop_date_time(stop_date_time):
     get_not_null(stop_date_time, 'arrival_date_time')
+    assert get_valid_datetime(stop_date_time['arrival_date_time'])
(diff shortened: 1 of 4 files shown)


source/jormungandr/tests/check_utils.py
```diff
@@ -824,3 +824,8 @@ def disruptions_filler(_, obj):
 
     return disruption_by_obj
 
+
+def is_valid_stop_date_time(stop_date_time):
+    get_not_null(stop_date_time, 'arrival_date_time')
```
antoine-de, line 829: those should be datetime no (or maybe just time  sometimes) ?

antoine-de:
nice tests :+1: 

Can you rename you PR according to the doc ? 
maybe like: 
Jormungandr: tag TAD StopDateTime on /departures

(or a better title if you want)

source/jormungandr/tests/check_utils.py
azime, line 829: OK

PR #1085, opened 2015-07-03
Kraken: fix accessible path

antoine-de:
the accessible tag was taken into account only for the stops and not the
vehicles.

fix http://jira.canaltp.fr/browse/NAVITIAII-1790

### DIFF
--- a/source/ed/build_helper.h
+++ b/source/ed/build_helper.h
@@ -83,7 +83,7 @@ struct SA {
 };
 
 
-struct builder{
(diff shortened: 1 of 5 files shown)


source/ed/build_helper.h
```diff
@@ -83,7 +83,7 @@ struct SA {
 };
 
 
-struct builder{
+struct builder {
```
TeXitoi, line 86: :+1:

PR #1086, opened 2015-07-03
raptor: direct path filtering

TeXitoi:
We want to use the direct path in the reader filtering, to avoid giving pt solution that is worst that the direct path solution.

fix http://jira.canaltp.fr/browse/NAVP-189

On fr-pdl with the geopal data for benchmark_full:
- before: 36.7s user for 1.411Go of RAM
- after: 33.6s user for 1.436Go of RAM


antoine-de:
could you add some figures about the performances evolution (ram / response time) ?

source/routing/raptor.cpp
```diff
@@ -619,11 +635,17 @@ void RAPTOR::boucleRAPTOR(const type::AccessibiliteParams& accessibilite_params,
...
+std::vector<Path> RAPTOR::compute(const type::StopArea *departure,
+                                  const type::StopArea *destination,
+                                  int departure_hour,
+                                  int departure_day,
+                                  DateTime borne,
+                                  bool disruption_active,
+                                  bool clockwise,
+                                  const type::AccessibiliteParams &accessibilite_params,
+                                  uint32_t max_transfers,
+                                  const std::vector<std::string> &forbidden_uri,
```
antoine-de: :sob:

source/routing/raptor_api.cpp
```diff
@@ -876,6 +888,10 @@ make_response(RAPTOR &raptor, const type::EntryPoint& origin,
 
 
     DateTime bound = clockwise ? DateTimeUtils::inf : DateTimeUtils::min;
+    typedef boost::optional<navitia::time_duration> OptTimeDur;
+    const OptTimeDur direct_path_dur = direct_path.path_items.empty() ?
+        OptTimeDur() :
+        OptTimeDur(direct_path.duration / origin.streetnetwork_params.speed_factor);
```
antoine-de, line 906: speed_factor there ? it isn't taken into account in the direct path computation ?

source/routing/tests/raptor_test.cpp
```diff
@@ -1984,3 +1984,77 @@ BOOST_AUTO_TEST_CASE(accessible_on_first_sp) {
...
+                                  10,
+                                  {},
+                                  {},
+                                  true,
+                                  15_s); // 15s direct path
+    BOOST_CHECK_EQUAL(res.size(), 0);
+
+    res = raptor.compute_all(departures,
+                             arrivals,
+                             DateTimeUtils::set(2, "08:03"_t),
```
antoine-de: why do you change the dt ?

kinnou02:
This will replace `_delete_non_optimal_journey` in jormungandr, so you can delete it

source/routing/raptor_api.cpp
TeXitoi, line 906: No it isn't.

source/routing/tests/raptor_test.cpp
TeXitoi: Bad copy and past from the reverse request I think.

TeXitoi:
`_delete_non_optimal_journey` removed.

source/routing/tests/raptor_test.cpp
TeXitoi: Corrected, github is just trying to be too cleaver.

TeXitoi:
Jenkins, test this please

PR #1088, opened 2015-07-06
Raptor api: explicit fall back when used for admin

pbougue:
Fall back for admin is now explicit when used.

We used to hide fall-back when starting from the center of an admin or a stop point out of the main stop area, which was not easy to understand for users.
For example a journey was impossible when walking (center too far from any stop point), and possible when biking but not mentioning any bike section in that case.

Partially processing http://jira.canaltp.fr/browse/NAVITIAII-1784

### DIFF
--- a/source/jormungandr/tests/ptref_tests.py
+++ b/source/jormungandr/tests/ptref_tests.py
@@ -301,7 +301,8 @@ class TestPtRefPlace(AbstractTestFixture):
 
         #the default is the search for all stops within 200m, so we should have A and C
         eq_(len(stops), 2)
-        assert ["stopA", "stopC"] == [s['name'] for s in stops]
+
+        assert set(["stopA", "stopC"]) == set([s['name'] for s in stops])
 
     def test_with_coord_distance_different(self):
         """same as test_with_coord, but with 300m radius. so we find all stops"""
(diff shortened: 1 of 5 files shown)


buildbotctp:

source/jormungandr/tests/ptref_tests.py
```diff
@@ -301,7 +301,8 @@ def test_with_coord(self):
 
         #the default is the search for all stops within 200m, so we should have A and C
         eq_(len(stops), 2)
-        assert ["stopA", "stopC"] == [s['name'] for s in stops]
+
+        assert set(["stopA", "stopC"]) == set([s['name'] for s in stops])
```
antoine-de, line 305: this is correct, but JFYI, I think you can also do {"stopA", "stopC"}

source/routing/tests/routing_api_test.cpp
```diff
@@ -1077,6 +1077,34 @@ BOOST_FIXTURE_TEST_CASE(biking_with_different_speed, streetnetworkmode_fixture<t
...
+// we should output a fall back section, as it is used for computation
+BOOST_FIXTURE_TEST_CASE(admin_explicit_fall_back, streetnetworkmode_fixture<test_speed_provider>) {
+
+    origin.uri = "admin:74435";
+    origin.type = b.data->get_type_of_id(origin.uri);
+    origin.streetnetwork_params.max_duration = navitia::seconds(100);
+    destination.streetnetwork_params.max_duration = navitia::seconds(100);
+
+    auto resp = make_response();
+    dump_response(resp, "admin_explicit_fall_back", true);
```
antoine-de: set it back to false no ?

PR #1090, opened 2015-07-07
streetnetwork: fix a bug when the origin and the destination where projected on the same edge

kinnou02:
It was possible to not count the cost of the edge and only count the
projection.

Example:

```
       * O
A-----------------------------------------B
                                     *D
```

We go from O to D: each point is projected on the two vertexes of the
edge. The calculation is done fine but when we reconstruct the path since we used only one
vertex we aren't able to reconstruct anything, so we were keeping only
the two smallest projection for each point.

With this commit, if we have a reverse path of only one element we used
the initialized values of the dijkstra for finding the right duration.


kinnou02:
Test will probably follow

PR #1096, opened 2015-07-09
streetnetwork: fix a bug when the origin and the destination where projected on the same edge

kinnou02:
It was possible to not count the cost of the edge and only count the
projection.

Example:

```
       * O
A-----------------------------------------B
                                     *D
```

We go from O to D: each point is projected on the two vertexes of the
edge. The calculation is done fine but when we reconstruct the path since we used only one
vertex we aren't able to reconstruct anything, so we were keeping only
the two smallest projection for each point.

With this commit, we calculate the distance to the edge, then the
distance to the projected point, and finally the distance from the
projected destination to the real destination.

### DIFF
--- a/source/georef/street_network.cpp
+++ b/source/georef/street_network.cpp
@@ -49,6 +49,18 @@ navitia::time_duration PathFinder::crow_fly_duration(const double distance) cons
     return navitia::seconds(distance / (default_speed[mode_] * speed_factor));
 }
 
+static bool is_projected_on_same_edge(const ProjectionData& p1, const ProjectionData& p2){
+    return (p1[source_e] == p2[source_e] && p1[target_e] == p2[target_e])
+                || (p1[source_e] == p2[target_e] && p1[target_e] == p2[source_e]);
+}
+
+navitia::time_duration PathFinder::path_duration_on_same_edge(const ProjectionData& p1, const ProjectionData& p2){
+    return crow_fly_duration(p1.real_coord.distance_to(p1.projected)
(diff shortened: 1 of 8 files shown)


source/georef/street_network.cpp
```diff
@@ -427,12 +469,12 @@ Path PathFinder::get_path(const ProjectionData& target, std::pair<navitia::time_
 void PathFinder::add_projections_to_path(Path& p, bool append_to_begin) const {
     //we need to find out which side of the projection has been used to compute the right length
 
-    //we check if the we already are the the arrival
+    //we check if the we already are the arrival
```
antoine-de: there is still grammar problems with the sentence :wink:

pbougue: there is still grammar problems with the sentence :wink:

TeXitoi:
OK for me. Waiting artemis check.

source/georef/street_network.h
```diff
@@ -225,7 +228,9 @@ struct StreetNetwork {
 };
 
 /// Build a path from a reverse path list
-Path create_path(const GeoRef& georef, std::vector<vertex_t> reverse_path, bool add_one_elt);
+Path create_path(const GeoRef& georef,
+                 std::vector<vertex_t> reverse_path,
```
xlqian: very minor note: Do we really want to pass a copy of a `vector<vertex_t>` to the function? maybe, i missed something...

TeXitoi: I think you're right: a const ref is much better here.

PR #1098, opened 2015-07-10
raptor: penalize transfer in walking time and desactivate same sa in journey

TeXitoi:


source/routing/raptor_solution_reader.cpp
```diff
@@ -162,7 +162,7 @@ bool is_valid(const RaptorSolutionReader<Visitor>& reader, const Journey& j) {
         }
     }
     // We don't want to pass to a sa several times during the journey.
-    if (pass_several_times_same_sa(reader, j)) { return false; }
+    //if (pass_several_times_same_sa(reader, j)) { return false; }
```
xlqian: useless commented code??

PR #1101, opened 2015-07-13
raptor: fix a bug when we can walk somewhere but go faster to it using pt

TeXitoi:
fix http://jira.canaltp.fr/browse/NAVITIAII-1776

### DIFF
--- a/source/routing/raptor.cpp
+++ b/source/routing/raptor.cpp
@@ -248,9 +248,11 @@ static void init_best_pts_snd_pass(const RAPTOR::vec_stop_point_duration& depart
                                    IdxMap<type::StopPoint, DateTime>& best_labels) {
     for (const auto& d: departures) {
         if (clockwise) {
-            best_labels[d.first] = departure_datetime + d.second.total_seconds() - 1;
+            best_labels[d.first] = std::min(best_labels[d.first],
+                                            departure_datetime + d.second.total_seconds() - 1);
         } else {
-            best_labels[d.first] = departure_datetime - d.second.total_seconds() + 1;
+            best_labels[d.first] = std::max(best_labels[d.first],
+                                            departure_datetime - d.second.total_seconds() + 1);
         }
     }
 }
@@ -315,8 +317,13 @@ RAPTOR::compute_all(const vec_stop_point_duration& departures_,
         best_labels_pts = best_labels_pts_for_snd_pass;
         best_labels_transfers = best_labels_transfers_for_snd_pass;
         boucleRAPTOR(accessibilite_params, !clockwise, disruption_active, max_transfers);
-        const auto reader_results = read_solutions(*this, !clockwise, departures_, destinations,
-                                                   disruption_active, accessibilite_params);
+        const auto reader_results = read_solutions(*this,
+                                                   !clockwise,
+                                                   departure_datetime,
+                                                   departures_,
+                                                   destinations,
+                                                   disruption_active,
+                                                   accessibilite_params);
         for (const auto& s: reader_results) { solutions.add(s); }
     }
 
--- a/source/routing/raptor_solution_reader.cpp
+++ b/source/routing/raptor_solution_reader.cpp
@@ -88,7 +88,14 @@ get_out_st_dt(const std::pair<const type::StopTime*, DateTime>& in_st_dt,
     return {nullptr, 0};
 }
 
-bool is_valid(const Journey& j) {
+template<typename Visitor>
+bool is_valid(const RaptorSolutionReader<Visitor>& reader, const Journey& j) {
+    // We don't want journeys that begins before (resp. after) the requested datetime
+    if (reader.v.clockwise()) {
+        if (j.departure_dt < reader.departure_datetime) { return false; }
+    } else {
+        if (j.arrival_dt > reader.departure_datetime) { return false; }
+    }
     // We don't want journeys with a transfert between 2 estimated stop times.
     for (auto it = j.sections.begin(), it_prev = it++; it != j.sections.end(); it_prev = it++) {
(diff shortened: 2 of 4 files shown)


source/routing/raptor_solution_reader.cpp
```diff
@@ -88,7 +88,14 @@ get_out_st_dt(const std::pair<const type::StopTime*, DateTime>& in_st_dt,
     return {nullptr, 0};
 }
 
-bool is_valid(const Journey& j) {
+template<typename Visitor>
+bool is_valid(const RaptorSolutionReader<Visitor>& reader, const Journey& j) {
+    // We don't want journeys that begins before (resp. after) the requested datetime
+    if (reader.v.clockwise()) {
+        if (j.departure_dt < reader.departure_datetime) { return false; }
```
xlqian: you can `return j.departure_dt >= reader.departure_datetime `.  same thing for L97

source/routing/raptor.cpp
```diff
@@ -248,9 +248,11 @@ static void init_best_pts_snd_pass(const RAPTOR::vec_stop_point_duration& depart
                                    IdxMap<type::StopPoint, DateTime>& best_labels) {
     for (const auto& d: departures) {
         if (clockwise) {
-            best_labels[d.first] = departure_datetime + d.second.total_seconds() - 1;
+            best_labels[d.first] = std::min(best_labels[d.first],
```
xlqian, line 251: are we sure `d.first` is always a valid key of `best_labels`?

source/routing/raptor_solution_reader.cpp
```diff
@@ -88,7 +88,14 @@ get_out_st_dt(const std::pair<const type::StopTime*, DateTime>& in_st_dt,
     return {nullptr, 0};
 }
 
-bool is_valid(const Journey& j) {
+template<typename Visitor>
+bool is_valid(const RaptorSolutionReader<Visitor>& reader, const Journey& j) {
```
xlqian, line 92: In this function, the only thing you need is `reader`'s `departure_datetime`. For me, it would be more readable (and avoid the template) if we pass only `reader.departure_datetime` than pass the entire `reader`. On the other hand, if we want to unit-test this function, we won't have to construct a complete `reader` for just testing purpose. :)

TeXitoi, line 92: no, we also need `reader.v` which is of the templated type.

source/routing/raptor.cpp
TeXitoi, line 251: yes, and that's almost checked by the type system! See utils/idx_map.h for more information.

source/routing/raptor_solution_reader.cpp
pbougue: You need reader also for clockwise.

xlqian, line 92: oops, sorry, my bad. And by reading other functions accepting `reader` as argument, i'm ok with that

source/routing/raptor.cpp
xlqian, line 251: ok!

source/routing/raptor_solution_reader.cpp
```diff
@@ -560,16 +571,17 @@ std::ostream& operator<<(std::ostream& os, const Journey& j) {
...
                          const bool disruption_active,
                          const type::AccessibiliteParams& accessibilite_params)
 {
     if (clockwise) {
-        return read_solutions(raptor, raptor_reverse_visitor(), deps, arrs,
+        return read_solutions(raptor, raptor_reverse_visitor(), departure_datetime, deps, arrs,
                               disruption_active, accessibilite_params);
     } else {
-        return read_solutions(raptor, raptor_visitor(), deps, arrs,
+        return read_solutions(raptor, raptor_visitor(), departure_datetime, deps, arrs,
```
pbougue, line 584: One call to `read_solution`, maybe with a ternary operator? :smiling_imp:

source/routing/tests/raptor_test.cpp
```diff
@@ -2149,3 +2149,45 @@ BOOST_AUTO_TEST_CASE(no_going_backward) {
     BOOST_CHECK_EQUAL(res[0].items[0].departure, time_from_string("2015-01-01 08:10:00"));
     BOOST_CHECK_EQUAL(res[0].items[0].arrival, time_from_string("2015-01-01 08:20:00"));
 }
+
+// 1: A--->B
+// 2:      B---->C
+// We are at A, and we can walk to B, but can't take 2.
```
pbougue: Comment not very clear to me : 
1 and 2 are VJ, and if we walk to B we can't take VJ 2?

pbougue:
Naïve question (maybe one more test) : what about end-to-end walk versus quicker PT? Was it affected by this bug also?

source/routing/tests/raptor_test.cpp
TeXitoi: Yeah, you can't because you'll be too late. If you have any idea to rewrite it, propose, please :)

TeXitoi:
No, or only in reverse clockwise.

source/routing/raptor_solution_reader.cpp
TeXitoi, line 584: You can't, `raptor_reverse_visitor()` and `raptor_visitor()` don't have the same type.

source/routing/tests/raptor_test.cpp
pbougue: Proposition :
// VJ 1: A--->B
// VJ 2:      B---->C
// Journey from A to C. We can walk to B, but then we arrive to late for VJ2 and no solution is possible.
// So we want A--vj1-->B--vj2-->C

pbougue:
Ok for me.

xlqian:
Ok for me

PR #1105, opened 2015-07-15
Tyr: fix bug on file type

antoine-de:
we'll need to hotfix with this and with #1104


kinnou02:
kill me.

antoine-de:
:sad:

wait a bit to merge it, I need to test a bit more

PR #1108, opened 2015-07-16
Jormun: use pypy as interpreter and some improvements

xlqian:
- remove keys() in jormungandr
- fix some unit tests to prepare the use of pypy
- some other improvements

Notes about using pypy:
- install pip for pypy
- install modules with `pypy -m pip install`
- pyzmq >= 3.0 and libzmq >= 3.0
- remember that if a lib calls c/c++ code in the back, it's not compatible with pypy. That's the reason why we cannot activate protobuf c++ bindings for pypy and why we replaced `ujson` by `simplejson`. 
- We can probably activate protobuf c++ interface by using `cppyy` and `reflex` but we need more investigations to figure out how that works and to test if we'll have an important gain(probably the answer is Yes). [see this link](http://docs.cython.org/src/userguide/pypy.html)
- some implementations in pypy are very different from python's. Like `datetime` module is completely reimplemented in pypy in pure python, so error messages are different in some cases. `dict` type is implemented in pypy in the way of `OrderedDict`. That's the reason why output of pypy is arranged in different order than the output of python. The `exception` class is also implemented in a different way. So remember to write also portable syntax.
-  `for loop` is not necessarily slower than `map` in pypy. so is `list comprehension`.

### DIFF
--- a/source/jormungandr/jormungandr/instance_manager.py
+++ b/source/jormungandr/jormungandr/instance_manager.py
@@ -214,7 +214,7 @@ class InstanceManager(object):
         return instances
 
     def region_exists(self, region_str):
-        if region_str in self.instances:
+        if region_str in self.instances.keys():
             return True
         else:
             raise RegionNotFound(region=region_str)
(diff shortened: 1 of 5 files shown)


buildbotctp:

antoine-de:
Ok to test

antoine-de:
Jenkins, test this please

antoine-de:
We need to bench it and thoroughly check for non regressions, so we postpone the merge for the moment

antoine-de:
Jenkins, test this please

buildbotctp:

xlqian:
Jenkins, test this please

PR #1111, opened 2015-07-17
Refactor jormungandr

xlqian:
- remove all `keys()`
- minor refactor

### DIFF
--- a/source/jormungandr/jormungandr/interfaces/v1/ResourceUri.py
+++ b/source/jormungandr/jormungandr/interfaces/v1/ResourceUri.py
@@ -181,14 +181,14 @@ class complete_links(object):
             if isinstance(elem, (list, tuple)):
                 queue.extend(elem)
             elif hasattr(elem, 'iterkeys'):
-                if elem.get('type', None) == collect_type:
+                if elem.get('type') == collect_type:
                     if collect_type == "notes":
                         result.append({"id": elem['id'], "value": elem['value'], "type": collect_type})
                     elif collect_type == "exceptions":
-                        type_ = "Add" if not elem['except_type'] else "Remove"
-                        result.append({"id": elem['id'], "date": elem['date'], "type" : type_})
+                        type_ = "Add" if elem['except_type'] == 0 else "Remove"
+                        result.append({"id": elem['id'], "date": elem['date'], "type": type_})
 
-                    [elem.pop(del_) for del_ in del_types]
+                    map(elem.pop, del_types)
                 else:
                     queue.extend(elem.itervalues())
         return result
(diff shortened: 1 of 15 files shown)


source/jormungandr/jormungandr/interfaces/v1/ResourceUri.py
```diff
@@ -172,25 +172,23 @@ def __init__(self, resource):
...
+        queue.extend(data.itervalues())
         collect_type = collect["type"]
+        del_types = collect["del"]
         while queue:
             elem = queue.pop()
             if isinstance(elem, (list, tuple)):
                 queue.extend(elem)
             elif hasattr(elem, 'iterkeys'):
-                if 'type' in elem and elem['type'] == collect_type:
+                if elem.get('type', None) == collect_type:
```
antoine-de: None is optional since it's the default value

source/jormungandr/jormungandr/interfaces/v1/ResourceUri.py
```diff
@@ -172,25 +172,23 @@ def __init__(self, resource):
...
                     if collect_type == "notes":
                         result.append({"id": elem['id'], "value": elem['value'], "type": collect_type})
                     elif collect_type == "exceptions":
-                        type_= "Remove"
-                        if elem['except_type'] == 0:
-                            type_ = "Add"
+                        type_ = "Add" if not elem['except_type'] else "Remove"
```
antoine-de: note sure that 

`not elem['except_type']` is the same than `elem['except_type'] == 0` no ?

source/jormungandr/jormungandr/interfaces/v1/ResourceUri.py
xlqian: yep

source/jormungandr/jormungandr/interfaces/v1/ResourceUri.py
xlqian: you're right.

source/jormungandr/jormungandr/interfaces/v1/ResourceUri.py
```diff
@@ -172,25 +172,23 @@ def __init__(self, resource):
...
                         result.append({"id": elem['id'], "date": elem['date'], "type" : type_})
-                    for del_ in collect["del"]:
-                        del elem[del_]
+
+                    [elem.pop(del_) for del_ in del_types]
```
xlqian: I'd have used `map`. After a little perf test with `python2.7`,  list comprehension(`[]`) is very little tiny faster than the 'for loop', but the `map` is 3 times faster...

source/jormungandr/jormungandr/scenarios/simple.py
```diff
@@ -166,7 +166,7 @@ def __stop_times(self, request, instance, departure_filter,
             st.interface_version = 0
         else:
             st.interface_version = request["interface_version"]
-        st.count = 10 if not "count" in request.keys() else request["count"]
+        st.count = 10 if "count" not in request else request["count"]
```
antoine-de: isn't it easier to rewrite it as `request.get('count', 10)` ?

PR #1117, opened 2015-07-20
Raptor: bug on main_stop_areas cost

pbougue:
Following http://jira.canaltp.fr/browse/NAVP-226
Some stop_areas had a cost while being main_stop_areas of given departure admin
To merge after (update artemis test) : CanalTP/artemis#179

### DIFF
--- a/source/georef/tests/georef_test.cpp
+++ b/source/georef/tests/georef_test.cpp
@@ -404,7 +404,7 @@ BOOST_AUTO_TEST_CASE(compute_nearest){
     navitia::routing::map_stop_point_duration tested_map;
     float_t walk_speed = default_speed[Mode_e::Walking] * 2;
     tested_map[navitia::routing::SpIdx(*sp1)] = navitia::seconds(60 / walk_speed);
-    BOOST_CHECK_EQUAL_COLLECTIONS(res.begin(), res.end(), tested_map.begin(), tested_map.end());
+    BOOST_CHECK_EQUAL_COLLECTIONS(res.cbegin(), res.cend(), tested_map.cbegin(), tested_map.cend());
 
     w.init(starting_point);
     res = w.find_nearest_stop_points(1000_s, pl, false);
@@ -412,7 +412,7 @@ BOOST_AUTO_TEST_CASE(compute_nearest){
     tested_map.clear();
     tested_map[navitia::routing::SpIdx(*sp1)] = navitia::seconds(60 / walk_speed);
     tested_map[navitia::routing::SpIdx(*sp2)] = navitia::seconds(50 / walk_speed) + 150_s;
-    BOOST_CHECK_EQUAL_COLLECTIONS(res.begin(), res.end(), tested_map.begin(), tested_map.end());
+    BOOST_CHECK_EQUAL_COLLECTIONS(res.cbegin(), res.cend(), tested_map.cbegin(), tested_map.cend());
 }
 
 // Récupérer les cordonnées d'un numéro impair :
--- a/source/georef/tests/georef_test.cpp
+++ b/source/georef/tests/georef_test.cpp
@@ -1059,9 +1059,6 @@ BOOST_AUTO_TEST_CASE(find_nearest_on_same_edge){
 
     w.init(starting_point);//not mandatory, but reinit to clean the distance table to get fresh dijsktra
     res = w.find_nearest_stop_points(180_s, pl, false);
-    for (auto& elt: res) {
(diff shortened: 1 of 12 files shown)


source/georef/tests/georef_test.cpp
```diff
@@ -398,19 +398,21 @@ BOOST_AUTO_TEST_CASE(compute_nearest){
 
     w.init(starting_point);//not mandatory, but reinit to clean the distance table to get fresh dijsktra
     res = w.find_nearest_stop_points(100_s, pl, false);
-    BOOST_REQUIRE_EQUAL(res.size(), 1);
-    BOOST_CHECK_EQUAL(res[0].first , 0);
-    BOOST_CHECK_EQUAL(res[0].second, navitia::seconds(60 / (default_speed[Mode_e::Walking] * 2))); //the projection is done with the same mean of transport, at the same speed
+    //the projection is done with the same mean of transport, at the same speed
+    navitia::routing::map_stop_point_duration tested_map;
+    tested_map.insert(navitia::routing::stop_point_duration(navitia::routing::SpIdx(0),
```
antoine-de: I think it's clearer if you build the SpIdx directly with the stop point

`navitia::routing::SpIdx(sp1)`

source/routing/raptor_utils.h
```diff
@@ -48,6 +49,9 @@ typedef Idx<type::JourneyPatternPoint> JppIdx;
 typedef Idx<type::JourneyPattern> JpIdx;
 typedef Idx<type::StopPoint> SpIdx;
 
+typedef boost::container::flat_map<SpIdx, navitia::time_duration> map_stop_point_duration;
+typedef map_stop_point_duration::value_type stop_point_duration;
```
xlqian: FYI: http://en.cppreference.com/w/cpp/language/type_alias 
PS: there are absolutely no differences between both.

source/routing/tests/routing_api_test.cpp
```diff
@@ -799,10 +799,11 @@ BOOST_FIXTURE_TEST_CASE(direct_path_under_the_limit, direct_path_routing_api_dat
...
 
-    BOOST_REQUIRE_EQUAL(sp.size(), 1);
-    auto res_sp = sp[0];
-    BOOST_CHECK_EQUAL(res_sp.second, to_duration(B.distance_to(S), navitia::type::Mode_e::Walking));
-    BOOST_CHECK_EQUAL(b.data->pt_data->stop_points[res_sp.first]->name, "stop_point:stopB");
+    navitia::routing::map_stop_point_duration tested_map;
+    tested_map.insert(navitia::routing::stop_point_duration(
+                        navitia::routing::SpIdx(*(b.data->pt_data->stop_points_map["stop_point:stopB"])),
+                        to_duration(B.distance_to(S), navitia::type::Mode_e::Walking)));
+    BOOST_CHECK_EQUAL_COLLECTIONS(sp.begin(), sp.end(), tested_map.begin(), tested_map.end());
```
xlqian: if possible: `sp.cbegin()/.cend()` .....

http://stackoverflow.com/questions/12001410/what-is-the-reason-behind-cbegin-cend

and using `std::begin()` is even better:

http://stackoverflow.com/questions/7593086/why-use-non-member-begin-and-end-functions-in-c11

source/routing/raptor_api.cpp
```diff
@@ -748,7 +747,7 @@ get_stop_points( const type::EntryPoint &ep, const type::Data& data,
                     const SpIdx sp_idx = SpIdx(*stop_point);
                     if (stop_points.find(sp_idx) == stop_points.end()) {
                         concerned_path_finder.distance_to_entry_point[sp_idx] = {};
-                        result.push_back({sp_idx, {}});
+                        result.insert(routing::stop_point_duration(sp_idx, {}));
```
antoine-de: I think you can remove `stop_points` no ?

source/routing/raptor_api.cpp
```diff
@@ -806,20 +805,22 @@ get_stop_points( const type::EntryPoint &ep, const type::Data& data,
         }
         const auto admin = data.geo_ref->admins[it_admin->second];
 
+        if (! admin->main_stop_areas.empty()) {
+            for (auto stop_area: admin->main_stop_areas) {
+                for(auto sp : stop_area->stop_point_list) {
+                    const SpIdx sp_idx = SpIdx(*sp);
```
xlqian: minor advice: `const SpIdx sp_idx{*sp};`   `{}` eliminates the most vexing parser and it tells us that `SpIdx` isn't a function.

PR #1121, opened 2015-07-22
Raptor: bug on cost of main_stop_area

pbougue:
as we use [], revert insert order
correct bug on #1117


pbougue:
After @patochectp check on CanalTP/artemis#181, and after the merge of this PR, it will be possible (and recommended) to merge on artemis.

PR #1127, opened 2015-07-29
kraken: switch back to local time stoptime for /vehicle_journeys

kinnou02:


source/type/pb_converter.cpp
```diff
@@ -648,7 +648,14 @@ void fill_pb_object(const nt::VehicleJourney* vj,
                            vehicle_journey->mutable_journey_pattern(), depth-1,
                            now, action_period, show_codes);
         }
-        for(const auto& stop_time : vj->stop_time_list) {
+        for(type::StopTime stop_time : vj->stop_time_list) {
```
antoine-de, line 651: I think it would be good to add a comment that the copy is wanted here no ?

PR #1128, opened 2015-07-29
update integration.rst

bbonnesoeur:
fix description for id and name attribute for stop area and stop point object


antoine-de:
thanks @bbonnesoeur !

PR #1130, opened 2015-08-03
Eitri: fix integration tests

antoine-de:
the helmi03 repository was broken, so we forked it.

fix the CMakeList to work with virtual env

add some requirements

### DIFF
--- a/source/eitri/requirements.txt
+++ b/source/eitri/requirements.txt
@@ -4,3 +4,4 @@ psycopg2==2.4.5
 clingon==0.1.4
 GeoAlchemy2==0.2.4
 SQLAlchemy==1.0.4
+alembic==0.6.7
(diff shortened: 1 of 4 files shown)


source/eitri/requirements.txt
```diff
@@ -2,3 +2,5 @@ docker-py==0.4.0
 retrying==1.2.3
 psycopg2==2.4.5
 clingon==0.1.4
+GeoAlchemy2==0.2.4
+SQLAlchemy==1.0.4
```
pbougue, line 6: Is the need for alembic covered by SQLAlchemy?
Nota: I only needed alembic, no SQLAlchemy

antoine-de, line 6: I think SqlAlchemy is also needed (even if the dependency might be carried by GeoAlchemy).

You're right for alembic, it has been added

PR #1131, opened 2015-08-03
All: Kraken headsign structure + fill in Ed

pbougue:
Headsign structure, aiming minimum memory print
- unit test
- test after bina (using docker test)

Tested on artemis/guichet_unique data, and working for headsign changes.

Memoy bench:
- fr-sncfnational: 520Mo versus 512Mo before PR (without stop_time headsign, we store vj's headsign only)
- artemis/guichet_unique: 564Mo vs 549Mo before PR (with stop_headsign)

### DIFF
--- a/source/ed/docker_tests/ed_integration_tests.cpp
+++ b/source/ed/docker_tests/ed_integration_tests.cpp
@@ -66,6 +66,23 @@ struct ArgsFixture {
    std::map<std::string, std::string> input_file_paths;
 };
 
+// check headsign value for stop times on given vjs
+static void check_headsigns(const navitia::type::Data& data, const std::string& headsign,
+                     size_t first_it_vj, size_t last_it_vj,
+                     size_t first_it_st = 0, size_t last_it_st = std::numeric_limits<size_t>::max()) {
+    const navitia::type::HeadsignHandler& headsigns = data.pt_data->headsign_handler;
+    auto& vj_vec = data.pt_data->vehicle_journeys;
+    for (size_t it_vj = first_it_vj; it_vj <= last_it_vj; ++it_vj) {
+        size_t real_last_it_st = vj_vec[it_vj]->stop_time_list.size() - 1;
+        if (real_last_it_st > last_it_st) {
+            real_last_it_st = last_it_st;
+        }
+        for (size_t it_st = first_it_st; it_st <= real_last_it_st; ++it_st) {
+            BOOST_CHECK_EQUAL(headsigns.get_headsign(vj_vec[it_vj]->stop_time_list[it_st]), headsign);
+        }
+    }
+}
+
 BOOST_FIXTURE_TEST_CASE(fusio_test, ArgsFixture) {
     const auto input_file = input_file_paths.at("ntfs_file");
     navitia::type::Data data;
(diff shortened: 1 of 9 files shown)


source/ed/docker_tests/ed_integration_tests.cpp
```diff
@@ -78,5 +78,55 @@ BOOST_FIXTURE_TEST_CASE(fusio_test, ArgsFixture) {
...
+    BOOST_CHECK_EQUAL(headsigns.get_headsign(vj_vec[4]->stop_time_list[2]), "vehiclejourney3");
+    BOOST_CHECK_EQUAL(headsigns.get_headsign(vj_vec[4]->stop_time_list[3]), "vehiclejourney3");
+    BOOST_CHECK_EQUAL(headsigns.get_headsign(vj_vec[4]->stop_time_list[4]), "vehiclejourney3");
+    BOOST_CHECK_EQUAL(headsigns.get_headsign(vj_vec[4]->stop_time_list[5]), "vehiclejourney3");
+    BOOST_CHECK_EQUAL(headsigns.get_headsign(vj_vec[5]->stop_time_list[0]), "vehiclejourney3");
+    BOOST_CHECK_EQUAL(headsigns.get_headsign(vj_vec[5]->stop_time_list[1]), "vehiclejourney3");
+    BOOST_CHECK_EQUAL(headsigns.get_headsign(vj_vec[5]->stop_time_list[2]), "vehiclejourney3");
+    BOOST_CHECK_EQUAL(headsigns.get_headsign(vj_vec[5]->stop_time_list[3]), "vehiclejourney3");
+    BOOST_CHECK_EQUAL(headsigns.get_headsign(vj_vec[5]->stop_time_list[4]), "vehiclejourney3");
+    BOOST_CHECK_EQUAL(headsigns.get_headsign(vj_vec[5]->stop_time_list[5]), "vehiclejourney3");
```
antoine-de: I think you can factorize a bit there :wink:

source/type/headsign_handler.cpp
```diff
@@ -0,0 +1,169 @@
...
+    vj.name = new_name;
+    headsign_mvj[vj.name].insert(vj.meta_vj);
+    update_headsign_mvj_after_remove(vj, prev_name);
+}
+
+void HeadsignHandler::update_headsign_mvj_after_remove(const VehicleJourney& vj,
+                                                       const std::string& removed_headsign) {
+    // unregister meta-vj from headsign only if necessary
+    bool meta_has_headsign = false;
+    for (const auto& it_headsign_vj: get_vj_from_headsign(removed_headsign)) {
```
antoine-de: a boost::find would be better

antoine-de: and early return might be better

source/type/headsign_handler.cpp
```diff
@@ -0,0 +1,169 @@
...
+    return (--it_headsign)->second;
+}
+
+bool HeadsignHandler::has_headsign_or_name(const VehicleJourney& vj,
+                                           const std::string& headsign) const {
+    if (vj.name == headsign) {
+        return true;
+    }
+
+    if(!navitia::contains(map_vj_map_stop_time_headsign_change, &vj)) {
```
antoine-de: it might be better to store the iterator there

source/type/headsign_handler.cpp
```diff
@@ -0,0 +1,169 @@
...
+        return false;
+    }
+
+    for (const auto& change: map_vj_map_stop_time_headsign_change.at(&vj)) {
```
antoine-de: a boost::find_if might be clearer there

source/type/headsign_handler.cpp
```diff
@@ -0,0 +1,169 @@
...
+        if (change.second == headsign) {
+            return true;
+        }
+    }
+    return false;
+}
+
+std::vector<const VehicleJourney*>
+HeadsignHandler::get_vj_from_headsign(const std::string& headsign) const {
+    using namespace std;
```
antoine-de: if you need only vector, I think it's clearer to only import it

source/type/headsign_handler.cpp
```diff
@@ -0,0 +1,169 @@
...
+    vector<const VehicleJourney*> res;
+    const auto& it_vj_set = headsign_mvj.find(headsign);
+    if (it_vj_set == headsign_mvj.end()) {
+        return res;
+    }
+
+    for (const MetaVehicleJourney* mvj: it_vj_set->second) {
+        vector<vector<VehicleJourney*>> vect_vect_vj{mvj->theoric_vj, mvj->adapted_vj, mvj->real_time_vj};
+        for (const auto vect_vj: vect_vect_vj) {
```
antoine-de: const auto&

source/type/headsign_handler.cpp
```diff
@@ -0,0 +1,169 @@
...
+HeadsignHandler::get_vj_from_headsign(const std::string& headsign) const {
+    using namespace std;
+    vector<const VehicleJourney*> res;
+    const auto& it_vj_set = headsign_mvj.find(headsign);
+    if (it_vj_set == headsign_mvj.end()) {
+        return res;
+    }
+
+    for (const MetaVehicleJourney* mvj: it_vj_set->second) {
+        vector<vector<VehicleJourney*>> vect_vect_vj{mvj->theoric_vj, mvj->adapted_vj, mvj->real_time_vj};
```
antoine-de: const

source/type/headsign_handler.cpp
```diff
@@ -0,0 +1,169 @@
...
+            }
+        }
+    }
+    return res;
+}
+
+void HeadsignHandler::affect_headsign_to_stop_time(const StopTime& stop_time,
+                                                   const std::string& headsign) {
+    const VehicleJourney* vj = stop_time.vehicle_journey;
+    headsign_mvj[vj->name].insert(vj->meta_vj); // for safety
```
antoine-de: I think it's better to add an assert here

source/type/headsign_handler.cpp
```diff
@@ -0,0 +1,169 @@
...
+    std::string prev_headsign_for_stop_time = get_headsign(stop_time);
+    if (headsign == prev_headsign_for_stop_time) {
+        return;
+    }
+
+    uint16_t order = stop_time.journey_pattern_point->order;
+    // erase change if exists
+    if (navitia::contains(map_vj_map_stop_time_headsign_change[vj], order)) {
```
antoine-de: I find this name a bit too long. 

Not sure but would `heasign_changes` be better ?

source/type/headsign_handler.cpp
```diff
@@ -0,0 +1,169 @@
...
+                                                       const std::string& removed_headsign) {
+    // unregister meta-vj from headsign only if necessary
+    bool meta_has_headsign = false;
+    for (const auto& it_headsign_vj: get_vj_from_headsign(removed_headsign)) {
+        if (it_headsign_vj->meta_vj == vj.meta_vj) {
+            meta_has_headsign = true;
+            break;
+        }
+    }
+    if (!meta_has_headsign && navitia::contains(headsign_mvj, removed_headsign)) {
```
antoine-de: not sure that the contains is relevant here since you want to use (twice) the indexed element

source/type/headsign_handler.cpp
```diff
@@ -0,0 +1,169 @@
...
+#include "headsign_handler.h"
+#include "utils/functions.h"
+
+namespace navitia { namespace type {
+
+void HeadsignHandler::change_name_and_register_as_headsign(VehicleJourney& vj,
+                                                           const std::string& new_name) {
+    std::string prev_name = vj.name;
+    vj.name = new_name;
+    headsign_mvj[vj.name].insert(vj.meta_vj);
```
xlqian, line 40: are you sure `vj.name` is always a valid key of `headsign_mvj`? Pay attention when using `[]` operator directly :)

source/type/headsign_handler.cpp
```diff
@@ -0,0 +1,169 @@
...
+std::vector<const VehicleJourney*>
+HeadsignHandler::get_vj_from_headsign(const std::string& headsign) const {
+    using namespace std;
+    vector<const VehicleJourney*> res;
+    const auto& it_vj_set = headsign_mvj.find(headsign);
+    if (it_vj_set == headsign_mvj.end()) {
+        return res;
+    }
+
+    for (const MetaVehicleJourney* mvj: it_vj_set->second) {
```
xlqian, line 120: optional `const auto*`

source/type/headsign_handler.h
```diff
@@ -0,0 +1,78 @@
...
+    }
+
+protected:
+    bool has_headsign_or_name(const VehicleJourney& vj, const std::string& headsign) const;
+    void update_headsign_mvj_after_remove(const VehicleJourney& vj,
+                                          const std::string& removed_headsign);
+
+    // for each VJ, map containing index of stop time and the new headsign (until next change)
+    // as new stop_time might be added in the future, if map_vj_map_stop_time_headsign_change[vj]
+    // exists, last change is always vj.name (and might happen after last stop_time)
```
antoine-de, line 71: I think an ascii art might explain that easily no ?

source/type/headsign_handler.cpp
xlqian: i think you can define a variable for `map_vj_map_stop_time_headsign_change[vj]`to avoid redundant research

source/type/headsign_handler.cpp
antoine-de, line 40: it's not a problem, it will create the set if needed no ?

xlqian, line 40: ok

source/type/tests/headsign_test.cpp
```diff
@@ -0,0 +1,153 @@
...
+}
+
+struct HeadsignHandlerTest: nt::HeadsignHandler {
+    std::unordered_map<const nt::VehicleJourney*, boost::container::flat_map<uint16_t, std::string>>
+    get_map_vj_map_stop_time_headsign_change() {return map_vj_map_stop_time_headsign_change;}
+    std::unordered_map<std::string, std::unordered_set<const nt::MetaVehicleJourney*>>
+    get_headsign_mvj() {return headsign_mvj;}
+};
+
+BOOST_AUTO_TEST_CASE(headsign_handler_internal_test) {
```
antoine-de: I think it would have been easier to use a fixture there, this way you can directly use the protected attribute

source/type/headsign_handler.h
pbougue, line 71: Yep, did it :smile:

source/type/tests/headsign_test.cpp
pbougue: Made a fixture as it is more compact, but not sure of what you meant by "use the protected attribute" :grey_question:

source/type/headsign_handler.h
antoine-de, line 71: yep I think it's clearer this way :+1:

source/type/tests/headsign_test.cpp
antoine-de: it's not relevant anymore (since you change the heasign_handler outside of the test), but a boost_test inherit from the fixture, so you can use the protected members (so your fixture could have inherited from the handler)

pbougue:
Jenkins, test this please

source/type/headsign_handler.cpp
```diff
@@ -0,0 +1,171 @@
...
+
+const std::string& HeadsignHandler::get_headsign(const StopTime& stop_time) const{
+    // if no headsign map for vj: return name
+    if (!navitia::contains(headsign_changes, stop_time.vehicle_journey)) {
+        return stop_time.vehicle_journey->name;
+    }
+
+    // otherwise use headsign change map
+    const auto& map_stop_time_headsign_change =
+            headsign_changes.at(stop_time.vehicle_journey);
```
antoine-de, line 71: not important at all, but the `at` is not needed here since it's checked before

edit: ah it's for the const ?

PR #1136, opened 2015-08-10
jormungandr: check the {sn}_speed for correctness

TeXitoi:


source/jormungandr/jormungandr/interfaces/parsers.py
```diff
@@ -36,6 +36,13 @@ def depth_argument(value, name):
     return int(value)
 
 
+def speed_value(value, name):
+    if float(value) <= 0:
```
xlqian: minor: the name may be `positive_float` to be more generic :)

xlqian:
ok for me

source/jormungandr/jormungandr/interfaces/parsers.py
francois-vincent: avoid double conversion:
`value = float(value)`

better:

``` python
message = "The {} argument must be float > 0, you gave : {}"
try:
    value = float(value)
except ValueError:
    raise ValueError(message.format(name, value))
if value <= 0:
    raise ValueError(message.format(name, value))
```

and add 'x' in the unitary test case list

TeXitoi: I can't do that, it'll change the message if the conversion fails.

Issue #1134, opened 2015-08-07
Comments on vj in stop_times links


eturck:
Hello,

In our process of timetable generation we, sometimes, add some comments on a vehicle_journey. It's mainly used to add an information about something we can't describe in the data.

It's working well with route_schedules but when we made some stop_schedules requests we were expecting to have those comments linked to the stop times, it was not the case.

I don't know if it's something you don't use or if you're proceeding differently. Since the vehicle_journey id is in the links, maybe you are doing another request to get the comments?

Since we were a little bit in a rush I have patched the code to have something like this : 
![stop_times](https://cloud.githubusercontent.com/assets/5136553/9137763/35d5525e-3d20-11e5-8c61-63899783d58b.png)

But before doing a PR I want to have your opinion on this.

Patch below (it's really simple) :

``` c++
diff --git a/source/type/pb_converter.cpp b/source/type/pb_converter.cpp
index 33a3dab..37c5692 100644
--- a/source/type/pb_converter.cpp
+++ b/source/type/pb_converter.cpp
@@ -1313,6 +1313,9 @@ void fill_pb_object(const navitia::type::StopTime* stop_time,
         if(!stop_time->vehicle_journey->odt_message.empty()){
             fill_pb_object(stop_time->vehicle_journey->odt_message, data, hn->add_notes(),max_depth,now,action_period);
         }
+        for (const auto& comment: data.pt_data->comments.get(*stop_time->vehicle_journey)) {
+            fill_pb_object(comment, data, hn->add_notes(), max_depth, now, action_period);
+        }
         rs_date_time->mutable_properties()->set_vehicle_journey_id(stop_time->vehicle_journey->uri);
     }
```

TeXitoi:
Why do you need the note on each stop time? on the header is enough, no? Because the stream is already really fat, adding a link on each stop time will be very heavy.

stifoon:
@TeXitoi : I think comments on route_schedule work well. @eturck want to add VJ comments on stop_time only on stop_schedule service.
We just have to be careful that there will not be redundant comments on route_schedule service after modifications on stop_schedule service :)

TeXitoi:
Ok, I had in mind the route schedule data, not the stop schedule one. Forget my question.

pbougue:
Ok, I'll have a look at that and check for possible redundancy.

pbougue:
Closed by #1138
Thank you!

PR #1138, opened 2015-08-13
Type: add comments on vj in stop_times links

xavierraffin:


xavierraffin:
Relative to issue : https://github.com/CanalTP/navitia/issues/1134

pbougue:
Thank for this PR!

We have few remarks to accept it though:
As @TeXitoi said, we don't want the fat stream of route schedules to increase too much, and this does not seem to be what you need (only stop schedules). But this PR affects both route and stop schedule. It would be better to find a way that only affects stop schedules : no notes on stop_times, just on vj for route schedules / notes on stop_times for stop schedules.

Also, if you could add some tests checking first that you have a proper note on stop_time in stop schedule, and second that this note is only on the vj for route schedules, it would be great.
For the test I guess the simpler would be to add a comment on a stop time in main_routing_test for example and then add an integration test in jormungandr test, calling stop schedule and route schedule.

pbougue:
Concerned APIs: 
- https://github.com/CanalTP/navitia/blob/dev/documentation/navitia.io/source/integration.rst#route-schedules-and-time-tables-route-schedules
- https://github.com/CanalTP/navitia/blob/dev/documentation/navitia.io/source/integration.rst#stop-schedules-and-other-kind-of-time-tables-stop-schedules

eturck:
As always thank you for the feedback.

The code is know in departure_board.cpp in order to just affect stop_schedules.
I've also added a small integration test in departure_board_tests.py checking that we have the note on date_times in stop_schedules and only in the header for route_schedules.
I hope the test is in the right place and good enough.

Let me know if I need to change anything else!

source/time_tables/departure_boards.cpp
```diff
@@ -78,8 +78,14 @@ render_v1(const std::map<uint32_t, pbnavitia::ResponseStatus>& response_status,
         //Now we fill the date_times
         for(auto dt_st : id_vec.second) {
             auto date_time = schedule->add_date_times();
+            auto vj = dt_st.second->vehicle_journey;
```
pbougue: I think we should test that stop_time is not null (as it is done in fill_pb_object).

source/time_tables/departure_boards.cpp
```diff
@@ -78,8 +78,14 @@ render_v1(const std::map<uint32_t, pbnavitia::ResponseStatus>& response_status,
...
             fill_pb_object(dt_st.second, data, date_time, 0,
                            current_time, action_period, dt_st.first, calendar_id);
+            if (vj != nullptr) {
+                for (const auto& comment: data.pt_data->comments.get(*vj)) {
+                    fill_pb_object(comment, data, date_time->mutable_properties()->add_notes(), 0, current_time, action_period);
```
pbougue: line length as we try to be under github's limit : 113 characters :smile:

source/time_tables/tests/departure_board_test_data.h
```diff
@@ -26,7 +26,9 @@ struct calendar_fixture {
         b.vj("line:A", "10101", "", true, "weekend")("stop1", 20 * 3600, 20 * 3600 + 10 * 60)("stop2", 21 * 3600, 21 * 3600 + 10 * 60);
 
         // and one everytime
-        b.vj("line:A", "1100101", "", true, "all")("stop1", 15 * 3600, 15 * 3600 + 10 * 60)("stop2", 16 * 3600, 16 * 3600 + 10 * 60);
+        auto& builder_vj = b.vj("line:A", "1100101", "", true, "all")("stop1", 15 * 3600, 15 * 3600 + 10 * 60)("stop2", 16 * 3600, 16 * 3600 + 10 * 60);
```
pbougue: line length too (but was previously :sob: ) you can use `("stop1", "15:00"_t, "15:10"_t)("stop2", "16:00"_t, "16:10"_t);`

source/jormungandr/tests/departure_board_tests.py
```diff
@@ -244,3 +244,38 @@ def test_departures(self):
...
+        assert len(response["notes"]) == 1
+        assert len(response["route_schedules"][0]["table"]["headers"]) == 4
+
+        for header in response["route_schedules"][0]["table"]["headers"]:
+            if len([l for l in header["links"] if l["type"] == "vehicle_journey" and l["id"] == "all"]) == 1:
+                assert len([l for l in header["links"] if l["type"] == "notes" and l["id"] == response["notes"][0]["id"]]) == 1
+
+        for row in response["route_schedules"][0]["table"]["rows"]:
+            for date_time in row["date_times"]:
+                assert len([l for l in date_time["links"] if l["type"] == "notes" and l["id"] == response["notes"][0]["id"]]) == 0
```
pbougue: Why do you test `l["id"] == response["notes"][0]["id"]]` ? Not sure it's necessary.
Plus if we have a `link.note` on another `id`, it's weird, probably another bug, but we're interested in knowing it.

pbougue: On all those tests, I think it would be more readable to declare, name and maybe comment the collections you test.
It would also shorten lines a bit :smile:

source/jormungandr/tests/departure_board_tests.py
```diff
@@ -244,3 +244,38 @@ def test_departures(self):
...
+        response = self.query_region("lines/line:A/route_schedules?from_datetime=20120615T080000")
+        assert 'error' not in response
+        assert 'route_schedules' in response
+        assert len(response["route_schedules"]) == 1
+        assert len(response["notes"]) == 1
+        assert len(response["route_schedules"][0]["table"]["headers"]) == 4
+
+        for header in response["route_schedules"][0]["table"]["headers"]:
+            if len([l for l in header["links"] if l["type"] == "vehicle_journey" and l["id"] == "all"]) == 1:
+                assert len([l for l in header["links"] if l["type"] == "notes" and l["id"] == response["notes"][0]["id"]]) == 1
```
pbougue: Similar as below: not sure it's mandatory to test `id` as you make sure before that there is only one `notes`
It's more secure that way, but messes with comprehension, so it's as you wish on this one (and above too).

pbougue:
First, sorry for the little delay to review it :-/
Otherwise, some formal comments, but if I understand well, that's exactly the kind of tests and code patch we had in mind, so thank you very much!

eturck:
Sorry for the huge delay to take your few comments into account... I wasn't working much on navitia those past couple weeks.
I have removed the check on the id for the note, it really is not necessary. I haven't done too much tests and I never know if I'm doing too many or too few check, so I'm often going a bit overboard :D.
I wasn't aware of the 113 columns rule, but I've shortened a few lines and configured my IDE to take that into account.

I have also rebased everything on the current dev branch.

pbougue:
Ok for me, I merge it now.
Thanks very much for the work!

The 113 columns rule is just to avoid horizontal scroll in unsplit view, and ease PR reading in split view if screen is big enough.

PR #1143, opened 2015-08-21
override profile's params by args in the request

xlqian:
override profile's params by args in the request

http://jira.canaltp.fr/browse/NAVP-229

Ex:

...&traveler_type=standard&walking_speed=42424242

standard's walking_speed now is overrided to 42424242


TeXitoi:
Thanks!

PR #1149, opened 2015-08-27
jormungandr: fix new_default with unicode uris

TeXitoi:


pbougue:
One little test? :)

TeXitoi:
Like that?

PR #1153, opened 2015-08-31
All: headsign output 

pbougue:
Headsigns are printed in departures, arrivals, route_schedules, journeys (display_informations) and vehicle_journeys (stop_time)

No headsign on stop_schedules, as discussed with @stifoon 
JIRA: http://jira.canaltp.fr/browse/NAVP-235

### DIFF
--- a/source/jormungandr/tests/ptref_tests.py
+++ b/source/jormungandr/tests/ptref_tests.py
@@ -495,3 +495,49 @@ class TestPtRef(AbstractTestFixture):
         assert 'error' not in response
         stop_areas = get_not_null(response, 'stop_areas')
         eq_(len(stop_areas), 1)
+
+    def test_headsign_stop_time_vj(self):
+        """test basic print of headsign in stop_times for vj"""
+        response = self.query_region('vehicle_journeys?filter=vehicle_journey.name="vehicle_journey 0"')
+        assert 'error' not in response
+        vjs = get_not_null(response, 'vehicle_journeys')
+        eq_(len(vjs), 1)
+        eq_(len(vjs[0]['stop_times']), 2)
+        eq_(vjs[0]['stop_times'][0]['headsign'], "A00")
+        eq_(vjs[0]['stop_times'][1]['headsign'], "vehicle_journey 0")
+
+    def test_headsign_display_info_journeys(self):
+        """test basic print of headsign in section for journeys"""
+        response = self.query_region('journeys?from=stop_point:stopB&to=stop_point:stopA&datetime=20120615T000000')
+        assert 'error' not in response
+        journeys = get_not_null(response, 'journeys')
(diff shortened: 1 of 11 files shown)


buildbotctp:

source/jormungandr/tests/ptref_tests.py
```diff
@@ -495,3 +495,48 @@ def test_multiple_resource_uri_with_final_collection_uri(self):
...
+        assert 'error' not in response
+        journeys = get_not_null(response, 'journeys')
+        eq_(len(journeys), 1)
+        eq_(len(journeys[0]['sections']), 1)
+        eq_(journeys[0]['sections'][0]['display_informations']['headsign'], "A00")
+
+    def test_headsign_display_info_departures(self):
+        """test basic print of headsign in display informations for departures"""
+        response = self.query_region('stop_points/stop_point:stopB/departures?from_datetime=20120615T000000')
+        assert 'error' not in response
```
antoine-de, line 521: I don't think it's mandatory, should be checked by query_region

source/routing/tests/routing_api_test_data.h
```diff
@@ -405,22 +405,18 @@ struct routing_api_data {
         b.sa("stopB", B.lon(), B.lat());
         if (activate_pt) {
             //we add a very fast bus (2 seconds) to be faster than walking and biking
-            b.vj("A")
-                ("stop_point:stopB", 8*3600 + 1*60, 8*3600 + 1 * 60)
-                ("stop_point:stopA", 8*3600 + 1 * 60 + 2, 8*3600 + 1*60 + 2)
+            b.vj("A")("stop_point:stopB", "08:01"_t)("stop_point:stopA", "08:01:02"_t)
```
antoine-de, line 408: :+1:

source/time_tables/route_schedules.cpp
```diff
@@ -288,7 +288,12 @@ route_schedule(const std::string& filter,
                     pbnavitia::Header* header = table->mutable_headers(j);
                     pbnavitia::PtDisplayInfo* vj_display_information = header->mutable_pt_display_informations();
                     auto vj = dt_stop_time.second->vehicle_journey;
-                    fill_pb_object(vj, d, vj_display_information, 0, now, action_period);
+                    fill_pb_object(vj, d, vj_display_information, dt_stop_time.second, nullptr,
+                                   0, now, action_period);
+                    vj_display_information->set_headsign(vj->name);
```
antoine-de, line 296: I think  a comment might be nice here to explain why it is done only for the API

source/jormungandr/tests/ptref_tests.py
pbougue, line 521: It's not checked, and not possible to add it, as sometimes we return a 200 with an error (no disruptions).

PR #1156, opened 2015-09-03
kraken: don't listen to request before data are loaded

kinnou02:
We also used a named queue for rabbitmq, this way if there is a small
disconnection we don't loose messages. The drawbacks are that we can't
anymore start two kraken for the same instance on the same host. And if
an host/instance is deleted, the queues associated won't be delete from
rabbitmq.

### DIFF
--- a/source/kraken/maintenance_worker.cpp
+++ b/source/kraken/maintenance_worker.cpp
@@ -37,6 +37,7 @@ www.navitia.io
 #include <boost/optional.hpp>
 #include <sys/stat.h>
 #include <signal.h>
+#include "utils/get_hostname.h"
 
 namespace nt = navitia::type;
 namespace pt = boost::posix_time;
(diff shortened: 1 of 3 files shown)


source/kraken/maintenance_worker.cpp
```diff
@@ -60,8 +61,14 @@ void MaintenanceWorker::load(){
 
 void MaintenanceWorker::operator()(){
     LOG4CPLUS_INFO(logger, "Starting background thread");
-    load();
 
+    try{
+        this->listen_rabbitmq();
+    }catch(const std::runtime_error& ex){
+        LOG4CPLUS_ERROR(logger, std::string("Connection to rabbitmq failed: ") << ex.what());
```
TeXitoi: you don't need to create a std::string here

PR #1162, opened 2015-09-04
kraken: DeleteQueue throw an exepton if the queue doesn't exist

kinnou02:


source/kraken/maintenance_worker.cpp
```diff
@@ -162,8 +162,13 @@ void MaintenanceWorker::init_rabbitmq(){
     if(!is_initialized){
         //first we have to delete the queues, binding can change between two run, and it's don't seem possible
         //to unbind a queue if we don't know at what topic it's subscribed
-        channel->DeleteQueue(queue_name_task);
-        channel->DeleteQueue(queue_name_rt);
+        //if the queue doesn't exist an exception is throw...
+        try{
+            channel->DeleteQueue(queue_name_task);
+        }catch(const std::runtime_error& ex){}
```
xlqian, line 168: add ex.what() to a log?

source/kraken/maintenance_worker.cpp
```diff
@@ -162,8 +162,13 @@ void MaintenanceWorker::init_rabbitmq(){
...
+        try{
+            channel->DeleteQueue(queue_name_rt);
+        }catch(const std::runtime_error& ex){}
```
antoine-de, line 171: add ex.what() to a log?

PR #1170, opened 2015-09-11
fix synonyms containing capital and special letters

xlqian:
http://jira.canaltp.fr/browse/NAVITIAII-1799

fix synonyms containing captical and special letters.

Example: 
CPAM <-> sécurité sociale

### DIFF
--- a/source/autocomplete/autocomplete.h
+++ b/source/autocomplete/autocomplete.h
@@ -441,44 +441,18 @@ struct Autocomplete
     }
 
     std::string strip_accents(std::string str) const {
-        std::vector< std::pair<std::string, std::string> > vec_str;
-        vec_str.push_back(std::make_pair("à","a"));
-        vec_str.push_back(std::make_pair("À","a"));
-        vec_str.push_back(std::make_pair("â","a"));
-        vec_str.push_back(std::make_pair("Â","a"));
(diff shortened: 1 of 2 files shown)


xlqian:
Jenkins, test this please

source/autocomplete/autocomplete.h
```diff
@@ -493,18 +471,18 @@ struct Autocomplete
...
         }
 
         //For each synonyms.value found in strFind add synonyms.key
         for(const auto& it : synonyms){
-            if  (boost::regex_search(strFind,boost::regex("\\<" + it.second + "\\>"))){
-                strTemp += " " + it.first;
+            if  (boost::regex_search(strFind,boost::regex("\\<" + strip_accents_and_lower(it.second) + "\\>"))){
+                strTemp += " " + strip_accents_and_lower(it.first);
             }
         }
```
kadhikari, line 488: The use of  function "strip_accents_and_lower" in the tokenizer is timeconsuming. It would be preferable if we can apply the function  for all the synonyms during binarization. Apart this it is fine for me.

xlqian, line 488: good point! I tested it, the performance is in the same order of the old version. :)

PR #1174, opened 2015-09-16
TMA:make disruption on stop point non blocking

xlqian:
make disruption on stop point non blocking, but disruption infos are still raised in API
http://jira.canaltp.fr/browse/NAVITIAII-1799

### DIFF
--- a/source/jormungandr/tests/chaos_disruptions_tests.py
+++ b/source/jormungandr/tests/chaos_disruptions_tests.py
@@ -576,9 +576,9 @@ class TestChaosDisruptionsStopPoint(ChaosDisruptionsFixture):
 @dataset([('main_routing_test', ['--BROKER.rt_topics='+chaos_rt_topic, 'spawn_maintenance_worker'])])
 class TestChaosDisruptionsStopArea(ChaosDisruptionsFixture):
     """
-    Add disruption on stop point:
+    Add disruption on stop area:
         test if the information is raised in response
-        test if stop point is blocked, for the moment the stop point should not be blocking
+        test if stop area is blocked, for the moment the stop area should not be blocking
     Then delete disruption:
         test if there is no more disruptions in response
     """
--- a/source/jormungandr/tests/chaos_disruptions_tests.py
+++ b/source/jormungandr/tests/chaos_disruptions_tests.py
@@ -266,6 +266,8 @@ class TestChaosDisruptionsBlocking(ChaosDisruptionsFixture):
         set_nb_disruptions(response['networks'])
         response = self.query_region('stop_areas')
         set_nb_disruptions(response['stop_areas'])
+        response = self.query_region('stop_points')
+        set_nb_disruptions(response['stop_points'])
         return nb_disruptions_map
 
     def run_check(self, object_id, type_):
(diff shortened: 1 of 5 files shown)


source/kraken/fill_disruption_from_chaos.cpp
```diff
@@ -490,7 +499,7 @@ struct add_impacts_visitor : public apply_impacts_visitor {
     void operator()(const nt::StopArea* stop_area) {
         this->log_start_action(stop_area->uri);
         for (const auto stop_point : stop_area->stop_point_list) {
-            this->operator()(stop_point);
+            handle_stop_point(stop_point);
```
kinnou02: this change seem to be the contrary of what we want

source/kraken/fill_disruption_from_chaos.cpp
```diff
@@ -444,6 +447,12 @@ struct add_impacts_visitor : public apply_impacts_visitor {
     }
 
     void operator()(const nt::StopPoint* stop_point) {
+        LOG4CPLUS_TRACE(log4cplus::Logger::getInstance("log"),
```
antoine-de: maybe in INFO no ?

source/jormungandr/tests/chaos_disruptions_tests.py
```diff
@@ -484,6 +484,41 @@ def test_disruption(self):
         for disruption in disrupt:
             assert disruption['uri'] != 'test_disruption', 'this disruption must have been deleted'
 
+@dataset([('main_routing_test', ['--BROKER.rt_topics='+chaos_rt_topic, 'spawn_maintenance_worker'])])
+class TestChaosDisruptionsStopPoint(ChaosDisruptionsFixture):
+    """
+    Add disruption on stop point, test if the information is raised in response
+    Then delete disruption, test if there is no more disruptions in response
+    """
+    def test_disruption_on_stop_point(self):
```
antoine-de, line 545: I think it would be nice to add a test on /journeys for the blocking stop_point (to check that for the moment it's not blocking)

source/kraken/fill_disruption_from_chaos.cpp
xlqian: hmmm, this function handles stop_point only for the case where the disruption is applied on stop_area(the behaviour of disruption on stop_area stays the same). However when the disruption is applied on stop_point, nothing happened, since `void operator()(const nt::StopPoint)` is empty. Am I misunderstanding something?

source/kraken/fill_disruption_from_chaos.cpp
xlqian: yep

source/chaos-proto
```diff
@@ -1 +1 @@
-Subproject commit 235484fe9fa9fb2fc7446e9d4d4cd74557d4f740
+Subproject commit bfbfaafd6d75ee05669b855cd50b302c122fea4a
```
pbougue: SHA-1 doesn't match with chaos-proto

source/kraken/fill_disruption_from_chaos.cpp
xlqian: ok you're right

source/chaos-proto
xlqian: done

source/jormungandr/tests/chaos_disruptions_tests.py
```diff
@@ -504,6 +533,87 @@ def test_disruption(self):
...
+
+        response = self.query_region(query)
+        disruptions = response.get('disruptions')
+        assert not disruptions
+
+
+@dataset([('main_routing_test', ['--BROKER.rt_topics='+chaos_rt_topic, 'spawn_maintenance_worker'])])
+class TestChaosDisruptionsStopArea(ChaosDisruptionsFixture):
+    """
+    Add disruption on stop point:
```
antoine-de: stop area no ?

antoine-de:
OK for me

source/jormungandr/tests/chaos_disruptions_tests.py
xlqian: yep

PR #1175, opened 2015-09-16
Type: add rt validity pattern

antoine-de:
vj's validity pattern are now a flat_enum_map, so a VJ have 3 VP:
- theoric
- adapted
- realtime

### DIFF
--- a/source/jormungandr/tests/chaos_disruptions_tests.py
+++ b/source/jormungandr/tests/chaos_disruptions_tests.py
@@ -348,6 +348,7 @@ class TestChaosDisruptionsBlocking(ChaosDisruptionsFixture):
 
 
 
+
 @dataset([("main_routing_test", ['--BROKER.rt_topics='+chaos_rt_topic, 'spawn_maintenance_worker'])])
 class TestChaosDisruptionsBlockingOverlapping(ChaosDisruptionsFixture):
     """
--- a/source/jormungandr/tests/chaos_disruptions_tests.py
+++ b/source/jormungandr/tests/chaos_disruptions_tests.py
@@ -346,9 +346,6 @@ class TestChaosDisruptionsBlocking(ChaosDisruptionsFixture):
         self.run_check('stopA', 'stop_area')
 
 
-
-
-
 @dataset([("main_routing_test", ['--BROKER.rt_topics='+chaos_rt_topic, 'spawn_maintenance_worker'])])
 class TestChaosDisruptionsBlockingOverlapping(ChaosDisruptionsFixture):
     """
--- a/source/kraken/fill_disruption_from_chaos.cpp
+++ b/source/kraken/fill_disruption_from_chaos.cpp
@@ -374,7 +374,7 @@ struct functor_add_vj {
         vj->utc_to_local_offset = vj_ref.utc_to_local_offset;
         // The validity_pattern is only active on the period of the impact
         nt::ValidityPattern tmp_vp;
-        nt::ValidityPattern tmp_ref_vp(*vj_ref.adapted_validity_pattern);
+        nt::ValidityPattern tmp_ref_vp(*vj_ref.validity_patterns[nt::RTLevel::Adapted]);
         for (auto period : impact->application_periods) {
             //we can impact a vj with a departure the day before who past midnight
             bg::day_iterator titr(period.begin().date() - bg::days(1));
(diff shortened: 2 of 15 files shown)


source/jormungandr/tests/chaos_disruptions_tests.py
```diff
@@ -348,6 +348,7 @@ def test_disruption_on_journey(self):
 
 
 
+
```
pbougue: useless line add?

source/type/pb_converter.cpp
```diff
@@ -675,15 +675,15 @@ void fill_pb_object(const nt::VehicleJourney* vj,
...
                        depth-1);
-        fill_pb_object(vj->adapted_validity_pattern, data,
+        fill_pb_object(vj->adapted_validity_pattern(), data,
                        vehicle_journey->mutable_adapted_validity_pattern(),
                        depth-1);
-        fill_pb_object(vj->validity_pattern, data, vehicle_journey->mutable_validity_pattern(), max_depth-1);
-        fill_pb_object(vj->adapted_validity_pattern, data, vehicle_journey->mutable_adapted_validity_pattern(), max_depth-1);
-        vptranslator::fill_pb_object(vptranslator::translate(*vj->validity_pattern),
+        fill_pb_object(vj->validity_pattern(), data, vehicle_journey->mutable_validity_pattern(), max_depth-1);
+        fill_pb_object(vj->adapted_validity_pattern(), data, vehicle_journey->mutable_adapted_validity_pattern(), max_depth-1);
```
pbougue: line length

source/type/type.h
```diff
@@ -722,9 +721,15 @@ struct VehicleJourney: public Header, Nameable, hasVehicleProperties, HasMessage
...
+
+    // validity pattern for all RTLevel
+    flat_enum_map<RTLevel, ValidityPattern*> validity_patterns = {{{nullptr, nullptr, nullptr}}};
+
     std::vector<VehicleJourney*> adapted_vehicle_journey_list; //REMOVE
-    VehicleJourney* theoric_vehicle_journey = nullptr; //REMOVE
+    VehicleJourney* theoric_vehicle_journey = nullptr;
+
+    ValidityPattern* adapted_validity_pattern() const { return validity_patterns[RTLevel::Adapted]; }
+    ValidityPattern* validity_pattern() const { return validity_patterns[RTLevel::Theoric]; }
```
pbougue: Not fan of implicit "not saying it's adapted or realtime so it's theoric"... I'd rather use `theoric_validity_pattern`
It might be refactored soon though, but still, it might ease the refactor.

pbougue: Not used in this PR, but how about adding `realtime_validity_pattern()` for future use and consistency?

source/kraken/fill_disruption_from_chaos.cpp
```diff
@@ -374,7 +374,7 @@ struct functor_add_vj {
         vj->utc_to_local_offset = vj_ref.utc_to_local_offset;
         // The validity_pattern is only active on the period of the impact
         nt::ValidityPattern tmp_vp;
-        nt::ValidityPattern tmp_ref_vp(*vj_ref.adapted_validity_pattern);
+        nt::ValidityPattern tmp_ref_vp(*vj_ref.validity_patterns[nt::RTLevel::Adapted]);
```
pbougue: use `adapted_validity_pattern()` instead?

source/routing/tests/next_stop_time_test.cpp
```diff
@@ -2047,7 +2047,7 @@ struct classic_freq_dataset {
     classic_freq_dataset() {
         vp = type::ValidityPattern(date(2012,7,7), "11");
 
-        vj.validity_pattern = &vp;
+        vj.validity_patterns[type::RTLevel::Theoric] = &vp;
```
pbougue, line 2050: same for all `validity_patterns[type::RTLevel::(.*)` -> `$1_validity_pattern()`

antoine-de, line 2050: It's not possible, you can't set with a getter (or rather I won't :wink: ) and I'm not a fan of putting a setter for this.
It's a draw,  @TeXitoi or @xlqian do you have a preference ?

pbougue, line 2050: Ok for not using a getter as setter :)
And sorry I was first thinking only getter.
Would be shorter with a setter in that case.

source/kraken/fill_disruption_from_chaos.cpp
```diff
@@ -561,7 +561,7 @@ struct delete_impacts_visitor : public apply_impacts_visitor {
     // We set all the validity pattern to the theorical one, we will re-apply
     // other disruptions after
     bool func_on_vj(nt::VehicleJourney& vj) {
-        vj.adapted_validity_pattern = vj.validity_pattern;
+        vj.validity_patterns[nt::RTLevel::Adapted] = vj.validity_patterns[nt::RTLevel::Theoric];
```
pbougue, line 564: right one is a getter I think

source/routing/tests/next_stop_time_test.cpp
pbougue, line 2050: The question is more on internalization of RTLevel (in vj and meta-vj), on a bigger picture.

TeXitoi:
Ok for me

PR #1178, opened 2015-09-18
fix Ubuntu 15.04 conditionnal

thiphariel:
Yup, I meant dependency _cough_

### DIFF
--- a/build_navitia.sh
+++ b/build_navitia.sh
@@ -179,7 +179,7 @@ then
     sudo pip install -r "$navitia_dir"/source/jormungandr/requirements.txt
     sudo pip install -r "$navitia_dir"/source/tyr/requirements.txt
     #we want a custom protobuff version
-    sudo pip install -U protobuf==2.5.0
+    sudo pip install -U protobuf==2.6.1
 fi
 
 #the build procedure is explained is the install documentation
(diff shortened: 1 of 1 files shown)


buildbotctp:

TeXitoi:
OK to test

build_navitia.sh
```diff
@@ -179,7 +179,7 @@ then
     sudo pip install -r "$navitia_dir"/source/jormungandr/requirements.txt
     sudo pip install -r "$navitia_dir"/source/tyr/requirements.txt
     #we want a custom protobuff version
-    sudo pip install -U protobuf==2.5.0
+    sudo pip install -U protobuf==2.6.1
```
TeXitoi: This line should just be removed, as it is managed by the requirements (as we want the distrib version)

thiphariel: done :)

build_navitia.sh
```diff
@@ -156,7 +156,7 @@ then
     version=`lsb_release -sr`
 
     # Fix Ubuntu 15.04 package
-    if [ "$distrib" = "Ubuntu" && "$version" = "15.04" ]; then
+    if [ "$distrib" = "Ubuntu" ] && [ "$version" = "15.04" ]; then
```
TeXitoi: You can use `[ test -a test ]` instead of `[ test ] && [ test ]`

thiphariel: Right. Will do that

PR #1179, opened 2015-09-21
install jormungandr database & scheme

thiphariel:
By default, jormungandr database is not install with the build script. I fix that.

The jormungandr database is called "jormungandr"

### DIFF
--- a/build_navitia.sh
+++ b/build_navitia.sh
@@ -199,7 +199,6 @@ echo "** setting up the database"
 kraken_db_name='navitia'
 db_owner='navitia'
 
-kraken_db_name='navitia'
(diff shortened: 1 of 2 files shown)


build_navitia.sh
```diff
@@ -224,6 +223,19 @@ cd "$navitia_dir"/source/sql
 PYTHONPATH=. alembic -x dbname="postgresql://$db_owner:$kraken_db_user_password@localhost/$kraken_db_name" upgrade head
 cd
 
+# Install jormungandr database and upgrade it's sheme
```
xlqian: sheme -> schema :)

buildbotctp:

build_navitia.sh
thiphariel: :°

PR #1180, opened 2015-09-22
Update integration.rst

stifoon:


documentation/navitia.io/source/integration.rst
```diff
@@ -276,6 +276,27 @@ Coverage (/coverage)
...
+
+.. code-block:: json
+
+    {
+        "regions": [
+            "fr-idf"
+        ],
+        address": {
+        ...,
+        "label": "17 Rue Hector Malot (Paris)",
```
pbougue: maybe 20?

documentation/navitia.io/source/integration.rst
```diff
@@ -276,6 +276,27 @@ Coverage (/coverage)
 You can easily navigate through regions covered by navitia.io, with the coverage api.
 The only arguments are the ones of `paging`_.
 
+Inverted geocoding (/coord)
+***************************
+In order to determinate the right coverage, you can request navitia with a WGS84 coordinate as:
```
antoine-de: I think we should clearly speak about both uses:
- Inverted geocoding (with explaination)
- coverages list

documentation/navitia.io/source/integration.rst
```diff
@@ -276,6 +276,65 @@ Coverage (/coverage)
...
+
+.. code-block:: json
+
+    {
+        "regions": [
+            "fr-idf"
+        ],
+        address": {
+        ...,
+        "label": "20 Rue Hector Malot (Paris)",
```
TeXitoi, line 300: I know @pbougue asked for 20, but the result is not coherent with the request...

pbougue, line 300: The goal is to advertise about Kisio Digital!
First, it's just an example
Second, do we ask datas to be coherent with the truth? ;)

PR #1182, opened 2015-09-24
batch processing for maintenance worker

xlqian:
http://jira.canaltp.fr/browse/NAVITIAII-1873

batch processing for maintenance worker

### DIFF
--- a/source/kraken/configuration.cpp
+++ b/source/kraken/configuration.cpp
@@ -60,7 +60,7 @@ po::options_description get_options_description(const boost::optional<std::strin
         ("BROKER.exchange", po::value<std::string>()->default_value("navitia"), "exchange used in rabbitmq")
         ("BROKER.rt_topics", po::value<std::vector<std::string>>(), "list of realtime topic for this instance")
         ("BROKER.timeout", po::value<int>()->default_value(100), "timeout for maintenance worker in millisecond")
-        ("BROKER.sleeptime", po::value<int>()->default_value(5), "sleeptime for maintenance worker in second")
+        ("BROKER.sleeptime", po::value<int>()->default_value(1), "sleeptime for maintenance worker in second")
 
         ("CHAOS.database", po::value<std::string>(), "Chaos database connection string");
 
--- a/source/kraken/configuration.cpp
+++ b/source/kraken/configuration.cpp
@@ -59,6 +59,8 @@ po::options_description get_options_description(const boost::optional<std::strin
         ("BROKER.vhost", po::value<std::string>()->default_value("/"), "vhost for rabbitmq")
         ("BROKER.exchange", po::value<std::string>()->default_value("navitia"), "exchange used in rabbitmq")
         ("BROKER.rt_topics", po::value<std::vector<std::string>>(), "list of realtime topic for this instance")
+        ("BROKER.timeout", po::value<int>()->default_value(100), "timeout for maintenance worker in millisecond")
+        ("BROKER.sleeptime", po::value<int>()->default_value(5), "sleeptime for maintenance worker in second")
 
         ("CHAOS.database", po::value<std::string>(), "Chaos database connection string");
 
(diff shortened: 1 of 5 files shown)


source/kraken/maintenance_worker.cpp
```diff
@@ -82,69 +85,110 @@ void MaintenanceWorker::operator()(){
...
-        LOG4CPLUS_WARN(logger, "Protobuf is not valid!");
-        return;
-    }
-    switch(task.action()){
-        case pbnavitia::RELOAD:
-            load(); break;
-        default:
-            LOG4CPLUS_TRACE(logger, "Task ignored");
+void MaintenanceWorker::handle_task_in_batch(const std::vector<AmqpClient::Envelope::ptr_t>& envelopes){
+    for (auto& envelope: envelopes) {
```
antoine-de, line 89: I don't know if we want to handle those tasks in batch since for the moment we only handle a 'reload' task

source/kraken/maintenance_worker.cpp
```diff
@@ -82,69 +85,110 @@ void MaintenanceWorker::operator()(){
...
+    assert(consume_tag != "");
+    assert(max_nb);
+
+    bool not_empty = true;
+    std::vector<AmqpClient::Envelope::ptr_t> envelopes;
+    envelopes.reserve(max_nb);
+    size_t consumed_nb = 0;
+    while(not_empty && consumed_nb < max_nb) {
+        AmqpClient::Envelope::ptr_t envelope = nullptr;
+        not_empty = channel->BasicConsumeMessage(consume_tag, envelope, timeout_ms);
```
antoine-de: wouldn't it be clearer to break on `! not_empty` (even if the double negation is not nice) ?

source/kraken/maintenance_worker.cpp
xlqian, line 89: I agree with you totally. In this case we can pass a vector that has only one task. Or, I can do a wrapper that takes only one message and delegate to this function?

source/kraken/maintenance_worker.cpp
```diff
@@ -82,69 +85,110 @@ void MaintenanceWorker::operator()(){
...
     data_manager.get_data()->is_connected_to_rabbitmq = true;
     while(true){
-        auto envelope = this->channel->BasicConsumeMessage(std::vector<std::string>({task_tag, rt_tag}));
-        if(envelope->ConsumerTag() == task_tag){
-            handle_task(envelope);
-        }else if(envelope->ConsumerTag() == rt_tag){
-            handle_rt(envelope);
-        }
+        size_t timeout_ms = conf.broker_timeout();
+        size_t max_batch_nb = 100; // Magic Number :)
```
kinnou02: it's an arbitrary value, or there is something that make it magical?

source/kraken/maintenance_worker.cpp
xlqian: ok

source/kraken/maintenance_worker.cpp
xlqian: Indeed, it's more like a arbitrary value not a magic value

source/kraken/configuration.cpp
```diff
@@ -59,6 +59,8 @@ po::options_description get_options_description(const boost::optional<std::strin
         ("BROKER.vhost", po::value<std::string>()->default_value("/"), "vhost for rabbitmq")
         ("BROKER.exchange", po::value<std::string>()->default_value("navitia"), "exchange used in rabbitmq")
         ("BROKER.rt_topics", po::value<std::vector<std::string>>(), "list of realtime topic for this instance")
+        ("BROKER.timeout", po::value<int>()->default_value(100), "timeout for maintenance worker in millisecond")
+        ("BROKER.sleeptime", po::value<int>()->default_value(5), "sleeptime for maintenance worker in second")
```
TeXitoi: 1 second should be enough.

TeXitoi:
OK for me after the comment is fixed.

PR #1189, opened 2015-09-30
kraken: Loading of realtime data from kirin

kinnou02:


source/kraken/maintenance_worker.cpp
```diff
@@ -60,6 +60,41 @@ void MaintenanceWorker::load(){
...
     }
+
+    load_realtime();
+}
+
+
+void MaintenanceWorker::load_realtime(){
+    LOG4CPLUS_INFO(logger, "loading realtime data");
+
+    std::string queue_name = channel->DeclareQueue("", false, false, true, true);
```
xlqian: hmmm, the queue is created, but i don't see it's bound to any exchange with any routing key, is that normal?

source/kraken/configuration.cpp
```diff
@@ -110,6 +112,14 @@ int Configuration::nb_thread() const{
     return this->vm["GENERAL.nb_threads"].as<int>();
 }
 
+bool Configuration::enable_realtime() const{
```
xlqian: function is not called... you may want it later?

source/kraken/maintenance_worker.cpp
xlqian, line 72: ok

source/navitia-proto
```diff
@@ -1 +1 @@
-Subproject commit 8d116b329cf9ce5e173d0e5c046022c4804d9449
+Subproject commit d22f76ceb354fd21e64d53743c95f15154b06dec
```
pbougue: one commit before master, should be good after rebase

source/kraken/configuration.cpp
pbougue: I would name it `[is_|get_|]realtime_enabled`

source/kraken/maintenance_worker.cpp
pbougue, line 72: minor: maybe add a comment on parameters (or name them, but too verbose)

source/kraken/configuration.cpp
```diff
@@ -51,6 +51,8 @@ po::options_description get_options_description(const boost::optional<std::strin
          "name of the instance")
 
         ("GENERAL.nb_threads", po::value<int>()->default_value(1), "number of workers threads")
+        ("GENERAL.enable_realtime", po::value<bool>()->default_value(false), "enable loading of realtime data")
+        ("GENERAL.kirin_timeout", po::value<int>()->default_value(10000), "timeout in ms for loading realtime data from kirin")
```
pbougue: line length

xlqian:
ok  for me

PR #1191, opened 2015-09-30
Ed: Find production date

azime:
Manage production date from different sources

### DIFF
--- a/source/ed/connectors/fusio_parser.h
+++ b/source/ed/connectors/fusio_parser.h
@@ -375,7 +375,7 @@ struct ObjectCodesFusioHandler: public GenericHandler {
  * simply define the list of elemental parsers to use
  */
 struct FusioParser : public GenericGtfsParser {
-    void parse_files(Data&);
+    void parse_files(Data&, const std::string& = "");
     FusioParser(const std::string & path) : GenericGtfsParser(path) {}
 };
 }
--- a/source/ed/connectors/fusio_parser.h
+++ b/source/ed/connectors/fusio_parser.h
@@ -375,7 +375,7 @@ struct ObjectCodesFusioHandler: public GenericHandler {
  * simply define the list of elemental parsers to use
  */
 struct FusioParser : public GenericGtfsParser {
-    void parse_files(Data&, const std::string& = "");
+    void parse_files(Data&, const std::string& beginning_date = "");
     FusioParser(const std::string & path) : GenericGtfsParser(path) {}
 };
 }
--- a/source/ed/connectors/gtfs_parser.cpp
+++ b/source/ed/connectors/gtfs_parser.cpp
@@ -1130,8 +1130,8 @@ GenericGtfsParser::GenericGtfsParser(const std::string & path) : path(path){
 }
 GenericGtfsParser::~GenericGtfsParser() {}
 
-void GenericGtfsParser::fill(Data& data, const std::string beginning_date) {
-
+void GenericGtfsParser::fill(Data& data, const std::string& beginning_date) {
+    /*
     try {
         gtfs_data.production_date = find_production_date(beginning_date);
     } catch (...) {
@@ -1140,8 +1140,8 @@ void GenericGtfsParser::fill(Data& data, const std::string beginning_date) {
             return;
         }
     }
-
-    parse_files(data);
+    */
+    parse_files(data, beginning_date);
 
     normalize_extcodes(data);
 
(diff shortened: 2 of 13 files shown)


source/ed/connectors/gtfs_parser.cpp
```diff
@@ -1351,7 +1342,42 @@ boost::gregorian::date_period GenericGtfsParser::find_production_date(const std:
...
+
+void GenericGtfsParser::manage_production_date(Data& data, const std::string& beginning_date){
+    auto start_it = data.feed_infos.find("feed_start_date");
+    auto end_it = data.feed_infos.find("feed_end_date");
+    boost::gregorian::date start_date(boost::gregorian::not_a_date_time), end_date(boost::gregorian::not_a_date_time);
+    if ((start_it != data.feed_infos.end()) && (end_it != data.feed_infos.end())) {
+        try{
+            start_date = boost::gregorian::from_undelimited_string(start_it->second);
+        }catch(const std::exception& e) {
+           LOG4CPLUS_WARN(logger, "manage_production_date, Unable to parse start_date :"<<start_it->second<<", error : "<<std::string(e.what()));
```
antoine-de: I don't think the string cast is necessary, it should be enough to do `<< e.what()`

you need also to take of the spacing, put some spaces around the `<<`  and `{`

source/ed/connectors/gtfs_parser.cpp
```diff
@@ -1130,18 +1130,9 @@ GenericGtfsParser::GenericGtfsParser(const std::string & path) : path(path){
 }
 GenericGtfsParser::~GenericGtfsParser() {}
 
-void GenericGtfsParser::fill(Data& data, const std::string beginning_date) {
+void GenericGtfsParser::fill(Data& data, const std::string& beginning_date) {
```
TeXitoi, line 1133: :+1:

source/ed/connectors/gtfs_parser.cpp
```diff
@@ -1351,7 +1342,45 @@ boost::gregorian::date_period GenericGtfsParser::find_production_date(const std:
...
+           LOG4CPLUS_WARN(logger, "manage_production_date, Unable to parse start_date :"
+                          << start_it->second << ", error : " << e.what());
+        }
+        try{
+            end_date = boost::gregorian::from_undelimited_string(end_it->second);
+        }catch(const std::exception& e) {
+           LOG4CPLUS_WARN(logger, "manage_production_date, Unable to parse end_date :"
+                          << end_it->second << ", error : " << e.what());
+        }
+        gtfs_data.production_date = complete_production_date(beginning_date, start_date, end_date);
```
TeXitoi: you don't want to do that if there is an exception before. The simplest would be to do only one try with the start, begin and complete.

source/ed/connectors/fusio_parser.h
```diff
@@ -375,7 +375,7 @@ struct ObjectCodesFusioHandler: public GenericHandler {
  * simply define the list of elemental parsers to use
  */
 struct FusioParser : public GenericGtfsParser {
-    void parse_files(Data&);
+    void parse_files(Data&, const std::string& = "");
```
pbougue: You can add string param name for readability

source/ed/connectors/gtfs_parser.h
```diff
@@ -392,7 +392,7 @@ class GenericGtfsParser {
     template <typename Handler>
     void parse(Data&); //some parser do not need a file since they just add default data
 
-    virtual void parse_files(Data&) = 0;
+    virtual void parse_files(Data&, const std::string& = "") = 0;
```
pbougue: You can add string param name for readability

source/ed/connectors/gtfs_parser.h
```diff
@@ -429,7 +433,7 @@ inline void GenericGtfsParser::parse(Data& data) {
  * simply define the list of elemental parsers to use
  */
 struct GtfsParser : public GenericGtfsParser {
-    virtual void parse_files(Data&);
+    virtual void parse_files(Data&, const std::string&);
```
pbougue: You can add string param name for readability

pbougue:
It would be nice to bench one bina with this PR vs with dev to see what we win :)
Not mandatory for merge though, but cool to know.

azime:
corrections

azime:
Test of this optimization on fr-pdl data:
before : 165779ms
After    : 154036 ms

PR #1198, opened 2015-10-07
Doc: remove old and useless french documentation

stifoon:


kinnou02:
Are you sure that all of the images aren't used?
And we have a vpp file, probably with the data model, don't we want to update it?

stifoon:
It's OK, we can delete all :)

PR #1205, opened 2015-10-13
Ed: for the GTFS create one route by line and direction_id

antoine-de:
with the journey pattern refactoring, we missed that the routes were not
associated with the VJ on the GTFS parser, thus gtfs2ed was crashing
during database insertion.

For the GTFS (since the route are explicitly given in the NTFS), we now
create one route by line and direction_id

NEEDS to be hotfixed
jira: http://jira.canaltp.fr/browse/NAVITIAII-1901

### DIFF
--- a/source/ed/connectors/gtfs_parser.cpp
+++ b/source/ed/connectors/gtfs_parser.cpp
@@ -893,6 +893,7 @@ void TripsGtfsHandler::init(Data& data) {
     wheelchair_c = csv.get_pos_col("wheelchair_accessible");
     bikes_c = csv.get_pos_col("bikes_allowed");
     shape_id_c = csv.get_pos_col("shape_id");
+    direction_id_c = csv.get_pos_col("direction_id");
 }
 
 void TripsGtfsHandler::finish(Data& data) {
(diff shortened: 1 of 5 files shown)


source/ed/connectors/gtfs_parser.cpp
```diff
@@ -902,18 +903,42 @@ void TripsGtfsHandler::finish(Data& data) {
...
+        return route;
+    }
+}
+
 void TripsGtfsHandler::handle_line(Data& data, const csv_row& row, bool) {
     auto it = gtfs_data.line_map.find(row[id_c]);
     if (it == gtfs_data.line_map.end()) {
-        LOG4CPLUS_WARN(logger, "Impossible to find the Gtfs route " + row[id_c]
-                       + " referenced by trip " + row[trip_c]);
+        LOG4CPLUS_WARN(logger, "Impossible to find the Gtfs line " << row[id_c]
```
pbougue, line 931: :+1:

xlqian:
ok for me

source/ed/tests/gtfsparser_test.cpp
```diff
@@ -368,6 +368,24 @@ static void check_gtfs_google_example(const ed::Data& data, const ed::connectors
     BOOST_REQUIRE(data.lines[4]->commercial_mode != nullptr);
     BOOST_CHECK_EQUAL(data.lines[4]->commercial_mode->uri, "3");
 
+    // we need to also check the number of routes created (since they are implicit in GTFS)
+    // we create one by line/direction id
+    BOOST_REQUIRE_EQUAL(data.routes.size(), 9);
+    BOOST_CHECK_EQUAL(data.routes[0]->uri, "AB:0");
```
pbougue: Is the order important?
Would it be more secure to check sets?

PR #1206, opened 2015-10-13
osm2ed: add a log for admins with an empty shape

kinnou02:


source/ed/osm2ed.cpp
```diff
@@ -377,11 +378,11 @@ void OSMCache::insert_relations() {
                           std::to_string(relation.level), coord, polygon_str,
                           "admin:"+std::to_string(relation.osm_id)});
         } else {
+            LOG4CPLUS_WARN(logger, "admin " << relation.name << " id: " << relation.osm_id << " of level " << relation.level << " won't be inserted since it has an empty polygon");
```
pbougue: line length

pbougue:
Is there a JIRA connected to that PR ?

kinnou02:
There is not really a linked jira, at least not one than is open

PR #1207, opened 2015-10-13
All: print vj.trip.id

pbougue:
JIRA: http://jira.canaltp.fr/browse/NAVP-319

### DIFF
--- a/source/ed/docker_tests/ed_integration_tests.cpp
+++ b/source/ed/docker_tests/ed_integration_tests.cpp
@@ -31,10 +31,11 @@ www.navitia.io
 #define BOOST_TEST_DYN_LINK
 #define BOOST_TEST_MODULE associated_calendar_test
 #include <boost/test/unit_test.hpp>
+#include <boost/algorithm/string.hpp>
 #include "utils/logger.h"
 #include "type/data.h"
 #include "type/pt_data.h"
-#include <boost/algorithm/string.hpp>
+#include "routing/raptor_utils.h"
 
 struct logger_initialized {
     logger_initialized()   { init_logger(); }
@@ -43,6 +44,7 @@ BOOST_GLOBAL_FIXTURE( logger_initialized )
 
 namespace pt = boost::posix_time;
 namespace nt = navitia::type;
+namespace nr = navitia::routing;
 
 struct ArgsFixture {
    ArgsFixture() {
@@ -124,8 +126,8 @@ BOOST_FIXTURE_TEST_CASE(fusio_test, ArgsFixture) {
     BOOST_CHECK(navitia::contains(headsigns.get_vj_from_headsign("vehiclejourney3"), vj_vec[5]));
 
     BOOST_CHECK_EQUAL(data.pt_data->meta_vjs.size(), 3);
-    for(auto& mvj : data.pt_data->meta_vjs) {
+    for (auto& mvj : data.pt_data->meta_vjs) {
         BOOST_CHECK_EQUAL(data.pt_data->meta_vjs[mvj->uri], mvj.get());
-        BOOST_CHECK_EQUAL(data.pt_data->meta_vjs[navitia::Idx<nt::MetaVehicleJourney>(mvj->idx)], mvj.get());
+        BOOST_CHECK_EQUAL(data.pt_data->meta_vjs[nr::MvjIdx(*mvj)], mvj.get());
     }
 }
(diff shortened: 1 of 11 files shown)


source/navitia-proto
```diff
@@ -1 +1 @@
-Subproject commit 485a92b929d8e6d164592a2f4cadf2fb23c6bb82
+Subproject commit 83c3be775c450322874c509e816ebe4035f70632
```
pbougue: waiting CanalTP/navitia-proto/pull/16 merge for exact SHA-1

source/ed/docker_tests/ed_integration_tests.cpp
```diff
@@ -122,4 +124,8 @@ BOOST_FIXTURE_TEST_CASE(fusio_test, ArgsFixture) {
     BOOST_CHECK(navitia::contains(headsigns.get_vj_from_headsign("vehiclejourney3"), vj_vec[5]));
 
     BOOST_CHECK_EQUAL(data.pt_data->meta_vjs.size(), 3);
+    for(auto& mvj : data.pt_data->meta_vjs) {
+        BOOST_CHECK_EQUAL(data.pt_data->meta_vjs[mvj->uri], mvj.get());
+        BOOST_CHECK_EQUAL(data.pt_data->meta_vjs[navitia::Idx<nt::MetaVehicleJourney>(mvj->idx)], mvj.get());
```
TeXitoi: maybe you can typedef that MvjIdx somewhere as it is done in raptor_utils.h for the Idx.

source/ed/docker_tests/ed_integration_tests.cpp
```diff
@@ -122,4 +124,8 @@ BOOST_FIXTURE_TEST_CASE(fusio_test, ArgsFixture) {
     BOOST_CHECK(navitia::contains(headsigns.get_vj_from_headsign("vehiclejourney3"), vj_vec[5]));
 
     BOOST_CHECK_EQUAL(data.pt_data->meta_vjs.size(), 3);
+    for(auto& mvj : data.pt_data->meta_vjs) {
```
TeXitoi: spacing

source/type/pb_converter.cpp
```diff
@@ -639,6 +639,15 @@ void fill_pb_object(const nt::ValidityPattern* vp, const nt::Data&,
     validity_pattern->set_days(vp->days.to_string());
 }
 
+void fill_pb_object(const nt::MetaVehicleJourney* nav_mvj,
```
TeXitoi, line 642: fill_pb_object takes a const nt::Data& as second argument

source/type/type.h
```diff
@@ -990,7 +990,7 @@ struct MetaVehicleJourney: public Header {
     std::map<std::string, AssociatedCalendar*> associated_calendars;
 
     template<class Archive> void serialize(Archive & ar, const unsigned int ) {
-        ar & base_vj & adapted_vj & real_time_vj & associated_calendars;
+        ar & idx & uri & base_vj & adapted_vj & real_time_vj & associated_calendars;
```
TeXitoi, line 993: Oups ;-)

source/jormungandr/tests/ptref_tests.py
```diff
@@ -579,3 +579,12 @@ def test_headsign_display_info_route_schedules(self):
         display_info = route_schedules[0]['table']['headers'][0]['display_informations']
         eq_(display_info['headsign'], "vjA")
         assert {"A00", "vjA"} == set(display_info['headsigns'])
+
+    def test_trip_id_vj(self):
```
antoine-de, line 583: you need to change the `is_valid_vehicle_journey` method

source/ed/docker_tests/ed_integration_tests.cpp
xlqian: Is `navitia::Idx<nt::MetaVehicleJourney>` a casting?

antoine-de: nop it is a constructor.

BTW you shouldn't provide the idx, you should use to constructor with a MVJ to be typesafe

source/type/pb_converter.cpp
pbougue, line 642: Usually yes, but we don't see the need for it in a close future so I vote to let it that way.

source/ed/docker_tests/ed_integration_tests.cpp
pbougue: @antoine-de will do that

source/jormungandr/tests/ptref_tests.py
pbougue, line 583: done

PR #1212, opened 2015-10-15
kraken: retry loading of realtime data if it fail the first time

kinnou02:


source/kraken/maintenance_worker.cpp
```diff
@@ -60,17 +60,27 @@ void MaintenanceWorker::load(){
...
     }
-    if(conf.is_realtime_enabled()){
-        load_realtime();
-    }
+    load_realtime();
 }
 
 
 void MaintenanceWorker::load_realtime(){
+    if(!conf.is_realtime_enabled()){
```
TeXitoi, line 68: spacing (and also after)

source/kraken/maintenance_worker.cpp
```diff
@@ -92,7 +102,8 @@ void MaintenanceWorker::load_realtime(){
     }else{
         LOG4CPLUS_WARN(logger, "no realtime data receive before timeout: going without it!");
     }
-
+    //no exception were throw, data has been loaded if there is any
```
TeXitoi: in the else clause just before, data has not be loaded...

xlqian:
Ok for me

PR #1213, opened 2015-10-15
jormun: handle bad format of date

kinnou02:
"T0800" seems to be a valid iso date, but the year is set to 800, we
won't be able to convert it to a timestamp.


xlqian:
Ok for me

PR #1215, opened 2015-10-16
Kraken: correcting compilation and warning

pbougue:
Did not compile on recent versions of gcc and clang

### DIFF
--- a/source/time_tables/thermometer.cpp
+++ b/source/time_tables/thermometer.cpp
@@ -140,7 +140,7 @@ struct VertexProperties
     VertexProperties() : idx(0) {}
     VertexProperties(type::idx_t const& i): idx(i) {}
 };
-using Graph = boost::adjacency_list<boost::vecS, boost::vecS, boost::directedS, boost::property<boost::vertex_bundle_t, VertexProperties> >;
+using Graph = boost::adjacency_list<boost::vecS, boost::vecS, boost::directedS, VertexProperties>;
 using Vertex = boost::graph_traits<Graph>::vertex_descriptor;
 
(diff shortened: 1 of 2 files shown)


source/time_tables/thermometer.cpp
```diff
@@ -140,7 +140,7 @@ struct VertexProperties
     VertexProperties() : idx(0) {}
     VertexProperties(type::idx_t const& i): idx(i) {}
 };
-using Graph = boost::adjacency_list<boost::vecS, boost::vecS, boost::directedS, boost::property<boost::vertex_bundle_t, VertexProperties> >;
+using Graph = boost::adjacency_list<boost::vecS, boost::vecS, boost::directedS, VertexProperties>;
```
pbougue, line 143: @xavierraffin we are going to merge that because it just did not compile on clang 3.6.0 and gcc 4.9.2, and tests are ok, but if there was a special purpose for that use of `vertex_bundle` we will have to do something else...

TeXitoi:
OK for me.

source/time_tables/thermometer.cpp
pbougue, line 143: Note @xavierraffin : you can reply to this comment even after merge :)

xavierraffin, line 143: OK for me, if that don't break tests.

Sample code from stack_overflow probably ^^

SouthWest french devs are so lazy you know ...

pbougue, line 143: Same here, I didn't bother to really look for a purpose, just waited for the SouthWest french devs to let me know ;)

PR #1217, opened 2015-10-16
jormun: fix a enormous bug in the retry of loading realtime

kinnou02:
We were setting "is_realtime_loaded" to false before checking if data
were loaded...


TeXitoi:
OK for me

PR #1223, opened 2015-10-20
ed2nav: optimisation in headsign handler

kinnou02:
the creation of the temporary vector was taking 12% of the time on
fr-idf.
Before this patch fr-idf was taking 4min30, after it took 3min50


pbougue:
Well seen :+1:

PR #1231, opened 2015-10-22
Jormun: Use pypy for jormungandr

xlqian:
fix some unit tests to prepare the use of pypy; some other improvements

This PR is a recovered version for: 

https://github.com/CanalTP/navitia/pull/1108

### DIFF
--- a/source/jormungandr/jormungandr/stat_manager.py
+++ b/source/jormungandr/jormungandr/stat_manager.py
@@ -209,7 +209,7 @@ class StatManager(object):
             stat_request.client = request.headers.getlist("X-Forwarded-For")[0]
         stat_request.path = request.path
 
-        stat_request.response_size = sys.getsizeof(call_result[0])
+        stat_request.response_size = 4242
 
         self.fill_info_response(stat_request.info_response, call_result)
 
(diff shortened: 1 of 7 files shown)


xlqian:
Jenkins, test this please

source/jormungandr/jormungandr/stat_manager.py
```diff
@@ -209,7 +209,7 @@ def fill_request(self, stat_request, call_result):
             stat_request.client = request.headers.getlist("X-Forwarded-For")[0]
         stat_request.path = request.path
 
-        stat_request.response_size = sys.getsizeof(call_result[0])
+        stat_request.response_size = 4242
```
francois-vincent, line 212: why is this necessary ?

xlqian, line 212: This line should be removed or commented. 

The intention of using `sys.getsizeof()` is to get the size of response in terms of 'length', but in fact this function returns the whole size of a python object in memory, the use may be wrong  and secondly this function is not yet implemented in pypy...

antoine-de:
close because it's been too long, and the new  solution will be to change flask/marshall to lima

PR #1234, opened 2015-10-23
Jormun: change scenario if the scenario hs been updated in db

kinnou02:
Previously we had to restart jormungandr for switching the scenario used
by an instance


antoine-de:
Jenkins, test this please

PR #1240, opened 2015-10-29
RT: connect apply disruption to the new VJs methods

antoine-de:
we want apply_disruption to handle chaos and kirin messages, so refactoring of apply_disruption to make it a bit more generic

### DIFF
--- a/source/kraken/apply_disruption.cpp
+++ b/source/kraken/apply_disruption.cpp
@@ -61,7 +61,7 @@ struct apply_impacts_visitor : public boost::static_visitor<> {
     virtual ~apply_impacts_visitor() {}
     apply_impacts_visitor(const apply_impacts_visitor&) = default;
 
-    virtual bool func_on_vj(nt::VehicleJourney&) = 0;
+    virtual void operator()(nt::MetaVehicleJourney*) = 0;
 
     void log_start_action(std::string uri) {
         LOG4CPLUS_TRACE(log4cplus::Logger::getInstance("log"),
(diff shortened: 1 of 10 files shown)


source/kraken/apply_disruption.cpp
```diff
@@ -99,63 +100,46 @@ struct apply_impacts_visitor : public boost::static_visitor<> {
...
-    void operator()(const nt::Route* route) {
+    void operator()(nt::Route* route) {
         this->log_start_action(route->uri);
-        route->for_each_vehicle_journey([&](nt::VehicleJourney& vj) {
-                return func_on_vj(vj);
+
+        // we cannot ensure that all VJ of a MetaVJ are on the same route,
+        // and since we want all actions to operate on MetaVJ, we collect all MetaVJ of the route
+        std::set<nt::MetaVehicleJourney*> mvjs;
+        route->for_each_vehicle_journey([&mvjs](nt::VehicleJourney& vj) {
```
kinnou02, line 108: we don't want to impact a vj that doesn't circulate on the impacted route

TeXitoi:
OK for me.

source/type/message.h
```diff
@@ -228,11 +228,15 @@ struct Tag {
 };
 
 struct Disruption {
+    Disruption() {}
+    Disruption(const std::string u, RTLevel lvl): uri(u), rt_level(lvl) {}
```
xlqian, line 232: `>>>>Troll<<<<`  `const std::string& u` :smiling_imp:

PR #1242, opened 2015-10-30
jormun: add a circuit breaker for not calling dead kraken each time 

kinnou02:
With this change we will call dead kraken only one time each minute, it
will prevent long response time on api who don't require a specific
coverage.


TeXitoi:
OK for me.

PR #1243, opened 2015-10-30
ed2nav: when using cities we try to use an admin already there

kinnou02:
Admin found by cities are search in georef by using there insee code, it
will reduce duplicate admin.

Refer to: http://jira.canaltp.fr/browse/NAVITIAII-1918

### DIFF
--- a/source/ed/ed2nav.cpp
+++ b/source/ed/ed2nav.cpp
@@ -59,6 +59,7 @@ struct FindAdminWithCities {
     boost::shared_ptr<pqxx::connection> conn;
     georef::GeoRef& georef;
     AdminMap added_admins;
+    AdminMap insee_admins_map;
     size_t nb_call = 0;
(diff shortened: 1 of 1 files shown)


source/ed/ed2nav.cpp
```diff
@@ -106,14 +118,17 @@ struct FindAdminWithCities {
         result_type res;
         for (auto it = result.begin(); it != result.end(); ++it) {
             const std::string uri = it["uri"].as<std::string>() + "extern";
-            navitia::georef::Admin*& admin = added_admins[uri];
+            const std::string insee = it["insee"].as<std::string>();
+            //we try to find the admin in georef by using it's insee code (only work in France)
+            navitia::georef::Admin*& admin = insee_admins_map[insee];
```
TeXitoi, line 123: It will not work if insee is empty...

PR #1246, opened 2015-11-02
Raptor: centralize definition of transfer penalty

pbougue:
Transfer penalty defined once and used everywhere needed ("les 2 minutes du peuple")

### DIFF
--- a/source/routing/raptor.cpp
+++ b/source/routing/raptor.cpp
@@ -270,6 +270,7 @@ make_starting_points_snd_phase(const RAPTOR& raptor,
 Journey convert_to_bound(const StartingPointSndPhase& sp,
                          uint32_t lower_bound_fb,
                          uint32_t lower_bound_conn,
+                         const navitia::time_duration& transfer_penalty,
                          bool clockwise) {
     Journey journey;
     journey.sections.resize(sp.count); // only the number of sections is part of the dominance function
@@ -283,7 +284,7 @@ Journey convert_to_bound(const StartingPointSndPhase& sp,
         journey.departure_dt = sp.end_dt;
     }
 
-    journey.transfer_dur = navitia::seconds(120 * sp.count + nb_conn * lower_bound_conn);
+    journey.transfer_dur = transfer_penalty * sp.count + navitia::seconds(nb_conn * lower_bound_conn);
     // provide best values on unknown criteria
     journey.min_waiting_dur = navitia::time_duration(boost::date_time::pos_infin);
     journey.nb_vj_extentions = 0;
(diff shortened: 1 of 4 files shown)


source/routing/raptor.cpp
```diff
@@ -360,7 +361,8 @@ RAPTOR::compute_all(const map_stop_point_duration& departures,
                     const std::vector<std::string>& forbidden_uri,
                     bool clockwise,
                     const boost::optional<navitia::time_duration>& direct_path_dur,
-                    const size_t max_extra_second_pass) {
+                    const size_t max_extra_second_pass,
+                    const navitia::time_duration& transfer_penalty) {
```
TeXitoi, line 365: I'd prefere max_extra_second_pass to be the last arg as we may remove it later.

source/routing/raptor_solution_reader.cpp
```diff
@@ -343,6 +345,7 @@ struct RaptorSolutionReader {
     const routing::map_stop_point_duration& sp_dur_arrs;// arrivals (not clockwise dependent)
     const type::RTLevel rt_level;
     const type::AccessibiliteParams& accessibilite_params;
+    navitia::time_duration transfer_penalty;
```
TeXitoi: you can const it.

source/routing/raptor.cpp
pbougue, line 365: It would modify some more lines as we do not care for transfer_penalty when calling compute_all so far and we would...

source/routing/raptor.cpp
```diff
@@ -440,7 +445,8 @@ RAPTOR::compute_all(const map_stop_point_duration& departures,
                                                    departures,
                                                    destinations,
                                                    rt_level,
-                                                   accessibilite_params);
+                                                   accessibilite_params,
```
xlqian: accessibilit~~E~~? Y?

source/routing/raptor.h
```diff
@@ -140,7 +140,8 @@ struct RAPTOR
                 const std::vector<std::string>& forbidden = std::vector<std::string>(),
                 bool clockwise = true,
                 const boost::optional<navitia::time_duration>& direct_path_dur = boost::none,
-                const size_t max_extra_second_pass = 0);
+                const size_t max_extra_second_pass = 0,
+                const navitia::time_duration& transfer_penalty = 2_min); // "les 2 minutes du peuple"
```
antoine-de, line 144: It may be easier to read if the default value is a bit more explicit (a static const above ? but as @TeXitoi is saying we don't want the default value to be used ...)

pbougue, line 144: We will probably have a struct containing all params at one point.
(And deal with exposition of this param in jormun also).
So we'll let it that way for this PR.

PR #1248, opened 2015-11-02
Kraken: clean deprecated proto

pbougue:


TeXitoi:
PR on navitia-proto created and merged.

PR #1252, opened 2015-11-03
doc: document realtime

stifoon:


documentation/navitia.io/source/integration.rst
```diff
@@ -926,9 +934,11 @@ Other parameters
 +----------+-----------------------+-----------+-------------------------------------------+-----------------+
 | nop      | max_nb_tranfers       | int       | Maximum of number transfers               | 10              |
 +----------+-----------------------+-----------+-------------------------------------------+-----------------+
-| nop      | disruption_active     | boolean   | If true the algorithm take the disruptions| False           |
+| nop      | disruption_active     | boolean   | For compatibility use only.               | False           |
+|          |                       |           | If true the algorithm take the disruptions| False           |
```
pbougue: One "False" is enough ?
Maybe mention that disruption_active=true => data_freshness=real_time ?
And as there is no strict equivalence with real_time, it's bit shaky but I don't know what to say instead.

TeXitoi:
OK for me if the useless False is removed

kinnou02:
It might be useful to add some explanation about realtime information, and not only update the parameters list of the api.

PR #1256, opened 2015-11-05
Tyr: New plan API

is06:


source/navitiacommon/navitiacommon/models.py
```diff
@@ -427,3 +434,9 @@ def find_by_uid(cls, uid):
     def __repr__(self):
         return '<DataSet %r>' % self.id
 
+
+class Plan(db.Model):
```
kinnou02: It's more a BillingPlan

source/navitiacommon/navitiacommon/models.py
```diff
@@ -427,3 +434,9 @@ def find_by_uid(cls, uid):
...
+    id = db.Column(db.Integer, primary_key=True)
+    name = db.Column(db.Text, nullable=False)
```
kinnou02, line 443: You will probably need to specify a default plan with a boolean

source/tyr/tyr/resources.py
```diff
@@ -701,3 +726,69 @@ def delete(self, name=None, traveler_type=None):
         except Exception:
             logging.exception("fail")
             raise
+
+
+class Plan(flask_restful.Resource):
+    def __init__(self):
```
kinnou02: If you don't need it, don't write it

source/tyr/tyr/resources.py
```diff
@@ -701,3 +726,69 @@ def delete(self, name=None, traveler_type=None):
...
+        pass
+
+    def get(self, plan_id=None):
+        if plan_id:
+            plan = models.Plan.query.get_or_404(plan_id)
+            return marshal(plan, plan_fields)
```
kinnou02: when you always use the same fields for marshaling you can use the decorator `marshall_with`

source/tyr/tyr/resources.py
```diff
@@ -701,3 +726,69 @@ def delete(self, name=None, traveler_type=None):
...
+        else:
+            plans = models.Plan.query.all()
+            return marshal(plans, plan_fields)
```
kinnou02: I don't think it's working, you have a list that you try to marshall as a single object

source/tyr/tyr/resources.py
```diff
@@ -701,3 +726,69 @@ def delete(self, name=None, traveler_type=None):
...
+            db.session.add(plan)
+            db.session.commit()
+            return marshal(plan, plan_fields)
+        except Exception:
+            logging.exception("fail")
+            raise
+
+    def put(self, plan_id=None):
+        plan = models.Plan.query.get_or_404(plan_id)
+        parser = reqparse.RequestParser()
```
kinnou02: It's seem to be the same parser used in post, you can create it in the ```_init__``` as a instance's attribute

kinnou02:
I don't see the migration file.

And there is a lot of code formatting (that I don't like, but it's not the point) that shouldn’t be in this PR

source/tyr/tyr/resources.py
is06: Like in... User class? :)

source/tyr/tyr/api.py
```diff
@@ -59,6 +58,10 @@
                  '/v0/instances/<string:name>/traveler_profiles/',
                  '/v0/instances/<string:name>/traveler_profiles/<string:traveler_type>')
 
+api.add_resource(resources.BillingPlan,
+                 '/v0/billing-plans/',
```
kinnou02: We use underscore in urls

is06:
Migrations added

source/tyr/migrations/versions/3e56c7e0a4a4_create_billing_plan_table.py
```diff
@@ -0,0 +1,43 @@
...
+
+def upgrade():
+
+    op.create_table('billing_plan',
+        sa.Column(
+            'id', sa.Integer(),
+            primary_key=True,
+            nullable=False
+        ),
+        sa.Column('name', sa.String(255), nullable=False),
```
kinnou02: use `Text` we aren't in 90's anymore

source/tyr/migrations/versions/3e56c7e0a4a4_create_billing_plan_table.py
```diff
@@ -0,0 +1,43 @@
...
+        sa.Column('max_request_count', sa.Integer(), nullable=True),
+        sa.Column('max_object_count', sa.Integer(), nullable=True),
+        sa.Column('default', sa.Boolean(), nullable=False)
+    )
+
+    op.execute("INSERT INTO billing_plan (name, max_request_count, max_object_count, \"default\") VALUES ('Developpeur (3000 requetes max)',3000,3000,true),('Entreprise (requetes illimitees)',NULL,NULL,false);")
```
kinnou02: I not agree with that! the enterprise plan should be the default plan, actually all media except the sncf one create user with unlimited calls.

source/tyr/migrations/versions/3e56c7e0a4a4_create_billing_plan_table.py
```diff
@@ -0,0 +1,43 @@
...
+            primary_key=True,
+            nullable=False
+        ),
+        sa.Column('name', sa.Text, nullable=False),
+        sa.Column('max_request_count', sa.Integer(), nullable=True),
+        sa.Column('max_object_count', sa.Integer(), nullable=True),
+        sa.Column('default', sa.Boolean(), nullable=False)
+    )
+
+    op.execute("INSERT INTO billing_plan (name, max_request_count, max_object_count, \"default\") VALUES ('developer_billing_plan',3000,3000,true),('enterprise_billing_plan',NULL,NULL,false);")
```
kinnou02: I not agree with that! the enterprise plan should be the default plan, actually all media except the sncf one create user with unlimited calls.

source/tyr/migrations/versions/3e56c7e0a4a4_create_billing_plan_table.py
```diff
@@ -0,0 +1,43 @@
...
+
+    op.add_column(u'user', sa.Column('billing_plan_id', sa.Integer(), nullable=True))
+    op.create_foreign_key(
```
kinnou02, line 41: This foreign_key isn't in the models

source/tyr/migrations/versions/3e56c7e0a4a4_create_billing_plan_table.py
```diff
@@ -0,0 +1,51 @@
...
+        sa.Column(
+            'id', sa.Integer(),
+            primary_key=True,
+            nullable=False
+        ),
+        sa.Column('name', sa.Text(), nullable=False),
+        sa.Column('max_request_count', sa.Integer(), nullable=True),
+        sa.Column('max_object_count', sa.Integer(), nullable=True),
+        sa.Column('default', sa.Boolean(), nullable=False),
+        sa.Column('end_point_id', sa.Integer(), nullable=False, server_default='1')
```
kinnou02: This is not in the model

source/tyr/migrations/versions/3e56c7e0a4a4_create_billing_plan_table.py
```diff
@@ -0,0 +1,51 @@
...
+    )
+
+    op.create_foreign_key(
+        "fk_billing_plan_end_point", "billing_plan",
+        "end_point", ["end_point_id"], ["id"]
+    )
+
+    op.execute("INSERT INTO end_point (name, \"default\") SELECT 'sncf',false WHERE NOT EXISTS (SELECT id FROM end_point WHERE id = 2 AND name = 'sncf');")
```
kinnou02: You shouldn't check with the id, there is no guarantee about it

source/tyr/migrations/versions/3e56c7e0a4a4_create_billing_plan_table.py
```diff
@@ -0,0 +1,51 @@
...
+    op.execute("INSERT INTO billing_plan (name, max_request_count, max_object_count, \"default\", end_point_id) VALUES ('nav_dev',3000,NULL,true,1),('nav_ent',NULL,NULL,false,1),('nav_ctp',NULL,NULL,false,1),('sncf_dev',3000,60000,true,2),('sncf_ent',NULL,NULL,false,2);")
```
kinnou02: again for now the default should be nav_ent

PR #1257, opened 2015-11-06
Build: patches for Ubuntu 15.10

antoine-de:
ubuntu 1.510 uses gcc 5.2 and boost 1.58 are there were several problems with the change.
- Boost
  - boost vector serialization is buggy (and has been patched in boost 1.59)
  - some deprecated includes were moved
  - eos_portable_archive was using deprecated stuff
  - BOOST_CLASS_EXPORT_GUID does not seems to work anymore, but at the same time does not seems to be mandatory anymore :sweat_smile: 
- GCC 5.2
  - gcc 5.2 seems to be quite aggresive with the optimization and when doing:
    `for (auto bob: make_a_range().advance_begin(1)) {...}`
    it optimize out the inner range, so the advance begin returns a reference on a deleted object
  - some tests on sorted vector for equivalent entry
  - some explicit constructor calls
  - a fixture file was containing a windows return line (\r), and was creating problems in the gtfs_parser

Compiled, unit-tested and docker-tested with:
- Ubuntu 15.04, clang 3.6.0 / gcc 4.9.2, boost 1.54, release
- Ubuntu 15.10, gcc 5.2, boost 1.58, release
- debian 7, gcc 4.7, boost 1.49, release
- Ubuntu 14.04,  gcc 4.8.4, boost 1.54, release

### DIFF
--- a/source/autocomplete/tests/test.cpp
+++ b/source/autocomplete/tests/test.cpp
@@ -406,14 +406,12 @@ BOOST_AUTO_TEST_CASE(Faute_de_frappe_One){
 
         auto res1 = ac.find_partial_with_pattern("gare patea", word_weight, nbmax, [](int){return true;}, ghostwords);
         BOOST_REQUIRE_EQUAL(res1.size(), 3);
-        BOOST_CHECK_EQUAL(res1.at(0).idx, 4);
-        BOOST_CHECK_EQUAL(res1.at(1).idx, 1);
+        std::initializer_list<unsigned> best_res = {1, 4}; // they are equivalent
+        BOOST_CHECK(in(res1.at(0).idx, best_res));
+        BOOST_CHECK(in(res1.at(1).idx, best_res));
         BOOST_CHECK_EQUAL(res1.at(2).idx, 0);
         BOOST_CHECK_EQUAL(res1.at(0).quality, 94);
-
-
     }
-
 
 /*
     > Le fonctionnement normal :> On prends que les autocomplete si tous les mots dans la recherche existent
--- a/source/autocomplete/tests/test.cpp
+++ b/source/autocomplete/tests/test.cpp
@@ -409,6 +409,7 @@ BOOST_AUTO_TEST_CASE(Faute_de_frappe_One){
         std::initializer_list<unsigned> best_res = {1, 4}; // they are equivalent
         BOOST_CHECK(in(res1.at(0).idx, best_res));
         BOOST_CHECK(in(res1.at(1).idx, best_res));
+        BOOST_CHECK_NE(res1.at(0).idx, res1.at(1).idx);
         BOOST_CHECK_EQUAL(res1.at(2).idx, 0);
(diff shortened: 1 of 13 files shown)


source/autocomplete/tests/test.cpp
```diff
@@ -406,15 +406,13 @@ BOOST_AUTO_TEST_CASE(Faute_de_frappe_One){
 
         auto res1 = ac.find_partial_with_pattern("gare patea", word_weight, nbmax, [](int){return true;}, ghostwords);
         BOOST_REQUIRE_EQUAL(res1.size(), 3);
-        BOOST_CHECK_EQUAL(res1.at(0).idx, 4);
-        BOOST_CHECK_EQUAL(res1.at(1).idx, 1);
+        std::initializer_list<unsigned> best_res = {1, 4}; // they are equivalent
+        BOOST_CHECK(in(res1.at(0).idx, best_res));
+        BOOST_CHECK(in(res1.at(1).idx, best_res));
```
xlqian, line 411: does it make sense to add `BOOST_CHECK_NE(res1.at(0).idx, res1.at(1).idx);`?

source/ed/connectors/fusio_parser.cpp
```diff
@@ -827,6 +827,11 @@ void CommercialModeFusioHandler::handle_line(Data& data, const csv_row& row, boo
                         "  file has more than one commercial mode and no commercial_mode_id column");
         throw InvalidHeaders(csv.filename);
     }
+    if (! has_col(name_c, row)) {
+        LOG4CPLUS_WARN(logger, "invalid line on " << csv.filename <<
+                        "  commercial mode " << row[id_c] << " is missing a name");
+        return;
+    }
```
xlqian, line 834: How come this part is concerned?

source/autocomplete/tests/test.cpp
antoine-de, line 411: yup you're right. Done.

source/ed/connectors/fusio_parser.cpp
antoine-de, line 834: yes it's strange, a fixture file was containing a windows return line (\r), and with the old gcc I think the vector was empty and is not empty now.

Anyway it makes sens to check the size of the vector before accessing it

source/routing/raptor_solution_reader.cpp
```diff
@@ -68,7 +68,8 @@ get_out_st_dt(const std::pair<const type::StopTime*, DateTime>& in_st_dt,
               const JourneyPatternContainer& jp_container)
 {
     DateTime cur_dt = in_st_dt.second;
-    for (const auto& st: raptor_visitor().st_range(*in_st_dt.first).advance_begin(1)) {
+    auto st_range = raptor_visitor().st_range(*in_st_dt.first);
+    for (const auto& st: st_range.advance_begin(1)) {
```
TeXitoi, line 72: Maybe adding a comment explaining why you create a variable?

TeXitoi:
OK for me.

PR #1262, opened 2015-11-09
jormun: don't call redis each time we need the model for the same request

kinnou02:
We were doing a lot of call to redis, in production for a /journey we do
50 call to redis, even with 1ms response time it's 50ms!


antoine-de:
OK for me

PR #1272, opened 2015-11-13
Ed: retrieving stop_times by chunks

pbougue:
Goal is to lighten memory on ed2nav.
Needs to be tested on artemis (maybe after merge on dev is enough), see if reading is not affected (but not much chance, as docker_tests are ok and tested it on some journeys).

Bench on stif ed2nav:
- before: VSZ 9.20 Go, 316s user, 6min23 total
- after: VSZ 5.16 Go, 317s user, 6min36 total

No change on postgres process memory.

### DIFF
--- a/source/ed/ed_reader.cpp
+++ b/source/ed/ed_reader.cpp
@@ -924,46 +924,56 @@ void EdReader::fill_stop_times(nt::Data& data, pqxx::work& work) {
         "FROM navitia.stop_time as st "
         "ORDER BY st.vehicle_journey_id, st.\"order\"";
 
-    pqxx::result result = work.exec(request);
+    pqxx::stateless_cursor<pqxx::cursor_base::read_only, pqxx::cursor_base::owned>
+        cursor( work, request, "stcursor", false );
+    size_t current_idx = 0;
+    size_t chunk_limit = 100000;
+    size_t nb_rows = cursor.size();
 
-    for(auto const_it = result.begin(); const_it != result.end(); ++const_it) {
-        const auto vj_id = const_it["vehicle_journey_id"].as<idx_t>();
-        auto& sts = sts_from_vj[vj_id];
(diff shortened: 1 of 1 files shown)


source/ed/ed_reader.cpp
```diff
@@ -924,46 +924,56 @@ void EdReader::fill_stop_times(nt::Data& data, pqxx::work& work) {
...
         "ORDER BY st.vehicle_journey_id, st.\"order\"";
 
-    pqxx::result result = work.exec(request);
+    pqxx::stateless_cursor<pqxx::cursor_base::read_only, pqxx::cursor_base::owned>
+        cursor( work, request, "stcursor", false );
+    size_t current_idx = 0;
+    size_t chunk_limit = 100000;
+    size_t nb_rows = cursor.size();
+
+    while(current_idx < nb_rows) {
```
TeXitoi:
``` c++
const size_t chunk_limit = 100000;
const size_t nb_rows = cursor.size();
for (size_t current_idx = 0; current_idx < nb_rows; current_idx += chunk_limit) {
```

source/ed/ed_reader.cpp
```diff
@@ -924,46 +924,56 @@ void EdReader::fill_stop_times(nt::Data& data, pqxx::work& work) {
...
+        pqxx::result result = cursor.retrieve( current_idx, std::min(current_idx + chunk_limit, nb_rows));
```
TeXitoi: spacing

source/ed/ed_reader.cpp
```diff
@@ -924,46 +924,56 @@ void EdReader::fill_stop_times(nt::Data& data, pqxx::work& work) {
...
+
+        for(auto const_it = result.begin(); const_it != result.end(); ++const_it) {
```
TeXitoi: spacing (`for (`)

PR #1275, opened 2015-11-16
RT: filter non consistent realtime messages

antoine-de:
we don't want VJ's with non ordered stop times, so we filter them


pbougue:
Ok for me, but we need another PR for forbidden pickup/dropoff that shouldn't be checked for inconsistencies

antoine-de:
It will be done in @xlqian PR

TeXitoi:
OK for me.

PR #1276, opened 2015-11-16
Ed: no more ORDER BY on stop_times in db

pbougue:
Bench on stif:
- before: 238s user, 5.17Go
- after: 230s user, 5.17Go


fixtures/ed/gtfs_google_example/stop_times.txt
```diff
@@ -1,6 +1,6 @@
 trip_id,arrival_time,departure_time,stop_id,stop_sequence,stop_headsign,pickup_type,drop_off_time,shape_dist_traveled
 STBA,6:00:00,6:00:00,STAGECOACH,1,,,,
-STBA,6:20:00,6:20:00,BEATTY_AIRPORT,2,,,,
+STBA,6:20:00,6:20:00,BEATTY_AIRPORT,23,,,,
```
TeXitoi: changing this file is problematic: it's the official gtfs example

source/ed/connectors/gtfs_parser.cpp
```diff
@@ -1046,6 +1046,10 @@ void StopTimeGtfsHandler::finish(Data& data) {
                 return st1->order < st2->order;
             });
 
+        for (size_t it_st = 0; it_st < vj->stop_time_list.size(); ++it_st) {
+            vj->stop_time_list[it_st]->order =  it_st;
```
antoine-de, line 1050: minor, you can use boost::for_each(vj->stop_time_list, Indexer());

pbougue, line 1050: Good to know, even if Indexer is working on idx, not order :(

PR #1279, opened 2015-11-18
kraken: fix loading of disruption with multiple application periods

kinnou02:
When loading disruptions from the database one message and one pt_object
was created for each application_period of an impact (or close).
We now use a set to check if the current message/pt_object has not
already be created.

### DIFF
--- a/source/kraken/fill_disruption_from_database.h
+++ b/source/kraken/fill_disruption_from_database.h
@@ -66,10 +66,11 @@ namespace navitia {
         chaos::Message* message = nullptr;
         chaos::Channel* channel = nullptr;
 
-        std::string last_message_id = "",
-                    last_ptobject_id = "",
-                    last_period_id = "",
+        std::string last_period_id = "",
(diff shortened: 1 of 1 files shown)


source/kraken/fill_disruption_from_database.h
```diff
@@ -93,25 +94,28 @@ namespace navitia {
...
                 channel = nullptr;
+                message_ids.clear();
             }
 
             if (impact && (last_period_id != const_it["application_id"].template as<std::string>())) {
                 fill_application_period(const_it);
             }
-            if (impact && (last_ptobject_id != const_it["ptobject_id"].template as<std::string>())) {
+            if (impact
+                    && (pt_object_ids.find(const_it["ptobject_id"].template as<std::string>()) == pt_object_ids.end())) {
```
antoine-de: you can use `! pt_object_ids.count(const_it["ptobject_id"].template as<std::string>())`

PR #1285, opened 2015-11-20
add license website in contributor object

azime:
add license and website in contributor object


antoine-de:
This is OK for me.

Can you add docker test @azime  ?

you need to add a check on the contributor in source/ed/docker_tests/ed_integration_tests.cpp and build the tests with `make docker_test`.

PR #1286, opened 2015-11-20
Jormungandr: remove useless API routes

antoine-de:
`v1/coverage/{collection}/id` is not a valid url, we remove it in the routing

it was useful to create links for API that does not need a /coverage (/journeys for example) thought,
so for these API we do not create links

fix http://jira.canaltp.fr/browse/NAVITIAII-1896


TeXitoi:
OK for me. @kinnou02 I let you review this.

kinnou02:
It seems good, let's hope there is no flask's magic somewhere

PR #1289, opened 2015-11-20
kraken: add a timeout to the rabbitmq connection

kinnou02:
Our beloved loadbalancer/firewall can choose to drop a session from
their table and never say anything to anyone.
Kraken don't know it, but he is already dead!
With this PR we will create a new connection if there is no message for
a period of time.

There is a small problem to solve before being able to merge this PR:
Actually the keep alive is handled by tyr, but when we deploy we stop
tyr_beat, so there is no more keep alive coming and all instances will
restart their connection periodically and more importantly reload all
realtime information: we don't want that to happen...


francois-vincent:
Jenkins, test this please

PR #1293, opened 2015-11-23
Jormun: refacto "is_open_data" computing

kinnou02:
This will  prevent a request to the database for each request to /status

And by default an instance is not in open_data.


antoine-de:
OK for me

TeXitoi:
OK for me.

PR #1297, opened 2015-11-24
Eitri: wait for the docker build to run eitri

antoine-de:
else if this was the first time eitri was run it did not work properly


TeXitoi:
OK for me.

PR #1303, opened 2015-11-27
RT: change delay enum

antoine-de:
this enum should be computed by kirin, but it is not the case, so for
the moment all TripDescriptor_ScheduleRelationship_SCHEDULED are delayed
train


TeXitoi:
OK for me.

PR #1218, opened 2015-10-19
TimeTable: Fix route schedule with calendar

antoine-de:
fixes #1161 (and http://jira.canaltp.fr/browse/NAVITIAII-1872)

/route_schedules with calendar are now without date and the sort have been corrected.

On the dataset given in #1161 the behaviour is now:

 classical route_schedules with a calendar:

`/routes/route:1_A/route_schedules?calendar=Y2FsZW5kYXI6OA`

```
{
  "schedule": [
    {
      "vj": "vehicle_journey:2",
      "dt": "005000"
    },
    {
      "vj": "vehicle_journey:3",
      "dt": "015000"
    },
    {
      "vj": "vehicle_journey:4",
      "dt": "025000"
    }
  ],
  "stop": "stop_point:1"
}
{
  "schedule": [
    {
      "vj": "vehicle_journey:2",
      "dt": "010500"
    },
    {
      "vj": "vehicle_journey:3",
      "dt": "020500"
    },
    {
      "vj": "vehicle_journey:4",
      "dt": "030500"
    }
  ],
  "stop": "stop_point:2"
}
{
  "schedule": [
    {
      "vj": "vehicle_journey:2",
      "dt": "011500"
    },
    {
      "vj": "vehicle_journey:3",
      "dt": "021500"
    },
    {
      "vj": "vehicle_journey:4",
      "dt": "031500"
    }
  ],
  "stop": "stop_point:3"
}


```

/route_schedules with a calendar, and sorted with a pivot time at midnight (same as above)
`/routes/route:1_A/route_schedules?calendar=Y2FsZW5kYXI6OA&from_datetime=20160704T000000`

```
{
  "schedule": [
    {
      "vj": "vehicle_journey:2",
      "dt": "005000"
    },
    {
      "vj": "vehicle_journey:3",
      "dt": "015000"
    },
    {
      "vj": "vehicle_journey:4",
      "dt": "025000"
    }
  ],
  "stop": "stop_point:1"
}
{
  "schedule": [
    {
      "vj": "vehicle_journey:2",
      "dt": "010500"
    },
    {
      "vj": "vehicle_journey:3",
      "dt": "020500"
    },
    {
      "vj": "vehicle_journey:4",
      "dt": "030500"
    }
  ],
  "stop": "stop_point:2"
}
{
  "schedule": [
    {
      "vj": "vehicle_journey:2",
      "dt": "011500"
    },
    {
      "vj": "vehicle_journey:3",
      "dt": "021500"
    },
    {
[...]

PR #1304, opened 2015-11-27
Pr route schedule over midnight

xavierraffin:
The previous PR #1218 correct only a part of the issue of services aver midnight.

It heavily depending on what you have in your data.
Hopefully, all tricky doomed data use case are present at Tisséo ;-)

I made a unit test to reproduce and provide code to solve the issue.

The test case is the following 

There is three services deserving three stops.
One integraly befor midnight, one integraly after midnight and on over midnight.

```
       A        B       C     
st1 |23:30   |23:50  |0:10   
st2 |23:40   |0:0    |0:20   
st3 |23:50   |0:10   |0:30   
```

and in seconds from midnight

```
       A        B       C     
st1 |86370   |86390  |10   
st2 |86380   |0      |20   
st3 |86390   |10     |30   
```

with a calendar switched by 1 for VJ C

We would like to display route_schedules with a calendar (TimeTable typical usecase).

Before this PR, the result was sorted like this 

```
      B       A      C     
st1 |23:50  |23:30  |0:10   
st2 |0:0    |23:40  |0:20   
st3 |0:10   |23:50  |0:30   
```

This is because the comparison used to sort column compare each cells and sort given the majority.

In PR #1218 we add 86400 to all VJ stop times of which the first VJ stop time is over midnight (or over a particular offset).

This is a good idea, but that does not save the sorting because the input is 

```
       A        B       C     
st1 |86370   |86390  |86410   
st2 |86380   |0      |86420
st3 |86390   |10     |86430
```

_This is a_ **return to the future** _sort of_

So I just change the code in order to add 86400 to all stop_times over midnight (or offset) for all cells and all cells after it.

The data before sort became

```
       A        B       C     
st1 |86370   |86390  |86410   
st2 |86380   |86400  |86420
st3 |86390   |86410  |86430
```

So after this correction the result is correctly sorted

```
       A        B       C     
st1 |23:30   |23:50  |0:10   
st2 |23:40   |0:0    |0:20   
st3 |23:50   |0:10   |0:30   
```

**I test it on a lot of real tricky use cases and with a lot of different offset . We are good !**

### DIFF
--- a/source/time_tables/tests/route_schedules_test.cpp
+++ b/source/time_tables/tests/route_schedules_test.cpp
@@ -448,8 +448,13 @@ BOOST_FIXTURE_TEST_CASE(test_get_all_route_stop_times_with_different_vp_and_hour
 // S1 23:30  23:50  00:10
 // S2 23:40  00:00  00:20
 // S3 23:50  00:10  00:30
-BOOST_AUTO_TEST_CASE(test_get_all_route_stop_times_with_different_vp_over_midnight) {
+BOOST_AUTO_TEST_CASE(test_route_schedule_with_different_vp_over_midnight) {
     ed::builder b = {"20151127"};
+    navitia::type::Calendar *c1;
+
+    boost::gregorian::date begin = boost::gregorian::date_from_iso_string("20150101");
+    boost::gregorian::date end = boost::gregorian::date_from_iso_string("20160101");
+
     b.vj("L", "111111", "", true, "A", "A")
         ("st1", "23:30"_t)
         ("st2", "23:40"_t)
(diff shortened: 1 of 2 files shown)


source/time_tables/tests/route_schedules_test.cpp
```diff
@@ -443,6 +443,66 @@ BOOST_FIXTURE_TEST_CASE(test_get_all_route_stop_times_with_different_vp_and_hour
...
+    boost::gregorian::date end = boost::gregorian::date_from_iso_string("20160101");
+
+    b.vj("L", "111111", "", true, "A", "A")
+        ("st1", "23:30"_t)
+        ("st2", "23:40"_t)
+        ("st3", "23:50"_t);
+    b.vj("L", "111111", "", true, "B", "B")
+        ("st1", "23:50"_t)
+        ("st2", "0:00"_t)
+        ("st3", "0:10"_t);
```
TeXitoi: It should be `("st1", "23:50"_t)("st2", "24:00"_t)("st3", "24:10"_t)`. Maybe you have the problem because your exports don't satisfy the GTFS specs on the stop_times (on which NTFS is based) : https://developers.google.com/transit/gtfs/reference#stop_times_fields

xavierraffin:
Because of the way ed stores stop times there is no relation with GTFS here.

But, I checked our GTFS, and we follow GTFS specs.

I am not asking you to believe me on my words, so I have build a custom GTFS for demonstration purpose.

# GTFS demonstration

There is a GTFS : http://web31srv.tisseo.fr/NTFS_1304.zip

It contains a very simple network, with only one line and three stops.
There is a unique grid_calendar WE from 7/1/2016 au 7/31/2016.
Stops are named 1 2 and 3.

The stop times are the same as described before : 

```
trip_id,stop_id,stop_sequence,arrival_time,departure_time,stop_headsign,pickup_type,drop_off_type,date_time_estimated
2,1,1,23:30:00,23:30:00,,0,0,1
2,2,2,23:40:00,23:40:00,,0,0,1
2,3,3,23:50:00,23:50:00,,0,0,1
3,1,1,23:50:00,23:50:00,,0,0,1
3,2,2,24:00:00,24:00:00,,0,0,1
3,3,3,24:10:00,24:10:00,,0,0,1
4,1,1,24:10:00,24:10:00,,0,0,1
4,2,2,24:20:00,24:20:00,,0,0,1
4,3,3,24:30:00,24:30:00,,0,0,1
```

You will see that with this GTFS, **I can reproduce the wrong behaviour**

So in ED (because of GMT offset) stop times become

```
=> select vehicle_journey_id, departure_time from stop_time;
 vehicle_journey_id | departure_time
--------------------+----------------
                  0 |          77400
                  0 |          78000
                  0 |          78600
                  1 |          78600
                  1 |          79200
                  1 |          79800
                  2 |          79800
                  2 |          80400
                  2 |          81000
```

So, now a request on route_schedules on the WE calendar with an offset at 4PM:

/v1/coverage/tisseo_fh/routes/route:1_A/route_schedules?from_datetime=20160704T030000&calendar=Y2FsZW5kYXI6OA&show_codes=true

With the current patch, answer is correct:

```
"date_times": [
{
    "date_time": "233000",
...
    "date_time": "235000",
...
    "date_time": "001000",      
```

Without my patch answer is wrong : 

```
"date_times": [
{
    "date_time": "235000",
...
    "date_time": "233000",
...
    "date_time": "001000",      
```

I hope this demonstration will be enought for you to merge this PR :-)

TeXitoi:
OK, that's because get_stop_times truncates in [0; 24h[. So you just have to change the test to validate the GTFS specs, and then I can merge this PR.

xavierraffin:
Done.

I tested before and after patch :

Before patch : 

```
      B       A      C     
st1 |23:50  |23:30  |0:10   
st2 |0:0    |23:40  |0:20   
st3 |0:10   |23:50  |0:30   
```

After patch

```
       A        B       C     
st1 |23:30   |23:50  |0:10   
st2 |23:40   |0:0    |0:20   
st3 |23:50   |0:10   |0:30   
```

TeXitoi:
Great! Thanks.

OK to merge (waiting for jenkins).

xavierraffin:
Do you think you could merge soon ?
I have got an other PR and it is complicated to make two PR at the same time (lazy dev technique).

TeXitoi:
Done!

PR #1305, opened 2015-11-27
Add the possibility to choose what type of poi import from osm

kinnou02:
close: http://jira.canaltp.fr/browse/NAVP-64

Osm2ed can now take in parameters what type of poi to import. 
It also force an order for poi with multiple type:
amenity > tourism > leisure > aeroway > railway > shop > others

It's possible to choose the type of poi for an instance in tyr with the api.

By default nothing change!

### DIFF
--- a/source/ed/osm2ed.cpp
+++ b/source/ed/osm2ed.cpp
@@ -845,11 +845,11 @@ void PoiHouseNumberVisitor::fill_poi(const u_int64_t osm_id, const CanalTP::Tags
         return;
 
     std::string ref_tag = "";
-    if (tags.find("amenity") != tags.end()) {
-            ref_tag = "amenity";
-    }
-    if (tags.find("leisure") != tags.end()) {
-            ref_tag = "leisure";
+    for(auto tags_type: tags_types){
+        if(tags.find(tags_type) != tags.end()){
+            ref_tag = tags_type;
+            break;
+        }
     }
     if (ref_tag.empty()) {
         return;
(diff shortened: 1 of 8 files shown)


source/ed/osm2ed.cpp
```diff
@@ -53,6 +53,21 @@ namespace pt = boost::posix_time;
 
 namespace ed { namespace connectors {
 
+bool poi_type_comp::operator()(const std::string& a, const std::string& b){
+    auto it_a = std::find(order.begin(), order.end(), a);
```
antoine-de: boost::find

source/ed/osm2ed.cpp
```diff
@@ -53,6 +53,21 @@ namespace pt = boost::posix_time;
 
 namespace ed { namespace connectors {
 
+bool poi_type_comp::operator()(const std::string& a, const std::string& b){
```
antoine-de, line 57: spacing

source/ed/osm2ed.cpp
```diff
@@ -907,11 +922,42 @@ void OSMCache::flag_nodes() {
...
+    std::map<std::string, std::string> result;
+    for(auto type: poi_types){
+        std::vector<std::string> strs;
+        boost::algorithm::split(strs, type, boost::is_any_of("="));
+        if(strs.size() == 1){
+            result[strs[0]] = strs[0];
+        }else if(strs.size() == 2){
+            result[strs[0]] = strs[1];
+        }else{
+            std::cout << "poi_type: \"" << type << "\" ignored!" << std::endl;
```
antoine-de, line 936: don't we have some loggers ?

kinnou02, line 936: it's used for parse the command line input, others case like this (--help, --version) output on cout too.
It seemed more logic to used cout for this too.

antoine-de, line 936: k

source/ed/osm2ed.cpp
```diff
@@ -845,11 +860,11 @@ void PoiHouseNumberVisitor::fill_poi(const u_int64_t osm_id, const CanalTP::Tags
         return;
 
     std::string ref_tag = "";
-    if (tags.find("amenity") != tags.end()) {
-            ref_tag = "amenity";
-    }
-    if (tags.find("leisure") != tags.end()) {
-            ref_tag = "leisure";
+    for(auto tags_type: tags_types){
```
TeXitoi: `for (const auto& tags_type: tags_types) {`

PR #1308, opened 2015-11-30
RT: fix disruption API for vehicle_journey

antoine-de:
the call vehicle_journey/bob/disruptions was not working, fix this

fix also the link to the disruption in the vj


TeXitoi:
OK for me.

PR #1309, opened 2015-12-01
kraken: fix journey pattern auto generation with multiple utc offset on vjs

xavierraffin:
I push this without test because I would like be sure that you won't fix the same issue at the same time.
Unit tests will follow if you don't already solve it yourself.

**So please tell me if you are already working on it or if I could continue**

The issue is quite simple to understand.

When there is different time offset on vehicule journey on a unique route, the journey pattern auto generation (overtake function) is wrong because it compares times off different offset.

Imagine you have a VJ starting at 15h with an offset of 3600 and a VJ starting at 16h with a 7200 offset (with different speeds).
Then overtake function will detect an overtake and create a journey pattern.

This additionnal JP creates very bad things in services like route_schedules.

### DIFF
--- a/source/routing/tests/journey_pattern_container_test.cpp
+++ b/source/routing/tests/journey_pattern_container_test.cpp
@@ -247,3 +247,47 @@ BOOST_AUTO_TEST_CASE(lightly_overtaking2) {
     BOOST_CHECK_EQUAL(jps.nb_jps(), 2);
 }
 
+// two vjs must not overtake together according to utc offset
+BOOST_AUTO_TEST_CASE(offset_no_overtaking) {
+    ed::builder b("20150101");
+    navitia::type::VehicleJourney *vj1, *vj2;
+
+    vj1 = b.vj("1", "000111","", true, "VJ1", "MVJ1")("A", "8:00"_t, "8:00"_t)("B", "8:10"_t, "8:10"_t)("C", "8:20"_t, "8:20"_t).make();
+    vj2 = b.vj("1", "000111","", true, "VJ2", "MVJ2")("A", "8:00"_t, "8:00"_t)("B", "8:05"_t, "8:05"_t)("C", "8:10"_t, "8:10"_t).make();
+
+    vj1->utc_to_local_offset = "01:00"_t;
+    vj2->utc_to_local_offset = "02:00"_t;
+
+    b.data->pt_data->index();
+    b.finish();
+    b.data->build_raptor();
+    b.data->build_uri();
+    const nt::PT_Data& d = *b.data->pt_data;
+
+    nr::JourneyPatternContainer jps;
+    jps.load(d);
+    BOOST_CHECK_EQUAL(jps.nb_jps(), 1);
+}
+
+// two vjs must overtake together according to utc offset
+BOOST_AUTO_TEST_CASE(offset_overtaking) {
+    ed::builder b("20150101");
+    navitia::type::VehicleJourney *vj1, *vj2;
+
+    vj1 = b.vj("1", "000111","", true, "VJ1", "MVJ1")("A", "8:00"_t, "8:00"_t)("B", "8:10"_t, "8:10"_t)("C", "8:20"_t, "8:20"_t).make();
+    vj2 = b.vj("1", "000111","", true, "VJ2", "MVJ2")("A", "8:50"_t, "8:50"_t)("B", "9:20"_t, "9:20"_t)("C", "9:40"_t, "9:40"_t).make();
+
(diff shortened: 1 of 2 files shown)


xavierraffin:
I just add two unit tests covering this issue.

TeXitoi:
By construction, this is not a bug as 2 vj on different timezone will not have their validity patterns that overlap (and thus can't overtake). But I suspect your problem is on calendar route schedule...

Your fix will not do the job as the validity patterns will not overlap, and thus, the overtake test will not be done https://github.com/CanalTP/navitia/blob/0566c4339e71373e1874b5948bd26045109b1527/source/routing/journey_pattern_container.cpp#L143

In another hand, we can't remove this test as it may increase the number of journey patterns, which is bad for performances.

I don't know how to fix that.

xavierraffin:
Yes, the issue is on calendar route schedule...

**I suggest to merge my PR as it is now :-)**

Effectively, addition to runtime code is very small and don't impact perfs (just few additions) and don't increase complexity too heavily.

The choice to keep my unit test is yours.

source/routing/tests/journey_pattern_container_test.cpp
```diff
@@ -247,3 +247,47 @@ BOOST_AUTO_TEST_CASE(lightly_overtaking2) {
...
+// two vjs must overtake together according to utc offset
+BOOST_AUTO_TEST_CASE(offset_overtaking) {
+    ed::builder b("20150101");
+    navitia::type::VehicleJourney *vj1, *vj2;
+
+    vj1 = b.vj("1", "000111","", true, "VJ1", "MVJ1")("A", "8:00"_t, "8:00"_t)("B", "8:10"_t, "8:10"_t)("C", "8:20"_t, "8:20"_t).make();
+    vj2 = b.vj("1", "000111","", true, "VJ2", "MVJ2")("A", "8:50"_t, "8:50"_t)("B", "9:20"_t, "9:20"_t)("C", "9:40"_t, "9:40"_t).make();
+
+    vj1->utc_to_local_offset = "02:00"_t;
+    vj2->utc_to_local_offset = "01:00"_t;
```
antoine-de, line 281: I don't think your test is realistic because in the GTFS you can have only one timezone, so vj's with different utc_to_local_offset must be on different period. In your test this will thus change the validitypattern of vj2 to "111000".

And as @TeXitoi said, if the vehicle journey are not on the same period we do not compare their time (the code you patched):
https://github.com/Tisseo/navitia/blob/pr-multiple-utc-offset-bugfix/source/routing/journey_pattern_container.cpp#L140-L145

Can you give us an example of your problem with the calendar route schedule ? I think your problem is elsewhere :wink:
(maybe in our vehicle journey split over DST ?)

xavierraffin:
I must recognize that the situation of validitypattern overlaping shouldn't occured.

Maybe this issue should be fixed in an other way.

This is what I understand :

In NTFS, we put two service

```
            vj1   vj2
stop A      0H    1H
stop B      1H    2H
```

The calendar associated with the trip is 01/12 -> 31/04 (111111....)

After import I see this in kraken:

```
       vj1_dst1(GMT+1)   vj1_dst2(GMT+2)  vj2_dst1(GMT+1)  vj2_dst2(GMT+2)
stop A        0H             23H               1H                0H
stop B        1H              0H               2H                1H
```

And in overtake, there is a colission (so I suppose that calendar overlapse).

I see this in route_schedules : (the vj1_dst1 appear twice)

```
{
  "schedule": [
    {
      "vj": "vehicle_journey:vj1_dst1",
      "dt": "000000"
    },
    {
      "vj": "vehicle_journey:vj1_dst1",
      "dt": "000000"
    },
    {
      "vj": "vehicle_journey:vj2_dst1",
      "dt": "000100"
    }
  ],
  "stop": "stop_point:A"
}
```

This is because vehicle_journey:vj1_dst1 is located on two autogenerated route.

I don't now exactly how it happens, but it is.
And if I fix "overtake" function with my patch it works.

I don't have so much time to investigate, but maybe do you have an idea about this ?
(I don't provide GTFS, because my real use case is a lot more complicated with all our network and I don't have time to reproduce on a simple GTFS)

antoine-de:
It's very strange, maybe there is a bug in the dst split ?

If there is no associated dataset it's a bit more complicated to analyse :confused:

xavierraffin:
The workaround is working for me.

Let this PR open, I'll try to create a small custom dataset to reproduce when I will have more time.

TeXitoi:
Open since too long, no solution, maybe no problem, then I close it

PR #1314, opened 2015-12-03
ed: unset unsound pickup and dropoff on borders

TeXitoi:
http://jira.canaltp.fr/browse/NAVITIAII-1940

At the same time, fix docker_tests


antoine-de:
OK for me

PR #1315, opened 2015-12-04
RT: add messages in stoptimes

antoine-de:
model changes to store a message at the stoptime level.

read also the message at the trip_update level.

### DIFF
--- a/source/kraken/realtime.cpp
+++ b/source/kraken/realtime.cpp
@@ -169,7 +169,7 @@ base_execution_period(const boost::gregorian::date& date, const nt::MetaVehicleJ
 static type::disruption::Message create_message(const std::string& str_msg) {
     type::disruption::Message msg;
     msg.text = str_msg;
-    //msg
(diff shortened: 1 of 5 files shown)


TeXitoi:
OK for me.

source/kraken/realtime.cpp
```diff
@@ -164,6 +166,14 @@ base_execution_period(const boost::gregorian::date& date, const nt::MetaVehicleJ
     return {boost::posix_time::ptime{date}, boost::posix_time::ptime{date}};
 }
 
+static type::disruption::Message create_message(const std::string& str_msg) {
+    type::disruption::Message msg;
+    msg.text = str_msg;
+    //do we want to add channel types ?
```
pbougue: I guess for now we don't, so we'll just see it when we need it.
This comment line deserves a 'TODO' or to be removed (I prefer removal)

pbougue:
Ok for me

PR #1316, opened 2015-12-04
RT:  impacts concerning a VJ

pbougue:
Also corrects bug on RT and test (delay a vj after it's base-schedule end, then cancel it)

### DIFF
--- a/source/type/type.cpp
+++ b/source/type/type.cpp
@@ -354,24 +354,27 @@ bool ValidityPattern::uncheck2(unsigned int day) const {
 template <typename F>
 static bool concerns_base_at_period(const VehicleJourney& vj,
                                     const std::vector<bt::time_period>& periods,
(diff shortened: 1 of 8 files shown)


source/type/type.cpp
```diff
@@ -756,6 +762,41 @@ type::hasOdtProperties Route::get_odt_properties() const{
...
+            return false; // we don't need to carry on when we find a day concerned
+        };
+        if (concerns_base_at_period(*this, impact_sptr->application_periods, vp_functor, false)) {
+            result.push_back(impact_sptr);
+        }
+    }
+    return result;
+}
+
+std::vector<idx_t> VehicleJourney::get_impacts_idx(const PT_Data& data) const {
```
antoine-de: can't we share the code with HasMessage.get_impacts_idx ?

PR #1334, opened 2015-12-15
type: fix a segfault in get_impacts when we don't have associated base vj

TeXitoi:
http://jira.canaltp.fr/browse/NAVITIAII-1982


source/type/type.cpp
```diff
@@ -358,9 +358,10 @@ static bool concerns_base_at_period(const VehicleJourney& vj,
             if (*titr < concerned_vp.beginning_date) { continue; }
             size_t day((*titr - concerned_vp.beginning_date).days());
             if (day >= concerned_vp.days.size() || ! concerned_vp.check(day)) { continue; }
+            if (*titr < concerned_vp.beginning_date) { continue; }
```
pbougue, line 361: Ooooh!

TeXitoi, line 361: I must say I didn't add this line, I just put it 2 lines higher. But I must admit I didn't see that the same line was present 3 lines upward.

PR #1340, opened 2015-12-16
experimental stuff on /journeys: splitting of pt and georef 

kinnou02:
For not breaking anything we add a new scenario, this way it's possible
to enable this new comportment at the request level.
In jormungandr every instance have two "adaptor" classes:
- planner: it will handle interaction with the PT journey planner. At this
  time only kraken is done and there is no plan for another one.
- georef: Handle all interaction with the street network system. A
  this time only kraken exist, but there is plan to replace it with a
  better routing system in a near future

Currently the experimental scenario has a lot of limitation:
- zonal TAD is not handle
- ~~only walking fallback is use~~
- ~~a lot of parameters are ignored~~
- direct path are not activated
- only crowfly are used: there is no streetnetwork section

Refactoring has be done in raptor_api, for limiting code duplication.

:warning: require https://github.com/CanalTP/navitia-proto/pull/31 to be merge before

### DIFF
--- a/source/jormungandr/tests/routing_tests_experimental.py
+++ b/source/jormungandr/tests/routing_tests_experimental.py
@@ -1,0 +1,112 @@
+# Copyright (c) 2001-2015, Canal TP and/or its affiliates. All rights reserved.
+#
+# This file is part of Navitia,
+#     the software to build cool stuff with public transport.
+#
+# Hope you'll enjoy and contribute to this project,
+#     powered by Canal TP (www.canaltp.fr).
+# Help us simplify mobility and open public transport:
+#     a non ending quest to the responsive locomotion way of traveling!
+#
+# LICENCE: This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU Affero General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+# GNU Affero General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with this program. If not, see <http://www.gnu.org/licenses/>.
+#
+# Stay tuned using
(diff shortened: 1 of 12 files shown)


antoine-de:
can you add the performance impact for some bench ?

kinnou02:
Currently it's two time slower :disappointed:

kinnou02:
jenkins, test this please

kinnou02:
Performance are back in line, the deepcopy was taking 40% of the time. Now we append the new sections then we sort them.

antoine-de:
:+1:

kinnou02:
Artemis is good!

source/jormungandr/tests/routing_tests_experimental.py
```diff
@@ -0,0 +1,112 @@
...
+
+    def test_error_on_journeys(self):
+        """ if we got an error with kraken, an error should be returned"""
+
+        query_out_of_production_bound = "journeys?from={from_coord}&to={to_coord}&datetime={datetime}"\
+            .format(from_coord="0.0000898312;0.0000898312",  # coordinate of S in the dataset
+            to_coord="0.00188646;0.00071865",  # coordinate of R in the dataset
+            datetime="20110614T080000")  # 2011 should not be in the production period
+
+        response, status = self.query_no_assert("v1/coverage/main_routing_test/" + query_out_of_production_bound)
```
antoine-de, line 80: It was be nice to have tests with bad coordinates too no ?

source/routing/raptor_api.cpp
```diff
@@ -856,6 +910,63 @@ parse_datetimes(RAPTOR& raptor,
...
+        return response;
+    }
+    auto datetime = datetimes.front();
+
+    routing::map_stop_point_duration departures;
+    routing::map_stop_point_duration arrivals;
+
+    for(const auto& origin: origins){
+        //todo check uri is valid
+        departures[SpIdx{*raptor.data.pt_data->stop_points_map[origin.uri]}] = navitia::seconds(origin.access_duration);
```
antoine-de: isn't it a bit dangerous to acces the map like that with uri from outside the kraken (even if it was the previous kraken self that got them :wink: )
I would prefer some checks

PR #1341, opened 2015-12-16
Jormun: fix next last previous links

xlqian:
http://jira.canaltp.fr/browse/NAVITIAII-1980

### DIFF
--- a/source/jormungandr/tests/ptref_tests.py
+++ b/source/jormungandr/tests/ptref_tests.py
@@ -40,10 +40,25 @@ class TestPtRef(AbstractTestFixture):
     """
     @staticmethod
     def _test_links(response, pt_obj_name):
+        # Test the validity of links of 'previous', 'next', 'last', 'first'
+        wanted_links_type = ['previous', 'next', 'last', 'first', pt_obj_name]
+        print response
         for l in response['links']:
-            if l['type'] in ['next', 'last', 'first']:
+            if l['type'] in wanted_links_type:
                 assert pt_obj_name in l['href']
-                
+
+        # Test the consistency between links
+        import difflib
+        wanted_links = [l['href'] for l in response['links'] if l['type'] in wanted_links_type]
+        if len(wanted_links) <= 1:
+            return
+        current_match = wanted_links[0]
+        for l in wanted_links[1:]:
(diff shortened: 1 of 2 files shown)


source/jormungandr/tests/ptref_tests.py
```diff
@@ -156,6 +161,8 @@ def test_line(self):
         assert line_group[0]['name'] == 'A group'
         assert line_group[0]['id'] == 'group:A'
 
+        self._test_links(response, 'lines')
```
antoine-de, line 182: isn't there a is_valid_routes (lines, ...) where you can add this ?

source/jormungandr/tests/ptref_tests.py
```diff
@@ -38,7 +38,12 @@ class TestPtRef(AbstractTestFixture):
     """
     Test the structure of the ptref response
     """
-
+    @staticmethod
+    def _test_links(response, pt_obj_name):
+        for l in response['links']:
+            if l['type'] in ['next', 'last', 'first']:
```
pbougue: don't you need to manage 'previous' also? (at least it would be nice to have one test on it)

pbougue:
It would be nice to add tests on consistency of next/previous/first/last (self?) if there are no tests on that, but that is not directly related, so...

source/jormungandr/tests/ptref_tests.py
xlqian: you're right

source/jormungandr/tests/ptref_tests.py
```diff
@@ -221,6 +245,8 @@ def test_route(self):
         assert com[0]['type'] == 'standard'
         assert com[0]['value'] == "I'm a happy comment"
 
+        self._test_links(response, 'routes')
```
pbougue, line 248: would have been nice to check on the second page of an API to see if previous is OK.

xlqian, line 248: too late :smiling_imp:

pbougue, line 248: You can still take care of my general remark on links consistency in a dedicated PR  :wink:

PR #1346, opened 2015-12-22
Timezone format for block until in tyr api updated.

Netmisa:
See with @kinnou02 
"user" table need to be without timezone to save UTC datetime.


kinnou02:
:+1:

PR #1347, opened 2015-12-23
Georef: crowfly length with speed_factor bug fix

pbougue:
Duration stored in path_items now uses speed_factor, and everything is managed as crowfly path_items
JIRA: http://jira.canaltp.fr/browse/NAVITIAII-1995

### DIFF
--- a/source/georef/street_network.cpp
+++ b/source/georef/street_network.cpp
@@ -97,8 +97,9 @@ Path StreetNetwork::get_path(type::idx_t idx, bool use_second) {
     if (! use_second) {
         result = departure_path_finder.get_path(idx);
 
-        if (! result.path_items.empty())
+        if (! result.path_items.empty()) {
             result.path_items.front().coordinates.push_front(departure_path_finder.starting_edge.projected);
+        }
     } else {
         result = arrival_path_finder.get_path(idx);
 
@@ -532,13 +533,14 @@ Path PathFinder::build_path(vertex_t best_destination) const {
     }
     reverse_path.push_back(best_destination);
 
-    return create_path(geo_ref, reverse_path, true);
+    return create_path(geo_ref, reverse_path, true, speed_factor);
 }
 
 
 Path create_path(const GeoRef& geo_ref,
                  const std::vector<vertex_t>& reverse_path,
-                 bool add_one_elt) {
+                 bool add_one_elt,
+                 double speed_factor) {
     Path p;
 
     // On reparcourt tout dans le bon ordre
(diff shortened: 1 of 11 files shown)


source/georef/street_network.cpp
```diff
@@ -573,8 +575,8 @@ Path create_path(const GeoRef& geo_ref,
         last_transport_carac = transport_carac;
         path_item.way_idx = edge.way_idx;
         path_item.transportation = transport_carac;
-        path_item.duration += edge.duration;
-        p.duration += edge.duration;
+        path_item.duration += edge.duration / speed_factor;
+        p.duration += path_item.duration;
```
antoine-de: oups you're adding to much duration here, not a good thing that the UT pass :wink:

pbougue: I don't understand your comment. The way I see it: we now compute a correct `path_item.duration` first (dealing with speed factor), then add this `path_item.duration` to path duration.
Would have been wrong to add `edge.duration`, yes, but that's not what we do.

antoine-de: nop, you're adding, at each edge, the current total path_item duration to the path. You should also add to the path `edge.duration / speed_factor`, else you'll add some edges multiple times.

pbougue: Oh I see... We have to test it on a path_item with multiple non-null edges. Don't know if it's easy to test.

xlqian:
ok for me

source/routing/tests/routing_api_test.cpp
```diff
@@ -1560,6 +1561,81 @@ BOOST_FIXTURE_TEST_CASE(biking_length_test, streetnetworkmode_fixture<normal_spe
...
+        if (resp.journeys(idx).sections().size() == 1) {
+            journey = resp.journeys(idx);
+        }
+    }
+
+    BOOST_REQUIRE_EQUAL(journey.sections_size(), 1);
+    pbnavitia::Section section = journey.sections(0);
+    BOOST_REQUIRE_EQUAL(section.type(), pbnavitia::SectionType::STREET_NETWORK);
+
+    dump_response(resp, "speed factor length test");
```
TeXitoi, line 1595: debug dump

source/routing/tests/routing_api_test.cpp
```diff
@@ -1560,6 +1561,81 @@ BOOST_FIXTURE_TEST_CASE(biking_length_test, streetnetworkmode_fixture<normal_spe
...
+
+    std::vector<pbnavitia::PathItem> path_items;
+    for (int s = 0; s < journey.sections_size() ; s++) {
```
TeXitoi: spacing

TeXitoi: and `++s` is better (even if that's the same here).

source/routing/tests/routing_api_test.cpp
pbougue, line 1595: You have to set it explicitly to `true` if you want the dump.
So useless code, but no useless trace. The point is to ease the debug, as on the other tests around.

TeXitoi, line 1595: OK, then

TeXitoi:
OK for me.

PR #1349, opened 2015-12-24
Tyr: handle raw files or zips

antoine-de:
For artemis not to need zips, tyr has to be changed.

Tyr can now handle either zips or directory

DO NOT MERGE IT YET, i've tested it only with artemis (and it works), but not with a real 'prod' env, and it not anymore the same thing

### DIFF
--- a/source/tyr/tyr/binarisation.py
+++ b/source/tyr/tyr/binarisation.py
@@ -51,6 +51,17 @@ from tyr.helper import get_instance_logger, get_named_arg
 from contextlib import contextmanager
 
 
+def unzip_if_needed(filename):
+    if not os.path.isdir(filename):
+        # if it's not a directory, we consider it's a zip, and we unzip it
+        working_directory = os.path.dirname(filename)
+        zip_file = zipfile.ZipFile(filename)
+        zip_file.extractall(path=working_directory)
+    else:
+        working_directory = filename
+    return working_directory
+
+
 def move_to_backupdirectory(filename, working_directory):
     """ If there is no backup directory it creates one in
         {instance_directory}/backup/{name}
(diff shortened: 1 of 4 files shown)


source/tyr/tyr/binarisation.py
```diff
@@ -127,10 +127,13 @@ def fusio2ed(self, instance_config, filename, job_id, dataset_uid):
 
     logger = get_instance_logger(instance)
     try:
-        working_directory = os.path.dirname(filename)
-
-        zip_file = zipfile.ZipFile(filename)
-        zip_file.extractall(path=working_directory)
+        if not os.path.isdir(filename):
```
pbougue: It's worth a utils function, no?

source/tyr/tyr/binarisation.py
```diff
@@ -127,10 +127,13 @@ def fusio2ed(self, instance_config, filename, job_id, dataset_uid):
...
+            # if it's not a directory, we consider it's a zip, and we unzip it
```
pbougue: Could we test if it ends by ".zip", or if it unzips correctly? We may allow a single file, like poi?

source/tyr/tyr/binarisation.py
```diff
@@ -127,10 +127,13 @@ def fusio2ed(self, instance_config, filename, job_id, dataset_uid):
...
+            working_directory = os.path.dirname(filename)
+            zip_file = zipfile.ZipFile(filename)
+            zip_file.extractall(path=working_directory)
+        else:
```
pbougue: not really up to date on tyr, but what if we get a dir that contains only a zip file?

pbougue:
Is it possible to test with eitri/docker_test?

source/tyr/tyr/binarisation.py
xlqian: @pbougue hmm, I think that test by trying unzip is simple enough

source/tyr/tyr/binarisation.py
```diff
@@ -166,10 +169,12 @@ def gtfs2ed(self, instance_config, gtfs_filename, job_id, dataset_uid):
 
     logger = get_instance_logger(instance)
     try:
-        working_directory = os.path.dirname(gtfs_filename)
-
-        zip_file = zipfile.ZipFile(gtfs_filename)
-        zip_file.extractall(path=working_directory)
+        if not os.path.isdir(gtfs_filename):
```
xlqian: it looks like repeated codes

source/tyr/tyr/binarisation.py
antoine-de: nop because not all ziped files end with .zip

antoine-de:
nop test tyr is a task (and not a short one) on its own.

pbougue:
Ok for me.

pbougue:
Tested by @antoine-de so I merge

PR #1354, opened 2016-01-06
Tyr: geopal and poi types were not handled without the zips

antoine-de:


pbougue:
What about osm and synonym?

antoine-de:
they cannot be included in zips

PR #1355, opened 2016-01-06
jormungandr: add a too long waiting filter in new_default

TeXitoi:
http://jira.canaltp.fr/browse/NAVITIAII-1993


pbougue:
Is it possible to add an integration test for that?
Otherwise OK for me

TeXitoi:
It'll need a new kraken mock with specific data. Or maybe it can be done on an artemis test ? (quite doable on a sncf instance)

pbougue:
If you have the courage when we add sncf instance on artemis why not

PR #1364, opened 2016-01-13
jormun: fix a bug in multithread context

kinnou02:
Some resources had their parser at the class level and not at the
instance level, so multiple request can used the same parser at the same
time. As you can guess it didn't do any good!


antoine-de:
nice catch :+1:

xlqian:
+1

pbougue:
:+1:

guikcd:
;)

PR #1371, opened 2016-01-18
Utils: maj utils sha1 for file rights

antoine-de:
we don't neet the write priviledge for reading files


pbougue:
Ok for me (SHA-1 is OK, waiting for tests)

PR #1380, opened 2016-01-20
Jormugnandr: CAR|PARK|WALKING = CAR|PARK for similar journeys

antoine-de:
also works for bss.

related to http://jira.canaltp.fr/browse/NAVITIAII-2016


TeXitoi:
OK for me.

PR #1382, opened 2016-01-20
Ed: normalize physical_modes uri

antoine-de:
Now the physical_modes uris are the same whenever they are from a GTFS or a NTFS

ie. the uri of the bus imported from the GTFS is no more `physical_mode:3` but `physical_mode:Bus`

**WARNING** this can break some API compatibility for navitia's instances loaded with GTFS (no dataset from api.navitia.io should be impacted)

Thanx @grote for the bug issue


grote:
Thanks again! I can confirm that this change solves the issue!

PR #1390, opened 2016-01-27
push unique impact into mvj's impacte_by list

xlqian:
http://jira.canaltp.fr/browse/NAVITIAII-2029

### DIFF
--- a/source/type/type.cpp
+++ b/source/type/type.cpp
@@ -506,6 +506,21 @@ std::vector<idx_t> MetaVehicleJourney::get(Type_e type, const PT_Data& data) con
     return result;
 }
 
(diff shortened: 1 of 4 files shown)


kinnou02:
:+1:

source/type/type.cpp
```diff
@@ -506,6 +506,21 @@ std::vector<idx_t> MetaVehicleJourney::get(Type_e type, const PT_Data& data) con
     return result;
 }
 
+bool MetaVehicleJourney::is_already_impacted_by(const boost::shared_ptr<disruption::Impact>& impact) {
+    auto find_impact = [&](const boost::weak_ptr<nt::disruption::Impact>& i) {
+            auto spt = i.lock();
+            return (spt) ? spt == impact : true;
```
TeXitoi: if spt is nullptr, you  want to return false, no? or you want to do your test on impact?

PR #1394, opened 2016-02-02
Jormungandr: Extract module schedule

pbougue:
also cleanup on imports and clang (cf. commits)

### DIFF
--- a/source/jormungandr/jormungandr/schedule.py
+++ b/source/jormungandr/jormungandr/schedule.py
@@ -1,0 +1,94 @@
+# coding=utf-8
+
+# Copyright (c) 2001-2014, Canal TP and/or its affiliates. All rights reserved.
+#
+# This file is part of Navitia,
+#     the software to build cool stuff with public transport.
+#
+# Hope you'll enjoy and contribute to this project,
+#     powered by Canal TP (www.canaltp.fr).
+# Help us simplify mobility and open public transport:
+#     a non ending quest to the responsive locomotion way of traveling!
+#
+# LICENCE: This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU Affero General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+# GNU Affero General Public License for more details.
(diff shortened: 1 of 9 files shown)


pbougue:
Any comment is welcome if you think it should be a larger PR.
PR on external-RT module + mixing will follow

source/jormungandr/jormungandr/schedule.py
```diff
@@ -0,0 +1,94 @@
...
+from navitiacommon import type_pb2, request_pb2
+from jormungandr.utils import date_to_timestamp
+
+
+class MixedSchedule(object):
+
+    def __init__(self, instance):
+        self.instance = instance
+
+    def __stop_times(self, request, api, departure_filter="", arrival_filter=""):
```
kinnou02, line 45: It would be nice to not pass the request dict, this way we aren't dependant of the input interface.

source/jormungandr/jormungandr/schedule.py
```diff
@@ -0,0 +1,94 @@
...
+# twitter @navitia
+# IRC #navitia on freenode
+# https://groups.google.com/d/forum/navitia
+# www.navitia.io
+
+from navitiacommon import type_pb2, request_pb2
+from jormungandr.utils import date_to_timestamp
+
+
+class MixedSchedule(object):
```
antoine-de, line 36: some class comments would be welcome

source/jormungandr/jormungandr/schedule.py
```diff
@@ -0,0 +1,94 @@
...
+            st.until_datetime = request["until_datetime"]
+        st.duration = request["duration"]
+        st.depth = request["depth"]
+        st.show_codes = request["show_codes"]
+        if "nb_stoptimes" not in request:
+            st.nb_stoptimes = 0
+        else:
+            st.nb_stoptimes = request["nb_stoptimes"]
+        if "interface_version" not in request:
+            st.interface_version = 0
```
antoine-de: not mandatory part of your PR, but I think this interface_version is not useful anymore

source/jormungandr/jormungandr/schedule.py
pbougue, line 45: would be nice indeed, but too big now, maybe in the end of feature (but I'm not sure I'll have the courage)

source/jormungandr/jormungandr/schedule.py
pbougue, line 36: Done

antoine-de:
OK for me.

is it ok for you @kinnou02 ?

PR #1395, opened 2016-02-02
kraken: periodically retry to load realtime data

kinnou02:


source/kraken/configuration.cpp
```diff
@@ -55,6 +55,8 @@ po::options_description get_options_description(const boost::optional<std::strin
                                         "enable loading of realtime data")
         ("GENERAL.kirin_timeout", po::value<int>()->default_value(60000),
                                   "timeout in ms for loading realtime data from kirin")
+        ("GENERAL.kirin_retry_timeout", po::value<int>()->default_value(5*60),
```
antoine-de: a kirin_timeout in ms and a kirin_retry_timeout in s is a bit strange no ?

source/kraken/maintenance_worker.cpp
```diff
@@ -242,7 +239,14 @@ void MaintenanceWorker::listen_rabbitmq(){
     LOG4CPLUS_INFO(logger, "start event loop");
     data_manager.get_data()->is_connected_to_rabbitmq = true;
     while(true){
+        auto now = pt::microsec_clock::universal_time();
+        //We don't want to try to load realtime data every second
+        if(now  > this->next_try_realtime_loading){
```
antoine-de: wrong spacing

source/kraken/maintenance_worker.cpp
```diff
@@ -242,7 +239,14 @@ void MaintenanceWorker::listen_rabbitmq(){
...
+            this->next_try_realtime_loading = now + pt::seconds(conf.kirin_retry_timeout());
+            this->load_realtime();
```
antoine-de, line 246: i don't understand, we don't want to try if we already have some realtime no ?

kinnou02, line 246: `load_realtime` already check that, and others things, here we just don't want to try a call each second

source/kraken/configuration.cpp
kinnou02: so, ms for everyone?

source/kraken/maintenance_worker.cpp
antoine-de, line 246: ok

source/kraken/configuration.cpp
antoine-de: as you wish

kinnou02:
Done.

kinnou02:
Jenkins, test this please.

PR #1397, opened 2016-02-02
Jormun: no from_datetime => current datetime and realtime

antoine-de:
linked to http://jira.canaltp.fr/browse/NAVP-275
- add a data_freshness parameter to the schedules API
- when no from_datetime is provided to those API, the current time is used (not 13h37 anymore), and the datafreshness (if not given) is set to realtime
- pass the realtime_level to kraken (it was hard coded before)

DO NOT merge, tests are coming

### DIFF
--- a/source/ed/build_helper.cpp
+++ b/source/ed/build_helper.cpp
@@ -624,9 +624,23 @@ void builder::build_blocks() {
     }
 }
 
- void builder::finish() {
+/*
+ * assign the last stop point a the first vj as the route's destination if none given
+ */
+void builder::fill_missing_destinations() {
+    for (auto r: data->pt_data->routes) {
+        if (r->destination) { continue; }
+        r->for_each_vehicle_journey([&r](nt::VehicleJourney& vj) {
+            if (vj.stop_time_list.empty()) { return true; }
+            r->destination = vj.stop_time_list.back().stop_point->stop_area;
+            return false; // we stop at the first
+        });
+    }
+}
 
+void builder::finish() {
      build_blocks();
+     fill_missing_destinations();
      for(navitia::type::VehicleJourney* vj : this->data->pt_data->vehicle_journeys) {
          if (vj->stop_time_list.empty()) {
              continue;
--- a/source/jormungandr/jormungandr/interfaces/v1/Schedules.py
+++ b/source/jormungandr/jormungandr/interfaces/v1/Schedules.py
@@ -146,10 +146,10 @@ class Schedules(ResourceUri, ResourceUtc):
 
         if not args['data_freshness']:
             # The data freshness depends on the API
-            # for route/stop schedule, by default we want the base schedule
-            if self.endpoint in ('route_schedules', 'departure_boards'):
+            # for route_schedule, by default we want the base schedule
+            if self.endpoint == 'route_schedules':
                 args['data_freshness'] = 'base_schedule'
-            # for previous/next departure/arrival, we want the freshest data by default
+            # for stop_schedule and previous/next departure/arrival, we want the freshest data by default
             else:
                 args['data_freshness'] = 'realtime'
 
--- a/source/jormungandr/jormungandr/interfaces/v1/Schedules.py
+++ b/source/jormungandr/jormungandr/interfaces/v1/Schedules.py
@@ -89,6 +89,14 @@ class Schedules(ResourceUri, ResourceUtc):
                                 description="Distance range of the query. Used only if a coord is in the query")
         parser_get.add_argument("show_codes", type=boolean, default=False,
                             description="show more identification codes")
+        #Note: no default param for data freshness, the default depends on the API
+        parser_get.add_argument("data_freshness",
+                                description='freshness of the data. '
+                                            'base_schedule is the long term planned schedule. '
+                                            'adapted_schedule is for planned ahead disruptions (strikes, '
+                                            'maintenances, ...). '
+                                            'realtime is to have the freshest possible data',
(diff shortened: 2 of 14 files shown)


source/jormungandr/jormungandr/interfaces/v1/Schedules.py
```diff
@@ -115,25 +123,35 @@ def get(self, uri=None, region=None, lon=None, lat=None):
...
+        # we save the original datetime for debuging purpose
+        args['original_datetime'] = args['from_datetime']
+        if args['from_datetime']:
             args['from_datetime'] = utils.date_to_timestamp(args['from_datetime'])
+        if args['until_datetime']:
+            args['until_datetime'] = utils.date_to_timestamp(args['until_datetime'])
+
+        if not args['data_freshness']:
+            # The data freshness depends on the API
+            # for route/stop schedule, by default we want the base schedule
```
kinnou02: @stifoon don't we want realtime in stop_schedule too?

antoine-de:
Jenkins, test this please

source/ed/build_helper.cpp
```diff
@@ -624,9 +624,23 @@ void builder::build_blocks() {
...
+/*
+ * assign the last stop point a the first vj as the route's destination if none given
+ */
+void builder::fill_missing_destinations() {
+    for (auto r: data->pt_data->routes) {
+        if (r->destination) { continue; }
+        r->for_each_vehicle_journey([&r](nt::VehicleJourney& vj) {
+            if (vj.stop_time_list.empty()) { return true; }
+            r->destination = vj.stop_time_list.back().stop_point->stop_area;
+            return false; // we stop at the first
```
pbougue, line 636: You fill only the first route-without-destination's destination ??

pbougue, line 636: My bad

antoine-de, line 636: still those :eyes:  :stuck_out_tongue_winking_eye:

source/jormungandr/tests/departure_board_tests.py
```diff
@@ -26,38 +26,37 @@
...
 from check_utils import *
 import datetime
 
 
-def check_departure_board(schedules, tester, only_time=False):
+def is_valid_stop_schedule(schedules, tester, only_time=False):
     """
     check the structure of a departure board response
     """
-    assert len(schedules) == 1, "there should be only one elt"
```
pbougue, line 38: there can be more than one element, now?

source/jormungandr/tests/departure_board_tests.py
```diff
@@ -26,38 +26,37 @@
...
-
-    datetimes = get_not_null(schedule, "date_times")
+        assert len(datetimes) != 0, "we have to have date_times"
+        for dt_wrapper in datetimes:
+            dt = dt_wrapper["date_time"]
+            if only_time:
+                get_valid_time(dt)
+            else:
+                get_valid_datetime(dt)
+            #TODO remove href_mandatory=False after link refactor, they should always be there :)
```
pbougue, line 55: maybe prepare that PR?

source/jormungandr/tests/departure_board_tests.py
pbougue, line 38: Ok I see, fine :)

antoine-de, line 38: this was some value check in a method that should only check the structure of the response
I thus moved this check in a specific test and let the structure check be generic

source/jormungandr/tests/departure_board_tests.py
antoine-de, line 55: this is an old comment (the function has just been moved)

And the PR cannot be prepared just like that. A link refactoring would be quite a big refactoring, and it will be difficult not break the API during this (so my guess is that it will wait for the v2 :smile: )

source/jormungandr/tests/departure_board_tests.py
```diff
@@ -355,3 +361,174 @@ def test_vj_comment(self):
...
+                                                      SchedDT(dt='20160101T100700',
+                                                              vj='A:vj2:modified:0:delay_vj2'),
+                                                      SchedDT(dt='20160101T110700',
+                                                              vj='A:vj3:modified:0:delay_vj3')]),
+                             StopSchedule(sp='S1', route='B:1',
+                                          date_times=[SchedDT(dt='20160101T113000', vj='B:vj1')])])
+
+    def test_stop_schedule_realtime(self):
+        """
+        test a stop schedule without a from_datetime parameter
```
pbougue: comment is not accurate I guess. You check that all is OK before testing without dt, do you?

antoine-de: yes you're right, thx!

source/time_tables/tests/departure_boards_test.cpp
```diff
@@ -484,3 +484,61 @@ BOOST_FIXTURE_TEST_CASE(test_journey, calendar_fixture) {
     //we must have a journey
     BOOST_REQUIRE_EQUAL(res1.size(), 1);
 }
+
+BOOST_FIXTURE_TEST_CASE(base_stop_schedule, departure_board_fixture) {
```
pbougue, line 491: minor: add a comment?

pbougue:
Ok for me.

source/time_tables/tests/departure_boards_test.cpp
antoine-de, line 491: don't you think it's self explained ?

but if you want I can add one

PR #1400, opened 2016-02-04
jormungandr: fix min_nb_journeys in the scenario new_default

TeXitoi:
fix http://jira.canaltp.fr/browse/NAVITIAII-2031

### DIFF
--- a/source/jormungandr/jormungandr/scenarios/new_default.py
+++ b/source/jormungandr/jormungandr/scenarios/new_default.py
@@ -149,6 +149,10 @@ def create_pb_request(requested_type, request, dep_mode, arr_mode):
     return req
 
 
+def _has_pt(j):
+    return any(s.type == response_pb2.PUBLIC_TRANSPORT for s in j.sections)
+
+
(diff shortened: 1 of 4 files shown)


source/jormungandr/jormungandr/scenarios/new_default.py
```diff
@@ -158,10 +162,16 @@ def create_next_kraken_request(request, responses):
     """
     one_minute = 60
     if request['clockwise']:
-        soonest_departure = min(j.departure_date_time for r in responses for j in r.journeys)
+        try:
+            soonest_departure = min(j.departure_date_time for r in responses for j in r.journeys if _has_pt(j))
```
kinnou02, line 166: If there is only non pt solution we won't do any next call (is it bad?) but we will raise some error since soonest_departure won't be define when calculating new_datetime

TeXitoi, line 166: We don't do any next call, but we return None, and then exit the loop if request is None.

pbougue:
Ok for me

PR #1401, opened 2016-02-04
Jormungandr: add a trip name

antoine-de:
which is for the sncf the headsign

Note: nav-proto sha1 need to be updated


kinnou02:
:+1:

pbougue:
Ok for me

antoine-de:
navitia proto updated

PR #1404, opened 2016-02-05
Jormugandr: load instance configuration as json

antoine-de:
to be able to pass more instance configuration (like realtime proxies),
.ini files are a bit too tight.

change those to .json files

here is the proposed json format:

``` json
{
    "key": "default",
    "zmq_socket": "ipc:///tmp/default_kraken",
    "realtime_proxies": [
        {
            "id": "code_du_systeme",
            "class": "jormungandr.realtime_schedule.timeo.Timeo",
            "args": {
                "service_url": "http://bob.com"
            }
        }
    ]
}

```

if the format is OK I'll change the .ini generated in the tests

[Review on Reviewable](https://reviewable.io/reviews/canaltp/navitia/1404)

### DIFF
--- a/source/jormungandr/jormungandr/__init__.py
+++ b/source/jormungandr/jormungandr/__init__.py
@@ -68,7 +68,7 @@ cache = Cache(app, config=app.config['CACHE_CONFIGURATION'])
 
 from jormungandr.instance_manager import InstanceManager
 
-i_manager = InstanceManager(ini_files=app.config.get('INI_FILE', None),
+i_manager = InstanceManager(conf_files=app.config.get('INI_FILE', None),
                             instances_dir=app.config.get('INSTANCES_DIR', None),
                             start_ping=app.config.get('START_MONITORING_THREAD', True))
 i_manager.initialisation()
--- a/source/jormungandr/jormungandr/__init__.py
+++ b/source/jormungandr/jormungandr/__init__.py
@@ -68,8 +68,7 @@ cache = Cache(app, config=app.config['CACHE_CONFIGURATION'])
 
 from jormungandr.instance_manager import InstanceManager
 
-i_manager = InstanceManager(conf_files=app.config.get('INI_FILE', None),
-                            instances_dir=app.config.get('INSTANCES_DIR', None),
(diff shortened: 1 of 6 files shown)


source/jormungandr/jormungandr/__init__.py
```diff
@@ -68,7 +68,7 @@
 
 from jormungandr.instance_manager import InstanceManager
 
-i_manager = InstanceManager(ini_files=app.config.get('INI_FILE', None),
+i_manager = InstanceManager(conf_files=app.config.get('INI_FILE', None),
```
TeXitoi: We also wa't to rename INI_FILE, no?

antoine-de: I was not sure to know every use of that parameter.

I know the tests use it, but I'll need to ask @kinnou02 if it's the only use.

kinnou02: `INI_FILE` seem to be never uses:

```
% ag INI_FILE
tests/tests_mechanism.py
119:        app.config['INI_FILES'] = [
131:        i_manager.ini_files = app.config['INI_FILES']

jormungandr/__init__.py
71:i_manager = InstanceManager(ini_files=app.config.get('INI_FILE', None),
```

source/jormungandr/jormungandr/instance_manager.py
```diff
@@ -115,14 +115,25 @@ def initialisation(self):
...
-            instance = Instance(self.context, conf.get('instance', 'key'))
-            instance.socket_path = conf.get('instance', 'socket')
+            if file_name.endswith('.ini'):
+                # Note: the ini configuration file is kept only temporarily, to migration all the
+                # production configuration slowly
+                conf = ConfigParser.ConfigParser()
+                conf.read(file_name)
+                instance = Instance(self.context, conf.get('instance', 'key'), conf.get('instance', 'socket'))
+            elif file_name.endswith('.json'):
+                config_data = json.loads(open(file_name).read())
```
kinnou02: It would be nice to close that file descriptor don't you think?

PR #1405, opened 2016-02-05
Ptref: add a period filter on /disruptions

antoine-de:
now since and until filter in the impacts application_periods

you can thus do `api.navitia.io/disruptions?since=20160205T1615`

to get all disruptions that have an application period after the since

we return the disruptions that have a non null intersection between one of ther application period and the [since, until] period

### DIFF
--- a/source/ptreferential/ptreferential.cpp
+++ b/source/ptreferential/ptreferential.cpp
@@ -394,9 +394,44 @@ static bool keep_vj(const nt::VehicleJourney* vj,
 
 static std::vector<idx_t>
 filter_vj_on_period(const std::vector<type::idx_t>& indexes,
-                    const boost::optional<bt::ptime>& since,
-                    const boost::optional<bt::ptime>& until,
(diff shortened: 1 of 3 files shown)


source/ptreferential/ptreferential.cpp
```diff
@@ -394,10 +394,45 @@ static bool keep_vj(const nt::VehicleJourney* vj,
...
+}
+
+static std::vector<idx_t>
+filter_impact_on_period(const std::vector<type::idx_t>& indexes,
+                    const  bt::time_period& period,
+                    const type::Data& data) {
+
+    std::vector<idx_t> res;
+    for (const idx_t idx: indexes) {
+        auto impact = data.pt_data->disruption_holder.get_weak_impacts()[idx].lock();
```
xlqian, line 416: shouldn't impact be tested before being de-referenced?

antoine-de, line 416: yes you're right, done

PR #1408, opened 2016-02-09
Jormun: in the interface we want to handle /coord and /coords the same way

kinnou02:


pbougue:
Don't you want to add a little test?

kinnou02:
Done.

pbougue:
Thanks!

PR #1412, opened 2016-02-12
jormungandr: api frames renamed to datasets

kadhikari:


source/jormungandr/jormungandr/interfaces/v1/converters_collection_type.py
```diff
@@ -38,7 +38,7 @@
     "pois": "poi", "poi_types": "poi_type",
     "connections": "connection", "vehicle_journeys": "vehicle_journey",
     "disruptions": "disruption", "trips": "trip", "contributors": "contributor",
-    "frames": "frame"}
+    "datasets": "frame"}
```
antoine-de, line 41: don't you want to change the value too ?

pbougue, line 41: I don't think so, cause it's pb part and we don't want to change it now.
Well we want, but we're in a hurry ;)

kadhikari, line 41: while using ptref like /datasets/id, kraken has to receive filter 'frame.uri=id'.

PR #1418, opened 2016-02-16
jormun: we want unicode everywhere and mostly for input parameters

### LINKED ISSUE #1417: Codec error when utf8 characters in filter
Hello,

We talked a little bit about it with @stifoon and @kinnou02 this morning on the phone.
If I do a request on api.navitia.io for the line 22bis with a filter on the code (https://api.navitia.io/v1/coverage/fr-idf/lines?filter=line.code=22bis&depth=0) I have a result : 

``` javascript
{
    code: "22bis",
    name: "22bis Marché",
    closing_time: "112500",
    opening_time: "083500",
    id: "line:OIF:067067022:22bisOIF376"
}
```

But if I want to filter on the name (https://api.navitia.io/v1/coverage/fr-idf/lines?filter=line.name="22bis%20March%C3%A9"&depth=0) I have an error :

``` javascript
{
    message: "'ascii' codec can't encode character u'\xe9' in position 21: ordinal not in range(128)"
}
```

I have the same problem on my local navitia instance, mainly when I want to filter by network name, since our network is called Tisséo. I use python 2.7.
In https://github.com/CanalTP/navitia/blob/dev/source/jormungandr/jormungandr/interfaces/v1/Uri.py#L106 I tried to change : 

``` python
if is_collection:
    parser.add_argument("filter", type=str, default="", description="The filter parameter")
```

To : 

``` python
if is_collection:
    parser.add_argument("filter", type=unicode, default="", description="The filter parameter")
```

It seems to work, I just need to use double quotes, but I think it's normal :
http://<server>/v1/coverage/tisseo/lines?filter=network.name=Tisséo => doesn't work (Unable to parse :Filter: Unable to parse the whole string. Not parsed: >>éo<<")
http://<server>/v1/coverage/tisseo/lines?filter=network.name="Tisséo" => :+1: 

I'm not sure the fix is that simple, but that is what I found :).

kinnou02:
Some test need to be done before using `from future import unicode_literals`
fix #1417

### DIFF
--- a/source/jormungandr/jormungandr/find_extrem_datetimes.py
+++ b/source/jormungandr/jormungandr/find_extrem_datetimes.py
@@ -43,9 +43,9 @@ def extremes(resp, request):
         if arg != "datetime" and arg != "clockwise":
             if isinstance(request.args.get(arg), type([])):
                 for v in request.args.get(arg):
-                    query_args += arg + "=" + str(v) + "&"
+                    query_args += arg + "=" + unicode(v) + "&"
             else:
-                query_args += arg + "=" + str(request.args.get(arg)) + "&"
+                query_args += arg + "=" + unicode(request.args.get(arg)) + "&"
 
     before = after = None
     if asap_journey.arrival_date_time and asap_journey.departure_date_time:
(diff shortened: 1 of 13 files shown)


pbougue:
Ok for me when test is good

PR #1421, opened 2016-02-18
All: Improvement of isochrone and remove nm api

kinnou02:
Remove the NM journey api, it's not used, not tested and probably not even work as expected.

Each journey in the isochrone response has a link to the detail of the journey (that will compute a usual journey), most of the time this link didn't include the `from` parameter, resulting to a link to another isochrone. `datetime_represents` has also been added to the link.

The link was created without any reference to the coverage used even if it was specifically requested in the request, I also added it if it was in the original request, it will ensure that you get a similar response since the same coverage will be used.

In the isochrone response only the field to was filled, even with a reverse isochrone.... With this change there always a `from` and a `to` sections, the drawback is that the response is bigger.

Since isochrone only used stop_points the journey link make you compute a journey to a specified stop_point, it would be nice to used stop_area in the response, it will provide better response and allow us to make journey link directly to the stop_area.

### DIFF
--- a/source/jormungandr/jormungandr/interfaces/v1/Journeys.py
+++ b/source/jormungandr/jormungandr/interfaces/v1/Journeys.py
@@ -588,15 +588,6 @@ class Journeys(ResourceUri, ResourceUtc):
 
         self.method_decorators.append(complete_links(self))
 
-        # manage post protocol (n-m calculation)
-        self.parsers["post"] = deepcopy(parser_get)
-        parser_post = self.parsers["post"]
-        parser_post.add_argument("details", type=boolean, default=False, location="json")
-        for index, elem in enumerate(parser_post.args):
-            if elem.name in ["from", "to"]:
-                parser_post.args[index].type = list
-                parser_post.args[index].dest = elem.name
-            parser_post.args[index].location = "json"
 
     @add_debug_info()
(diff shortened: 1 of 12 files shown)


source/jormungandr/jormungandr/interfaces/v1/Journeys.py
```diff
@@ -291,20 +291,18 @@ def wrapper(*args, **kwargs):
...
                 del kwargs["lon"]
                 del kwargs["lat"]
             for journey in objects[0]['journeys']:
-                if "sections" not in journey:
+                if "sections" not in journey:#this mean it's an isochrone...
                     kwargs["datetime"] = journey["requested_date_time"]
                     kwargs["to"] = journey["to"]["id"]
+                    kwargs["from"] = journey["from"]["id"]
+                    if 'datetime_represents' in request.args:
+                        kwargs['datetime_represents'] = request.args['datetime_represents']
```
antoine-de: I think it would be easier to just deepcopy the request.args dict and just add a `to` or a `from` if there is none

TeXitoi:
Contrary to the description, kraken code is removed. Is it intentional ?

Else, OK for me.

kinnou02:
I forgot to edit the description, yes it is intentional.

PR #1422, opened 2016-02-18
kraken: optimization in fill_section

azime:
ptref is not optimized.


xlqian:
A little explanation about this optimisation is welcome :)
Ok for me

PR #1424, opened 2016-02-18
RT: add base departure/arrival in journey

pbougue:
JIRA: http://jira.canaltp.fr/browse/NAVP-342

### DIFF
--- a/source/routing/raptor_api.cpp
+++ b/source/routing/raptor_api.cpp
@@ -288,7 +288,7 @@ static bt::ptime get_base_dt(const nt::StopTime* st_orig, const nt::StopTime* st
     auto validity_pattern_dt_day = dt_orig.date();
     validity_pattern_dt_day -= boost::gregorian::days(st_orig->vehicle_journey->shift);
     auto hour_of_day_orig = (is_departure ? st_orig->departure_time : st_orig->arrival_time);
-    validity_pattern_dt_day -= boost::gregorian::days(hour_of_day_orig / (60*60*24));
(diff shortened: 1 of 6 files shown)


pbougue:
Check submodule before merge

source/routing/raptor_api.cpp
```diff
@@ -253,6 +253,37 @@ static void add_direct_path(PbCreator& pb_creator,
...
+        auto logger = log4cplus::Logger::getInstance("log");
+        LOG4CPLUS_DEBUG(logger, "Ignored stop_time finding: impossible to match exactly one base stop_time");
+        return nullptr;
+    }
+    return st_base;
+}
+
+static bt::ptime get_base_dt(const nt::StopTime* st_orig, const nt::StopTime* st_base,
+                             const bt::ptime& dt_orig, bool is_departure) {
+    auto validity_pattern_dt_day = dt_orig.date();
```
antoine-de, line 288: a bit of comments here would be nice

source/routing/raptor_api.cpp
```diff
@@ -253,6 +253,37 @@ static void add_direct_path(PbCreator& pb_creator,
...
+    validity_pattern_dt_day -= boost::gregorian::days(st_orig->vehicle_journey->shift);
+    auto hour_of_day_orig = (is_departure ? st_orig->departure_time : st_orig->arrival_time);
+    validity_pattern_dt_day -= boost::gregorian::days(hour_of_day_orig / (60*60*24));
```
antoine-de: it's better to use SECONDS_PER_DAY

TeXitoi:
OK for me with tests.

PR #1425, opened 2016-02-19
jormun: instrument send_and_receive with newrelic

kinnou02:


kinnou02:
it's possible to do the same thing with only configuration

PR #1429, opened 2016-02-23
Jormungandr: night_bus_filter_max_factor is a float

pbougue:
JIRA: http://jira.canaltp.fr/browse/NAVITIAII-2056

### DIFF
--- a/source/jormungandr/jormungandr/interfaces/v1/Journeys.py
+++ b/source/jormungandr/jormungandr/interfaces/v1/Journeys.py
@@ -577,7 +577,7 @@ class Journeys(ResourceUri, ResourceUtc):
 
         parser_get.add_argument("_walking_transfer_penalty", type=int)
         parser_get.add_argument("_night_bus_filter_base_factor", type=int)
-        parser_get.add_argument("_night_bus_filter_max_factor", type=float_gt_0)
+        parser_get.add_argument("_night_bus_filter_max_factor", type=float)
         parser_get.add_argument("_min_car", type=int)
         parser_get.add_argument("_min_bike", type=int)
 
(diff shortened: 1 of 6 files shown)


source/jormungandr/jormungandr/interfaces/v1/Journeys.py
```diff
@@ -577,7 +577,7 @@ def __init__(self):
 
         parser_get.add_argument("_walking_transfer_penalty", type=int)
         parser_get.add_argument("_night_bus_filter_base_factor", type=int)
-        parser_get.add_argument("_night_bus_filter_max_factor", type=int)
+        parser_get.add_argument("_night_bus_filter_max_factor", type=float_gt_0)
```
TeXitoi: We should manage a 0 value. A test can be great.

TeXitoi:
OK for me.

source/jormungandr/tests/routing_tests.py
```diff
@@ -307,6 +307,34 @@ def test_journeys_wheelchair_profile(self):
...
+
+        query = "journeys?from={from_coord}&to={to_coord}&datetime={d}&" \
+                         "_night_bus_filter_max_factor={_night_bus_filter_max_factor}"\
+            .format(from_coord=s_coord, to_coord=r_coord, d="20120614T08",
+                    _night_bus_filter_max_factor=2.8)
+
+        response = self.query_region(query)
+        is_valid_journey_response(response, self.tester, query)
+
+        #and the second should be 0 initialized
```
antoine-de: I don't understand this comment

source/jormungandr/tests/routing_tests.py
```diff
@@ -307,6 +307,34 @@ def test_journeys_wheelchair_profile(self):
...
+        journeys = get_not_null(response, "journeys")
+        assert journeys[0]["requested_date_time"] == "20120614T080000"
```
antoine-de: I don't also understand this assert, what do you want to check ?

PR #1438, opened 2016-03-01
Type: fix circular reference on vj/datasets

antoine-de:
BOOST_CLASS_EXPORT_IMPLEMENT is brokne in boost 158, so we trick boost
serialize not to use it

[Review on Reviewable](https://reviewable.io/reviews/canaltp/navitia/1438)

### DIFF
--- a/source/type/type.h
+++ b/source/type/type.h
@@ -319,7 +319,7 @@ struct Dataset : public Header, Nameable{
     std::vector<VehicleJourney*> vehiclejourney_list;
 
     template<class Archive> void serialize(Archive & ar, const unsigned int ) {
-        ar & idx & uri & contributor & realtime_level & validation_period & desc & system & vehiclejourney_list;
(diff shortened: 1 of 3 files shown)


source/type/type.h
```diff
@@ -563,14 +563,30 @@ struct VehicleJourney: public Header, Nameable, hasVehicleProperties {
     std::vector<boost::shared_ptr<disruption::Impact>> get_impacts() const;
 
     bool operator<(const VehicleJourney& other) const;
-    template<class Archive> void serialize(Archive& ar, const unsigned int ) {
+
+    template<class Archive> void save(Archive& ar, const unsigned int ) const {
+        ar & name & uri & route & physical_mode & company & validity_patterns
```
kinnou02, line 568: it's the same in load no? we should add a method for the serialize

antoine-de, line 568: I don't think we can since the method is const in one case, and not const in the other one

xlqian:
ok for me

PR #1439, opened 2016-03-02
navitia-proto: replace vj by route in Passage

TeXitoi:


antoine-de:
OK for me, but we'll have to be carefull with artemis

PR #1440, opened 2016-03-02
jormungandr: activate realtime proxy in departures

TeXitoi:
TODO: tests

### DIFF
--- a/source/jormungandr/jormungandr/schedule.py
+++ b/source/jormungandr/jormungandr/schedule.py
@@ -33,7 +33,7 @@ import pytz
 from jormungandr import utils
 
 from navitiacommon import type_pb2, request_pb2, response_pb2
-from jormungandr.utils import date_to_timestamp
+from jormungandr.utils import date_to_timestamp, pb_del_if
 import datetime
 from copy import deepcopy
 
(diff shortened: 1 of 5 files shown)


source/jormungandr/jormungandr/schedule.py
```diff
@@ -89,11 +89,58 @@ def update_passages(stop_schedule, next_realtime_passages):
...
+    del template.pt_display_informations.headsigns[:]
+    template.stop_date_time.ClearField("arrival_date_time")
+    template.stop_date_time.ClearField("departure_date_time")
+    template.stop_date_time.ClearField("base_arrival_date_time")
+    template.stop_date_time.ClearField("base_departure_date_time")
+    template.stop_date_time.ClearField("properties")
+
+    # filter passages with entries of the asked route_point
+    idx_to_del = []
+    for i, p in enumerate(passages):
```
antoine-de: I think it would be good to have a filter_method that takes a collection and a lambda to do this

source/jormungandr/jormungandr/schedule.py
```diff
@@ -89,11 +89,58 @@ def update_passages(stop_schedule, next_realtime_passages):
...
+        if RoutePoint(p.stop_point, p.route) == route_point:
+            idx_to_del.append(i)
+    for i in idx_to_del[::-1]:
```
antoine-de: `reversed(idx_to_del)` is not more straightforward ?

source/jormungandr/jormungandr/schedule.py
```diff
@@ -169,7 +239,31 @@ def next_arrivals(self, request):
...
-        return self.__stop_times(request, api=type_pb2.NEXT_DEPARTURES, departure_filter=request["filter"])
+        log = logging.getLogger(__name__)
+
+        resp = self.__stop_times(request, api=type_pb2.NEXT_DEPARTURES, departure_filter=request["filter"])
+        if request['data_freshness'] != 'realtime':
+            return resp
+
+        route_points = set()
+        for passage in resp.next_departures:
+            route_points.add(RoutePoint(stop_point=passage.stop_point, route=passage.route))
```
antoine-de: I think it would be clearer with

```
route_points = {RoutePoint(stop_point=passage.stop_point, route=passage.route) for passage in resp.next_departures}
```

PR #1442, opened 2016-03-02
Kraken: secure max cache use

pbougue:


pbougue:
Jenkins, test this please.

pbougue:
The cache datetime window filtering seems OK already: https://github.com/CanalTP/navitia/blob/dev/source/routing/next_stop_time.cpp#L404

PR #1444, opened 2016-03-03
kraken: show_codes = true for all APIs

azime:
Jira: http://jira.canaltp.fr/browse/NAVP-505

### DIFF
--- a/source/jormungandr/tests/routing_tests.py
+++ b/source/jormungandr/tests/routing_tests.py
@@ -407,6 +407,16 @@ class TestLongWaitingDurationFilter(AbstractTestFixture):
         eq_(response['journeys'][1]['arrival_date_time'], "20120614T160000")
         eq_(response['journeys'][1]['type'], "best")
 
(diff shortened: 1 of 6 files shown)


azime:
jenkins test this

azime:
Jenkins, test this please

source/jormungandr/tests/routing_tests.py
```diff
@@ -407,6 +407,16 @@ def test_novalidjourney_on_first_call_debug(self):
         eq_(response['journeys'][1]['arrival_date_time'], "20120614T160000")
         eq_(response['journeys'][1]['type'], "best")
 
+    def test_codes_on_object(self):
```
antoine-de: can you add a comment on the method ?

And maybe add a method that do the same test but with `show_codes=false`

azime: OK

azime:
corrections

PR #1445, opened 2016-03-03
Jormungandr: forbidden_id/_uris debug

pbougue:
JIRA: http://jira.canaltp.fr/browse/NAVITIAII-2060

### DIFF
--- a/source/jormungandr/jormungandr/interfaces/v1/Schedules.py
+++ b/source/jormungandr/jormungandr/interfaces/v1/Schedules.py
@@ -81,7 +81,7 @@ class Schedules(ResourceUri, ResourceUtc):
                                 stop_point/route")
         parser_get.add_argument("forbidden_id[]", type=unicode,
                                 description="DEPRECATED, replaced by forbidden_uris[]",
-                                dest="forbidden_id[]",
+                                dest="__temporary_forbidden_id[]",
                                 default=[],
                                 action='append')
         parser_get.add_argument("forbidden_uris[]", type=unicode,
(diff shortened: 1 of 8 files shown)


source/jormungandr/jormungandr/interfaces/v1/Schedules.py
```diff
@@ -80,13 +80,16 @@ def __init__(self, endpoint):
                                 description="Maximum number of schedule per\
                                 stop_point/route")
         parser_get.add_argument("forbidden_id[]", type=unicode,
-                                description="forbidden ids",
-                                dest="forbidden_uris[]",
-                                action="append")
+                                description="DEPRECATED, replaced by forbidden_uris[]",
+                                dest="forbidden_id[]",
```
antoine-de: Minor: I would have name this temporary list with a less usual name (`__temporary_id` ?)  to make sure it is not used in the other part of the code (it's somewhat difficult to know which one to use between forbidden_id and forbidden_uri :wink: )

source/jormungandr/jormungandr/interfaces/v1/Schedules.py
```diff
@@ -109,6 +112,11 @@ def __init__(self, endpoint):
 
     def get(self, uri=None, region=None, lon=None, lat=None):
         args = self.parsers["get"].parse_args()
+
+        # for retrocompatibility purpose
+        for forbid_id in args.get('forbidden_id[]', []):
+            args.get('forbidden_uris[]', []).append(forbid_id)
```
antoine-de: minor: you can do `args['forbidden_uris[]']` (and same above) because I think flask add by default all params in args

It's better because if it is not a valid flask argument you'll get an error

TeXitoi:
OK for me.

pbougue:
Jenkins, test this please.

PR #1449, opened 2016-03-04
test for the PR https://github.com/CanalTP/navitia/pull/1447

kinnou02:
It was caused by "complex" anticlockwise journey with vjs that don't have the same validity pattern but these vjs have extensions between them.
The reader didn't check the returned value of next_stoptime


pbougue:
Ok for me

PR #1459, opened 2016-03-07
Jormungandr: Departures, arrivals without filter

azime:
Jira: http://jira.canaltp.fr/browse/NAVITIAII-2025


source/jormungandr/jormungandr/interfaces/v1/Schedules.py
```diff
@@ -121,6 +122,8 @@ def get(self, uri=None, region=None, lon=None, lat=None):
         args["interface_version"] = 1
 
         if uri is None:
+            if not args['filter']:
+                raise InvalidArguments('filter')
```
antoine-de, line 126: I think a better error would be nice

edit: my bad, I just saw that the invalid argument gives a good error message :smile:

PR #1465, opened 2016-03-10
jormun: first batch of compat python 3 without breaking anything (hopefuly)!

kinnou02:
There is no ugly hack, we only use future for compatibility.
The preferred syntax is python 3. But there is some problems with
unicode.

This PR doesn't work with python 3, it's just preparatory work.


buildbotctp:
Build finished. 1021 tests run, 0 skipped, 1 failed.

source/jormungandr/jormungandr/interfaces/v1/Uri.py
```diff
@@ -76,7 +78,7 @@ def __init__(self, is_collection, collection, *args, **kwargs):
                             default=[],
                             action="append")
         parser.add_argument("forbidden_uris[]", type=unicode,
-                            description="forbidden uris",
+                            description="forbidden ids",
```
antoine-de: uris

source/jormungandr/jormungandr/scenarios/journey_filter.py
```diff
@@ -26,13 +26,17 @@
...
 import logging
 import itertools
 import datetime
 from jormungandr.scenarios.utils import compare, get_pseudo_duration
 from navitiacommon import response_pb2
 from jormungandr.utils import pb_del_if
+from builtins import range, zip
+
+#we can't use reverse(enumerate(list)) without creating a temporary list, so we define our own reverse enumerate
+reverse_enumerate = lambda l: zip(range(len(l)-1, -1, -1), reversed(l))
```
antoine-de: where is the old one ? maybe it's a merge problem

antoine-de:
OK for me

source/jormungandr/jormungandr/scenarios/journey_filter.py
TeXitoi: I move it not so long ago in jormungandr/jormungandr/utils.py

PR #1467, opened 2016-03-10
jormungandr: manage empty route point in proxy rt for departures

TeXitoi:


source/time_tables/passages.cpp
```diff
@@ -0,0 +1,194 @@
+/* Copyright © 2001-2014, Canal TP and/or its affiliates. All rights reserved.
```
antoine-de: a bit old :)

source/time_tables/passages.cpp
```diff
@@ -0,0 +1,194 @@
...
+
+namespace pt = boost::posix_time;
+using navitia::routing::StopEvent;
+
+namespace navitia { namespace timetables {
+
+namespace {
+struct PassagesVisitor {
+    pbnavitia::API api_pb;
+    const type::Data &data;
```
antoine-de: :sob:

source/time_tables/passages.cpp
```diff
@@ -0,0 +1,194 @@
...
+        case pbnavitia::PREVIOUS_DEPARTURES:
+            return StopEvent::pick_up;
+        case pbnavitia::NEXT_ARRIVALS:
+        case pbnavitia::PREVIOUS_ARRIVALS:
+            return StopEvent::drop_off;
+        default:
+            throw std::logic_error("bad api_pb in passages");
+        }
+    }
+    bool is_useless(const routing::JppIdx& jpp_idx) const {
```
antoine-de, line 76: is_useless ? I think a comment would be nice

antoine-de:
OK for me

PR #1470, opened 2016-03-11
routing: optimize memory of CachedNextStopTime

TeXitoi:
Performance:
- time: (small gain) (ms/request)
  - stif: 99 -> 94
  - ch: 490 -> 440
- memory: (improve 1.5 for stif, 1.9 for ch) (Mo/cache)
  - stif: [147-254], avg 202 -> [47-212], avg 130
  - ch: [362-396], avg 389 -> [129-368], avg 209

http://jira.canaltp.fr/browse/NAVP-511

### DIFF
--- a/source/routing/next_stop_time.h
+++ b/source/routing/next_stop_time.h
@@ -230,10 +230,18 @@ struct CachedNextStopTime {
 
 private:
     struct DtStFromJpp {
+        DtStFromJpp(const JppIdxMap& jpp_idx_map);
+        // Returns the range corresponding to jpp_idx_map[jpp_idx],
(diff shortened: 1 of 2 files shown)


source/routing/next_stop_time.h
```diff
@@ -226,6 +227,16 @@ struct CachedNextStopTime {
                    const JppIdx jpp_idx,
                    const DateTime dt,
                    const bool clockwise) const;
+
+private:
+    struct DtStFromJpp {
```
antoine-de, line 234: I think it would be nice to have an ASCII art to explain the structure
(and maybe a reminder to explain the range creation in DtStFromJpp::operator[])

PR #1472, opened 2016-03-11
Jormungandr: add cache on timeo

antoine-de:
Add a cache on timeo's reponse

Important note: we now deactivate cache by default. (else the tests uses the cache :smile:  )


kinnou02:
:+1:

PR #1473, opened 2016-03-11
RT: enhance schedules API

kadhikari:


source/jormungandr/tests/proxy_realtime_tests.py
```diff
@@ -148,3 +150,50 @@ def test_departures(self):
...
+        query = 'stop_areas/S42/departures?from_datetime=20160102T1000&show_codes=true&count=7'
+        response = self.query_region(query)
+
+        assert "departures" in response
+        assert len(response["departures"]) == 7
+
+        dep_data_freshness = [(d['route']['name'], d['stop_date_time']['data_freshness'])
+                      for d in response['departures']]
+        expected_data_freshness = [
+            ("J", "realtime"),
```
antoine-de: I think we can mutualize some code with my PR https://github.com/CanalTP/navitia/pull/1471

https://github.com/CanalTP/navitia/pull/1471/files#diff-a60f1454a9e1e2aa1d57f92ff5ea84c9R153

source/time_tables/passages.cpp
```diff
@@ -176,9 +177,19 @@ void passages(PbCreator& pb_creator,
...
         passage->mutable_stop_date_time()->set_departure_date_time(departure_date);
         passage->mutable_stop_date_time()->set_arrival_date_time(arrival_date);
+
+        //find base datetime
+        auto base_st = get_base_stop_time(dt_stop_time.second);
+        if (base_st != nullptr) {
+            auto base_ptime = navitia::to_posix_time(dt_stop_time.first, pb_creator.data);
+            auto base_dep_dt = get_base_dt(dt_stop_time.second, base_st, base_ptime, true);
+            passage->mutable_stop_date_time()->set_base_departure_date_time(navitia::to_posix_timestamp(base_dep_dt));
+            passage->mutable_stop_date_time()->set_base_arrival_date_time(navitia::to_posix_timestamp(base_dep_dt));
```
antoine-de, line 187: you need the arrival there no ?

kadhikari, line 187: In the vector we don't have arrival and departure values so same datetime used for both.

source/jormungandr/jormungandr/schedule.py
```diff
@@ -41,6 +41,7 @@
 from copy import deepcopy
 
 RT_PROXY_PROPERTY_NAME = 'realtime_system'
+RT_PROXY_DATA_FRESHNESS = 'real_time'
```
kinnou02: realtime in a single word, and I don't see where this is used

kadhikari:
us concerned: http://jira.canaltp.fr/browse/NAVP-307

source/time_tables/passages.cpp
pbougue, line 187: minor: It would be nice to give it another name than departure, maybe just `base_dt`
And it's worth doing the same above with `departure_date` and `arrival date`

PR #1489, opened 2016-03-18
kraken: change to loadbalalancing system between kraken thread

### LINKED ISSUE #348: Load balancing with variations in request processing time
The broker mechanism in https://github.com/CanalTP/navitia/blob/dev/source/kraken/kraken_zmq.cpp is connecting a ROUTER socket to a DEALER socket via a ZMQ_QUEUE device. This may not be the optimal load balancing approach.

The DEALER socket's outgoing routing strategy is round-robin: one message is sent to each worker 1..N, then the DEALER starts over again at worker 1. Conceptually, if not internally, the messages are pre-assigned to a worker where they queue up. For a request/response based system (like Navitia or RRRR) this can work well enough if each work unit takes almost the same amount of time to complete. However, if there is significant variance in computation time per work unit and/or a limited number of worker processes, clients will perceive unnecessary latency when their request is stuck in line behind a slow calculation on their assigned worker. In this case, there is an advantage to using a load-balancing pattern where the workers signal the broker that they are available, and work units are handed out to the workers in the order that the workers become available.

By way of analogy, compare a store where separate queues form at each cash register (and there is no 10-items or less register), versus a public administration or railway ticket office where an agent is dynamically assigned to each client as the agents become available (using paper tickets and screens). The second option gives clients faster service when some operations take much longer than others. 

Note 1: I am more familiar with the ZMQ C API, so it is possible you have already accounted for this and I am misunderstanding something about zmq::device.

Note 2: If there are no outlier requests that take significantly longer than the mean processing time, then there is no need for a more sophisticated load balancing mechanism.

kinnou02:
With the previous system (zmq_proxy) request where staticaly associated
to a thread. With this system incomming request are affected to the
first availaible thread.
The mechanism used is describe here: http://zguide.zeromq.org/php:chapter3#header-73
Fixes #348
require: https://github.com/CanalTP/utils/pull/40

### DIFF
--- a/source/kraken/kraken_zmq.cpp
+++ b/source/kraken/kraken_zmq.cpp
@@ -38,7 +38,7 @@ www.navitia.io
 #include <iostream>
 #include "utils/init.h"
 #include "kraken_zmq.h"
-#include "utils/zmq_compat.h"
+#include "utils/zmq.h"
 
 
 int main(int argn, char** argv){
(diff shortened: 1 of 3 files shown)


kinnou02:
tests should'nt work :(

kinnou02:
It could be nice to do some bench for ensuring there is no regression in performance.

stifoon:
Hey, a new issue will be done :) https://github.com/CanalTP/navitia/issues/348

source/kraken/kraken_zmq.cpp
```diff
@@ -72,13 +72,12 @@ int main(int argn, char** argv){
     boost::thread_group threads;
     // Prepare our context and sockets
     zmq::context_t context(1);
```
antoine-de, line 74: why is the context outside the LoadBalancer ?

PR #1497, opened 2016-03-22
Jormungandr: next = one second after best journey

pbougue:


source/jormungandr/jormungandr/scenarios/new_default.py
```diff
@@ -159,22 +159,36 @@ def create_next_kraken_request(request, responses):
...
+    (if anticlockwise, remove one second to the arrival of the journey with latest departure, then shortest)
     """
-    one_minute = 60
+    one_second = 1
+    best_arrival = best_departure = None
     if request['clockwise']:
-        try:
-            soonest_departure = min(j.departure_date_time for r in responses for j in r.journeys if _has_pt(j))
-        except ValueError:
+        for r in responses :
```
antoine-de, line 167: seems like dupplicated code, I think it would be nice to factorize it (use a function for the purpose to be clearer ?

antoine-de:
I think it's important to test this

TeXitoi:
OK for me.

source/jormungandr/jormungandr/scenarios/tests/new_default_tests.py
```diff
@@ -247,3 +248,89 @@ def merge_responses_on_errors_test():
...
+    section.street_network.mode = response_pb2.Walking
+    section = journey.sections.add()
+    section.type = response_pb2.PUBLIC_TRANSPORT
+    section = journey.sections.add()
+    section.type = response_pb2.STREET_NETWORK
+    section.street_network.mode = response_pb2.Walking
+
+
+def create_next_kraken_request_test():
+    request_clock = {}
```
antoine-de: minor: but if you want to win some lines you can do:
`request_clock = {'datetime':100000, 'clockwise':True }`

PR #1500, opened 2016-03-24
Doc migration slate

stifoon:


documentation/slate/source/includes/welcome.md
```diff
@@ -23,20 +23,26 @@ Have a look at the examples below to learn what services we provide and how to u
...
-#### First,
+#### First step,
 get a token here <http://navitia.io/register/>
 
-#### Second,
-use the token : if you use a web browser, you only have to **paste it in the user area**,
-with **no password**. 
-Or, in a simplier way, you can add your token in the address bar like :
+#### Second step,
+Go to the API. The simpliest way is to use [firefox browser](#http://www.getfirefox.com)
```
antoine-de: maybe we should be less pushy about firefox.

eg: 

``` raw
Go to the API. The simpliest way is to use a web browser (our humble opinion is that [firefox browser](#http://www.getfirefox.com) and a json viewer extension like [JSON-DataView](#https://addons.mozilla.org/en-us/firefox/addon/json-dataview) is a good setup)
```

PR #1510, opened 2016-03-31
OSM: do not import private nameless way

antoine-de:
they will be useless, so we don't import them

linked to http://jira.canaltp.fr/browse/NAVITIAII-2074

---

This change is [Reviewable](https://reviewable.io/reviews/canaltp/navitia/1510)

### DIFF
--- a/source/ed/osm2ed.cpp
+++ b/source/ed/osm2ed.cpp
@@ -152,11 +152,16 @@ void ReadRelationsVisitor::relation_callback(uint64_t osm_id, const CanalTP::Tag
     }
 }
 
+ReadWaysVisitor::~ReadWaysVisitor() {
+    LOG4CPLUS_INFO(logger, filtered_private_way << " way filtered because they were nameless and private");
+}
+
 /*
  * We stores ways they are streets.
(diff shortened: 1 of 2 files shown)


source/ed/osm2ed.cpp
```diff
@@ -152,11 +152,16 @@ void ReadRelationsVisitor::relation_callback(uint64_t osm_id, const CanalTP::Tag
...
+    LOG4CPLUS_INFO(logger, filtered_private_way << " way filtered because they were nameless and private");
+}
+
 /*
  * We stores ways they are streets.
  * We store ids of needed nodes
  */
-void ReadWaysVisitor::way_callback(uint64_t osm_id, const CanalTP::Tags &tags, const std::vector<uint64_t> & nodes_refs) {
+void ReadWaysVisitor::way_callback(uint64_t osm_id, const CanalTP::Tags &tags,
+                                   const std::vector<uint64_t> & nodes_refs) {
```
TeXitoi, line 164: If you reformat code, be kind with & :smile:

TeXitoi:
OK for me.

PR #1513, opened 2016-04-01
BSS web service call

ooga:
mock atos webservice


source/jormungandr/jormungandr/realtime_place/bss/bss_provider_manager.py
```diff
@@ -0,0 +1,67 @@
...
+    def __init__(self):
+        self.instances = []
+        self.log = logging.getLogger(__name__)
+        for configuration in app.config["BSS_PROVIDER"]:
+            arguments = configuration.get('args', {})
+            self.instances.append(self.init_class(configuration["class"], arguments))
+
+    def handle_places(self, places):
+        if places:
+            for place in places:
```
kinnou02: Too much `if`, you can do:

```
for place in places or []:
    ...
```

source/jormungandr/jormungandr/realtime_place/bss/bss_provider_manager.py
```diff
@@ -0,0 +1,67 @@
...
+                if (place["embedded_type"] == "poi") & (place["poi"]["poi_type"]["id"] == 'poi_type:amenity:bicycle_rental'):
```
kinnou02: Parenthesis aren't required, for a logical and you should use `and`
And by the way the preferred style for string is simple quote.

source/jormungandr/jormungandr/interfaces/v1/Places.py
```diff
@@ -359,4 +360,8 @@ def get(self, region=None, lon=None, lat=None, uri=None):
         self._register_interpreted_parameters(args)
         response = i_manager.dispatch(args, "places_nearby",
                                       instance_name=self.region)
-        return response, 200
+        response = marshal(response, places_nearby)
+        if args["bss_stands"]:
+            bss_manager = BssProviderManager()
+            response["places_nearby"] = bss_manager.handle_places(response["places_nearby"])
+        return response, 200
```
pbougue: missing final newline

source/jormungandr/jormungandr/realtime_place/bss/atos.py
```diff
@@ -0,0 +1,58 @@
...
+    WS_URL_TEMPLATE = "https://{}.velossimo.com/services/terminal"
+    OPERATOR = "Keolis"
+
+    def __init__(self, id_ao, network, subdomain = "admin"):
+        self.id_ao = id_ao
+        self.network = network
+        self.WS_URL = self.WS_URL_TEMPLATE.format(subdomain)
+
+    def is_supported(self, poi):
+        if "properties" in poi:
```
pbougue: I'd rather do an early return here:

```
if "properties" not in poi:
    return false
```

Which also covers better your "else" case.

source/jormungandr/jormungandr/default_settings.py
```diff
@@ -54,6 +54,9 @@
     }
 }
 
+# Bike self-service configuration
+BSS_PROVIDER = []
```
xlqian: if `BSS_PROVIDER` is constant, you can use `()` instead of `[]`

source/jormungandr/jormungandr/realtime_place/bss/atos.py
```diff
@@ -0,0 +1,58 @@
...
+
+    WS_URL_TEMPLATE = "https://{}.velossimo.com/services/terminal"
+    OPERATOR = "Keolis"
+
+    def __init__(self, id_ao, network, subdomain = "admin"):
+        self.id_ao = id_ao
+        self.network = network
+        self.WS_URL = self.WS_URL_TEMPLATE.format(subdomain)
+
+    def is_supported(self, poi):
```
pbougue: Not fan of the name, i'd rather name it more explicitly.
No big idea, `is_connector_supported` `is_provider_supported`?

source/jormungandr/jormungandr/realtime_place/bss/atos.py
```diff
@@ -0,0 +1,58 @@
...
+        if "properties" in poi:
+            operator = False
+            network = False
+            for type, value in poi["properties"].items():
+                if type == "operator":
+                    operator = value == self.OPERATOR
```
pbougue: not a fan of `=` `==` but that's personal.
I'd rather have it parenthesized, or here in the if.

source/jormungandr/jormungandr/realtime_place/bss/atos.py
```diff
@@ -0,0 +1,58 @@
...
+                    continue
+                if type == "network":
+                    network = value.encode("utf-8") == self.network
+                    continue
+            return operator & network
```
pbougue: you use the logical op `&` which should work here but via casts, what you really mean is `and` in python

source/jormungandr/jormungandr/realtime_place/bss/atos.py
```diff
@@ -0,0 +1,58 @@
...
+        self.WS_URL = self.WS_URL_TEMPLATE.format(subdomain)
+
+    def is_supported(self, poi):
+        if "properties" in poi:
+            operator = False
+            network = False
+            for type, value in poi["properties"].items():
+                if type == "operator":
+                    operator = value == self.OPERATOR
+                    continue
```
pbougue: i'd rather have an `elif` than those `continue` here, easier to read

source/jormungandr/jormungandr/realtime_place/bss/atos.py
```diff
@@ -0,0 +1,58 @@
...
+                if type == "network":
+                    network = value.encode("utf-8") == self.network
```
pbougue: you only have encoding problems in network, not operator?
@kinnou02 what should we use for these encoding problems?

source/jormungandr/jormungandr/realtime_place/bss/tests/bss_provider_manager_test.py
```diff
@@ -0,0 +1,134 @@
...
+            "id": "poi_type:amenity:bicycle_rental"
+        }
+    }
+    app.config["BSS_PROVIDER"] = CONFIG
+    manager = BssProviderManager()
+    provider = manager.find_provider(poi)
+    assert provider == manager.instances[0]
+    poi["properties"]["operator"] = "Bad_operator"
+    provider = manager.find_provider(poi)
+    assert provider is None
```
pbougue: missing newline

source/jormungandr/jormungandr/realtime_place/bss/atos.py
antoine-de: I think the problem is more the it should be call `supports` no ? (or `concerns` ?)

with this for me it can be quite clear at the use like

`bss_provider.concerns(poi)` no ? but `concerns_poi` would be fine for me if you like more explicit stuff

source/jormungandr/jormungandr/realtime_place/bss/atos.py
antoine-de: I personally don't have a problem with this

source/jormungandr/jormungandr/realtime_place/bss/atos.py
```diff
@@ -0,0 +1,58 @@
...
+    def __init__(self, id_ao, network, subdomain = "admin"):
+        self.id_ao = id_ao
+        self.network = network
+        self.WS_URL = self.WS_URL_TEMPLATE.format(subdomain)
+
+    def is_supported(self, poi):
+        if "properties" in poi:
+            operator = False
+            network = False
+            for type, value in poi["properties"].items():
```
antoine-de: not very important and be do also this in some places, but `type` is a reserved keyword, we should try not to use it a a variable

source/jormungandr/jormungandr/realtime_place/bss/atos.py
antoine-de: I think you can rewrite your method like: (not sure it's completly valid, it's just the idea):

``` python
def concerns(self, poi):
    return properties.get('type') == self.self.OPERATOR and properties.get('network') == self.network
```

source/jormungandr/jormungandr/realtime_place/bss/bss_provider_manager.py
```diff
@@ -0,0 +1,67 @@
...
+# www.navitia.io
+from jormungandr import app
+from importlib import import_module
+import logging
+
+
+class BssProviderManager(object):
+
+    def __init__(self):
+        self.instances = []
```
antoine-de: I think `bss_providers` is a better name, there are already too much `instances` in the codes :smile:

source/jormungandr/jormungandr/realtime_place/bss/tests/atos_test.py
```diff
@@ -0,0 +1,56 @@
...
+#
+# Stay tuned using
+# twitter @navitia
+# IRC #navitia on freenode
+# https://groups.google.com/d/forum/navitia
+# www.navitia.io
+from jormungandr.realtime_place.bss.atos import AtosProvider
+
+
+def realtime_place_atos_supported_test():
```
antoine-de: :+1:

source/jormungandr/jormungandr/realtime_place/bss/tests/bss_provider_manager_test.py
```diff
@@ -0,0 +1,134 @@
...
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with this program. If not, see <http://www.gnu.org/licenses/>.
+#
+# Stay tuned using
+# twitter @navitia
+# IRC #navitia on freenode
+# https://groups.google.com/d/forum/navitia
+# www.navitia.io
+from jormungandr.realtime_place.bss.bss_provider_manager import BssProviderManager
```
antoine-de: why is it stored in jormungandr.relatime_place ?

source/jormungandr/jormungandr/realtime_place/bss/tests/bss_provider_manager_test.py
```diff
@@ -0,0 +1,134 @@
...
+        {
+            "class": "jormungandr.realtime_place.bss.BadProvider",
+            "args": {
+                "id_ao": "38",
+                "network": "Vlille"
+            }
+        }
+    ]
+    try:
+        manager = BssProviderManager()
```
antoine-de: for this you can wrap you test method with `@raises(YourExceptionClass)`

source/jormungandr/jormungandr/default_settings.py
antoine-de: it's not a constant, in the real configuration file there will be several bss providers

xlqian: yeah, so `BSS_PROVIDER` is read only...

source/jormungandr/jormungandr/realtime_place/bss/bss_provider_manager.py
```diff
@@ -0,0 +1,67 @@
...
+        if places:
+            for place in places:
+                if (place["embedded_type"] == "poi") & (place["poi"]["poi_type"]["id"] == 'poi_type:amenity:bicycle_rental'):
+                    provider = self.find_provider(place["poi"])
+                    if provider:
+                        place["poi"]["stands"] = provider.get_informations(place["poi"])
+        return places
+
+    def find_provider(self, poi):
+        for instance in self.instances:
```
xlqian: what happens if no instance support poi? 

what you want is `return next((i for i in self.instances if i.is_supported(poi)), None)`

source/jormungandr/jormungandr/default_settings.py
antoine-de: you're right

source/jormungandr/jormungandr/realtime_place/bss/bss_provider_manager.py
antoine-de: it does not matter that no provider can handle it.

source/jormungandr/jormungandr/realtime_place/bss/tests/bss_provider_manager_test.py
ooga: `jormungandr.realtime_bss.bss_provider_manager` will be better ?

source/jormungandr/jormungandr/realtime_place/bss/__init__.py
```diff
@@ -0,0 +1,30 @@
...
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with this program. If not, see <http://www.gnu.org/licenses/>.
+#
+# Stay tuned using
+# twitter @navitia
+# IRC #navitia on freenode
+# https://groups.google.com/d/forum/navitia
+# www.navitia.io
+from atos import AtosProvider
```
kinnou02: you should add: 

```
```

This way your import will fail (and your code will become compatible with python 3)

source/jormungandr/jormungandr/realtime_place/bss/atos.py
```diff
@@ -0,0 +1,69 @@
...
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with this program. If not, see <http://www.gnu.org/licenses/>.
+#
+# Stay tuned using
+# twitter @navitia
+# IRC #navitia on freenode
+# https://groups.google.com/d/forum/navitia
+# www.navitia.io
+from bss_provider import BssProvider
```
kinnou02: same here

source/jormungandr/jormungandr/realtime_place/bss/atos.py
```diff
@@ -0,0 +1,69 @@
...
+               properties.get('network', '').encode('utf-8') == self.network
+
+    def get_informations(self, poi):
+        try:
+            all_stands = self.get_all_stands()
+            ref = poi.get('properties', {}).get('ref', '')
+            stands = all_stands.get(ref)
+            if stands:
+                return stands
+        except Exception:
```
kinnou02: A catch for all `Exception` without any log? It's seems strange

source/jormungandr/jormungandr/realtime_place/bss/atos.py
```diff
@@ -0,0 +1,69 @@
...
+            pass
+        return None
+
+    def get_all_stands(self):
+        client = self.get_client()
+        all_stands = client.service.getSummaryInformationTerminals(self.id_ao)
+        result = {}
```
kinnou02: This  could be done with list (actually dict) comprehension:

```
return {stands.libelle: Stands(stands.nbPlacesDispo, stands.nbVelosDispo) for stands in all_stands}
```

source/jormungandr/jormungandr/realtime_place/bss/atos.py
```diff
@@ -0,0 +1,76 @@
...
+        self.id_ao = id_ao
+        self.network = network
+        self.WS_URL = self.WS_URL_TEMPLATE.format(subdomain)
+
+    def __repr__(self):
+        return self.WS_URL + str(self.id_ao)
+
+    def support_poi(self, poi):
+        properties = poi.get('properties', {})
+        return properties.get('operator', '') == self.OPERATOR and \
```
antoine-de: it is not necessary to put the second `get` argument, the default is `None`, and I think it's what you want

source/jormungandr/jormungandr/interfaces/v1/Places.py
```diff
@@ -359,4 +360,8 @@ def get(self, region=None, lon=None, lat=None, uri=None):
         self._register_interpreted_parameters(args)
         response = i_manager.dispatch(args, "places_nearby",
                                       instance_name=self.region)
+        response = marshal(response, places_nearby)
+        if args["bss_stands"]:
+            bss_manager = BssProviderManager()
```
antoine-de: I think you should initialize it only once no ? you can put this in **init**.py in jormungandr

source/jormungandr/jormungandr/realtime_place/bss/atos.py
```diff
@@ -0,0 +1,76 @@
...
+               properties.get('network', '') == self.network
+
+    def get_informations(self, poi):
+        try:
+            all_stands = self.get_all_stands()
+            ref = poi.get('properties', {}).get('ref', '')
```
antoine-de: same here, the second `get` argument is not mandatory:
`ref = poi.get('properties', {}).get('ref')`

source/jormungandr/jormungandr/realtime_place/bss/atos.py
```diff
@@ -0,0 +1,76 @@
...
+            stands = all_stands.get(ref)
+            if stands:
```
antoine-de: since you'll return None if stand is None, you can just return stands no ?

pbougue:
It may be nicer to call it `parking_space_availability` than `realtime_place`, as:
- it may be confusing with realtime in next_schedule or in journeys
- this information might not be realtime and still be interesting (total spots in a bss station for example)
- place may be a little confusing with available tickets on a train journey for example

source/jormungandr/jormungandr/realtime_place/bss/atos.py
```diff
@@ -0,0 +1,76 @@
...
+                return stands
+        except URLError as e:
+            logging.getLogger(__name__).info(str(e))
+        except suds.WebFault as e:
+            logging.getLogger(__name__).info('{} in document {}'.format(e.fault, e.document))
+        return None
```
xlqian: `finally` : https://docs.python.org/2/tutorial/errors.html#defining-clean-up-actions

source/jormungandr/jormungandr/interfaces/v1/Places.py
ooga: Not sure it's a good idea to initialize it when we request ```journeys``` api

xlqian:
Ok for me :) :+1:

antoine-de:
great work, OK for me, but you'll need to rebase first :wink:

ooga:
@antoine-de rebase done

PR #1521, opened 2016-04-05
Set favicon in meta tag in index.html

alcalyn:
Last PR which add a favicon.ico worked with middleman, but once deployed on `canaltp.github.io/navitia`, it is in a subfolder, so it needs a meta tag in index.html.


antoine-de:
great, thx!

PR #1522, opened 2016-04-05
fix empty timezone in /places when disruptions are presents

xlqian:
http://jira.canaltp.fr/browse/NAVITIAII-2088


kinnou02:
Have you checked all the others API?

xlqian:
@kinnou02 
`/journeys` is ok, `/lines` is ok, `/networks` is ok, `/stop_areas` is ok, `/stop_points` is ok, `/vehicle_journeys` is ok

If you see other api I should test

PR #1533, opened 2016-04-11
fix new_default when max_nb_journeys equals to nb_journeys_mas_have

xlqian:
http://jira.canaltp.fr/browse/NAVITIAII-2091


xlqian:
Jenkins, test this please!

TeXitoi:
OK for me, but that's not the bug we look at (maybe that's just NMP that lost configuration).

PR #1535, opened 2016-04-11
jormungandr: add a ratelimiter on synthese proxy

kinnou02:
The rate-limiter come from https://github.com/DomainTools/rate-limit


antoine-de:
Ok for me

don't forget to change fabric too

PR #1536, opened 2016-04-12
Add Jcdecaux webservice for bss

pthegner:


source/jormungandr/jormungandr/parking_space_availability/bss/jcdecaux.py
```diff
@@ -0,0 +1,65 @@
...
+# Stay tuned using
+# twitter @navitia
+# IRC #navitia on freenode
+# https://groups.google.com/d/forum/navitia
+# www.navitia.io
+from __future__ import absolute_import, print_function, unicode_literals, division
+from jormungandr.parking_space_availability.bss.bss_provider import BssProvider
+from jormungandr.parking_space_availability.bss.stands import Stands
+from jormungandr import cache, app
+from urllib2 import URLError, HTTPError, urlopen
```
kinnou02: We don't like urllib (1, 2 or 3, they are all pretty awful) [request](http://docs.python-requests.org/en/master/) is our preferred solution for dealing with HTTP

source/jormungandr/jormungandr/parking_space_availability/bss/jcdecaux.py
```diff
@@ -0,0 +1,65 @@
...
+        self.api_key = api_key
+
+    def support_poi(self, poi):
+        properties = poi.get('properties', {})
+        return properties.get('operator') == self.OPERATOR and \
+               properties.get('network') == self.network
+
+    def _call_webservice(self, station_id):
+        try:
+            response = urlopen(self.WS_URL_TEMPLATE.format(station_id, self.contract, self.api_key))
```
kinnou02: Two things here: we always want to specify a timeout when calling an external services, not that we don't trust them, but we don't trust them :)
and we also add a circuit breaker that will prevent us to call a webservice that is dead.
There is an example here: https://github.com/CanalTP/navitia/commit/d05e7ccc2f0129928a925a783a3b0f902d9db516

antoine-de:
@pthegner can you notify us when you made the changes ?

It's up to you, but there is a navitia release tomorrow, if you want this change to be shipped in it, it would be nice for the PR to be changed today :wink:

pthegner:
@antoine-de it's done

PR #1538, opened 2016-04-12
[tyr]: manage directories in autocomplete_parameters

kadhikari:
api post creats a directory named autocomplete_parameter.name
api delete deletes directory and it's contents named
autocomplete_parameter.name

### DIFF
--- a/source/navitiacommon/navitiacommon/models.py
+++ b/source/navitiacommon/navitiacommon/models.py
@@ -42,6 +42,7 @@ from navitiacommon.utils import street_source_types, address_source_types, \
     poi_source_types, admin_source_types
 
 from navitiacommon import default_values
+import os
 
 db = SQLAlchemy()
 
@@ -508,13 +509,13 @@ class AutocompleteParameter(db.Model, TimestampMixin):
         self.admin = admin
 
     def main_dir(self, root_path):
-        return '{}/{}'.format(root_path, self.name)
+        return os.path.join(root_path, self.name)
 
     def source_dir(self, root_path):
-        return '{}/{}/{}'.format(root_path, self.name, "source")
+        return os.path.join(self.main_dir(root_path), "source")
 
     def backup_dir(self, root_path):
-        return '{}/{}/{}'.format(root_path, self.name, "backup")
+        return os.path.join(self.main_dir(root_path), "backup")
 
     def __repr__(self):
         return '<AutocompleteParameter %r>' % self.name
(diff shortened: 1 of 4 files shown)


source/tyr/tyr/tasks.py
```diff
@@ -276,3 +276,24 @@ def heartbeat():
             producer = connection.Producer(exchange=exchange)
             producer.publish(task.SerializeToString(), routing_key='{}.task.heartbeat'.format(instance.name))
 
+
+@celery.task()
+def create_autocomplete_depot(name):
+    autocomplete_path = current_app.config['TYR_AUTOCOMPLETE_DIR']
+
+    if os.path.exists(autocomplete_path):
```
antoine-de: I don't know if we don't want to create this directory, but at least we should warn if the directory is not correctly created

kadhikari: @antoine-de 
i modified to add an error message if the root disrectory is absent.

source/tyr/tyr/resources.py
```diff
@@ -998,6 +999,7 @@ def post(self):
             autocomplete_parameter.admin = args['admin']
             db.session.add(autocomplete_parameter)
             db.session.commit()
+            create_autocomplete_depot(autocomplete_parameter.name)
```
kinnou02: for an asynchronous call you need to use the `delay()` method: http://docs.celeryproject.org/en/latest/userguide/calling.html

source/tyr/tyr/tasks.py
```diff
@@ -276,3 +276,26 @@ def heartbeat():
...
+        name_path = os.path.join(autocomplete_path, name)
```
antoine-de: can you abstract this a bit ?

maybe put in AutocompleteParameter a `source_dir` / `backup_dir` method ?

kadhikari:
@antoine-de 
This correction manages also source and backup directory

source/navitiacommon/navitiacommon/models.py
```diff
@@ -507,6 +507,15 @@ def __init__(self, name=None, street=None, address=None,
         self.poi = poi
         self.admin = admin
 
+    def main_dir(self, root_path):
+        return '{}/{}'.format(root_path, self.name)
+
+    def source_dir(self, root_path):
+        return '{}/{}/{}'.format(root_path, self.name, "source")
```
antoine-de: minor: you can use as previously os.path.join(root_path, self.name, "source") it will handle the '/' a bit better

kadhikari:
@antoine-de 
done

PR #1544, opened 2016-04-18
Doc: URL correction

stifoon:


documentation/slate/source/includes/apis.md
```diff
@@ -204,7 +204,7 @@ period. Both parameters "until" and "since" are optional.
 
 Example:
 
--   <https://api.navitia.io/v1/coverage/fr_idf/vehicle_journeys?since=20150912T120000&until=20150913T110000>
+-   <https://api.navitia.io/v1/coverage/fr-idf/vehicle_journeys?since=20150912T120000&until=20150913T110000>
```
antoine-de: there is an invalid period for the url, shouldn't we change it also ?

PR #1545, opened 2016-04-18
TYR: automatic creation fo main directory for autocomplete

antoine-de:
this way we don't have to handle it with fabric

---

This change is [Reviewable](https://reviewable.io/reviews/canaltp/navitia/1545)

### DIFF
--- a/source/tyr/tyr/tasks.py
+++ b/source/tyr/tyr/tasks.py
@@ -378,18 +378,15 @@ def heartbeat():
 @celery.task()
 def create_autocomplete_depot(name):
     autocomplete_dir = current_app.config['TYR_AUTOCOMPLETE_DIR']
-    if os.path.exists(autocomplete_dir):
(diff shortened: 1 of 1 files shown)


source/tyr/tyr/tasks.py
```diff
@@ -378,18 +378,15 @@ def heartbeat():
...
-                os.mkdir(main_dir)
-                os.mkdir(autocomplete.source_dir(autocomplete_dir))
-                os.mkdir(autocomplete.backup_dir(autocomplete_dir))
-            except OSError:
-                logging.error('create directory {} failed'.format(main_dir))
-    else:
-        logging.error('directory {} does not exist'.format(autocomplete_dir))
+    autocomplete = models.AutocompleteParameter.query.filter_by(name=name).first_or_404()
+    main_dir = autocomplete.main_dir(autocomplete_dir)
+    if not os.path.exists(main_dir):
```
TeXitoi: What if main dir exists but not source and backup ?

PR #1546, opened 2016-04-20
[TYR] update login (User model)

lrocheWB:
See with @kinnou02, we can change user login, as:
- [PUT] v0/users/<ID>?login=newValue

### DIFF
--- a/source/tyr/tests/integration/users_test.py
+++ b/source/tyr/tests/integration/users_test.py
@@ -261,8 +261,7 @@ def test_update_user(create_multiple_users):
     resp = api_put('/v0/users/{}'.format(create_multiple_users['user1']), data=json.dumps(user),
                    content_type='application/json')
     assert resp['id'] == create_multiple_users['user1']
-    #login can't be updated, we keep the original value
-    assert resp['login'] == 'foo'
+    assert resp['login'] == user['login']
     assert resp['email'] == user['email']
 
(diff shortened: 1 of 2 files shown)


alcalyn:
PATCH should be most pertinent
PUT: http://restcookbook.com/HTTP%20Methods/put-vs-post/,
PATCH: http://restcookbook.com/HTTP%20Methods/patch/

source/tyr/tests/integration/users_test.py
```diff
@@ -261,8 +261,7 @@ def test_update_user(create_multiple_users):
     resp = api_put('/v0/users/{}'.format(create_multiple_users['user1']), data=json.dumps(user),
                    content_type='application/json')
     assert resp['id'] == create_multiple_users['user1']
-    #login can't be updated, we keep the original value
-    assert resp['login'] == 'foo'
+    assert resp['login'] == user['login']
```
pbougue, line 264: This test is now broken, as we get a 409 "duplicate user" error (generated in resource.py)
@lrocheWB did you try pass tyr tests? Would you know how to fix this? It would be nice to talk about that.

lrocheWB, line 264: @pbougue as see in console output "100% tests passed, 0 tests failed out of 58", so for me there is no error. We can check this tomorrow...

PR #1552, opened 2016-04-26
Doc: explanations on qualification process

stifoon:


documentation/slate/source/includes/stuff.md
```diff
@@ -26,6 +26,13 @@ Journey qualification process
 
 Since Navitia can return several journeys, it tags them to help the user
 choose the best one that matches their needs.
+Here are some tagging rules:
+-    there is only one "Best" itinerary. It is the ASAP one, on best time
```
pbougue: I would remove caps -> "best" (in case someone really try to match Best)

documentation/slate/source/includes/stuff.md
```diff
@@ -26,6 +26,13 @@ Journey qualification process
...
+-    the other kinds of itineraries are displayed only if the are relevant
```
pbougue: if theY

documentation/slate/source/includes/stuff.md
pbougue: I don't understand the end of the sentence. You mean "It's the ASAP one, and if 2 are arriving at the same time the shortest one" ?

buildbotctp:
Since the tags depend on the scenario I think it a bit tricky to speak about them in the doc.
I think we should discuss this

documentation/slate/source/includes/stuff.md
pbougue: And you would remove "kind of"

documentation/slate/source/includes/stuff.md
```diff
@@ -26,6 +26,13 @@ Journey qualification process
...
+-    and for a specific tag, you may find many relevant itinerary
+
+For example, for a specific request, you can find the "best" itinerary, 2 "less_fallback_walk" ones 
+(with less walking, but taking more time) and no "comfort" (the "best" one is the same as the "comfort" one for example).
```
pbougue, line 35: Which brings the question of what is an "comfort" itinerary... :)

stifoon, line 35: @antoine-de  i've been asked for that many times : "why Navitia doesn't compute systemically the less connection itinerary". And these tags are normalized
@pbougue ok, but later.

PR #1554, opened 2016-04-26
Tyr: fix test 'test_update_block_until' (duplicate user)

lrocheWB:
@pbougue, see together (fix 409 error 'duplicate user')

### DIFF
--- a/source/tyr/tests/integration/users_test.py
+++ b/source/tyr/tests/integration/users_test.py
@@ -268,7 +268,7 @@ def test_update_block_until(create_multiple_users):
     """
     we update a user
     """
-    user = {'login': 'user1', 'block_until': '20160128T111200Z'}
+    user = {'block_until': '20160128T111200Z'}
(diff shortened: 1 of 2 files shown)


pbougue:
This is ok for tests, but I don't know if we should do more... The "duplicate user" error is actually an integrity error, I think, but not sure. @kinnou02 any opinion?

kinnou02:
It's not what we want:
In the resources we do this:

```
parser.add_argument('login', type=unicode, required=False,
                case_sensitive=False, help='user identifier', location=('json', 'values'))
```

then that: 

```
user.login = args['login']
```

So, if the login is not set we get a None value for `user.login`.
In fact the error message wrong, in reality it's something about null value blablabla, but all integrity error are treated the same way in tyr.

So there are two ways to fix that:
- the login is required
- the default value for the login is set to the previous value

lrocheWB:
@kinnou02 done, thanks...

PR #1556, opened 2016-04-27
jormun: StatManager didn't handle response with error set to None

kinnou02:


source/jormungandr/jormungandr/stat_manager.py
```diff
@@ -170,7 +170,7 @@ def fill_info_response(self, stat_info_response, call_result):
             stat_info_response.object_count = call_result[0]['pagination']['items_on_page']
 
     def fill_result(self, stat_request, call_result):
-        if 'error' in call_result[0]:
+        if 'error' in call_result[0] and call_result[0]['error']:
```
pbougue, line 173: wouldn't it be better with is not None? You want to match `""` ?

PR #1557, opened 2016-04-27
Optimisation traffic_reports api

azime:


source/disruption/traffic_reports_api.cpp
```diff
@@ -292,6 +291,10 @@ void TrafficReport::disruptions_list(const std::string& filter,
                         const type::Data& d,
                         const boost::posix_time::ptime now){
 
+    if (d.pt_data->disruption_holder.get_weak_impacts().empty()){
```
antoine-de, line 295: can you add a comment there ?

source/ptreferential/ptreferential.cpp
```diff
@@ -106,7 +107,7 @@ struct select_r: qi::grammar<Iterator, std::vector<Filter>(), qi::space_type>
             [qi::_val = boost::phoenix::construct<Filter>(qi::_1, qi::_2)];
         after_clause = ("AFTER(" >> text >> ')')[qi::_val = boost::phoenix::construct<Filter>(qi::_1)];
         args_clause = (word|escaped_string|bracket_string)[boost::phoenix::push_back(qi::_val, qi::_1)] % ',';
-        method_clause = (word >> "." >> word >> "(" >> args_clause >> ")")
+        method_clause = (word >> "." >> word >> "("  >> -(args_clause) >> ")" )
```
antoine-de, line 113: same here, I think it would be nice to add a comment explaining that args_clause is optional

source/ptreferential/ptreferential.cpp
```diff
@@ -302,6 +355,10 @@ Indexes get_indexes(Filter filter,  Type_e requested_type, const Data & d) {
             for (const VehicleJourney* vj: d.pt_data->headsign_handler.get_vj_from_headsign(filter.args.at(0))) {
                 indexes.insert(vj->idx); //TODO bulk insert ?
             }
+        } else if ((filter.object == "vehicle_journey")
+                   && (filter.method == "has_disruption")
+                   && (filter.args.size() == 0)) {
+                indexes = get_indexes_by_impacts(d, type::Type_e::VehicleJourney);
```
antoine-de, line 365: same here I think it's nice to explain why we get the indexes by the impacts

source/ptreferential/tests/ptref_test.cpp
```diff
@@ -187,6 +187,14 @@ BOOST_AUTO_TEST_CASE(parser_method) {
     BOOST_CHECK_EQUAL(filters[0].args, std::vector<std::string>{"john"});
 }
 
+BOOST_AUTO_TEST_CASE(parser_method_has_disruption) {
```
antoine-de, line 190: I think we should also test that we are able to get the vj that have disruptions no ?

azime:
corrections

PR #1558, opened 2016-04-28
Handle disruptions from Chaos on line section

eturck:
Hello!

The goal of this pull request is to fully handle disruptions on line section coming from Chaos in kraken.
A line section is defined by : 
- One line,
- Two stop areas, one for the start of the section one for the end of the section,
- 0 to N routes. Listing all the routes impacted by this disruption.

Since it's not a standard pt_object it wasn't as easy to map to the objects in kraken so I made some choices : 
- If we have no routes in the Chaos database, or in a message, associated to the line_section I apply the disruption on every route of the line (since the line is required in the proto),
- I only do something in apply_disruption if the severity is Effect::NO_SERVICE, I don't really know how to handle Effect::SIGNIFICANT_DELAYS so I tried to keep it simple,
- If we apply a line_section disruption on a route but this route have no vj passing by those two stop_areas, I remove the impact from the route,
- On the other end if the line_section disable all the stop_times of a vj, I simply cancel the vj,
- If I have a disruption from the stop_area S1 to the stop_area S2 I'm assuming this means the disruption is from the **first** stop_time passing by S1 to the **last** stop_time passing by S2. Pretty important for line with loops.

Some of them are probably not the best choices, and there is probably some use cases I didn't think of, but I tried to keep it simple and logical.

On the technical side, to apply the disruption I was greatly inspired by apply_disruption for the stop_points, since there is some similarities. I tried to comment the part of the code, especially when I was making those choices above.
I also tried to add enough unit tests to cover most of the cases, but if I forgot some I'm of course willing to add some more.

I hope I didn't forget too many things or made too much mistakes.
I will try to be around as much as possible to quickly make changes that need to be made. And I'm on IRC if it's easier to discuss about something.

Have a nice day!


antoine-de:
Thank you very much @eturck , we're in a bit of a rush now, we'll look at the PR next week.

source/ed/build_helper.cpp
```diff
@@ -426,12 +452,25 @@ VJ builder::vj(const std::string& line_name,
                const std::string& uri,
                const std::string& meta_vj,
                const std::string& physical_mode){
-    return vj_with_network("base_network", line_name, validity_pattern, block_id,
+    return vj_with_network("base_network", line_name, "", validity_pattern, block_id,
+                           wheelchair_boarding, uri, meta_vj, physical_mode);
+}
+
+VJ builder::vj_with_route(const std::string& line_name,
```
antoine-de: It's really a minor point and it's not mandatory for you to change this since it will change a lot of code, but for these builders we recently added some way to better define them based on linked setters.

this way you only define the parameters you want to change in the builder.

eg, you can replace 

```
b.vj_with_route("line:A", "route:A1", "111111", "", true, "vj:1")(some stoptimes...)
```

to 

```
b.vj("line:A").route("route:A1").uri("vj:1")(some stoptimes...)
```

to do this you only have to create a '_route' field en `ed::VJ` and add the 'route' setter in the class
defined as `VJ& route(const std::string& r) { _route = r; return *this; }` 

This way you don't have to provide lot's of default values

source/kraken/apply_disruption.cpp
```diff
@@ -201,6 +195,149 @@ struct add_impacts_visitor : public apply_impacts_visitor {
         log_end_action(mvj->uri);
     }
 
+    void operator()(nt::disruption::LineSection& ls) {
+        std::string uri =
+            "line section (" +  ls.line->uri  + " : " + ls.start_point->uri + "/" + ls.end_point->uri + ")";
+        this->log_start_action(uri);
+        if (impact->severity->effect == nt::disruption::Effect::NO_SERVICE) {
```
antoine-de: minor: for readability I personally prefer early return with a log for this

source/kraken/apply_disruption.cpp
```diff
@@ -201,6 +195,149 @@ struct add_impacts_visitor : public apply_impacts_visitor {
...
+            LOG4CPLUS_TRACE(log, "canceling " << uri);
+
+            // Get all impacted VJs and compute the corresponding base_canceled vp
+            using SectionBounds = const std::pair<const nt::StopTime*, const nt::StopTime*>;
+            std::vector<std::tuple<const nt::VehicleJourney*, nt::ValidityPattern, SectionBounds>> vj_vp_pairs;
+
+            // Computing a validity_pattern of impact used to pre-filter concerned vjs later
+            // Loop on every (even partial) days of the period
+            type::ValidityPattern impact_vp{meta.production_date.begin()}; // bitset are all initialised to 0
```
antoine-de: I think you can put this in a function in the impact no ?

it is already in `apply_disruption.cpp:212` and in this function we consider a day before the impact period in order to handle overmidnight impact. don't you think it could help you ?

source/kraken/tests/apply_disruption_test.cpp
```diff
@@ -1161,3 +1161,499 @@ BOOST_AUTO_TEST_CASE(stop_point_deletion_test) {
...
+
+    auto vjmap = b.data->pt_data->vehicle_journeys_map;
+
+    BOOST_REQUIRE_EQUAL(b.data->pt_data->lines.size(), 1);
+    BOOST_REQUIRE_EQUAL(b.data->pt_data->routes.size(), 1);
+    // Only 5 vj, vj:3 shouldn't be affected
+    BOOST_REQUIRE_EQUAL(b.data->pt_data->vehicle_journeys.size(), 5);
+
+    // Check the original vj
+    auto* vj = vjmap["vj:1"];
```
antoine-de: minor: not 100% sure, but I don't think you need those map, I think you can access a vj by name with 
`b.get<VehicleJourney>("vj:1");`

source/kraken/apply_disruption.cpp
```diff
@@ -201,6 +195,149 @@ struct add_impacts_visitor : public apply_impacts_visitor {
...
+                        for(const auto& period : impact->application_periods) {
+                            // get the vp of the section
+                            new_vp.days |= vj.get_vp_for_section(bounds_st, rt_level, period).days;
+                        }
+                        // If there is effective days for the adapted vp we're keeping it
+                        if(!new_vp.days.none()){
+                            LOG4CPLUS_TRACE(log, "vj "<< vj.uri << " is affected, keeping it.");
+                            new_vp.days >>= vj.shift;
+                            has_affected_vj = true;
+                            vj_vp_pairs.emplace_back(&vj, new_vp, bounds_st);
```
antoine-de: Hum maybe i'm missing something but this seems quite error prone.

we take an address from a vehicle_journey in the route's vj's vector.

but afterwards, in the impact application (line 311), we create a new vj with the modification and we add the new vj in the route's vector.

We thus can activate a vector reallocation and invalidate the other vj's addresses :confused: no ?

nvm: you take a vj ptr, it does not depends on the route's vector it's handled with a vector of unique_ptr, I think it will be ok

source/kraken/make_disruption_from_chaos.cpp
```diff
@@ -144,7 +144,30 @@ make_line_section(const chaos::PtObject& chaos_section,
                        << pb_section.end_point().uri() << " in LineSection invalid!");
         return boost::none;
     }
-    if (impact) line->add_impact(impact);
+    if(pb_section.routes().size()) {
+        for(auto pb_route: pb_section.routes()) {
```
antoine-de: you should take `pb_route` as a `const&`

source/kraken/make_disruption_from_chaos.cpp
```diff
@@ -144,7 +144,30 @@ make_line_section(const chaos::PtObject& chaos_section,
                        << pb_section.end_point().uri() << " in LineSection invalid!");
         return boost::none;
     }
-    if (impact) line->add_impact(impact);
+    if(pb_section.routes().size()) {
```
antoine-de: very minor, but it's better to check `! empty()`

source/kraken/tests/apply_disruption_test.cpp
```diff
@@ -1161,3 +1161,499 @@ BOOST_AUTO_TEST_CASE(stop_point_deletion_test) {
...
+
+    // Check the adapted vj
+    vj = vjmap["vj:1:Adapted:0:line_section_on_line:A_diverted"];
+    adapted_vp = vj->adapted_validity_pattern()->days;
+    base_vp = vj->base_validity_pattern()->days;
+    BOOST_CHECK_MESSAGE(ba::ends_with(adapted_vp.to_string(), "000110"), adapted_vp);
+    BOOST_CHECK_MESSAGE(ba::ends_with(base_vp.to_string(), "000000"), base_vp);
+    // The adapted vj should have only 2 stop_times
+    BOOST_REQUIRE_EQUAL(vj->stop_time_list.size(), 2);
+    BOOST_REQUIRE_EQUAL(vj->stop_time_list.front().stop_point->uri, "stop_point:10");
```
antoine-de, line 1293: not for all stop_times, but we can maybe check for the first VJ that the times are not changed (8h10 -> 8h40 in this case) no ?

source/ed/build_helper.cpp
eturck: Thanks for the tip :+1:
It's easier to use and I don't have to modify any unit tests to make it works anymore.

source/kraken/tests/apply_disruption_test.cpp
eturck: Thanks for the tip again :+1:

source/kraken/tests/apply_disruption_test.cpp
eturck, line 1293: I'll add some checks on the stop_times :)

source/kraken/apply_disruption.cpp
```diff
@@ -201,6 +195,149 @@ struct add_impacts_visitor : public apply_impacts_visitor {
...
+                const auto* vj = std::get<0>(vj_vp_section);
+                auto& new_vp = std::get<1>(vj_vp_section);
+                auto& bounds_st = std::get<2>(vj_vp_section);
+
+                bool ignore_stop(false);
+                for (const auto& st : vj->stop_time_list) {
+                    // Ignore stop if it's the range of impacted stop_times
+                    if (!ignore_stop && &st == bounds_st.first ){
+                        ignore_stop = true;
+                    }
```
xlqian, line 245: ignore_stop = !ignore_stop && &st == bounds_st.first;

source/kraken/apply_disruption.cpp
```diff
@@ -201,6 +195,149 @@ struct add_impacts_visitor : public apply_impacts_visitor {
...
+                    if(ignore_stop) {
+                        LOG4CPLUS_TRACE(log, "Ignoring stop " << st.stop_point->uri << "on " << vj->uri);
+                        // Mandatory to output impacted stops in traffic_report
+                        ls.impacted_stop_points_by_route[vj->route->uri].insert(st.stop_point);
+                        // Reset ignore_stop if it's the end stop_time
+                        if(&st == bounds_st.second) {
+                            ignore_stop = false;
+                        }
```
xlqian: ignore_stop = !(&st == bounds_st.second)     :)

source/type/pb_converter.cpp
```diff
@@ -71,8 +71,19 @@ struct PbCreator::Filler::PtObjVisitor: public boost::static_visitor<> {
         add_pt_object(bo);
     }
     void operator()(const nt::disruption::LineSection& line_section) const {
-        //TODO: for the moment a line section is only a line, but later we might want to output more stuff
-        add_pt_object(line_section.line);
+        for(const nt::Route* route: line_section.routes) {
+            auto* pobj = add_pt_object(route);
```
antoine-de: we want to output line sections as routes with stop impacted stops ?

I'm quite OK with this, but @stifoon what do you think about this ?

Isn't it a pity to loose the `line section` information ?

source/kraken/apply_disruption.cpp
```diff
@@ -201,6 +195,149 @@ struct add_impacts_visitor : public apply_impacts_visitor {
...
+
+            // Get all impacted VJs and compute the corresponding base_canceled vp
+            using SectionBounds = const std::pair<const nt::StopTime*, const nt::StopTime*>;
+            std::vector<std::tuple<const nt::VehicleJourney*, nt::ValidityPattern, SectionBounds>> vj_vp_pairs;
+
+            // Computing a validity_pattern of impact used to pre-filter concerned vjs later
+            // Loop on every (even partial) days of the period
+            type::ValidityPattern impact_vp{meta.production_date.begin()}; // bitset are all initialised to 0
+            for (const auto& period: impact->application_periods) {
+                for (bg::day_iterator it(period.begin().date()); it <= period.end().date() ; ++it) {
```
xlqian: hmm, the pre-filtering doesn't consider vjs that pass midnight, can that be a problem?

eturck: Yes, it's a problem. I talk about it with @antoine-de on IRC a bit, I'm going to change a few things :).

source/kraken/apply_disruption.cpp
eturck, line 245: Not really. When ignore_stop is set to true I don't want it to change anymore.
With this code when ignore_stop is set to true, on the next step we're going to have : 

``` cpp
ignore_stop = !true && ... // ignore_stop will be set to false again.
```

But I can indeed change it to something a bit more concise like : 

``` cpp
ignore_stop |= (&st == bounds_st.first);
```

Thank you!

source/kraken/apply_disruption.cpp
eturck: Thanks ;)

source/type/type.cpp
```diff
@@ -242,6 +242,78 @@ ValidityPattern VehicleJourney::get_vp_of_sp(const StopPoint& sp,
...
+     *  - s1/s2 from the 5th to the 10th,
+     *  - s1/s2/s5 from the 11th to the 15th
+     * */
+    auto* base_vj = this->get_corresponding_base();
+    for(auto& st: (base_vj ? base_vj->stop_time_list : this->stop_time_list)) {
+        // The line section is using stop_areas so we make sure we have one
+        if(st.stop_point && st.stop_point->stop_area) {
+            if(!bounds_st.first && st.stop_point->stop_area->uri == start_stop->uri) {
+                bounds_st.first = &st;
+            }
```
xlqian, line 306: I think we can add a `else` here

source/kraken/apply_disruption.cpp
```diff
@@ -201,6 +195,149 @@ struct add_impacts_visitor : public apply_impacts_visitor {
...
+
+            // Loop on each affected vj
+            for (auto& vj_vp_section : vj_vp_pairs) {
+                std::vector<nt::StopTime> new_stop_times;
+                const auto* vj = std::get<0>(vj_vp_section);
+                auto& new_vp = std::get<1>(vj_vp_section);
+                auto& bounds_st = std::get<2>(vj_vp_section);
+
+                bool ignore_stop(false);
+                for (const auto& st : vj->stop_time_list) {
```
xlqian: Minor suggestion for the readability:

Maybe you can get `bounds_st`directly by their `order()` instead of pointer so that you can test if the `st` in the line section by the order. Something like `bounds_st.first <= st.order() <= bounds_st.second`

What do you think?

source/type/type.cpp
eturck, line 306: Between the two "if" to set bounds_st.first and bounds_st.second?
If that's what you're referring to no we can't since it's possible to have a line_section where start_stop = end_stop, so bounds_st.first == bounds_st.second.

source/kraken/apply_disruption.cpp
eturck: That's what I did first.
But I had a pair of uint_16 then, and I didn't really know what to set as a default value, std::numeric_limits<uint16_t>::max()? I need to have this default value to detect if one the stop isn't found in the vj. I switch to pointer for the easiness to compare to nullptr and direct access to the stop_time in the loop you commented on.

But I'm open to change it if you think it would be better to juste store the rank of the stop_times. I just need to know if you think storing std::numeric_limits<uint16_t>::max() as a default value is ok or if you suggest a different way to do it.

source/kraken/apply_disruption.cpp
xlqian, line 245: OK! :+1:

source/kraken/apply_disruption.cpp
xlqian: I actually doubted you've done this for a reason I ignored :) 

Yep, I thought of that too, you can do a `boost::Optional<uint16_t>`, how about that?

eturck: Yes! boost::optional, I should have thought about it... Thanks :smile:

source/kraken/apply_disruption.cpp
```diff
@@ -201,6 +195,149 @@ struct add_impacts_visitor : public apply_impacts_visitor {
...
+                    new_stop_times.push_back(std::move(new_st));
+                }
+
+                auto mvj = vj->meta_vj;
+                mvj->push_unique_impact(impact);
+
+                // If all stop times have been ignored
+                if(new_stop_times.empty()) {
+                    LOG4CPLUS_DEBUG(log, "All stop times has been ignored on "<< vj->uri << ". Cancelling it.");
+                    mvj->cancel_vj(rt_level, impact->application_periods, pt_data);
```
xlqian: :+1:

xlqian:
Thank you for the PR and your effort on it!

source/kraken/apply_disruption.cpp
xlqian: Can you add a test where `new_stop_times` is empty plz?

eturck: Good call, I'm going to add one.

eturck:
I made changes : 
- Remove default route_name value and added a .route method on VJ for the builder,
- Early return in apply_disruption if it's not a severity NO_SERVICE to improve readability,
- Moved the impact vp calculation in the impact object since I use the same code than apply_disruption on a stop_point,
- Use b.get<VehicleJourney> instead of using a map to get vjs by their uri in unit tests,
- Loop on a const reference and use empty instead of size in make_disruption_from_chaos,
- Add some checks on stop_times departure / arrival time in apply_disruption_test,
- More concise way to set ignore_stop in apply_disruption,
- Handle vj pass midnight,
- Use order of stop_times instead of pointer in the bounds of the section,
- Added a test for vj pass midnight and line_section cancelling completely a vj.

The only thing left is about what do we output for a line_section :).
I changed it from just the line to a list of routes with, for each routes, a list of impacted stops, but it could break something expecting a line.

Let me know if you see anything else!

antoine-de:
great, thanks!

The PR is fine for me, great job!

For the output of the section, I'm not completely convinced by the current state because there is a confusion between a `stop_point` and a `stop_time`

it will be easier to understand with some json :smile: 

`GET /disruptions`:

The current state is not great, we hide lot's of information

``` json
{
    "disruptions": [
        {
            "impacted_objects": [   {
                    "pt_object": {
                        "embedded_type": "line", 
                        "id": "bob", 
                        "line": {
...
                        }
                    }
                }
            ], 
            "uri": "6985324e-3616-41e8-8af4-a96949154218"
        }

```

what you propose is way better, but I think there is an ambiguity between the `impacted_stops` being either a stop_time if the impacted object is a `trip` or a stop_point if the impacted object is a `route` (or in reallity a `line_section`)

``` json
{
    "disruptions": [
        {
            "impacted_objects": [   {
                    "impacted_stops": [   {
                            "stop_point": {
                                "coord": {
                                    "lat": "49.59962", 
                                    "lon": "6.134022"
                                }, 
                                "id": "stop_point:OCE:SP:TGV-82001000", 
                                "label": "gare de Luxembourg (Luxembourg)", 
                            }, 
                            "stop_time_effect": "deleted"
                        }, 
                    ], 
                    "pt_object": {
                        "embedded_type": "route", 
                        "id": "bob", 
                        "name": "bob", 
                        "route": {
...
                        }
                    }
                }
            ], 
            "uri": "6985324e-3616-41e8-8af4-a96949154218"
        }
```

To better handle it, one solution would be to add a `line_section` object to navitia.
A downside would be that this object would not be a "static" object like the other navitia's object, but only stored when a disruption on this line_section is available

In this case the response would be:

``` json
{
    "disruptions": [
        {
            "impacted_objects": [   {
                    "pt_object": {
                        "embedded_type": "line_section", 
                        "id": "bob", 
                        "name": "bob", 
                        "line_section": {
...
                        }
                    }
                }
            ], 
            "uri": "6985324e-3616-41e8-8af4-a96949154218"
        }
```

in order to respect the API contract we would have to create a new end_point:
`GET /line_sections/{id}`:

``` json
{
    "line_sections": [ {
...
}]
}
```

one way to represent the line_section would be to do as you suggest and output all impacted routes with there stop_points.
ex:

``` json
{
    "line_sections": [ {
  "id": "toto",
   "name": "line section of RER A around Chatelet and Gare de lyon",
  "sections": [
       "route": {
            a route
         },
        "stop_points": [
           some stop_points
         ]
    ]
}]
}
```

This seems a bit trouble some, but I think it would be cleaner. We would of course not ask you to implement this as I don't think you need to use the `/disruptions` API :grin: 
if we decide to do this solution, I think for the moment we can keep the old way to represent an impact on a line section thus we won't break the integration when we'll implement the clean way.

what do you think about this @eturck @stifoon  ?

eturck:
There is indeed a confusion between stop_points and stop_times in the impacted_stops.
It's tricky because a line_section is not really an object but a group of objects... I don't really know what's the best way to describe it.

I have something like that currently : 

``` json
"impacted_stops": [
    {
        "amended_arrival_time": "000000",
        "stop_point": {
            "name": "François Verdier",
            "links": [ ],
            "coord": {
                "lat": "43.600549",
                "lon": "1.452131"
            },
            "label": "François Verdier (Toulouse)",
            "equipments": [
                "has_wheelchair_boarding"
            ],
            "id": "stop_point:2520"
        },
        "stop_time_effect": "deleted",
        "amended_departure_time": "000000",
        "base_arrival_time": "000000",
        "cause": "",
        "base_departure_time": "000000"
    },
    (...)
]
```

I guess I see 3 way to make it "better" : 
- At least make the departure_time / arrival_time optional? The stop_time_effect is 'deleted' but since we don't give any time it concerns all stop_times? That's still a bit ambiguous...
- You could have something else after `repeated StopTimeUpdate impacted_stops` in type.proto, like a `repeated StopPoint impacted_stop_points`, but from what I see the StopTimeUpdate is close to the GTFS-RT reference, and there is nothing close to what we try to represent in GTFS-RT,
- Make line_section an object. It would certainly be easier to handle and the source code would be a bit cleaner since currently line_section are a specific case almost every time, but it's quite a bit of work :smile:.

I don't need to use the disruption API... yet :), but I'm sure in the future it's going to change ;).
It's ok for me to keep the old way in order to have something better in the future :).

eturck:
I reverted the output of the line_section to the current one.
I removed a useless map, changed back the test in jormungandr and added the impact on the line again.

Let me know if I need to change anything else.

antoine-de:
great PR, thank you very much!

we'll see who need a better line_section output to manage the work.

Thanks again for the great work.

PR #1561, opened 2016-04-29
Add note about `traveler_type` setting fallback modes even despite `forbidden_uris[]` set.

daaain:
Small note about something which had me scratching my head for a while!

As discussed in https://groups.google.com/forum/#!topic/navitia/6jjwjNVVL80


buildbotctp:

antoine-de:
@daaain , thanks!

I think I'll open another PR to add more details on the `forbidden_uris[]` section as it does not seems very clear.

PR #1564, opened 2016-05-03
[tyr]: add param --import-way in osm2mimir

kadhikari:


kadhikari:
param value deleted for import-way and param import-admin added

source/tyr/tyr/binarisation.py
```diff
@@ -534,6 +534,10 @@ def osm2mimir(self, autocomplete_instance, filename, job_id, dataset_uid):
         for lvl in autocomplete_instance.admin_level:
             params.append('--level')
             params.append(str(lvl))
+        if autocomplete_instance.admin:
```
kinnou02: The value need to be checked.

kadhikari:
@kinnou02 , i modified on your comments

PR #1565, opened 2016-05-03
Fix pre-filter of vjs for impact on stop_points

eturck:
While discussing about the PR #1558 @antoine-de told me I had indeed found a bug.
The use case is : 
- your production date starts on the 1st,
- your impact starts on the 2nd at 9am,
- your impact ends on the 3rd at 8:30am.

The code using time_iterator was missing the last day because we had : 
1. "it" was initialized at  period.begin() - one_day => 1st at 9am,
2. second loop, we increment of one_day => it = 2st at 9am,
3. third loop, we increment of one day => it = 3rd at 9am.

And on the third loop it < period.end() is false, since it = 3rd at 9am and period.end() = 3rd at 8:30am.
We were never adding the 3rd to the impact_vp, causing some vj to be ignored when they should not.

I added a little unit test failing before the fix, and working afterwards.
Let me know if there is something to change. I'll rebase the code of the PR #1558 afterwards and move this code in the impact since I'm using it when applying disruption on line_sections.


antoine-de:
Great work @eturck !

PR #1569, opened 2016-05-04
Jormungandr: change isochron default duration

antoine-de:


source/jormungandr/jormungandr/interfaces/v1/Journeys.py
```diff
@@ -546,6 +546,16 @@ def get(self, region=None, lon=None, lat=None, uri=None):
                     abort(503, message="Unable to compute journeys "
                                        "from this object")
 
+        if args['destination'] and args['origin']:
+            api = 'journeys'
+        else:
+            api = 'isochrone'
+
+        if api == 'isochrone':
```
xlqian, line 554: maybe this block can be moved after L552?

xlqian:
ok for me

pbougue:
ok for me, maybe just add a mention of that in doc?

PR #1573, opened 2016-05-06
Jormungandr: /stop_schedules: do not activate realtime if datetime

antoine-de:
no realtime activated if a datetime is provided in /departures,
/arrivals and /stop_schedules

### DIFF
--- a/source/jormungandr/jormungandr/interfaces/v1/Schedules.py
+++ b/source/jormungandr/jormungandr/interfaces/v1/Schedules.py
@@ -150,12 +150,19 @@ class Schedules(ResourceUri, ResourceUtc):
 
             if not args['data_freshness']:
                 args['data_freshness'] = 'realtime'
-        elif not args.get('calendar'):
-            #if a calendar is given all times will be given in local (because the calendar might span over dst)
-            if args['from_datetime']:
-                args['from_datetime'] = self.convert_to_utc(args['from_datetime'])
-            if args['until_datetime']:
-                args['until_datetime'] = self.convert_to_utc(args['until_datetime'])
+        else:
+            if not args.get('calendar'):
+                # if a calendar is given all times will be given in local (because the calendar might span
+                # over dst)
+                if args['from_datetime']:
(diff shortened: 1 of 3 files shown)


antoine-de:
added the do_not_merge flag, I'm not sure this is the right way to handle it, I think it's the proxy responsibility to filter the date

kinnou02:
Can we close it?

antoine-de:
I don't know, it depends on whether we want to change the default values or not. @stifoon ?

stifoon:
OK, after lots of discussions with re-users, here is conclusions:
- /stop_schedules will use by default "data_fresness=base_schedule" parameter
- /departures and /arrivals will use by default "data_fresness=real_time" parameter

source/jormungandr/jormungandr/interfaces/v1/Schedules.py
```diff
@@ -150,12 +150,19 @@ def get(self, uri=None, region=None, lon=None, lat=None):
...
+            if not args.get('calendar'):
+                # if a calendar is given all times will be given in local (because the calendar might span
+                # over dst)
+                if args['from_datetime']:
+                    args['from_datetime'] = self.convert_to_utc(args['from_datetime'])
+                if args['until_datetime']:
+                    args['until_datetime'] = self.convert_to_utc(args['until_datetime'])
+
+            if args['data_freshness'] in (None, 'realtime'):
+                # If a data has been provided, we cannot give realtime results
```
xlqian, line 163: `If a data_freshness ....`?

kinnou02:
where are we?

pbougue:
As far as I understood, the description of need that is up to date is the one in @stifoon 's last comment.
So this PR as to be changed a lot functionnaly, and I don't know about tests.
Anyway, it's a schedule-team's problem, I guess :)

pbougue:
This PR been dangling for too long, and no real need on that.
The point is, we want default values to be simple, so no default depending on other param.
Then it seems preferable to have the same behaviour on /departures and on /stop_schedules, all realtime and it's what we are doing right now.
So we are closing this, as there has been no problems or requests since we implemented RT.

PR #1583, opened 2016-05-11
Isochrons : create the new API /isochrons

ZoeBrunet:
Response time for different duration of isochrons

| duration | without Merge | With Merge |
| --- | --- | --- |
| 10 min | 0m1.624s | 0m0.545s |
| 1 h | 0m45.845s | 0m7.351s |
| 5h | 2m3.535s | 0m3.886s |
| 10h | 2m4.424s | 0m2.967s |

PR #1593, opened 2016-05-20
Jormungander: add tests for isochrons

ZoeBrunet:
Needs PR #1583 
Check the geojson when the datetime field is empty and when we change the speed_factor


source/jormungandr/tests/graphical_isochrons_tests.py
```diff
@@ -85,6 +85,56 @@ def test_graphical_isochrons_to_stop_point(self):
         assert (multi_poly.contains(stopA))
         is_valid_graphical_isochron(response, self.tester, q)
 
+    def test_graphical_isochrons_no_datetime(self):
+        q = "v1/coverage/main_routing_test/isochrons?from={}&duration={}&_current_datetime={}"
+        q = q.format(s_coord, '3600', '20120614T080000')
+        response = self.query(q)
```
antoine-de: I think it would be better to have more explicit names

source/jormungandr/tests/graphical_isochrons_tests.py
```diff
@@ -85,6 +85,56 @@ def test_graphical_isochrons_to_stop_point(self):
...
+        query = "v1/coverage/main_routing_test/isochrons?from={}&datetime={}&duration={}"
+        query = query.format(s_coord, '20120614T080000', '3600')
+        resp = self.query(query)
+
+        is_valid_graphical_isochron(response, self.tester, q)
+        assert len(response['isochrons']) == len(resp['isochrons'])
+
+        for i in range(0, len(response['isochrons'])):
```
antoine-de: a more pythonic way to write it would be to use `zip`

something like: `for isochron_no_dt, isochron in zip(response_no_dt, isochron):`

antoine-de:
nice tests!

PR #1596, opened 2016-05-23
jormun: set timezone of the request after a call to kraken

kinnou02:
It's ugly to do that in Instance, but I don't see a solution that
doesn't imply to modify all API.

### DIFF
--- a/source/jormungandr/jormungandr/timezone.py
+++ b/source/jormungandr/jormungandr/timezone.py
@@ -34,26 +34,36 @@ import pytz
 from flask import g
 from jormungandr.exceptions import TechnicalError, RegionNotFound
 
-
 def set_request_timezone(region):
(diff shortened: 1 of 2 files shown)


source/jormungandr/jormungandr/timezone.py
```diff
@@ -34,26 +34,36 @@
...
     tz = pytz.timezone(instance.timezone)
 
     if not tz:
-        logger.warn("impossible to find timezone: '{}' for region {}".format(instance.timezone, region))
+        logger.warn("impossible to find timezone: '{}' for region {}".format(instance.timezone, instance.name))
 
-    g.timezone = tz
+    try:
+        g.timezone = tz
+    except RuntimeError:
```
antoine-de, line 65: isn't it an AttributeError ?

PR #1598, opened 2016-05-24
Isochron : add min_duration

ZoeBrunet:
Add min duration argument and some tests to make isochron between two durations.

min_duration

### DIFF
--- a/source/kraken/tests/worker_test.cpp
+++ b/source/kraken/tests/worker_test.cpp
@@ -73,10 +73,11 @@ static pbnavitia::Request create_request(bool wheelchair, const std::string& des
     return req;
 }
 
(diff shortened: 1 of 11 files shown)


source/kraken/tests/worker_test.cpp
```diff
@@ -73,10 +73,11 @@ static pbnavitia::Request create_request(bool wheelchair, const std::string& des
     return req;
 }
 
-static pbnavitia::Request create_isochron_request(int max_duration, const std::string& origine) {
+static pbnavitia::Request create_isochron_request(int max_duration, int min_duration, const std::string& origine) {
```
antoine-de: typo on origin**e**

PR #1602, opened 2016-05-26
ed: "NULL" value of optional attribut in ntfs files managed

kadhikari:
some tests added


kadhikari:
Ticket concerned: http://jira.canaltp.fr/browse/NAVITIAII-2130

kadhikari:
I pushed a version with merged version of lotus (default null value = **NULL_LOTUS_VALUE**)

antoine-de:
OK for me, thanks

PR #1619, opened 2016-06-01
Isochrone: catch boost exception

ZoeBrunet:
Catch boost exception when we merge polygon and add a warning


antoine-de:
The IsochronException is not used for the moment, but since I think it will be usefull soon, OK for me

PR #1621, opened 2016-06-01
Isochrone : Add an e to isochron

ZoeBrunet:
Change the name of the API
http://jira.canaltp.fr/browse/NAVP-597


source/kraken/worker.cpp
```diff
@@ -695,7 +695,7 @@ pbnavitia::Response Worker::pt_ref(const pbnavitia::PTRefRequest &request,
                                     current_datetime);
 }
 
-pbnavitia::Response Worker::graphical_isochron(const pbnavitia::GraphicalIsochronRequest &request_iso,
+pbnavitia::Response Worker::graphical_isochrone(const pbnavitia::GraphicalIsochroneRequest &request_iso,
                                                const boost::posix_time::ptime& current_datetime) {
```
pbougue: minor (very): alignment

source/navitia-proto
```diff
@@ -1 +1 @@
-Subproject commit 1f28df6b5e914e3d5d44332e16a5250a74513131
+Subproject commit 4c29c8f78d67cfd8037aa264b2f303641ddc762c
```
pbougue: don't forget to update submodule sha-1 now that it's merged ;)

source/routing/isochrone.h
```diff
@@ -37,12 +37,12 @@ www.navitia.io
...
-    IsochronException() = default;
-    IsochronException(const IsochronException&) = default;
-    IsochronException& operator=(const IsochronException&) = default;
-    virtual ~IsochronException() noexcept = default;
+struct IsochroneException: public recoverable_exception {
+    IsochroneException(const std::string& msg): recoverable_exception(msg) {}
+    IsochroneException() = default;
+    IsochroneException(const IsochroneException&) = default;
+    IsochroneException& operator=(const IsochroneException&) = default;
+    virtual ~IsochroneException() noexcept = default;
```
pbougue: You should define that method in .cpp
Bit tricky but if you don't do that, the compiler embeds vtable everywhere and compiling is slow enough already.
(I didn't just caught that, my compiler raised a warning)
So in .h:

```
virtual ~IsochroneException() noexcept;
```

And in .cpp:

```
IsochroneException::~IsochroneException() noexcept = default;
```

source/utils
```diff
@@ -1 +1 @@
-Subproject commit 9f21f114981ff9ed5fdc0baa4f7c1906136de735
+Subproject commit 97c292b6284b22ab8b0054c03a148d064a4e9d90
```
pbougue: It's weird that you have that submodule changed. You didn't touch utils, did you?
You should probably do a `git submodule update` ;)
This may be the reason for the conflicts, so you should do that before rebase on dev.

pbougue:
Should we worry about conf files, also?

ZoeBrunet:
We should rename GRAPHICAL_ISOCHRON, GRAPHICAL_ISOCHRON**E** in conf files.

PR #1624, opened 2016-06-02
Jormungandr: for integration tests check if a kraken is dead

antoine-de:
for integration tests check if a kraken is dead before killing it to
have better error messages


antoine-de:
Jenkins, test this please

xlqian:
ok for me

PR #1625, opened 2016-06-03
ed: Tests modified to valid docker_test on optional NULL  value

kadhikari:


source/ed/tests/fusioparser_test.cpp
```diff
@@ -221,10 +221,10 @@ BOOST_AUTO_TEST_CASE(parse_small_ntfs_dataset) {
     // headsign of vj and stop_times
     BOOST_REQUIRE_EQUAL(vj1->name.substr(0,6), "NULL");
 
-    BOOST_REQUIRE_EQUAL(vj1->stop_time_list[0]->headsign, "HS1");
```
antoine-de, line 224: why did you change this headsign ?

kadhikari, line 224: This was a headsign value that i had added in my last pull request. Infact it's better to have some empty, NULL and some other texts also. I will do that.

kadhikari:
i modified some values in data and test file.

antoine-de:
Jenkins, test this please

pbougue:
Ok for me.

PR #1626, opened 2016-06-03
Jormungandr: check if the test instances are up at the beginning

antoine-de:
hope this make the very slow CI platform work


TeXitoi:
OK for me.

source/jormungandr/tests/tests_mechanism.py
```diff
@@ -34,6 +34,8 @@
 from operator import itemgetter
 from flask import logging
 import mock
+import retrying
```
xlqian, line 37: can this line be removed?

antoine-de, line 37: nop it's used below

PR #1628, opened 2016-06-03
Jormun: fix max_successive_physical_mode

pbougue:
JIRA: http://jira.canaltp.fr/browse/NAVITIAII-2135

works on my stif instance

### DIFF
--- a/source/jormungandr/jormungandr/instance.py
+++ b/source/jormungandr/jormungandr/instance.py
@@ -222,6 +222,11 @@ class Instance(object):
         return get_value_or_default('factor_too_long_journey', instance_db, self.name)
 
     @property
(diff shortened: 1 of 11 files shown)


source/jormungandr/jormungandr/instance.py
```diff
@@ -222,6 +222,11 @@ def factor_too_long_journey(self):
         return get_value_or_default('factor_too_long_journey', instance_db, self.name)
 
     @property
+    def bus_physical_mode_id(self):
```
antoine-de: Since it is parametrized, I think I would have parametrized a `successive_physical_mode_limit` to be able to limit some the number of trains or planes taken

PR #1629, opened 2016-06-07
ed: adding multi object_code support

pbougue:
JIRA: http://jira.canaltp.fr/browse/NAVITIAII-2089

TODO:
- [x] automatic tests (docker, has_code(), external_code=)
- [x] prepare PR for artemis_reference and artemis_data

needs https://github.com/CanalTP/artemis_references/pull/31 and https://github.com/CanalTP/artemis/pull/228 to be merged right after, and a push on artemis_data of new version also

### DIFF
--- a/source/jormungandr/jormungandr/schedule.py
+++ b/source/jormungandr/jormungandr/schedule.py
@@ -174,11 +174,14 @@ class RoutePoint(object):
 
     @staticmethod
     def _get_code(obj, object_id_tag):
-        if len([c.value for c in obj.codes if c.type == object_id_tag]) > 1:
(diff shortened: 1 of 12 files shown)


source/jormungandr/jormungandr/schedule.py
```diff
@@ -174,6 +174,10 @@ def __hash__(self):
 
     @staticmethod
     def _get_code(obj, object_id_tag):
+        if len([c.value for c in obj.codes if c.type == object_id_tag]) > 1:
```
antoine-de: since you compute the array, use it for the return no ?

antoine-de:
great test coverage

antoine-de:
OK for me apart from the small remark

PR #1635, opened 2016-06-10
jormungandr: delete autocomplete from bragi json

azime:
- delete "addr:" from address id

### DIFF
--- a/source/jormungandr/jormungandr/autocomplete/elastic_search.py
+++ b/source/jormungandr/jormungandr/autocomplete/elastic_search.py
@@ -283,8 +283,9 @@ class AddressId(fields.Raw):
             return None
         geocoding = obj.get('properties', {}).get('geocoding', {})
         id = geocoding.get('id')
-        if id.startswith("addr:"):
-            return id[5:]
+        prefix = "addr:"
+        if id.startswith(prefix):
+            return id[len(prefix):]
         return id
 
 geocode_addr = {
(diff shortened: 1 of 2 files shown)


source/jormungandr/jormungandr/autocomplete/elastic_search.py
```diff
@@ -275,10 +276,21 @@ def output(self, key, obj):
...
 }
 
+
+class AddressId(fields.Raw):
+    def output(self, key, obj):
+        if not obj:
+            return None
+        geocoding = obj.get('properties', {}).get('geocoding', {})
+        id = geocoding.get('id')
+        if id.startswith("addr:"):
```
xlqian: minor:

```
prefix = "addr:"
if id.startswith(prefix):
     return id[len(prefix):]
```

azime: ok

source/jormungandr/jormungandr/autocomplete/elastic_search.py
```diff
@@ -296,10 +308,20 @@ def format(self, place):
...
+    "places": fields.List(GeocodejsonFeature, attribute='features')
 }
 
 
+class delete_attribute_autocomplete():
+    def __call__(self, f):
+        @wraps(f)
+        def wrapper(*args, **kwargs):
+            objects = f(*args, **kwargs)
+            if 'Autocomplete' in objects:
```
xlqian: you can do 

```
return objects.get('Autocomplete') or objects
```

azime: ok

PR #1637, opened 2016-06-13
Fix remove "addr:" from address id in places api

azime:


source/jormungandr/jormungandr/autocomplete/elastic_search.py
```diff
@@ -48,6 +48,20 @@ def __init__(self, val):
     def output(self, key, obj):
         return self.val
 
+
+def delete_prefix(value, prefix="addr:"):
```
antoine-de: AFAIK I think it's better not to put a default parameter here and put it in each calling function

PR #1642, opened 2016-06-15
Doc: Plugins section added in navitia.io documentation

lrocheWB:
@pbougue as you suggested during the sprint review, this PR adds "plugins" section.

Result:
screen-plugins-2


stifoon:
Could you please rebase, in order to check the result in slate :)

PR #1647, opened 2016-06-16
Doc: Example for stop_schedule section.

Netmisa:
# Description

Examples added in documentation for stop_schedule section.
# Screenshot

screenshot from 2016-06-16 10 08 58


pbougue:
@Netmisa @alcalyn @lrocheWB @stifoon : Nice work on those examples, I think we have to consider add a tab on result for clarity and to avoid big chunks in the right column... Not sure if this is the best idea.
Also on all those commits, there is no comment on what is the response/the request. Minor but the new tab or the comment would help.

PR #1658, opened 2016-06-21
Kraken: Fix redundant stop schedules when stop points impacted

xlqian:
http://jira.canaltp.fr/browse/NAVITIAII-2144

### DIFF
--- a/source/routing/get_stop_times.cpp
+++ b/source/routing/get_stop_times.cpp
@@ -95,7 +95,7 @@ get_stop_times(const std::vector<routing::JppIdx>& journey_pattern_points,
                const uint32_t begining_time,
                const uint32_t max_time,
                const type::Data& data,
-               const std::string calendar_id,
+               const std::string& calendar_id,
                const type::AccessibiliteParams& accessibilite_params) {
     std::vector<datetime_stop_time> result;
 
@@ -141,7 +141,7 @@ get_stop_times(const std::vector<routing::JppIdx>& journey_pattern_points,
 std::vector<std::pair<uint32_t, const type::StopTime*>>
 get_all_stop_times(const routing::JourneyPattern& jp,
                    const routing::JourneyPatternPoint& jpp,
-                   const std::string calendar_id,
+                   const std::string& calendar_id,
                    const type::VehicleProperties& vehicle_properties) {
 
     std::set<const type::MetaVehicleJourney*> meta_vjs;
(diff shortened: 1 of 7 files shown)


source/routing/get_stop_times.cpp
```diff
@@ -141,15 +141,19 @@ get_stop_times(const std::vector<routing::JppIdx>& journey_pattern_points,
...
                    const type::VehicleProperties& vehicle_properties) {
 
     std::set<const type::MetaVehicleJourney*> meta_vjs;
     auto insert_meta_vj = [&](const nt::VehicleJourney& vj ) {
         assert(vj.meta_vj);
         meta_vjs.insert(vj.meta_vj);
     };
-    for (const auto* vj: jp.discrete_vjs) { insert_meta_vj(*vj); }
+    for (const auto* vj: jp.discrete_vjs) {
+        if(vj->isBaseScheduled()) {
```
pbougue: You have to check more than that, no?
Not sure, but we have to know the freshness we want, and provide the corresponding one, I'd say (defaulting to base).
The departure board uses this if I'm correct.

stifoon: In a calendar mode, we don't want other freshness than base_shedule. We don't want real_time for a "days of scholar week".

xlqian: we care only about the base scheduled VJs when `calendar_id` is provided

source/type/type.h
```diff
@@ -510,7 +510,9 @@ struct VehicleJourney: public Header, Nameable, hasVehicleProperties {
     int16_t utc_to_local_offset() const; //in seconds
 
     RTLevel realtime_level = RTLevel::Base;
-
+    bool isBaseScheduled() const{
```
kinnou02: should be in snake case

xlqian: oops  :scream:

TeXitoi:
OK for me.

source/routing/get_stop_times.cpp
pbougue: as orally debated, a rename of get_all_stop_times and get_stop_times mentioning calendar would be nice

source/type/type.h
pbougue: and I would remove the 'd': is_base_schedule

pbougue:
ok for me

PR #1659, opened 2016-06-22
ed: optimisation of ed_reader

kinnou02:
When using column name with libpqxx, every call resolve the name to the
position with `PQfnumber` it's fairly slow on stoptimes. So we
translate the name of the column to the position one time manually

On NL dataset without osm (and with #1657):
before: 2,32min
after:: 2,06min


TeXitoi:
OK for me.

PR #1664, opened 2016-06-23
Journeys: fix bug on max_duration_to_pt

ZoeBrunet:


kinnou02:
A few tests would be great don't you think?

PR #1665, opened 2016-06-23
Osm2ed: simplify geometry of admin in osm2ed

kinnou02:
For nl dataset we gain 5 minutes, we are down to 15min.


kinnou02:
It"s probably not really related to this PR, but some way aren't always merge like they should be. In some case it's better before this change, and in some others it's better with it.

PR #1668, opened 2016-06-23
osm2ed: remove a lot of useless node

kinnou02:
A lot of node where inserted in the database but weren't used and had to
be removed in ed2nav.

In green we have the nodes before the modification, and in blue/purple after:
node_osm


antoine-de:
whaou!

what's the impact on the data processing time?

kinnou02:
There is almost no change for fr-idf.

PR #1671, opened 2016-06-27
jormungandr: fix label in geocodejson

TeXitoi:


source/jormungandr/jormungandr/.cache/v/cache/lastfailed
```diff
@@ -0,0 +1 @@
+{}
```
antoine-de: you shouldn't commit this file

PR #1672, opened 2016-06-28
Isochrones: delete useless circles

ZoeBrunet:


source/routing/isochrone.cpp
```diff
@@ -161,6 +161,39 @@ struct InfoCircle {
         duration_left(duration_left) {}
 };
 
+static bool within_info_circle(const InfoCircle& circle,
+                              const std::vector<InfoCircle>& multi_poly,
+                              const double& speed) {
```
xlqian: & can be removed

source/routing/isochrone.cpp
```diff
@@ -161,6 +161,39 @@ struct InfoCircle {
...
+    if (multi_poly.empty()) {return false;}
+    auto begin = multi_poly.begin();
+    const auto end = multi_poly.end();
```
xlqian, line 168:
``` c++
auto begin = std::begin(multi_poly);
auto end = std::end(multi_poly);
```

source/routing/isochrone.cpp
```diff
@@ -161,6 +161,39 @@ struct InfoCircle {
...
+    double circle_radius = circle.duration_left * speed;
+    double coslat = cos(circle.center.lat() * type::GeographicalCoord::N_DEG_TO_RAD);
+    return any_of(begin, end, [speed, circle, circle_radius, coslat](const InfoCircle& it) {
```
xlqian: :+1:

source/routing/isochrone.cpp
```diff
@@ -200,10 +233,11 @@ type::MultiPolygon build_single_isochrone(RAPTOR& raptor,
             circles_classed.push_back(to_add);
         }
     }
-    boost::sort(circles_classed,
+    std::vector<InfoCircle> circles_check = delete_useless_circle(circles_classed, speed);
```
xlqian: it seems that `circles_classed` is consumed by `delete_useless_circle`, so you can do `auto circles_check = delete_useless_circle(std::move(circles_classed), speed);`

source/routing/isochrone.cpp
```diff
@@ -161,6 +161,39 @@ struct InfoCircle {
...
+    double circle_radius = circle.duration_left * speed;
+    double coslat = cos(circle.center.lat() * type::GeographicalCoord::N_DEG_TO_RAD);
+    return any_of(begin, end, [speed, circle, circle_radius, coslat](const InfoCircle& it) {
+        double it_radius = it.duration_left * speed;
+        return sqrt(circle.center.approx_sqr_distance(it.center, coslat)) + circle_radius < it_radius;
+    });
+}
+
+static std::vector<InfoCircle> delete_useless_circle(std::vector<InfoCircle> circles,
+                                                     const double& speed) {
```
xlqian: & can be removed

source/routing/isochrone.cpp
```diff
@@ -161,6 +161,39 @@ struct InfoCircle {
...
+    std::vector<InfoCircle> circles_utils;
```
xlqian: `circles_utils`? `useful_circles`?

source/routing/isochrone.cpp
```diff
@@ -161,6 +161,39 @@ struct InfoCircle {
...
+    if (!circles.empty()) {
+        boost::sort(circles,
+                    [](const InfoCircle& a, const InfoCircle& b) {return a.duration_left > b.duration_left;});
+        auto it = circles.begin();
```
xlqian: prefer using `std::begin()` as suggested above :)

source/routing/isochrone.cpp
TeXitoi, line 168: That's better in generic programming, but in this context, I find `.begin()` and `.end()` more readable.

source/routing/isochrone.cpp
TeXitoi: You can `[=](){}` or `[&](){}` when there is lots of arguments.

source/routing/isochrone.cpp
```diff
@@ -161,6 +161,39 @@ struct InfoCircle {
...
+        circles_utils.push_back(*it);
+        const auto end = circles.end();
+        ++it;
+        for (; it != end; ++it) {
+            if (!within_info_circle(*it, circles_utils, speed)) {
+                circles_utils.push_back(*it);
```
xlqian: i don't know if it'd be better to move instead of copy? `circles_utils.push_back(std::move(*it));`@TeXitoi

source/routing/isochrone.cpp
```diff
@@ -161,6 +161,39 @@ struct InfoCircle {
         duration_left(duration_left) {}
 };
 
+static bool within_info_circle(const InfoCircle& circle,
+                              const std::vector<InfoCircle>& multi_poly,
+                              const double& speed) {
+    if (multi_poly.empty()) {return false;}
```
TeXitoi: You don't need a special case.

source/routing/isochrone.cpp
xlqian: `std::move` won't invalid iterators...

source/routing/isochrone.cpp
xlqian: You can `[=](){}` or `[&](){}` when you want to capture **ALL** by copy or ref :)

source/routing/isochrone.cpp
```diff
@@ -161,6 +161,39 @@ struct InfoCircle {
...
+    return any_of(begin, end, [speed, circle, circle_radius, coslat](const InfoCircle& it) {
+        double it_radius = it.duration_left * speed;
+        return sqrt(circle.center.approx_sqr_distance(it.center, coslat)) + circle_radius < it_radius;
+    });
+}
+
+static std::vector<InfoCircle> delete_useless_circle(std::vector<InfoCircle> circles,
+                                                     const double& speed) {
+    std::vector<InfoCircle> circles_utils;
+    if (!circles.empty()) {
```
TeXitoi: You don't need special case

source/routing/isochrone.cpp
xlqian: it'd be better that you 'move' instead of 'copy' http://ideone.com/VK5tEi we can discuss about it...

source/routing/isochrone.cpp
```diff
@@ -161,6 +161,36 @@ struct InfoCircle {
...
+static std::vector<InfoCircle> delete_useless_circle(std::vector<InfoCircle> circles,
+                                                     const double speed) {
+    std::vector<InfoCircle> useful_circles;
+    boost::sort(circles,
+                [](const InfoCircle& a, const InfoCircle& b) {return a.duration_left > b.duration_left;});
+    auto it = circles.begin();
+    useful_circles.push_back(*it);
+    const auto end = circles.end();
+    ++it;
+    for (; it != end; ++it) {
```
TeXitoi: You don't need this first element case. any_of with an empty list will return false. Thus, you can just use a range for.

source/routing/isochrone.cpp
```diff
@@ -163,35 +163,30 @@ struct InfoCircle {
...
+    std::vector<InfoCircle> useful_circles;
+    boost::sort(circles,
+                [](const InfoCircle& a, const InfoCircle& b) {return a.duration_left > b.duration_left;});
+    auto it = circles.begin();
+    const auto end = circles.end();
+    for (; it != end; ++it) {
+        if (!within_info_circle(*it, useful_circles, speed)) {
+            useful_circles.push_back(std::move(*it));
         }
     }
```
TeXitoi, line 186:
``` c++
for (auto& circle: circles) {
    if (!within_info_circle(circle, useful_circles, speed)) {
        useful_circles.push_back(std::move(circle));
    }
}
```

source/routing/isochrone.cpp
```diff
@@ -163,35 +163,30 @@ struct InfoCircle {
...
                               const std::vector<InfoCircle>& multi_poly,
-                              const double& speed) {
-    if (multi_poly.empty()) {return false;}
+                              const double speed) {
     auto begin = multi_poly.begin();
     const auto end = multi_poly.end();
     double circle_radius = circle.duration_left * speed;
     double coslat = cos(circle.center.lat() * type::GeographicalCoord::N_DEG_TO_RAD);
-    return any_of(begin, end, [speed, circle, circle_radius, coslat](const InfoCircle& it) {
+    return any_of(begin, end, [=](const InfoCircle& it) {
```
TeXitoi: That's OK for this case. In general, that's better to use `[&](){}`

PR #1678, opened 2016-07-05
jormungandr: add ecologic tag in scenario new_default

TeXitoi:
To help the understanding of CO2 emission, we propose to add a picto on the journey summary if the journey is ecologic.

An ecologic journey is ecologic if it emits less than 50% of the journey by car.

This first version only tags the journey if we have a direct path by car. In a followup PR, we'll add an approximation based on the flying journey.

### DIFF
--- a/source/jormungandr/jormungandr/scenarios/helpers.py
+++ b/source/jormungandr/jormungandr/scenarios/helpers.py
@@ -185,6 +185,17 @@ def is_non_pt_walk(journey):
 def is_non_pt_bike(journey):
     return journey.type == 'non_pt_bike'
 
+def is_car_direct_path(journey):
+    for section in journey.sections:
+        if section.type not in [response_pb2.STREET_NETWORK, response_pb2.PARK,
+                                response_pb2.LEAVE_PARKING]:
+            return False
+        if section.type != response_pb2.STREET_NETWORK:
+            continue
+        if section.street_network.mode not in [response_pb2.Walking, response_pb2.Car]:
+            return False
+    return True
+
 max_duration_fallback_modes = {'walking': [response_pb2.Walking],
                                'bss': [response_pb2.Walking, response_pb2.Bss],
                                'bike': [response_pb2.Walking, response_pb2.Bss, response_pb2.Bike],
--- a/source/jormungandr/jormungandr/scenarios/new_default.py
+++ b/source/jormungandr/jormungandr/scenarios/new_default.py
@@ -32,7 +32,7 @@ from copy import deepcopy
 import itertools
 import logging
 from flask.ext.restful import abort
-from jormungandr.scenarios import simple, journey_filter
+from jormungandr.scenarios import simple, journey_filter, helpers
 from jormungandr.scenarios.utils import journey_sorter, change_ids, updated_request_with_default, get_or_default, \
     fill_uris, gen_all_combin, get_pseudo_duration
 from navitiacommon import type_pb2, response_pb2, request_pb2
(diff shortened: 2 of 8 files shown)


source/jormungandr/jormungandr/scenarios/new_default.py
```diff
@@ -620,6 +642,7 @@ def fill_journeys(self, request_type, api_request, instance):
 
         sort_journeys(pb_resp, instance.journey_order, api_request['clockwise'])
         tag_journeys(pb_resp)
+        journey_filter.delete_journeys((pb_resp,), api_request)
```
kinnou02, line 633: The next calls to kraken would be create when we haven't yet deleted the filtered journeys, there is a risk that the next call will be based on a "invalid" journey.

kinnou02:
I will need a little context for understand this PR. What is the purpose of this change?

source/jormungandr/jormungandr/scenarios/new_default.py
```diff
@@ -156,11 +156,33 @@ def sort_journeys(resp, journey_order, clockwise):
         resp.journeys.sort(journey_sorter[journey_order](clockwise=clockwise))
 
 
+def _is_car_direct_path(journey):
```
kinnou02: A lot of function of this kind are already in scenarios/helpers.py

TeXitoi:
Description of the PR updated to explain the goal of this PR.

source/jormungandr/jormungandr/scenarios/new_default.py
TeXitoi, line 633: We corrected it.

source/jormungandr/jormungandr/scenarios/new_default.py
```diff
@@ -156,11 +156,33 @@ def sort_journeys(resp, journey_order, clockwise):
...
 def tag_journeys(resp):
     """
     qualify the journeys
     """
-    pass
+    car = next((j for j in resp.journeys if _is_car_direct_path(j)), None)
+    if car is None:
+        return
+    for j in resp.journeys:
+        if not j.HasField('co2_emission'):
```
kinnou02: the ```co2_emission``` of a physical mode is operational in ntfs, what happen if it's not set?

source/jormungandr/jormungandr/scenarios/new_default.py
```diff
@@ -160,7 +160,17 @@ def tag_journeys(resp):
     """
     qualify the journeys
     """
-    pass
+    car = next((j for j in resp.journeys if helpers.is_car_direct_path(j)), None)
+    if car is None:
+        return
+    for j in resp.journeys:
+        if not j.HasField('co2_emission'):
```
kinnou02, line 167: the `co2_emission` of a physical mode is optional in ntfs, what happen if it's not set?

source/jormungandr/jormungandr/scenarios/helpers.py
```diff
@@ -185,6 +185,17 @@ def is_non_pt_walk(journey):
 def is_non_pt_bike(journey):
     return journey.type == 'non_pt_bike'
 
+def is_car_direct_path(journey):
```
kinnou02, line 188: This function should be tested

source/jormungandr/jormungandr/scenarios/new_default.py
TeXitoi, line 167: I'll add a test for the field in the car section. If there is no data for car, we'll do not add any ecologic tag. Thus, if you add the CO2 for car, you'll need to add it to all the non 0 other physical mode.

kinnou02, line 167: seems fair

source/jormungandr/jormungandr/scenarios/helpers.py
TeXitoi, line 188: done

TeXitoi:
@kinnou02 OK?

kinnou02:
Ok for me

TeXitoi:
Then I think you can merge ;-)

PR #1680, opened 2016-07-06
tyr: add the ability to no wait for the end of a import_last_dataset command

kinnou02:
It wasn't possible to run the jobs on the workers and to wait for them
to finish. So in production all job were only execute by tyr1 :(


TeXitoi:
OK for me.

PR #1682, opened 2016-07-07
Jorm: add an API "geo_status" providing stats on georef data

kinnou02:
This PR is related to http://jira.canaltp.fr/browse/NAVITIAII-2150
The goal of this API is to provide statistics on georef data, it isn't
at destination of our users but more for the data team, it will help
them see the impacts of their update

Maybe georef_stats would be better
require: https://github.com/CanalTP/navitia-proto/pull/63

More is coming

### DIFF
--- a/source/jormungandr/jormungandr/interfaces/v1/GeoStatus.py
+++ b/source/jormungandr/jormungandr/interfaces/v1/GeoStatus.py
@@ -38,7 +38,7 @@ geo_status = {
             'nb_admins_from_cities': fields.Raw,
             'nb_ways': fields.Raw,
             'nb_addresses': fields.Raw,
-            'nb_poi': fields.Raw,
+            'nb_pois': fields.Raw,
             'poi_sources': fields.List(fields.String),
         })
 }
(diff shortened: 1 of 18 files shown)


source/jormungandr/jormungandr/interfaces/v1/GeoStatus.py
```diff
@@ -0,0 +1,51 @@
...
+            'nb_ways': fields.Raw,
+            'nb_addresses': fields.Raw,
+            'nb_poi': fields.Raw,
+        })
+}
+
+class GeoStatus(Resource):
+    @marshal_with(geo_status)
+    def get(self, region):
+        response = protobuf_to_dict(i_manager.dispatch({}, "geo_status", instance_name=region), use_enum_labels=True)
```
antoine-de: why do you need this ? can't you directly marshal the protobuf ?

antoine-de:
you have conflicts

antoine-de:
OK for me

antoine-de:
you need to rebase

kinnou02:
I have rename the api in `_geo_status` since it's a private api as we can say

PR #1688, opened 2016-07-13
Isochrone : fix bug section_mode

ZoeBrunet:
Change the radius of circles if the section_mode[] change

### DIFF
--- a/source/kraken/worker.cpp
+++ b/source/kraken/worker.cpp
@@ -746,13 +746,26 @@ pbnavitia::Response Worker::graphical_isochrone(const pbnavitia::GraphicalIsochr
     for(int i = 0; i < request_iso.boundary_duration_size(); ++i) {
         boundary_duration.push_back(request_iso.boundary_duration(i));
     }
+    auto streetnetwork = request.streetnetwork_params();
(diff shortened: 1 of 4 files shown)


source/kraken/worker.cpp
```diff
@@ -746,13 +746,26 @@ pbnavitia::Response Worker::graphical_isochrone(const pbnavitia::GraphicalIsochr
     for(int i = 0; i < request_iso.boundary_duration_size(); ++i) {
         boundary_duration.push_back(request_iso.boundary_duration(i));
     }
+    auto streetnetwork = request.streetnetwork_params();
```
xlqian, line 749: auto&

source/kraken/worker.cpp
```diff
@@ -746,13 +746,26 @@ pbnavitia::Response Worker::graphical_isochrone(const pbnavitia::GraphicalIsochr
...
+    auto mode_iso = request.clockwise() ? streetnetwork.destination_mode() : streetnetwork.origin_mode();
```
xlqian, line 750: auto&

TeXitoi:
Finally, what we said about that?

PR #1689, opened 2016-07-13
All: expose car_co2_emission in context

xlqian:
http://jira.canaltp.fr/browse/ITI-17

display car co2 emission in the response API

which gives:

``` json
{
    "context": {
        "car_direct_path" : {
            "co2_emission": {
                "value": 42,
                "unit": "gEC"
            }
        }
    },
    "disruptions": [],
    "exceptions": [],
    "feed_publishers": [],
    "journeys": [ ... ],
    "links": [ ... ],
    "notes": [],
    "tickets": []
}
```

### DIFF
--- a/source/jormungandr/jormungandr/georef.py
+++ b/source/jormungandr/jormungandr/georef.py
@@ -61,4 +61,14 @@ class Kraken(object):
         response = self.instance.send_and_receive(req)
         return response.places[0]
 
+    def get_car_co2_emission_on_crow_fly(self, origin, destination):
+        req = request_pb2.Request()
+        req.requested_api = type_pb2.car_co2_emission
+        req.car_co2_emission.origin.place = origin
+        req.car_co2_emission.origin.access_duration = 0
+        req.car_co2_emission.destination.place = destination
+        req.car_co2_emission.destination.access_duration = 0
+
+        response = self.instance.send_and_receive(req)
+        return response.car_co2_emission
 
--- a/source/jormungandr/jormungandr/georef.py
+++ b/source/jormungandr/jormungandr/georef.py
@@ -28,7 +28,7 @@
 # www.navitia.io
 
 from __future__ import absolute_import, print_function, unicode_literals, division
-from navitiacommon import request_pb2, type_pb2
+from navitiacommon import request_pb2, response_pb2, type_pb2
 
 class Kraken(object):
 
@@ -70,5 +70,8 @@ class Kraken(object):
         req.car_co2_emission.destination.access_duration = 0
 
         response = self.instance.send_and_receive(req)
+        if response.error and response.error.id == \
+                response_pb2.Error.error_id.Value('no_solution'):
+            return None
         return response.car_co2_emission
 
--- a/source/jormungandr/jormungandr/georef.py
+++ b/source/jormungandr/jormungandr/georef.py
@@ -29,6 +29,7 @@
 
 from __future__ import absolute_import, print_function, unicode_literals, division
 from navitiacommon import request_pb2, response_pb2, type_pb2
+import logging
 
 class Kraken(object):
 
@@ -62,6 +63,7 @@ class Kraken(object):
         return response.places[0]
 
     def get_car_co2_emission_on_crow_fly(self, origin, destination):
+        logger = logging.getLogger(__name__)
         req = request_pb2.Request()
         req.requested_api = type_pb2.car_co2_emission
         req.car_co2_emission.origin.place = origin
@@ -72,6 +74,8 @@ class Kraken(object):
         response = self.instance.send_and_receive(req)
         if response.error and response.error.id == \
                 response_pb2.Error.error_id.Value('no_solution'):
+            logger.error("Cannot compute car co2 emission from {} to {}"
+                         .format(origin, destination))
             return None
         return response.car_co2_emission
 
--- a/source/jormungandr/jormungandr/interfaces/v1/Journeys.py
+++ b/source/jormungandr/jormungandr/interfaces/v1/Journeys.py
@@ -38,7 +38,7 @@ from jormungandr import i_manager, app
 from jormungandr.interfaces.v1.fields import disruption_marshaller, Links
 from jormungandr.interfaces.v1.fields import display_informations_vj, error, place,\
     PbField, stop_date_time, enum_type, NonNullList, NonNullNested,\
-    SectionGeoJson, Co2Emission, PbEnum, feed_publisher, Durations
+    SectionGeoJson, PbEnum, feed_publisher, Durations
 
 from jormungandr.interfaces.parsers import option_value, date_time_format, default_count_arg_type, date_time_format
 from jormungandr.interfaces.v1.ResourceUri import ResourceUri, complete_links
@@ -194,7 +194,10 @@ section = {
     "base_departure_date_time": DateTime(attribute="base_begin_date_time"),
     "arrival_date_time": DateTime(attribute="end_date_time"),
     "base_arrival_date_time": DateTime(attribute="base_end_date_time"),
-    "co2_emission": Co2Emission(),
+    "co2_emission": NonNullNested({
+        'value': fields.Raw,
+        'unit': fields.String
+        }),
 }
 
 cost = {
@@ -222,7 +225,10 @@ journey = {
     'tags': fields.List(fields.String),
     "status": fields.String(attribute="most_serious_disruption_effect"),
     "calendars": NonNullList(NonNullNested(calendar)),
-    "co2_emission": Co2Emission(),
+    "co2_emission": NonNullNested({
+        'value': fields.Raw,
+        'unit': fields.String
+        }),
     "durations": Durations(),
     "debug": JourneyDebugInfo()
 }
--- a/source/jormungandr/jormungandr/interfaces/v1/Journeys.py
+++ b/source/jormungandr/jormungandr/interfaces/v1/Journeys.py
@@ -237,10 +237,12 @@ ticket = {
 }
 
 context = {
-    'car_co2_emission': NonNullNested({
-        'value': fields.Raw,
-        'unit': fields.String
-    })
+    'car_direct_path': {
+        'co2_emission': NonNullNested({
+            'value': fields.Raw,
+            'unit': fields.String
+        }, attribute="car_co2_emission")
+    }
 }
 
 journeys = {
--- a/source/jormungandr/jormungandr/interfaces/v1/Journeys.py
+++ b/source/jormungandr/jormungandr/interfaces/v1/Journeys.py
@@ -236,6 +236,9 @@ ticket = {
     "links": TicketLinks(attribute="section_id")
 }
 
+context = {
+    "car_co2_emission": Co2Emission()
+}
 
 journeys = {
     "journeys": NonNullList(NonNullNested(journey)),
@@ -244,6 +247,7 @@ journeys = {
     "disruptions": fields.List(NonNullNested(disruption_marshaller), attribute="impacts"),
     "feed_publishers": fields.List(NonNullNested(feed_publisher)),
     "links": fields.List(Links()),
+    "context": context,
 }
 
 
--- a/source/jormungandr/jormungandr/interfaces/v1/Journeys.py
+++ b/source/jormungandr/jormungandr/interfaces/v1/Journeys.py
@@ -237,7 +237,10 @@ ticket = {
 }
 
 context = {
-    "car_co2_emission": Co2Emission()
+    'car_co2_emission': NonNullNested({
+        'value': fields.Raw,
+        'unit': fields.String
+    })
 }
 
 journeys = {
--- a/source/jormungandr/jormungandr/interfaces/v1/fields.py
+++ b/source/jormungandr/jormungandr/interfaces/v1/fields.py
@@ -426,18 +426,6 @@ class MultiPolyGeoJson(fields.Raw):
 
         return response
 
-
-class Co2Emission(fields.Raw):
-    def output(self, key, obj):
-        if not obj.HasField(key):
-            return
-        co2_emission = getattr(obj, key)
-        return {
-            'value': co2_emission.value,
-            'unit': co2_emission.unit
-        }
-
-
 class Durations(fields.Raw):
     def output(self, key, obj):
         if not obj.HasField(b"durations"):
--- a/source/jormungandr/jormungandr/interfaces/v1/fields.py
+++ b/source/jormungandr/jormungandr/interfaces/v1/fields.py
@@ -429,11 +429,12 @@ class MultiPolyGeoJson(fields.Raw):
 
 class Co2Emission(fields.Raw):
     def output(self, key, obj):
-        if not obj.HasField(b"co2_emission"):
+        if not obj.HasField(key):
             return
+        co2_emission = getattr(obj, key)
         return {
-            'value': obj.co2_emission.value,
-            'unit': obj.co2_emission.unit
+            'value': co2_emission.value,
+            'unit': co2_emission.unit
         }
 
 
--- a/source/jormungandr/jormungandr/scenarios/new_default.py
+++ b/source/jormungandr/jormungandr/scenarios/new_default.py
@@ -177,18 +177,15 @@ def tag_journeys(resp):
     tag the journeys
     """
     # if there is no available car_co2_emission in resp, no tag will be assigned
-    if not resp.car_co2_emission.value and \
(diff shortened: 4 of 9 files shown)


source/jormungandr/jormungandr/scenarios/new_default.py
```diff
@@ -156,20 +156,36 @@ def sort_journeys(resp, journey_order, clockwise):
...
-        return
+        # if there is no car journey found, we request kraken to give us an estimation of
+        # co2 emission
+        from jormungandr.georef import Kraken
+        co2_estimation = Kraken(instance).get_car_co2_emission_on_crow_fly(request['origin'], request['destination'])
+        co2_emission_value = co2_estimation.value
+        co2_emission_unit = co2_estimation.unit
+    else:
+        co2_emission_value = car.co2_emission.value
+        co2_emission_unit = car.co2_emission.unit
```
TeXitoi: I think this should not be done here: first we should compute the `context.car_co2_emission`, and then use it here.

source/jormungandr/jormungandr/scenarios/new_default.py
```diff
@@ -156,20 +156,36 @@ def sort_journeys(resp, journey_order, clockwise):
...
             j.tags.append('ecologic')
             continue
-        if j.co2_emission.unit != car.co2_emission.unit:
+        if j.co2_emission.unit != co2_emission_unit:
             continue
-        if j.co2_emission.value < car.co2_emission.value * 0.5:
+
+        # Assign car_co2_emission into the resp, these value will be exposed in the final result
+        resp.car_co2_emission.value = co2_emission_value
+        resp.car_co2_emission.unit = co2_emission_unit
```
TeXitoi: Doing that for every journey is just useless.

source/jormungandr/jormungandr/scenarios/new_default.py
TeXitoi: We must take care of the case "there is no data for co2 for car. In this case, there is no tag.

source/jormungandr/tests/routing_tests_new_default.py
```diff
@@ -161,6 +161,42 @@ def test_ecologic_tag_with_car(self):
...
+                .format(from_coord=s_coord, to_coord=r_coord, datetime="20120614T075500")
+        response = self.query_region(query)
+        check_journeys(response)
+        assert len(response["journeys"]) == 2
+
+        assert 'context' in response
+        assert 'car_co2_emission' in response['context']
+
+        assert response['context']['car_co2_emission']['value'] == 34.224
+        assert response['context']['car_co2_emission']['unit'] == 'gEC'
```
TeXitoi: Lack tests of the tag in this case.

source/kraken/worker.cpp
```diff
@@ -737,6 +737,38 @@ pbnavitia::Response Worker::graphical_isochrone(const pbnavitia::GraphicalIsochr
 
 }
 
+pbnavitia::Response Worker::car_co2_emission_on_crow_fly(const pbnavitia::CarCO2EmissionRequest& request) {
+    const auto data = data_manager.get_data();
+    init_worker_data(data);
+
+    auto get_geographical_coord = [&](const pbnavitia::LocationContext& location){
+        std::cout << location.place() << std::endl;
```
TeXitoi: to remove

source/kraken/worker.cpp
```diff
@@ -737,6 +737,38 @@ pbnavitia::Response Worker::graphical_isochrone(const pbnavitia::GraphicalIsochr
...
+        auto origin_type = data->get_type_of_id(location.place());
+        auto origin = type::EntryPoint(origin_type, location.place(), location.access_duration());
+
+        if (origin_type == type::Type_e::Address || origin_type == type::Type_e::Admin
+                || origin_type == type::Type_e::StopArea || origin_type == type::Type_e::StopPoint
+                || origin_type == type::Type_e::POI) {
+            auto coordinates = coord_of_entry_point(origin, data);
+            return navitia::type::GeographicalCoord{coordinates.lon(), coordinates.lat()};
+        }
+        return navitia::type::GeographicalCoord{};
```
TeXitoi: the test is useless, you can directly call `coord_of_entry_point`.

source/kraken/worker.cpp
```diff
@@ -737,6 +737,38 @@ pbnavitia::Response Worker::graphical_isochrone(const pbnavitia::GraphicalIsochr
...
+    };
+    auto origin = get_geographical_coord(request.origin());
+    auto destin = get_geographical_coord(request.destination());
```
TeXitoi, line 771: If the coordinates are invalid, we must not compute the co2

source/kraken/worker.cpp
```diff
@@ -737,6 +737,38 @@ pbnavitia::Response Worker::graphical_isochrone(const pbnavitia::GraphicalIsochr
...
+    auto distance = origin.distance_to(destin);
+    pbnavitia::Response r;
+    auto co2_emission = r.mutable_car_co2_emission();
```
TeXitoi: better to create the object only if we compute it.

source/kraken/worker.cpp
```diff
@@ -737,6 +737,38 @@ pbnavitia::Response Worker::graphical_isochrone(const pbnavitia::GraphicalIsochr
...
+
+    if (data->pt_data->physical_modes_map["physical_mode:Car"]->co2_emission) {
+        const double estimation_coeff = 1.3;
```
TeXitoi: 1.35

source/kraken/worker.cpp
```diff
@@ -737,6 +737,38 @@ pbnavitia::Response Worker::graphical_isochrone(const pbnavitia::GraphicalIsochr
...
+        co2_emission->set_unit("gEC");
+        co2_emission->set_value(estimation_coeff * distance / 1000.0 * data->pt_data->physical_modes_map["physical_mode:Car"]->co2_emission.get());
```
TeXitoi: long line

source/kraken/worker.cpp
```diff
@@ -737,6 +737,38 @@ pbnavitia::Response Worker::graphical_isochrone(const pbnavitia::GraphicalIsochr
...
+    }
+    return r;
+
```
TeXitoi: useless empty line.

source/kraken/worker.cpp
```diff
@@ -777,6 +809,7 @@ pbnavitia::Response Worker::dispatch(const pbnavitia::Request& request) {
     case pbnavitia::place_code : response = place_code(request.place_code()); break;
     case pbnavitia::nearest_stop_points : response = nearest_stop_points(request.nearest_stop_points()); break;
     case pbnavitia::graphical_isochrone : response = graphical_isochrone(request.isochrone(), current_datetime); break;
+    case pbnavitia::car_co2_emission: response = car_co2_emission_on_crow_fly(request.car_co2_emission()); break;
```
TeXitoi: long line

source/kraken/worker.h
```diff
@@ -120,6 +120,8 @@ class Worker {
         pbnavitia::Response nearest_stop_points(const pbnavitia::NearestStopPointsRequest& request);
         pbnavitia::Response graphical_isochrone(const pbnavitia::GraphicalIsochroneRequest &request,
                                                const boost::posix_time::ptime& current_datetime);
+        pbnavitia::Response car_co2_emission_on_crow_fly(const pbnavitia::CarCO2EmissionRequest& request);
```
TeXitoi, line 122: long line

source/kraken/worker.h
```diff
@@ -120,6 +120,8 @@ class Worker {
...
+
```
TeXitoi: useless empty line

TeXitoi:
You have conflicts

source/jormungandr/jormungandr/scenarios/new_default.py
```diff
@@ -156,20 +156,36 @@ def sort_journeys(resp, journey_order, clockwise):
...
     """
     car = next((j for j in resp.journeys if helpers.is_car_direct_path(j)), None)
+    co2_emission_value = 0
+    co2_emission_unit = 0
     if car is None or not car.HasField('co2_emission'):
-        return
+        # if there is no car journey found, we request kraken to give us an estimation of
+        # co2 emission
+        from jormungandr.georef import Kraken
+        co2_estimation = Kraken(instance).get_car_co2_emission_on_crow_fly(request['origin'], request['destination'])
```
kinnou02: The instance should already have a georef object, if not you need to add it. T
here Is no point to abstract kraken if we instantiate the class every time we need it.

source/jormungandr/jormungandr/georef.py
```diff
@@ -61,4 +61,14 @@ def place(self, place):
...
+    def get_car_co2_emission_on_crow_fly(self, origin, destination):
+        req = request_pb2.Request()
+        req.requested_api = type_pb2.car_co2_emission
+        req.car_co2_emission.origin.place = origin
+        req.car_co2_emission.origin.access_duration = 0
+        req.car_co2_emission.destination.place = destination
+        req.car_co2_emission.destination.access_duration = 0
+
+        response = self.instance.send_and_receive(req)
+        return response.car_co2_emission
```
kinnou02, line 80: Since it's a simple response you could use a custom structure for the response, this way it will be easier when we will move this feature from kraken.
Or not, you are already returning only the co2_emission part, it will be easy to comform to this "interface" latter.

source/kraken/worker.cpp
```diff
@@ -737,6 +737,38 @@ pbnavitia::Response Worker::graphical_isochrone(const pbnavitia::GraphicalIsochr
...
+        }
+        return navitia::type::GeographicalCoord{};
+    };
+    auto origin = get_geographical_coord(request.origin());
+    auto destin = get_geographical_coord(request.destination());
+    auto distance = origin.distance_to(destin);
+    pbnavitia::Response r;
+    auto co2_emission = r.mutable_car_co2_emission();
+
+    if (data->pt_data->physical_modes_map["physical_mode:Car"]->co2_emission) {
```
kinnou02: The physical_mode "physical_mode:Car" always exist? I don't think that we create it automatically.

source/jormungandr/jormungandr/scenarios/new_default.py
xlqian: ok, I see

source/kraken/worker.cpp
xlqian, line 771: How can you tell if coordinates are valid? By testing its type?

xlqian, line 771: EntryPoint raises indeed an exception when coordinates are invalid... but the exception is caught inside of the constructor...

source/kraken/worker.cpp
xlqian: I checked that with @nlehuby that "physical_mode:Car" should always exist **in theory**

source/kraken/worker.h
xlqian, line 122: Not too long on my screen and IDE :)

source/kraken/worker.cpp
kinnou02: it should exist for data provided by fusio, sometime we don't use fusio, typically for opendata on country not using latin alphabet since fusio can't support these.
So I propose to deal with **the reality**.

TeXitoi: There is no such physical_mode if we import a GTFS. The actual solution:

https://github.com/CanalTP/navitia/blob/dev/source/type/pb_converter.cpp#L1681
https://github.com/CanalTP/navitia/blob/dev/source/type/pb_converter.cpp#L1377

xlqian:
@kinnou02 Can you take a look at this PR plz? thx

source/jormungandr/jormungandr/georef.py
```diff
@@ -61,4 +61,17 @@ def place(self, place):
...
+    def get_car_co2_emission_on_crow_fly(self, origin, destination):
+        req = request_pb2.Request()
+        req.requested_api = type_pb2.car_co2_emission
+        req.car_co2_emission.origin.place = origin
+        req.car_co2_emission.origin.access_duration = 0
+        req.car_co2_emission.destination.place = destination
+        req.car_co2_emission.destination.access_duration = 0
+
+        response = self.instance.send_and_receive(req)
+        if response.error and response.error.id == \
```
antoine-de, line 75: don't you want to log the errors ?

source/jormungandr/jormungandr/interfaces/v1/Journeys.py
```diff
@@ -236,6 +236,9 @@ def output(self, key, obj):
     "links": TicketLinks(attribute="section_id")
 }
 
+context = {
+    "car_co2_emission": Co2Emission()
```
antoine-de: do you really need this ?
doesn't it work with only a flask dict ?
like (speudo code):

```
"car_co2_emission": NonNullNester({
'value': fields.String,
)}
...
```

source/jormungandr/jormungandr/scenarios/new_default.py
```diff
@@ -156,20 +156,38 @@ def sort_journeys(resp, journey_order, clockwise):
...
+
 def tag_journeys(resp):
     """
-    qualify the journeys
+    tag the journeys
     """
-    car = next((j for j in resp.journeys if helpers.is_car_direct_path(j)), None)
-    if car is None or not car.HasField('co2_emission'):
+    # if there is no available car_co2_emission in resp, no tag will be assigned
+    if not resp.car_co2_emission.value and \
```
antoine-de: isn't it very co2 tag specific ?
since we have only those tag it works for the moment, but I think it's counter intuitive, I would have nested the following for loop in the `if resp.car_co2_emission.value and resp.car_co2_emission.unit:` condition for this to be clearer

source/jormungandr/tests/places_test.py
```diff
@@ -157,3 +157,10 @@ def test_wrong_places_nearby(self):
 
         eq_(status, 404)
         assert(get_not_null(response, 'message'))
+
+    def test_non_existent_places_nearby(self):
```
antoine-de, line 161: :+1:

antoine-de:
I'm not sure `context/car_co2_emission` at the `journeys` level is straightforward.

at least wouldn't it be better to nest it inside a `car`, `car_direct_path`, `personal_car` or `full_car` section ?
this would allow us to add more information later on (like estimated duration ? / price ?)

source/kraken/worker.cpp
```diff
@@ -756,6 +755,31 @@ pbnavitia::Response Worker::graphical_isochrone(const pbnavitia::GraphicalIsochr
...
+        auto origin_type = data->get_type_of_id(location.place());
+        auto origin = type::EntryPoint{origin_type, location.place(), location.access_duration()};
+        auto coordinates = coord_of_entry_point(origin, data);
+        return navitia::type::GeographicalCoord{coordinates.lon(), coordinates.lat()};
+    };
+    auto origin = get_geographical_coord(request.origin());
+    auto destin = get_geographical_coord(request.destination());
+    auto distance = origin.distance_to(destin);
+    pbnavitia::Response r;
+    if (data->pt_data->physical_modes_map["physical_mode:Car"]->co2_emission) {
```
TeXitoi: It will crash if we don't have physical_mode:Car, no?

TeXitoi:
``` c++
const auto* car = find_or_default(data->pt_data->physical_modes_map, "physical_mode:Car");
if (car && car->co2_emission) {
```

xlqian: it won't crash but create a new entry in the map...

xlqian: not great either... :)

TeXitoi: It'll create a new entry with a nullptr and then dereference it... so crash.

source/jormungandr/jormungandr/georef.py
TeXitoi, line 75: error logged

source/jormungandr/jormungandr/interfaces/v1/fields.py
```diff
@@ -429,11 +429,12 @@ def format(self, value):
 
 class Co2Emission(fields.Raw):
```
TeXitoi: Thus, this object can be removed, no?

TeXitoi:
OK for me.

source/jormungandr/jormungandr/interfaces/v1/fields.py
xlqian: this class is still used by co2_emission in journeys, i'll see if i can remove it

antoine-de:
OK for me

PR #1690, opened 2016-07-15
jormungandr: do not filter journeys when adding car

TeXitoi:
See http://jira.canaltp.fr/browse/ITI-7 for the specs.
- [x] add a tag by mode to the journeys
- [x] change the rule

### DIFF
--- a/source/jormungandr/jormungandr/scenarios/journey_filter.py
+++ b/source/jormungandr/jormungandr/scenarios/journey_filter.py
@@ -30,7 +30,7 @@ from __future__ import absolute_import, print_function, unicode_literals, divisi
 import logging
 import itertools
 import datetime
-from jormungandr.scenarios.utils import compare, get_pseudo_duration, get_or_default
+from jormungandr.scenarios.utils import compare, get_pseudo_duration, get_or_default, mode_weight
 from navitiacommon import response_pb2
 from jormungandr.utils import pb_del_if
 
@@ -65,7 +65,7 @@ def filter_journeys(response_list, instance, request):
 
     _filter_similar_vj_journeys(journeys, request)
 
-    _filter_not_coherent_journeys(journeys, instance, request)
+    _filter_too_long_journeys(journeys, instance, request)
 
     _filter_too_long_waiting(journeys, request)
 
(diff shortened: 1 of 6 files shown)


source/jormungandr/jormungandr/scenarios/journey_filter.py
```diff
@@ -270,66 +270,74 @@ def get_nb_connections(journey):
...
-    we check for each journey the difference between the requested datetime and the arrival datetime
-    (the other way around for non clockwise)
+def _get_mode_weight(journey):
+    if 'car' in journey.tags:
+        return 4
+    if 'bike' in journey.tags:
+        return 3
+    if 'bss' in journey.tags:
+        return 2
+    return 1
```
xlqian:
```
_MODE_WEIGHT = {'car': 4, 'bike': 3, 'bss': 2}
def _get_mode_weight(journey):
    return max(_MODE_WEIGHT.get(mode, 1) for mode in journey.tags)
```

source/jormungandr/jormungandr/scenarios/tests/journey_compare_tests.py
```diff
@@ -519,13 +519,85 @@ def test_not_too_late_journeys():
...
+    responses.append(response_pb2.Response())
+    car = responses[-1].journeys.add()
+    car.departure_date_time = 1001042
+    car.arrival_date_time = 1016070 # bike * 2 + 10
+    car.tags.append('car')
+
+    journey_filter._filter_too_long_journeys(list(journeys_gen(responses)),
+                                             MockInstance(), request)
+
+    assert sum('to_delete' not in j.tags for r in responses for j in r.journeys) == 1
```
xlqian: I'd prefer `['to_delete' not in j.tags for r in responses for j in r.journeys].count(True) == 1` but your code is nice, too

source/jormungandr/jormungandr/scenarios/new_default.py
```diff
@@ -424,6 +424,39 @@ def combinations_sorter(v):
     journey_filter.delete_journeys((resp,), request)
 
 
+def _tag_journey_by_mode(journey):
+    mode_weight = {
```
xlqian: hmmm, the mode_weight is used in two places... can we put them in one file?

source/jormungandr/jormungandr/scenarios/journey_filter.py
```diff
@@ -270,66 +270,71 @@ def get_nb_connections(journey):
...
-        comp_func = max
-    keeped = [j for j in journeys if not to_be_deleted(j)]
-    if not keeped:
-        #min and max don't like empty list
-        return
-    asap_journey = comp_func(keeped, key=comp_value)
-
-    for j in journeys:
-        if to_be_deleted(j):
+    for j1 in journeys:
```
xlqian: can we do something like this? 

```
import itertools
for (j1,j2) in itertools.permutations(journeys, 2): 
    if to_be_deleted(j1) or to_be_deleted(j2):
        continue
    if way_later(request, j1, j2):
         mark_as_dead(j1, 'too_long', 'too_long_compared_to_{}'.format(j2.internal_id))
```

xlqian:
ok for me @antoine-de @kinnou02 @pbougue  you want to take a look?

antoine-de:
ok for me

PR #1694, opened 2016-07-20
Enrichment of first/last section mode in new_default scenario

azime:
jira: http://jira.canaltp.fr/browse/ITI-28


TeXitoi:
OK for me. @antoine-de @kinnou02 @pbougue ?

antoine-de:
OK for me

pbougue:
ok, we have to think of bss->walking and walking->bss, but still it's better!

PR #1695, opened 2016-07-21
All: catch exception and don't compute co2 when there is no solution

xlqian:
- catch exception 
- don't compute co2 when there is no solution

### DIFF
--- a/source/kraken/worker.cpp
+++ b/source/kraken/worker.cpp
@@ -85,8 +85,8 @@ static type::GeographicalCoord coord_of_entry_point(
             return poi->second->coord;
         }
     }
-    LOG4CPLUS_ERROR(log4cplus::Logger::getInstance(LOG4CPLUS_TEXT("logger")),
-            "The entry point: " + entry_point.uri + " is not valid");
+    LOG4CPLUS_DEBUG(log4cplus::Logger::getInstance(LOG4CPLUS_TEXT("logger")),
+            "The entry point: " << entry_point.uri << " is not valid");
     throw navitia::recoverable_exception{"The entry point: " + entry_point.uri + " is not valid"};
 }
 
@@ -484,7 +484,7 @@ pbnavitia::Response Worker::place_uri(const pbnavitia::PlaceUriRequest &request,
         type::GeographicalCoord  coord;
         try{
             coord = coord_of_entry_point(ep, data);
-        }catch(const std::exception& e) {
+        }catch(const navitia::recoverable_exception& e) {
             pbnavitia::Response r;
             fill_pb_error(pbnavitia::Error::no_solution, e.what(), r.mutable_error());
             return r;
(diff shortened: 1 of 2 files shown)


xlqian:
I assume that catch the exception is a little gross but at least, it's not too difficult to understand... I've thought about returning an optional value instead of catching exception, but with `navitia::JourneysArg arg` it'll be complicated

source/kraken/worker.cpp
```diff
@@ -85,6 +85,8 @@ static type::GeographicalCoord coord_of_entry_point(
             return poi->second->coord;
         }
     }
+    LOG4CPLUS_ERROR(log4cplus::Logger::getInstance(LOG4CPLUS_TEXT("logger")),
```
kinnou02: I don't want error log, it's not an error of the application!

source/kraken/worker.cpp
```diff
@@ -85,6 +85,8 @@ static type::GeographicalCoord coord_of_entry_point(
...
+            "The entry point: " + entry_point.uri + " is not valid");
```
antoine-de: minor: it's better to always use the `<<`

source/kraken/worker.cpp
```diff
@@ -415,7 +417,14 @@ pbnavitia::Response Worker::proximity_list(const pbnavitia::PlacesNearbyRequest
                                            const boost::posix_time::ptime& current_datetime) {
     const auto data = data_manager.get_data();
     type::EntryPoint ep(data->get_type_of_id(request.uri()), request.uri());
-    auto coord = coord_of_entry_point(ep, data);
+    type::GeographicalCoord  coord;
+    try{
+        coord = coord_of_entry_point(ep, data);
+    }catch(const std::exception& e) {
```
TeXitoi: catching such an anonymous exception will cause problem. I'm not fond of this catch everywhere, but at least it's cleaner with a dedicated exception.

source/kraken/worker.cpp
xlqian: OK, changed to a debug log

source/kraken/worker.cpp
xlqian: thx

source/kraken/worker.cpp
xlqian: done

xlqian:
Jenkins, test this please!

source/kraken/worker.cpp
```diff
@@ -415,7 +417,14 @@ pbnavitia::Response Worker::proximity_list(const pbnavitia::PlacesNearbyRequest
                                            const boost::posix_time::ptime& current_datetime) {
     const auto data = data_manager.get_data();
     type::EntryPoint ep(data->get_type_of_id(request.uri()), request.uri());
-    auto coord = coord_of_entry_point(ep, data);
+    type::GeographicalCoord  coord;
+    try{
+        coord = coord_of_entry_point(ep, data);
+    }catch(const navitia::recoverable_exception& e) {
```
TeXitoi: `recoverable_exception` can be something else. You'll need a dedicated type.

TeXitoi:
OK for me.

PR #1701, opened 2016-07-26
Ed: Gtfs fix

pbougue:
fix for direction as it is not required
fix the way bad stop-times are not handled

TODO (in a perfect world):
- [X] tests (done manually)

### DIFF
--- a/source/ed/connectors/gtfs_parser.cpp
+++ b/source/ed/connectors/gtfs_parser.cpp
@@ -975,11 +975,19 @@ void StopTimeGtfsHandler::finish(Data& data) {
         for (auto it_next = it_st + 1; it_next != vj->stop_time_list.end(); it_next++, it_st ++) {
             const auto* st1 = *it_st;
             const auto* st2 = *it_next;
+            bool is_valid_vj = true;
             if (! (st1->arrival_time <= st1->departure_time &&
                   st1->departure_time <= st2->arrival_time &&
                   st2->arrival_time <= st2->departure_time)) {
-                LOG4CPLUS_INFO(logger, "invalid vj " << vj->uri << ", the stop times "
-                                        "are not correcly ordered "
+                is_valid_vj = false;
+                LOG4CPLUS_INFO(logger, "invalid vj " << vj->uri << ", stop times are not correctly ordered");
+            }
+            if (st1->arrival_time < 0 || st1->departure_time < 0
+                    || st2->arrival_time < 0 || st2->departure_time < 0) {
+                LOG4CPLUS_INFO(logger, "invalid vj " << vj->uri << ", stop times are not correct");
+            }
+            if (! is_valid_vj) {
+                LOG4CPLUS_INFO(logger, "concerned stop times : "
                                << "stop time " << st1->order << " [" << st1->arrival_time
                                << ", " << st1->departure_time <<"]"
                                << " stop time " << st2->order << " [" << st2->arrival_time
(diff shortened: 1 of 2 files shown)


source/ed/connectors/gtfs_parser.cpp
```diff
@@ -970,11 +975,19 @@ void StopTimeGtfsHandler::finish(Data& data) {
...
                   st1->departure_time <= st2->arrival_time &&
                   st2->arrival_time <= st2->departure_time)) {
-                LOG4CPLUS_INFO(logger, "invalid vj " << vj->uri << ", the stop times "
-                                        "are not correcly ordered "
+                is_valid_vj = false;
+                LOG4CPLUS_INFO(logger, "invalid vj " << vj->uri << ", stop times are not correctly ordered");
+            }
+            if (st1->arrival_time < 0 || st1->departure_time < 0
+                    || st2->arrival_time < 0 || st2->departure_time < 0) {
+                LOG4CPLUS_INFO(logger, "invalid vj " << vj->uri << ", stop times are not correct");
```
antoine-de, line 989: shouldn't you also tag here as invalid_vj ?

source/ed/connectors/gtfs_parser.cpp
```diff
@@ -989,7 +1002,11 @@ void StopTimeGtfsHandler::finish(Data& data) {
 }
 
 static int to_utc(const std::string& local_time, int utc_offset) {
-    return time_to_int(local_time) - utc_offset;
+    int local = time_to_int(local_time);
```
antoine-de, line 1007: I don't understand this.

what is the `-1` value ?

source/ed/connectors/gtfs_parser.cpp
pbougue, line 989: Oh yes, forgot that one :)

source/ed/connectors/gtfs_parser.cpp
pbougue, line 1007: time_to_int returns `-1` when the provided string is not correctly parsed

source/ed/connectors/gtfs_parser.cpp
```diff
@@ -861,7 +861,12 @@ void TripsGtfsHandler::handle_line(Data& data, const csv_row& row, bool) {
 
     nm::Line* line = it->second;
 
-    nm::Route* route = get_or_create_route(data, {line, row[direction_id_c]});
+    // direction_id is optional (and possible values "0" or "1"), so defaulting to "0"
+    std::string direction_id = "0";
```
TeXitoi: As we are a bit laxist here by purpose (you can put anything in direction, we just group by string), I suggest to use the empty string as default value.

pbougue: empty string is not gtfs complient, no?
this was a quick fix, but if both work...

antoine-de: gtfs spec says that it should be either `0` or `1`, but since it is a string, we decided to be a bit more laxist and accept any direction.

With it, even with GTFS, you can create any number of routes for a given line (as we can do with ntfs)

pbougue: ok, I'll change that. It's fine by me, the point is just that it does not crash

source/ed/connectors/gtfs_parser.cpp
pbougue, line 1007: switched it to -infinite because actual times can be negative after utc offset, and also for clarity of debug

antoine-de, line 1007: :+1:

pbougue:
tested locally

PR #1702, opened 2016-07-27
jormun: use dict in place of orderedDict in marshal for improving memory consumption

kinnou02:
Explanation here: https://github.com/CanalTP/flask-restful/pull/1

fix: http://jira.canaltp.fr/browse/NAVITIAII-1522

### DIFF
--- a/source/jormungandr/jormungandr/interfaces/v1/ResourceUri.py
+++ b/source/jormungandr/jormungandr/interfaces/v1/ResourceUri.py
@@ -140,12 +140,14 @@ class add_computed_resources(object):
                     data['links'].append({
                         "href": url_for("v1." + api, **kwargs),
                         "rel": api,
+                        "type": api,
                         "templated": templated
                     })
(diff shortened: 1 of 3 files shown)


source/jormungandr/jormungandr/interfaces/v1/ResourceUri.py
```diff
@@ -140,12 +140,14 @@ def wrapper(*args, **kwargs):
...
                         "href": url_for("v1." + api, **kwargs),
                         "rel": api,
+                        "type": api,
                         "templated": templated
                     })
             if collection in ['stop_areas', 'stop_points', 'addresses']:
                 data['links'].append({
                     "href": url_for("v1.journeys", **kwargs),
                     "rel": "journeys",
+                    "type": "journeys",
```
antoine-de: I think type should be `journey`

antoine-de:
OK for me

PR #1710, opened 2016-08-02
Kraken: fix partial terminus

antoine-de:
in /stop_schedule, the route point  status was computed using all jpp of the stop_point, not only with the route's jpp.

It was thus leading to the first stop of a route being a partial terminus because it was the terminus of the backward route...


antoine-de:
Jenkins, test this please

PR #1711, opened 2016-08-02
type: fix order of stoptime and route

kinnou02:
A few tests were failing randomly on stop_schedule since the tests are
dependant of the order of the response


antoine-de:
OK for me

PR #1712, opened 2016-08-02
proto: specify syntax "proto2" for protocol buffer 3 compatibility

kinnou02:
require https://github.com/CanalTP/chaos-proto/pull/15 and https://github.com/CanalTP/navitia-proto/pull/70

### DIFF
--- a/source/CMakeLists.txt
+++ b/source/CMakeLists.txt
@@ -60,8 +60,8 @@ FIND_PATH(PQXX_INCLUDE_DIR pqxx/pqxx)
 FIND_LIBRARY(PQXX_LIB pqxx)
 
 #add current compilation dir to include path to handle config.h
-include_directories("${CMAKE_CURRENT_BINARY_DIR}"
-                     "${CMAKE_SOURCE_DIR}"
+include_directories(SYSTEM "${CMAKE_CURRENT_BINARY_DIR}")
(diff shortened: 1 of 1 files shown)


kinnou02:
This can be merged :)

source/CMakeLists.txt
```diff
@@ -60,8 +60,8 @@ FIND_PATH(PQXX_INCLUDE_DIR pqxx/pqxx)
 FIND_LIBRARY(PQXX_LIB pqxx)
 
 #add current compilation dir to include path to handle config.h
-include_directories("${CMAKE_CURRENT_BINARY_DIR}"
-                     "${CMAKE_SOURCE_DIR}"
+include_directories(SYSTEM "${CMAKE_CURRENT_BINARY_DIR}")
+include_directories("${CMAKE_SOURCE_DIR}"
```
TeXitoi, line 64: Why 2 include_directories?

kinnou02: Because I don't want our headers to be considered has "system".
Protobuf header generate a lot of warning, this way we don't have to see them every time they are imported.

TeXitoi: OK. And it works?

kinnou02: totally!

TeXitoi: Great! :-)

PR #1713, opened 2016-08-03
Jormun: remove a crash with bss availability when depth=0 on place_nearby

kinnou02:


TeXitoi:
Strange, test are failling

kinnou02:
My bad, I have added a description of the test before committing, and a doc string has to be properly indented too.

PR #1715, opened 2016-08-04
Jormun: Fix 500 error when token contains unicode character

kinnou02:
Base64 doesn't seem to handle unicode character very well since our token can
never contain unicode characters since they are generated from an
uuid we can simply catch the error and return a 401


antoine-de:
great tests :+1:

PR #1717, opened 2016-08-05
Georef: Accurate street network geometries

eturck:
Hi!

I started this long ago, but I finally took the time to finalize it, fix a few bugs and add some unit tests. I know that you are in the process of changing almost everything on the street network routing algorithm so I hope it's not too late...

There is two commits fixing some bugs : 
- In routing_api_test  an edge added in the way was reversed compare to the one added via boost::add_edge. It was breaking the unit test,
- In osm2ed the geometry of the reversed edge wasn't reversed (since it's not currently used except to get its length there is currently no impact).

Otherwise the main changes are : 
- Export geometries of edges in ed2nav. To reduce memory usage we store the geometry once in the way, and reference it on each type of edge (pedestrian, bike, car),
- Improve projection and distance from both vertex by using the geometry of the edge to find the closest point,
- way::make_multiline use them too,
- Better geojson for street network sections : 
  - Each section use the geometry of the edge,
  - When adding custom projection to the path we split the first and last edge geometries and just keep the part before / after the projected point,
  - Handle the case of two point projected on the same edge too. 

As discussed a few month ago, it's completely optional since it's quite memory intensive. You need to pass an argument to ed2nav otherwise geometries are not exported. On our (small) network : 
- ~15k vj, ~350k stop_times,
- OSM data just on the PTU (https://perspectivedestransports.files.wordpress.com/2013/08/carte-toulouse_rc3a9seau-toulouse.jpg) => 64Mo pbf file.

We have :

|  | lz4 size | RAM used by Kraken at startup |
| --- | --- | --- |
| Without geometries | 44Mo | ~190Mo |
| With geometries | 69Mo | ~240Mo |

And all of this to get : 
with_geom

Instead of :
without_geom

Let me know what you think about all of this.

### DIFF
--- a/source/ed/ed2nav.cpp
+++ b/source/ed/ed2nav.cpp
@@ -156,6 +156,7 @@ int main(int argc, char * argv[])
     auto logger = log4cplus::Logger::getInstance("log");
     std::string output, connection_string, region_name, cities_connection_string;
     double min_non_connected_graph_ratio;
+    bool export_georef_edges_geometries;
     po::options_description desc("Allowed options");
     desc.add_options()
         ("help,h", "Show this message")
@@ -168,6 +169,10 @@ int main(int argc, char * argv[])
         ("min_non_connected_ratio,m",
          po::value<double>(&min_non_connected_graph_ratio)->default_value(0.1),
          "min ratio for the size of non connected graph")
+        ("full_street_network_geometries", po::value<bool>(&export_georef_edges_geometries)->default_value(false),
+         "If true export street network geometries allowing kraken to return accurate geojson for street network sections. "
+         "Also improve projections accuracy. "
+         "WARNING : memory intensive. The lz4 can more than double in size and kraken will consume significantly more memory.")
         ("connection-string", po::value<std::string>(&connection_string)->required(),
          "database connection parameters: host=localhost user=navitia dbname=navitia password=navitia")
         ("cities-connection-string", po::value<std::string>(&cities_connection_string)->default_value(""),
@@ -216,7 +221,7 @@ int main(int argc, char * argv[])
     }
 
     try {
-        reader.fill(data, min_non_connected_graph_ratio);
+        reader.fill(data, min_non_connected_graph_ratio, export_georef_edges_geometries);
     }
     catch (const navitia::exception& e) {
         LOG4CPLUS_ERROR(logger, "error while reading the database "  << e.what());
--- a/source/ed/ed2nav.cpp
+++ b/source/ed/ed2nav.cpp
@@ -156,7 +156,6 @@ int main(int argc, char * argv[])
     auto logger = log4cplus::Logger::getInstance("log");
     std::string output, connection_string, region_name, cities_connection_string;
     double min_non_connected_graph_ratio;
-    bool export_georef_edges_geometries;
     po::options_description desc("Allowed options");
     desc.add_options()
         ("help,h", "Show this message")
@@ -169,9 +168,8 @@ int main(int argc, char * argv[])
         ("min_non_connected_ratio,m",
          po::value<double>(&min_non_connected_graph_ratio)->default_value(0.1),
          "min ratio for the size of non connected graph")
-        ("full_street_network_geometries", po::value<bool>(&export_georef_edges_geometries)->default_value(false),
-         "If true export street network geometries allowing kraken to return accurate geojson for street network sections. "
-         "Also improve projections accuracy. "
+        ("full_street_network_geometries", "If true export street network geometries allowing kraken to return accurate"
+         "geojson for street network sections. Also improve projections accuracy. "
          "WARNING : memory intensive. The lz4 can more than double in size and kraken will consume significantly more memory.")
         ("connection-string", po::value<std::string>(&connection_string)->required(),
          "database connection parameters: host=localhost user=navitia dbname=navitia password=navitia")
@@ -180,6 +178,7 @@ int main(int argc, char * argv[])
 
     po::variables_map vm;
     po::store(po::parse_command_line(argc, argv, desc), vm);
+    bool export_georef_edges_geometries(vm.count("full_street_network_geometries"));
 
     if(vm.count("version")){
         std::cout << argv[0] << " " << navitia::config::project_version << " "
--- a/source/ed/ed_reader.cpp
+++ b/source/ed/ed_reader.cpp
@@ -1428,7 +1428,8 @@ EdReader::get_duration (nt::Mode_e mode, float len, uint64_t source, uint64_t ta
 void EdReader::fill_graph(navitia::type::Data& data, pqxx::work& work) {
     std::string request = "select e.source_node_id, target_node_id, e.way_id, "
                           "ST_LENGTH(the_geog) AS leng, e.pedestrian_allowed as pede, "
-                          "e.cycles_allowed as bike,e.cars_allowed as car from georef.edge e;";
+                          "e.cycles_allowed as bike,e.cars_allowed as car, "
+                          "ST_ASTEXT(the_geog) AS geometry from georef.edge e;";
     pqxx::result result = work.exec(request);
     size_t nb_edges_no_way = 0;
     int nb_walking_edges(0), nb_biking_edges(0), nb_driving_edges(0);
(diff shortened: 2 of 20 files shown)


kinnou02:
Great job.
We have started to look for the replacement of this module, but it will take some time before it can be fully remove (if we can remove it)

There is almost nobody next week so we won't probably able to look at it before two weeks.

xlqian:
Jenkins, test this please!

source/ed/ed_reader.cpp
```diff
@@ -1425,10 +1425,11 @@ EdReader::get_duration (nt::Mode_e mode, float len, uint64_t source, uint64_t ta
     }
 }
 
-void EdReader::fill_graph(navitia::type::Data& data, pqxx::work& work) {
+void EdReader::fill_graph(navitia::type::Data& data, pqxx::work& work, bool export_georef_edges_geometries) {
     std::string request = "select e.source_node_id, target_node_id, e.way_id, "
                           "ST_LENGTH(the_geog) AS leng, e.pedestrian_allowed as pede, "
-                          "e.cycles_allowed as bike,e.cars_allowed as car from georef.edge e;";
+                          "e.cycles_allowed as bike,e.cars_allowed as car, "
+                          "ST_ASTEXT(the_geog) AS geometry from georef.edge e;";
```
kinnou02: It will be nice to not generate the wkt if we do not need it, on fr-nw we loose 40sec.
If your are on something else I can make a PR on your PR

source/ed/ed2nav.cpp
```diff
@@ -168,6 +169,10 @@ int main(int argc, char * argv[])
         ("min_non_connected_ratio,m",
          po::value<double>(&min_non_connected_graph_ratio)->default_value(0.1),
          "min ratio for the size of non connected graph")
+        ("full_street_network_geometries", po::value<bool>(&export_georef_edges_geometries)->default_value(false),
```
kinnou02: it's minor but you don't have to define a type, it's more a flag that we want to activate, without a type we can do this:

```
ed2nav --full_street_network_geometries <blabla>
```

then you use `vm.count("full_street_network_geometries")` for setting the value of `export_georef_edges_geometries`

source/georef/georef.h
```diff
@@ -82,10 +82,11 @@ struct Vertex {
 
 struct Edge {
     nt::idx_t way_idx = nt::invalid_idx; //< indexe vers le nom de rue
+    nt::idx_t geom_idx = nt::invalid_idx; // geometry index
     navitia::time_duration duration = {}; // duration of the edge
 
     template<class Archive> void serialize(Archive & ar, const unsigned int) {
-        ar & way_idx & duration;
+        ar & way_idx & geom_idx & duration;
```
kinnou02, line 89: You need to increment `Data::data_version` in type/data.cpp
Without this an old kraken can try to load a new dataset, in most case it will lead to a segmentation fault

kinnou02:
Great work! I will make this configurable in tyr.

source/ed/osm2ed.cpp
```diff
@@ -338,6 +338,7 @@ void OSMCache::insert_edges() {
             "cycles_allowed", "cars_allowed"});
     std::stringstream geog;
     geog << std::cout.precision(10);
+    std::vector<std::string> coords;
```
TeXitoi: you can use `navitia::type::LineString` (which is just a `typedef std::vector<GeographicalCoord`). Then, you can use boost to generate the wkt.

source/georef/georef.cpp
```diff
@@ -169,7 +169,16 @@ nt::GeographicalCoord Way::nearest_coord(const int number, const Graph& graph) c
 nt::MultiLineString Way::make_multiline(const Graph& graph) const {
     nt::MultiLineString multiline;
     for (auto edge: this->edges) {
-        multiline.push_back({graph[edge.first].coord, graph[edge.second].coord});
+        Edge e = graph[boost::edge(edge.first, edge.second, graph).first];
+        /*
+            Check if the edge way is the same that the current way.
+            Since boost::edge return only one edge between two vertices, if there is parallel edges, we aren't
+            sure it's the right one, and the geom_idx could be out of bounds.
+        */
```
TeXitoi: If you prefer, to handle parallel edges, you can loop on `boost::out_egdes` as in nearest_edge to find the shape.

source/georef/georef.cpp
```diff
@@ -324,10 +333,18 @@ void ProjectionData::init(const type::GeographicalCoord & coord, const GeoRef &
...
-    distances[Direction::Target] = projected.distance_to(vertex2_coord);
+    Edge e = sn.graph[nearest_edge];
+    if(e.geom_idx != nt::invalid_idx) {
+        auto& geom = sn.ways[e.way_idx]->geoms[e.geom_idx];
+        this->projected = type::project(geom, coord);
+        distances[Direction::Source] = type::real_distance_from_extremity(geom, projected, false);
+        distances[Direction::Target] = type::real_distance_from_extremity(geom, projected, true);
+    } else {
+        this->projected = coord.project(vertex1_coord, vertex2_coord).first;
+        // On calcule la distance « initiale » déjà parcourue avant d'atteindre ces extrémité d'où on effectue le calcul d'itinéraire
```
TeXitoi: Please translate this comment in english (I know, that's not your code).

eturck:
Sorry I was away for the last two weeks. I will be back to work on this PR as soon as possible. 
Thank you for the feedback!

pbougue:
You have to rebase (but maybe in the end)

eturck:
Hello again :).

I've made a few changes from the past feedback : 
- full_street_network_geometries is configurable in tyr, thanks to @kinnou02 ;),
- full_street_network_geometries is now a flag,
- use nt::LineString and boost in osm2ed to generate the wkt,
- don't call ST_ASTEXT if we don't want to export geometries,
- use out_edges in make_multiline to handle parallel edges,
- bump data_version to 62,

I also made some new changes. Those changes are made to handle parallel edges or edges with source and target on the same vertex : 
- In ProjectionData I've added the Edge property, this way when there is a parallel edge we know what was the exact Edge we projected on. I haven't just added the edge descriptor because it's not serializable by itself, and I didn't manage to make it serializable,
- In is_projected_on_same_edge we don't just check source and target we also check the duration and the way idx are the same,
- I made a few changes here and there to handle edges with the same vertex for source and target (dead ends ending with a roundabout with only one entrance/exit).

The last change is the biggest one. Currently, if two points are projected on the same edge we consider that the path using only this edge is the fastest way to reach those 2 points. But it's not always the case, to get from 'o' to 'd' it's faster to use v1-v2 : 

```
.------------------o----|v1
|                       |
'------------------d----|v2
```

I've changed the current way get_path works. Currently it does : 
- If we are on the same edge take a direct path,
- Otherwise use the path returned by build_path and add projection

And now I do : 
- Get the path returned by build_path and store its duration as base_duration,
- Add the projections to this path to get the total duration,
- Check if it's better to use a direct path. It is the case if : 
  - we are on the same edge
  - AND
    - the path returned by build path have a base_duration of 0 second AND it's not an edge with the same source and target (make sure we have at least an edge in the path returned by Dijkstra except if source and target are the same)
    - OR 
    - the duration on the same edge is smaller than the total duration of the path with projection

It works great when using the API but I encoutered a problem when adding a unit test. I used before "compute_path" defined in georef_test which call update_path in street_network.cpp. But when using it, I never have the path v1-v2 returned by Dijkstra.
I switched to using a StreetNetwork worker like in find_nearest_on_same_edge or the API and it's working, but only when setting a max_duration on the entry point, if I keep the default 1 second value it doesn't work.

My questions are :
- Are you ok with all those changes :)?,
- Is this a problem to have an edge property instead of an edge_t in ProjectionData? I would love to have an edge_t but I didn't find a way to make it serializable, and Google wasn't helping much...
- Is it normal to have to set a max_duration bigger than one second to have the correct path? Or is there a bug on my end?
- Is it normal that it's not working with "compute_path"? I didn't see where it was used except in the unit tests so I don't really know if that is impacting other things. What I saw is that it is not calling Djikstra the same way.

A few screenshots of what it did before : 
parallel_before
And after : 
parallel

And an illustration of the use case above : 
parralel_routing
parallel_routing_two

I guess we'll talk about it in a few minutes ;)

source/tyr/tyr/binarisation.py
```diff
@@ -457,6 +457,8 @@ def ed2nav(self, instance_config, job_id, custom_output_dir):
         argv = ["-o", output_file, "--connection-string", connection_string]
         if 'CITIES_DATABASE_URI' in current_app.config and current_app.config['CITIES_DATABASE_URI']:
             argv.extend(["--cities-connection-string", current_app.config['CITIES_DATABASE_URI']])
+        if instance.full_sn_geometries:
+            argv.extend(['--full_street_network_geometries', 'true'])
```
kinnou02: we doen't need the `true` anymore since `--full_street_network_geometries` is now a flag

kinnou02:
Seems good to me.
I think we should wait after the release to merge this PR since Artemis will probably register some changes.

source/georef/georef.h
pbougue: Done

pbougue (approved):

PR #1719, opened 2016-08-05
All: Add direct path api for experimental

TeXitoi:


source/jormungandr/jormungandr/georef.py
```diff
@@ -62,6 +62,28 @@ def place(self, place):
...
+        req.direct_path.origin.place = origin
+        req.direct_path.origin.access_duration = 0
+        req.direct_path.destination.place = destination
+        req.direct_path.destination.access_duration = 0
+        req.direct_path.datetime = datetime
+        req.direct_path.clockwise = clockwise
+        req.direct_path.streetnetwork_params.origin_mode = mode
+        req.direct_path.streetnetwork_params.destination_mode = mode
+        req.direct_path.streetnetwork_params.walking_speed = self.instance.walking_speed
+        req.direct_path.streetnetwork_params.max_walking_duration_to_pt = self.instance.max_walking_duration_to_pt
```
kinnou02, line 78: The "x2" coefficient should be here

TeXitoi, line 78: In the code we don't do x2, we do origin_max_duration + destination_max_duration. But I can do that.

TeXitoi, line 78: Looking at the code I don't think it's a good idea to x2 here, lots of code will need special casing.

xlqian:
Jenkins, test this please!

source/jormungandr/jormungandr/georef.py
kinnou02, line 78: ok

kinnou02:
ok for me

PR #1720, opened 2016-08-05
Jormungandr: change for python 3

kinnou02:
This is work in progress, it should not break anything for python 2 but more tests are required.
Some libs have been updated.

suds must be upgraded to suds-jurko for python 3, but the interface seems to have change.

I will continue this after my holidays.

### DIFF
--- a/source/jormungandr/jormungandr/instance.py
+++ b/source/jormungandr/jormungandr/instance.py
@@ -399,7 +399,7 @@ class Instance(object):
         """
         update the property of an instance from a response if the metadatas field if present
         """
-        if response.HasField(b"metadatas") and response.publication_date != self.publication_date:
+        if response.HasField(str("metadatas")) and response.publication_date != self.publication_date:
             with self.lock as lock:
                 if response.metadatas.shape and response.metadatas.shape != "":
                     try:
@@ -423,7 +423,7 @@ class Instance(object):
             resp = self.send_and_receive(req, timeout=1000, quiet=True)
             self.update_property(resp)
             #the instance is automatically updated on a call
-            if resp.HasField(b'publication_date') and self.publication_date != resp.publication_date:
+            if resp.HasField(str('publication_date')) and self.publication_date != resp.publication_date:
                 self.publication_date = resp.publication_date
                 return True
         except DeadSocketException:
--- a/source/jormungandr/jormungandr/realtime_schedule/timeo.py
+++ b/source/jormungandr/jormungandr/realtime_schedule/timeo.py
@@ -201,7 +201,7 @@ class Timeo(RealtimeProxy):
             return None
 
         # timeo can only handle items_per_schedule if it's < 5
-        count = min(count, 5) or 5  # if no value defined we ask for 5 passages
+        count = min(count or 5, 5) or 5  # if no value defined we ask for 5 passages
 
         # if a custom datetime is provided we give it to timeo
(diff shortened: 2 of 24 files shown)


xlqian:
Jenkins, test this please!

kinnou02:
Jenkins, test this please.

kinnou02:
Jenkins, test this please.

source/jormungandr/jormungandr/realtime_schedule/timeo.py
```diff
@@ -200,7 +200,7 @@ def _make_url(self, route_point, count=None, from_dt=None):
             return None
 
         # timeo can only handle items_per_schedule if it's < 5
-        count = min(count, 5) or 5  # if no value defined we ask for 5 passages
+        count = min(count or 5, 5) or 5  # if no value defined we ask for 5 passages
```
TeXitoi: The final `or 5` is useless.

source/jormungandr/jormungandr/travelers_profile.py
```diff
@@ -79,7 +79,8 @@ def override_params(self, args):
                               ('destination_mode',           self.last_section_mode),
                               ('wheelchair',                 self.wheelchair))
 
-        def override((arg, profile_attr)):
+        def override(tupe):
```
pbougue: not sure: `tuple` ?

kinnou02: `tuple` is a "reserved" word, I can override it but I don't really like it, but I can rename it `pair` it will be better

source/jormungandr/jormungandr/instance.py
```diff
@@ -393,7 +393,7 @@ def update_property(self, response):
         """
         update the property of an instance from a response if the metadatas field if present
         """
-        if response.HasField(b"metadatas") and response.publication_date != self.publication_date:
+        if response.HasField(str("metadatas")) and response.publication_date != self.publication_date:
```
xlqian, line 402: hmm, `str` may be useless, no?

source/navitiacommon/navitiacommon/ratelimit.py
```diff
@@ -33,7 +33,7 @@
 import math
 import redis
 import time
-from itertools import izip
+from builtins import zip
```
xlqian, line 36: not necessary i think

xlqian (approved):

source/jormungandr/jormungandr/instance.py
kinnou02: `str()` ensure that we have the desired string type no matter what version of python we are using.
`unicode_literals` is enable so every quoted string is a `unicode` object, Protobuf doesn't always like them, that's kinda strange. With `str()` we have a ascii string aka `bytes` in python2 and a unicode string in python 3.

source/navitiacommon/navitiacommon/ratelimit.py
kinnou02: In python 3 `zip()` is the equivalent of `izip()`, but in python 2 they doesn't work the same way, `builtins` form the `futures` package ensure that we always have the desired function accross python version.

pbougue (approved):

kinnou02:
Jenkins, test this please.

TeXitoi (approved):

PR #1723, opened 2016-08-08
Atos connector updated. New connectors added.

Netmisa:
# Introduction

New operators added:
- Keolis
- Effia
- Effia transport
- Kéolis
# How to test
## 1. Run unit tests

_Execute command:_

``` shell
JORMUNGANDR_CONFIG_FILE="<navitia_path>/run/jormungandr_settings.py" PYTHONPATH="<navitia_path>/source/navitiacommon:<navitia_path>/source/jormungandr" KRAKEN_BUILD_DIR=<navitia_path>/bin/release  py.test parking_space_availability/bss/tests/atos_test.py 
```

PR #1724, opened 2016-08-11
Atos connector updated. New connectors added.

ooga:
# Introduction

New operators added:
- Keolis
- Effia
- Effia transport
- Kéolis
# How to test
## 1. Run unit tests

_Execute command:_

``` shell
JORMUNGANDR_CONFIG_FILE="<navitia_path>/run/jormungandr_settings.py" PYTHONPATH="<navitia_path>/source/navitiacommon:<navitia_path>/source/jormungandr" KRAKEN_BUILD_DIR=<navitia_path>/bin/release  py.test parking_space_availability/bss/tests/atos_test.py 
```

### DIFF
--- a/source/jormungandr/jormungandr/parking_space_availability/bss/atos.py
+++ b/source/jormungandr/jormungandr/parking_space_availability/bss/atos.py
@@ -31,18 +31,17 @@ from __future__ import absolute_import, print_function, unicode_literals, divisi
 from jormungandr.parking_space_availability.bss.bss_provider import BssProvider
 from jormungandr.parking_space_availability.bss.stands import Stands
 from jormungandr import cache, app
-from urllib2 import URLError
 import suds
 import logging
 
 
 class AtosProvider(BssProvider):
 
-    def __init__(self, id_ao, network, url, operators, timeout=5):
+    def __init__(self, id_ao, network, url, operators={'keolis'}, timeout=5):
         self.id_ao = id_ao
-        self.network = network
+        self.network = network.lower()
         self.WS_URL = url
-        self.operators = operators
+        self.operators = [o.lower() for o in operators]
         self.timeout = timeout
         self._client = None
 
--- a/source/jormungandr/jormungandr/parking_space_availability/bss/atos.py
+++ b/source/jormungandr/jormungandr/parking_space_availability/bss/atos.py
@@ -38,12 +38,11 @@ import logging
 
 class AtosProvider(BssProvider):
 
-    OPERATORS = ('keolis', 'effia', 'effia transport', u'kéolis')
-
-    def __init__(self, id_ao, network, url, timeout=5):
+    def __init__(self, id_ao, network, url, operators, timeout=5):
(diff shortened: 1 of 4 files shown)


ooga:
Update of #1723

source/jormungandr/jormungandr/parking_space_availability/bss/atos.py
```diff
@@ -37,12 +38,11 @@
...
 class AtosProvider(BssProvider):
 
-    OPERATOR = 'Keolis'
-
-    def __init__(self, id_ao, network, url, timeout=5):
+    def __init__(self, id_ao, network, url, operators, timeout=5):
         self.id_ao = id_ao
         self.network = network
         self.WS_URL = url
+        self.operators = map(unicode.lower, operators)
```
antoine-de: isn't it better to store them as lower ?
and in this case you can store the network as lower too

source/jormungandr/jormungandr/parking_space_availability/bss/atos.py
```diff
@@ -37,12 +38,11 @@
 
 class AtosProvider(BssProvider):
 
-    OPERATOR = 'Keolis'
-
-    def __init__(self, id_ao, network, url, timeout=5):
+    def __init__(self, id_ao, network, url, operators, timeout=5):
```
antoine-de: :+1:

source/jormungandr/jormungandr/parking_space_availability/bss/atos.py
```diff
@@ -42,7 +42,7 @@ def __init__(self, id_ao, network, url, operators, timeout=5):
         self.id_ao = id_ao
         self.network = network
         self.WS_URL = url
-        self.operators = map(unicode.lower, operators)
+        self.operators = operators
```
antoine-de: I'm really sorry, it's only in python 3 that this is lazy computed so your previous version was fine :wink: (but since we plan to move to python 3 soon we couln't have kept the map).

I'm fine with this version.
If you wanted to ensure the lowerness you could have also done:
`self.operators = [o.lower() for o in operators]`

source/jormungandr/jormungandr/parking_space_availability/bss/atos.py
```diff
@@ -51,8 +51,8 @@ def __repr__(self):
 
     def support_poi(self, poi):
         properties = poi.get('properties', {})
-        return properties.get('operator') == self.OPERATOR and \
-               properties.get('network') == self.network
+        return properties.get('operator').lower() in self.operators and \
+               properties.get('network').lower() == self.network
```
kinnou02, line 54: You need to apply a lower on `self.network` too, ideally in the `__init__`

source/jormungandr/jormungandr/parking_space_availability/bss/atos.py
kinnou02: It would help to set a default value to``operators` this way we won't break anything

PR #1725, opened 2016-08-17
RT proxy: custom Synthese route lookup

antoine-de:
linked to http://jira.canaltp.fr/browse/HOR-10

since we have some bad synchronization between synthese routes code and our route code, implementation of a custom lookup
- we first look if we can find a route with the synthese code of our route in it's external codes (it can have several if the route is a fusion of many routes)
- else, we query navitia to get all routes that pass by the stoppoint for the line of the route point
  - if we get only one route, we search for this route's line in the synthese response (because lines synthese code are move coherent)
    - we concatenate all synthese passage of this line (since we have have fusioned some routes)
- else we failed, we return the base schedule

On T2C data with this we now have 68% of realtime data

### DIFF
--- a/source/jormungandr/jormungandr/ptref.py
+++ b/source/jormungandr/jormungandr/ptref.py
@@ -31,6 +31,19 @@ from __future__ import absolute_import, print_function, unicode_literals, divisi
 from navitiacommon import request_pb2, type_pb2
 from flask import logging
 
+# Mapping to know with response field to consider for a given pb_type
+PT_TYPE_RESPONSE_MAPPING = {
+    type_pb2.STOP_AREA: 'stop_areas',
+    type_pb2.STOP_POINT: 'stop_points',
+    type_pb2.POI: 'pois',
+    type_pb2.ADMINISTRATIVE_REGION: 'administrative_regions',
+    type_pb2.LINE: 'lines',
+    type_pb2.NETWORK: 'networks',
+    type_pb2.COMMERCIAL_MODE: 'commercial_modes',
+    type_pb2.ROUTE: 'routes',
+    type_pb2.LINE_GROUP: 'line_groups',
+}
+
 
 class PtRef(object):
     def __init__(self, instance):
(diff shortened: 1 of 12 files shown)


source/jormungandr/jormungandr/ptref.py
```diff
@@ -58,3 +71,28 @@ def get_stop_point(self, line_uri, code_key, code_value):
...
+        req.ptref.count = 20
+        req.ptref.start_page = 0
+        req.ptref.depth = 1
+        req.ptref.filter = filter
+
+        while True:
+            result = self.instance.send_and_receive(req)
+
+            objects = getattr(result, PT_TYPE_RESPONSE_MAPPING[pb_type])
+            if not objects:
```
kinnou02, line 91: You could also check if your aren't on the last page, it will prevent a useless call

source/jormungandr/jormungandr/realtime_schedule/synthese.py
```diff
@@ -138,11 +141,9 @@ def _get_next_passage_for_route_point(self, route_point, count=None, from_dt=Non
...
 
         logging.getLogger(__name__).debug("synthese response: {}".format(r.text))
-        stop_point_id = str(route_point.fetch_stop_id(self.object_id_tag))
-        route_id = str(route_point.fetch_route_id(self.object_id_tag))
-        route_point = SyntheseRoutePoint(route_id, stop_point_id)
-        m = self._get_synthese_passages(r.content)
-        return m.get(route_point)# if there is nothing from synthese, we keep the base
+        passages = self._get_synthese_passages(r.content)
+
+        return self._find_route_point(route_point, passages)
```
kinnou02: don't really like the name of this function, I would expect a route point as result while it will return passages

source/jormungandr/jormungandr/realtime_schedule/synthese.py
```diff
@@ -222,3 +225,62 @@ def _timestamp_to_date(self, timestamp):
...
+         * else we query navitia to get all routes that pass by the stoppoint for the line of the route point
+            * if we get only one route, we search for this route's line in the synthese response
+                (because lines synthese code are move coherent)
+                -> we concatenate all synthese passages on this line
+         -> else we return the base schedule
+        """
+        log = logging.getLogger(__name__)
+        stop_point_id = str(route_point.fetch_stop_id(self.object_id_tag))
+        is_same_route = lambda syn_rp: syn_rp.syn_route_id in route_point.fetch_all_route_id(self.object_id_tag)
+        p = next((p for syn_rp, p in passages.items()
```
kinnou02: We only return the passages of one route point here no?
Since route have been merge we need to return the passages of all the synthése's routes of the navitia route

source/jormungandr/jormungandr/realtime_schedule/synthese.py
```diff
@@ -222,3 +225,62 @@ def _timestamp_to_date(self, timestamp):
...
+                                                      line=route_point.fetch_line_uri()))
+
+        first_routes = list(itertools.islice(routes_gen, 2))
+
+        if len(first_routes) == 1:
+            # there is only one route that pass through our stoppoint for the line of the routepoint
+            # we can concatenate all synthese's route of this line
+            line_passages = [p for syn_rp, p in passages.items() if syn_rp.syn_line_id ==
+                             route_point.fetch_line_id(self.object_id_tag)]
+
```
kinnou02, line 270: A log here would probably be useful

source/jormungandr/jormungandr/ptref.py
antoine-de, line 91: almost done below

source/jormungandr/jormungandr/realtime_schedule/synthese.py
antoine-de, line 270: done above :wink:

PR #1728, opened 2016-08-19
jormungandr: add a ratelimiter to timeo

kinnou02:
it's for preventing major outage when we ask us to set timeout value so
big that setting no timeout will have the same effect.
This PR is more of a quick bypass, a better will be needed, probably
something like a distributed semaphore, than complicate for almost
nothing, if you have better idea, I am listening


source/jormungandr/jormungandr/realtime_schedule/timeo.py
```diff
@@ -73,6 +74,17 @@ def __init__(self, id, service_url, service_args, timezone,
         # Note: if the timezone is not know, pytz raise an error
         self.timezone = pytz.timezone(timezone)
 
+        if kwargs.get('redis_host', None) and kwargs.get('rate_limit_count', None):
```
antoine-de: `, None)` is the default

antoine-de:
apart from the default values, ok for me

PR #1730, opened 2016-08-22
osm2ed: fix a bug with private way part of an administrative boundary

kinnou02:
We didn't import private way even when they were part of a boundary, we
now keep them.

There is room for improvement, a lot of useless way seems to be
imported, all ways part of a boundary are imported, they aren't use in
the streetnetwork but we insert them.
Since this component should progressively disappear and is not covered
by any test I think the best call is to let it this way.

### DIFF
--- a/source/ed/osm2ed.cpp
+++ b/source/ed/osm2ed.cpp
@@ -306,7 +306,7 @@ void OSMCache::insert_ways(){
     size_t n_inserted = 0;
     const size_t max_n_inserted = 50000;
     for (const auto& way : ways) {
-        if (!way.is_used() || way.properties.none()) {
+        if (!way.is_used() || !way.is_street()) {
             continue;
         }
         std::vector<std::string> values;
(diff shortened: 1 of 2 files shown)


TeXitoi:
OK for me.

kinnou02:
We had to remove some useless way

source/ed/osm2ed.cpp
```diff
@@ -306,7 +306,7 @@ void OSMCache::insert_ways(){
     size_t n_inserted = 0;
     const size_t max_n_inserted = 50000;
     for (const auto& way : ways) {
-        if (!way.is_used()) {
+        if (!way.is_used() || way.properties.none()) {
```
TeXitoi: Can't we have a `way.is_street()` for better readability?

kinnou02: Oh, I didn't see this sorry, I think it's possible ;)

TeXitoi:
OK for me.

kinnou02:
Jenkins, test this please.

PR #1737, opened 2016-08-26
Isochrone: Add min_date_time and max_date_time in json

ZoeBrunet:


source/navitia-proto
```diff
@@ -1 +1 @@
-Subproject commit bcd5b9f1000397aea94f1c8f499e7ce424b03667
+Subproject commit 67cc20e526204facbc8bf10f3a51151ee404da8c
```
TeXitoi: to be updated

source/routing/raptor_api.cpp
```diff
@@ -1286,6 +1291,12 @@ static boost::optional<pbnavitia::Response> fill_isochrone_common(IsochroneCommo
     return boost::none;
 }
 
+DateTime set_isochrone_date (const DateTime& init_dt,
```
TeXitoi: `make_` or `get_` is better here as you don't set anything

TeXitoi:
OK for me.

PR #1739, opened 2016-08-29
tyr: execute purge of instances automatically every night

kinnou02:


TeXitoi:
How it was done before ?

kinnou02:
It was a cron, the main drawback is that we can't modify it easily since it's managed by puppet.

TeXitoi:
OK. Thanks.

PR #1743, opened 2016-08-30
Jormun: add support for async call to kraken

kinnou02:
There is one major warning: each green thread has it own thread local
storage so stuff like flask.request or flask.g won't work as expected.

### DIFF
--- a/source/jormungandr/jormungandr/scenarios/new_default.py
+++ b/source/jormungandr/jormungandr/scenarios/new_default.py
@@ -715,28 +715,27 @@ class Scenario(simple.Scenario):
         # TODO: call first bss|bss and do not call walking|walking if no bss in first results
         resp = []
         logger = logging.getLogger(__name__)
-        futures = {}
+        futures = []
+        def worker(dep_mode, arr_mode, instance, request, request_id):
+            return (dep_mode, arr_mode, instance.send_and_receive(request, request_id=request_id))
+
         for dep_mode, arr_mode in krakens_call:
             pb_request = create_pb_request(request_type, request, dep_mode, arr_mode)
 
             #we spawn a new green thread, it won't have access to our thread local request object so we set request_id
-            futures[(dep_mode, arr_mode)] = gevent.spawn(instance.send_and_receive,
-                                                         pb_request,
-                                                         request_id=flask.request.id)
+            futures.append(gevent.spawn(worker, dep_mode, arr_mode, instance, pb_request, request_id=flask.request.id))
 
-        gevent.joinall(futures.values())
-        for (dep_mode, arr_mode), future in futures.items():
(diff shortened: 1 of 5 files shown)


source/jormungandr/jormungandr/scenarios/new_default.py
```diff
@@ -713,13 +715,20 @@ def call_kraken(self, request_type, request, instance, krakens_call):
...
-            self.nb_kraken_calls += 1
 
-            local_resp = instance.send_and_receive(pb_request)
+            #we spawn a new green thread, it won't have access to our thread local request object so we set request_id
+            futures[(dep_mode, arr_mode)] = gevent.spawn(instance.send_and_receive,
+                                                         pb_request,
+                                                         request_id=flask.request.id)
 
+        gevent.joinall(futures.values())
+        for (dep_mode, arr_mode), future in futures.items():
```
kinnou02: Without the log line 727|736 I can improve thing a little bit by not waiting that all threads had finished before starting to handle the results.
But i like this log :(

TeXitoi: I don't understand why you can't log if you don't joinall.

kinnou02: I can use `iwait()` for working on them as they finish, this function take a list of future/greenlet.
If I use it I won't be able to know what are the mode used as fallback for this call

TeXitoi: You can just remove the joinall. `future.get()` will wait. That's not as optimal as `iwait()`, but better than `joinall()`.

Or you can spawn a closure returning `(dep_mode, arr_mode, instance.send_and_receive(...))`

TeXitoi:
OK for me.

source/jormungandr/jormungandr/scenarios/new_default.py
```diff
@@ -713,21 +715,27 @@ def call_kraken(self, request_type, request, instance, krakens_call):
...
+        def worker(dep_mode, arr_mode, instance, request, request_id):
+            return (dep_mode, arr_mode, instance.send_and_receive(request, request_id=request_id))
+
         for dep_mode, arr_mode in krakens_call:
             pb_request = create_pb_request(request_type, request, dep_mode, arr_mode)
-            self.nb_kraken_calls += 1
 
-            local_resp = instance.send_and_receive(pb_request)
+            #we spawn a new green thread, it won't have access to our thread local request object so we set request_id
+            futures.append(gevent.spawn(worker, dep_mode, arr_mode, instance, pb_request, request_id=flask.request.id))
```
xlqian: why don't you do a deepcopy in this loop?

xlqian:
:+1:

PR #1748, opened 2016-09-01
jormungandr: fix a few error on bss connectors

kinnou02:
If a POI didn't have all the required fields (operator, network and ref) it could cause some errors leading to incomplete responses.

### DIFF
--- a/source/jormungandr/jormungandr/parking_space_availability/bss/tests/atos_test.py
+++ b/source/jormungandr/jormungandr/parking_space_availability/bss/tests/atos_test.py
@@ -69,6 +69,11 @@ def parking_space_availability_atos_support_poi_test():
     poi['properties']['operator'] = 'Bad_operator'
     assert not provider.support_poi(poi)
 
+    invalid_poi = {}
+    assert not provider.support_poi(invalid_poi)
+    invalid_poi = {'properties': {}}
+    assert not provider.support_poi(invalid_poi)
+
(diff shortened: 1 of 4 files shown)


source/jormungandr/jormungandr/parking_space_availability/bss/tests/atos_test.py
```diff
@@ -81,6 +86,12 @@ def parking_space_availability_atos_get_informations_test():
     provider = AtosProvider(u'10', u'vélitul', u'https://webservice.atos.com?wsdl', {'keolis'})
     provider.get_all_stands = MagicMock(return_value=all_stands)
     assert provider.get_informations(poi) == stands
+    invalid_poi = {}
+    assert provider.get_informations(invalid_poi) is None
+
+    poi_blur_ref = {'properties': {'ref': '02'}}
+    assert provider.get_informations(poi_blur_ref) == stands
```
pbougue, line 93: I don't get this test

kinnou02: The ref of the poi returned by the connector is `2`, this test check that we ignore leading 0

pbougue: Ah ok. The stands/all_stands declaration is a little confusing.

PR #1749, opened 2016-09-01
Stat: fix TZ bug when launching tests on a London platform

pbougue:
we now consider kraken TZ to build UTC datetimes for all stats
 (used to do it considering jormun TZ for journey_request, and sending local kraken datetimes in journey)
also removed some clang warnings

### DIFF
--- a/source/jormungandr/jormungandr/stat_manager.py
+++ b/source/jormungandr/jormungandr/stat_manager.py
@@ -41,6 +41,7 @@ from jormungandr import utils
 import re
 from threading import Lock
 
+import pytz
 import time
 import sys
 import kombu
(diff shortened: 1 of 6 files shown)


source/jormungandr/jormungandr/stat_manager.py
```diff
@@ -296,14 +297,19 @@ def fill_journey(self, stat_journey, resp_journey):
         of the node journeys in the result.
         """
         init_journey(stat_journey)
+
+        tz = utils.get_timezone()
         if 'requested_date_time' in resp_journey:
-            stat_journey.requested_date_time = utils.str_to_time_stamp(resp_journey['requested_date_time'])
+            dt = tz.normalize(tz.localize(utils.str_to_dt(resp_journey['requested_date_time'])))
+            stat_journey.requested_date_time = utils.date_to_timestamp(dt.astimezone(pytz.utc))
```
TeXitoi: Can't this be abstracted in a function?

PR #1750, opened 2016-09-02
kraken and jormungandr: Add direct path duration to raptor call for Scenario Experimental

kadhikari:


source/jormungandr/jormungandr/planner.py
```diff
@@ -68,4 +70,7 @@ def journeys(self, origins, destinations, datetime, clockwise, journey_parameter
         for uri in journey_parameters.forbidden_uris:
             req.journeys.forbidden_uris.append(uri)
 
+        if journey_parameters.direct_path_duration:
```
TeXitoi: `is not None`, else if the duration is 0, it will not be given.

source/jormungandr/jormungandr/scenarios/experimental.py
```diff
@@ -146,13 +146,26 @@ def call_kraken(self, request_type, request, instance, krakens_call):
...
+                journey_parameters.direct_path_duration = direct_path.journeys[0].durations.total
+                resp.append(direct_path)
 
             local_resp = instance.planner.journeys(g.origins_fallback[dep_mode],
                                                    g.destinations_fallback[arr_mode],
                                                    request['datetime'],
                                                    request['clockwise'],
                                                    journey_parameters)
 
+            if local_resp.HasField(b"error") and local_resp.error.id == response_pb2.Error.error_id.\
```
TeXitoi: strange line break

kadhikari:
@TeXitoi 
I pushed modifications on your comments

source/kraken/worker.cpp
```diff
@@ -872,6 +873,7 @@ static type::EntryPoint make_sn_entry_point(const std::string& place,
     case type::Type_e::StopPoint:
     case type::Type_e::POI:
         entry_point.coordinates = coord_of_entry_point(entry_point, data); // StopPoint doesn't use street network
+    default:
```
TeXitoi, line 877: strange added line

source/kraken/worker.cpp
```diff
@@ -881,6 +883,7 @@ static type::EntryPoint make_sn_entry_point(const std::string& place,
     case type::Type_e::POI:
     case type::Type_e::StopArea:
         entry_point.streetnetwork_params.mode = type::static_data::get()->modeByCaption(mode);
+    default:
```
TeXitoi, line 888: strange added line

source/kraken/worker.cpp
kadhikari: I added this line to delete warning while compiling kraken

source/kraken/worker.cpp
kadhikari: I added this line to delete warning while compiling kraken

source/kraken/worker.cpp
TeXitoi: Then, prefer add a break before:

``` c++
switch (toto) {
case toto:
    a = b;
    break;
default:
    break;
}
```

source/kraken/worker.cpp
TeXitoi: same

kadhikari:
@TeXitoi i push one more modification

source/jormungandr/jormungandr/planner.py
```diff
@@ -29,21 +29,23 @@ def __init__(self,
...
+        self.direct_path_duration = direct_path_duration
 
 
 class Kraken(object):
 
     def __init__(self, instance):
         self.instance = instance
 
-    def journeys(self, origins, destinations, datetime, clockwise, journey_parameters):
+    def journeys(self, origins, destinations, datetime, clockwise, journey_parameters, direct_path_duration=None):
```
kinnou02: this parameter is not used

source/jormungandr/jormungandr/scenarios/experimental.py
```diff
@@ -146,13 +146,26 @@ def call_kraken(self, request_type, request, instance, krakens_call):
...
             #todo: this is probably shared between multiple thread
             self.nb_kraken_calls += 1
+            direct_path = self._get_direct_path(instance,
+                                                dep_mode,
+                                                request['origin'],
+                                                request['destination'],
+                                                request['datetime'],
+                                                request['clockwise'])
+            if direct_path.journeys:
+                journey_parameters.direct_path_duration = direct_path.journeys[0].durations.total
```
kinnou02, line 156: Since the journey_parameters is keep between call to kraken you need to reset the value of `direct_path_duration` if no direct path has been found

kadhikari: ok

PR #1752, opened 2016-09-06
Doc: Add contributors and datasets endpoints

pbougue:
As a lot of people ask for already used datas, let's explain them how to search if we already have them.

### DIFF
--- a/documentation/slate/source/includes/apis.md
+++ b/documentation/slate/source/includes/apis.md
@@ -20,6 +20,39 @@ The only arguments are the ones of [paging](#paging).
 | `coverage/{lon;lat}`                    | Information about a specific region |
 
 
+<a name="datasets"></a>Datasets
+-------------------------------
+
+Very simple endpoint providing the sets of data that are used in the given coverage.
+Contributor associated to every dataset is also provided.
+Very usefull to know all the datas that form a coverage.
+
+The only arguments are the ones of [paging](#paging).
+
+### Accesses
+
+| url | Result |
+|----------------------------------------------|-------------------------------------------|
+| `coverage/{region_id}/datasets               | List of the datasets of a specific region |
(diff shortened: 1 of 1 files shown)


documentation/slate/source/includes/apis.md
```diff
@@ -20,6 +20,39 @@ The only arguments are the ones of [paging](#paging).
 | `coverage/{lon;lat}`                    | Information about a specific region |
 
 
+<a name="datasets"></a>Datasets
+-------------------------------
+
+Very simple endpoint providing the sets of data that are used in the given coverage.
```
antoine-de, line 32: I think it is important to add some explanation on what is exactly a contributor and a dataset

pbougue: Done

documentation/slate/source/includes/apis.md
```diff
@@ -20,6 +26,48 @@ The only arguments are the ones of [paging](#paging).
...
+-------------------------------
+
+Very simple endpoint providing the contributors of data for the given coverage.
+
+A contributor is a data provider (typically a transport authority), and can provide multiple [datasets](#datasets).
+For example, the contributor Italian Railways will provide a dataset for the national train and some others for the regional trains.
+We will try to put them in the same [coverage](#coverage) so that we assemble them in the same journey search, using both.
+
+Very usefull to know which contributors are used in the datasets forming a coverage.
+
```
stifoon, line 60: Could be nice to add a sentence about contributors feeding feedpublisher property on every response. With licence and etc.

TeXitoi: First, it should be activated on prod...

pbougue: I agree with both... Maybe it should be part of a new PR (or part of #1754 as it should allow feed_publisher in prod).
We merge this and make a new PR? @stifoon if you don't have the time to I can prepare it.

PR #1753, opened 2016-09-07
Experimental: Use direct path in building journeys

xlqian:
use kraken's direct_path to compute fallback direct path in experimental


source/georef/street_network.cpp
```diff
@@ -364,7 +365,13 @@ PathFinder::get_duration_with_dijkstra(const navitia::time_duration& radius,
     if (dest_coords.empty()) {
         return {};
     }
-    auto offset = geo_ref.offsets[mode];
+    nt::idx_t offset;
+    if(mode == type::Mode_e::Car){
+        //on direct path with car we want to arrive on the walking graph
+        offset = geo_ref.offsets[nt::Mode_e::Walking];
+    } else {
+        offset = geo_ref.offsets[mode];
```
TeXitoi, line 375: Test with BSS

source/georef/street_network.cpp
```diff
@@ -510,6 +511,7 @@ PathFinder::get_path(const ProjectionData& target,
                      const std::pair<navitia::time_duration, ProjectionData::Direction>& nearest_edge)
 {
     if (! computation_launch || ! target.found || nearest_edge.first == bt::pos_infin)
+
```
TeXitoi: strange empty line

source/jormungandr/jormungandr/scenarios/experimental.py
```diff
@@ -106,10 +105,10 @@ class Scenario(new_default.Scenario):
     def __init__(self):
         super(Scenario, self).__init__()
 
-    def _get_direct_path(self, instance, mode, pt_object_origin, pt_object_destination, datetime, clockwise):
+    def _get_direct_path(self, instance, mode, pt_object_origin, pt_object_destination, datetime, clockwise, reverse_sections = False):
```
TeXitoi: long line

source/georef/street_network.cpp
xlqian: tested

source/jormungandr/jormungandr/scenarios/experimental.py
```diff
@@ -118,18 +101,103 @@ def update_crowfly_duration(instance, fallback_list, mode, stop_area_uri):
...
+            elif not g.fallback_direct_path.get(departure_dp_key):
+                g.fallback_direct_path[departure_dp_key] = self._get_direct_path(instance,
+                                                                                 dep_mode,
+                                                                                 _from,
+                                                                                 departure,
+                                                                                 journey.departure_date_time,
+                                                                                 True)
+            if origins[departure.uri] != 0:
+                departure_direct_path = copy.deepcopy(g.fallback_direct_path[departure_dp_key])
+                _extend_pt_sections_with_direct_path(journey, departure_direct_path)
```
TeXitoi, line 181: This section mixes several things.

First, better to abstract the cache somewhere else, in `self._get_direct_path` for example.

Then, I don't think a duration of 0 is enough for the crow fly. 2 cases when it's no the case:
- when you need to walk less that a second (from coord to stop point).
- when there is no street network, kraken will use a crow fly as an approximation (see sncf coverage for example) (but maybe this case is better handled in get_direct_path of kraken)

source/jormungandr/jormungandr/scenarios/experimental.py
```diff
@@ -118,18 +101,103 @@ def update_crowfly_duration(instance, fallback_list, mode, stop_area_uri):
...
+                                                   o,
+                                                   d,
+                                                   journey.arrival_date_time,
+                                                   False)
+                if reverse_sections:
+                    dp_journey = _reverse_journeys(dp_journey)
+                g.fallback_direct_path[arrival_dp_key] = dp_journey
+            if destinations[arrival.uri] != 0:
+                arrival_direct_path = copy.deepcopy(g.fallback_direct_path[arrival_dp_key])
+                _extend_pt_sections_with_direct_path(journey, arrival_direct_path)
```
TeXitoi, line 204: Mmm, déjà-vu! ;-)

source/jormungandr/jormungandr/scenarios/experimental.py
xlqian: - [x] move the cache `into _get_direct_path`
- [ ] handle the crowfly in particular cases

TeXitoi:
OK for me.

PR #1754, opened 2016-09-07
kraken: improve performance of add_contributor

kinnou02:
For a ptref on linelist we would insert the contributor n² time, with n
being the number of lines, this was pretty slow, in production it would
take 6sec for fr-pdl, with these change we are down to less than 100ms.
We had to add another shortcut for the dataset to the networks.

### DIFF
--- a/source/ptreferential/ptref_graph.cpp
+++ b/source/ptreferential/ptref_graph.cpp
@@ -145,6 +145,11 @@ Jointures::Jointures() {
     boost::add_edge(vertex_map.at(Type_e::Dataset), vertex_map.at(Type_e::Route), Edge(2), g);
     // From a StopPoint we can have datasets
     boost::add_edge(vertex_map.at(Type_e::Dataset), vertex_map.at(Type_e::StopPoint), Edge(3), g);
+    // We use a weight to avoid the usage of this path: we want it to be used only when we look for the dataset
+    // or the contributor of a network. If we want the vehicle_journeys of a network we have to use the path:
+    // Network -> Line -> Route -> VJ with a cost a 3 and not the path using the dataset:
+    // Network -> Dataset -> VJ that has a cost of 4
+    boost::add_edge(vertex_map.at(Type_e::Dataset), vertex_map.at(Type_e::Network), Edge(3), g);
 
     // edges for the impacts. for the moment we only need unilateral links,
     // we don't need from an impact all the impacted objects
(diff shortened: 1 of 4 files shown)


source/ptreferential/ptref_graph.cpp
```diff
@@ -145,6 +145,7 @@ Jointures::Jointures() {
     boost::add_edge(vertex_map.at(Type_e::Dataset), vertex_map.at(Type_e::Route), Edge(2), g);
     // From a StopPoint we can have datasets
     boost::add_edge(vertex_map.at(Type_e::Dataset), vertex_map.at(Type_e::StopPoint), Edge(3), g);
+    boost::add_edge(vertex_map.at(Type_e::Dataset), vertex_map.at(Type_e::Network), Edge(3), g);
```
buildbotctp, line 152: why the weight at 3 ? I think it would be nice to comment it :wink:

kinnou02: I don't care about the saying of a bot!

antoine-de, line 152: gnagnagna

source/ptreferential/ptref_graph.cpp
```diff
@@ -145,6 +145,11 @@ Jointures::Jointures() {
     boost::add_edge(vertex_map.at(Type_e::Dataset), vertex_map.at(Type_e::Route), Edge(2), g);
     // From a StopPoint we can have datasets
     boost::add_edge(vertex_map.at(Type_e::Dataset), vertex_map.at(Type_e::StopPoint), Edge(3), g);
+    // We use a weight to avoid the usage of this path: we want it to be used only when we look for the dataset
+    // or the contributor of a network. If we want the vehicle_journeys of a network we have to use the path:
+    // Network -> Line -> Route -> VJ with a cost a 3 and not the path using the dataset:
+    // Network -> Dataset -> VJ that has a cost of 4
```
antoine-de, line 151: but then why the path wouldn't be 
`Network -> Line -> Route -> Contributor`
instead of `Network -> Contributor`

antoine-de:
OK for me

source/ptreferential/ptref_graph.cpp
TeXitoi: 2.5 seems more appropriate.

pbougue: `Route ---2---> Contributor` so we wouldn't use `Network -> Line -> Route -> Contributor` (weight 4).
I guess we can avoid decimal weight as much as possible, then.

PR #1756, opened 2016-09-09
doc: Use JSFiddle directly to avoid bug with external resources not executed on page load

ooga:
Also add resources to get JSFiddle feedback

image


documentation/slate/source/includes/examples.md
```diff
@@ -100,7 +100,7 @@ In the ``links`` section there is for example this link: ``"href": "https://api.
 
 This link is about lines (according to its ``rel`` attribute) and is templated which means that it needs additional parameters.<br>
 The parameters are identified with the ``{`` ``}`` syntax.
-In this case it needs a region id. This id can the found in the ``regions`` section. 
+In this case it needs a region id. This id can the found in the ``regions`` section.
```
TeXitoi: This id can **be** found

ooga: I just remove the trailing whitespace.
I can fix it if you want but it's not the purpose of this PR :)

xlqian:
ok for me

PR #1760, opened 2016-09-12
routing: (almost) support street network when starting to a stop_point

TeXitoi:
This PR makes `max_duration_to_pt` completely forbid using the street network.

Everything is here to activate the street network at the stop points, just remove the 2 TODO lines and update the test by uncommenting them.

### DIFF
--- a/source/jormungandr/tests/ptref_tests.py
+++ b/source/jormungandr/tests/ptref_tests.py
@@ -755,7 +755,7 @@ class TestPtRefRoutingCov(AbstractTestFixture):
 
     def test_headsign_display_info_journeys(self):
         """test basic print of headsign in section for journeys"""
-        response = self.query_region('journeys?from=stop_point:stopB&to=stop_point:stopA&datetime=20120615T000000')
+        response = self.query_region('journeys?from=stop_point:stopB&to=stop_point:stopA&datetime=20120615T000000&max_duration_to_pt=0')
         assert 'error' not in response
         journeys = get_not_null(response, 'journeys')
         eq_(len(journeys), 1)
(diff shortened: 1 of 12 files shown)


source/jormungandr/tests/ptref_tests.py
```diff
@@ -755,7 +755,7 @@ def test_headsign_stop_time_vj(self):
 
     def test_headsign_display_info_journeys(self):
         """test basic print of headsign in section for journeys"""
-        response = self.query_region('journeys?from=stop_point:stopB&to=stop_point:stopA&datetime=20120615T000000')
+        response = self.query_region('journeys?from=stop_point:stopB&to=stop_point:stopA&datetime=20120615T000000&max_duration_to_pt=0')
```
xlqian, line 758: :+1:

source/routing/routing.cpp
```diff
@@ -64,10 +64,13 @@ std::string PathItem::print() const {
     return ss.str();
 }
 
+bool is_same_stop_point(const type::EntryPoint& point, const type::StopPoint* stop_point) {
```
xlqian: static bool ...

TeXitoi: nop, exported and used in raptor_api

source/routing/raptor_api.cpp
```diff
@@ -554,7 +554,9 @@ static void add_pathes(PbCreator& pb_creator,
             const auto& departure_stop_point = path.items.front().stop_points.front();
             georef::Path sn_departure_path = worker.get_path(departure_stop_point->idx);
 
-            if (use_crow_fly(origin, departure_stop_point, sn_departure_path, pb_creator.data)){
+            if (is_same_stop_point(origin, departure_stop_point)) {
+                // nothing in this case
+            } else if (use_crow_fly(origin, departure_stop_point, sn_departure_path, pb_creator.data)){
```
xlqian: How about?

``` c++
if (! is_same_stop_point(origin, departure_stop_point) && use_crow_fly(origin, departure_stop_point, sn_departure_path, pb_creator.data))
```

source/routing/routing.cpp
xlqian: ok

source/routing/routing.cpp
```diff
@@ -64,10 +64,13 @@ std::string PathItem::print() const {
...
+    return point.type == type::Type_e::StopPoint && point.uri == stop_point->uri;
```
xlqian: just a detail, if stop_point can't be null, I prefer use a ref, if stop_point can be null, it should be checked :)

source/routing/raptor_api.cpp
TeXitoi: It'll create a walking section.

xlqian:
ok for me!

PR #1764, opened 2016-09-15
kraken: parameter type modified to float

kadhikari:
ticket concerned: http://jira.canaltp.fr/browse/ITI-121


pbougue:
Wah! Would be nice to add some tests on that...

kadhikari:
@pbougue i pushed some modifications

TeXitoi (approved):
OK for me.

PR #1460, opened 2016-03-07
All: Remove all show_codes

xlqian:
http://jira.canaltp.fr/browse/NAVP-502

PR #1766, opened 2016-09-15
Kraken: Fix disruption status when calling /lines

eturck:
Hello!

We notice a strange behavior while playing with disruptions. When we were adding a disruption on a line the disruption returned by the API when calling /lines always had a status set to "past".
Since PR #1460 action_period is not passed anymore to get_response for line, line_groups and validity pattern. It's probably just a mistake that [has been made while removing show_codes](https://github.com/CanalTP/navitia/pull/1460/files#diff-1dfed77d1f8e2a82b800972f0b7e7d54L52)
I added a jormungandr integration test failing without the second commit.

Let me know if I am wrong or if I need to change anything.

### DIFF
--- a/source/jormungandr/tests/ptref_tests.py
+++ b/source/jormungandr/tests/ptref_tests.py
@@ -311,6 +311,22 @@ class TestPtRef(AbstractTestFixture):
         lg = line_groups[0]
         is_valid_line_group(lg)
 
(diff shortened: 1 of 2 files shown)


antoine-de (approved):
great catch, thanks!

TeXitoi:
OK for me.

PR #1767, opened 2016-09-15
Jormun: limit number of concurrent greenlet

kinnou02:
Currently load balancing of request between kraken isn't fair, since the
introduction of parallel requests to kraken we sometime encounter some
timeouts for fast request with 1sec timeout.
I think that some request get queued behind long request so this PR aim
to limit the number of concurrent call made to kraken during journeys.
The number 3 has been choose "randomly": by default kraken has 4 thread,
so even if all request are routed the same one, we will at least keep
one thread available


TeXitoi:
OK for me.

TeXitoi (approved):
OK for me.

PR #1770, opened 2016-09-16
Test: clean up volumes after teardown

antoine-de:
remove the docker's volume at the teardown of the docker. else the
volume will take some space on the disk.

we should really make have a shared python utils lib to share this
across all repo :wink:


TeXitoi (approved):
OK for me.

kinnou02:
Jenkins, test this please.

PR #1772, opened 2016-09-19
jormun: remove geojson from stop_schedule too!

kinnou02:


pbougue (approved):
Ok for me. Maybe we should just do it everywhere, no?
How about some doc in the same PR? Would be niiiice :)

TeXitoi:
Yeah, and particularly on coverage.

pbougue (changes requested):
Can we add some doc?
(To test this feature also)

TeXitoi (commented):
Could be great:
- support this on autocomplete (pt_objects and places)
- support this on coverage

pbougue (approved):

PR #1779, opened 2016-09-21
CMake: add python protobuf files to CMake protobuf target

pbougue:
also:
- some doc
- some python correction
- some warning removal

### DIFF
--- a/documentation/slate/source/includes/welcome.md
+++ b/documentation/slate/source/includes/welcome.md
@@ -98,7 +98,9 @@ To help you in the building of your project, we have implemented tools to query 
 About the data
 --------------
 
-The street network is extracted from [OpenStreetMap](http://www.openstreetmap.org). The public transport data are provided by networks that provide their timetables as open data. Some data improvements are achieved by Kisio Digital and are published back there <https://navitia.opendatasoft.com>
+The street network is extracted from [OpenStreetMap](http://www.openstreetmap.org). The public transport data are provided by networks that provide their timetables as open data. Some data improvements are achieved by Kisio Digital and are published back there <https://navitia.opendatasoft.com>.
+
+Want to know if your city is in Navitia? Know if a special contributor is used? You can either search in [datasets](#datasets) of the different [coverages](#coverage). Or use the filter provided on our data catalog <https://navitia.opendatasoft.com>.
 
 <aside class="success">
     We are gradually supporting more and more cities. If your city has open public transport data and is missing, drop us a note.
@@ -109,7 +111,7 @@ The street network is extracted from [OpenStreetMap](http://www.openstreetmap.or
 Getting help
 ------------
 
-All available functions are documented in [integration part](#navitia-documentation-v1-interface)
+All available functions are documented in [integration part](#interface)
 
 A mailing list is available to ask question: <a href="mailto:navitia@googlegroups.com">navitia@googlegroups.com</a>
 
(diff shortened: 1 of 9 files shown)


documentation/slate/source/includes/welcome.md
```diff
@@ -98,7 +98,9 @@ To help you in the building of your project, we have implemented tools to query
 About the data
 --------------
 
-The street network is extracted from [OpenStreetMap](http://www.openstreetmap.org). The public transport data are provided by networks that provide their timetables as open data. Some data improvements are achieved by Kisio Digital and are published back there <https://navitia.opendatasoft.com>
+The street network is extracted from [OpenStreetMap](http://www.openstreetmap.org). The public transport data are provided by networks that provide their timetables as open data. Some data improvements are achieved by Kisio Digital and are published back there <https://navitia.opendatasoft.com>.
+
+Want to know if your city is in Navitia? Know if a special contributor is used? You can either search in [datasets](#datasets) of the different [coverages](#coverage). Or use the filter provided on our data catalog <https://navitia.opendatasoft.com>.
```
stifoon, line 104: Tx a lot for that :)
Maybe add a Playground link in the third column?
http://canaltp.github.io/navitia-playground/play.html?request=https%3A%2F%2Fapi.navitia.io%2Fv1%2Fcoverage%2Fsandbox%2Fcontributors%2F%2F%3F&token=3b036afe-0110-4202-b9ed-99718476c2e0

stifoon (changes requested):

stifoon (approved):

source/type/CMakeLists.txt
```diff
@@ -27,38 +27,62 @@ SET(PROTO_SRCS ${CMAKE_CURRENT_BINARY_DIR}/chaos.pb.cc
...
-    WORKING_DIRECTORY "${CMAKE_SOURCE_DIR}/chaos-proto" VERBATIM)
+    WORKING_DIRECTORY "${CMAKE_SOURCE_DIR}/chaos-proto" VERBATIM
+    DEPENDS ${PROTO_FILES})
+
+add_custom_target(protobuf_files DEPENDS ${PROTO_HDRS} ${PROTO_SRCS}
+                  ${NAVITIACOMMON_PROTO_PY} ${MONITOR_KRAKEN_PROTO_PY} ${JORMUNGANDR_TEST_PY})
+set_source_files_properties(${PROTO_HDRS} ${PROTO_SRCS} PROPERTIES GENERATED TRUE)
+
+
+add_library(pb_lib ${PROTO_SRCS} pb_converter.cpp ${NAVITIACOMMON_PROTO_PY} ${MONITOR_KRAKEN_PROTO_PY} ${JORMUNGANDR_TEST_PY})
```
kinnou02: Do you really need to add the python file to the source of the c++ library? ><

pbougue: Nope, I just put it where it was. And it's only a dependancy, not a source.
But I'm open to change that, I just don't see to what other target I should attach that dependancy.

pbougue: @TeXitoi any opinion on that?

TeXitoi: you can put a depedency on all and on jormungandr related tests

pbougue: Maybe just to the 'test' target... but then we'd have to document that better (expose the target and even the command itself in install processes and scripts)

pbougue: So : added "lonely" to the `all` target, and still no dependency between `all` and `test` (like there is no dependency between `main_routing_test` and `jormungandr_integration_test`).
The advice is still to run `make && make test` instead of simply `make test`...

documentation/slate/source/includes/welcome.md
pbougue: Not so found of it, perharps more an output example, but this is pretty self-explanatory so not in this PR :smile:

PR #1783, opened 2016-09-23
Jormun: Can't explicitly disable bss_stands or disable_geojson

eturck:
We can't explicitly disable those 2 parameters when calling the API since the type of the argument is set to `bool` with add_argument. As long as the parameter is passed to the API with a value the argument will be set to True, even if this value is false. I changed it to use boolean, declared in flask.ext.restful.inputs, already used for other boolean parameters.

### DIFF
--- a/source/jormungandr/tests/ptref_tests.py
+++ b/source/jormungandr/tests/ptref_tests.py
@@ -225,7 +225,7 @@ class TestPtRef(AbstractTestFixture):
 
     def test_line(self):
         """test line formating"""
-        response = self.query_region("v1/lines")
+        response = self.query_region("v1/lines?disable_geojson=false")
 
         lines = get_not_null(response, 'lines')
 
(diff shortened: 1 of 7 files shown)


source/jormungandr/tests/ptref_tests.py
```diff
@@ -225,7 +225,7 @@ def test_dataset_by_contributor(self):
 
     def test_line(self):
         """test line formating"""
-        response = self.query_region("v1/lines")
```
pbougue, line 228: I'd rather have a dedicated test just testing that part (copy paste of the beginning of this one), but...

pbougue (approved):
Very nice catch! Thanks :blush:

source/jormungandr/tests/ptref_tests.py
eturck: I wasn't sure if you were trying to have a limited number of test or not :).
I moved it in a separate one.

pbougue:
Also from the commits: :+1:  TDD

PR #1784, opened 2016-09-23
Tyr: post data

prhod:
Adding a /data/instances/string:instance_name/ API (POST only) to allow sending data by API. It may reduce Devops actions when configuring shared directories.
It will be used by tartare in a near future to send geo_data to several tyr environnements.

### DIFF
--- a/source/tyr/Procfile
+++ b/source/tyr/Procfile
@@ -1,4 +1,3 @@
 web: ./manage_tyr.py runserver
 worker: celery worker -A tyr.tasks
 scheduler: celery beat -A tyr.tasks
-reloader_at: ./manage_tyr.py at_reloader
--- a/source/tyr/Procfile
+++ b/source/tyr/Procfile
@@ -1,3 +1,4 @@
 web: ./manage_tyr.py runserver
 worker: celery worker -A tyr.tasks
 scheduler: celery beat -A tyr.tasks
+reloader_at: ./manage_tyr.py at_reloader
--- a/source/tyr/tests/integration/datapost_test.py
+++ b/source/tyr/tests/integration/datapost_test.py
@@ -4,16 +4,17 @@ import pytest
 import os
 from navitiacommon import models
 from tyr import app
+import logging
 
-@pytest.fixture
-def create_instance2():
+@pytest.fixture(scope="module")
+def create_instance_fr():
     with app.app_context():
         instance = models.Instance('fr')
         models.db.session.add(instance)
         models.db.session.commit()
         return instance.id
 
-def test_post_pbf(create_instance2):
+def test_post_pbf(create_instance_fr):
     filename = 'empty_pbf.osm.pbf'
     path = os.path.join(os.path.dirname(os.path.dirname(os.path.dirname(__file__))), 'tests/fixtures/', filename)
     files = {'file': (open(path, 'rb'), filename)}
@@ -21,4 +22,16 @@ def test_post_pbf(create_instance2):
     tester = app.test_client()
     resp = tester.post('/v0/data/instances/fr', data=files)
     assert resp.status_code == 200
-    #assert os.path.isfile()
+    assert os.path.isfile('/tmp/empty_pbf.osm.pbf')
+    os.remove('/tmp/empty_pbf.osm.pbf')
+    assert (not os.path.isfile('/tmp/empty_pbf.osm.pbf'))
+
+def test_post_no_file(create_instance_fr):
+    tester = app.test_client()
+    resp = tester.post('/v0/data/instances/fr')
+    assert resp.status_code == 404
+
+def test_get(create_instance_fr):
+    tester = app.test_client()
+    resp = tester.get('/v0/data/instances/fr')
+    assert resp.status_code == 400
--- a/source/tyr/tests/integration/datapost_test.py
+++ b/source/tyr/tests/integration/datapost_test.py
@@ -1,0 +1,24 @@
+from tests.check_utils import api_get, api_post, api_delete, api_put, _dt
+import json
+import pytest
+from navitiacommon import models
+from tyr import app
+
+@pytest.fixture
+def create_instance():
+    with app.app_context():
+        instance = models.Instance('fr')
+        models.db.session.add(instance)
+        models.db.session.commit()
+        return instance.id
+
+def test_post_pbf(create_instance):
+    filename = 'empty_pbf.osm.pbf'
+    path = os.path.join(os.path.dirname(os.path.dirname(os.path.dirname(__file__))), 'fixtures/', filename)
+    files = {'file': (open(path, 'rb'), filename)}
+
+    resp = api_post('/v0/data/instances/{}'.fomat(create_instance), data=files)
+    assert resp.status_code == 200
+    assert len(resp) == 1
+    assert resp[0]['name'] == 'fr'
+    assert resp[0]['id'] == create_instance
--- a/source/tyr/tests/integration/datapost_test.py
+++ b/source/tyr/tests/integration/datapost_test.py
@@ -1,4 +1,4 @@
-from tests.check_utils import api_get, api_post, api_delete, api_put, _dt
+from tests.check_utils import api_get, api_post, api_delete, api_put, _dt, _to_json
 import json
 import pytest
 import os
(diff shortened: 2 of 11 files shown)


source/tyr/tests/integration/datapost_test.py
```diff
@@ -0,0 +1,24 @@
...
+
+def test_post_pbf(create_instance2):
+    filename = 'empty_pbf.osm.pbf'
+    path = os.path.join(os.path.dirname(os.path.dirname(os.path.dirname(__file__))), 'tests/fixtures/', filename)
+    files = {'file': (open(path, 'rb'), filename)}
+
+    tester = app.test_client()
+    resp = tester.post('/v0/data/instances/fr', data=files)
+    assert resp.status_code == 200
+    #assert os.path.isfile()
```
pbougue: to be activated, verifying that the files exists, then remove (I guess we should do that in a docker for safety and cleanness)

source/tyr/tyr/default_settings.py
```diff
@@ -23,7 +23,7 @@
 
 
 #Path to the directory where the configuration file of each instance of ed are defined
-INSTANCES_DIR = '.'
+INSTANCES_DIR = './tests/fixtures/'
```
pbougue: maybe it should be moved to a specific test_settings.py (like we do in jormun)

source/tyr/tyr/resources.py
```diff
@@ -1113,3 +1116,25 @@ def delete(self, name=None):
...
+        if instance is None:
+            return {'message': 'bad instance {}'.format(instance_name)}, 404
+
+        if not request.files:
+            return {'message': 'the Data file is missing'}, 400
+        content = request.files['file']
+        logger = logging.getLogger(__name__)
+        logger.info('content received: %s', content)
+
+        instance = load_instance_config('fr')
```
pbougue: should be called once (probably at startup of api) worst case when no config is loaded, but not reload at each call.

source/tyr/Procfile
```diff
@@ -1,3 +1,4 @@
 web: ./manage_tyr.py runserver
 worker: celery worker -A tyr.tasks
 scheduler: celery beat -A tyr.tasks
+reloader_at: ./manage_tyr.py at_reloader
```
pbougue: reloader_at stands for what? especially "at"

pbougue (changes requested):
this will one day save devops :) nice!

source/tyr/tests/integration/datapost_test.py
prhod: fixed

source/tyr/tyr/default_settings.py
prhod: fixed

source/tyr/tyr/resources.py
prhod: I'm not sure of how to do that the clean way. Can we discuss it ? :)

source/tyr/Procfile
prhod: I just followed the doc : https://github.com/CanalTP/navitia/blob/dev/source/tyr/readme.md#example-with-honcho
If it's no use, I can delete it ^^

source/tyr/tyr/resources.py
pbougue: I think we should discuss that with @kinnou02

source/tyr/Procfile
kinnou02: reloader_at is dead since a long time :)

source/tyr/tyr/resources.py
```diff
@@ -1113,3 +1116,28 @@ def delete(self, name=None):
             raise
         return ({}, 204)
 
+class Data(flask_restful.Resource):
+    def post(self, instance_name):
+        instance = models.Instance.query_existing().filter_by(name=instance_name).first_or_404()
+        if instance is None:
```
kinnou02: the `first_or_404()` should do that

source/tyr/tyr/resources.py
kinnou02: The `fr` should be `instance.name`. 
There is not real advantage to load the configuration only one time at startup, reading a file is way faster than sending a few hundreds Mo over the network
You reuse the variable instance for storing something else, I don't like that: it's error prone.

source/tyr/tyr/resources.py
```diff
@@ -1113,3 +1116,28 @@ def delete(self, name=None):
             raise
         return ({}, 204)
 
+class Data(flask_restful.Resource):
+    def post(self, instance_name):
```
kinnou02: Maybe this function should be part of the `Job`api since the purpose of it is to create a new job.

source/tyr/tyr/resources.py
```diff
@@ -1113,3 +1116,28 @@ def delete(self, name=None):
...
+        instance = models.Instance.query_existing().filter_by(name=instance_name).first_or_404()
+        if instance is None:
+            return {'message': 'bad instance {}'.format(instance_name)}, 404
+
+        if not request.files:
+            return {'message': 'the Data file is missing'}, 400
+        content = request.files['file']
+        logger = logging.getLogger(__name__)
```
kinnou02: You should use `get_instance_logger`, it will configure the logger so we can keep trace of the instance: using a the instance name as tag in syslog, or using a file named from the instance with file log storage

source/tyr/tests/integration/datapost_test.py
```diff
@@ -0,0 +1,37 @@
...
+    filename = 'empty_pbf.osm.pbf'
+    path = os.path.join(os.path.dirname(os.path.dirname(os.path.dirname(__file__))), 'tests/fixtures/', filename)
+    files = {'file': (open(path, 'rb'), filename)}
+
+    tester = app.test_client()
+    resp = tester.post('/v0/data/instances/fr', data=files)
+    assert resp.status_code == 200
+    assert os.path.isfile('/tmp/empty_pbf.osm.pbf')
+    os.remove('/tmp/empty_pbf.osm.pbf')
+    assert (not os.path.isfile('/tmp/empty_pbf.osm.pbf'))
```
kinnou02, line 33: You should assert that the file doesn't exist at first, we want to be sure that we are the one having created this file

source/tyr/tests/integration/datapost_test.py
```diff
@@ -0,0 +1,37 @@
...
+    with app.app_context():
+        instance = models.Instance('fr')
+        models.db.session.add(instance)
+        models.db.session.commit()
+        return instance.id
+
+def test_post_pbf(create_instance_fr):
+    filename = 'empty_pbf.osm.pbf'
+    path = os.path.join(os.path.dirname(os.path.dirname(os.path.dirname(__file__))), 'tests/fixtures/', filename)
+    files = {'file': (open(path, 'rb'), filename)}
```
kinnou02: Does flask's tester close the file handler?

kinnou02 (changes requested):

source/tyr/tyr/resources.py
prhod: In the first place, we wanted to let the directory scan create the job. Is there a problem with that ?

source/tyr/Procfile
prhod: removed from ProcFile and readme.md

source/tyr/tyr/resources.py
prhod: removed

source/tyr/tyr/resources.py
prhod: fixed

source/tyr/tests/integration/datapost_test.py
prhod: the close of the file handler is added (i don't know how to check the flask's tester behaviour)

source/tyr/tests/integration/datapost_test.py
prhod: tests have been completed

source/tyr/tyr/resources.py
prhod: the `fr` have been fixed (thanks for the notification !). I'm not sure if I need to do something else ^^

source/tyr/tyr/resources.py
```diff
@@ -1113,3 +1116,26 @@ def delete(self, name=None):
...
+            return {'message': 'the Data file is missing'}, 400
+        content = request.files['file']
+        logger = get_instance_logger(instance)
+        logger.info('content received: %s', content)
+
+        instance = load_instance_config(instance_name)
+        if not os.path.exists(instance.source_directory):
+            return ({'error': 'input folder unavailable'}, 500)
+
+        content.save(os.path.join(instance.source_directory, content.filename + ".tmp"))
```
kinnou02: you should use `full_file_name` here

prhod: fixed

prhod:
After discussion with @kinnou02 , the /data/ api has been moved to the POST method of the /jobs/<instance> API

source/tyr/tests/integration/datapost_test.py
```diff
@@ -0,0 +1,43 @@
...
+        models.db.session.add(instance)
+        models.db.session.commit()
+        return instance.id
+
+def test_post_pbf(create_instance_fr):
+    assert (not os.path.isfile('/tmp/empty_pbf.osm.pbf'))
+
+    filename = 'empty_pbf.osm.pbf'
+    path = os.path.join(os.path.dirname(os.path.dirname(os.path.dirname(__file__))), 'tests/fixtures/', filename)
+    f = open(path, 'rb')
```
kinnou02, line 22: what you have done works, for the next time you should use `with` it's better

kinnou02 (approved):

pbougue (approved):

PR #1787, opened 2016-09-26
Tyr: Update shape on user

patochectp:


source/tyr/tests/integration/users_test.py
```diff
@@ -415,17 +416,20 @@ def test_update_invalid_user(mock_rabbit):
     assert status == 404
     assert mock_rabbit.call_count == 0
 
-def test_update_user(create_multiple_users, mock_rabbit):
+def test_update_user(create_multiple_users, mock_rabbit, geojson_feature):
     """
     we update a user
     """
-    user = {'login': 'user1', 'email': 'user1@example.com'}
+    user = {'login': 'user1', 'email': 'user1@example.com', 'shape': geojson_feature}
```
pbougue, line 433: Not sure : you don't test shape, here, but you should (`has_shape` also)? Maybe also test the status?

pbougue (commented):
This should be rebased
And also use ujson in test?

source/tyr/tests/integration/users_test.py
pbougue: Maybe we should wait for the merge of the "display" PR to add tests...
And would be nice to test that updating shape with `{}` doesn't change the shape.

datanel: +1 for the status

> And would be nice to test that updating shape with {} doesn't change the shape.

This will be achieved posting a json without shape attribute

pbougue: To sum it up (As @datanel and I) said:
- Test that passing `{}` as shape doesn't change the shape
- Test that passing no shape doesn't change the shape
- Test that passing None deletes the shape
- Test that it is possible to pass `has_shape` and it doesn't crash (it does nothing, but also doesn't crash)

The main idea is when you GET, then PUT the result of this GET : OK and nothing change

patochectp: api_put test that status =200

PR #1789, opened 2016-09-26
Tyr: List users with shape

patochectp:


source/tyr/tyr/resources.py
```diff
@@ -454,11 +469,19 @@ def map_args_to_instance(attr_name):
 
 class User(flask_restful.Resource):
     def get(self, user_id=None):
+        parser = reqparse.RequestParser()
+        parser.add_argument('disable_geojson', type=bool,
+                            default=False,
```
pbougue, line 474: default is True (will change tests) : we don't want to output shape by default

source/tyr/tests/integration/users_test.py
```diff
@@ -494,3 +504,49 @@ def test_deletion_keys_and_auth(create_instance, mock_rabbit):
...
     assert len(resp_auth['authorizations']) == 0
     assert mock_rabbit.called
+
+
+def test_get_user_with_shape(create_user, geojson_feature_collection):
+    """
+    We start by creating the user with a shape,
+    and we test that the attribute shape is returned and has_shape = True
+    """
+    resp = api_get('/v0/users/{}'.format(create_user))
```
pbougue, line 514: will need explicit `disable_geojson=False`

source/tyr/tests/integration/users_test.py
```diff
@@ -494,3 +504,49 @@ def test_deletion_keys_and_auth(create_instance, mock_rabbit):
...
+
+    assert resp['has_shape'] == True
+    assert resp['shape'] == geojson_feature_collection
+
+
+def test_get_user_with_shape_and_disable_geojson_param_true(create_user, geojson_feature_collection):
```
pbougue, line 520: would be nice to test the same but not providing param, see that we default to `disable_geojson=True`

pbougue (changes requested):
Looks good to me except for the default value

source/tyr/tyr/resources.py
datanel: > we don't want to output shape by default

why?

copied from **https://github.com/CanalTP/navitia/blob/dev/source/jormungandr/jormungandr/interfaces/v1/Disruptions.py#L91**

datanel: I merge and will make another PR if we change this behaviour

PR #1790, opened 2016-09-27
georef: remove curly brackets

patochectp:
curly brackets are not compatible with libboost 1.58.0.1


TeXitoi:
That's because of explicit constructor, but conversion operator is functional.

TeXitoi (approved):

PR #1792, opened 2016-09-28
Tyr: change shape validation error

patochectp:


pbougue:
Jenkins, test this please

source/navitiacommon/navitiacommon/parser_args_type.py
```diff
@@ -28,7 +28,8 @@
 # IRC #navitia on freenode
 # https://groups.google.com/d/forum/navitia
 # www.navitia.io
-
+import json
```
kinnou02: In jormungandr we use `ujson` it's way faster than `json`

patochectp: done

PR #1795, opened 2016-09-29
Experimental: neglect fallback when max_duration_to_pt or max_duration equals zero

xlqian:


source/jormungandr/jormungandr/scenarios/experimental.py
```diff
@@ -155,6 +155,14 @@ def _get_stop_points(self, instance, place, mode, max_duration, reverse=False, m
         # we use place_nearby of kraken at the first place to get stop_points around the place, then call the
         # one_to_many(or many_to_one according to the arg "reverse") service to take street network into consideration
         # TODO: reverse is not handled as so far
+
+        # When max_duration is 0, there is no need to compute the fallback to pt, except if place is a stop_point or a
+        # stop_area
+        if max_duration == 0:
+            if 'stop_point' in place.uri:
```
TeXitoi: We can't rely on string to determine the type.

TeXitoi (changes requested):

TeXitoi (approved):

PR #1804, opened 2016-10-03
Doc: Add isochrone.boundary_duration[] parameter

stifoon:


documentation/slate/source/includes/apis.md
```diff
@@ -1029,7 +1029,9 @@ direction       | int                    | Angle (in degree) between the previou
 <a name="isochrones_api"></a>Isochrones (currently in Beta)
 ---------------------------------------
 
->[Try it on Navitia playground (click on "MAP" buttons for "wow effect")](http://canaltp.github.io/navitia-playground/play.html?request=https%3A%2F%2Fapi.navitia.io%2Fv1%2Fcoverage%2Fsandbox%2Fisochrones%3Ffrom%3D2.377097%3B48.846905%26datetime%3D20160505T080000%26max_duration%3D2000%26min_duration%3D1000)
+>[Try a simple example on Navitia playground](http://canaltp.github.io/navitia-playground/play.html?request=https%3A%2F%2Fapi.navitia.io%2Fv1%2Fcoverage%2Fsandbox%2Fisochrones%3Ffrom%3D2.377097%3B48.846905%26max_duration%3D2000%26min_duration%3D1000)
```
pbougue: You should add the sandbox token here

pbougue: Maybe talk of the "MAP" button in the first of the 2 explainations?

pbougue: And the reverse-geocoder is fun : we are Ruelle Fraisier, now ;)

stifoon: I can't do more accurate coordinates: it depends on geographical data. It seems there is no longer Paris OSM data :(

PR #1809, opened 2016-10-04
jormungandr: fix error on use of user shape in autocomplete

patochectp:


source/jormungandr/jormungandr/interfaces/v1/Places.py
```diff
@@ -249,8 +250,8 @@ def get(self, region=None, lon=None, lat=None):
             if global_autocomplete:
                 user = authentication.get_user(token=authentication.get_token(), abort_if_no_token=False)
                 shape = None
-                if user:
-                    shape = user.shape
+                if user and user.shape :
+                    shape = ujson.loads(user.shape)
```
pbougue: Isn't there a `else` to output `null` (not string) if no shape is there?

pbougue: Oups, ok :)
Can we test that? ... Not sure it's easy

source/jormungandr/jormungandr/interfaces/v1/Places.py
```diff
@@ -49,6 +49,7 @@
 from flask_restful import marshal, marshal_with
 import datetime
 from jormungandr.parking_space_availability.bss.stands_manager import ManageStands
+import ujson
```
kinnou02: for simplicity we always alias ujson as json:

```
import ujson as json
```

PR #1810, opened 2016-10-05
Jormun: Don't clear the cache on kraken error, it's only making all thing slower

kinnou02:
Currently a dead socket will always trigger a purge of the cache.
If an instance is dead this will trigger a purge of the cache every ten seconds. 
In most of the case it only a transient error and the next request will create a new connection that should be routed to a healthy kraken.

On the production during the last six hours we have purged the cache 100 000 times making it useless.
Without cache call to /journeys with ids but without coverage is much slower since it need to call each instances.


kinnou02:
This is a first step, I aim at the simplification of the instance initialization process

PR #1812, opened 2016-10-05
jormungandr: removal of thread ping

kinnou02:
Currently each jormun call every instance each 10 seconds to check if
it's still alive. It's useless since we don't do anything of this
information and it generate a lot of unneeded requests to kraken (remember
there are 60 jormungandr in production)

I kept the thread just to ensure that we call each instance at least one
time after the startup, this way we have the geom and the timezone of
the instance.

With this change after the startup jormungandr will only call kraken
when handling a request. This will allow us to spawn kraken only when
we need it (something based on xinetd probably) on some testing environments.

### DIFF
--- a/source/jormungandr/jormungandr/instance.py
+++ b/source/jormungandr/jormungandr/instance.py
@@ -83,7 +83,7 @@ class Instance(object):
         self.name = name
         self.timezone = None  # timezone will be fetched from the kraken
         self.publication_date = -1
-        self.is_initialized = False#kraken hasn't been call yet we don't have geom nor timezone
+        self.is_initialized = False #kraken hasn't been called yet we don't have geom nor timezone
         self.breaker = pybreaker.CircuitBreaker(fail_max=app.config['CIRCUIT_BREAKER_MAX_INSTANCE_FAIL'],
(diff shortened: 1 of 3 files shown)


source/jormungandr/jormungandr/instance.py
```diff
@@ -83,7 +83,7 @@ def __init__(self, context, name, zmq_socket, street_network_configuration=None,
         self.name = name
         self.timezone = None  # timezone will be fetched from the kraken
         self.publication_date = -1
-        self.is_up = True
+        self.is_initialized = False#kraken hasn't been call yet we don't have geom nor timezone
```
pbougue: minor: spacing + `called`

source/jormungandr/jormungandr/instance_manager.py
```diff
@@ -168,11 +168,18 @@ def init_kraken_instances(self):
...
         for instance in self.instances.values():
-            purge_cache_needed = instance.init() or purge_cache_needed
-        if purge_cache_needed:
-            self._clear_cache()
+            if not instance.is_initialized:
+                futures.append(gevent.spawn(instance.init))
+
+        gevent.wait(futures)
+        for future in futures:
+            #we check if an instance need the cache to be purged
```
pbougue: need`s`

pbougue (approved):

PR #1817, opened 2016-10-06
Tyr: Fix shape value when is "null" in database

datanel:
In database, shape is a TEXT. 
So we convert the json to a string with `json.dumps(shape)`
but `json.dumps(None) = "null"` so in database we have a string 'null'

When we get a user with `shape = "null"` we convert the string with `json.loads(shape)`
but `json.loads("null") = {}` and not `None`

There is a better fix https://github.com/CanalTP/navitia/pull/1813 for postgres >= 9.2

### DIFF
--- a/source/tyr/tests/integration/users_test.py
+++ b/source/tyr/tests/integration/users_test.py
@@ -172,6 +172,43 @@ def test_get_users_empty():
     resp = api_get('/v0/users/')
     assert resp == []
 
+
+def test_add_user_without_shape(mock_rabbit):
+    """
+    creation of a user without shape
+    When we get this user, we should see
+    shape = None and has_shape = False
+    """
+    user = {'login': 'user1', 'email': 'user1@example.com'}
+    data = json.dumps(user)
+    resp = api_post('/v0/users/', data=data, content_type='application/json')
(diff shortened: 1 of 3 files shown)


source/tyr/tests/integration/users_test.py
```diff
@@ -172,6 +172,42 @@ def test_get_users_empty():
...
+        gen = (k for k in user if k is not 'shape')
+        for k in gen:
+            assert u[k] == user[k]
+        assert u['end_point']['name'] == 'navitia.io'
+        assert u['type'] == 'with_free_instances'
+        assert u['block_until'] == None
+
+    check(resp)
+    assert resp['shape'] == None
+    assert mock_rabbit.called
```
pbougue, line 196: you can test "has_shape" also

pbougue (changes requested):
As we said, the PUT doesn't work without a db migration (shape "null" instead of NULL by default)

source/tyr/tests/integration/users_test.py
datanel: done

pbougue (approved):

PR #1822, opened 2016-10-12
Jormungandr: use speed in experimental

xlqian:


source/jormungandr/tests/routing_tests_experimental.py
```diff
@@ -74,37 +74,45 @@ def check_previous_datetime_link(dt, response, clockwise):
...
-        super(JourneyCommon, self).test_sp_to_sp()
+        super(TestJourneysExperimental, self).test_sp_to_sp()
 
-    @skip("temporarily disabled")
     def test_speed_factor_direct_path(self):
-        super(JourneyCommon, self).test_speed_factor_direct_path()
+        """
+        This feature is no longer supported
+        """
+        pass
```
TeXitoi: Looking at the title of the test, you should just use another way to get the direct path that don't use type filter. Using max_duration=0 should do the trick

source/jormungandr/tests/routing_tests_new_default.py
```diff
@@ -75,25 +75,35 @@ def check_previous_datetime_link(dt, response, clockwise):
...
+        """
+        pass
 
-    @skip("temporarily disabled")
     def test_speed_factor_direct_path(self):
-        super(JourneyCommon, self).test_speed_factor_direct_path()
+        """
+        This feature is no longer supported
+        """
+        pass
```
TeXitoi: same as in experimental.

TeXitoi (changes requested):

source/jormungandr/tests/routing_tests_experimental.py
xlqian: you're right, this test shouldn't be overwritten...

source/jormungandr/tests/routing_tests_new_default.py
xlqian: ok

TeXitoi (approved):

TeXitoi:
It should fix a lot of artemis tests.

PR #1825, opened 2016-10-13
tyr: add retrying when redis error while locking

kinnou02:
We retry the full task if the error happens while locking. If it's while
releasing the lock we only retry to release it and if we don't success
after 5 times we continue. Hopefuly we won't need to do something on that
instance before the timeout of the lock :(
Refer to: http://jira.canaltp.fr/browse/NAVITIAII-2171

### DIFF
--- a/source/tyr/tyr/binarisation.py
+++ b/source/tyr/tyr/binarisation.py
@@ -53,6 +53,8 @@ from navitiacommon import models
 from tyr.helper import get_instance_logger, get_named_arg
 from contextlib import contextmanager
 import glob
+from redis.exceptions import ConnectionError
+import retrying
 
 
 def unzip_if_needed(filename):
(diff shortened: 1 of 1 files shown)


source/tyr/tyr/binarisation.py
```diff
@@ -103,20 +105,25 @@ def wrapper(*args, **kwargs):
...
+                logger.info('lock on %s retry %s in %s sec', job.instance.name, func.__name__, countdown)
                 task.retry(countdown=countdown, max_retries=10)
             else:
                 try:
                     logger.debug('lock acquired on %s for %s', job.instance.name, func.__name__)
                     return func(*args, **kwargs)
                 finally:
                     logger.debug('release lock on %s for %s', job.instance.name, func.__name__)
-                    lock.release()
+                    #sometime we have been disctonnected from redis when we want to release the lock, so we retry only the release
```
pbougue: minor: sometime`s`, `disconnected`, line length

pbougue (approved):
For the test part, are there some on the usual part? (not on the error and lock, but check that nominal is still ok).
I guess there is still artemis.

kinnou02:
There is currently no automatic tests on this part except for Artemis

PR #1826, opened 2016-10-13
Jormungandr: fix test_sp_to_sp for experimental

xlqian:


source/jormungandr/tests/journey_common_tests.py
```diff
@@ -376,11 +376,22 @@ def test_sp_to_sp(self):
...
+            assert 'non_pt_walking' in journeys[0]['tags']
 
         # with street network activated
         response = self.query_region(query + "&max_duration_to_pt=1")
-        eq_(len(response['journeys']), 1)
+        if current_scenario in ('default', 'new_default'):
+            eq_(len(response['journeys']), 1)
+        elif current_scenario == 'experimental':
+            eq_(len(response['journeys']), 2)
+
```
TeXitoi, line 394: I'd prefer to check that the first (resp last) section is a public_transport section with from (resp to) = request from (resp to).

TeXitoi (changes requested):

source/jormungandr/tests/journey_common_tests.py
TeXitoi: Sorry, that's not the problem. But we shouldn't have this journey if we want to be isofunctional. Maybe that should be done in another ticket.

TeXitoi (approved):

source/jormungandr/tests/journey_common_tests.py
xlqian: i created a ticket for that

TeXitoi:
Will be fixed globally in another PR

PR #1829, opened 2016-10-17
Jormungandr: simplify the "no token" message

stifoon:
As a great idea from our entertainer @gcrouigneau \o/

### DIFF
--- a/source/jormungandr/jormungandr/authentication.py
+++ b/source/jormungandr/jormungandr/authentication.py
@@ -161,7 +161,7 @@ def get_user(token, abort_if_no_token=True):
             #a token is mandatory for non public jormungandr
             if not current_app.config.get('PUBLIC', False):
                 if abort_if_no_token:
-                    flask_restful.abort(401, message='no token. You can get one at http://www.navitia.io')
+                    flask_restful.abort(401, message='no token. You can get one at http://www.navitia.io or contact your support if you’re using the opensource version of Navitia https://github.com/CanalTP/navitia')
                 else:
                     return None
             else:  # for public one we allow unknown user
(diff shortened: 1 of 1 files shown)


source/jormungandr/jormungandr/authentication.py
```diff
@@ -161,7 +161,7 @@ def get_user(token, abort_if_no_token=True):
             #a token is mandatory for non public jormungandr
             if not current_app.config.get('PUBLIC', False):
                 if abort_if_no_token:
-                    flask_restful.abort(401, message='no token')
+                    flask_restful.abort(401, message='no token. You can get one at http://www.navitia.io')
```
antoine-de: I think this should be a parameter, because I don't think self hosted versions will like this message :stuck_out_tongue_winking_eye:

source/jormungandr/jormungandr/authentication.py
```diff
@@ -161,7 +161,7 @@ def get_user(token, abort_if_no_token=True):
             #a token is mandatory for non public jormungandr
             if not current_app.config.get('PUBLIC', False):
                 if abort_if_no_token:
-                    flask_restful.abort(401, message='no token')
+                    flask_restful.abort(401, message='no token. You can get one at http://www.navitia.io or contact your support if you’re using the opensource version of Navitia https://github.com/CanalTP/navitia')
```
antoine-de, line 164: I though more of something like:

 `flask_restful.abort(401, message=current_app.config.get(NO_TOKEN_MSG, 'no token. You can get one at http://www.navitia.io'`

pbougue: Boarf, you heard of growth hacking?... Don't forget he's in marketing now ;)

PR #1831, opened 2016-10-20
Kraken: Fix crowfly when last stop_time has arrival != departure

TeXitoi:
fix http://jira.canaltp.fr/browse/ITI-204

Waiting for artemis scenario tests to be merged because some artemis test will fail.


pbougue:
May also fix http://jira.canaltp.fr/browse/NAVITIAII-2173 (to close also if that's the case)

TeXitoi:
This PR will break artemis

PR #1832, opened 2016-10-24
jormungandr: parameters in request used for direct_path

kadhikari:
Since parameters of an instance always have default values,
we should use parameter values of the request.


source/jormungandr/tests/routing_tests_experimental.py
```diff
@@ -118,21 +114,8 @@ def test_max_duration_to_pt_equals_to_0(self):
...
-
-        # second is car
-        assert('car' in response['journeys'][1]['tags'])
-        eq_(len(response['journeys'][1]['sections']), 3)
-
-        # last is walking
-        assert('walking' in response['journeys'][-1]['tags'])
-        eq_(len(response['journeys'][-1]['sections']), 1)
+        # pas de solution
+        assert('journeys' not in response)
```
TeXitoi: All these tests should be in JourneyCommon

TeXitoi (changes requested):

kadhikari:
@TeXitoi i made some corrections

TeXitoi (approved):

PR #1835, opened 2016-10-26
Kraken: Add disable_geojson to pb_converter to reduce protobuf payload size

eturck:
Hello :)

On our navitia instances we are requesting sometime a lot of data at once and the geometries are representing a good 85% of the total size of the response sometimes. The disable_geojson you added is really good since we often don't need those geometries.
On my (crappy) development machine, when requesting all our lines (not much more than a 100) : 

| request | total time | kraken request time | size |
| --- | --- | --- | --- |
| lines?count=200 | 5s | 1s | 5MB |
| lines?count=200&disable_geojson=True | 4s | 1s | 220KB |

We gain 1s and a lot of byte, but it was still taking a lot of time, mainly in protobuf. After some discussions, @kinnou02 proposed to forward disable_geojson to kraken to completely remove the geojson from the protobuf response.

With my (still crappy) machine I have now : 

| request | total time | kraken request time | size |
| --- | --- | --- | --- |
| lines?count=200 | 5s | 1s | 5MB |
| lines?count=200&disable_geojson=True | 300ms | < 100ms | 220KB |

It's way quicker since the protobuf payload is a lot smaller. I've added a disable_geojson attribute to PbCreator and use it for all the timetable and ptref calls, and also for ptobjects. I've also modified navitia-proto (see CanalTP/navitia-proto#76).
Let me know if you need me to change or add anything.


kinnou02:
Jenkins, test this please.

kinnou02:
Jenkins, test this please.

source/jormungandr/jormungandr/scenarios/simple.py
```diff
@@ -148,6 +148,7 @@ def pt_objects(self, request, instance):
         req.requested_api = type_pb2.pt_objects
         req.pt_objects.q = request['q']
         req.pt_objects.depth = request['depth']
+        req.pt_objects.disable_geojson = request['disable_geojson']
```
kinnou02, line 151: At first I wanted to sugest the usage of `get()` to prevent an error in case of the key `disable_geosjon` not being set in the request dict, this should not happens since request parser always fill it. But we don't do it for others attributes and it will require to duplicate all the default values. 

So I don't know what to do, the request parser only handle http request, if we do an internal call we still have to check for the key. I think there is a lot of refactoring to do on this part of jormungandr: Resources, instance manager, instances, scenarios are complex and a lot of thing are duplicate in theses. So I think I will just add this to my todo list

kinnou02 (approved):

TeXitoi (approved):

TeXitoi (changes requested):
You need to update navitia-proto to the master commit. Then, we merge.

Thanks again!

eturck:
I've updated the navitia-proto commit.

TeXitoi (approved):

PR #1840, opened 2016-11-08
Tyr: Update api  poi_types

datanel:
## TODO

- [x] Documentation
- [x] Call osm2ed for validation (to be done in another PR)

### DIFF
--- a/source/navitiacommon/navitiacommon/models.py
+++ b/source/navitiacommon/navitiacommon/models.py
@@ -219,6 +219,19 @@ class PoiType(db.Model):
     def __init__(self, uri, name=None, instance=None):
         self.uri = uri
         self.name = name
+
+
+class PoiTypeJson(db.Model):
+    poi_types_json = db.Column(db.Text, nullable=True)
+    instance_id = db.Column(db.Integer,
+                            db.ForeignKey('instance.id'),
+                            primary_key=True,
+                            nullable=False)
+
+    __tablename__ = 'poi_type_json'
+
+    def __init__(self, poi_types_json, instance=None):
+        self.poi_types_json = poi_types_json
         self.instance = instance
 
 
@@ -235,6 +248,9 @@ class Instance(db.Model):
 
     poi_types = db.relationship('PoiType', backref=backref('instance'),
                                lazy='dynamic', cascade='save-update, merge, delete')
+
+    poi_type_json = db.relationship('PoiTypeJson', uselist=False, backref=backref('instance'),
+                                 cascade='save-update, merge, delete, delete-orphan')
     # ============================================================
     # params for jormungandr
     # ============================================================
--- a/source/tyr/migrations/versions/5ad2dddab995_poi_types_json.py
+++ b/source/tyr/migrations/versions/5ad2dddab995_poi_types_json.py
@@ -1,4 +1,4 @@
-"""empty message
+"""Add poi_type_json table
 
 Revision ID: 5ad2dddab995
 Revises: c5cd5a41def
--- a/source/tyr/migrations/versions/5ad2dddab995_poi_types_json.py
+++ b/source/tyr/migrations/versions/5ad2dddab995_poi_types_json.py
@@ -1,0 +1,27 @@
+"""empty message
+
+Revision ID: 5ad2dddab995
+Revises: c5cd5a41def
+Create Date: 2016-11-04 17:07:25.934771
+
+"""
+
+# revision identifiers, used by Alembic.
+revision = '5ad2dddab995'
+down_revision = 'c5cd5a41def'
+
+from alembic import op
+import sqlalchemy as sa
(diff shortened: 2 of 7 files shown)


source/tyr/readme.md
```diff
@@ -740,52 +740,50 @@ If nothing if configured, default POI are extracted. At this time the list is as
 - uri: "leisure:garden" name: "Jardin"
 - uri: "leisure:park" name: "Zone Parc. Zone verte ouverte, pour déambuler. habituellement municipale"
 
-Let say you want to see all type of poi for the instance fr-bre, you have to call the PoiTypes endpoint for this
+Let say you want to see all poi types for the instance fr-bre, you have to call the PoiTypes endpoint for this
```
pbougue: Let`'s`

source/tyr/readme.md
```diff
@@ -740,52 +740,50 @@ If nothing if configured, default POI are extracted. At this time the list is as
...
-        },
-        {
-            "name": "foo\u00e9",
-            "uri": "foo:baff"
-        }
-    ]
+
+  "poi_types": [
+    {"id": "pdv", "name": "Point de vente"},
+    {"id": "mairie", "name": "Hôtel de ville"}
```
pbougue: We now ask the conf to have at least 2 poi-types with id `amenity:parking` and `amenity:bicycle_rental`.
Not mandatory to use them in rules, but as their use is hard-coded (for the switch with walking in journeys), it would be better if documented.

pbougue (commented):
Doing the call to osm2ed in this PR would be nice.
Maybe we should do all in one PR with kraken #1843 ... To see, because the migration won't be simple.
I guess the simplest for migration is have all being usable... but not easy.

We will also have to open (tagged do_not_merge) the PR removing the old poi conf.

source/tyr/readme.md
pbougue: I guess the easiest is to put the default json (see in #1843 )

source/tyr/readme.md
datanel: done

source/tyr/readme.md
datanel: replaces by default values from https://github.com/CanalTP/navitia/pull/1843

pbougue:
Jenkins, test this please

source/navitiacommon/navitiacommon/models.py
```diff
@@ -219,6 +219,19 @@ class PoiType(db.Model):
...
+
+
+class PoiTypeJson(db.Model):
+    poi_types_json = db.Column(db.Text, nullable=True)
+    instance_id = db.Column(db.Integer,
+                            db.ForeignKey('instance.id'),
+                            primary_key=True,
+                            nullable=False)
+
+    __tablename__ = 'poi_type_json'
```
pbougue, line 231: Can't we just create a new field in instance table?

datanel: make sense as it is a one to one relation

source/tyr/migrations/versions/5ad2dddab995_poi_types_json.py
```diff
@@ -0,0 +1,27 @@
+"""empty message
```
azime: Minor: can you add a message?

source/navitiacommon/navitiacommon/models.py
pbougue: It would be a big field when reading in the table, so we let it as is

source/tyr/readme.md
```diff
@@ -740,52 +740,110 @@ If nothing if configured, default POI are extracted. At this time the list is as
...
-        {
-            "name": "foo",
-            "uri": "foo:ba"
-        },
-        {
-            "name": "foo\u00e9",
-            "uri": "foo:baff"
-        }
-    ]
+  "poi_types": [
```
pbougue, line 749: For memory: I will just put a link to the file in my PR (easier to maintain)

source/tyr/readme.md
```diff
@@ -740,52 +740,110 @@ If nothing if configured, default POI are extracted. At this time the list is as
...
 ```
 
-If you want to add a poi type for an instance you have to POST it:
-
-    curl 'http://localhost:5000/v0/instances/fr-bre/poi_types/foo:baff?name=foo' -X POST
-
-The name of a poi type is optional and can also be passed in a json like this:
-
-    curl 'http://localhost:5000/v0/instances/fr-bre/poi_types/foo:json' -X POST -H 'content-type: application/json' -d '{"name": "json"}'
+If you want to add poi types for an instance you have to POST it:
```
pbougue: For memory (to be changed in my PR also): "you have to POST again all the json (previous AND new poi types)"

pbougue (approved):

PR #1842, opened 2016-11-09
Jormungandr: correctly handle no solution in valhalla

TeXitoi:


source/jormungandr/jormungandr/street_network/valhalla.py
```diff
@@ -237,6 +237,11 @@ def check_response(self, response):
     def direct_path(self, mode, pt_object_origin, pt_object_destination, datetime, clockwise, request):
         data = self._make_data(mode, pt_object_origin, [pt_object_destination], request, api='route')
         r = self._call_valhalla('{}/{}'.format(self.service_url, 'route'), requests.post, data)
+        if r is not None and r.status_code != 200:
```
antoine-de: wouldn't it be great to log the valhalla response ?

TeXitoi: This case should just be "no solution" case. Maybe the test is too large.

xlqian: @antoine-de we can, but be aware that this response will be very "fat"

antoine-de: even for error the response can be fat ?

my point @TeXitoi if it happens that the test is a bit broad I think it would be nice to at least see it in the logs

TeXitoi: I prefer a better test to useless log. I look if I can do better

source/jormungandr/jormungandr/street_network/tests/valhalla_test.py
```diff
@@ -361,20 +361,45 @@ def direct_path_func_without_response_valhalla_test():
...
+                                 destination,
+                                 str_to_time_stamp('20161010T152000'),
+                                 True,
+                                 MOCKED_REQUEST)
+        assert str(excinfo.value) == '500: Internal Server Error'
+        assert str(excinfo.value.data['message']) == 'Valhalla service unavailable, impossible to query : http://bob.com/route'
+        assert str(excinfo.typename) == 'TechnicalError'
+
+
+def direct_path_func_with_no_response_valhalla_test():
```
xlqian, line 384: :+1:

xlqian (approved):

pbougue:
Jenkins, test this please

pbougue (commented):
No tests on n*m? Don't know if it's easily doable, though.

TeXitoi:
@pbougue done

xlqian (approved):

pbougue:
needs #1845 to be merged

PR #1843, opened 2016-11-09
Ed: refacto osm2ed to allow one poi-type from multi osm tags

pbougue:
POI-type params will now:
* allow to create one POI-type from multiple OSM-tags (all are required for the OSM-object to to be typed)
* allow to use OSM-tag key or value that includes a `:`
* fill POI name with its POI-type label if no name is provided by OSM
* allow a simple check of params: call `osm2ed -p {<my json poi-type params>}` and test that it returns OK (`0`). It may also be possible to get standard output and print it in tyr json.

Default poi-type param: see source/ed/default_poi_types.h

WARNING: does not compile on debian7, as it uses `boost::property_tree::read_json` which has the bug https://svn.boost.org/trac/boost/ticket/6785 in boost 1.49, corrected in 1.50.
So after the merge of this PR, no coming back on debian7 vanilla.

As discussed with @eturck we upgrade the requirement for boost to 1.55 (version from debian 8, should also allow merge of #1003)

TODO:

- [x] cleanup (pass on "poi-type" and "amenity" elsewhere, rm to_string)
- [x] allow simple json-param check
- [x] test that we have the same nb of poi before/after on paris osm
- [x] call correctly osm2ed from tyr (for json check and regular use)
- [x] tests (jsons for `parse_json_poi_types` and osm tags for `get_applicable_poi_rule`)
- [x] rebase once #1845 is merged
- [x] wait for @eturck to test that boost 1.55 is OK for them (then update utils SHA-1)


source/ed/default_poi_types.h
```diff
@@ -0,0 +1,222 @@
...
+https://groups.google.com/d/forum/navitia
+www.navitia.io
+*/
+
+#pragma once
+
+namespace ed { namespace connectors {
+
+/*
+  To be generated from json bellow, replacing s#"#\"#g and s#\n# \\\n#g
```
antoine-de: I don't see the point of this comment

source/ed/osm2ed.cpp
```diff
@@ -704,11 +707,6 @@ void PoiHouseNumberVisitor::node_callback(uint64_t osm_id, double lon, double la
  * We read another time ways to insert housenumbers and poi
  */
 void PoiHouseNumberVisitor::way_callback(uint64_t osm_id, const CanalTP::Tags &tags, const std::vector<uint64_t> & refs) {
-    if (tags.find("addr:housenumber") == tags.end() &&
```
antoine-de, line 707: do you know the performance impact of this modification ?

source/ed/osm2ed.cpp
```diff
@@ -906,49 +904,55 @@ void PoiHouseNumberVisitor::fill_poi(const u_int64_t osm_id, const CanalTP::Tags
...
-    }
-    for(auto st : tags){
-        if(properties_to_ignore.find(st.first) != properties_to_ignore.end()){
-            if(st.first == "addr:housenumber"){
-                poi.address_number = st.second;
+
+    // the first rule that matches osm object decides poi-type
+    // match : osm object has all osm tags required by rule
+    for (const auto& osm_rule: poi_params.rules) {
+        bool is_valid_rule = true;
```
antoine-de: can't you use `contains_if` there ?

source/ed/osm2ed.cpp
```diff
@@ -968,44 +972,57 @@ void OSMCache::flag_nodes() {
...
+
+    auto logger = log4cplus::Logger::getInstance("log");
+    LOG4CPLUS_INFO(logger, json_params);
+    pt::ptree poi_root;
+    std::istringstream iss (json_params);
+    pt::read_json(iss, poi_root);
+
+    ed::connectors::PoiTypeParams params;
+
+    for (const pt::ptree::value_type &json_poi_type : poi_root.get_child("poi_types")) {
```
antoine-de: spacing

source/ed/default_poi_types.h
```diff
@@ -0,0 +1,222 @@
...
+        {"key": "leisure", "value": "park"}
+      ],
+      "poi_type_id": "leisure:park"
+    }
+  ]
+}
+ */
+
+static std::string DEFAULT_JSON_POI_TYPES =
+ "{ \
```
antoine-de: can't you use a string literal ?

http://en.cppreference.com/w/cpp/language/string_literal

source/ed/osm2ed.cpp
pbougue: nope, will have a look at that

source/ed/osm2ed.cpp
```diff
@@ -902,52 +900,66 @@ void PoiHouseNumberVisitor::fill_housenumber(const uint64_t osm_id,
...
+ * the first rule that matches OSM object decides POI-type
+ * match : OSM object has all OSM tags required by rule
+ */
+const RuleOsmTag2PoiType* PoiHouseNumberVisitor::get_applicable_poi_rule(const CanalTP::Tags& tags) {
+    for (const auto& osm_rule: poi_params.rules) {
+        bool is_valid_rule = true;
+        for (const auto& osm_rule_tag: osm_rule.osm_tag_filters) {
+            const auto it_poi_tag = tags.find(osm_rule_tag.first);
+            if (it_poi_tag == tags.end() || it_poi_tag->second != osm_rule_tag.second) {
+                is_valid_rule = false;
```
pbougue: `contains_if` seems hard to use here as we also want to check that the value is good

TeXitoi: contains_if will loop over every elements...

source/ed/osm2ed.cpp
pbougue: Done

antoine-de: and there is no impact ?

source/ed/osm2ed.cpp
antoine-de: boost::find_if does not short circuit ?

source/ed/osm2ed.cpp
pbougue: On paris.osm.pbf, the new version (filtering using `get_applicable_poi_rule` that I just changed) is 2.4s longer than the old ugly filter (no judgement).

source/ed/osm2ed.cpp
pbougue: I don't see how, it uses the whole pair to apply predicate

pbougue:
You can now review, but before any merge we will have to see how we migrate and how we interface with tyr PR #1840

pbougue:
Jenkins, test this please

pbougue:
Jenkins, test this please

source/ed/osm2ed.cpp
antoine-de: ah there might have been a misunderstanding, it's the above for loop that can be rewrote with contains_if

here is a raw example (I've not compiled it)

```
 const auto not_valid = contains_if(osm_rule.osm_tag_filters, [&](const auto& osm_rule_tag) {
            const auto it_poi_tag = tags.find(osm_rule_tag.first);
            return it_poi_tag == tags.end() || it_poi_tag->second != osm_rule_tag.second);
});
```

TeXitoi: Ha, then +1

antoine-de:
for the debian 7 non compatibility you need to check with @kinnou02, It will impact the other users (for example @eturck) , we need to announce it clearly (but I'm completely for upgrading gcc + boost minimum required version!)

pbougue:
@antoine-de agreed for the debian 7 part. That's one more reason to delay merge and it's ok for me.

source/ed/osm2ed.cpp
pbougue: Done

source/ed/osm2ed.cpp
```diff
@@ -902,52 +888,62 @@ void PoiHouseNumberVisitor::fill_housenumber(const uint64_t osm_id,
...
-        if(tags.find(tags_type) != tags.end()){
-            ref_tag = tags_type;
-            break;
+/**
+ * the first rule that matches OSM object decides POI-type
+ * match : OSM object has all OSM tags required by rule
+ */
+const RuleOsmTag2PoiType* PoiTypeParams::get_applicable_poi_rule(const CanalTP::Tags& tags) const {
+    for (const auto& osm_rule: rules) {
+        auto poi_misses_a_tag = [&](const std::pair<const std::string, std::string>& osm_rule_tag) {
```
antoine-de: If we bump definitely to debian 7, we can now support c++14 and use generic lambda no ? (implemented in gcc 4.9 and it's the debian jessy's version)

pbougue: It may not be the same impact to migrate only boost (if @eturck already has a custom boost version).
But yes, adding a auto is always nicer than a pair :)

eturck: We are indeed stuck with debian 7 for probably at least more than a year. I did try to use a custom boost version (1.55) in the past when you told us you might switch to this version, and I think it was working pretty well but it was a quick test, we are still on 1.49 currently.
Switching to debian 8 and/or gcc 4.9 would be probably a bit harder :).

pbougue:
Jenkins, test this please

source/ed/osm2ed.cpp
```diff
@@ -970,71 +966,100 @@ void OSMCache::flag_nodes() {
...
-        ("input,i", po::value<std::string>(&input)->required(), "Input file (must be an OSM one)")
+        ("input,i", po::value<std::string>(&input),
+                    "Input file (must be an OSM one). "
+                    "If none provided, will check that json poi-type param (if provided) is correct.")
         ("connection-string", po::value<std::string>(&connection_string)->required(),
              "Database connection parameters: host=localhost user=navitia"
              " dbname=navitia password=navitia")
-        ("poi-type,p", po::value<std::vector<std::string>>(&poi_types)->default_value(default_poi_types, ""),
-                       "a poi_type like amenity:college with optionally a name separated with an equals, "
-                       "ex: amenity:college=école");
```
TeXitoi, line 1019: That'll be problematic to stop supporting the old way. I think the old way should generate a filter to have the same behavior as before.

source/ed/osm2ed.cpp
```diff
@@ -970,71 +966,100 @@ void OSMCache::flag_nodes() {
...
+            LOG4CPLUS_ERROR(logger, std::string("Incorrect json poi-type parameter: unknown problem"));
+            return 1;
+        }
+        LOG4CPLUS_INFO(logger, "json poi-type parameter OK");
+        return 0;
+    }
+
+    if (! vm.count("poi-type")) {
+        json_poi_types = ed::connectors::DEFAULT_JSON_POI_TYPES;
+    }
```
TeXitoi, line 1004: Can't you give a default value for that ?

source/ed/osm2ed.cpp
```diff
@@ -970,71 +966,100 @@ void OSMCache::flag_nodes() {
...
-    for(auto type: poi_types){
-        std::vector<std::string> strs;
-        boost::algorithm::split(strs, type, boost::is_any_of("="));
-        if(strs.size() == 1){
-            result[strs[0]] = strs[0];
-        }else if(strs.size() == 2){
-            result[strs[0]] = strs[1];
-        }else{
-            std::cout << "poi_type: \"" << type << "\" ignored!" << std::endl;
+static ed::connectors::PoiTypeParams parse_json_poi_types(const std::string& json_params){
```
TeXitoi: I'd make that a constructor of `PoiTypeParams`

source/ed/osm2ed.cpp
pbougue: I have to distinguish if it's given or not to know if I have to perform a check or a regular bina.

source/ed/osm2ed.cpp
pbougue: Yep, in my test part, I moved that

source/ed/osm2ed.cpp
pbougue: In discussion

source/ed/connectors/osm_tags_reader.cpp
```diff
@@ -224,4 +226,55 @@ std::bitset<8> parse_way_tags(const std::map<std::string, std::string> & tags){
...
+    pt::ptree poi_root;
+    std::istringstream iss (json_params);
+    pt::read_json(iss, poi_root);
+
+    for (const pt::ptree::value_type& json_poi_type: poi_root.get_child("poi_types")) {
+        std::string poi_type_id = json_poi_type.second.get<std::string>("id");
+        if (poi_types.find(poi_type_id) != poi_types.end()) {
+            throw std::invalid_argument("POI type id " + poi_type_id + " is defined multiple times");
+        }
+        poi_types[poi_type_id] = json_poi_type.second.get<std::string>("name");
```
xlqian:
```c++
auto it = poi_types.find(poi_type_id);
if (it != poi_types.end()){
    throw std::invalid_argument("POI type id " + poi_type_id + " is defined multiple times");
}
it->second = json_poi_type.second.get<std::string>("name");
```
(codes not tested)

source/ed/connectors/osm_tags_reader.cpp
```diff
@@ -224,4 +226,55 @@ std::bitset<8> parse_way_tags(const std::map<std::string, std::string> & tags){
...
+        throw std::invalid_argument("The 2 POI types id=amenity:parking and "
+                                    "id=amenity:bicycle_rental must be defined");
+    }
+
+    for (const pt::ptree::value_type& json_rule: poi_root.get_child("rules")) {
+        rules.emplace_back();
+        auto& rule = rules.back();
+        rule.poi_type_id = json_rule.second.get<std::string>("poi_type_id");
+        if (poi_types.find(rule.poi_type_id) == poi_types.end()) {
+            throw std::invalid_argument("Using an undefined POI type id (" + rule.poi_type_id + ") in rules");
```
xlqian, line 254: minor: you may want to throw the exception **before** you create a new element in `rules`

source/ed/connectors/osm_tags_reader.cpp
```diff
@@ -224,4 +226,55 @@ std::bitset<8> parse_way_tags(const std::map<std::string, std::string> & tags){
...
+    }
+}
+
+/**
+ * the first rule that matches OSM object decides POI-type
+ * match : OSM object has all OSM tags required by rule
+ */
+const RuleOsmTag2PoiType* PoiTypeParams::get_applicable_poi_rule(const CanalTP::Tags& tags) const {
+    for (const auto& osm_rule: rules) {
+        auto poi_misses_a_tag = [&](const std::pair<const std::string, std::string>& osm_rule_tag) {
```
xlqian: minor: lambda can be moved out of the loop :)

source/ed/connectors/osm_tags_reader.h
```diff
@@ -40,7 +42,24 @@ constexpr uint8_t CAR_BWD = 3;
 constexpr uint8_t FOOT_FWD = 4;
 constexpr uint8_t FOOT_BWD = 5;
 
-/// Retourne les propriétés quant aux modes et sens qui peuvent utiliser le way
+/// return properties on modes and directions that are possible on a way
```
xlqian, line 45: 👍

source/ed/osm2ed.cpp
```diff
@@ -902,52 +885,45 @@ void PoiHouseNumberVisitor::fill_housenumber(const uint64_t osm_id,
...
-                poi.address_name = st.second;
-            }
-        }else{
-            poi.properties[st.first] = st.second;
+    //properties and other attributes
+    for(auto poi_tag: tags){
+        if(poi_tag.first == "addr:housenumber") {
+            poi.address_number = poi_tag.second;
+        }
+        if(poi_tag.first == "addr:street") {
```
xlqian: `else if`?

source/ed/osm2ed.h
```diff
@@ -381,46 +384,26 @@ inline std::string to_string(OsmObjectType t) {
...
-    PoiHouseNumberVisitor(EdPersistor& persistor, /*const*/ OSMCache& cache,
-            Georef& data, const bool parse_pois, const std::map<std::string, std::string>& poi_types) :
-        persistor(persistor), cache(cache), data(data), parse_pois(parse_pois)  {
+    PoiHouseNumberVisitor(EdPersistor& persistor, /*const*/ OSMCache& cache, Georef& data,
+                          const bool parse_pois, const PoiTypeParams& poi_params) :
+                              persistor(persistor), cache(cache), data(data),
+                              parse_pois(parse_pois), poi_params(poi_params) {
         uint32_t idx = 0;
-        for(auto type: poi_types){
+        for(auto type: poi_params.poi_types){
```
xlqian: const auto& type

source/ed/tests/osm_tags_reader_test.cpp
```diff
@@ -0,0 +1,259 @@
...
+            {
+              "osm_tags_filters": [
+                {"key": "amenity", "value": "bicycle_rental"}
+              ],
+              "poi_type_id": "amenity:bicycle_rental"
+            },
+            {
+              "osm_tags_filters": [
+                {"key": "amenity", "value": "bicycle_rental"},
+                {"key": "bbs_type", "value": "velib"}
```
xlqian: ~~bbs~~? bss?

source/ed/connectors/osm_tags_reader.cpp
pbougue: There is a `try` around the call to that method to output the problem when just checking that json is OK. Not around the regular osm2ed call, because crashing seems OK to me in that case...
(if that was the question)

source/ed/connectors/osm_tags_reader.cpp
pbougue: see previous reply, but I guess it's not really worrisome

source/ed/connectors/osm_tags_reader.cpp
pbougue: that was not the question :)

pbougue: We want to create it if it doesn't exist, find doesn't create a new element if missing.

TeXitoi:
```c++
const auto& poi_type_id = json_poi_type.second.get<std::string>("id");
const auto& poi_type_name = json_poi_type.second.get<std::string>("name");
if (! poi_types.insert(poi_type_id, poi_type_name).second) {
    throw std::invalid_argument("POI type id " + poi_type_id + " is defined multiple times");
}
```

source/ed/connectors/osm_tags_reader.cpp
TeXitoi: That the same, the constructor will abort, and everything will be cleaned.

source/ed/connectors/osm_tags_reader.h
```diff
@@ -40,7 +42,24 @@ constexpr uint8_t CAR_BWD = 3;
...
+                        std::unordered_map<std::string, ed::types::PoiType>& poi_types);
+
+struct RuleOsmTag2PoiType {
+    std::string poi_type_id;
+    std::map<std::string, std::string> osm_tag_filters;
+};
+
+struct PoiTypeParams {
+    std::map<std::string, std::string> poi_types;
+    std::vector<RuleOsmTag2PoiType> rules;
```
TeXitoi, line 59: Maybe these should be private.

source/ed/osm2ed.cpp
```diff
@@ -902,52 +885,45 @@ void PoiHouseNumberVisitor::fill_housenumber(const uint64_t osm_id,
...
+    // we have found a POI and we know it's type
     ed::types::Poi poi;
-    poi.poi_type = it->second;
+    //type
+    poi.poi_type = data.poi_types[applicable_rule->poi_type_id];
+    //name
     if(tags.find("name") != tags.end()){
         poi.name = tags.at("name");
+    } else {
+        poi.name = poi.poi_type->name;
```
TeXitoi, line 911: :+1:

source/ed/osm2ed.cpp
TeXitoi: OK

source/navitiacommon/navitiacommon/launch_exec.py
```diff
@@ -105,11 +106,16 @@ def launch_exec(exec_name, args, logger):
...
                 break
     finally:
         proc.stdout.close()
         proc.stderr.close()
 
-    return proc.returncode
+    return proc.returncode, traces
+
+def launch_exec(exec_name, args, logger):
+    code, traces = launch_exec_traces(exec_name, args, logger)
```
datanel: you don't use traces

TeXitoi (changes requested):

source/ed/osm2ed.cpp
pbougue: So we decided it's ok to have a downtime for few non-standard poi-types at deploy.
Then we let it as is.

source/ed/connectors/osm_tags_reader.h
pbougue: it's used in PoiHouseNumberVisitor and as the philosophy is no getter... ;)

TeXitoi (approved):

PR #1844, opened 2016-11-10
Jormun: Async experimental scenario

azime:
Support async call kraken in experimental scenario

### DIFF
--- a/source/jormungandr/jormungandr/scenarios/experimental.py
+++ b/source/jormungandr/jormungandr/scenarios/experimental.py
@@ -100,12 +100,12 @@ def create_parameters(request):
 
 def _update_crowfly_duration(instance, mode, stop_area_uri):
     stop_points = instance.georef.get_stop_points_for_stop_area(stop_area_uri)
-    crowfly = set()
+    crowfly_sps = set()
     fallback_list = collections.defaultdict(dict)
     for stop_point in stop_points:
         fallback_list[mode][stop_point.uri] = 0
-        crowfly.add(stop_point.uri)
-    return crowfly, fallback_list
+        crowfly_sps.add(stop_point.uri)
+    return crowfly_sps, fallback_list
 
 
 def _rename_journey_sections_ids(start_idx, sections):
@@ -166,50 +166,12 @@ def _sn_routing_matrix(instance, place, places_crowfly, mode, max_duration, requ
                                                                                            request,
                                                                                            **kwargs)
     if not sn_routing_matrix.rows[0].duration:
-            return {mode: {}}
+        return {mode: {}}
     import numpy as np
     durations = np.array(sn_routing_matrix.rows[0].duration)
     valid_duration_idx = np.argwhere((durations > -1) & (durations < max_duration)).flatten()
     return {mode: dict(zip([places[i].uri for i in valid_duration_idx],
                        durations[(durations > -1) & (durations < max_duration)].flatten()))}
-
-
-def worker_routing_matrix(instance, dep_arr, place, places_crowfly, mode, max_duration, request, **kwargs):
-    return mode, dep_arr, _sn_routing_matrix(instance, place, places_crowfly, mode, max_duration, request, **kwargs)
-
-
-def worker_crowfly(instance, place, dep_arr, mode, request, max_nb_crowfly=5000, **kwargs):
-    return (mode, dep_arr, _get_places_crowfly(instance, mode, place, get_max_fallback_duration(request, mode),
-                                               max_nb_crowfly, **kwargs))
-
-
-def worker_update_crowfly_duration(instance, dep_arr, mode, place):
-    return mode, dep_arr, _update_crowfly_duration(instance,  mode, place)
-
-
-def worker_direct_path(func, fallback_direct_path, instance, mode, origin, destination, datetime,
-                       clockwise, request):
-    return (mode, func(instance, mode, origin, destination, datetime, clockwise, request,
-                       fallback_direct_path))
-
-
-def worker_journey(instance, dep_mode, arr_mode, pt_object_origin, pt_object_destination, request,
-                 fallback_direct_path, fall_back_origins, fall_back_destinations, journey_parameters,
-                 reverse_sections=False):
-    dp_key = (dep_mode, pt_object_origin.uri, pt_object_destination.uri, request['datetime'],
-              request['clockwise'], reverse_sections)
-    dp = fallback_direct_path.get(dp_key)
-    if dp.journeys:
-        journey_parameters.direct_path_duration = dp.journeys[0].durations.total
-    else:
-        journey_parameters.direct_path_duration = None
-    origins = fall_back_origins.get(dep_mode)
-    destinations = fall_back_destinations.get(arr_mode)
-    if not fall_back_origins or not fall_back_destinations or not request.get('max_duration', 0):
-        return (dep_mode, arr_mode, None)
-    return (dep_mode, arr_mode, instance.planner.journeys(origins, destinations,
-                                                          request['datetime'], request['clockwise'],
-                                                          journey_parameters))
 
 
 def worker_from(journey, instance, fisrt_section, func_extend_journey, _from, crow_fly_stop_points, request, dep_mode,
@@ -223,7 +185,7 @@ def worker_from(journey, instance, fisrt_section, func_extend_journey, _from, cr
                                      journey.sections[0].begin_date_time)])
         else:
             func_extend_journey(instance, journey, dep_mode, _from, departure, journey.departure_date_time,
-                                 origins_fallback, True, request, fallback_direct_path)
+                                origins_fallback, True, request, fallback_direct_path)
     return journey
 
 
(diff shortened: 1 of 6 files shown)


source/jormungandr/jormungandr/scenarios/experimental.py
```diff
@@ -131,16 +134,240 @@ def _reverse_journeys(res):
...
+        fallback[mode] = {}
+        return
+    import numpy as np
+    durations = np.array(sn_routing_matrix.rows[0].duration)
+    valid_duration_idx = np.argwhere((durations > -1) & (durations < max_duration)).flatten()
+    fallback[mode] = dict(zip([places[i].uri for i in valid_duration_idx],
+                    durations[(durations > -1) & (durations < max_duration)].flatten()))
+
+
+import gevent, gevent.pool
```
kinnou02: I like import to be on the top of the file.

source/jormungandr/jormungandr/scenarios/experimental.py
```diff
@@ -131,16 +134,240 @@ def _reverse_journeys(res):
...
+    dp_key = (dep_mode, pt_object_origin.uri, pt_object_destination.uri, request['datetime'],
+              request['clockwise'], reverse_sections)
+    dp = fallback_direct_path.get(dp_key)
+    if dp.journeys:
+        journey_parameters.direct_path_duration = dp.journeys[0].durations.total
+    else:
+        journey_parameters.direct_path_duration = None
+    origins = fall_back_origins.get(dep_mode)
+    destinations = fall_back_destinations.get(arr_mode)
+    if not fall_back_origins or not fall_back_destinations or not request.get('max_duration', 0):
```
kinnou02: Almost everywhere we use fallback in one word not two.

source/jormungandr/jormungandr/scenarios/experimental.py
```diff
@@ -131,16 +134,240 @@ def _reverse_journeys(res):
...
+        return futures
+
+
 class Scenario(new_default.Scenario):
 
     def __init__(self):
         super(Scenario, self).__init__()
 
-    def _get_direct_path(self, instance, mode, pt_object_origin, pt_object_destination, datetime, clockwise, request, reverse_sections=False):
+    def _get_direct_path(self, instance, mode, pt_object_origin, pt_object_destination, datetime, clockwise,
```
kinnou02: I am a little confuse about all of this, so this function act as a cache to not compute the same direct multiple time right?

source/jormungandr/jormungandr/scenarios/experimental.py
```diff
@@ -249,85 +408,67 @@ def call_kraken(self, request_type, request, instance, krakens_call):
...
-                                                                         reverse=True)
-
-               #Fetch all the stop points of this stop_area and replaces all the durations by 0 in the table
-               #g.destinations_fallback[arr_mode]
-               _update_crowfly_duration(instance, g.destinations_fallback, arr_mode, request['destination'],
-                                        crow_fly_stop_points)
+        worker = Worker(instance, krakens_call, request)
+
+        if request.get('max_duration', 0):
+            futures = worker._get_crowfly_futures(g.requested_origin,
```
kinnou02: This seems to be mostly based on side effect, I don't like side effect and async, it's complex and error prone

source/jormungandr/jormungandr/scenarios/experimental.py
```diff
@@ -249,85 +408,67 @@ def call_kraken(self, request_type, request, instance, krakens_call):
...
+                                            g.requested_destination, g.origins_places_crowfly,
+                                            g.destinations_places_crowfly, 5000)
+            gevent.joinall(futures)
+
+            futures = worker._get_routing_matrix_fustures(g.requested_origin, g.requested_destination,
```
kinnou02: There is a lot of abstraction level and absolute not explanation about what happen, why at that moment.
I have count at least seven steps, why do we do one before the other?

kinnou02 (changes requested):

source/jormungandr/jormungandr/scenarios/experimental.py
```diff
@@ -249,90 +396,88 @@ def call_kraken(self, request_type, request, instance, krakens_call):
...
+                mode, dep_arr, resp = future.get()
+                if dep_arr == "dep":
+                    g.origins_fallback.update(resp)
+                else:
+                    g.destinations_fallback.update(resp)
+
+            futures = worker.get_duration_futures()
+            for future in gevent.iwait(futures):
+                mode, dep_arr, resp = future.get()
+                crow_fly_stop_points = set(list(crow_fly_stop_points) + resp[0])
```
xlqian: `crow_fly_stop_points.add(resp[0])` ?

xlqian:
Jenkins, test this please!

source/jormungandr/jormungandr/scenarios/experimental.py
```diff
@@ -29,16 +29,20 @@
...
-from jormungandr.utils import get_uri_pt_object
+from jormungandr.utils import get_uri_pt_object, generate_id
 from jormungandr import app
-
+import gevent
+import gevent.pool
+import collections
+# http://www.gevent.org/intro.html#monkey-patching
+import gevent.monkey
+gevent.monkey.patch_socket()
```
kinnou02: We should do the monkey patching in the init of jormun, else importing the scenario would have induce a side effect

source/jormungandr/jormungandr/scenarios/experimental.py
```diff
@@ -244,90 +400,89 @@ def call_kraken(self, request_type, request, instance, krakens_call):
...
-               #Fetch all the stop points of this stop_area and replaces all the durations by 0 in the table
-               #g.destinations_fallback[arr_mode]
-               _update_crowfly_duration(instance, g.destinations_fallback, arr_mode, request['destination'],
-                                        crow_fly_stop_points)
+        worker = AsyncWorker(instance, krakens_call, request)
+
+        if request.get('max_duration', 0):
+            orig_futures, dest_futures = worker.get_crowfly_futures(g.requested_origin, g.requested_destination)
+            gevent.joinall(orig_futures + dest_futures)
+            for future in gevent.iwait(orig_futures):
```
kinnou02: You don't have to do a `joinall` if you use `iwait`

source/jormungandr/jormungandr/scenarios/experimental.py
```diff
@@ -131,16 +139,232 @@ def _reverse_journeys(res):
...
+            called_modes |= {(arr_mode, destination.uri), (dep_mode, origin.uri)}
+        return origin_futures, destination_futures
+
+    def get_crowfly_futures(self, origin, destination):
+        origin_futures = []
+        destination_futures = []
+
+        called_modes = set()
+        for dep_mode, arr_mode in self.krakens_call:
+            if (dep_mode, origin.uri) not in called_modes:
```
kinnou02: We don't need the `origin.uri` here no?

source/jormungandr/jormungandr/scenarios/experimental.py
```diff
@@ -244,90 +400,89 @@ def call_kraken(self, request_type, request, instance, krakens_call):
...
-                                                                         reverse=True)
-
-               #Fetch all the stop points of this stop_area and replaces all the durations by 0 in the table
-               #g.destinations_fallback[arr_mode]
-               _update_crowfly_duration(instance, g.destinations_fallback, arr_mode, request['destination'],
-                                        crow_fly_stop_points)
+        worker = AsyncWorker(instance, krakens_call, request)
+
+        if request.get('max_duration', 0):
+            orig_futures, dest_futures = worker.get_crowfly_futures(g.requested_origin, g.requested_destination)
```
kinnou02, line 501: comment to add:
First we search all stop_point reachable for each modes from the origin and the destination

source/jormungandr/jormungandr/scenarios/experimental.py
```diff
@@ -244,90 +400,89 @@ def call_kraken(self, request_type, request, instance, krakens_call):
...
+            gevent.joinall(orig_futures + dest_futures)
+            for future in gevent.iwait(orig_futures):
+                g.origins_places_crowfly.update(future.get())
+            for future in gevent.iwait(dest_futures):
+                g.destinations_places_crowfly.update(future.get())
+
+            orig_futures, dest_futures = worker.get_routing_matrix_futures(g.requested_origin,
```
kinnou02, line 516: Comment to add:
Then we want to put a arrival time for each of this stop points.
As a side note this won't work the day when our ETA will be impacted by the datetime of the journey, at least for the arrival when doing a "departure after" request.

source/jormungandr/jormungandr/scenarios/experimental.py
```diff
@@ -244,90 +400,89 @@ def call_kraken(self, request_type, request, instance, krakens_call):
...
+                                                                           g.requested_destination,
+                                                                           g.origins_places_crowfly,
+                                                                           g.destinations_places_crowfly)
+            gevent.joinall(orig_futures + dest_futures)
+            for future in gevent.iwait(orig_futures):
+                g.origins_fallback.update(future.get())
+            for future in gevent.iwait(dest_futures):
+                g.destinations_fallback.update(future.get())
+
+            orig_futures, dest_futures = worker.get_update_crowfly_duration_futures()
```
kinnou02, line 528: If I understand correctly we update the stop_points of the requested stop_area to set the ETA at 0, if so this call to kraken can have be made earlier.

source/jormungandr/jormungandr/scenarios/experimental.py
```diff
@@ -244,90 +400,89 @@ def call_kraken(self, request_type, request, instance, krakens_call):
...
-                if direct_path not in resp:
-                    direct_path.journeys[0].internal_id = "dp_{resp}-{j}".format(resp=self.nb_kraken_calls, j=0)
-                    resp.append(direct_path)
-            else:
-                journey_parameters.direct_path_duration = None
 
-            fall_back_origins = g.origins_fallback.get(dep_mode)
-            fall_back_destinations = g.destinations_fallback.get(arr_mode)
-            if not fall_back_origins or not fall_back_destinations or not request.get('max_duration', 0):
+        futures = worker.get_direct_path_futures(self._get_direct_path,
```
kinnou02: Comment to add:
we compute the direct path for this journey, their time will be used to initialized our PT calls

source/jormungandr/jormungandr/scenarios/experimental.py
```diff
@@ -244,90 +400,89 @@ def call_kraken(self, request_type, request, instance, krakens_call):
...
+                                                 g.requested_destination)
+        for future in gevent.iwait(futures):
+            self.nb_kraken_calls += 1
+            dep_mode, resp_direct_path = future.get()
+            if resp_direct_path.journeys:
+                if resp_direct_path not in resp:
+                    resp_direct_path.journeys[0].internal_id = str(generate_id())
+                    resp.append(resp_direct_path)
+
+        futures = worker.get_journey_futures(g.requested_origin, g.requested_destination,
```
kinnou02: Comment to add:
We launch the PT journey planning

This calls could be launched earlier, as soon as we have the direct_path duration for the requested mode

source/jormungandr/jormungandr/scenarios/experimental.py
```diff
@@ -244,90 +400,89 @@ def call_kraken(self, request_type, request, instance, krakens_call):
...
                 continue
-
-            local_resp = instance.planner.journeys(fall_back_origins,
-                                                   fall_back_destinations,
-                                                   request['datetime'],
-                                                   request['clockwise'],
-                                                   journey_parameters)
+            dp_key = (dep_mode, g.requested_origin.uri, g.requested_destination.uri, request['datetime'],
+                      request['clockwise'], False)
+            direct_path = g.fallback_direct_path.get(dp_key)
```
kinnou02, line 573: I didn't see where we compute the fallback direct_path

source/jormungandr/jormungandr/scenarios/experimental.py
```diff
@@ -244,90 +400,89 @@ def call_kraken(self, request_type, request, instance, krakens_call):
...
-                                    request)
-
+            for j in local_resp.journeys:
+                j.internal_id = str(generate_id())
+                map_response.append((dep_mode, arr_mode, j))
             resp.append(local_resp)
-            logger.debug("for mode %s|%s we have found %s journeys", dep_mode, arr_mode, len(local_resp.journeys))
-            logger.debug("for mode %s|%s we have found %s direct path", dep_mode, arr_mode, len(direct_path.journeys))
+
+        futures = worker.build_journeys(self._extend_journey, map_response, crow_fly_stop_points)
```
kinnou02: Comment to add:
We compute the fallback part of our PT journeys

source/jormungandr/jormungandr/scenarios/experimental.py
```diff
@@ -244,90 +400,89 @@ def call_kraken(self, request_type, request, instance, krakens_call):
...
+        for future in gevent.iwait(futures):
```
kinnou02, line 592: Why not do that direclty in `build_journey`?

kinnou02 (changes requested):

source/jormungandr/jormungandr/scenarios/experimental.py
xlqian: done

source/jormungandr/jormungandr/scenarios/experimental.py
xlqian: I have simplified that logic :)

source/jormungandr/jormungandr/scenarios/experimental.py
xlqian: Yep, I'm adding a lot of comments :)

source/jormungandr/jormungandr/scenarios/experimental.py
xlqian: I imagined a scenario as followed: 

We need to compute some crowfly stop_points with "walking" and "bike" respectively for "from" and "to", in this case we'll get 4 requests for stop_points.

**WITH** `joinall`: all requests are launched at the same time thanks to the concatenation
**WITHOUT** `joinall`: only 2 requests are parallelized 

That's why i did with `joinall`

source/jormungandr/jormungandr/scenarios/experimental.py
xlqian: Hmm, this functions update the sn_routing_matrix if the stop_points is contained by stop_area, so we can only do this after the computation of sn_routing_matrix

source/jormungandr/jormungandr/scenarios/experimental.py
xlqian: `This calls could be launched earlier, as so......`
Nope, we need the `global` direct_path duration to launch this pt computation

source/jormungandr/jormungandr/scenarios/experimental.py
xlqian: my bad, i'm improving this

source/jormungandr/jormungandr/scenarios/experimental.py
xlqian: I'm improving this (which is not that easy)

source/jormungandr/jormungandr/scenarios/experimental.py
xlqian: I removed the iwait :)

source/jormungandr/jormungandr/scenarios/experimental.py
```diff
@@ -265,71 +477,98 @@ def call_kraken(self, request_type, request, instance, krakens_call):
...
+        # Now we compute the direct path with all requested departure mode
+        # their time will be used to initialized our PT calls
+        futures = worker.get_direct_path_futures(g.fallback_direct_path,
+                                                 g.requested_origin,
+                                                 g.requested_destination,
+                                                 request['datetime'],
+                                                 request['clockwise'],
+                                                 False)
+        for future in gevent.iwait(futures):
+            self.nb_kraken_calls += 1
```
azime: to delete ?

xlqian:
I benched on `new_default` w/wo `gevet.monkey.patch_socket()` There is no apparent regression in performance.

kinnou02 (approved):

PR #1845, opened 2016-11-15
kraken: fix a test on georef_test

kinnou02:
A `georef.init()` was missing since the change from https://github.com/CanalTP/navitia/commit/70e3ea351c78b9684ea2575d4fa42bd1cde9758a

### DIFF
--- a/source/georef/georef.h
+++ b/source/georef/georef.h
@@ -257,7 +257,7 @@ struct GeoRef {
     flat_enum_map<nt::Mode_e, nt::idx_t> offsets;
 
     /// number of vertex by transportation mode
-    nt::idx_t nb_vertex_by_mode;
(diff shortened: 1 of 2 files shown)


pbougue (commented):
No georef init in those tests (maybe it's on purpose but we change the graph):
L779
L880
L969

And maybe more... That being said we can just wait for the next test break and improve then.

kinnou02:
These tests doesn't use graphbuilder or georef object, so we don't build it obviously :)

TeXitoi:
maybe assigning that to 0 will help be shure everything is OK:

https://github.com/CanalTP/navitia/blob/dev/source/georef/georef.h#L260

TeXitoi (approved):

PR #1849, opened 2016-11-16
All: no feedpublisher creation for georef.get_crow_fly (experimental)

TeXitoi:
- [x] merge CanalTP/navitia-proto#78 (I can't do it, YOU can!)
- [x] update with the merged commit
- [x] validate https://github.com/CanalTP/navitia/pull/1849/files#diff-05cbd102661a0da9e75756727f9715bfR86 or no test or do something else (what?)

### DIFF
--- a/source/jormungandr/jormungandr/georef.py
+++ b/source/jormungandr/jormungandr/georef.py
@@ -79,9 +79,12 @@ class Kraken(object):
         req.places_nearby.depth = 1
         req.places_nearby.count = max_nb_crowfly
         req.places_nearby.start_page = 0
+        req.disable_feedpublisher = True
         # we are only interested in public transports
         req.places_nearby.types.append(type_pb2.STOP_POINT)
-        return self.instance.send_and_receive(req).places_nearby
+        res = self.instance.send_and_receive(req)
+        assert len(res.feed_publishers) == 0 # didn't find any better way to test that...
+        return res.places_nearby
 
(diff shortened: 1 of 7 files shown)


source/jormungandr/jormungandr/georef.py
```diff
@@ -79,9 +79,12 @@ def get_crow_fly(self, origin, streetnetwork_mode, max_duration, max_nb_crowfly)
         req.places_nearby.depth = 1
         req.places_nearby.count = max_nb_crowfly
         req.places_nearby.start_page = 0
+        req.disable_feedpublisher = True
         # we are only interested in public transports
         req.places_nearby.types.append(type_pb2.STOP_POINT)
-        return self.instance.send_and_receive(req).places_nearby
+        res = self.instance.send_and_receive(req)
+        assert len(res.feed_publishers) == 0 # didn't find any better way to test that...
```
kinnou02: this will fail in production too,that's not nice :(
But I don't see an easy way to test this method

xlqian: indeed, difficult to test a local var of a function... maybe log it?

TeXitoi: So? what should I do?

TeXitoi:
I've replaced the assert by an error log.

xlqian (approved):

PR #1854, opened 2016-11-18
jormun: fix on /v1/status

kinnou02:
`is_free` and `realtime_proxies` were always set to null on /v1/status,
they were filled only on /v1/coverage/foo/status.
This PR fix that by filling them on both.


TeXitoi (approved):

pbougue (commented):
Not a must: add tests (and maybe also on coverage one if there are none)

xlqian (approved):

PR #1860, opened 2016-11-23
kraken: checking for zonal stop_points added

kadhikari:
journey arrival_time filled with last section arrival_time value.


kadhikari:
Ticket concerned: http://jira.canaltp.fr/browse/ITI-284

TeXitoi (approved):

PR #1866, opened 2016-11-28
Kraken: fix disruption update on lines

xlqian:
http://jira.canaltp.fr/browse/NAVITIAII-2170

### DIFF
--- a/source/kraken/apply_disruption.cpp
+++ b/source/kraken/apply_disruption.cpp
@@ -523,11 +523,13 @@ void apply_impact(boost::shared_ptr<nt::disruption::Impact> impact,
 using impact_wptr = boost::weak_ptr<nt::disruption::Impact>;
 
 static auto comp = [](const impact_wptr& lhs, const impact_wptr& rhs){
-    if(lhs.expired() || rhs.expired()) {
-        return false;
-    }
     auto i_lhs = lhs.lock();
     auto i_rhs = rhs.lock();
+
+    if(!i_lhs || !i_rhs) {
+       return false;
+    }
+
     return i_lhs->updated_at < i_rhs->updated_at || i_lhs->uri < i_rhs->uri;
 };
 
--- a/source/kraken/apply_disruption.cpp
+++ b/source/kraken/apply_disruption.cpp
@@ -522,7 +522,7 @@ void apply_impact(boost::shared_ptr<nt::disruption::Impact> impact,
 
 using impact_wptr = boost::weak_ptr<nt::disruption::Impact>;
 
-auto comp = [](const impact_wptr& lhs, const impact_wptr& rhs){
+static auto comp = [](const impact_wptr& lhs, const impact_wptr& rhs){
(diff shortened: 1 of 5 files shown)


pbougue (commented):
not sure utils is on the right commit

pbougue:
pay attention to utils, not sure it's on the right commit

xlqian:
@pbougue Fixed! thx!

xlqian:
Jenkins, test this please!

TeXitoi (approved):

TeXitoi:
@kinnou02 can you look at it?

source/kraken/apply_disruption.cpp
```diff
@@ -520,15 +520,31 @@ void apply_impact(boost::shared_ptr<nt::disruption::Impact> impact,
...
 }
 
+using impact_wptr = boost::weak_ptr<nt::disruption::Impact>;
+
+static auto comp = [](const impact_wptr& lhs, const impact_wptr& rhs){
+    if(lhs.expired() || rhs.expired()) {
+        return false;
+    }
+    auto i_lhs = lhs.lock();
+    auto i_rhs = rhs.lock();
```
kinnou02, line 527: `lock()` can return an empty/invalid pointer, it will be safer to check the resulting `shared_ptr`

source/kraken/apply_disruption.cpp
```diff
@@ -566,9 +582,7 @@ struct delete_impacts_visitor : public apply_impacts_visitor {
         decltype(mvj->impacted_by) impacted_by_moved;
         boost::swap(impacted_by_moved, mvj->impacted_by);
         for (auto i: impacted_by_moved) {
-            if (auto spt = i.lock()) {
-                apply_disruption(*spt->disruption, pt_data, meta);
-            }
+            disruptions_collection.insert(i);
```
kinnou02: You could do:
```
disruptions_collection.insert(begin(impacted_by_moved), end(impacted_by_moved));
```

kinnou02 (approved):
My understanding is that before we were resetting a vj then reapplying all the impact on it. 
WIth this PR we will reset all vjs then reapply the disruptions, it will prevent doing the same job multiple times.

Am I right?

source/kraken/apply_disruption.cpp
xlqian: I checked that at L526... Isn't that enough? (Hopefully, if weak_ptr is not expired, the resulting shared_ptr is not null neither)

source/kraken/apply_disruption.cpp
xlqian: Yep!

xlqian:
@kinnou02 Yes! you're right!

source/kraken/apply_disruption.cpp
kinnou02: It can expire between the check, we are in a multi threaded context

PR #1883, opened 2016-12-02
jormun: add is_open_service in /status

### LINKED ISSUE #1855: Add new coverage property
There is a "is_open_data" property on /coverage/{region_id}/status which is usefull.
But it would be great to have an "is_open_service" for Regions who give their data to navitia.io but not in real open_data.

kinnou02:
This PR close #1855. `is_free` was used to indicate that a coverage was using opendata, this isn't always true as a coverage can be used by anyone but not be based on opendata. This will be the case of `fr-bre` in a few weeks.

So we have now two flag:
   - `is_open_service`: this indicate a coverage usable by anyboby (`is_free` in database)
   - `is_open_data`: the coverage is using opendata only

By default is_open_data will be initialized with the value of is_free, this value can then be updated by the instances api of tyr.

This PR also add the management of `is_free` with the tyr API.

I don't like the name `is_open_service` so if any of you has a better idea I will be glad to change it.


antoine-de (approved):

pbougue:
`is_fully_open_data` or `is_open_data_only` ?

TeXitoi (approved):

prhod (approved):
ok for `is_open_service` for me

PR #1886, opened 2016-12-06
Jormun: Admin to admin in experimental

antoine-de:
implement the admin to admin journeys in experimental.

we can start(finish) from(to) either 
 * the stop points of the admin main stop areas at a cost of 0 
 * or we can walk from the center of the admin

A crow fly section is created from the admin to the stop points

linked to http://jira.canaltp.fr/browse/ITI-279

The tests will be done in artemis

TODO:
- [x] wait until some artemis tests are done locally

### DIFF
--- a/source/jormungandr/jormungandr/scenarios/experimental.py
+++ b/source/jormungandr/jormungandr/scenarios/experimental.py
@@ -48,7 +48,7 @@ def create_crowfly(_from, to, begin, end, mode='walking'):
     section.type = response_pb2.CROW_FLY
     section.origin.CopyFrom(_from)
     section.destination.CopyFrom(to)
-    section.duration = end-begin
+    section.duration = end - begin
     section.begin_date_time = begin
     section.end_date_time = end
     section.street_network.mode = response_pb2.Walking
@@ -98,8 +98,25 @@ def create_parameters(request):
                              forbidden_uris=request['forbidden_uris[]'])
 
 
-def _update_crowfly_duration(instance, mode, stop_area_uri, requested_point):
-    stop_points = instance.georef.get_stop_points_for_stop_area(stop_area_uri)
+def _update_crowfly_duration(instance, mode, requested_point):
+    """
+    compute the list of stoppoint that can be accessed for free by crowfly
+    for a stoparea it's the list of it's stoppoints
+    for an admin it's the list of it's main_stop_area's stoppoints
+
+    we also look if there are on demand transport (odt) stoppoints in the area
+
+    :return
+        the list of the freely accessed stoppoints
+        the list of odt stoppoints
+        the fallback duration for all those stoppoints
+    """
+    stop_points = []
+    if requested_point.embedded_type == type_pb2.STOP_AREA:
+        stop_points = instance.georef.get_stop_points_for_stop_area(requested_point.uri)
+    elif requested_point.embedded_type == type_pb2.ADMINISTRATIVE_REGION:
+        stop_points = [sp for sa in requested_point.administrative_region.main_stop_areas
+                       for sp in sa.stop_points]
     crowfly_sps = set()
     odt_stop_points = set()
     fallback_list = collections.defaultdict(dict)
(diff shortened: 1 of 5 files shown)


source/jormungandr/jormungandr/scenarios/experimental.py
```diff
@@ -30,6 +30,7 @@
 from __future__ import absolute_import, print_function, unicode_literals, division
 import logging
 from jormungandr.scenarios import new_default
+from jormungandr.scenarios.default import is_admin
```
TeXitoi: useless?

source/type/pb_converter.cpp
```diff
@@ -429,6 +429,14 @@ void PbCreator::Filler::fill_pb_object(const ng::Admin* adm, pbnavitia::Administ
...
         admin->set_insee(adm->insee);
     }
+    for (const auto& sa: adm->main_stop_areas) {
+        auto* pb_sa = admin->add_main_stop_areas();
+        fill_pb_object(sa, pb_sa);
+        for (const auto& sp: sa->stop_point_list) {
+            auto* pb_sp = pb_sa->add_stop_points();
+            fill_pb_object(sp, pb_sp);
+        }
+    }
```
TeXitoi, line 443: It can be great if we limit the size of that to the bare minimum. depth=0 for the 2 I think.

TeXitoi (commented):
Great!

source/type/pb_converter.cpp
antoine-de: yes you're right!

source/jormungandr/jormungandr/scenarios/experimental.py
```diff
@@ -97,8 +97,25 @@ def create_parameters(request):
                              forbidden_uris=request['forbidden_uris[]'])
 
 
-def _update_crowfly_duration(instance, mode, stop_area_uri, requested_point):
-    stop_points = instance.georef.get_stop_points_for_stop_area(stop_area_uri)
+def _update_crowfly_duration(instance, mode, requested_point):
```
pbougue: `requested_place` maybe

source/jormungandr/jormungandr/scenarios/experimental.py
```diff
@@ -118,7 +135,7 @@ def _update_crowfly_duration(instance, mode, stop_area_uri, requested_point):
         for stop_point in odt_sps:
             fallback_list[mode][stop_point.uri] = 0
             odt_stop_points.add(stop_point.uri)
-    
+
     return crowfly_sps, odt_stop_points, fallback_list
```
pbougue, line 139: @TeXitoi this is always free, so I see no point in `fallback_list` here

pbougue (approved):
Adding tests would be nice, but I can't imagine interesting ones

source/jormungandr/jormungandr/scenarios/experimental.py
TeXitoi: I don't understand your remark.

source/jormungandr/jormungandr/scenarios/experimental.py
TeXitoi: I suspect that's `requested_(entry_)point`

source/jormungandr/jormungandr/scenarios/experimental.py
pbougue: The time is always 0 in fallback

TeXitoi: https://github.com/CanalTP/navitia/blob/dev/source/jormungandr/jormungandr/scenarios/experimental.py#L502

That's the same format as `g.*_places_crowfly`

pbougue: ok, that's fine as it is

antoine-de:
artemis tests done locally

TeXitoi (approved):

PR #1887, opened 2016-12-07
tyr: add a command that return the last imported dataset

kinnou02:
this will be used during upgrade to check if they are still available

### DIFF
--- a/source/tyr/tyr/command/last_dataset.py
+++ b/source/tyr/tyr/command/last_dataset.py
@@ -27,11 +27,19 @@
 # https://groups.google.com/d/forum/navitia
 # www.navitia.io
 
-from tyr.command.reload_kraken import ReloadKrakenCommand
-from tyr.command.build_data import BuildDataCommand
-from tyr.command.load_data import LoadDataCommand
-import tyr.command.purge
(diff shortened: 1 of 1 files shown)


source/tyr/tyr/command/last_dataset.py
```diff
@@ -0,0 +1,45 @@
...
+    print the last dataset imported for each family of dataset.
+    This will be the dataset imported when doing a import_last_dataset
+    """
+    instance = models.Instance.query_existing().filter_by(name=instance_name).first()
+
+    if not instance:
+        raise Exception("cannot find instance {}".format(instance_name))
+
+    for d in instance.last_datasets(1):
+        print('{}|{}'.format(d.name, d.family_type))
```
antoine-de, line 45: isn't it easier to do this via API ?

kinnou02: it's for fabric, it aim to replace the sql request that is used to check the dataset. It's easier for fabric to interact with cli than web api.

antoine-de (approved):
not fully convinced (I'd prefer an API), but I understand the will to do a simple thing here

PR #1889, opened 2016-12-08
Jormungandr: fix performance regression caused by gevent monkey patch

xlqian:


source/jormungandr/jormungandr/scenarios/experimental.py
```diff
@@ -39,6 +39,15 @@
...
 import gevent
 import gevent.pool
+import gevent.socket
+import urllib3
+
+# This line replaces the gevent.monkey.patch_socket()
+# the reason why we don't use patch_socket() at the very beginning of jormungandr is
+# that it caused a mysterious strong performance regression for certain requests, thus we patch
+# only at places where asynchronisation is needed
+urllib3.connection.connection.socket = gevent.socket
```
kinnou02: Doing this in experimental will have some strange side effects when this scenario will be imported for the first time. I prefer monkey patch to be done globally, so we have consistent behavior no matter what call have been made previously.

xlqian:
@kinnou02 i moved that to `api.py` and added an option in the settings

source/jormungandr/jormungandr/default_settings.py
```diff
@@ -109,3 +109,5 @@
     
 GRAPHICAL_ISOCHRONE = False
 HEAT_MAP = False
+
+PATCH_WITH_GEVENT_SOCKET = False
```
pbougue: no new line, and the name is a bit confusing, maybe more something with parallel_street_network_calls?

source/jormungandr/jormungandr/api.py
```diff
@@ -41,10 +41,17 @@
...
-# http://www.gevent.org/intro.html#monkey-patching
-#import gevent.monkey
-#gevent.monkey.patch_socket()
-#gevent.monkey.patch_ssl()
+
+if rest_api.app.config.get('PATCH_WITH_GEVENT_SOCKET', False):
+    import gevent.socket
+    import urllib3
+    logger = logging.getLogger('jormungandr.patch_gevent_socket')
+    logger.info("Attention! You'are patching urllib3.connection.connection.socket with gevent.socket")
```
pbougue: Same here, maybe add a message on the goal of the patch.

xlqian:
jenkins, test this please!

source/jormungandr/jormungandr/default_settings.py
xlqian: Hmm, here is the tricky thing:

For kraken, with/without the patch, parallelization is always activated
For valalla, without the patch,  parallelization is not (always) activated

source/jormungandr/jormungandr/api.py
xlqian: Yep

source/jormungandr/jormungandr/default_settings.py
```diff
@@ -109,3 +109,6 @@
     
 GRAPHICAL_ISOCHRONE = False
 HEAT_MAP = False
+# This parameter are used to apply gevent's monkey patch
+# The Goal is to activate parallel calling valhalla, without the patch, parallel http calling may not work
+PATCH_WITH_GEVENT_SOCKET = False
```
pbougue: no new line

pbougue: also, should we activate it on artemis for tests?

xlqian: @pbougue theoretically, this won't affect anything functional but **probably** the performance, which is not handled by artemis

source/jormungandr/tests/integration_tests_settings.py
```diff
@@ -59,3 +59,5 @@
 
 GRAPHICAL_ISOCHRONE = True
 HEAT_MAP = True
+
+PATCH_WITH_GEVENT_SOCKET = False
```
kinnou02: we should enable it on the test at least

source/jormungandr/jormungandr/default_settings.py
pbougue: If it is the code used on prod, it should be the one mainly tested... If we can make it crash on artemis, that's nicer than on prod :)

TeXitoi:
Jenkins, test this please

xlqian:
Jenkins, test this please

kinnou02 (approved):

antoine-de:
I don't really see how this will fix stuff (apart from being able to test the it)

source/jormungandr/jormungandr/api.py
```diff
@@ -41,10 +41,18 @@
...
+if rest_api.app.config.get('PATCH_WITH_GEVENT_SOCKET', False):
+    import gevent.socket
+    import urllib3
+    logger = logging.getLogger('jormungandr.patch_gevent_socket')
+    logger.info("Attention! You'are patching urllib3.connection.connection.socket with gevent.socket, "
+                "parallel calling valhalla is activated")
+    # This line replaces the gevent.monkey.patch_socket()
+    # the reason why we don't use patch_socket() at the very beginning of jormungandr is
+    # that it caused a mysterious performance regression for certain requests, thus we patch
+    # only at places where asynchronisation is needed
```
antoine-de, line 54: Ok @kinnou02 explained this to me.

maybe it would be good to have a bit more precise explanation in this comment (which is already nice btw!) explaining that we monkey_patch only http request because we want asynchronisation on those, but we don't want to activate it for reddis because it causes performance regression

xlqian: ok

xlqian: @antoine-de can you re-check this comment please? thx

antoine-de (approved):

PR #1891, opened 2016-12-09
Doc: fix for prices.csv

azime:


prhod (commented):
Why didn't you described the "currency" field ?

PR #1899, opened 2016-12-14
Kraken: fix disruption update crash

xlqian:
fix disruption update crash

http://jira.canaltp.fr/browse/NAVITIAII-2170

### DIFF
--- a/source/kraken/apply_disruption.cpp
+++ b/source/kraken/apply_disruption.cpp
@@ -586,11 +586,7 @@ struct delete_impacts_visitor : public apply_impacts_visitor {
         // it with an empty vector.
         decltype(mvj->impacted_by) impacted_by_moved;
         boost::swap(impacted_by_moved, mvj->impacted_by);
-        
-        // note: insert(it_begin, it_end) crashed...
-        for(const auto& i: impacted_by_moved) {
-            disruptions_collection.insert(i);
-        }
+        disruptions_collection.insert(std::begin(impacted_by_moved), std::end(impacted_by_moved));
     }
 
     void operator()(nt::StopPoint* stop_point) {
--- a/source/kraken/apply_disruption.cpp
+++ b/source/kraken/apply_disruption.cpp
@@ -526,11 +526,14 @@ static auto comp = [](const impact_wptr& lhs, const impact_wptr& rhs){
     auto i_lhs = lhs.lock();
     auto i_rhs = rhs.lock();
 
+    // lexical sort by pointer then update datetime then uri
     if(!i_lhs || !i_rhs) {
-       return false;
+       return i_lhs.get() < i_rhs.get();
     }
-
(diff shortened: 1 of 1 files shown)


source/kraken/apply_disruption.cpp
```diff
@@ -583,7 +586,11 @@ struct delete_impacts_visitor : public apply_impacts_visitor {
         // it with an empty vector.
         decltype(mvj->impacted_by) impacted_by_moved;
         boost::swap(impacted_by_moved, mvj->impacted_by);
-        disruptions_collection.insert(std::begin(impacted_by_moved), std::end(impacted_by_moved));
+        
+        // note: insert(it_begin, it_end) crashed...
+        for(const auto& i: impacted_by_moved) {
```
pbougue: With a correct "complete" comparator it still crashes?
Problem of references or something like that?
Would be interesting to investigate a little on that even after merging, this can be a real problem (memcheck errors?)

xlqian: yep

kinnou02:
it isn't possible to add tests?

TeXitoi:
Making a simple test that crash is nearly impossible :-(

kinnou02:
ok :(

source/kraken/apply_disruption.cpp
```diff
@@ -583,7 +586,11 @@ struct delete_impacts_visitor : public apply_impacts_visitor {
         // it with an empty vector.
         decltype(mvj->impacted_by) impacted_by_moved;
         boost::swap(impacted_by_moved, mvj->impacted_by);
-        disruptions_collection.insert(std::begin(impacted_by_moved), std::end(impacted_by_moved));
+        
+        // note: insert(it_begin, it_end) crashed...
```
TeXitoi: If that's the case, the problem is not solved.

TeXitoi (changes requested):

source/kraken/apply_disruption.cpp
```diff
@@ -526,11 +526,14 @@ static auto comp = [](const impact_wptr& lhs, const impact_wptr& rhs){
     auto i_lhs = lhs.lock();
     auto i_rhs = rhs.lock();
 
+    // lexical sort by pointer then update datetime then uri
```
TeXitoi: Nop, lexical sort by is_null then updated datetime then uri

source/kraken/apply_disruption.cpp
```diff
@@ -526,11 +526,14 @@ static auto comp = [](const impact_wptr& lhs, const impact_wptr& rhs){
...
     if(!i_lhs || !i_rhs) {
-       return false;
+       return i_lhs.get() < i_rhs.get();
```
TeXitoi: better to do a set of shared_ptr, to avoid this test and be sure the comparator is stable.

TeXitoi (changes requested):

TeXitoi (approved):

PR #1900, opened 2016-12-15
Jormungandr: fix no reverse geocoding in experimental

TeXitoi:
http://jira.canaltp.fr/browse/ITI-216

### DIFF
--- a/source/jormungandr/jormungandr/georef.py
+++ b/source/jormungandr/jormungandr/georef.py
@@ -48,8 +48,14 @@ class Kraken(object):
         if place.startswith("coord:"):
             # In some cases, the given "from" is not always a findable address by kraken
             # we just forward the given coord and return a pt object
+            _, lon, lat = place.split(':')
+            uri = '{};{}'.format(lon, lat)
             p = type_pb2.PtObject()
-            p.uri = place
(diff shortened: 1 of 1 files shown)


source/jormungandr/jormungandr/georef.py
```diff
@@ -48,8 +48,14 @@ def place(self, place):
         if place.startswith("coord:"):
             # In some cases, the given "from" is not always a findable address by kraken
             # we just forward the given coord and return a pt object
+            _, lon, lat = place.split(':')
```
antoine-de, line 51: wouldn't it be better to remove the old `coord:` thing ?

TeXitoi: That's a much bigger refactor.

pbougue: Seems a bit big but might be nice

TeXitoi: And it's opensource :innocent:

antoine-de: rohhhh:

```
 ❯ git grep '"coord:' | wc -l
21
```

but I'm kidding, it's ok to do this later

PR #1901, opened 2016-12-15
kraken: fix disable feedpublisher

TeXitoi:


xlqian (approved):

source/type/pb_converter.cpp
```diff
@@ -1946,6 +1946,7 @@ pbnavitia::Metadatas* PbCreator::mutable_metadatas(){
 
 void PbCreator::clear_feed_publishers(){
     response.clear_feed_publishers();
+    contributors.clear();
```
pbougue, line 1948: Not sure about a possible double use with: https://github.com/CanalTP/navitia/blob/dev/source/type/pb_converter.cpp#L1777 ?

TeXitoi: The clear you point is to not duplicate contributors if you call get_response several times. It's done after importing the contributors in the pb feedpublisher. This clear just removes the contributors. When you'll call get_response, contributors will then be empty. That was the bug, we were cleaning the pb but the pb was not filled.

PR #1909, opened 2016-12-23
Tyr: import stops in global autocomplete

antoine-de:
when importing NTGF or GTFS import the stops in the global autocomplete
(mimir)

the import is very quick for the moment and is done with `ed2nav`

It is not activated by default and can be activated for one instance
with `import_stops_in_mimir` in the instance API

### DIFF
--- a/source/tyr/migrations/versions/361f3993960e_import_stops_in_global_autocomplete_.py
+++ b/source/tyr/migrations/versions/361f3993960e_import_stops_in_global_autocomplete_.py
@@ -8,7 +8,7 @@ Create Date: 2016-12-23 11:48:47.003811
 
 # revision identifiers, used by Alembic.
 revision = '361f3993960e'
-down_revision = '57ee5581cc30'
+down_revision = '502b7c6fd00b'
 
 from alembic import op
 import sqlalchemy as sa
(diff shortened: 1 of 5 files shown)


source/tyr/migrations/versions/361f3993960e_import_stops_in_global_autocomplete_.py
```diff
@@ -0,0 +1,24 @@
...
+revision = '361f3993960e'
+down_revision = '57ee5581cc30'
+
+from alembic import op
+import sqlalchemy as sa
+
+
+def upgrade():
+    op.add_column('instance', sa.Column('import_stops_in_mimir', sa.Boolean(), nullable=True))
+    op.execute('update instance set import_stops_in_mimir=false')
```
xlqian, line 19: is this supposed to set the default value in db? have you tried `op.add_column(..., server_default=xx, ...)`?

antoine-de: this is done to set the value and then put the field as mandatory in the DB

pbougue (approved):

PR #1910, opened 2016-12-23
Jormun: marshall stop_area from bragi

pbougue:


source/jormungandr/jormungandr/interfaces/v1/Places.py
```diff
@@ -189,6 +189,36 @@ def output(self, key, obj):
                 create_administrative_regions_field(geocoding) or create_admin_field(geocoding),
         }
 
+
+class StopAreaField(fields.Raw):
+    def output(self, key, obj):
+        if not obj:
+            return None
+
+        coordinates = obj.get('geometry', {}).get('coordinates', [])
```
antoine-de: maybe we can factorize this

PR #1911, opened 2016-12-23
Jormungandr: activate autocomplete on instance

antoine-de:
add the ability to use the global autocomplete (mimir) instead of the
instance's autocomplete
(ie `/v1/coverage/<instance>/places `uses mimir)

By default it's the instance autocomplete that is used.

To activate it add to the configuration:
```json
"autocomplete": {
        "class":
"jormungandr.autocomplete.geocodejson.GeocodeJson",
        "args" : {
            "host": "the host"
        }
    },
```

Note: I don't like that we use the v1's marshall in the autocomplete,
but it's what was done for kraken and I don't see a nice way to avoid
this

Note2:
I changed the way the autocomplete was given in the configuration (to be consistent with the realtimeproxies), thus we'll need to change the jormungandr configuration (but it's not used yet, so no breakdown)

Note 3: lots of files changed because of some refactors for the tests


pbougue (commented):
Maybe add to test's conf the use of kraken in autocomplete, just to test conf loading (cause result would obviously be the same).

As for the conf naming it would be nice to have something consistent (and retrocompatible) :+1: 
I have no opinion on what's the best naming though.

antoine-de:
I updated the PR with some tests.

It impacts way more file due to some refactoring needed along the way..

pbougue (approved):

kinnou02 (approved):

PR #1917, opened 2017-01-04
Jormungandr: fix manhattan distance in experimental

azime:


source/kraken/worker.cpp
```diff
@@ -933,6 +933,8 @@ void Worker::street_network_routing_matrix(const pbnavitia::StreetNetworkRouting
                 k->set_routing_status(pbnavitia::RoutingStatus::unreached);
                 break;
             default:
+                k->set_duration(street_network_worker->departure_path_finder.crow_fly_duration(
+                                    entry_point.coordinates.distance_to(coord), true).total_seconds());
```
TeXitoi, line 937: I don't like that because with that, we will need to do the same thing in valhalla and geovelo.

PR #1918, opened 2017-01-04
Jormun: refactor admin marshall from bragi

pbougue:
* use city object from bragi, not administrative_region sub-object (can be empty)
* return str for lat,lon as in kraken

### DIFF
--- a/source/jormungandr/jormungandr/interfaces/v1/Places.py
+++ b/source/jormungandr/jormungandr/interfaces/v1/Places.py
@@ -103,13 +103,14 @@ def create_administrative_regions_field(geocoding):
     response = []
     for admin in administrative_regions:
         coord = admin.get('coord', {})
-        lat = coord.get('lat') if coord else None
-        lon = coord.get('lon') if coord else None
+        lat = str(coord.get('lat')) if coord and coord.get('lat') else None
+        lon = str(coord.get('lon')) if coord and coord.get('lon') else None
         zip_codes = admin.get('zip_codes', [])
         response.append({
             "insee": admin.get('insee'),
             "name": admin.get('label'),
-            "level": int(admin.get('level')),
+            "level":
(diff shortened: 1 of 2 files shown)


source/jormungandr/jormungandr/interfaces/v1/Places.py
```diff
@@ -122,13 +123,13 @@ def create_administrative_regions_field(geocoding):
 
 
 def get_lon_lat(obj):
-    if not obj:
+    if not obj or not obj.get('geometry') or not obj.get('geometry').get('coordinates'):
```
antoine-de, line 126: I don't see the point of this check

source/jormungandr/jormungandr/interfaces/v1/Places.py
```diff
@@ -142,8 +143,25 @@ class AdministrativeRegionField(fields.Raw):
...
+            "level":
+                int(geocoding.get('level')) if geocoding.get('level') else None ,
+            "name": geocoding.get('name'),
+            "label": geocoding.get('label'),
+            "id": geocoding.get('id'),
+            "coord": {
+                "lat": lat,
+                "lon": lon
+            },
+            "zip_code": geocoding.get('postcode'),
```
antoine-de, line 161: shouldn't we `format_zip_code` here ?

antoine-de (approved):

source/jormungandr/jormungandr/interfaces/v1/Places.py
pbougue: just (over-)protecting code bellow

source/jormungandr/jormungandr/interfaces/v1/Places.py
pbougue: in geocodejson it's a string not a table

PR #1920, opened 2017-01-05
jormungandr: Choosing journey on waiting durations

kadhikari:


kadhikari:
concerned Ticket: http://jira.canaltp.fr/browse/ITI-203

source/jormungandr/jormungandr/scenarios/journey_filter.py
```diff
@@ -276,6 +279,18 @@ def _get_mode_weight(journey):
     return max(mode_weight.get(mode, 1) for mode in journey.tags)
 
 
+def get_min_waiting(journey):
+    """
+    Returns min waiting time in a journey
+    """
+    waiting = 3600*4
```
antoine-de: why this 4h waiting ?

kadhikari: This is the max waiting value allowed while filtering journeys

antoine-de: but this value is not relevant for your method (and if we change the max waiting value we'll forget to update this method :wink: and a journey with no sections will have a 4h waiting)


and I think it will be more pythonic to write this (not tested): 
```python
return 0 if not journey.sections 
             else min(s.duration for s in journey.sections if s.type == response_pb2.WAITING)
```

or you can write this (but it's less straightforward :stuck_out_tongue_winking_eye: )
```python
return min((s.duration for s in journey.sections if s.type == response_pb2.WAITING) or [0])
```

TeXitoi: 0 is a strange default value.

TeXitoi: and the first version doesn't work

TeXitoi: (neither the second)

source/jormungandr/jormungandr/scenarios/journey_filter.py
```diff
@@ -101,6 +101,7 @@ def _get_worst_similar(j1, j2, request):
      - asap
      - less fallback
      - duration
+     - smaller value among min waiting duration
```
TeXitoi, line 105: greater value amoung min waiting duration

source/jormungandr/jormungandr/scenarios/tests/journey_compare_tests.py
```diff
@@ -396,6 +395,111 @@ def test_similar_journeys_different_transfer():
...
+
+    journey_filter._filter_similar_vj_journeys(list(journeys_gen(responses)), {})
+
+    assert 'to_delete' not in journey2.tags
+    assert 'to_delete' in journey1.tags
+
+
+def test_similar_journeys_multi_trasfer_and_different_waiting_durations():
+    """
+     If 2 journeys take the same vjs, same number of sections and plusieurs waitings with different waiting durations,
```
TeXitoi: s/plusieurs/several/

source/jormungandr/jormungandr/scenarios/journey_filter.py
kadhikari: This function returns worst among two journeys not the one we keep.

source/jormungandr/jormungandr/scenarios/journey_filter.py
```diff
@@ -283,13 +287,7 @@ def get_min_waiting(journey):
...
     Returns min waiting time in a journey
     """
-    waiting = 3600*4
-    for s in journey.sections:
-        if s.type != response_pb2.WAITING:
-            continue
-        waiting = min(waiting, s.duration)
-    return waiting
-
+    return min(([s.duration for s in journey.sections if s.type == response_pb2.WAITING]) or [0])
```
TeXitoi: Useless () around []

TeXitoi (changes requested):

antoine-de (approved):

source/jormungandr/jormungandr/scenarios/journey_filter.py
```diff
@@ -276,6 +283,12 @@ def _get_mode_weight(journey):
     return max(mode_weight.get(mode, 1) for mode in journey.tags)
 
 
+def get_min_waiting(journey):
+    """
+    Returns min waiting time in a journey
+    """
+    return min([s.duration for s in journey.sections if s.type == response_pb2.WAITING] or [0])
```
antoine-de, line 290: Ah no it won't work, the default value need to be +infinity because we want to max this (but then we should rename this function no ?)

TeXitoi: finally, no, as we use this function with same number of transfer. So, if there is a waiting in one and no waiting in the other, that's because the waiting is 0.

TeXitoi: But, in practice, this case will just not exists.

TeXitoi (approved):

PR #1922, opened 2017-01-06
Jormun: pass lon and lat to geocodejson autocomplete

datanel:
Add the possibility to give a coordinate to the autocomplete's API to prioritize the objects around the coordinate.

### Example
http://127.0.0.1:5000/v1/coverage/default/places?q=parking&from=2.34;48.85

### DIFF
--- a/source/jormungandr/jormungandr/autocomplete/geocodejson.py
+++ b/source/jormungandr/jormungandr/autocomplete/geocodejson.py
@@ -34,7 +34,6 @@ import logging
 from jormungandr.autocomplete.abstract_autocomplete import AbstractAutocomplete
 import requests
 from jormungandr.exceptions import TechnicalError
-from urllib import urlencode
 
 
 class GeocodeJson(AbstractAutocomplete):
@@ -59,13 +58,11 @@ class GeocodeJson(AbstractAutocomplete):
         if request["from"]:
             params["lon"], params["lat"] = self.get_coords(request["from"])
 
-        url = '{endpoint}?{query}'.format(endpoint=self.external_api, query=urlencode(params))
-
         try:
             if shape:
-                raw_response = requests.post(url, timeout=self.timeout, json=shape)
+                raw_response = requests.post(self.external_api, timeout=self.timeout, json=shape, params=params)
             else:
-                raw_response = requests.get(url, timeout=self.timeout)
+                raw_response = requests.get(self.external_api, timeout=self.timeout, params=params)
 
         except requests.Timeout:
             logging.getLogger(__name__).error('autocomplete request timeout')
@@ -84,4 +81,8 @@ class GeocodeJson(AbstractAutocomplete):
         raise NotImplementedError
 
     def get_coords(self, param):
+        """
+        Get coordinates (longitude, latitude).
+        For moment we consider that the param can only be a coordinate.
+        """
         return param.split(";")
--- a/source/jormungandr/jormungandr/autocomplete/geocodejson.py
+++ b/source/jormungandr/jormungandr/autocomplete/geocodejson.py
@@ -55,7 +55,7 @@ class GeocodeJson(AbstractAutocomplete):
             "count": request["count"]
         }
 
-        if request["from"]:
+        if request.get("from"):
             params["lon"], params["lat"] = self.get_coords(request["from"])
 
         try:
--- a/source/jormungandr/jormungandr/autocomplete/geocodejson.py
+++ b/source/jormungandr/jormungandr/autocomplete/geocodejson.py
@@ -52,7 +52,7 @@ class GeocodeJson(AbstractAutocomplete):
 
         params = {
             "q": request["q"],
-            "count": request["count"]
+            "limit": request["count"]
         }
 
         if request.get("from"):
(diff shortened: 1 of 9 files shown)


source/jormungandr/jormungandr/autocomplete/geocodejson.py
```diff
@@ -50,9 +51,16 @@ def get(self, request, instance, shape=None):
...
-                                                      count=request['count'])
+        params = {
+            "q": request["q"],
+            "count": request["count"]
+        }
+
+        if request["from"]:
+            params["lon"], params["lat"] = self.get_coords(request["from"])
+
+        url = '{endpoint}?{query}'.format(endpoint=self.external_api, query=urlencode(params))
```
antoine-de: I think it's better to use the `requests` build in params no ?

`requests.post(url, timeout=self.timeout, json=shape, params=params)`

source/jormungandr/jormungandr/autocomplete/geocodejson.py
```diff
@@ -75,4 +83,5 @@ def get(self, request, instance, shape=None):
     def geo_status(self, instance):
         raise NotImplementedError
 
-
+    def get_coords(self, param):
+        return param.split(";")
```
antoine-de, line 88: I think it would be nice to add a comment saying that for moment we consider that the param can only be a coordinate

source/jormungandr/jormungandr/interfaces/parsers.py
```diff
@@ -74,3 +74,18 @@ def unsigned_integer(value):
         return d
     except ValueError as e:
         raise ValueError("Unable to evaluate, {}".format(e))
+
+
+def coord_format(coord):
```
antoine-de, line 79: :+1:

source/jormungandr/jormungandr/interfaces/v1/Places.py
```diff
@@ -313,6 +314,9 @@ def __init__(self, *args, **kwargs):
         self.parsers['get'].add_argument("disable_geojson", type=boolean, default=False,
                             description="remove geojson from the response")
 
+        self.parsers['get'].add_argument("from", type=coord_format,
+                                         description="Coorddinates longitude:latitude from where you want to search")
```
antoine-de: longitude **;** latitude no ?
and typo on coordinate

in this wouldn't it be nice to add the fact that the coordinate is used to prioritize the objects around this coordinate ?

source/jormungandr/jormungandr/autocomplete/geocodejson.py
```diff
@@ -50,9 +51,16 @@ def get(self, request, instance, shape=None):
         if not self.external_api:
             raise TechnicalError('global autocomplete not configured')
 
-        url = '{endpoint}?q={q}&limit={count}'.format(endpoint=self.external_api,
-                                                      q=request['q'],
-                                                      count=request['count'])
+        params = {
```
antoine-de, line 53: for me (it has not been done in your PR) we screwd on those parameters.

the `request` should only be the requested word to autocomplete (`request["q"]`)

And we should add another param with the `count` and the `from`

I think thus the `AbstractAutocomplete.get(..)` should be written with only the core parameters:
``` python
def get(self, q, instance, count, **kwargs):
```

and add the needed additional params in the specific class (for Autocomplete.Kraken the `search_type`, the `_current_datetime`, ...)

what do you think @kinnou02 @TeXitoi @pbougue ?

source/jormungandr/jormungandr/interfaces/parsers.py
```diff
@@ -74,3 +74,18 @@ def unsigned_integer(value):
...
+    """
+    Validate coordinates format (lon;lat)
+    """
+    lon_lat_splitted = coord.split(";")
+    if len(lon_lat_splitted) != 2:
+        raise ValueError('Invalid coordinate parameter. It must be lon;lat where lon and lat are floats.')
+
+    lon, lat = lon_lat_splitted
+    float(lon)
+    float(lat)
```
xlqian: Would be it useful if we check `lon` and `lat` 's ranges...

source/jormungandr/jormungandr/interfaces/parsers.py
```diff
@@ -74,3 +74,18 @@ def unsigned_integer(value):
...
+
+def coord_format(coord):
+    """
+    Validate coordinates format (lon;lat)
+    """
+    lon_lat_splitted = coord.split(";")
+    if len(lon_lat_splitted) != 2:
+        raise ValueError('Invalid coordinate parameter. It must be lon;lat where lon and lat are floats.')
+
+    lon, lat = lon_lat_splitted
```
xlqian, line 87: `map(float, lon_lat_splitted)`

xlqian: you can put all together:
```
if len(map(float, lon_lat_splitted)) != 2:
...
```
(not sure it's clearer)

antoine-de: yes I'm not convinced this is clearer

source/jormungandr/jormungandr/autocomplete/geocodejson.py
pbougue: I agree but that does not seems so urgent.

pbougue (approved):

PR #1926, opened 2017-01-11
jormun: don't use circuit breaker on init

kinnou02:


source/jormungandr/jormungandr/instance.py
```diff
@@ -444,7 +444,7 @@ def init(self):
         req = request_pb2.Request()
         req.requested_api = type_pb2.METADATAS
         try:
-            resp = self.send_and_receive(req, timeout=1000, quiet=True)
+            resp = self._send_and_receive(req, timeout=1000, quiet=True)
```
antoine-de, line 448: can you add a comment there ?

antoine-de (approved):

PR #1927, opened 2017-01-11
Ed: change uri of admin to match mimir

pbougue:
id of admin is now `admin:fr:<insee>` if admin has insee code (France),
`admin:osm:<osmid>` otherwise

WARNING: this is a breaking change for anyone using admin's ids (which, as we frequently stress, is not a good practice)

TODO:
- [x] Test docker? all is tested in artemis
- [x] Prepare artemis that should break (add tests to ensure cities, osm and geopal is ok)
- [x] Check that the cities database is updated on deploy (to be seen with @Hoshiyo and @fdegomme-canaltp ) : ticket on HOR board for that
- [x] Check why 63/1.5million addresses are lost on artemis destineo geo_status: https://github.com/CanalTP/navitia/pull/1927/files#diff-c2e4a88ca2720be59f239d69571c963aR291

Needs https://github.com/CanalTP/artemis/pull/249 and https://github.com/CanalTP/artemis_references/pull/91 to be merged for artemis

### DIFF
--- a/source/cities/cities.cpp
+++ b/source/cities/cities.cpp
@@ -178,10 +178,14 @@ void OSMCache::insert_relations() {
             std::string polygon_str = polygon_stream.str();
             const auto coord = "POINT(" + std::to_string(relation.second.centre.get<0>())
                 + " " + std::to_string(relation.second.centre.get<1>()) + ")";
+
+            auto uri = "admin:osm:" + std::to_string(relation.first);
+            if ( ! relation.second.insee.empty()) {
+                uri = "admin:fr:" + relation.second.insee;
+            }
             lotus.insert({std::to_string(relation.first), relation.second.name,
                           relation.second.postal_code(), relation.second.insee,
-                          std::to_string(relation.second.level), coord, polygon_str,
-                          "admin:"+std::to_string(relation.first)});
+                          std::to_string(relation.second.level), coord, polygon_str, uri});
         } else {
             ++nb_empty_polygons;
         }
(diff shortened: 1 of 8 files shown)


source/cities/cities.cpp
```diff
@@ -178,10 +178,14 @@ void OSMCache::insert_relations() {
             std::string polygon_str = polygon_stream.str();
             const auto coord = "POINT(" + std::to_string(relation.second.centre.get<0>())
                 + " " + std::to_string(relation.second.centre.get<1>()) + ")";
+
+            auto uri = "admin:osm:"+std::to_string(relation.first);
```
xlqian: spacing

source/ed/osm2ed.cpp
```diff
@@ -395,9 +395,12 @@ void OSMCache::insert_relations() {
             std::string polygon_str = polygon_stream.str();
             const auto coord = "POINT(" + std::to_string(relation.centre.get<0>())
                 + " " + std::to_string(relation.centre.get<1>()) + ")";
+            auto uri = "admin:osm:"+std::to_string(relation.osm_id);
```
xlqian: same here :)

source/ed/ed_persistor.cpp
```diff
@@ -130,7 +130,8 @@ void EdPersistor::insert_admins(const ed::Georef& data){
     for(const auto& itm : data.admins){
         if(itm.second->is_used){
             this->lotus.insert({std::to_string(itm.second->id), itm.second->name, itm.second->insee,
-                               itm.second->level, this->to_geographic_point(itm.second->coord),"admin:" + itm.second->insee});
+                               itm.second->level, this->to_geographic_point(itm.second->coord),
```
antoine-de, line 133: I find it strange to do stuff here.

you are sure it is needed ? why isn't it done in osm2ed ?

pbougue: This is the geopal2ed part that calls `persist()`, that calls `insert_admins()`.
So the code is not so consistently stored, but to me it's needed.

PR #1929, opened 2017-01-13
Tyr: add the tyr tests to the `make docker_test` target

antoine-de:
This way the tests will hopefully be more run


pbougue (commented):
Cool PR!
I see 1 possible downside: cmake does not manage virtualenv and if we have conflicts between eitri and tyr, we will pay it when using this target.
I see 1 possible little improvment: add a intermediate target for eitri_test, called in docker_test. That way we could call them separately (might be usefull if virtualenv needs separate treatment or if one target is too long).

antoine-de:
we want to keep the number of 'real' targets as little as possible.

having a `make docker_test` is already a trade off (not to have a strong dependency on docker)

we don't want 4 commands to run all the tests, that one of the purpose of the makefile 😉 

So the additional target would have to be justified by being necessary (and we'll be able to add it later on)

have you checked that there is sub target yet ?

The pip dependencies are however indeed a problem, I don't know the canonical way to handle them

source/jormungandr/requirements_dev.txt
```diff
@@ -1,3 +1,4 @@
+-r requirements.txt
```
xlqian, line 1: 👍

PR #1930, opened 2017-01-13
Experimental: Configure engine per mode

xlqian:
http://jira.canaltp.fr/browse/ITI-342

### DIFF
--- a/source/jormungandr/jormungandr/instance.py
+++ b/source/jormungandr/jormungandr/instance.py
@@ -115,11 +115,12 @@ class Instance(object):
     @staticmethod
     def _set_default_street_network_config(street_network_configs):
         default_sn_class = 'jormungandr.street_network.kraken.Kraken'
-        modes_in_configs = set(list(itertools.chain.from_iterable(config.get("modes", []) for config in street_network_configs)))
-        modes_not_set = set(STREET_NETWORK_MODES) - modes_in_configs
 
-        street_network_configs.append({"modes": list(modes_not_set),
-                                       "class": default_sn_class})
+        modes_in_configs = set(list(itertools.chain.from_iterable(config['modes'] for config in street_network_configs)))
+        modes_not_set = set(STREET_NETWORK_MODES) - modes_in_configs
+        if modes_not_set:
+            street_network_configs.append({"modes": list(modes_not_set),
+                                           "class": default_sn_class})
     def get_models(self):
         if self.name not in g.instances_model:
             g.instances_model[self.name] = self._get_models()
--- a/source/jormungandr/jormungandr/instance.py
+++ b/source/jormungandr/jormungandr/instance.py
@@ -479,9 +479,6 @@ class Instance(object):
         return False
 
     def get_street_network_routing_matrix(self, origins, destinations, mode, max_duration_to_pt, request, **kwargs):
-        if mode not in self.street_network_services:
-            abort(400, "{} is not handled, only {} are valid street network modes".format(mode,
-                                                                                          STREET_NETWORK_MODES))
         return self.street_network_services[mode].get_street_network_routing_matrix(origins,
                                                                                     destinations,
                                                                                     mode,
@@ -490,9 +487,6 @@ class Instance(object):
                                                                                     **kwargs)
 
     def direct_path(self, mode, pt_object_origin, pt_object_destination, datetime, clockwise, request, **kwargs):
-        if mode not in self.street_network_services:
-            abort(400, "{} is not handled, only {} are valid street network mode".format(mode,
-                                                                                         STREET_NETWORK_MODES))
         return self.street_network_services[mode].direct_path(mode,
                                                               pt_object_origin,
                                                               pt_object_destination,
--- a/source/jormungandr/jormungandr/instance.py
+++ b/source/jormungandr/jormungandr/instance.py
@@ -70,6 +70,18 @@ def _init_g():
     g.instances_model = {}
 
 
+def _set_default_street_network_config(street_network_configs):
+    street_network_configs = street_network_configs or []
+    default_sn_class = 'jormungandr.street_network.kraken.Kraken'
+
+    modes_in_configs = set(list(itertools.chain.from_iterable(config['modes'] for config in street_network_configs)))
+    modes_not_set = set(STREET_NETWORK_MODES) - modes_in_configs
+    if modes_not_set:
+        street_network_configs.append({"modes": list(modes_not_set),
+                                       "class": default_sn_class})
+    return street_network_configs
+
+
 class Instance(object):
 
     def __init__(self,
@@ -96,9 +108,7 @@ class Instance(object):
         self.georef = georef.Kraken(self)
         self.planner = planner.Kraken(self)
 
-        if street_network_configurations is None:
-            street_network_configurations = []
-        self._set_default_street_network_config(street_network_configurations)
+        street_network_configurations = _set_default_street_network_config(street_network_configurations)
         self.street_network_services = street_network.StreetNetwork.get_street_network_services(self,
                                                                                                 street_network_configurations)
         self.ptref = ptref.PtRef(self)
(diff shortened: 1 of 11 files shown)


xlqian:
the config file becomes
```json
{
  "key": "stif",
  "zmq_socket": "ipc:///tmp/stif_kraken",
  "street_network": [{
    "modes": ["car", "walking"],
    "class": "jormungandr.street_network.valhalla.Valhalla"
  },
  {
    "modes": ["bss"],
    "class": "jormungandr.street_network.kraken.Kraken"
  }]
}
```

Note that:
* street network mode are grouped by engine 
* modes that are not configured in config files will take kraken as default engine

xlqian:
This is a very first version to make street network engine configurable by its mode.
Challenge this PR as much as you can

xlqian:
Jenkins, test this please!

source/jormungandr/jormungandr/street_network/kraken.py
```diff
@@ -32,8 +32,10 @@
 from jormungandr.exceptions import TechnicalError
 from navitiacommon import request_pb2, type_pb2
 from jormungandr.utils import get_uri_pt_object
+from jormungandr.street_network.street_network import IStreetNetwrokService
 
-class Kraken(object):
+
+class Kraken(IStreetNetwrokService):
```
antoine-de: typo on Network

source/jormungandr/jormungandr/street_network/street_network.py
```diff
@@ -31,32 +31,52 @@
...
 from importlib import import_module
 import logging
+import abc
+
+# Using abc.ABCMeta in a way it is compatible both with Python 2.7 and Python 3.x
+# http://stackoverflow.com/a/38668373/1614576
+ABC = abc.ABCMeta(str("ABC"), (object,), {})
+
+
+class IStreetNetwrokService(ABC):
```
antoine-de: I'm not a fan of the **I**Street... notation.

bad memories :stuck_out_tongue_winking_eye:

source/jormungandr/jormungandr/street_network/street_network.py
```diff
@@ -31,32 +31,52 @@
...
-        except AttributeError:
-            log.critical('impossible to build StreetNetwork, cannot find class: {}'.format(cls))
-            raise AttributeError('impossible to build StreetNetwork, cannot find class: {}'.format(cls))
-
-        log.info('** StreetNetwork {} used for direct_path **'.format(name))
-        return attr(instance=instance, url=service_url, **args)
+        street_network_services = {}
+        for config in street_network_configurations:
+            try:
+                cls = config['class']
```
antoine-de: there is the `create_object` method in utils to do this, maybe you can use it

source/jormungandr/jormungandr/instance.py
```diff
@@ -94,8 +95,12 @@ def __init__(self,
                                                 reset_timeout=app.config['CIRCUIT_BREAKER_INSTANCE_TIMEOUT_S'])
         self.georef = georef.Kraken(self)
         self.planner = planner.Kraken(self)
-        self.street_network_service = street_network.StreetNetwork.get_street_network(self,
-                                                                                      street_network_configuration)
+
+        if street_network_configurations is None:
+            street_network_configurations = []
+        self._set_default_street_network_config(street_network_configurations)
```
antoine-de: you can make it a free function (It might be a bit clearer that it changes only the input dict) (and then you can move the 2 line above in the method)

source/jormungandr/jormungandr/instance.py
```diff
@@ -458,3 +471,26 @@ def init(self):
             # the next request. We don't want to purge all our cache for a small error.
             logging.getLogger(__name__).debug('timeout on init for %s', self.name)
         return False
+
+    def get_street_network_routing_matrix(self, origins, destinations, mode, max_duration_to_pt, request, **kwargs):
+        if mode not in self.street_network_services:
+            abort(400, "{} is not handled, on {} are valid street network mode".format(mode,
```
antoine-de: I think you meant "on**ly**"

source/jormungandr/jormungandr/instance.py
```diff
@@ -458,3 +471,26 @@ def init(self):
...
+        return self.street_network_services[mode].get_street_network_routing_matrix(origins,
+                                                                                    destinations,
+                                                                                    mode,
+                                                                                    max_duration_to_pt,
+                                                                                    request,
+                                                                                    **kwargs)
+
+    def direct_path(self, mode, pt_object_origin, pt_object_destination, datetime, clockwise, request, **kwargs):
+        if mode not in self.street_network_services:
+            abort(400, "{} is not handled, on {} are valid street network mode".format(mode,
```
antoine-de: same here, I think you meant "only"

source/jormungandr/jormungandr/street_network/street_network.py
```diff
@@ -57,26 +57,28 @@ def get_street_network_services(instance, street_network_configurations):
         for config in street_network_configurations:
             try:
                 cls = config['class']
+                modes = config['modes']
             except KeyError, TypeError:
                 log.critical('impossible to build a routing, missing mandatory field in configuration')
```
antoine-de: wouldn't it be better to log the exception (`log.exception`) ?

source/jormungandr/jormungandr/street_network/street_network.py
pbougue: And "Network"

xlqian: @antoine-de how about `AbstractStreetNetworkService` ? Too long maybe...

source/jormungandr/jormungandr/instance.py
xlqian: 1. Ok for the free function
2. I cannot put L99 and L100 into the free function. because `_set_default_street_network_config` will take the reference of `street_network_configurations` and when it's `None`(instead of a real object), it's not modifiable

source/jormungandr/jormungandr/street_network/street_network.py
xlqian: @pbougue 👍

source/jormungandr/jormungandr/instance.py
```diff
@@ -62,23 +63,35 @@
...
 }
 
+STREET_NETWORK_MODES = ('walking', 'car', 'bss', 'bike')
+
 @app.before_request
 def _init_g():
     g.instances_model = {}
 
 
+def _set_default_street_network_config(street_network_configs):
```
antoine-de, line 75: can you add a comment there ?

source/jormungandr/jormungandr/street_network/street_network.py
```diff
@@ -31,32 +31,56 @@
...
-            log.critical('impossible to build a routing, missing mandatory field in configuration')
-            raise KeyError('impossible to build a routing, missing mandatory field in configuration')
-
-        args = street_network_configuration.get('args', {})
-        service_url = args.get('service_url', None)
-        try:
+        street_network_services = {}
+        for config in street_network_configurations:
+            try:
+                cls = config['class']
```
antoine-de: there is the create_object method in utils to do this, can't you use it ?

source/jormungandr/jormungandr/instance.py
```diff
@@ -458,3 +474,26 @@ def init(self):
             # the next request. We don't want to purge all our cache for a small error.
             logging.getLogger(__name__).debug('timeout on init for %s', self.name)
         return False
+
+    def get_street_network_routing_matrix(self, origins, destinations, mode, max_duration_to_pt, request, **kwargs):
+        if mode not in self.street_network_services:
+            abort(400, "{} is not handled, only {} are valid street network modes".format(mode,
```
kinnou02: I don't like the abort like that, in the middle of our application code

source/jormungandr/jormungandr/street_network/street_network.py
xlqian: nice!!!

source/jormungandr/jormungandr/instance.py
```diff
@@ -476,9 +479,6 @@ def init(self):
         return False
 
     def get_street_network_routing_matrix(self, origins, destinations, mode, max_duration_to_pt, request, **kwargs):
-        if mode not in self.street_network_services:
```
antoine-de: hum, even if the `mode` is checked by the API, I think it would be nice not to crash here (if we add another mode  for exemple)

antoine-de (approved):

PR #1934, opened 2017-01-17
jormungandr: marshall citycode from bragi response

patochectp:


antoine-de (approved):
:+1: 

can you also update the feature in:

https://github.com/CanalTP/navitia/blob/dev/source/jormungandr/tests/bragi_autocomplete_tests.py
?

PR #1936, opened 2017-01-17
jormungandr: fix multimode in experimental + activate lots of asserts

TeXitoi:
Will make one more Artemis fail (but that's a good one). Artemis can be then fixed.

### DIFF
--- a/source/jormungandr/jormungandr/scenarios/experimental.py
+++ b/source/jormungandr/jormungandr/scenarios/experimental.py
@@ -429,6 +429,7 @@ class AsyncWorker(object):
         for dep_mode, arr_mode in self.krakens_call:
             dp_key = make_direct_path_key(dep_mode, origin.uri, destination.uri, datetime, clockwise, reverse_sections)
             dp = fallback_direct_path.get(dp_key)
+            journey_parameters = copy.deepcopy(journey_parameters)
             if dp.journeys:
                 journey_parameters.direct_path_duration = dp.journeys[0].durations.total
             else:
@@ -437,14 +438,20 @@ class AsyncWorker(object):
             origins = origins_fallback.get_locations_contexts(dep_mode)
             destinations = destinations_fallback.get_locations_contexts(arr_mode)
 
-            def worker_journey():
+            # default argument to workaround late binding
+            # http://stackoverflow.com/questions/3431676/creating-functions-in-a-loop
+            def worker_journey(dep_mode=dep_mode,
+                               arr_mode=arr_mode,
+                               journey_parameters=journey_parameters,
+                               origins=origins,
+                               destinations=destinations):
                 if not origins or not destinations or not self.request.get('max_duration', 0):
                     return dep_mode, arr_mode, None
                 return dep_mode, arr_mode, instance.planner.journeys(origins, destinations,
                                                                      datetime, clockwise,
                                                                      journey_parameters)
             futures_jourenys.append(self.pool.spawn(worker_journey))
-            return futures_jourenys
+        return futures_jourenys
 
     @staticmethod
     def _extend_journey(pt_journey, mode, pb_from, pb_to, departure_date_time, nm, clockwise,
--- a/source/jormungandr/tests/journey_common_tests.py
+++ b/source/jormungandr/tests/journey_common_tests.py
@@ -34,7 +34,7 @@ import mock
 from .check_utils import *
 
 
-def check_journeys(resp):
+def check_best(resp):
     assert not resp.get('journeys') or sum((1 for j in resp['journeys'] if j['type'] == "best")) == 1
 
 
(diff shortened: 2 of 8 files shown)


xlqian:
GG!

source/jormungandr/jormungandr/scenarios/experimental.py
```diff
@@ -437,14 +438,20 @@ def get_pt_journey_futures(self, origin, destination, fallback_direct_path, orig
...
             destinations = destinations_fallback.get_locations_contexts(arr_mode)
 
-            def worker_journey():
+            # default argument to workaround late binding
+            # http://stackoverflow.com/questions/3431676/creating-functions-in-a-loop
+            def worker_journey(dep_mode=dep_mode,
+                               arr_mode=arr_mode,
+                               journey_parameters=journey_parameters,
+                               origins=origins,
+                               destinations=destinations):
```
xlqian, line 447: 👍 
Since it's a tricky bug to find, maybe henceforth we should avoid using functions defined inside a loop especially when using async calling. I suggest to using `functools.partial`, it's kind of `std::bind` in C++

source/jormungandr/jormungandr/scenarios/experimental.py
```diff
@@ -437,14 +438,20 @@ def get_pt_journey_futures(self, origin, destination, fallback_direct_path, orig
...
                 if not origins or not destinations or not self.request.get('max_duration', 0):
                     return dep_mode, arr_mode, None
                 return dep_mode, arr_mode, instance.planner.journeys(origins, destinations,
                                                                      datetime, clockwise,
                                                                      journey_parameters)
             futures_jourenys.append(self.pool.spawn(worker_journey))
-            return futures_jourenys
+        return futures_jourenys
```
xlqian, line 454: moving a large bloc of codes out may avoid the miss indentation too :)  (yes, these are my codes and i'm criticizing myself :sweat:)

source/jormungandr/tests/journey_common_tests.py
```diff
@@ -818,6 +902,8 @@ def test_journeys_without_show_codes(self):
             .format(from_sa="A", to_sa="D", datetime="20120614T080000")
 
         response = self.query_region(query, display=False)
+        check_best(response)
+        #self.is_valid_journey_response(response, query)# linestring with 1 value (0,0)
```
xlqian, line 925: still useful?

source/jormungandr/tests/journey_common_tests.py
```diff
@@ -837,6 +923,8 @@ def test_journeys_with_show_codes(self):
...
```
xlqian, line 946: still useful?

source/jormungandr/tests/routing_tests_experimental.py
```diff
@@ -49,35 +48,14 @@ class TestJourneysExperimental(JourneyCommon, DirectPath, AbstractTestFixture):
...
-
     def test_journey_with_different_fallback_modes(self):
         """
         Test when departure/arrival fallback modes are different
         """
         query = journey_basic_query + "&first_section_mode[]=walking&last_section_mode[]=car&debug=true"
         response = self.query_region(query)
-        check_journeys(response)
+        check_best(response)
+        #self.is_valid_journey_response(response, query)# linestring with 1 value (0,0)
```
xlqian, line 58: still useful?

source/jormungandr/tests/tests_mechanism.py
```diff
@@ -236,8 +236,14 @@ def query_no_assert(self, url, display=False):
     def check_journeys_links(self, response, query_dict):
         journeys_links = get_links_dict(response)
         clockwise = query_dict.get('datetime_represents', 'departure') == "departure"
+        has_pt = any(s for j in response['journeys'] for s in j['sections'] if s['type'] == 'public_transport')
+        print(query_dict)
```
xlqian: still useful?

source/jormungandr/tests/tests_mechanism.py
```diff
@@ -236,8 +236,14 @@ def query_no_assert(self, url, display=False):
     def check_journeys_links(self, response, query_dict):
         journeys_links = get_links_dict(response)
         clockwise = query_dict.get('datetime_represents', 'departure') == "departure"
+        has_pt = any(s for j in response['journeys'] for s in j['sections'] if s['type'] == 'public_transport')
```
xlqian: I think what you want is this: `has_pt = any(s['type'] == 'public_transport' for j in response['journeys'] for s in j['sections'])`

source/jormungandr/tests/journey_common_tests.py
TeXitoi: I can't activate the check because the linestring contains only one (0,0) element. The check asserts that at least 2 elements is needed. I don't know if I should correct the check or correct the code. As that's only adding test, I let that like that for the future.

xlqian (approved):

source/jormungandr/tests/tests_mechanism.py
```diff
@@ -299,7 +304,7 @@ def is_valid_journey_response(self, response, query_str):
         # to have forwarded all params, (and the time must be right)
         self.check_journeys_links(response, query_dict)
 
-        feed_publishers = get_not_null(response, "feed_publishers")
+        feed_publishers = response.get("feed_publishers", [])
```
antoine-de, line 307: there can be no feed_publisher ?

antoine-de (approved):

source/jormungandr/tests/tests_mechanism.py
TeXitoi: If there is no license yes.

PR #1938, opened 2017-01-20
kraken: fix a crash due to a data lifetime bug

TeXitoi:


xlqian (approved):

kinnou02 (approved):

kinnou02:
I don't like the interaction between doWork, the worker and the pbCreator, but it's minor we will look into it later.

PR #1941, opened 2017-01-24
!BREAKING CHANGE! Jormungandr: enable configuration from environment variable

kinnou02:
This PR also introduce the configuration of kraken from environment
variables and of course make the configuration file optional.

These variables are very badly named since we want to be able to read old
configuration files and a new option_descriptions or a mapping table
between environment variable and file's key is error prone since every
change has to be made two time at least.

So we have ugly names with the following form: `KRAKEN_{SECTION}_{FIELD_NAME}`
This will converted to `{SECTION}.{FIELD_NAME}`.

So to set the zmq_socket we need to define a variable named
`KRAKEN_GENERAL_zmq_socket`

Since the configuration file was also used for logger configuration, I
added two new properties so we can configure the logging with
environment variables too.

https://github.com/CanalTP/utils/pull/47 need to be merde before this one.

### DIFF
--- a/source/jormungandr/jormungandr/default_settings.py
+++ b/source/jormungandr/jormungandr/default_settings.py
@@ -1,9 +1,11 @@
 # encoding: utf-8
 
 import logging
+import os
+import json
 
 # path of the configuration file for each instances
-INSTANCES_DIR = '/etc/jormungandr.d'
+INSTANCES_DIR = os.getenv('JORMUNGANDR_INSTANCES_DIR', '/etc/jormungandr.d')
 
 # Start the thread at startup, True in production, False for test environments
 START_MONITORING_THREAD = True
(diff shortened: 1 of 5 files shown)


antoine-de (approved):
great!

So you think we are ready to loose the support for `.ini` files ? 
we need to at least add comment in the changelog for this breaking change.

we also need to document a bit this configuration, but I don't really know where we should do this

source/jormungandr/jormungandr/default_settings.py
```diff
@@ -1,35 +1,40 @@
...
-HTTP_BASIC_AUTH_REALM = 'Token Required'
+HTTP_BASIC_AUTH_REALM = os.getenv('JORMUNGANDR_HTTP_BASIC_AUTH_REALM', 'Token Required')
 
 from jormungandr.logging_utils import IdFilter
 
+log_level = os.getenv('JORMUNGANDR_LOG_LEVEL', 'DEBUG')
+log_format = os.getenv('JORMUNGANDR_LOG_FORMAT', '[%(asctime)s] [%(request_id)s] [%(levelname)5s] [%(process)5s] [%(name)10s] %(message)s')
+
 # logger configuration
 LOGGER = {
```
antoine-de, line 32: don't you also want to give a way to overload all the LOGGER ? (but maybe we can do this if a need arise)

kinnou02: It will require a json, and I want to avoid having json in a environment variable, it's ugly.
We almost never update the logger configuration

antoine-de (approved):
great!

we just need to remember to add the breaking change of dropping the old ini configuration files in the documentation

TeXitoi:
> CanalTP/utils#47 need to be merde before this one.

Yeah, that'll become a pill of shit :rofl:

PR #1952, opened 2017-01-31
Jormungandr: limit first section radius using direct path duration in experimental

TeXitoi:
For 20 rue Hector Malot to Gare de Lyon using car as first section mode and valhalla (friendly case for this optimization):
 - before: 3.3s
 - after: 1.6s

We limit the radius only on the first section because having a bound on the last section is more complicated (we need to know if we have one direct path for each mode and, if this is the case, the bound is the maximum of every duration of the direct paths).

In practice, we use long street network section at the begining as we never put car and bike at the end.

Request: http://localhost:5001/v1/coverage/valhalla/journeys?from=2.37715%3B48.846781&to=stop_area%3A0%3ASA%3A8768600&first_section_mode%5B%5D=car&debug=true&_override_scenario=experimental&

### DIFF
--- a/source/jormungandr/jormungandr/scenarios/experimental.py
+++ b/source/jormungandr/jormungandr/scenarios/experimental.py
@@ -66,15 +66,11 @@ class SectionSorter(object):
             return -1 if a.end_date_time < b.end_date_time else 1
 
 
-def get_max_fallback_duration(request, mode):
-    if mode == 'walking':
-        return request['max_walking_duration_to_pt']
-    if mode == 'bss':
-        return request['max_bss_duration_to_pt']
-    if mode == 'bike':
-        return request['max_bike_duration_to_pt']
-    if mode == 'car':
-        return request['max_car_duration_to_pt']
+def get_max_fallback_duration(request, mode, dp_durations={}):
+    if mode in ['walking', 'bss', 'bike', 'car']:
+        max_duration = request['max_{}_duration_to_pt'.format(mode)]
+        dp_duration = dp_durations.get(mode, max_duration)
+        return min(max_duration, dp_duration)
     raise ValueError('unknown mode: {}'.format(mode))
 
 
@@ -82,6 +78,14 @@ def make_direct_path_key(dep_mode, orig_uri, dest_uri, datetime, clockwise, reve
     # datetime is not taken into consideration because we assume that
     # a direct path from A to B remains the same even the departure time are different
     return dep_mode, orig_uri, dest_uri, clockwise, reverse_sections
+
+
+def make_direct_path_duration_by_mode(fallback_direct_path):
+    res = {}
+    for key, dp in fallback_direct_path.items():
+        if dp.journeys:
+            res[key[0]] = dp.journeys[0].durations.total
+    return res
 
 
 def get_direct_path_if_exists(direct_path_pool, mode, orig_uri, dest_uri, datetime, clockwise, reverse_sections):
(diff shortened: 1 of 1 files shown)


kadhikari (approved):

source/jormungandr/jormungandr/scenarios/experimental.py
```diff
@@ -66,15 +66,11 @@ def __call__(self, a, b):
...
-    if mode == 'walking':
-        return request['max_walking_duration_to_pt']
-    if mode == 'bss':
-        return request['max_bss_duration_to_pt']
-    if mode == 'bike':
-        return request['max_bike_duration_to_pt']
-    if mode == 'car':
-        return request['max_car_duration_to_pt']
+def get_max_fallback_duration(request, mode, dp_durations={}):
+    if mode in ['walking', 'bss', 'bike', 'car']:
```
xlqian, line 70: 👍

source/jormungandr/jormungandr/scenarios/experimental.py
```diff
@@ -84,6 +80,14 @@ def make_direct_path_key(dep_mode, orig_uri, dest_uri, datetime, clockwise, reve
     return dep_mode, orig_uri, dest_uri, clockwise, reverse_sections
 
 
+def make_direct_path_duration_by_mode(fallback_direct_path):
+    res = {}
+    for key, dp in fallback_direct_path.items():
+        if dp.journeys:
+            res[key[0]] = dp.journeys[0].durations.total
```
xlqian, line 87: what's key[0]?

source/jormungandr/jormungandr/scenarios/experimental.py
```diff
@@ -337,7 +341,7 @@ def get_routing_matrix_futures(self, origin, destination, origins_places_crowfly
 
         return origin_futures, destination_futures
 
-    def get_crowfly_futures(self, origin, destination):
+    def get_crowfly_futures(self, origin, destination, dp_durations):
```
xlqian, line 344: i think it would be useful to add some comments for dp_durations as a future reminding

source/jormungandr/jormungandr/scenarios/experimental.py
```diff
@@ -623,10 +627,32 @@ def call_kraken(self, request_type, request, instance, krakens_call):
 
         worker = AsyncWorker(instance, krakens_call, request)
 
+        resp = []
+
+        # Now we compute the direct path with all requested departure mode
+        # their time will be used to initialized our PT calls
```
xlqian, line 633: update for comments: "their time will also be used to optimize the computation of NM"

source/jormungandr/jormungandr/scenarios/experimental.py
```diff
@@ -623,10 +627,32 @@ def call_kraken(self, request_type, request, instance, krakens_call):
...
+            resp_key, resp_direct_path = future.get()
+            g.fallback_direct_path[resp_key] = resp_direct_path
+            if resp_direct_path.journeys:
+                resp_direct_path.journeys[0].internal_id = str(generate_id())
+                resp.append(resp_direct_path)
+
         if request.get('max_duration', 0):
+            direct_path_duration_by_mode = make_direct_path_duration_by_mode(g.fallback_direct_path)
+
             # Get all stop_points around the requested origin within a crowfly range
```
xlqian: Sorry for bothering, i think we can actually call `get_max_fallback_duration` immediately after the computation of the "general" direct_path, instead of calling it repeatedly in `get_crowfly_futures` and `get_routing_matrix_futures` (we can do it later, i'm ok with that )

xlqian (approved):
👍

PR #1953, opened 2017-01-31
kraken: optimization of direct path

kadhikari:
* For direct path of 10 to 20minutes in a query the processing time is reduced from 14ms to 6 ms.
* For direct path of 11 to 22 minutes in another query the processing time is reduced from "slow request" (> 1000 ms) to  between 42ms and 217ms  

*concerned ticket: http://jira.canaltp.fr/browse/ITI-373

### DIFF
--- a/source/georef/street_network.h
+++ b/source/georef/street_network.h
@@ -141,6 +141,7 @@ struct PathFinder {
     void init(const type::GeographicalCoord& start_coord, nt::Mode_e mode, const float speed_factor);
 
     void start_distance_dijkstra(const navitia::time_duration& radius);
+    void start_distance_and_target_dijkstra(const navitia::time_duration& radius, const std::vector<vertex_t>& destinations);
 
     /// compute the reachable stop points within the radius
     routing::map_stop_point_duration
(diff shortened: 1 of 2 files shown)


source/georef/street_network.h
```diff
@@ -377,5 +378,21 @@ struct target_unique_visitor : public boost::dijkstra_visitor<> {
...
+        distance_visitor(max_dur, dur), target_all_visitor(destinations){}
+    template <typename graph_type>
+    void finish_vertex(vertex_t u, const graph_type& g){
+        target_all_visitor::finish_vertex(u, g);
+    }
+
+    template <typename G>
+    void examine_vertex(typename boost::graph_traits<G>::vertex_descriptor u, const G& g) {
+        distance_visitor::examine_vertex(u, g);
+    }
```
TeXitoi, line 394: The inheritance is not enough?

TeXitoi (commented):
:+1:

source/georef/street_network.h
TeXitoi: OK it's needed.

TeXitoi (approved):

xlqian (approved):

PR #1963, opened 2017-02-03
Jormun: Call bragi with type

azime:


source/jormungandr/tests/bragi_autocomplete_tests.py
```diff
@@ -90,7 +95,8 @@ def test_autocomplete_call(self):
                 }, 200)
         })
         with mock.patch('requests.get', mock_requests.get):
-            response = self.query_region('places?q=bob&pt_dataset=main_autocomplete_test')
+            response = self.query_region("places?q=bob&pt_dataset=main_autocomplete_test&type[]=['stop_area', "
```
antoine-de: hum I think we do this more:
`type[]=stop_area&type[]=adress&...`

datanel:
does the documentation need to be updated?

TeXitoi:
@datanel no that's isofunctional

source/jormungandr/tests/bragi_autocomplete_tests.py
```diff
@@ -118,8 +124,13 @@ def test_autocomplete_call_override(self):
         """"
         test that the _autocomplete param switch the right autocomplete service
         """
+        url = 'https://host_of_bragi/autocomplete'
+        kwargs = {'params': {u'q': u'bob', u'type[]': [u"['stop_area', 'address', 'poi', 'administrative_region']"],
```
datanel: If we don't specify a type, `type[]=stop_area&type[]=address&type[]=poi&type[]=administrative_region` is sent to bragi

but administrative_region = city in bragi

do we need to check bragi types as we do here https://github.com/CanalTP/navitia/blob/dev/source/jormungandr/jormungandr/autocomplete/kraken.py#L64 ?

TeXitoi: It must be isofunctional, thus:
 - by default, type[]=stop_area&type[]=address&type[]=poi&type[]=administrative_region
 - the type[] must be checked in {stop_area, stop_point, address, poi, administrative_region}
 - a type "conversion" may be needed to request bragi
 - for stop points, 2 possibilities:
    - or we silently remove stop point in the request to bragi
    - or we error with unsupported object

antoine-de: I'll vote for the conversion before sending to bragi and a warning section added to the response if asked for stop_points. It's a bit like silently removing the stoppoints but at the the caller can have a feedback. If we error on this we might break integrations :confounded:

source/jormungandr/tests/bragi_autocomplete_tests.py
datanel: `type[]=stop_area&type[]=adress` does not work in the tests as `requests` expects to have `type[]=['stop_area', adress']`

but `type[]=stop_area&type[]=adress` works perfectly from a browser

source/jormungandr/jormungandr/autocomplete/geocodejson.py
```diff
@@ -55,6 +57,22 @@ def make_params(self, request, instance):
         if request.get("from"):
             params["lon"], params["lat"] = self.get_coords(request["from"])
 
+        if request.get("type[]"):
+            types = []
+            for type in request.get("type[]"):
+                if type == 'stop_point':
```
antoine-de, line 61: can you add a log telling this type is not handled ?

source/jormungandr/jormungandr/autocomplete/geocodejson.py
```diff
@@ -55,6 +57,22 @@ def make_params(self, request, instance):
...
+                    continue
+
+                if type not in pb_type:
```
antoine-de: isn't this checked in the api ?

source/jormungandr/jormungandr/autocomplete/geocodejson.py
```diff
@@ -55,6 +57,22 @@ def make_params(self, request, instance):
...
+                    abort(422, message="{} is not an acceptable type".format(type))
+
+                if type == 'administrative_region':
```
antoine-de: cna you add a comment explaining this ?

antoine-de (approved):
great tests!

antoine-de (approved):

PR #1964, opened 2017-02-06
Doc: since/until parameters on disruptions

kinnou02:


kinnou02:
It would be nice if this can be merge before tomorrow morning :p

PR #1969, opened 2017-02-08
ed: fix timezone bug when there is no dst

### LINKED ISSUE #1968: UTC Time zone without Daylight Saving Time (DST)
I have a super simple GTFS[0] which has as time zone "America/Costa_Rica" defined at agency.txt.  After importing the file on navitia, running my own server and using navitia docker-compose, I got some weird times results, for example a departure that should be at 05:45 is appearing at 23:51 of the day before. 

In the IRC I got help and yesterday we make some tests changing the GTFS Timezone and it seems to work with timezones that do have Daylight Saving Time[1],  I successfully tested using America/Mexico_City (UTC-6), America/Sao_Paulo(UTC-3), Europe/Paris (UTC+1). But when using Timezones that do not use DST is not working, for example: America/Costa_Rica (UTC-6), America/Guatemala (UTC-6) or America/Jujuy (UTC-3).

[0] Super simple GTFS with only 1 route available at: ic-itcr.ac.cr/~jgutierrez/incofer/cr-incofer_prueba.zip
[1] When I say Timezones that do have DST I mean that are defined here: https://github.com/CanalTP/navitia/blob/dev/source/ed/connectors/tz_db_wrapper.cpp

kinnou02:
for timezone without dst the offset was in minutes in place of seconds,
fix #1968

### DIFF
--- a/source/ed/tests/gtfsparser_test.cpp
+++ b/source/ed/tests/gtfsparser_test.cpp
@@ -785,6 +785,31 @@ BOOST_AUTO_TEST_CASE(split_over_dst_one_year_new_york) {
     BOOST_CHECK_EQUAL(summer_periods.at(1).last(), "20160730"_d);
 }
 
+/*
+ * Configuration in tz_db_wrapper.h :
+ * America/Costa_Rica,CST,CST,,,-06:00:00,+00:00:00,,,,+00:00:00
+ * there is no dst in costa rica, offset is always -6 hours
+ */
+BOOST_AUTO_TEST_CASE(split_over_dst_costa_rica) {
+    ed::connectors::GtfsData gtfs_data;
+    auto tz_pair = gtfs_data.tz.get_tz("America/Costa_Rica");
+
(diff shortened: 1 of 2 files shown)


TeXitoi (approved):

source/ed/tests/gtfsparser_test.cpp
```diff
@@ -785,6 +785,31 @@ BOOST_AUTO_TEST_CASE(split_over_dst_one_year_new_york) {
...
+    boost::gregorian::date_period vj_validity_period {"20120201"_d, "20121105"_d};
+
+    ed::EdTZWrapper tz_wrapper {tz_pair.first, tz_pair.second};
+
+    auto split_periods = tz_wrapper.split_over_dst(vj_validity_period);
+
+    BOOST_REQUIRE_EQUAL(split_periods.size(), 1);
+
+    const auto winter_paris_utc_shift = -6 * 60 * 60;
+    const auto& period = split_periods.at(winter_paris_utc_shift);
```
pbougue: `costa_rica_utc_shift`

pbougue (approved):
:+1:

PR #1970, opened 2017-02-08
RFC: line section

antoine-de:
Thanks to @eturck we know how to impact a `line_section` now we needs to
know how to show them :grin:

What do you think about it ?

Note:
here is a try on a RFC to openly track the discussion on the model of the line sections. we'll see if that's useful.

[rendered ](https://github.com/antoine-de/navitia/blob/rfc_line_sections/documentation/rfc/line_sections.md)

TODO:
- [ ] wait for code, to update this RFC accordingly

### DIFF
--- a/documentation/rfc/line_sections.md
+++ b/documentation/rfc/line_sections.md
@@ -98,7 +98,7 @@ We display only the stoppoints impacted by the line section
 
 `/stop_points/D_1_of_route_1_of_line_1`  --> YES
 
-`/stop_points/E_1_of_route_1_of_line_1`  --> NO
+`/stop_points/E_1_of_route_1_of_line_1`  --> YES
 
 `/stop_points/F_1_of_route_1_of_line_1`  --> NO
 
@@ -187,6 +187,5 @@ struct LineSection {
 ```
 
 but a line section disruption will be stored in:
-* the line
-* each vehicle_journeys of each routes
+* each impacted vehicle_journeys of each routes
 * each stoppoints in [start_point, end_point] for each vj for each routes
--- a/documentation/rfc/line_sections.md
+++ b/documentation/rfc/line_sections.md
@@ -6,6 +6,8 @@ Here's a proposal on how to handle them:
 
 # representation
 
+## chaos like
+
 I think the easiest way to show them would be as a classic pt_object
 ```javascript
 {
@@ -38,11 +40,15 @@ I think the easiest way to show them would be as a classic pt_object
 ```
 
 with this we need to choose whether to dump the stops/line objects (in depth 0) or just make a link
-I think the depth 0 is more coherent with the rest of the API
 
-This would imply that we provide a `/line_sections/<uri>` api to be coherent, but I think we can skip the whole ptref stuf (`/stop_areas/<uri>/line_sections`) as the `line_section` is not really a referential object
+I think dumping the sub objects as depth 0 is more coherent with the rest of the API
 
-another option would be to do:
+This would imply that we provide a `/line_sections/<uri>` api to be coherent, but I think we can skip the whole 
+ptref stuf (`/stop_areas/<uri>/line_sections`) as the `line_section` is not really a referential object
+
+## custom line section
+
+Another option would be to add a distinct field (apart from the `impacted_objects`:
 
 ```javascript
 {
@@ -71,7 +77,154 @@ another option would be to do:
   "id":"002def5c-dc76-11e6-8542-005056a47b86"
 }
 ```
-This would act the fact that the line section are different than the other objects (thus we can skip the `/line_sections` api), but I don't like adding a custom `impacted_line_sections` as it is not coherent with the other impacted objects (and it does not compose well if we want to impact more than a line section).
+This would act the fact that the line section are different than the other objects (thus we can skip the `/line_sections` api), 
+but I don't like adding a custom `impacted_line_sections` as it is not coherent with the other impacted objects 
+and more important it does not compose well if we want to impact more than a line section.
+
+## hybrid custom line section
+
+There could be a hybrid approach, impacting the line as the main object and adding an `impacted_line_sections` 
+(a bit like how delays are handled with `impacted_objects` being a `trip` and `impacted_stops` being the stop_times of this trip)
+
+
+```javascript
+{
+  "status":"active",
+  // lots of additional fields
+  "impacted_objects":[
+  {
+     "pt_object": {
+        "embedded_type": "line",
+        "line": {
+          // a line
+        }
+        "id":"id_of_the_line",
+     }
+  }
+  ]
+  "impacted_line_sections":[
+  {
+      "line_section":{
+          "from": {
+              // a stop area
+          },
+          "to": {
+              // a stop area
+          },
+          "line": {
+              // the line
+          }
+          // routes ?
+      },
+  },
+  ]
+
+```
+
+With this we can skip the `/line_sections` api as it is not treated as a ptobject. I find it also a bit more coherent.
+
+The downside is that the integration is harder, the user needs to check wheter or not a field `impacted_line_sections` is present to treat the impacted line differently.
+
+As several objects can be impacted it arise several other problems:
+ * is a bit tricky to link the line to its line section
+ * if a impact is link to the "line_1" AND to a line section of this line (I don't see the point of doing this, but it's technically possible), it's difficult to represent this
+
+
+## hybrid custom line section #2
+
+It's the same as above, but @pbougue proposed to add a `non_pt_objects` aside the `impacted_objects`
+
+I don't really like this as it will either break the delay api (`impacted_stops`), or have another field aside.
+
+## route sections
+
+The real impacted objects on the line_sections are routes's sections
+
+Thus we could represent the line section as a routes section
+
+```javascript
+{
+  "status":"active",
+  // lots of additional fields
+  "impacted_objects":[
+    {
+      "pt_object":{
+        "embedded_type":"route_section", // <--- here we add a new embedded_type
+        "line_section":{
+            "from": {
+              // a stop area
+            },
+            "to": {
+              // a stop area
+            },
+            "routes": [{
+               // one route
+            }, {
+               // another route
+            }]
+            // to ease integration we could also add the impacted lines here
+        },
+        "quality":0,
+        "id":"id_of_the_route_section",
+        "name":"Route section"
+      }
+    }
+  ]
+  "id":"002def5c-dc76-11e6-8542-005056a47b86"
+}
+```
+
+I find it less constrained by how the disruption is created in the TR ui but one major problem with this is that the client asking us to work 
+on line section can a lot's of impacted routes (+100), thus it can flood the response and cause major performance problems
+
+## hybrid route sections
+
+The same hybrid approach can be used for route sections
+
+```javascript
+{
+  "status":"active",
+  // lots of additional fields
+  "impacted_objects":[
+  {
+     "pt_object": {
+        "embedded_type": "route",
+        "route": {
+          // a route
+        }
+        "id":"id_of_the_route",
+     }
+  },
+  {
+     "pt_object": {
+        "embedded_type": "route",
+        "route": {
+          // another route
+        }
+        "id":"id_of_the_route_2",
+     }
+  }
+  ]
+  "impacted_route_sections":[
+  {
+      "route_section":{
+          "from": {
+              // a stop area
+          },
+          "to": {
+              // a stop area
+          },
+          "route": {
+              // the line
+          }
+      },
+  },
+  ]
+
+```
+
+As there can be many many routes I don't think this approach can work
+
 
 # when to display the impact on a line section
 
--- a/documentation/rfc/line_sections.md
+++ b/documentation/rfc/line_sections.md
@@ -105,7 +105,7 @@ We display only the stoppoints impacted by the line section
 
 `/routes/route_2_of_the_line_1`  --> NO
 
-`/lines/line_1`  -->  YES
+`/lines/line_1`  -->  NOT_SPECIFIED
 
 `/lines/line_2`  -->  NO
 
--- a/documentation/rfc/line_sections.md
+++ b/documentation/rfc/line_sections.md
@@ -249,7 +249,7 @@ examples:
 
 this is the navitia call and YES means we display the disruption, NO means we do not display it
 
-`/routes/route_1_of_the_line_1` --> YES (? not sure about this one if the line section is represented as linked to the line)
+`/routes/route_1_of_the_line_1` --> YES
 
 `/stop_areas/A`  --> NO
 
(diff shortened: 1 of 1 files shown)


documentation/rfc/line_sections.md
```diff
@@ -0,0 +1,165 @@
...
+Here's a proposal on how to handle them:
+
+# representation
+
+I think the easiest way to do show them would be as a classic pt_object
+```javascript
+{
+  "status":"active",
+  // lots of additional fields
+  "impacted_objects":[
```
kinnou02: I vote for this one

documentation/rfc/line_sections.md
```diff
@@ -0,0 +1,165 @@
...
+`/stop_areas/A`  --> NO
+`/stop_areas/B`  --> NO
+`/stop_areas/C`  --> YES
+`/stop_areas/D`  --> YES
+`/stop_areas/E`  --> YES
+`/stop_areas/F`  --> NO
+
+`/routes/route_2_of_the_line_1`  --> NO
+
+`/lines/line_1`  -->  YES (? not sure about this one if we represent the line section as linked to 2 stops and a list of routes)
```
kinnou02: This **must** work (mostly on `/lines/line_1/disruptions`)

documentation/rfc/line_sections.md
```diff
@@ -0,0 +1,165 @@
+The line sections is a way to impact the some routes between 2 stops areas
```
kadhikari: ".. to impact the some routes .." -> ".. to impact some routes .."

documentation/rfc/line_sections.md
kadhikari: A good choice. It's same as in chaos..

documentation/rfc/line_sections.md
```diff
@@ -0,0 +1,165 @@
...
+`/lines/line_2`  -->  NO
+
+If we decide to display the line section disruptions based on the last object of the query it will lead to some difficult to understand behaviour:
+
+`/stop_areas/A/routes/route_1_of_the_line_1`  --> YES
+`/stop_areas/C/routes/route_2_of_the_line_1`  --> NO
+
+`/networks/network_of_the_route_1 `  --> NO
+
+`/stop_points/C_1`  --> NO
```
kadhikari: What if stop_point C_1 is related to route_1_of_the_line_1 ?

documentation/rfc/line_sections.md
```diff
@@ -0,0 +1,165 @@
...
+I don't think we need to do some stuff here, the same rules as the common cases can be used.
+
+## journeys
+
+If a journey use a VJ of an impacted route on at least one stop point impacted by the line section we display it.
+
+
+# modelisation
+
+TODO, we need to first agree on the other questions before doing this
```
kadhikari: What i understood: In the API /disruption we display impact on object directly related to line section whereas in /traffic_reports if any child object is impacted
Example:
/networks/network_of_the_route_1/disruptions --> NO
/networks/network_of_the_route_1/traffic_reports  --> YES

TeXitoi:
# representation

I prefer the second solution, but with no veto on the first.

I prefer it because a line section is not a pt_object for me.

# when to display the impact on a line section

A line section impacts vj on a list of stop points. Thus, I think that the disruption must be present on the impacted vj.

Then, it should also be present on the impacted stop points.

That's the technical impacted object. Now, functionally, we want to impact lines and stop areas. Thus I propose that the impact touch all the objects between them : 
 - vj -> route -> line
 - stop point -> stop area

Thus, our disruption on a line section is present on the modified vj, on the stop points removed from a vj, on the route and the line of the vj, and on the stop areas of the precited stop points.

The path of the request must not be taken into account to display or not the disruption, they are just a mean to filter objects.

So, to conclude, I'm OK with the proposition, except than I propose that the disruption is also present on the vj, stop point and route.

documentation/rfc/line_sections.md
```diff
@@ -0,0 +1,165 @@
...
+            "from": {
+              // a stop area
+            },
+            "to": {
+              // a stop area
+            },
+            "line": {
+               // the line
+            }
+            // for the moment the idea is not to dump the associated routes
```
pbougue: And we would add it and just check that it's on the correct line.

documentation/rfc/line_sections.md
```diff
@@ -0,0 +1,165 @@
...
+        "name":"Line section"
+      }
+    }
+  ]
+  "id":"002def5c-dc76-11e6-8542-005056a47b86"
+}
+```
+
+with this we need to choose whether to dump the stops/line objects (in depth 0) or just make a link
+I think the depth 0 is more coherent with the rest of the API
```
pbougue: I guess you mean dump `impacted_objects` in depth 0.

documentation/rfc/line_sections.md
```diff
@@ -0,0 +1,165 @@
...
+            }
+            // for the moment the idea is not to dump the associated routes
+        },
+        "quality":0,
+        "id":"id_of_the_line_section",
+        "name":"Line section"
+      }
+    }
+  ]
+  "id":"002def5c-dc76-11e6-8542-005056a47b86"
```
pbougue: this hash is a good thing.
though we might have to think about stability over reload, so maybe hash a key "line#from#to"

documentation/rfc/line_sections.md
```diff
@@ -0,0 +1,165 @@
...
+# representation
+
+I think the easiest way to do show them would be as a classic pt_object
+```javascript
+{
+  "status":"active",
+  // lots of additional fields
+  "impacted_objects":[
+    {
+      "pt_object":{
```
pbougue: A third way could be to have another field `"non_pt_object"[ { "embedded_type":"line_section"... } ]` (or even to have a `line_sections` field aside).
The bright side is we know it's not PT (not possible to query), and we can compose.
The downside is that it can break few things (navitia and intergration) to offer the possibility of no `pt_object`.

Or we have the line as `pt_object` and like for sncf delays (it concerns a trip, not a line) we have an outside field describing what happens on this line (so the from/to).

documentation/rfc/line_sections.md
antoine-de: In my idea C_1 belong to the route_1, but we do not display the disruption

documentation/rfc/line_sections.md
antoine-de: yes that's what is done for the other objects for the moment

antoine-de:
Note: @TeXitoi for the specific comments I think it's easier if you comment on the PR it's easier to reply to them :wink: 

> I prefer it because a line section is not a pt_object for me.

I do agree, but It does not compose well at all if a disruption is on a full line and another line's section is added, thus I'll vote for the 1st model

> So, to conclude, I'm OK with the proposition, except than I propose that the disruption is also present on the vj, stop point and route.

while I agree it seems better that is not what is done for the other objects :disappointed: 
If we impact a line, we'll not find the impact on the vehicle_journeys of this line.
I think it's like this because nobody gets the disruptions from ptobjects :grin:  

We can change this but it widen a bit the scope of the modification :roll_eyes:

documentation/rfc/line_sections.md
antoine-de: shouldn't we also dump the routes ?

documentation/rfc/line_sections.md
```diff
@@ -0,0 +1,165 @@
...
+`/routes/route_1_of_the_line_1` --> YES (? not sure about this one if the line section is represented as linked to the line)
+
+`/stop_areas/A`  --> NO
+`/stop_areas/B`  --> NO
+`/stop_areas/C`  --> YES
+`/stop_areas/D`  --> YES
+`/stop_areas/E`  --> YES
+`/stop_areas/F`  --> NO
+
+`/routes/route_2_of_the_line_1`  --> NO
```
pbougue: as long as there is no route in sections, I'd say it's better to display on all routes than none.
We can also process if the from/to appears in that order on the route, but not sure it's easy, not sure it's urgent too.

TeXitoi:
> while I agree it seems better that is not what is done for the other objects

That's quite different. A disruption on a line is a disruption on a line.

A disruption on a line section is
  - technically a disruption on portions of vj
 - functionnally a disruption on an interval on a line

Thus, what to do with the object in between. That's seems strange to not impact vj and stop points as this is the object impacted technically, and that seems strange to not impact the line and stop areas as this is the objects impacted functionnally. Then, between the two, you have the routes.

documentation/rfc/line_sections.md
pbougue: I'd say yes here (because I want it linked to stop_point, but maybe not)

documentation/rfc/line_sections.md
```diff
@@ -0,0 +1,165 @@
...
+
+If we decide to display the line section disruptions based on the last object of the query it will lead to some difficult to understand behaviour:
+
+`/stop_areas/A/routes/route_1_of_the_line_1`  --> YES
+`/stop_areas/C/routes/route_2_of_the_line_1`  --> NO
+
+`/networks/network_of_the_route_1 `  --> NO
+
+`/stop_points/C_1`  --> NO
+`/vehicle_journeys/vj1_of_route_1_of_line_1`  --> NO
```
pbougue: that's a bigger problem (and maybe trips also)

documentation/rfc/line_sections.md
```diff
@@ -0,0 +1,165 @@
...
+
+`/routes/route_2_of_the_line_1`  --> NO
+
+`/lines/line_1`  -->  YES (? not sure about this one if we represent the line section as linked to 2 stops and a list of routes)
+`/lines/line_2`  -->  NO
+
+If we decide to display the line section disruptions based on the last object of the query it will lead to some difficult to understand behaviour:
+
+`/stop_areas/A/routes/route_1_of_the_line_1`  --> YES
+`/stop_areas/C/routes/route_2_of_the_line_1`  --> NO
```
pbougue: I'd say those are ok, as we have the links in routes/lines/stops.

antoine-de:
Maybe we don't care about it, but I'm a bit bothered by the `LineSection` current model.
It's called LineSection because at first what could be impacted in TR UI was really a line's section, but since we want to filter a some routes of the line isn't what we really want a `RoutesSection` ?

this way we could technically represent a routes sections as a list of routes and 2 stop areas.
would the clients be bothered by it being:
```javascript
        "route_section":{
            "from": {
              // a stop area
            },
            "to": {
              // a stop area
            },
            "routes": [{
               // one route
            }, {
            // another route
            }
            ]
            // to ease integration we would add a "lines" section computed but I'm not fond of this
        },
```
?

documentation/rfc/line_sections.md
```diff
@@ -0,0 +1,165 @@
...
+`/stop_areas/B/traffic_reports`  --> NO
+`/stop_areas/C/traffic_reports`  --> YES
+
+`/lines/line_1/traffic_reports ` -->  YES
+`/lines/line_2/traffic_reports`  -->  NO
+
+`/routes/route_1_of_the_line_1/traffic_reports`  --> YES
+`/routes/route_2_of_the_line_1/traffic_reports`  --> NO
+
+`/stop_areas/A/routes/route_1_of_the_line_1/traffic_reports`  --> YES (I think NO would be better it but this will be quite difficult for not that much gain)
```
pbougue: agreed

documentation/rfc/line_sections.md
```diff
@@ -0,0 +1,165 @@
...
+`/stop_areas/C/routes/route_2_of_the_line_1/traffic_reports`  --> NO
+
+`/networks/network_of_the_route_1/traffic_reports `  --> YES
+`/stop_points/C_1/traffic_reports`  --> YES since it's stop_area is part of an impacted line section
```
pbougue: you mix sp and sa, I think it's worth describe both (and I'd say YES for saC, spC1 and NO for spC2)

documentation/rfc/line_sections.md
```diff
@@ -0,0 +1,165 @@
...
+
+`/vehicle_journeys/vj1_of_route_1_of_line_1/traffic_reports`  --> NO (It would be better to display it, but we do not display other disruptions there, would it be worth it to change this ?)
```
pbougue: I'd say YES, as we change it, it's one of the most impacted object.

documentation/rfc/line_sections.md
```diff
@@ -0,0 +1,165 @@
...
+
+
+## schedules
+
+I don't think we need to do some stuff here, the same rules as the common cases can be used.
```
pbougue, line 169: So depends on the last object... I agree.

documentation/rfc/line_sections.md
pbougue: This summary of differences clarifies the doc and it would be nice to add it somewhere, to understand better differences between disruptions vs traffic_reports.

documentation/rfc/line_sections.md
antoine-de: hum not sure to understand how you want me to  describe them more

documentation/rfc/line_sections.md
antoine-de: this hash is the id from Chaos

documentation/rfc/line_sections.md
antoine-de: that's indeed 2 other options

antoine-de:
damn corrected some typo and added some lints so some still valid comments appear as outdated :crying_cat_face:

documentation/rfc/line_sections.md
pbougue: up

documentation/rfc/line_sections.md
pbougue: add in doc the 2 examples
`/stop_points/C_2/traffic_reports`
`/stop_areas/C/traffic_reports`

documentation/rfc/line_sections.md
pbougue: For me it's not a good idea to rely on external module for uri.
But displaying chaos's id is another thing...
Pretty sure we have to discuss about communication with chaos.

kinnou02:
A "RouteSection" won't be usable in the case that matter to us right now, 120 routes that nobody except us know won't help anyone.

antoine-de:
I wrote down the different propositions that arose, I think we need to discuss them (and do a poll ?)

documentation/rfc/line_sections.md
```diff
@@ -0,0 +1,192 @@
...
+struct LineSection {
+    Line* line;
+    StopArea* start_point;
+    StopArea* end_point;
+    std::vector<Route*> routes;
+};
+```
+
+but a line section disruption will be stored in:
+* the line
```
pbougue: I vote NO to that if possible (cleaner functional model).

kinnou02 (approved):

documentation/rfc/line_sections.md
```diff
@@ -0,0 +1,192 @@
...
+* each vehicle_journeys of each routes
```
TeXitoi: each **impacted** vehicle_journeys

documentation/rfc/line_sections.md
```diff
@@ -0,0 +1,191 @@
...
+`/stop_points/C_1_of_route_1_of_line_1`  --> YES
+
+`/stop_points/C_2_of_route_2_of_line_1`  --> NO
+
+`/stop_points/C_3_of_line_2`  --> NO
+
+
+`/stop_points/D_1_of_route_1_of_line_1`  --> YES
+
+`/stop_points/E_1_of_route_1_of_line_1`  --> NO
```
TeXitoi: YES?

antoine-de: you're right, my bad, I corrected this

PR #1971, opened 2017-02-08
Jormungandr: POC of a SIRI lite connector for realtime

kinnou02:
This connector is based on the STIF implementation of StopMonitoring from SIRI 2.0 as defined by
the local agreement IDF 2.4
(http://www.normes-donnees-tc.org/wp-content/uploads/2014/05/Profil_Siri_IDF_V2-4-STIF-20130712.pdf)

### DIFF
--- a/source/jormungandr/jormungandr/realtime_schedule/siri_lite.py
+++ b/source/jormungandr/jormungandr/realtime_schedule/siri_lite.py
@@ -1,0 +1,148 @@
+# coding=utf-8
+
+# Copyright (c) 2001-2016, Canal TP and/or its affiliates. All rights reserved.
+#
+# This file is part of Navitia,
+#     the software to build cool stuff with public transport.
+#
+# Hope you'll enjoy and contribute to this project,
+#     powered by Canal TP (www.canaltp.fr).
+# Help us simplify mobility and open public transport:
+#     a non ending quest to the responsive locomotion way of traveling!
+#
+# LICENCE: This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU Affero General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+# GNU Affero General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with this program. If not, see <http://www.gnu.org/licenses/>.
+#
+# Stay tuned using
+# twitter @navitia
+# IRC #navitia on freenode
+# https://groups.google.com/d/forum/navitia
+# www.navitia.io
+from jormungandr.realtime_schedule.realtime_proxy import RealtimeProxy
+from flask import logging
+import pybreaker
+import pytz
+import requests as requests
+from jormungandr import cache, app
+from jormungandr.schedule import RealTimePassage
+from datetime import datetime
+
+
+class SiriLite(RealtimeProxy):
+    """
+    class managing calls to a siri lite external service providing real-time next passages
+    """
+    def __init__(self, id, service_url, object_id_tag=None, timeout=10, **kwargs):
+        self.service_url = service_url
+        self.timeout = timeout  # timeout in seconds
+        self.rt_system_id = id
+        self.object_id_tag = object_id_tag if object_id_tag else id
+        self.breaker = pybreaker.CircuitBreaker(fail_max=app.config.get('CIRCUIT_BREAKER_MAX_SIRILITE_FAIL', 5),
+                                                reset_timeout=app.config.get('CIRCUIT_BREAKER_SIRILITE_TIMEOUT_S', 60))
+
+    def __repr__(self):
+        """
+         used as the cache key. we use the rt_system_id to share the cache between servers in production
+        """
+        return self.rt_system_id
+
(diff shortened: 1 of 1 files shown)


source/jormungandr/jormungandr/realtime_schedule/siri_lite.py
```diff
@@ -0,0 +1,143 @@
...
+    """
+    class managing calls to a siri lite external service providing real-time next passages
+    """
+    def __init__(self, id, service_url, object_id_tag=None, timeout=10, **kwargs):
+        self.service_url = service_url
+        self.timeout = timeout  # timeout in seconds
+        self.rt_system_id = id
+        self.object_id_tag = object_id_tag if object_id_tag else id
+        self.breaker = pybreaker.CircuitBreaker(fail_max=app.config.get('CIRCUIT_BREAKER_MAX_SIRILITE_FAIL', 5),
+                                                reset_timeout=app.config.get('CIRCUIT_BREAKER_CLEVERAGE_TIMEOUT_S', 60))
```
antoine-de: CIRCUIT_BREAKER_SIRILITE_TIMEOUT_S

source/jormungandr/jormungandr/realtime_schedule/siri_lite.py
```diff
@@ -0,0 +1,143 @@
...
+        line_id = route_point.fetch_line_id(self.object_id_tag)
+
+        if not stop_id or not line_id:
+            # one a the id is missing, we'll not find any realtime
+            logging.getLogger(__name__).debug('missing realtime id for {obj}: stop code={s}'.
+                                              format(obj=route_point, s=stop_id))
+            self.record_internal_failure('missing id')
+            return None
+
+        url = "{base_url}{stop_id}".format(base_url=self.service_url, stop_id=stop_id)
```
antoine-de, line 93: so the line_id is not used ?

source/jormungandr/jormungandr/realtime_schedule/siri_lite.py
```diff
@@ -0,0 +1,143 @@
...
+        return url
+
+    def _get_dt(self, datetime_str):
+        return pytz.utc.localize(datetime.strptime(datetime_str, "%Y-%m-%dT%H:%M:%S.%fZ"))
+
+    def _get_passages(self, route_point, resp):
+        logging.getLogger(__name__).debug('sirilite response: {}'.format(resp))
+
+        line_code = route_point.fetch_line_id(self.object_id_tag)
+        monitored_stop_visit = resp['siri']['serviceDelivery']['stopMonitoringDelivery']['monitoredStopVisit']
```
antoine-de: are the fields mandatory ? shouldn't you use `.get` ?

source/jormungandr/jormungandr/realtime_schedule/siri_lite.py
```diff
@@ -0,0 +1,143 @@
...
+        schedules = next((vj for vj in monitored_stop_visit if vj['monitoredVehicleJourney']['lineRef']['value'] == line_code), None)
```
antoine-de: schedules doesn't seems used

antoine-de (commented):
great work!

It would be nice to add test though (with mocks to have examples of for the siri response)

antoine-de:
so do you think it will be possible to filter to only have the route's passage ?

kinnou02:
there is no route in the response, and even so there isn't a shared referential of route(that is required to have meaningful id) , we could use the destination to find, then

source/jormungandr/jormungandr/realtime_schedule/siri_lite.py
kinnou02: actually we can't use the `lineRef` parameter, but we filter manually the response

source/jormungandr/jormungandr/realtime_schedule/siri_lite.py
kinnou02: what did you not understand in "POC"? :D

antoine-de: :mask:

source/jormungandr/jormungandr/realtime_schedule/siri_lite.py
antoine-de: ok can you thus add a comment there?

source/jormungandr/jormungandr/realtime_schedule/siri_lite.py
```diff
@@ -0,0 +1,143 @@
...
+    def _make_url(self, route_point):
+        """
+        The url returns something like a departure on a stop point
+        """
+
+        stop_id = route_point.fetch_stop_id(self.object_id_tag)
+        line_id = route_point.fetch_line_id(self.object_id_tag)
+
+        if not stop_id or not line_id:
+            # one a the id is missing, we'll not find any realtime
```
pbougue: `if` the id is missing ?

source/jormungandr/jormungandr/realtime_schedule/siri_lite.py
```diff
@@ -0,0 +1,143 @@
...
+        return url
+
+    def _get_dt(self, datetime_str):
+        return pytz.utc.localize(datetime.strptime(datetime_str, "%Y-%m-%dT%H:%M:%S.%fZ"))
+
+    def _get_passages(self, route_point, resp):
+        logging.getLogger(__name__).debug('sirilite response: {}'.format(resp))
+
+        line_code = route_point.fetch_line_id(self.object_id_tag)
+        monitored_stop_visit = resp.get('siri', {}).get('serviceDelivery', {}).get('stopMonitoringDelivery', {}).get('monitoredStopVisit', [])
```
pbougue: line length

antoine-de:
Jenkins, test this please

PR #1975, opened 2017-02-10
jormungandr: use user shape on instance autocomplete

patochectp:


pbougue (approved):

source/jormungandr/jormungandr/interfaces/v1/Places.py
```diff
@@ -351,6 +351,13 @@ def get(self, region=None, lon=None, lat=None):
         if args['disable_geojson']:
             g.disable_geojson = True
 
+        user = authentication.get_user(token=authentication.get_token(), abort_if_no_token=False)
+
+        shape = None
+        if user and user.shape:
+            shape = json.loads(user.shape)
```
antoine-de: I think you can remove the `shape` and directly do `args['shape'] =...`

antoine-de (commented):
It's not mandatory but if it's not too bothering it might be nice to add some integration tests with a mock user providing a shape.

source/jormungandr/jormungandr/interfaces/v1/Places.py
```diff
@@ -351,6 +351,10 @@ def get(self, region=None, lon=None, lat=None):
         if args['disable_geojson']:
             g.disable_geojson = True
 
+        user = authentication.get_user(token=authentication.get_token(), abort_if_no_token=False)
+
+        args['shape'] = json.loads(user.shape) if user and user.shape else None
```
datanel, line 357: why do you move this here?
it means kraken autocomplete will also receive shape in the request?

antoine-de: yes I think it's more coherent, the different autocompletes system might not know how to handle all the parameters but it's the same as the coordinate (to boost the results near the coordinate)

pbougue: Yep, and each autocomplete-connector manages if it passes the shape along

source/jormungandr/jormungandr/tests/utils_test.py
```diff
@@ -63,3 +66,108 @@ def get(self, url, *args, **kwargs):
...
+        """
+        Return True if user is blocked else False
+        """
+        return self._is_blocked
+
+def geojson():
+    return '{"type": "Feature", "geometry": ' \
+           '{"type": "Point", "coordinates": [102.0, 0.5]}, "properties": {"prop0": "value0"}}'
+
+user_in_db = {
```
antoine-de: I don't think this should be in utils it's specific to a test no ?

source/jormungandr/jormungandr/tests/utils_test.py
```diff
@@ -63,3 +66,108 @@ def get(self, url, *args, **kwargs):
...
+        "departure_board_test": {'ALL': True},
+        "empty_routing_test": {'ALL': True}
+    },
+}
+
+
+@contextmanager
+def user_set(app, user_name):
+    """
+    add user
```
antoine-de: I think it's worth putting a better comment

source/jormungandr/tests/bragi_autocomplete_tests.py
```diff
@@ -263,3 +264,40 @@ def http_get(url, *args, **kwargs):
...
+                assert False
+
+            def http_post(url, *args, **kwargs):
+                json = kwargs.pop('json')
+                assert json.get('geometry')
+                return MockResponse({}, 200, '{}')
+
+            with mock.patch('requests.get', http_get) as mock_method:
+                with mock.patch('requests.post', http_post) as mock_method:
+                    self.query('v1/coverage/main_routing_test/places?q=toto&_autocomplete=bragi')
```
antoine-de: can you add a boolean to assert (or a custom `mock` function) that the http_post method as bee used ?

source/jormungandr/jormungandr/tests/utils_test.py
pbougue: like what?
maybe remove it, but what would you say?

source/jormungandr/jormungandr/tests/utils_test.py
pbougue: hmm, we can talk about it, but to me it looks like it implies a copy/paste (or I just create a fake_user_base.py containing this part of utils file)

antoine-de: in your bragi test you only need 2 users no ?

I'm not a fan of putting test specific stuff in utils, I think it makes the tests less readable

antoine-de (approved):

pbougue:
Jenkins, test this please

PR #1979, opened 2017-02-14
Jormungandr: serialize poi_type in poi from geocodejson

TeXitoi:


antoine-de:
there are some mock bragi responses in the  tests, you should add this field to them

source/jormungandr/jormungandr/interfaces/v1/Places.py
```diff
@@ -199,6 +199,7 @@ def output(self, key, obj):
 
         lon, lat = get_lon_lat(obj)
         geocoding = obj.get('properties', {}).get('geocoding', {})
+        poi_type = geocoding.get('poi_types', [{}])[0]
```
xlqian: (hmmm, i'm not sure this is possible)
if `geocoding = {..., 'poi_types': [], ...}`,  
then `geocoding.get('poi_types', [{}])[0]` will crash...

xlqian: how about `geocoding.get('poi_types', [None])[0]`?

TeXitoi: That's the same, if poi_types is [], you'll have [], and boom.

TeXitoi: I have a correction for that.

TeXitoi:
@antoine-de test added

antoine-de (approved):

PR #1984, opened 2017-02-16
Kraken: disruption status is only based on 'now'

antoine-de:
the disruption 'status' field is now only based on the current_datetime (the datetime of the query)

some tests have been added to explain a bit the different cases


kinnou02 (approved):

TeXitoi (approved):
:+1:

PR #1986, opened 2017-02-16
Jormungandr: Geovelo management

pbougue:
first version without a real direct path

### DIFF
--- a/source/jormungandr/jormungandr/scenarios/experimental.py
+++ b/source/jormungandr/jormungandr/scenarios/experimental.py
@@ -278,12 +278,12 @@ def _get_duration(resp, place, mode, **kwargs):
 def _sn_routing_matrix(instance, origins, destinations, is_orig_center, mode, max_duration_to_pt, request, **kwargs):
     # When max_duration_to_pt is 0, there is no need to compute the fallback to pt, except if place is a stop_point or a
     # stop_area
-    center_isochron = origins[0] if is_orig_center else destinations[0]
-    places_isochron = destinations if is_orig_center else origins
+    center_isochrone = origins[0] if is_orig_center else destinations[0]
+    places_isochrone = destinations if is_orig_center else origins
     if max_duration_to_pt == 0:
         # When max_duration_to_pt is 0, we can get on the public transport ONLY if the place is a stop_point
-        if instance.georef.get_stop_points_from_uri(center_isochron.uri):
-            return {mode: {center_isochron.uri: {"duration": 0, "status": response_pb2.reached}}}
+        if instance.georef.get_stop_points_from_uri(center_isochrone.uri):
+            return {mode: {center_isochrone.uri: {"duration": 0, "status": response_pb2.reached}}}
         else:
             return {mode: {}}
     sn_routing_matrix = instance.get_street_network_routing_matrix(origins,
@@ -298,9 +298,9 @@ def _sn_routing_matrix(instance, origins, destinations, is_orig_center, mode, ma
     result = {mode: {}}
     for pos, r in enumerate(sn_routing_matrix.rows[0].routing_response):
         if r.routing_status != response_pb2.unreached:
-            duration = _get_duration(r, places_isochron[pos], mode, **kwargs)
+            duration = _get_duration(r, places_isochrone[pos], mode, **kwargs)
             if duration < max_duration_to_pt:
-                result[mode].update({places_isochron[pos].uri: {'duration': duration, 'status': r.routing_status}})
+                result[mode].update({places_isochrone[pos].uri: {'duration': duration, 'status': r.routing_status}})
     return result
 
 
--- a/source/jormungandr/jormungandr/scenarios/experimental.py
+++ b/source/jormungandr/jormungandr/scenarios/experimental.py
@@ -275,31 +275,32 @@ def _get_duration(resp, place, mode, **kwargs):
     return map_response[resp.routing_status]
 
 
-def _sn_routing_matrix(instance, place, places_crowfly, mode, max_duration_to_pt, request, **kwargs):
+def _sn_routing_matrix(instance, origins, destinations, is_orig_center, mode, max_duration_to_pt, request, **kwargs):
     # When max_duration_to_pt is 0, there is no need to compute the fallback to pt, except if place is a stop_point or a
     # stop_area
+    center_isochron = origins[0] if is_orig_center else destinations[0]
+    places_isochron = destinations if is_orig_center else origins
     if max_duration_to_pt == 0:
         # When max_duration_to_pt is 0, we can get on the public transport ONLY if the place is a stop_point
-        if instance.georef.get_stop_points_from_uri(place.uri):
-            return {mode: {place.uri: {"duration": 0, "status": response_pb2.reached}}}
+        if instance.georef.get_stop_points_from_uri(center_isochron.uri):
+            return {mode: {center_isochron.uri: {"duration": 0, "status": response_pb2.reached}}}
         else:
             return {mode: {}}
-    places = places_crowfly[mode]
-    sn_routing_matrix = instance.get_street_network_routing_matrix([place],
-                                                                   places,
+    sn_routing_matrix = instance.get_street_network_routing_matrix(origins,
+                                                                   destinations,
                                                                    mode,
                                                                    max_duration_to_pt,
                                                                    request,
                                                                    **kwargs)
-    if not len(sn_routing_matrix.rows[0].routing_response):
+    if not len(sn_routing_matrix.rows) or not len(sn_routing_matrix.rows[0].routing_response):
         return {mode: {}}
 
     result = {mode: {}}
     for pos, r in enumerate(sn_routing_matrix.rows[0].routing_response):
         if r.routing_status != response_pb2.unreached:
-            duration = _get_duration(r, places[pos], mode, **kwargs)
+            duration = _get_duration(r, places_isochron[pos], mode, **kwargs)
             if duration < max_duration_to_pt:
-                result[mode].update({places[pos].uri: {'duration': duration, 'status': r.routing_status}})
+                result[mode].update({places_isochron[pos].uri: {'duration': duration, 'status': r.routing_status}})
     return result
 
 
@@ -326,8 +327,9 @@ class AsyncWorker(object):
             if dep_mode not in called_dep_modes:
                 origin_futures.append(self.pool.spawn(_sn_routing_matrix,
                                                       self.instance,
-                                                      origin,
-                                                      origins_places_crowfly,
+                                                      [origin],
+                                                      origins_places_crowfly[dep_mode],
+                                                      True,
                                                       dep_mode,
                                                       get_max_fallback_duration(self.request, dep_mode, dp_durations),
                                                       self.request,
@@ -336,8 +338,9 @@ class AsyncWorker(object):
             if arr_mode not in called_arr_modes:
                 destination_futures.append(self.pool.spawn(_sn_routing_matrix,
                                                            self.instance,
-                                                           destination,
-                                                           destinations_places_crowfly,
+                                                           destinations_places_crowfly[arr_mode],
+                                                           [destination],
+                                                           False,
                                                            arr_mode,
                                                            get_max_fallback_duration(self.request, arr_mode),
                                                            self.request,
--- a/source/jormungandr/jormungandr/street_network/geovelo.py
+++ b/source/jormungandr/jormungandr/street_network/geovelo.py
@@ -1,0 +1,187 @@
+#  Copyright (c) 2001-2017, Canal TP and/or its affiliates. All rights reserved.
+#
+# This file is part of Navitia,
+#     the software to build cool stuff with public transport.
+#
+# Hope you'll enjoy and contribute to this project,
+#     powered by Canal TP (www.canaltp.fr).
+# Help us simplify mobility and open public transport:
+#     a non ending quest to the responsive locomotion way of traveling!
+#
+# LICENCE: This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU Affero General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+# GNU Affero General Public License for more details.
+#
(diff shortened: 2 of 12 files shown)


source/jormungandr/jormungandr/street_network/geovelo.py
```diff
@@ -0,0 +1,187 @@
...
+import json
+from navitiacommon import response_pb2
+from jormungandr import app
+from jormungandr.exceptions import TechnicalError, InvalidArguments
+from jormungandr.street_network.street_network import AbstractStreetNetworkService
+from jormungandr.utils import get_pt_object_coord, is_url, record_external_failure
+
+
+class Geovelo(AbstractStreetNetworkService):
+    sn_system_id = 'geovelo'
```
antoine-de: doesn't pep8 enforce this variable to be caps locked ?

source/jormungandr/jormungandr/scenarios/experimental.py
```diff
@@ -275,31 +275,32 @@ def _get_duration(resp, place, mode, **kwargs):
     return map_response[resp.routing_status]
 
 
-def _sn_routing_matrix(instance, place, places_crowfly, mode, max_duration_to_pt, request, **kwargs):
+def _sn_routing_matrix(instance, origins, destinations, is_orig_center, mode, max_duration_to_pt, request, **kwargs):
```
antoine-de, line 278: Not sure it's related to your PR but I think experimental is very hard to understand and will need at least a refactoring

source/jormungandr/jormungandr/street_network/geovelo.py
```diff
@@ -0,0 +1,187 @@
...
+
+    def __init__(self, instance, service_url, timeout=10, **kwargs):
+        self.instance = instance
+        if not is_url(service_url):
+            raise ValueError('service_url is invalid, you give {}'.format(service_url))
```
antoine-de: shouldn't this be a TechnicalError ?

and not important at all but maybe the error message would be better like `service_url {} is not a valid url`

source/jormungandr/jormungandr/street_network/geovelo.py
```diff
@@ -0,0 +1,190 @@
...
+        self.api_key = api_key
+        self.timeout = timeout
+        self.breaker = pybreaker.CircuitBreaker(fail_max=app.config['CIRCUIT_BREAKER_MAX_GEOVELO_FAIL'],
+                                                reset_timeout=app.config['CIRCUIT_BREAKER_GEOVELO_TIMEOUT_S'])
+
+    def _pt_object_summary(self, pt_object):
+        coord = get_pt_object_coord(pt_object)
+        return [coord.lat, coord.lon, getattr(pt_object, 'uri', None)]
+
+    def _make_data(self, origins, destinations):
```
antoine-de: I think `_make_request_arguments` is a better name

source/jormungandr/jormungandr/street_network/geovelo.py
```diff
@@ -0,0 +1,190 @@
...
+        data = {}
+        data['starts'] = []
+        for o in origins:
+            data['starts'].append(self._pt_object_summary(o))
```
antoine-de: `data['starts'] = [self._pt_object_summary(o) for o in origins]`

you even rewrite the whole function:
```python
return {
'start': [self._pt_object_summary(o) for o in origins],
'ends': [self._pt_object_summary(o) for o in destinations],
}
```

source/jormungandr/jormungandr/street_network/geovelo.py
```diff
@@ -0,0 +1,190 @@
...
+        self.instance = instance
+        if not is_url(service_url):
+            raise ValueError('service_url is invalid, you give {}'.format(service_url))
+        self.service_url = service_url
+        self.api_key = api_key
+        self.timeout = timeout
+        self.breaker = pybreaker.CircuitBreaker(fail_max=app.config['CIRCUIT_BREAKER_MAX_GEOVELO_FAIL'],
+                                                reset_timeout=app.config['CIRCUIT_BREAKER_GEOVELO_TIMEOUT_S'])
+
+    def _pt_object_summary(self, pt_object):
```
antoine-de: can be either a free function of a static method

source/jormungandr/jormungandr/street_network/geovelo.py
antoine-de: can be either a free function of a static method

source/jormungandr/jormungandr/street_network/geovelo.py
```diff
@@ -0,0 +1,190 @@
...
+            self.record_external_failure('circuit breaker open')
+        except requests.Timeout as t:
+            logging.getLogger(__name__).error('Geovelo routing service dead (error: {})'.format(t))
+            self.record_external_failure('timeout')
+        except Exception as e:
+            logging.getLogger(__name__).exception('Geovelo routing error')
+            self.record_external_failure(str(e))
+        return None
+
+    def _get_matrix(self, json_response):
```
antoine-de, line 84: I think it's worth adding a comment there

source/jormungandr/jormungandr/street_network/geovelo.py
```diff
@@ -0,0 +1,190 @@
...
+                routing.duration = e[2]
+                routing.routing_status = response_pb2.reached
+            else:
+                routing.duration = -1
+                routing.routing_status = response_pb2.unknown
+        return sn_routing_matrix
+
+    def _check_response(self, response):
+        if response == None:
+            raise TechnicalError('impossible to access geovelo service')
```
antoine-de: we don't want to return a response if geovelo is down ?

source/jormungandr/jormungandr/street_network/geovelo.py
```diff
@@ -0,0 +1,190 @@
...
+            if e[2]:
+                routing.duration = e[2]
+                routing.routing_status = response_pb2.reached
+            else:
+                routing.duration = -1
+                routing.routing_status = response_pb2.unknown
+        return sn_routing_matrix
+
+    def _check_response(self, response):
+        if response == None:
```
antoine-de: should be `is None`

source/jormungandr/jormungandr/street_network/geovelo.py
```diff
@@ -0,0 +1,190 @@
...
+            raise TechnicalError('impossible to access geovelo service')
+        if response.status_code != 200:
+            logging.getLogger(__name__).error('Geovelo service unavailable, impossible to query : {}'
```
antoine-de: Isn't the `error` level a bit high for this ?

source/jormungandr/jormungandr/street_network/geovelo.py
```diff
@@ -0,0 +1,190 @@
...
+            routing = row.routing_response.add()
+            if e[2]:
+                routing.duration = e[2]
+                routing.routing_status = response_pb2.reached
+            else:
+                routing.duration = -1
+                routing.routing_status = response_pb2.unknown
+        return sn_routing_matrix
+
+    def _check_response(self, response):
```
antoine-de: can be either a free function of a static method

source/jormungandr/jormungandr/street_network/geovelo.py
```diff
@@ -0,0 +1,190 @@
...
+        self._check_response(r)
+        resp_json = r.json()
+        return self._get_matrix(resp_json)
+
+    def _get_response(self, json_response, mode, pt_object_origin, pt_object_destination, datetime, clockwise):
+        resp = response_pb2.Response()
+        resp.status_code = 200
+        resp.response_type = response_pb2.ITINERARY_FOUND
+
+        duration = 13371337
```
antoine-de: Note sure to understand this code.

there can be only one duration returned by geovelo ?

I think there should at least be more comments :wink: but why do you init this a this value ?

source/jormungandr/jormungandr/street_network/geovelo.py
```diff
@@ -0,0 +1,190 @@
...
+
+        section = journey.sections.add()
+        section.type = response_pb2.STREET_NETWORK
+
+        section.begin_date_time = journey.departure_date_time
+        section.end_date_time = journey.arrival_date_time
+
+        section.id = 'section_0'
+        section.duration = journey.duration
+        section.length = journey.duration*3
```
antoine-de: ?

source/jormungandr/jormungandr/street_network/geovelo.py
```diff
@@ -0,0 +1,190 @@
...
+        pt_coord.lat, pt_coord.lon = coord.lat, coord.lon
+
+        return resp
+
+    def direct_path(self, mode, pt_object_origin, pt_object_destination, datetime, clockwise, request):
+        if mode != "bike":
+            logging.getLogger(__name__).error('Geovelo, mode {} not implemented'.format(mode))
+            raise InvalidArguments('Geovelo, mode {} not implemented'.format(mode))
+
+        data = {}
```
antoine-de: can't you use `_make_data` here ?

source/jormungandr/jormungandr/street_network/geovelo.py
```diff
@@ -0,0 +1,190 @@
...
+            logging.getLogger(__name__).exception('Geovelo routing error')
+            self.record_external_failure(str(e))
+        return None
+
+    def _get_matrix(self, json_response):
+        sn_routing_matrix = response_pb2.StreetNetworkRoutingMatrix()
+        row = sn_routing_matrix.rows.add()
+        for e in json_response:
+            if e[2] == 'duration':
+                continue
```
TeXitoi: better to
```python
check_headers(json_response[0])
for e in json_response[1:]:
    ...
```

source/jormungandr/jormungandr/street_network/street_network.py
```diff
@@ -40,6 +40,7 @@
 
 
 class AbstractStreetNetworkService(ABC):
+    connector_type = 'streetnetwork'
```
xlqian: is this variable used?

source/jormungandr/jormungandr/street_network/tests/streetnetwork_test_utils.py
```diff
@@ -0,0 +1,52 @@
...
+#
+# Stay tuned using
+# twitter @navitia
+# IRC #navitia on freenode
+# https://groups.google.com/d/forum/navitia
+# www.navitia.io
+from navitiacommon import type_pb2
+
+
+def get_pt_object(embedded_type, lon, lat, uri=None):
```
TeXitoi: that's more a make than a get

source/jormungandr/jormungandr/street_network/geovelo.py
xlqian: since `sn_system_id` is a common variable for `GeoVelo` `Kraken` and `Valhalla`, i'm wondering would it be better if we define an abstract `get_system_id` in abstract class and implement it in every inheriting class?

source/jormungandr/jormungandr/scenarios/experimental.py
pbougue: agreed (and also as you say, it's not the point of this PR to refactor, just a side effect when needed, so later :smiley: )

source/jormungandr/jormungandr/street_network/geovelo.py
pbougue: copied on valhalla management... crowfly for all? may be managed outside of connector, though.

source/jormungandr/jormungandr/street_network/geovelo.py
pbougue: copied on valhalla too... depends on how we handle errors, for now it's fine for me

source/jormungandr/jormungandr/street_network/geovelo.py
pbougue: the point is to get only one duration, the same provided by 1-n request, yes.
and this code is dirty, but don't read too much in it, the next PR really uses direct_path call to geovelo

source/jormungandr/jormungandr/street_network/geovelo.py
pbougue: arbitrary value to provide something "pretty" realistic from the tests I did manually : it looks like the speed is close from 3m/s (again, next PR breaks that)

source/jormungandr/jormungandr/street_network/geovelo.py
pbougue: Done

TeXitoi (approved):

source/jormungandr/jormungandr/street_network/geovelo.py
```diff
@@ -34,42 +34,42 @@
...
-        for o in origins:
-            data['starts'].append(self._pt_object_summary(o))
-        data['ends'] = []
-        for d in destinations:
-            data['ends'].append(self._pt_object_summary(d))
-        return data
+    @staticmethod
+    def _make_request_arguments(origins, destinations):
+        return {
+            'starts': [Geovelo._pt_object_summary(o) for o in origins],
```
antoine-de: this is valid, but you can also make it a classmethod and take `cls` as the first argument of the method

source/jormungandr/jormungandr/street_network/geovelo.py
```diff
@@ -82,9 +82,17 @@ def _call_geovelo(self, url, method=requests.post, data=None):
...
+        as it is ordered, we only consider the third
+        '''
         sn_routing_matrix = response_pb2.StreetNetworkRoutingMatrix()
         row = sn_routing_matrix.rows.add()
-        for e in json_response:
+        if json_response[0] != ["start_reference", "end_reference", "duration"]:
+            logging.getLogger(__name__).error('Geovelo parsing error. Response: {}'.format(json_response))
+            raise UnableToParse('Geovelo parsing error. Response: {}'.format(json_response))
+        for e in json_response[1:]:
             if e[2] == 'duration':
```
antoine-de: you can remove this no ?

source/jormungandr/jormungandr/street_network/geovelo.py
```diff
@@ -82,9 +82,17 @@ def _call_geovelo(self, url, method=requests.post, data=None):
...
     def _get_matrix(self, json_response):
+        '''
+        build the 1-n response matrix from geovelo table
+        each element is ["start", "end", duration] (first being the header)
+        as it is ordered, we only consider the third
+        '''
         sn_routing_matrix = response_pb2.StreetNetworkRoutingMatrix()
         row = sn_routing_matrix.rows.add()
-        for e in json_response:
+        if json_response[0] != ["start_reference", "end_reference", "duration"]:
```
antoine-de: can you add a comment saying this is the header of the response ?

PR #1990, opened 2017-02-21
Doc: warnings on object's order and id/uri

pbougue:


documentation/slate/source/includes/interface.md
```diff
@@ -126,3 +126,29 @@ That means :
 * where you can find the details of your object ( **"id": "edc46f3a-ad3d-11e4-a5e1-005056a44da2"** ).
 
 
+
+Objects order
+------------
+
+Unless specified otherwise objects are not sorted and stability of object's order is not guaranteed.
```
stifoon: would prefer
"Unless specified, objects are not sorted and stability of object's order is not guaranteed."
no?

stifoon: plurial
objects' order

documentation/slate/source/includes/interface.md
```diff
@@ -126,3 +126,29 @@ That means :
...
+
+Examples of sorted objects:
+* journeys in a '/journeys' response
+* '/departures' and '/arrivals'
+* stop_points in '/routes/<id_route>/stop_points'
+* '/places_nearby'
+* '/places'
+
```
stifoon: should explain the complex sort of "stop_schedules" ;)
and "traffic_reports" \o/

documentation/slate/source/includes/interface.md
```diff
@@ -126,3 +126,29 @@ That means :
...
+* stop_points in '/lines/<id_line>/stop_points'
+* pretty much everything else...
+
+
+
+Lifetime of id
+--------------
+
+The way id (or uri) of objects are generated is not garanteed stable, nor are the data processed.
+So we advise to limit to the minimum storing id/uri of objects.
```
stifoon: ids/uris?

documentation/slate/source/includes/interface.md
```diff
@@ -126,3 +126,29 @@ That means :
...
+Examples of unsorted responses:
+* stop_points in '/lines/<id_line>/stop_points'
+* pretty much everything else...
+
+
+
+Lifetime of id
+--------------
+
+The way id (or uri) of objects are generated is not garanteed stable, nor are the data processed.
```
stifoon: ids/uris

documentation/slate/source/includes/interface.md
pbougue: To me it's not the point of this part of doc... Can be in the same PR though, but I have to dig into it a bit before writing it.

PR #1991, opened 2017-02-23
Jormungandr: better display of line_sections in /journeys and schedules

antoine-de:
For /journeys:
we display a line section impact if:
 * a vj we use is impacted on the stops we use in the journey section
 * a stop_point we use is impacted by a line section

for schedules (/departures, /arrivals, /stop_schedules...):
* we display a line section if the (stop_points, vj) used is part of a line sections

to better understand this you can read the tests :wink: 

Note:
I'm not a fan of line section detection (`fill_messages(const VjStopTimes* vj_stoptimes, PtDisplayInfo* pt_display_info)`), but I don't see a better way :confused:

### DIFF
--- a/source/time_tables/passages.cpp
+++ b/source/time_tables/passages.cpp
@@ -191,7 +191,8 @@ void passages(PbCreator& pb_creator,
 
         const type::VehicleJourney* vj = dt_stop_time.second->vehicle_journey;
         passage->mutable_stop_date_time()->set_data_freshness(to_pb_realtime_level(vj->realtime_level));
-        const auto& vj_st = navitia::VjStopTimes(vj, dt_stop_time.second, nullptr);
+        const auto sts = std::vector<const nt::StopTime*>{dt_stop_time.second};
+        const auto& vj_st = navitia::VjStopTimes(vj, sts);
         pb_creator.fill(&vj_st, passage->mutable_pt_display_informations(), 1);
         fill_route_point(pb_creator, depth, vj->route, dt_stop_time.second->stop_point, passage);
     }
--- a/source/time_tables/route_schedules.cpp
+++ b/source/time_tables/route_schedules.cpp
@@ -398,7 +398,8 @@ void route_schedule(PbCreator& pb_creator, const std::string& filter,
                     pbnavitia::Header* header = table->mutable_headers(j);
                     pbnavitia::PtDisplayInfo* vj_display_information = header->mutable_pt_display_informations();
                     auto vj = dt_stop_time.second->vehicle_journey;
-                    const auto& vj_st = navitia::VjStopTimes(vj, dt_stop_time.second, nullptr);
+                    auto sts = std::vector<const nt::StopTime*>{dt_stop_time.second};
+                    const auto& vj_st = navitia::VjStopTimes(vj, sts);
                     pb_creator.fill(&vj_st, vj_display_information, 0);
                     // as we only issue headsign for vj in route_schedules:
                     // - need to override headsign with trip headsign (i.e. vj.name)
--- a/source/type/pb_converter.cpp
+++ b/source/type/pb_converter.cpp
@@ -1153,8 +1153,51 @@ void PbCreator::Filler::fill_pb_object(const VjOrigDest* vj_orig_dest, pbnavitia
     }
 }
 
+/*
+ * override fill_messages for journey sections to handle line sections impacts
+ *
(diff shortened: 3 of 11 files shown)


source/time_tables/passages.cpp
```diff
@@ -191,7 +191,8 @@ void passages(PbCreator& pb_creator,
 
         const type::VehicleJourney* vj = dt_stop_time.second->vehicle_journey;
         passage->mutable_stop_date_time()->set_data_freshness(to_pb_realtime_level(vj->realtime_level));
-        const auto& vj_st = navitia::VjStopTimes(vj, dt_stop_time.second, nullptr);
+        const auto sts = std::vector<const nt::StopTime*>{dt_stop_time.second};
+        const auto& vj_st = navitia::VjStopTimes(vj, sts);
```
xlqian, line 195: you are copying the vector twice... is that wanted? :confused:

source/time_tables/route_schedules.cpp
```diff
@@ -398,7 +398,8 @@ void route_schedule(PbCreator& pb_creator, const std::string& filter,
                     pbnavitia::Header* header = table->mutable_headers(j);
                     pbnavitia::PtDisplayInfo* vj_display_information = header->mutable_pt_display_informations();
                     auto vj = dt_stop_time.second->vehicle_journey;
-                    const auto& vj_st = navitia::VjStopTimes(vj, dt_stop_time.second, nullptr);
+                    auto sts = std::vector<const nt::StopTime*>{dt_stop_time.second};
+                    const auto& vj_st = navitia::VjStopTimes(vj, sts);
```
xlqian: same here

source/type/pb_converter.cpp
```diff
@@ -1153,8 +1153,51 @@ void PbCreator::Filler::fill_pb_object(const VjOrigDest* vj_orig_dest, pbnavitia
...
+                        break;
+                    }
+                }
+                if (found) { break; }
+            }
+            if (! found) {
+                // we haven't found a stoppoint impacted by this line section impact, it does not concern this vj
+                continue;
+            }
+        }
```
xlqian: a little suggestion to get rid of the doctrine of `break`:
```c++
    bool found = [](){
       if (line_section_impacted_obj_it != message->informed_entities().end()) {
            // note in this we take the premise that an impact cannot impact a line section AND a vj
            const auto& line_section = *line_section_impacted_obj_it;
            for (const auto& st: vj_stoptimes->stop_times) {
                // if one stop point of the stoptimes is impacted by the same impact
                // it means the section is impacted
                for (const auto& sp_message : st->stop_point->get_applicable_messages(pb_creator.now,
                                                                            pb_creator.action_period)) {
                    if (sp_message == message) {
                        return true;
                    }
                }
            }
            return false;
        }}();
      // we haven't found a stoppoint impacted by this line section impact, it does not concern this vj
      if (! found) continue;
```

source/time_tables/passages.cpp
antoine-de: hum Actually I don't think it is, but maybe we should :grin: 

`VjStopTimes` keeps a `const&` on the vector, so it is not copied, but it can thus lead to some c++ *$£$ù$* with a ref to a dead object :cry: 

what do you think about this ?

source/type/pb_converter.cpp
xlqian: @antoine-de  code not tested, use with precaution XD

source/time_tables/passages.cpp
xlqian: oops, I didn't see the `const&`, you're right, my bad :)

xlqian: i don't understand why you did a copy at L194? since it's declared `const`, why don't you pass the ref of `dt_stop_time.second` to `navitia::VjStopTimes` directly?

source/type/pb_converter.cpp
```diff
@@ -1153,8 +1153,51 @@ void PbCreator::Filler::fill_pb_object(const VjOrigDest* vj_orig_dest, pbnavitia
     }
 }
 
+/*
+ * override fill_messages for journey sections to handle line sections impacts
+ *
+ * for a journey section, we get:
+ *  - all of the meta journey impacts but the line sections impacts
+ *  - for the line sections impacts we check that theyintersect the journey section
```
TeXitoi: lack space

TeXitoi (approved):

PR #2003, opened 2017-03-02
Kraken: fix disruption.updated_at

pbougue:
Fixing http://jira.canaltp.fr/browse/NAVITIAII-2198

We where using `created_at` field from chaos, using now `updated_at`

Last commit corrects https://github.com/CanalTP/navitia/pull/1991/files#diff-9db5f8b2306c44a2a59af3215f2beb47R1187

### DIFF
--- a/source/jormungandr/tests/chaos_disruptions_tests.py
+++ b/source/jormungandr/tests/chaos_disruptions_tests.py
@@ -143,7 +143,7 @@ class TestPlacesWithDisruptions(ChaosDisruptionsFixture):
         assert response['places'][0]['id'] == 'stopB'
 
         assert len(response['disruptions']) == 1
-        assert response['disruptions'][0]['updated_at'] == u'20160405T150623'
+        assert response['disruptions'][0]['updated_at'] == u'20160405T150733'
 
 
 @dataset(MAIN_ROUTING_TEST_SETTING)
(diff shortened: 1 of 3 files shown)


source/jormungandr/tests/chaos_disruptions_tests.py
```diff
@@ -684,7 +684,8 @@ def make_mock_chaos_item(disruption_name, impacted_obj, impacted_obj_type, start
     impact = disruption.impacts.add()
     impact.id = "impact_" + disruption_name + "_1"
     enums_impact = gtfs_realtime_pb2.Alert.DESCRIPTOR.enum_values_by_name
-    impact.created_at = 1459868783 # u'20160405T150623' 
+    impact.created_at = 1459868783 # u'20160405T150623'
```
antoine-de: for this (I know it has not been done everywhere), I prefer to use a timestamp function (there even might be one in utils):
`impact.created_at = my_time_stamp_function('20160405T150623')`
it is clearer and less error prone

source/type/pb_converter.cpp
```diff
@@ -1184,8 +1184,9 @@ void PbCreator::Filler::fill_messages(const VjStopTimes* vj_stoptimes,
                         }
                     }
                 }
-                return false;
-            }}();
+            }
+            return false;
+            }();
           // we haven't found a stoppoint impacted by this line section impact, it does not concern this vj
           if (! found) continue;
```
TeXitoi: Please indent all that consistently.

xlqian (approved):

antoine-de (approved):

PR #2005, opened 2017-03-03
Jormungandr: fix physical_modes with serpy

TeXitoi:


pbougue (approved):
Would be nice to add tests on that, though

antoine-de (approved):

PR #2010, opened 2017-03-08
jormungandr: fix encoding problem in SIRI RT proxy

TeXitoi:


source/jormungandr/jormungandr/realtime_schedule/siri.py
```diff
@@ -73,7 +73,7 @@ def _get_next_passage_for_route_point(self, route_point, count, from_dt, current
         if not siri_response or siri_response.status_code != 200:
             return None
         logging.getLogger(__name__).debug('siri for {}: {}'.format(stop, siri_response.text))
```
pbougue, line 75: content here also?

pbougue: only text works in logger

pbougue (approved):

PR #2019, opened 2017-03-16
Jormun: manage places/uri from mimir

azime:
- [x] Tests Bragi Error

### DIFF
--- a/source/jormungandr/jormungandr/autocomplete/geocodejson.py
+++ b/source/jormungandr/jormungandr/autocomplete/geocodejson.py
@@ -50,18 +50,51 @@ class GeocodeJson(AbstractAutocomplete):
     TYPE_STREET = "street"
 
     def __init__(self, **kwargs):
-        self.external_api = kwargs.get('host')
+        self.host = kwargs.get('host')
         self.timeout = kwargs.get('timeout', 10)
 
+    def call_bragi(self, url, method, **kwargs):
+        try:
+            return method(url, **kwargs)
+        except requests.Timeout:
+            logging.getLogger(__name__).error('autocomplete request timeout')
+            raise TechnicalError('external autocomplete service timeout')
+        except:
+            logging.getLogger(__name__).exception('error in autocomplete request')
+            raise TechnicalError('impossible to access external autocomplete service')
+
+    def response_marshaler(self, response_bragi):
+        json_response = response_bragi.json()
+        from flask.ext.restful import marshal
+        from jormungandr.interfaces.v1.Places import geocodejson
+
+        return marshal(json_response, geocodejson)
+
+    def make_url(self, end_point, uri=None):
+
+        if end_point not in ['autocomplete', 'features']:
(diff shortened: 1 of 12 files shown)


source/jormungandr/jormungandr/autocomplete/geocodejson.py
```diff
@@ -50,18 +50,51 @@ class GeocodeJson(AbstractAutocomplete):
...
+
+        if not self.host:
+            raise TechnicalError('global autocomplete not configured')
+
+        url = "{host}/{end_point}".format(host=self.host, end_point=end_point)
+        if uri:
+            url = '{url}/{uri}'.format(url=url, uri=uri)
+        return url
+
+    def basic_params(self, instance):
```
antoine-de, line 98: I'm not against, but by curiosity why did you change this ?

source/jormungandr/jormungandr/autocomplete/geocodejson.py
```diff
@@ -50,18 +50,51 @@ class GeocodeJson(AbstractAutocomplete):
...
+        try:
+            return method(url, **kwargs)
+        except requests.Timeout:
+            logging.getLogger(__name__).error('autocomplete request timeout')
+            raise TechnicalError('external autocomplete service timeout')
+        except:
+            logging.getLogger(__name__).exception('error in autocomplete request')
+            raise TechnicalError('impossible to access external autocomplete service')
+
+    def response_marshaler(self, response_bragi):
```
antoine-de: you can make this either a free function of a static method I think

source/jormungandr/jormungandr/autocomplete/geocodejson.py
azime: This method should return the common parameters to "get" and "get_uri", common_params ?

source/jormungandr/jormungandr/autocomplete/geocodejson.py
azime: ok

source/jormungandr/jormungandr/autocomplete/geocodejson.py
```diff
@@ -50,18 +50,63 @@ class GeocodeJson(AbstractAutocomplete):
...
+            raise TechnicalError('external autocomplete service timeout')
+        except:
+            logging.getLogger(__name__).exception('error in autocomplete request')
+            raise TechnicalError('impossible to access external autocomplete service')
+
+    @classmethod
+    def _check_response(cls, response, uri):
+        if response == None:
+            raise TechnicalError('impossible to access autocomplete service')
+        if response.status_code == 404:
```
antoine-de, line 71: I don't think it is an invalid argument, it's just that we don't know it :wink: 

aren't there an UnknownObject error (or something like this) ?

azime: ok

antoine-de (approved):

PR #2024, opened 2017-03-20
Ed: connexe component by mode

antoine-de:
We now compute the connexe components by mode.

This way we can disable some routing mode on the edges

This is done so we don't have any walking edges accessible only by car edges (like in a parking)

fixes http://jira.canaltp.fr/browse/NAVITIAII-2199

**Warning**: I did not reimplement the old behavior to filter the `Way` that do not have any edges any more

This will lead certain way to be found (like with the autocomplete) but we'll be unable to project  on them (thus we'll project to a near way).

I don't think this is bad (and it might even lead to better behavior)

Note: I did not do any automated tests as I do not know how to do them
easily

On the ile de france dataset:
total number of nodes: 587289

|                        | walking | bike    | car      |
|-------------------|-----------|---------|----------|
|useless nodes | 31699   |92076 |117237|
|disable edges  | 33796 | 14021  | 9001   |

| | before | now |
|--- | --- | ---- |
|total number of disable nodes | 19429 | 22900|
| computation time | 15.7 | 3.2s |

example of filtering:
near gare de l'est in paris, green is the classical graph, red is the filtered edges for walking.
The star is the starting point of the ticket example

selection_002

### DIFF
--- a/source/ed/ed_reader.cpp
+++ b/source/ed/ed_reader.cpp
@@ -53,7 +53,7 @@ void EdReader::fill(navitia::type::Data& data, const double min_non_connected_gr
 
     pqxx::work work(*conn, "loading ED");
 
-    this->fill_vector_to_ignore(data, work, min_non_connected_graph_ratio);
+    this->fill_vector_to_ignore(work, min_non_connected_graph_ratio);
     this->fill_meta(data, work);
     // TODO merge fill_feed_infos, fill_meta
     this->fill_feed_infos(data, work);
@@ -625,7 +625,7 @@ void EdReader::fill_lines(nt::Data& data, pqxx::work& work){
     for(auto const_it = property_result.begin(); const_it != property_result.end(); ++const_it){
         auto line_it = this->line_map.find(const_it["object_id"].as<idx_t>());
         if(line_it != this->line_map.end()) {
-            line_it->second->properties[const_it["property_name"].as<std::string>()] = const_it["property_value"].as<std::string>(); 
+            line_it->second->properties[const_it["property_name"].as<std::string>()] = const_it["property_value"].as<std::string>();
         }
     }
 }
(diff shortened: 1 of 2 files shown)


source/ed/ed_reader.cpp
```diff
@@ -625,7 +625,7 @@ void EdReader::fill_lines(nt::Data& data, pqxx::work& work){
     for(auto const_it = property_result.begin(); const_it != property_result.end(); ++const_it){
         auto line_it = this->line_map.find(const_it["object_id"].as<idx_t>());
         if(line_it != this->line_map.end()) {
-            line_it->second->properties[const_it["property_name"].as<std::string>()] = const_it["property_value"].as<std::string>(); 
+            line_it->second->properties[const_it["property_name"].as<std::string>()] = const_it["property_value"].as<std::string>();
```
pbougue: minor: length

source/ed/ed_reader.cpp
```diff
@@ -1296,51 +1296,73 @@ void EdReader::fill_house_numbers(navitia::type::Data& data, pqxx::work& work){
...
-    navitia::georef::GeoRef geo_ref_temp;
-    std::unordered_map<idx_t, uint64_t> map_idx_to_id;
+struct ComponentVertice {
+    uint64_t db_id;
+};
+struct ComponentEdge {
+    navitia::flat_enum_map<nt::Mode_e, bool> modes;
+};
+
+typedef boost::adjacency_list<boost::vecS, boost::vecS, boost::directedS, ComponentVertice, ComponentEdge> ComponentGraph;
```
pbougue: Instead of `Component`Graph (which tells the usage, not what it is) I'd be more fan of `SNMode`Graph (or something like that), telling that it stores info on Mode.
(Same for sub-types)
@TeXitoi if you have an opinion...

source/ed/ed_reader.cpp
```diff
@@ -1475,16 +1512,28 @@ void EdReader::fill_graph(navitia::type::Data& data, pqxx::work& work, bool expo
                 way->geoms.push_back(geometry);
             }
         }
+        auto edge_id = std::make_pair(const_it["source_node_id"].as<uint64_t>(), const_it["target_node_id"].as<uint64_t>());
```
pbougue: minor: length

pbougue (approved):

TeXitoi (approved):

PR #2027, opened 2017-03-21
Doc: add a sandBox token in playground uri

stifoon:


pbougue (changes requested):
To be closed and added to #2029

pbougue (changes requested):
We don't want to have one Jenkins build each time a line is changed, so maybe you can do one by one and snip Jenkins build? Or all in one PR :)

PR #2028, opened 2017-03-21
Doc: add sandbox token in a playground uri

stifoon:


pbougue (changes requested):
To be closed and added to #2029

Issue #2001, opened 2017-03-02
Previous departure/arrival return wrong results on the first production day


eturck:
While writing the new passages unit test in #1980 I stumbled upon this bug.
You can make a request to get the previous departures somewhere, let's say something like that : 
http://api.navitia.io/v1/coverage/fr-idf/stop_areas/stop_area:OIF:SA:59410/departures?until_datetime=20170215T010000&count=100

You get : 
```json
"pagination": {
    "start_page": 0,
    "items_on_page": 56,
    "items_per_page": 100,
    "total_result": 56
}
```

But if you make the same request, one second earlier (http://api.navitia.io/v1/coverage/fr-idf/stop_areas/stop_area:OIF:SA:59410/departures?until_datetime=20170215T005959&count=100) you get : 
```json
"pagination": {
    "start_page": 0,
    "items_on_page": 100,
    "items_per_page": 100,
    "total_result": 100
}
```
And you don't even get the previous departures, but the next ones :sweat_smile:. And it's because the start production day is 2017/02/14.

The problem is in this function : https://github.com/CanalTP/navitia/blob/dev/source/time_tables/request_handle.cpp#L37
Where we only check that datetime.date() and (datetime + duration).date() is in the production period, not (datetime - duration).date() (the check should depends on the clockwise parameter).

Since max_datetime is an uint_32 and we substract a duration > datetime we have a max_datetime with a value of MAX_INT - (duration - datetime). And since get_stop_times compare dt & max_dt to determine if we are clockwise or not (https://github.com/CanalTP/navitia/blob/dev/source/routing/get_stop_times.cpp#L47), it thinks we are clockwise.

There is two ways I see to fix this, both pretty easy : 
- Keep the current idea, if (datetime - duration).date() is not in the production period just throw an out_of_bounds exception when we're not clockwise,
- Or limit the value of max_dt to min(dt + duration, production_date.end()) in clockwise and max(0, dt - duration) in non clockwise.

I only propose the second solution because I don't find it pretty obvious, even with next departures / arrival, that this request (http://api.navitia.io/v1/coverage/fr-idf/stop_areas/stop_area:OIF:SA:59410/departures?from_datetime=20170312T120000) is throwing a date_out_of_bounds error but this one doesn't (http://api.navitia.io/v1/coverage/fr-idf/stop_areas/stop_area:OIF:SA:59410/departures?from_datetime=20170312T120000&duration=3600) when getting departures on the last production day.

[...]

PR #2032, opened 2017-03-23
Change behavior of passages on first and last production day

eturck:
Following issue #2001. This is fixing the bug I encountered and also change the behavior of max_datetime calculation. Now : 
- We're throwing an out of bound error only if the requested datetime is outside the production_period,
- In clockwise we limit max_datetime at production_period.end() at midnight,
- In !clockwise we limit max_datetime at 0 if duration is greater than the requested datetime.

This way, even if we're not returning all the departures for the requested duration we are at least returning the departures we have.


antoine-de (approved):
great, thanks for the fix!

PR #2033, opened 2017-03-23
Jormungandr: Put some logics into connector

xlqian:
put some logics into connector


TeXitoi (approved):

pbougue (changes requested):
Instead of kwargs, we should use an enum on FallbakBeforePT, FallbackAfterPT, DirectNoPT (names can change but you have the meaning).

source/jormungandr/jormungandr/scenarios/experimental.py
```diff
@@ -77,17 +71,17 @@ def get_max_fallback_duration(request, mode, dp_durations={}):
     raise ValueError('unknown mode: {}'.format(mode))
 
 
-def make_direct_path_key(dep_mode, orig_uri, dest_uri, is_fallback_at_end, fallback_extremity):
+def make_direct_path_key(dep_mode, orig_uri, dest_uri, direct_path_type, fallback_extremity):
     '''
     :param orig_uri, dest_uri, mode: matters obviously
-    :param is_fallback_at_end: whether it's a fallback at the beginning or end of journey also matters
+    :param direct_path_type: whether it's a fallback at the beginning or end of journey also matters
```
pbougue: you should update this comment : it can also be a directNoPT

source/jormungandr/jormungandr/street_network/street_network.py
```diff
@@ -38,16 +38,23 @@
 # http://stackoverflow.com/a/38668373/1614576
 ABC = abc.ABCMeta(str("ABC"), (object,), {})
 
+# we need to "invert" a direct path when it's a ending fallback by car if and only if it's returned by kraken
+# In other case, it's ignored
+class DirectPathType:
+    NORMAL = 0
```
pbougue: Too vague to me

source/jormungandr/jormungandr/street_network/street_network.py
```diff
@@ -38,16 +38,23 @@
...
+    ENDING_CAR_FALLBACK_BY_KRAKEN = 1
```
pbougue: We cannot afford a so connector-specific name/behaviour in generic class / experimental scenario.
Functionnally we have 3 types of directPath (cf. previous comment), then cache (and so make_direct_path_key) must be connector-specific, and in the connector we can choose to manage 2 of the 3 DirectPathType's the same way...
But it's the connector's responsability.

pbougue (changes requested):
Not sure it's worth doing it now as it's maybe too much code to move. But the exercise is useful before full refactor.

xlqian:
jenkins, test this please!

xlqian:
jenkins, test this please!

xlqian:
jenkins, test this please!

source/jormungandr/jormungandr/street_network/kraken.py
```diff
@@ -29,19 +29,41 @@
...
+                d = copy.deepcopy(s.destination)
+                s.origin.CopyFrom(d)
+                s.destination.CopyFrom(o)
+                s.end_date_time = previous_section_begin
+                previous_section_begin = s.begin_date_time = s.end_date_time - s.duration
+            j.sections.sort(utils.SectionSorter())
+        return response
+
+    def direct_path(self, mode, pt_object_origin, pt_object_destination, fallback_extremity, request, direct_path_type):
+        should_invert_journey = mode == 'car' and direct_path_type == DirectPathType.ENDING_FALLBACK
```
pbougue: with all thos `=`, `()` would be welcome

source/jormungandr/jormungandr/street_network/street_network.py
```diff
@@ -38,16 +38,23 @@
...
     @abc.abstractmethod
     def get_street_network_routing_matrix(self, origins, destinations, street_network_mode, max_duration, request, **kwargs):
         pass
 
     @abc.abstractmethod
-    def direct_path(self, mode, pt_object_origin, pt_object_destination, fallback_extremity, request):
+    def direct_path(self, mode, pt_object_origin, pt_object_destination, fallback_extremity, request, direct_path_type):
         '''
         :param fallback_extremity: is a PeriodExtremity (a datetime and it's meaning on the fallback period)
+        :param direct_path_type : we need to "invert" a direct path when it's a ending fallback by car if and only if
```
pbougue: I'd put the kraken part of this comment into kraken connector (if it's not already the case). Although it's good to explain here a bit what this param is, and have an example of what can be its use.
Here you're a bit to definitive about its use (it might someday be used by another connector).

source/jormungandr/jormungandr/utils.py
```diff
@@ -388,3 +388,11 @@ def decode_polyline(encoded):
...
 # (mostly used for fallback management in experimental scenario)
 PeriodExtremity = namedtuple('PeriodExtremity', ['datetime', 'represents_start'])
+
+
+class SectionSorter(object):
+    def __call__(self, a, b):
+        if a.begin_date_time != b.begin_date_time:
+            return -1 if a.begin_date_time < b.begin_date_time else 1
+        else:
+            return -1 if a.end_date_time < b.end_date_time else 1
```
pbougue: missing final `\n`

source/jormungandr/jormungandr/street_network/street_network.py
```diff
@@ -38,16 +38,23 @@
...
+    ENDING_FALLBACK = 2
 
 class AbstractStreetNetworkService(ABC):
     @abc.abstractmethod
     def get_street_network_routing_matrix(self, origins, destinations, street_network_mode, max_duration, request, **kwargs):
         pass
 
     @abc.abstractmethod
-    def direct_path(self, mode, pt_object_origin, pt_object_destination, fallback_extremity, request):
+    def direct_path(self, mode, pt_object_origin, pt_object_destination, fallback_extremity, request, direct_path_type):
```
pbougue, line 53: You probably have to do a grep for `direct_path(` in connectors: geovelo should be affected by this change too.
tests also (kraken and geovelo?)

pbougue (changes requested):
:+1: psychopath happy :)
requesting minor changes though

source/jormungandr/jormungandr/street_network/kraken.py
xlqian: ok

source/jormungandr/jormungandr/street_network/street_network.py
xlqian: yep you are right

source/jormungandr/jormungandr/utils.py
pbougue: Done

source/jormungandr/jormungandr/street_network/street_network.py
xlqian: done

pbougue (approved):

PR #2037, opened 2017-03-24
All: add allowed_id[] parameter for journeys

TeXitoi:


source/jormungandr/jormungandr/scenarios/default.py
```diff
@@ -130,9 +130,10 @@ def parse_journey_request(self, requested_type, request):
             req.journeys.streetnetwork_params.origin_mode = "bss"
         if req.journeys.streetnetwork_params.destination_mode == "bike_rental":
             req.journeys.streetnetwork_params.destination_mode = "bss"
-        if "forbidden_uris[]" in request and request["forbidden_uris[]"]:
-            for forbidden_uri in request["forbidden_uris[]"]:
-                req.journeys.forbidden_uris.append(forbidden_uri)
+        for forbidden_uri in get_or_default(request, "forbidden_uris[]", []):
```
antoine-de, line 133: I'm not sure what `get_or_default` do, but I think you can just do 
`for forbidden_uri in request.get("forbidden_uris[]", []):`

source/routing/raptor.cpp
```diff
@@ -520,88 +526,115 @@ RAPTOR::isochrone(const map_stop_point_duration& departures,
...
+    boost::dynamic_bitset<> valid_journey_pattern_points(jp_container.nb_jpps());
+    valid_journey_pattern_points.set();
+    valid_stop_points.set();
+
+    auto forbidden_objs = ObjsFromIds(forbidden, jp_container, data);
+    valid_journey_patterns &= forbidden_objs.jps.flip();
+    valid_journey_pattern_points &= forbidden_objs.jpps.flip();
+    valid_stop_points &= forbidden_objs.sps.flip();
+
+    const auto allowed_objs = ObjsFromIds(allowed, jp_container, data);
```
antoine-de, line 630: you can maybe add a comment here explaining the mechanism

source/jormungandr/jormungandr/scenarios/default.py
TeXitoi:
```python
def get_or_default(request, val, default):
    """
    Flask add all API param in the request dict with None by default,
    so here is a simple helper to get the default value is the key
    is not in the dict or if the value is None
    """
    val = request.get(val, default)
    if val is not None:
        return val
    return default
```

Looks like I can't

documentation/slate/source/includes/apis.md
```diff
@@ -935,6 +936,43 @@ The [isochrones](#isochrones) service exposes another response structure, which
 | nop     | wheelchair           | boolean | If true the traveler is considered to be using a wheelchair, thus only accessible public transport are used<br>be warned: many data are currently too faint to provide acceptable answers with this parameter on       | False       |
 | nop     | debug                | boolean | Debug mode<br>No journeys are filtered in this mode     | False       |
 
+### Precisions on `forbidden_uris[]` and `allowed_id[]`
+
+These parameters are filtering the vehicle journeys and the stop points use to compute the journeys.  The journeys can only use allowed vehicle journeys (as present in the `public_transport` or `on_demand_transport` sections). They also can only use the allowed stop points (as present in the `street_network`, `waiting` and `crow_fly` sections).
```
antoine-de, line 941: :+1: seems quite clear for me.

Hovewer I don't find the hand drawn picture cristal clear, I think a computer generated schema would be better

source/routing/raptor.cpp
antoine-de: done

documentation/slate/source/includes/apis.md
TeXitoi: if you don't like the new one, tune it yourself:
```dot
graph G {
  rankdir=LR;
  SPA -- SPB [label="LA"];
  SPA -- SPB [label="LB"];
  SPA -- SPC [label="LC"];
  SPA -- SPC [label="LD"];
  SPC -- SPB [label="LE"];
  SPC -- SPB [label="LF"];
}
```

antoine-de (approved):

PR #2039, opened 2017-03-28
Jormungandr: Places uri in journeys

antoine-de:
handle id from multiple autocomplete system in journeys

When a journey is requested, a query is done to first the instance's autocomplete system, then to the global autocomplete to have more detail on the requested object.

from this detail an ID is extracted (either the original ID, for stops and admin, or the object coordinates) for kraken to be able to compute the journey

At the end of the journey computation, the journeys are modified to put back the user's requested object.  

TODO:
- [x] journey modification in new default
- [x] experimental


source/jormungandr/jormungandr/interfaces/v1/transform_id.py
```diff
@@ -31,13 +31,4 @@
...
-    if len(splitted_coord) == 2:
-        try:
-            # check if lon and lat are convertible to float
-            float(splitted_coord[0])
-            float(splitted_coord[1])
-            return "coord:" + id.replace(";", ":")
-        except ValueError:
-            pass
-    return id
+    return id  # TODO remove the call to this method
```
antoine-de, line 34: Note: this will be done in the PR removing completly `coord:`

source/jormungandr/tests/journey_common_tests.py
```diff
@@ -963,48 +963,6 @@ def test_datetime_error(self):
             assert error["message"] == "Unable to parse datetime, Not naive datetime (tzinfo is already set)"
             assert error["id"] == "unable_to_parse"
 
-    def test_journeys_without_show_codes(self):
```
antoine-de, line 966: those have been removed because we don't care about show_codes anymore and this was causing problems because in the dataset there and a stop_point AND a stop_area called 'A', and since we do now an autocomplete query we had the stop_area as the from

source/jormungandr/jormungandr/utils.py
```diff
@@ -55,6 +55,15 @@ def get_uri_pt_object(pt_object):
 def kilometers_to_meters(distance):
     return distance * 1000.0
 
+def is_coord(uri):
+    # for the moment we do a simple check
+    return uri.count(';') == 1
```
TeXitoi: I don't like that.

source/jormungandr/jormungandr/utils.py
```diff
@@ -236,7 +245,7 @@ def add_elt(name, elt):
 
         add_elt(elem[0], elem[1])
 
-
+
```
TeXitoi: Spaces

source/jormungandr/tests/journey_common_tests.py
```diff
@@ -348,7 +348,7 @@ def test_journeys_do_not_loose_precision(self):
 
         assert(len(response['journeys']) > 0)
         for j in response['journeys']:
-            assert(j['sections'][0]['from']['id'] == id)
+            eq_(j['sections'][0]['from']['id'], id)
```
TeXitoi: remove the change or change all the function.

source/kraken/worker.cpp
```diff
@@ -43,6 +43,7 @@ www.navitia.io
 #include "calendar/calendar_api.h"
 #include "routing/raptor.h"
 #include "type/meta_data.h"
+#include <regex>
```
TeXitoi: useless include

source/kraken/worker.h
```diff
@@ -48,6 +48,7 @@ namespace routing{
 
 #include <memory>
 #include <limits>
+#include <boost/optional.hpp>
```
TeXitoi: useless include

source/jormungandr/jormungandr/autocomplete/abstract_autocomplete.py
```diff
@@ -44,13 +45,34 @@ def get(self, query, instance):
...
+        if nothing is found, raise an ObjectUnkown exception
+        :return the list of objects found
+        """
         pass
 
     @abstractmethod
     def geo_status(self, instance):
         pass
 
+    def get_object_by_uri(self, uri, instance=None, current_datetime=None):
```
xlqian, line 61: I was thinking of suggesting doing `def __get__object_by_uri` which is too ugly to do, so i searched on google and find this how to prevent function from being overrided [link](http://stackoverflow.com/a/2425818/1614576)

source/jormungandr/jormungandr/interfaces/v1/fields.py
```diff
@@ -554,9 +554,15 @@ def format(self, value):
     "text_color": fields.String(attribute="text_color"),
 }
 
+
+class DoubleToStringField(fields.Raw):
+    def format(self, value):
+        # we don't want to loose precision while converting a double to string
+        return "{:.16g}".format(value)
```
xlqian, line 561: 👍

source/type/type.cpp
```diff
@@ -945,28 +945,54 @@ std::string to_string(ExceptionDate::ExceptionType t) {
...
-               this->coordinates.set_lat(boost::lexical_cast<double>(uri.substr(pos2+1)));
-           }
-       }catch(boost::bad_lexical_cast){
-           this->coordinates.set_lon(0);
-           this->coordinates.set_lat(0);
-       }
-   }
+    if (type == Type_e::Address){
+        std::vector<std::string> vect;
+        vect = split_string(uri, ":");
```
xlqian: auto vect = split_string(uri, ":");

source/jormungandr/jormungandr/scenarios/new_default.py
```diff
@@ -673,6 +700,21 @@ def fill_journeys(self, request_type, api_request, instance):
 
         krakens_call = get_kraken_calls(api_request)
 
+        # sometimes we need to change the entrypoint id (eg if the id is from another autocomplete system)
+        # we store the original requested id to put them back in the journey's response
+        api_request['requested_entrypoint'] = {
```
xlqian: it seems this field is not used... can you remove this?

TeXitoi (approved):

xlqian (approved):

PR #2043, opened 2017-03-30
kraken: added missing include

datanel:


pbougue (changes requested):
will need utils update when https://github.com/CanalTP/utils/pull/49 merged

pbougue (approved):

PR #2044, opened 2017-03-31
Doc: Adding RailShuttle to physical_mode list

prhod:


TeXitoi:
And which picto I put in navitia-playground for that? How use that?

documentation/ntfs/ntfs_changelog.md
```diff
@@ -43,3 +43,4 @@
     * ajout des temps d'embarquement et de débarquement dans le fichier stop_times.txt.
 * Version 0.6.3 du 25/01/2017
     * ajout du champ dataset_extrapolation dans le fichier datasets.txt
+* Version 0.6.4 du 31/03/2017
```
pbougue, line 46: missing the change itself, no?

pbougue: (side idea) And maybe we should follow the "breaking.feature.bugfix" versionning thing from : semver.org

prhod: changelog fixed. I used the 3rd identifier because there is no real change in the spec

pbougue (changes requested):
Looks like you commited outdated submodules files (and http url), is it intentional?

prhod:
Sorry for the miss, I fixed the PR

pbougue (approved):

PR #2048, opened 2017-04-06
Jormungandr: add shape param in places for bragi

TeXitoi:


source/jormungandr/tests/autocomplete_tests.py
```diff
@@ -114,3 +114,19 @@ def test_place_uri_coords(self):
         places = get_not_null(response, 'places')
         assert len(places) == 1
         assert places[0]['id'] == 'stop_point:Becharles'
+
+    def test_place_shape(self):
```
antoine-de: can you add some comments ?

antoine-de: and maybe split it in multiple tests ?

source/navitiacommon/navitiacommon/parser_args_type.py
```diff
@@ -82,26 +82,24 @@ def to_return(value):
...
+        return 'valid' in valid and (valid['valid'] == 'yes' or valid['valid'] == '')
+
+    if value:
+        if not isinstance(value, dict):
+            raise ValueError('invalid json')
+
+        if not is_geometry_valid(value) :
+            raise ValueError('invalid geojson')
+
+        geometry= value.get('geometry', {}).get('type')
```
antoine-de: spacing

source/tyr/tyr/resources.py
```diff
@@ -657,6 +657,9 @@ def put(self, user_id):
         if not billing_plan:
             return ({'error': 'billing_plan doesn\'t exist'}, 400)
 
+        if args['shape'] == {}:
+            args['shape'] = ujson.loads(user.shape)
```
antoine-de: isn't this done by the default value ?

TeXitoi: No, that's the tricky special case of the empty object saying "no modification", thus you don't have default value as you provided the empty object

PR #2054, opened 2017-04-11
TYR : Reimport last datasets stops to mimir

kadhikari:
Concerned Ticket: http://jira.canaltp.fr/browse/NAVP-654

### DIFF
--- a/source/navitiacommon/navitiacommon/models.py
+++ b/source/navitiacommon/navitiacommon/models.py
@@ -331,13 +331,16 @@ class Instance(db.Model):
         if jobs:
             self.jobs = jobs
 
-    def last_datasets(self, nb_dataset=1):
+    def last_datasets(self, nb_dataset=1, family_type=None):
         """
         return the n last dataset of each family type loaded for this instance
         """
         family_types = db.session.query(func.distinct(DataSet.family_type)) \
-            .filter(Instance.id == self.id) \
-            .all()
+            .filter(Instance.id == self.id)
+        if family_type:
+            family_types = family_types.filter(DataSet.family_type == family_type)
+
+        family_types = family_type.all()
 
         result = []
         for family_type in family_types:
--- a/source/navitiacommon/navitiacommon/models.py
+++ b/source/navitiacommon/navitiacommon/models.py
@@ -340,7 +340,7 @@ class Instance(db.Model):
         if family_type:
             family_types = family_types.filter(DataSet.family_type == family_type)
 
-        family_types = family_type.all()
+        family_types = family_types.all()
 
         result = []
         for family_type in family_types:
--- a/source/navitiacommon/navitiacommon/models.py
+++ b/source/navitiacommon/navitiacommon/models.py
@@ -335,12 +335,12 @@ class Instance(db.Model):
         """
         return the n last dataset of each family type loaded for this instance
         """
-        family_types = db.session.query(func.distinct(DataSet.family_type)) \
+        query = db.session.query(func.distinct(DataSet.family_type)) \
             .filter(Instance.id == self.id)
         if family_type:
-            family_types = family_types.filter(DataSet.family_type == family_type)
+            query = query.filter(DataSet.family_type == family_type)
 
-        family_types = family_types.all()
+        family_types = query.all()
 
         result = []
         for family_type in family_types:
--- a/source/tyr/tyr/binarisation.py
+++ b/source/tyr/tyr/binarisation.py
@@ -574,18 +574,21 @@ def osm2mimir(self, autocomplete_instance, filename, job_id, dataset_uid):
 
 
 @celery.task(bind=True)
-def stops2mimir(self, instance_config, input, job_id, dataset_uid):
+def stops2mimir(self, instance_config, input, job_id=None, dataset_uid=None):
     """
     launch stops2mimir
 
     Note: this is temporary, this will be done by tartare when tartare will be available
     """
-    job = models.Job.query.get(job_id)
-    instance = job.instance
-
-    logger = get_instance_logger(instance)
+    if job_id:
+        job = models.Job.query.get(job_id)
+        instance = job.instance
+        logger = get_instance_logger(instance)
+    else:
+        logger = logging.getLogger("autocomplete")
     cnx_string = current_app.config['MIMIR_URL']
 
+    logger.exception('')
     working_directory = os.path.dirname(input)
 
     stops_file = os.path.join(working_directory, 'stops.txt')
@@ -601,6 +604,8 @@ def stops2mimir(self, instance_config, input, job_id, dataset_uid):
             raise ValueError('stops2mimir failed')
     except:
         logger.exception('')
-        job.state = 'failed'
-        models.db.session.commit()
+        if job_id:
+            job.state = 'failed'
+            models.db.session.commit()
+
         raise
--- a/source/tyr/tyr/binarisation.py
+++ b/source/tyr/tyr/binarisation.py
@@ -580,6 +580,7 @@ def stops2mimir(self, instance_config, input, job_id=None, dataset_uid=None):
 
     Note: this is temporary, this will be done by tartare when tartare will be available
     """
+    # We don't have job_id while doing a reimport of all instances with import_stops_in_mimir = true
     if job_id:
         job = models.Job.query.get(job_id)
         instance = job.instance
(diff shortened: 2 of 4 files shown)


source/tyr/tyr/command/import_last_autocomplete_dataset.py
```diff
@@ -56,12 +56,38 @@ def import_last_autocomplete_dataset(instance_name, wait=True):
...
+    if wait and future:
+        future.wait()
+    logger.info('last datasets reimport finished for %s', instance.name)
+
+
+@manager.command
 def refresh_autocomplete_data(wait=True):
     """
-    reimport all the last autocomplete datas
+    reimport all the last autocomplete datas alongwith the last dataset of all instances
```
antoine-de: missing space alongwith

source/tyr/tyr/command/import_last_autocomplete_dataset.py
```diff
@@ -56,12 +56,38 @@ def import_last_autocomplete_dataset(instance_name, wait=True):
...
+    instance = models.Instance.query_existing().filter_by(name=instance_name).first()
+
+    if not instance:
+        raise Exception("cannot find instance {}".format(instance_name))
+
+    files = [d.name for d in instance.last_datasets(1)]
+    logger = logging.getLogger(__name__)
+    logger.info('we reimport to mimir the last dataset of %s, composed of: %s',
+                instance.name, files)
+    future = tasks.import_in_mimir(files, instance, async=wait)
```
antoine-de: this won't work, you'll send some osm/synonym data too mimir too, and the function `import_in_mimir` does not check the type of data

source/tyr/tyr/tasks.py
```diff
@@ -235,6 +235,34 @@ def import_autocomplete(files, autocomplete_instance, async=True, backup_file=Tr
...
 @celery.task()
+def import_in_mimir(files, instance, async=True):
+    """
+    Import pt data stops to autocomplete
+    """
+    current_app.logger.debug("Import pt data to mimir")
+    job = models.Job()
+    actions = []
+    job.instance = instance
+    job.send_to_mimir = True
```
antoine-de: this is useless as it is no ?

source/tyr/tyr/tasks.py
```diff
@@ -235,6 +235,34 @@ def import_autocomplete(files, autocomplete_instance, async=True, backup_file=Tr
 
 
 @celery.task()
+def import_in_mimir(files, instance, async=True):
```
antoine-de: I'm not convinced by this, you create a new `Job` and attach it to the instance, so the job will be use in the next import no ?

source/tyr/tyr/binarisation.py
```diff
@@ -574,18 +574,21 @@ def osm2mimir(self, autocomplete_instance, filename, job_id, dataset_uid):
...
-    logger = get_instance_logger(instance)
+    if job_id:
+        job = models.Job.query.get(job_id)
+        instance = job.instance
+        logger = get_instance_logger(instance)
+    else:
+        logger = logging.getLogger("autocomplete")
     cnx_string = current_app.config['MIMIR_URL']
 
+    logger.exception('')
```
antoine-de: debug line

source/tyr/tyr/binarisation.py
```diff
@@ -574,18 +574,21 @@ def osm2mimir(self, autocomplete_instance, filename, job_id, dataset_uid):
...
     """
     launch stops2mimir
 
     Note: this is temporary, this will be done by tartare when tartare will be available
     """
-    job = models.Job.query.get(job_id)
-    instance = job.instance
-
-    logger = get_instance_logger(instance)
+    if job_id:
```
antoine-de: maybe add a comment telling why this can be null

source/tyr/tyr/command/import_last_autocomplete_dataset.py
```diff
@@ -71,10 +71,9 @@ def import_last_stop_dataset(instance_name, wait=True):
     logger = logging.getLogger(__name__)
     logger.info('we reimport to mimir the last dataset of %s, composed of: %s',
                 instance.name, files)
-    future = tasks.import_in_mimir(files, instance, async=wait)
+    future = tasks.import_in_mimir(files, instance)
     if wait and future:
         future.wait()
-    logger.info('last datasets reimport finished for %s', instance.name)
```
antoine-de: having a log line at the end of the process might be handy

source/navitiacommon/navitiacommon/models.py
```diff
@@ -331,13 +331,16 @@ def __init__(self, name=None, is_free=False, authorizations=None,
         if jobs:
             self.jobs = jobs
 
-    def last_datasets(self, nb_dataset=1):
+    def last_datasets(self, nb_dataset=1, family_type=None):
         """
         return the n last dataset of each family type loaded for this instance
         """
         family_types = db.session.query(func.distinct(DataSet.family_type)) \
```
antoine-de: I think it would be easier to read if the first variable was named `query`

source/tyr/tyr/command/import_last_autocomplete_dataset.py
```diff
@@ -56,12 +56,38 @@ def import_last_autocomplete_dataset(instance_name, wait=True):
...
+    reimport the last datasets of of all instances with import_stops_in_mimir = true
+
+    you can set wait=False if you don't want to wait for the result
+    """
+    instance = models.Instance.query_existing().filter_by(name=instance_name).first()
+
+    if not instance:
+        raise Exception("cannot find instance {}".format(instance_name))
+
+    files = [d.name for d in instance.last_datasets(1, 'pt')]
```
antoine-de: I think it's more readable to name the parameters in this case

source/tyr/tyr/tasks.py
```diff
@@ -235,6 +235,25 @@ def import_autocomplete(files, autocomplete_instance, async=True, backup_file=Tr
...
+    """
+    Import pt data stops to autocomplete
+    """
+    current_app.logger.debug("Import pt data to mimir")
+    actions = []
+    instance_config = load_instance_config(instance.name)
+
+    for _file in files:
```
antoine-de: I don't really think that this makes sens to pass a function here, I think it's better to check beforehand that there is only one element and give only one file here
and thus you can just say (check if this works)

```python
    action = stops2mimir.si(instance_config, _file, None)
    if async:
        return actions.delay()
    else:
        # all job are run in sequence and import_in_mimir will only return when all the jobs are finish
        return action.apply()
```

source/tyr/tyr/tasks.py
```diff
@@ -235,6 +235,25 @@ def import_autocomplete(files, autocomplete_instance, async=True, backup_file=Tr
...
+        actions.append(stops2mimir.si(instance_config, _file, None))
```
antoine-de: you can remove the `None` no ?

source/tyr/tyr/binarisation.py
kadhikari: ok

source/tyr/tyr/binarisation.py
kadhikari: ok

source/tyr/tyr/command/import_last_autocomplete_dataset.py
kadhikari: ok

source/navitiacommon/navitiacommon/models.py
kadhikari: ok

source/tyr/tyr/tasks.py
kadhikari: yes

source/tyr/tyr/tasks.py
```diff
@@ -235,6 +235,25 @@ def import_autocomplete(files, autocomplete_instance, async=True, backup_file=Tr
...
+    """
+    current_app.logger.debug("Import pt data to mimir")
+    actions = []
+    instance_config = load_instance_config(instance.name)
+
+    for _file in files:
+        filename = _file
+
+    if async:
+        return chain(stops2mimir.si(instance_config, filename)).delay()
```
antoine-de: you need the chain ?

source/tyr/tyr/tasks.py
```diff
@@ -235,6 +235,25 @@ def import_autocomplete(files, autocomplete_instance, async=True, backup_file=Tr
...
+def import_in_mimir(files, instance, async=True):
+    """
+    Import pt data stops to autocomplete
+    """
+    current_app.logger.debug("Import pt data to mimir")
+    actions = []
+    instance_config = load_instance_config(instance.name)
+
+    for _file in files:
+        filename = _file
```
antoine-de: you can't do this, you need to pass only one file as argument of import_in_mimir

antoine-de (approved):

PR #2047, opened 2017-04-05
Jormungandr: use serpy for api places (geocodejson)

datanel:
### TODO
- [x] refactoring 
- [x] add warning property
```json
"warnings": [
    {
        "message": "This service is under construction. You can help through github.com/CanalTP/navitia",
        "id": "beta_endpoint"
    }
]
```

PR #2055, opened 2017-04-13
Jormungandr: /schema to expose JSON Schema of resources responses

ooga:
this PR aims to provide an automated swagger definition of the navitia's API.

# key architecture
* add a new route:
  * `api.navitia.io/v1/<some_route>` will be described with the http verb `OPTIONS`
             (eg `curl -X OPTIONS api.navitia.io/v1/coverage/bob/stop_points/toto` will give the schema of the called API
  * `api.navitia.io/v1/schema` to get the global schema (of all API described with the `OPTIONS` verb)

* use of flask's parser to describe the input query parameters (`api.navitia.io/v1/places?q=<query_parameter>`)
* use of flask's converter to describe the input path parameters (`api.navitia.io/v1/coverage/<path_parameter>`)
* use of serpy to describe the output
   * we'll use the defined serpy's type when we can
   * if no type is provided, we'll need to define a `schema_type` on the serpy's field
   * all fields can also have a schema_metadata dict to add swagger information (like a validation regex, min, max, ...)

To have a rendered view of the current work:
copy the documentation in  `documentation/rfc/schema/global_swagger.json` in the [swagger editor](http://editor.swagger.io/#/)

TODO:
* authentication description
* describe all the navitia API :stuck_out_tongue_winking_eye: 
* add better descriptions


source/jormungandr/jormungandr/interfaces/v1/serializer/JSONSchema.py
```diff
@@ -0,0 +1,165 @@
...
+    'int': {
+        'type': 'integer'
+    },
+    'bool': {
+        'type': 'boolean',
+    },
+}
+
+class JSONSchema(Serializer):
+    properties = fields.MethodField('get_properties')
```
antoine-de: I think you can remove the `get_properties` it's the default value (`get_` + name of the field

source/jormungandr/jormungandr/interfaces/v1/serializer/JSONSchema.py
```diff
@@ -0,0 +1,165 @@
...
+    type = fields.MethodField('get_constant_object')
```
antoine-de: if you wait for https://github.com/CanalTP/navitia/pull/2047 , you'll get a `LiteralField` to do this

source/jormungandr/jormungandr/interfaces/v1/JSONSchema.py
```diff
@@ -0,0 +1,45 @@
...
+
+class ApiSchema(Serializer):
+    ptr = api.PTReferentialSerializer()
+
+class Schema(ResourceUri):
+    def __init__(self, **kwargs):
+        ResourceUri.__init__(self, authentication=False, links=False, **kwargs)
+
+    def get(self):
+        return JSONSchema.JSONSchema(ApiSchema(), root=True).data, 200
```
antoine-de: missing new line

source/jormungandr/jormungandr/interfaces/v1/JSONSchema.py
```diff
@@ -0,0 +1,45 @@
...
+
+from jormungandr.interfaces.v1.ResourceUri import ResourceUri
+from jormungandr.interfaces.v1.serializer import api, JSONSchema
+
+from serpy import Serializer
+
+class ApiSchema(Serializer):
+    ptr = api.PTReferentialSerializer()
+
+class Schema(ResourceUri):
```
antoine-de: why do you inherit from ResourceUri ?

source/jormungandr/jormungandr/interfaces/v1/serializer/JSONSchema.py
```diff
@@ -0,0 +1,165 @@
...
+    required = fields.MethodField('get_required')
+    # Weird name to ensure it will be processed a the end
+    _definitions = fields.MethodField('get_definitions', display_none=False, label='definitions')
+
+    def __init__(self, instance=None, many=False, data=None, context=None, root=False, **kwargs):
```
antoine-de: can't you just use `root` ?

```python
def __init__(self, root, **kwargs):
    super(JSONSchema, self).__init__(**kwargs)
    self.root = root
    self.definitions = {}
```

source/jormungandr/jormungandr/interfaces/v1/serializer/JSONSchema.py
```diff
@@ -0,0 +1,165 @@
...
+        return 'object'
+
+    def get_definitions(self, obj):
+        if self.root:
+            properties = {}
+            definitions = self.definitions
+
+            while definitions:
+                field_name, field = definitions.popitem()
+                schema, nestedDefinitions = self.__class__._from_nested_schema(field)
```
antoine-de: it python you can just do `self.from_nested_schema(field)` (even if it's a classmethod)

source/jormungandr/jormungandr/interfaces/v1/serializer/JSONSchema.py
```diff
@@ -0,0 +1,165 @@
...
+        mapping[fields.BoolField] = 'bool'
+        # Map the 2 next fields to avoid exceptions for now
+        mapping[fields.Field] = 'str'
+        # mapping[fields.MethodField] = 'method'
+        properties = {}
+
+        for field_name, field in obj._field_map.items():
+            searchObj = field
+            preTypeMethodName = '_jsonschema_pre_type_mapping'
+            if field.__class__ == fields.MethodField:
```
antoine-de: I think it's more pythonic to use `isinstance`, especially with Serpy which extensively use (multiple) inheritance

source/jormungandr/jormungandr/interfaces/v1/serializer/JSONSchema.py
```diff
@@ -0,0 +1,165 @@
...
+
+    def get_properties(self, obj):
+        mapping = {}
+        mapping[fields.StrField] = 'str'
+        mapping[fields.IntField] = 'int'
+        mapping[fields.FloatField] = 'float'
+        mapping[fields.BoolField] = 'bool'
+        # Map the 2 next fields to avoid exceptions for now
+        mapping[fields.Field] = 'str'
+        # mapping[fields.MethodField] = 'method'
```
antoine-de: by curiosity how do you want to handle those ?

source/jormungandr/jormungandr/interfaces/v1/serializer/JSONSchema.py
```diff
@@ -0,0 +1,165 @@
...
+        properties = {}
+
+        for field_name, field in obj._field_map.items():
```
antoine-de: I don't really understand what you are doing here

source/jormungandr/jormungandr/interfaces/v1/serializer/JSONSchema.py
```diff
@@ -0,0 +1,165 @@
...
+                schema, definitions = self.__class__._from_nested_schema(field, onlyRef=True)
+            else:
+                raise ValueError('unsupported field type %s for attr %s in object %s' % (field, field_name, obj.__class__.__name__))
+
+            name = field.label or field_name
+            properties[name] = schema
+
+        return properties
+
+    def get_required(self, obj):
```
antoine-de: minor: it's more pythonic to rewrite this:
```python
return [field.label or field_name for field_name, field in obj._field_map.items() if field.required]
```

source/jormungandr/jormungandr/interfaces/v1/serializer/base.py
```diff
@@ -70,6 +70,36 @@ def to_value(self, value):
...
 
+    """
+    Need to find a way to read protobuf
+    """
+    def _jsonschema_type_mapping(self):
+        value = {
+            "type": "string"
+        }
+        enum = []
+        if self.attr == 'id':
```
antoine-de: I'm not a fan of this, can't we do this without the hardcoded list ?

source/jormungandr/jormungandr/interfaces/v1/serializer/pt.py
```diff
@@ -242,6 +270,9 @@ class RouteSerializer(GenericSerializer):
     line = serpy.MethodField(display_none=False)
     stop_points = StopPointSerializer(many=True, display_none=False)
 
+    def get_line_jsonschema_pre_type_mapping(self):
```
antoine-de: I think it's better to add an attribute to MethodField (but we either need to change serpy or to use a custom MethodField)

source/jormungandr/jormungandr/interfaces/v1/serializer/time.py
```diff
@@ -48,13 +48,25 @@ def to_value(self, value):
...
     """
     custom date format from timestamp
     """
     def to_value(self, value):
         return timestamp_to_str(value)
 
+    def _jsonschema_type_mapping(self):
+        return {
+            'type': 'string',
+            'pattern': '\d{4}\d{2}\d{2}T\d{2}\d{2}\d{2}'
```
antoine-de: I don't really know how to do this, but I would prefer to abstract a bit the JsonSchema stuff

source/jormungandr/jormungandr/interfaces/v1/serializer/JSONSchema.py
```diff
@@ -0,0 +1,165 @@
...
+#
+# Stay tuned using
+# twitter @navitia
+# IRC #navitia on freenode
+# https://groups.google.com/d/forum/navitia
+# www.navitia.io
+
+from serpy import fields, Serializer
+
+TYPE_MAP = {
```
antoine-de: I think it would be better to remove the `type`, and add it at the TYPE_MAP use

source/jormungandr/jormungandr/interfaces/v1/serializer/JSONSchema.py
```diff
@@ -0,0 +1,165 @@
...
+        'type': 'boolean',
+    },
+}
+
+class JSONSchema(Serializer):
+    properties = fields.MethodField('get_properties')
+    type = fields.MethodField('get_constant_object')
+    required = fields.MethodField('get_required')
+    # Weird name to ensure it will be processed a the end
+    _definitions = fields.MethodField('get_definitions', display_none=False, label='definitions')
```
antoine-de: I don't really understand why is done, but I think we can do something else to not be dependant of the process order

antoine-de (commented):
:+1: 

It needs  some polish on the edges but the work is really promising!

source/jormungandr/jormungandr/interfaces/v1/serializer/JSONSchema.py
ooga: We call ```method.__name__ + '_jsonschema_pre_type_mapping' ``` to get a description of the output
like https://github.com/CanalTP/navitia/pull/2055/files#diff-8616a8e682cc2f9747d29e22bea695c7R73

source/jormungandr/jormungandr/interfaces/v1/serializer/JSONSchema.py
ooga: I search for a method to call for describing output
- method ```_jsonschema_pre_type_mapping``` for a description to be processed
- method ```_jsonschema_type_mapping``` for raw jsonschema description

source/jormungandr/jormungandr/interfaces/v1/JSONSchema.py
ooga: Because I just do a dumb copy/paste :)

source/jormungandr/jormungandr/interfaces/v1/serializer/base.py
ooga: 👍 I would find a way to read protobuf

source/jormungandr/jormungandr/interfaces/v1/JSONSchema.py
```diff
@@ -0,0 +1,45 @@
...
+# https://groups.google.com/d/forum/navitia
+# www.navitia.io
+
+from flask.ext.restful import Resource
+from jormungandr.interfaces.v1.serializer import api, jsonschema
+import serpy
+
+class ApiSchema(serpy.Serializer):
+    ptr = api.PTReferentialSerializer()
+    test = jsonschema.Field(schema_type=int)
```
pbougue: leftover from a test?

source/jormungandr/jormungandr/interfaces/v1/serializer/jsonschema/serializer.py
```diff
@@ -0,0 +1,163 @@
...
+    'bool': {
+        'type': 'boolean',
+    },
+}
+
+class JsonSchemaSerializer(serpy.Serializer):
+    properties = serpy.MethodField()
+    type = serpy.MethodField()
+    required = serpy.MethodField(display_none=False)
+    # Weird name to ensure it will be processed a the end
```
pbougue: `at` the end

pbougue: and not fan of these tricks, but not sure there's better

antoine-de:
for discussion purposes I think we should have a concrete proposition of how it will impact the code (I think it will help others understand the needs too)

The idea is to change the serpy serializer to add all the information needed to understand (by a machine and hopefully by a human) the output of navitia

as an example I propose a virtual 2 serializer (with all the different case I can think of, feel free to react and propose other corner cases)

Base example:
```python
class SeveritySerializer(PbNestedSerializer):
    name = serpy.Field()
    effect = EnumListField(attr='effect_types')
    color = serpy.Field()
    priorities = serpy.Field(many=True)


class SeverityContainer(GenericSerializer):
    quality = serpy.MethodField()
    severities = SeveritySerializer(many=True, display_none=False)
    updated_at = DateTimeField()

    def get_quality():
        # lots of obscure stuff
```

What we need for each field:
 * type
 * description
 * arity
 * whether the field is required or optional
 * (optional): more constraints like:
    * regexp
    * values if enum

first proposal:
add `type`and eventually `required` as first class citizen in the fields and a dict (or an object if it's not much noise) for the other fields

```python
class SeveritySerializer(PbNestedSerializer):
    name = Field(type=string, schema={'description': 'The name of the severity})
    effect = EnumListField(attr='effect_types', schema={'description': 'blabla'})  # auto generation of the enum values from the protobuf + definition in EnumListField of the type
    color = Field(type=Color, schema={'description': 'bla'})  # custom type for better json schema ?
    priorities = Field(many=True, type=int, schema={'constraints': {'multiple_of': 10}})


class SeverityContainer(GenericSerializer):
    quality = MethodField(type=QualitySerializer, required=False)
    severities = SeveritySerializer(many=True, display_none=False)
    updated_at = DateTimeField()  # definition in DateTimeField of a custom type for better json schema ?

    def get_quality():
        # lots of obscure stuff
```

another approach would be to use decorators to make it clearer that the fields are used only for documentation purpose:

```python
class SeveritySerializer(PbNestedSerializer):
    @schema(description='The name of the severity', type=string)
    name = Field()

    @schema(description='bla')
    effect = EnumListField(attr='effect_types')  # auto generation of the enum values from the protobuf + definition in EnumListField of the type

    @schema(type=Color, description='bla')
    color = Field()

    @schema(type=int, description='bla', multiple_of=10)
    priorities = Field(many=True)
```

It might be a bit clearer, but since we need fields from both the decorator and the Field, it might be a bit more difficult to read...

source/jormungandr/jormungandr/interfaces/v1/JSONSchema.py
ooga: Yep, I'm going to list all endpoints and params from this endpoint

source/jormungandr/jormungandr/interfaces/v1/serializer/jsonschema/fields.py
```diff
@@ -0,0 +1,95 @@
...
+        self.schema_type = schema_type
+        self.schema_metadata = schema_metadata
+
+
+class StrField(serpy.StrField):
+    """
+    A :class:`StrField` that hold metadata for schema.
+    """
+
+    def __init__(self, schema_metadata={}, **kwargs):
```
antoine-de: isn't it possible to store the additional field at class level ? it would save memory and be faster

source/jormungandr/jormungandr/interfaces/v1/serializer/jsonschema/serializer.py
```diff
@@ -0,0 +1,164 @@
...
+    },
+    'int': {
+        'type': 'integer'
+    },
+    'bool': {
+        'type': 'boolean',
+    },
+}
+
+class JsonSchemaSerializer(serpy.Serializer):
```
antoine-de: I'm not convinced by the fact that you model this as a serpy.Serializer, wouldn't it be better to have an intermediate representation ?

@kinnou02 @TeXitoi @pbougue ?

source/jormungandr/jormungandr/interfaces/v1/StatedResource.py
```diff
@@ -61,3 +62,15 @@ def wrapper(*args, **kwargs):
                 register_used_coverages([region])
             return result
         return wrapper
+
+    def dispatch_request(self, *args, **kwargs):
```
antoine-de: hum I don't understand why you need this

source/jormungandr/jormungandr/interfaces/v1/serializer/fields.py
```diff
@@ -50,20 +73,20 @@ def to_value(self, value):
...
+    name = jsonschema.Field(schema_type=str)
+    value = jsonschema.Field(schema_type=str)
 
 
 class FeedPublisherSerializer(PbNestedSerializer):
-    id = serpy.Field()
-    name = serpy.Field()
-    url = serpy.Field()
-    license = serpy.Field()
+    id = serpy.StrField()
```
antoine-de, line 59: you need a Field there (and all the below fields)

source/jormungandr/jormungandr/interfaces/v1/serializer/fields.py
```diff
@@ -72,20 +95,27 @@ class CoordSerializer(serpy.Serializer):
...
-class FirstCommentField(serpy.Field):
+class FirstCommentField(jsonschema.Field):
     """
     for compatibility issue we want to continue to output a 'comment' field
     even if now we have a list of comments, so we take the first one
     """
+
+    def __init__(self, schema_metadata={}, **kwargs):
+        schema_type = CommentSerializer()
+        for key, value in kwargs.items() :
```
antoine-de: do you need this ?

source/jormungandr/jormungandr/interfaces/v1/serializer/fields.py
```diff
@@ -97,17 +127,53 @@ def to_value(self, item):
...
 
 
-class LinkSerializer(serpy.Field):
+class LinkSerializer(jsonschema.Field):
     """
     Add link to disruptions on a pt object
     """
+
+    def __init__(self, schema_metadata={}, **kwargs):
+        schema_metadata.update({
```
antoine-de: I think we should never have raw jsonschema

source/jormungandr/jormungandr/interfaces/v1/StatedResource.py
ooga: This is to deactivate decorators for OPTIONS method.
The decorators are actually expect the GET result in input

source/jormungandr/jormungandr/interfaces/v1/serializer/jsonschema/fields.py
ooga: Maybe for the type but metadata are specific to the instance

source/jormungandr/jormungandr/interfaces/v1/StatedResource.py
antoine-de: I think we can use: https://github.com/flask-restful/flask-restful/pull/532/files

modifying  `self` seems dangerous

antoine-de:
this PR aims to provide a automated swagger definition of the navitia's API.

# key architecture
* add a new route:
  * `api.navitia.io/v1/<some_route>` will be described with the http verb `OPTIONS`
             (eg `curl -X OPTION api.navitia.io/v1/coverage/bob/stop_points/toto` will give the schema of the called API
  * `api.navitia.io/v1/schema` to get the global schema (of all API described with the `OPTIONS` verb)

* use of flask's parser to describe the input query parameters (`api.navitia.io/v1/places?q=<query_parameter>`)
* use of flask's converter to describe the input path parameters (`api.navitia.io/v1/coverage/<path_parameter>`)
* use of serpy to describe the output
   * we'll use the defined serpy's type when we can
   * if no type is provided, we'll need to define a `schema_type` on the serpy's field
   * all fields can also have a schema_metadata dict to add swagger information (like a validation regex, min, max, ...)

To have a rendered view of the current work:
copy the documentation in  `documentation/rfc/schema/global_swagger.json` in the [swagger editor](http://editor.swagger.io/#/)

TODO:
* describe all the navitia API :stuck_out_tongue_winking_eye:

source/jormungandr/jormungandr/interfaces/v1/serializer/pt.py
```diff
@@ -264,26 +273,32 @@ class PlaceSerializer(GenericSerializer):
...
     codes = CodeSerializer(many=True, display_none=False)
     direction = PlaceSerializer()
     geojson = MultiLineStringField(display_none=False)
-    links = LinkSerializer(attr='impact_uris')
-    line = serpy.MethodField(display_none=False)
+    links = LinkSerializer(attr='impact_uris', display_none=True)
+    line = jsonschema.MethodField(schema_type='get_line_jsonschema')
     stop_points = StopPointSerializer(many=True, display_none=False)
 
+    def get_line_jsonschema(self):
```
kinnou02: this seems to be deprecated

source/jormungandr/jormungandr/interfaces/v1/swagger_schema.py
```diff
@@ -0,0 +1,350 @@
...
+    schema = {
+        "type": "object",
+        "properties": properties_schema,
+    }
+    if required_properties:  # swagger is a bit rigid not to have empty required properties
+        schema['required'] = required_properties
+    return schema, definitions
+
+
+def get_schema_properties(serializer):
```
kinnou02, line 171: A little bit of comment would help here

TeXitoi:
CORS in coverage is now broken :-(

PR #2059, opened 2017-04-18
Kraken: use boost for regex

pbougue:
On gcc (4.7) for debian 7, std::regex is not stable, so it's crashing tests


antoine-de (approved):

pbougue:
Jenkins, test this please

PR #2062, opened 2017-04-20
jormungandr: priority to journey with bike_in_pt while eliminating similar journey

kadhikari:
Concerned ticket: http://jira.canaltp.fr/browse/NAVITIAII-2221

Hand-tested only (OK whether tagged `bike_in_pt` or not)

### DIFF
--- a/source/jormungandr/jormungandr/scenarios/new_default.py
+++ b/source/jormungandr/jormungandr/scenarios/new_default.py
@@ -800,7 +800,6 @@ class Scenario(simple.Scenario):
         resp = []
         logger = logging.getLogger(__name__)
         futures = []
-        self.nb_kraken_calls = 0
 
         def worker(dep_mode, arr_mode, instance, request, request_id):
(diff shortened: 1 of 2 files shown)


source/jormungandr/jormungandr/scenarios/new_default.py
```diff
@@ -800,6 +800,7 @@ def call_kraken(self, request_type, request, instance, krakens_call):
         resp = []
         logger = logging.getLogger(__name__)
         futures = []
+        self.nb_kraken_calls = 0
```
antoine-de: why did you add this ?

kadhikari:
jenkins test this please

source/jormungandr/jormungandr/scenarios/new_default.py
kadhikari: This initializes the counter of kraken call and is not concerned to bike_in_pt.

xlqian:
Jenkins, test this please!

TeXitoi (approved):

TeXitoi (approved):

PR #2063, opened 2017-04-21
Kraken: added a help message

datanel:
`kraken -h`
```shell
Usage:
	kraken --help		Show this message.
	kraken [config_file]	Specify the path of the configuration file (default: kraken.ini in the current path)
```

### DIFF
--- a/source/kraken/kraken_zmq.cpp
+++ b/source/kraken/kraken_zmq.cpp
@@ -1,28 +1,28 @@
 /* Copyright © 2001-2014, Canal TP and/or its affiliates. All rights reserved.
-  
+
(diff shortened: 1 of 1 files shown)


source/kraken/kraken_zmq.cpp
```diff
@@ -39,7 +39,16 @@ www.navitia.io
 #include "utils/init.h"
 #include "kraken_zmq.h"
 #include "utils/zmq.h"
+#include <iostream>
 
+static void show_usage(std::string name)
```
xlqian: const std::string& name

TeXitoi (approved):

PR #2064, opened 2017-04-21
jormungandr: modifications for same_journey_schedules

kadhikari:
1. Only one value of first_section_mode[] and last_section_mode[] if
present in url.

2. StopPoints instead of StopAreas added in allowed_id[]

3. Ticket concerned: https://jira.canaltp.fr/browse/NAVP-657

4. parameter min_nb_transfers added in the links and used to filter journeys to have same journey schedules.

### DIFF
--- a/source/jormungandr/jormungandr/interfaces/v1/Journeys.py
+++ b/source/jormungandr/jormungandr/interfaces/v1/Journeys.py
@@ -294,9 +294,17 @@ class add_journey_href(object):
                 return objects
             for journey in objects[0]['journeys']:
                 args = dict(request.args)
-                ids = {o['stop_point']['stop_area']['id']
+                ids = {o['stop_point']['id']
                        for s in journey.get('sections', []) if 'from' in s
                        for o in (s['from'], s['to']) if 'stop_point' in o}
+
+                #only one first_section_mode is allowed if present in the request
+                #same for last_section_mode also
+                if 'first_section_mode[]' in args:
+                    args['first_section_mode[]'] = journey['sections'][0]['mode']
+                if 'last_section_mode[]' in args:
+                    args['last_section_mode[]'] = journey['sections'][-1]['mode']
+
                 if 'region' in kwargs:
                     args['region'] = kwargs['region']
                 if "sections" not in journey:#this mean it's an isochrone...
--- a/source/jormungandr/jormungandr/interfaces/v1/Journeys.py
+++ b/source/jormungandr/jormungandr/interfaces/v1/Journeys.py
@@ -294,17 +294,9 @@ class add_journey_href(object):
                 return objects
             for journey in objects[0]['journeys']:
                 args = dict(request.args)
-                ids = {o['stop_point']['id']
+                allowed_ids = {o['stop_point']['id']
                        for s in journey.get('sections', []) if 'from' in s
                        for o in (s['from'], s['to']) if 'stop_point' in o}
-
-                #only one first_section_mode is allowed if present in the request
-                #same for last_section_mode also
-                if 'first_section_mode[]' in args:
-                    args['first_section_mode[]'] = journey['sections'][0]['mode']
-                if 'last_section_mode[]' in args:
-                    args['last_section_mode[]'] = journey['sections'][-1]['mode']
-                args['min_nb_transfers'] = journey['nb_transfers']
 
                 if 'region' in kwargs:
                     args['region'] = kwargs['region']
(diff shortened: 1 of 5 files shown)


source/jormungandr/jormungandr/interfaces/v1/Journeys.py
```diff
@@ -294,9 +294,18 @@ def wrapper(*args, **kwargs):
                 return objects
             for journey in objects[0]['journeys']:
                 args = dict(request.args)
-                ids = {o['stop_point']['stop_area']['id']
+                ids = {o['stop_point']['id']
```
pbougue: minor: I'd renamed it `allowed_ids` for clarity

source/jormungandr/jormungandr/interfaces/v1/Journeys.py
```diff
@@ -294,9 +294,18 @@ def wrapper(*args, **kwargs):
...
                        for s in journey.get('sections', []) if 'from' in s
                        for o in (s['from'], s['to']) if 'stop_point' in o}
+
+                #only one first_section_mode is allowed if present in the request
+                #same for last_section_mode also
+                if 'first_section_mode[]' in args:
+                    args['first_section_mode[]'] = journey['sections'][0]['mode']
```
pbougue: Not sure this works for `bss`

source/jormungandr/jormungandr/scenarios/journey_filter.py
```diff
@@ -257,6 +259,22 @@ def _filter_too_much_connections(journeys, instance, request):
                 mark_as_dead(j, "too_much_connections")
 
 
+def _filter_min_transfers(journeys, instance, request):
+    """
+    eliminates journeys with number of connections more then max_nb_transfers among journeys
```
pbougue: Isn't it redundant with `_max_additional_connections`?
Looks like this line is a leftover of comment.
Maybe the 2 functions should be merged in one?

source/jormungandr/jormungandr/interfaces/v1/Journeys.py
TeXitoi: It doesn't work for BSS, it's sure.

source/jormungandr/jormungandr/interfaces/v1/Journeys.py
```diff
@@ -294,9 +294,18 @@ def wrapper(*args, **kwargs):
...
             for journey in objects[0]['journeys']:
                 args = dict(request.args)
-                ids = {o['stop_point']['stop_area']['id']
+                ids = {o['stop_point']['id']
                        for s in journey.get('sections', []) if 'from' in s
                        for o in (s['from'], s['to']) if 'stop_point' in o}
+
+                #only one first_section_mode is allowed if present in the request
+                #same for last_section_mode also
+                if 'first_section_mode[]' in args:
```
TeXitoi: Better to set a first_section_mode[] even if there is no one setted.

source/jormungandr/jormungandr/interfaces/v1/Journeys.py
```diff
@@ -294,9 +294,18 @@ def wrapper(*args, **kwargs):
...
+                    args['first_section_mode[]'] = journey['sections'][0]['mode']
+                if 'last_section_mode[]' in args:
```
TeXitoi: Same here.

source/jormungandr/jormungandr/interfaces/v1/Journeys.py
```diff
@@ -294,9 +294,18 @@ def wrapper(*args, **kwargs):
...
+                    args['last_section_mode[]'] = journey['sections'][-1]['mode']
+                args['min_nb_transfers'] = journey['nb_transfers']
```
TeXitoi: It will not be enough for direct  journey, as there is 0 transfers in direct pt journeys and non pt journeys.

TeXitoi (changes requested):

source/jormungandr/jormungandr/interfaces/v1/Journeys.py
```diff
@@ -294,9 +294,24 @@ def wrapper(*args, **kwargs):
...
+                    args['first_section_mode[]'] = 'bss'
+                else:
+                    args['first_section_mode[]'] = journey['sections'][0]['mode']
+
+                if next(iter(journey['sections'][-2::-1]), {}).get('type', '').startswith('bss'):
+                    args['last_section_mode[]'] = 'bss'
+                else:
+                    args['last_section_mode[]'] = journey['sections'][-1]['mode']
+
+                args['min_nb_transfers'] = journey['nb_transfers']
```
TeXitoi: It will not be enough for direct journey, as there is 0 transfers in direct pt journeys and non pt journeys. I don't have easy solution for that just now.

source/jormungandr/tests/tests_mechanism.py
```diff
@@ -367,6 +367,10 @@ def is_valid_journey_response(self, response, query_str):
                 assert j['links'][0]['type'] == 'journeys'
                 assert '/journeys?' in j['links'][0]['href']
                 assert 'allowed_id%5B%5D=' in j['links'][0]['href']
+                if 'first_secion_mode%5B%5D=' in j['links'][0]['href']:
+                    assert j['links'][0]['href'].count('first_secion_mode%5B%5D=') == 1
```
antoine-de: assert j['links'][0]['href'].count('first_secion_mode%5B%5D=') <= 1 no ?

and can you add a comment ?

source/jormungandr/jormungandr/scenarios/journey_filter.py
```diff
@@ -283,6 +285,21 @@ def _filter_too_much_connections(journeys, instance, request):
                 mark_as_dead(j, "too_much_connections")
 
 
+def _filter_min_transfers(journeys, instance, request):
+    """
+    eliminates journeys with number of connections less then min_nb_transfers among journeys
+    """
+    logger = logging.getLogger(__name__)
+    min_nb_transfers = get_or_default(request, 'min_nb_transfers', 0)
```
pbougue, line 293: On that, a non-ambiguous way may be to have a `min_nb_pt_section` param, maybe deprecate `max_nb_transfers` to a straightly tied (+1) param `max_nb_pt_section`.
It's always that difference between what's easy to integrate and what we think should be shown to traveler.

pbougue (approved):

antoine-de (approved):

TeXitoi (approved):

PR #2069, opened 2017-04-26
Tyr: Update elasticsearch data

azime:
Add a POST for mimir data


source/tyr/tests/integration/datapost_test.py
```diff
@@ -6,14 +6,18 @@
 from tyr import app
 import logging
 
-@pytest.fixture(scope="module")
+@pytest.fixture
 def create_instance_fr():
     with app.app_context():
         instance = models.Instance('fr')
+        ac_instance = models.AutocompleteParameter(name='fr', street='OSM', address='BANO', poi='OSM', admin='OSM',
```
antoine-de: I think it's a good practice not to have the same name for autocomplete instance and navitia instances, it's already confusing :stuck_out_tongue_winking_eye:

source/tyr/tests/integration/datapost_test.py
```diff
@@ -32,12 +36,33 @@ def test_post_pbf(create_instance_fr):
...
 
+
 def test_post_bad_instance(create_instance_fr):
     tester = app.test_client()
     resp = tester.post('/v0/jobs/fr_ko')
     assert resp.status_code == 404
+
+
+def test_post_pbf_autocomplete(create_instance_fr):
+    assert (not os.path.isfile('/tmp/empty_pbf.osm.pbf'))
```
antoine-de: I've seen that this is done in  the other tests, but I don't like this hardcoded path to tmp, I would prefer to use pytest's `tmpdir` (I'm not sure it's possible though)

source/tyr/tests/integration/datapost_test.py
```diff
@@ -32,12 +36,33 @@ def test_post_pbf(create_instance_fr):
...
+
+    filename = 'empty_pbf.osm.pbf'
+    path = os.path.join(os.path.dirname(os.path.dirname(os.path.dirname(__file__))), 'tests/fixtures/', filename)
+    f = open(path, 'rb')
```
antoine-de: I'd prefer to use context manager for file handlers

source/tyr/tyr/helper.py
```diff
@@ -191,3 +192,24 @@ def get_named_arg(arg_name, func, args, kwargs):
...
             return None
+
+
+def get_tempfile(file_storage):
+    """
+    Save stream file in temp directory
+    :param file_storage: stream file
+    :return: fileneme
+    """
+    tmp_file = os.path.join(tempfile.gettempdir(), file_storage.filename)
```
antoine-de: maybe `save_in_tmp` would be a better name ?

source/tyr/tyr/helper.py
```diff
@@ -191,3 +192,24 @@ def get_named_arg(arg_name, func, args, kwargs):
...
+    file_storage.save(tmp_file)
+    return tmp_file
+
+
+def rm_tempfile(filename):
```
antoine-de: I don't see why we would need this

source/tyr/tyr/resources.py
```diff
@@ -269,7 +87,7 @@ def post(self, instance_name):
         instance = models.Instance.query_existing().filter_by(name=instance_name).first_or_404()
 
         if not request.files:
-            return {'message': 'the Data file is missing'}, 400
+            return {'message': 'the Data file is missing'}, 404
```
antoine-de: I'd say 400 is better, it's a bad request, not a not found

source/tyr/tyr/decorators.py
```diff
@@ -0,0 +1,62 @@
...
+from navitiacommon import models
+from tyr.helper import get_instance_logger, rm_tempfile, get_tempfile
+from tyr.fields import one_job_fields, error_fields
+
+
+class update_data(object):
+    def __init__(self, is_autocomplete=False):
+        self.is_autocomplete = is_autocomplete
+
+    def __call__(self, func):
```
antoine-de: Hum I'm not convinced by a wrapper for this, I would prefer a function but as I don't know how to make it cleanly generic over either the navitia's instances and the autocomplete's instances I would put all this in `AutocompleteUpdateData::Post`

source/tyr/tyr/decorators.py
```diff
@@ -0,0 +1,62 @@
...
+                    filter_by(name=kwargs.get('instance_name')).first_or_404()
+
+            if not request.files:
+                return marshal({'error': {'message': 'the Data file is missing'}}, error_fields), 404
+            content = request.files['file']
+            logger = get_instance_logger(instance)
+            logger.info('content received: %s', content.filename)
+            filename = get_tempfile(content)
+            feature, job = func(*args, files=[filename], instance=instance, **kwargs)
+            rm_tempfile(filename)
```
antoine-de: I don't think this is needed, this should be handled by the import function

source/tyr/tyr/decorators.py
```diff
@@ -0,0 +1,62 @@
...
+                instance = models.AutocompleteParameter.query.\
+                    filter_by(name=kwargs.get('ac_instance_name')).first_or_404()
+            else:
+                instance = models.Instance.query_existing().\
+                    filter_by(name=kwargs.get('instance_name')).first_or_404()
+
+            if not request.files:
+                return marshal({'error': {'message': 'the Data file is missing'}}, error_fields), 404
+            content = request.files['file']
+            logger = get_instance_logger(instance)
```
antoine-de: you can check https://github.com/CanalTP/navitia/pull/2068, I changed the instance_logger

source/tyr/tests/integration/datapost_test.py
azime: OK

source/tyr/tyr/helper.py
azime: OK

source/tyr/tyr/helper.py
azime: Remove tempfile, no ?

source/tyr/tyr/resources.py
azime: correction of [this](https://github.com/CanalTP/navitia/blob/dev/source/tyr/tests/integration/datapost_test.py#L38) test

source/tyr/tyr/decorators.py
azime: :+1:

source/tyr/tyr/resources.py
antoine-de: the test does not seems correct, but we can ask the other about it

source/tyr/tyr/helper.py
antoine-de: no it should be handled by  the different task no ? (isn't there one moving the file to the backup ?)

source/tyr/tests/integration/datapost_test.py
```diff
@@ -32,12 +41,56 @@ def test_post_pbf(create_instance_fr):
...
+            json_response = api_post('/v0/autocomplete_parameters/fr/update_data', data=files)
+            assert 'job' in json_response
+            job = json_response['job']
+            assert 'id' in job
+            assert type(job['id']) == int
+
+            jobs = get_jobs_from_db()
+            assert len(jobs) == 1
+            assert jobs[0].id == job['id']
+            shutil.rmtree(ac_path)#Remove autocomplete directory
```
antoine-de: I don't like this test, if it fail, we'll need to hand remove the file.

I think pytest's tempdir can help us  no?

source/tyr/tests/integration/datapost_test.py
```diff
@@ -32,12 +41,56 @@ def test_post_pbf(create_instance_fr):
...
+        path = os.path.join(os.path.dirname(os.path.dirname(os.path.dirname(__file__))), 'tests/fixtures/', filename)
+        with open(path, 'rb') as f:
+            files = {'file': (f, filename)}
+            json_response = api_post('/v0/autocomplete_parameters/fr/update_data', data=files)
+            assert 'job' in json_response
+            job = json_response['job']
+            assert 'id' in job
+            assert type(job['id']) == int
+
+            jobs = get_jobs_from_db()
```
antoine-de: hum event if for the moment it's only for test purpose, I think this is quite dangerous because there can be a loooots of job (I don't think we clean them in the db).

I would have more implemented a `get` to fetch a job by it's id (and we'll need this method anyway)

source/tyr/tests/integration/datapost_test.py
```diff
@@ -32,12 +41,56 @@ def test_post_pbf(create_instance_fr):
...
+
+def test_post_pbf_autocomplete_without_files(create_instance_fr):
+    with app.app_context():
+        ac_path = current_app.config['TYR_AUTOCOMPLETE_DIR']
+        assert not os.path.exists(ac_path)
+        filename = 'empty_pbf.osm.pbf'
+        path = os.path.join(os.path.dirname(os.path.dirname(os.path.dirname(__file__))), 'tests/fixtures/', filename)
+        with open(path, 'rb') as f:
+            response = api_post('/v0/autocomplete_parameters/fr/update_data', check=False)
+            assert response[1] == 404
```
antoine-de: I think it's clearer to deconstruct the tuple
`response, status = api_post('/v0/autocomplete_parameters/fr/update_data', check=False)`

antoine-de: I think we should have a 400 here

source/tyr/tyr/decorators.py
```diff
@@ -0,0 +1,63 @@
...
+        def wrapper(*args, **kwargs):
+            if self.is_autocomplete:
+                instance = models.AutocompleteParameter.query.\
+                    filter_by(name=kwargs.get('ac_instance_name')).first_or_404()
+            else:
+                instance = models.Instance.query_existing().\
+                    filter_by(name=kwargs.get('instance_name')).first_or_404()
+
+            if not request.files:
+                return marshal({'error': {'message': 'the Data file is missing'}}, error_fields), 404
```
pbougue: Should be a 400 here also, non? (if we agree on the other 400)

source/tyr/tyr/decorators.py
```diff
@@ -0,0 +1,63 @@
...
+            content = request.files['file']
+            logger = get_instance_logger(instance)
+            logger.info('content received: %s', content.filename)
+            filename = save_in_tmp(content)
+            feature, job = func(*args, files=[filename], instance=instance, **kwargs)
+            job = models.db.session.merge(job) #reatache the object
+            rm_tempfile(filename)
```
pbougue: are you sure you are not deleting the file too soon? I guess it goes with @antoine-de remark on the method, have to check vs kraken instance tasks

pbougue (commented):
Just to be sure, TODO:
- [ ] hand-run tyr and docker tests
- [ ] hand-test that update is really done with a light pbf file (and maybe also bano if it's usefull to test)

pbougue:
Jenkins, test this please

PR #2070, opened 2017-04-26
jormungandr: Ignore jorm .ini config files

rchoquet:
This PR removes some dead code in the instance_manager and catches error in case of an invalid .json config file.
Not sure about the second part as i don't know if the abort with a raw exception in case of a malformed json config file was intended.

### DIFF
--- a/source/jormungandr/jormungandr/instance_manager.py
+++ b/source/jormungandr/jormungandr/instance_manager.py
@@ -81,13 +81,8 @@ class InstanceManager(object):
     """
 
     def __init__(self, instances_dir=None, start_ping=False):
-        # if a configuration file is defined in the settings we take it
-        # else we load all .ini/.json files found in the INSTANCES_DIR
-        if instances_dir:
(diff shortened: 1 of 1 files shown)


source/jormungandr/jormungandr/instance_manager.py
```diff
@@ -121,14 +116,13 @@ def initialisation(self):
...
-            if file_name.endswith('.json'):
+            try:
                 with open(file_name) as f:
                     config_data = json.load(f)
                     self.register_instance(config_data)
-
-            else:
-                logging.getLogger(__name__).warn('impossible to init an instance with the configuration '
-                                                 'file %s', file_name)
+            except ValueError:
```
antoine-de: hum I'm not sure about catching this, if the configuration file is not correct it might be better to fail fast

antoine-de (commented):
:+1: thanks!

pbougue (approved):

PR #2078, opened 2017-05-04
Jormungandr: Direct path filter

pbougue:
add a param to filter: `direct_path = [indifferent, none, only]`
and use it in `same_journey_schedule` link


antoine-de:
can you add an integration test ? I think it's important, especially with the `tag` based mechanism (it might change)

:white_check_mark: Done

PR #2082, opened 2017-05-09
Jormungandr: migrate vehicle journeys serialization to serpy

antoine-de:
also the migration of journey pattern and journey pattern point

Hopefully this will fix Artemis (since it's now in serpy)

### DIFF
--- a/source/jormungandr/jormungandr/interfaces/v1/decorators.py
+++ b/source/jormungandr/jormungandr/interfaces/v1/decorators.py
@@ -40,8 +40,11 @@ map_serializer = {
    "stop_areas": api.StopAreasSerializer,
    "routes": api.RoutesSerializer,
    "line_groups": api.LineGroupsSerializer,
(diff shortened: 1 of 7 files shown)


pbougue (approved):
looks good, but it would be nice that someone with "serpy-experience" check that...

source/jormungandr/jormungandr/interfaces/v1/decorators.py
```diff
@@ -40,8 +40,11 @@
    "stop_areas": api.StopAreasSerializer,
    "routes": api.RoutesSerializer,
    "line_groups": api.LineGroupsSerializer,
-   "lines": api.LinesSerializer,
```
azime, line 43: why delete this line?

antoine-de: it was defined above :stuck_out_tongue_winking_eye:

PR #2084, opened 2017-05-12
streetnetwork: tools for log

antoine-de:
add quantum gis stuff to debug direct path

also add a small log to display all kept connected components

### DIFF
--- a/source/ed/ed_reader.cpp
+++ b/source/ed/ed_reader.cpp
@@ -1379,17 +1379,26 @@ static std::vector<component_vertex_t> get_useless_nodes(const filtered_graph& g
 
     LOG4CPLUS_INFO(log, "the biggest has " << principal_component->second << " nodes");
 
+    auto big_enough = [&](const size_t size) {
+        return size / principal_component->second >= min_non_connected_graph_ratio;
+    };
+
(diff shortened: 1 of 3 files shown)


source/ed/ed_reader.cpp
```diff
@@ -1379,17 +1379,26 @@ static std::vector<component_vertex_t> get_useless_nodes(const filtered_graph& g
...
+        if (big_enough(nb_elt_in_component)) {
             continue; //big enough, we skip
         }
         useless_nodes.push_back(vertex_idx);
     }
     LOG4CPLUS_INFO(log, "for mode " << mode << ", " << useless_nodes.size() << " useless nodes");
+    for (const auto& comp_and_size: component_size) {
+        if (big_enough(comp_and_size.second)) {
+            LOG4CPLUS_INFO(log, "we kept a component with " << comp_and_size.second << " nodes");
+        }
```
pbougue, line 1400: maybe add a log on what comp we delete, too?

pbougue (approved):

source/ed/ed_reader.cpp
antoine-de: I dont' think so, on idf, there are about 100k components, we don't want to log them all 😜

PR #2085, opened 2017-05-12
Kraken: fix cornercase on dijkstra

antoine-de:
fix https://jira.canaltp.fr/browse/NAVITIAII-2217

the problem was that the starting edge was a 'round' one (from a vertex
A to the same vertex A)

with a small street network enhancement when the projection was exactly
on the vertex, it was leading dijkstra to think it was finished.


xlqian:
test?( if not complicated.... )

antoine-de:
I think they are complicated :rofl:

TeXitoi (approved):

PR #2087, opened 2017-05-16
PtRef: Display line section impact in lines

xlqian:
https://jira.canaltp.fr/browse/NAVITIAII-2211

### DIFF
--- a/source/type/pb_converter.cpp
+++ b/source/type/pb_converter.cpp
@@ -636,7 +636,7 @@ void PbCreator::Filler::fill_pb_object(const nt::Line* l, pbnavitia::Line* line)
          * the behavior too much.
          * */
         std::set<boost::shared_ptr<nt::disruption::Impact>> added_impact;
-        auto pred = [&](const nt::VehicleJourney& vj) {
+        auto fill_line_section_message = [&](const nt::VehicleJourney& vj) {
             for(const auto& impact: vj.meta_vj->impacted_by) {
                 auto impact_ptr = impact.lock();
                 if (! impact_ptr)
@@ -652,7 +652,7 @@ void PbCreator::Filler::fill_pb_object(const nt::Line* l, pbnavitia::Line* line)
             return true;
         };
         for(const auto* route: l->route_list) {
-            route->for_each_vehicle_journey(pred);
+            route->for_each_vehicle_journey(fill_line_section_message);
         }
     }
 }
--- a/source/type/pb_converter.cpp
+++ b/source/type/pb_converter.cpp
@@ -343,19 +343,19 @@ void PbCreator::Filler::add_contributor(const T* nav) {
 template<typename NAV, typename PB>
 void PbCreator::Filler::fill(NAV* nav_object, PB* pb_object) {
     if (nav_object == nullptr) { return; }
-    copy(depth-1, dump_message).fill_pb_object(nav_object, get_sub_object(nav_object, pb_object));
+    copy(depth-1, dump_message, false).fill_pb_object(nav_object, get_sub_object(nav_object, pb_object));
 }
 template<typename NAV, typename F>
 void PbCreator::Filler::fill_with_creator(NAV* nav_object, F creator) {
     if (nav_object == nullptr) { return; }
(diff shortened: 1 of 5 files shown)


source/type/pb_converter.h
```diff
@@ -322,6 +322,13 @@ struct PbCreator {
                 fill_pb_object(nav_obj, pb_list->Add());
             }
         }
+        template<typename Nav, typename Pb>
+        void fill_line_with_line_seciton_message(const std::vector<Nav*>& nav_list,
```
antoine-de: typo on seciton

source/type/pb_converter.cpp
```diff
@@ -624,6 +624,35 @@ void PbCreator::Filler::fill_pb_object(const nt::Line* l, pbnavitia::Line* line)
     }
 
 }
+void PbCreator::Filler::fill_line_with_line_section_message(const nt::Line* l , pbnavitia::Line* line){
```
antoine-de: wouldn't it be better to have a `Filler` flag (like DumpMessage) activated only with `PbCreator::pb_fill` (but for all objects) ?

TeXitoi (approved):

antoine-de (commented):
erf it's way more verbose then :disappointed: 

what do the others think about it ?

even if it's a bool, I prefer a bit an enum for sutff like this to 'name' the parameter (like DumpMessage)

source/type/pb_converter.cpp
```diff
@@ -623,37 +623,34 @@ void PbCreator::Filler::fill_pb_object(const nt::Line* l, pbnavitia::Line* line)
...
+                    continue;
+                for (const auto& entity: impact_ptr->informed_entities()) {
+                    if (boost::get<nt::disruption::LineSection>(&entity) &&
+                            added_impact.insert(impact_ptr).second ) {
+                        fill_messages(vj.meta_vj, line);
+                        return true;
+                    }
                 }
             }
+            return true;
```
pbougue: shouldn't we return `false` here?

source/type/pb_converter.cpp
```diff
@@ -623,37 +623,34 @@ void PbCreator::Filler::fill_pb_object(const nt::Line* l, pbnavitia::Line* line)
...
-                    fill_messages(vj.meta_vj, line);
-                    return true;
+    if (dump_message == DumpMessage::Yes && dump_line_section_message) {
+        /*
+         * Here we dump the impacts which impact LineSection.
+         * We could have link the LineSection impact with the line, but that would change the code and
+         * the behavior too much.
+         * */
+        std::set<boost::shared_ptr<nt::disruption::Impact>> added_impact;
+        auto pred = [&](const nt::VehicleJourney& vj) {
```
pbougue: `pred` is a bit too generic name (predicate, right?)

pbougue (commented):
As for the too verbose (and typing the boolean), a solution might be to wrap it in a struct with `dump_message` (as both are defaulted, it would probably do the job)

source/type/pb_converter.cpp
```diff
@@ -623,6 +623,32 @@ void PbCreator::Filler::fill_pb_object(const nt::Line* l, pbnavitia::Line* line)
...
+            for(const auto& impact: vj.meta_vj->impacted_by) {
+                auto impact_ptr = impact.lock();
+                if (! impact_ptr)
+                    continue;
+                for (const auto& entity: impact_ptr->informed_entities()) {
+                    if (boost::get<nt::disruption::LineSection>(&entity) &&
+                            added_impact.insert(impact_ptr).second ) {
+                        fill_messages(vj.meta_vj, line);
+                        return true;
```
pbougue, line 648: isn't it too early to return? if there are multiple line_sections?

xlqian: we can have multiple line_sections on the impact, but we add only once the impact :)

source/type/pb_converter.cpp
xlqian: `false` will stop the loop, which is not wanted...

antoine-de (approved):

PR #2090, opened 2017-05-17
All: remove fastlz useless lib

antoine-de:
again champignones! :tada:


pbougue (commented):
you can also update the type/readme.txt

pbougue (approved):

PR #2093, opened 2017-05-23
Kraken: upgrade of rabbitmq-c lib

kinnou02:
This version enable tcpkeepalive on the rabbitmq socket, so
disconnection will trigger an error on the kraken side, so it will be
able to reconnect.
No more "tu ne le sais pas encore, mais tu es deja mort" syndrome.

This PR replace #1289.
Let's wait for the release before merging it.

### DIFF
--- a/source/kraken/maintenance_worker.cpp
+++ b/source/kraken/maintenance_worker.cpp
@@ -237,8 +237,9 @@ void MaintenanceWorker::listen_rabbitmq(){
     }
     bool no_local = true;
     bool no_ack = false;
-    std::string task_tag = this->channel->BasicConsume(this->queue_name_task, "", no_local, no_ack);
-    std::string rt_tag = this->channel->BasicConsume(this->queue_name_rt, "", no_local, no_ack);
+    bool exclusive = false;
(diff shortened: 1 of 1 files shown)


source/kraken/maintenance_worker.cpp
```diff
@@ -237,8 +237,9 @@ void MaintenanceWorker::listen_rabbitmq(){
     }
     bool no_local = true;
     bool no_ack = false;
-    std::string task_tag = this->channel->BasicConsume(this->queue_name_task, "", no_local, no_ack);
-    std::string rt_tag = this->channel->BasicConsume(this->queue_name_rt, "", no_local, no_ack);
+    bool exclusive = false;
```
xlqian, line 240: hmmm, i'm just wondering why you need to change this...

kinnou02: After a timeout rabbitmq doesn't let create a new consumer without it.

xlqian: ok! understood

antoine-de (approved):

PR #2097, opened 2017-05-30
Jormungandr: add is_journeys_schedules for stif

xlqian:
https://jira.canaltp.fr/browse/NAVITIAII-2230

### DIFF
--- a/source/jormungandr/jormungandr/interfaces/v1/Journeys.py
+++ b/source/jormungandr/jormungandr/interfaces/v1/Journeys.py
@@ -321,7 +321,7 @@ class add_journey_href(object):
 
                     args['min_nb_transfers'] = journey['nb_transfers']
                     args['direct_path'] = 'only' if 'non_pt' in journey['tags'] else 'none'
-                    args["min_nb_journeys"] = 5
+                    args['min_nb_journeys'] = 5
                     args['is_journey_schedules'] = True
                     allowed_ids.update(args.get('allowed_id[]', []))
                     args['allowed_id[]'] = list(allowed_ids)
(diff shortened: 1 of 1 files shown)


source/jormungandr/jormungandr/interfaces/v1/Journeys.py
```diff
@@ -479,6 +481,11 @@ def _set_specific_params(mod):
         else:
             _set_specific_params(default_values)
 
+        # set parameters when is_journey_schedules is set to True
+        if args.get("is_journey_schedules"):
+            args["min_nb_journeys"] = 5
```
antoine-de: I don't think this should override the `min_nb_journeys`

And I'm not sure we should add this parameter under the hood, I would prefer this to be explicit in the `same_journey_schedule`

xlqian: @TeXitoi

source/jormungandr/jormungandr/interfaces/v1/Journeys.py
```diff
@@ -321,7 +321,8 @@ def wrapper(*args, **kwargs):
 
                     args['min_nb_transfers'] = journey['nb_transfers']
                     args['direct_path'] = 'only' if 'non_pt' in journey['tags'] else 'none'
-                    args['min_nb_journeys'] = 5
+                    args["min_nb_journeys"] = 5
```
pbougue: minor: you can let the simple quotes

source/jormungandr/jormungandr/interfaces/v1/Journeys.py
```diff
@@ -411,6 +412,8 @@ def __init__(self):
         parser_get.add_argument("count", type=default_count_arg_type)
         parser_get.add_argument("_min_journeys_calls", type=int)
         parser_get.add_argument("_final_line_filter", type=boolean)
+        parser_get.add_argument("is_journey_schedules", type=boolean, default=False,
+                                description="True when '/journeys' is called to compute the same journey schedules")
```
pbougue: maybe we can add that it (de)activates appropriate parameters for that use.

pbougue (approved):

antoine-de (approved):

PR #2099, opened 2017-05-30
Jormungandr: documentation of /places

antoine-de:
add the swagger documentation to /places

I updated flask_restful to the lastest version to be able to avoid method decorators on `OPTIONS`

the descriptions on all the output needs to be done

the swagger output needs to be better checked :wink:

### DIFF
--- a/source/jormungandr/jormungandr/interfaces/parsers.py
+++ b/source/jormungandr/jormungandr/interfaces/parsers.py
@@ -30,6 +30,8 @@ from dateutil import parser
 from navitiacommon import parser_args_type
 
 # TODO: to be moved completely into navitiacommon
+from navitiacommon.parser_args_type import ParameterDescription
+
 depth_argument = parser_args_type.depth_argument
 
 float_gt_0 = parser_args_type.float_gt_0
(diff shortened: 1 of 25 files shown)


source/jormungandr/jormungandr/interfaces/parsers.py
```diff
@@ -76,20 +78,24 @@ def unsigned_integer(value):
...
+        if not (-90.0 <= lat <= 90.0):
+            raise ValueError("lat should be between -90 and 90")
+        lon = float(lon)
+        if not (180.0 >= lon >= -180.0):
+            raise ValueError("lon should be between -180 and 180")
+
+        return coord
+
+    def description(self):
+        return ParameterDescription(type=str, metadata={'pattern': '.*;.*'})
```
pbougue, line 101: genuine question (don't know where the cursor is): don't we wanna precise more that it's float?

source/jormungandr/jormungandr/interfaces/v1/Places.py
```diff
@@ -61,38 +66,38 @@
...
+class geojson_argument(object):
+    def __call__(self, value):
+        decoded = json.loads(value)
+        if not decoded:
+            raise ValueError('invalid shape')
+
+        return parser_args_type.geojson_argument(decoded)
+
+    def description(self):
+        return ParameterDescription(type=str)  # TODO a better description of the geojson
```
pbougue, line 78: maybe just add a link in description to https://geojson.org/geojson-spec.html (or the RFC if readable enough) ?

source/jormungandr/jormungandr/interfaces/v1/Places.py
```diff
@@ -61,38 +66,38 @@
...
         self.parsers["get"].add_argument("count", type=default_count_arg_type, default=10,
                                          description="The maximum number of places returned")
         self.parsers["get"].add_argument("search_type", type=int, default=0,
                                          description="Type of search: firstletter or type error")
         self.parsers["get"].add_argument("admin_uri[]", type=unicode,
                                          action="append",
-                                         description="If filled, will\
-                                         restrained the search within the\
-                                         given admin uris")
+                                         description="If filled, will restrained the search within the "
```
pbougue: `restrain` (no "-ed")

pbougue (commented):
seems ok, but not sure I have a sharp eye on that

source/jormungandr/jormungandr/interfaces/parsers.py
antoine-de: I'll let you write the regex for a float :grin: 

more seriously I think we should not overengineer this sort of stuff and wait for the SDK input to fit their needs

source/jormungandr/jormungandr/interfaces/v1/Places.py
antoine-de: hum I don't think we can do this, I don't see how the SDK will be able to handle this parameter (but maybe we don't need it and we can do a simple thing as giving a link to the documentation as you said)

PR #2106, opened 2017-06-01
jormungandr: deactivate by default the schema on OPTIONS

antoine-de:
since an OPTIONS call is very common, we do not want to add a big
overhead on it.

so by default we do not dump a route's schema with the OPTIONS verb, we
need the param `schema=true`)

Note: since they touch the same files, this PR needs https://github.com/CanalTP/navitia/pull/2104 you can review only the last commit

### DIFF
--- a/source/jormungandr/jormungandr/interfaces/v1/Coverage.py
+++ b/source/jormungandr/jormungandr/interfaces/v1/Coverage.py
@@ -89,9 +89,4 @@ class Coverage(StatedResource):
         return resp, 200
 
     def options(self, **kwargs):
-        from flask import request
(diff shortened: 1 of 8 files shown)


source/jormungandr/jormungandr/interfaces/v1/Coverage.py
```diff
@@ -90,5 +90,8 @@ def get(self, region=None, lon=None, lat=None):
 
     def options(self, **kwargs):
         from flask import request
+        if not request.headers.get('DUMP_SCHEMA'):
```
kinnou02: You must cast it to a bool

kinnou02 (changes requested):

PR #2107, opened 2017-06-02
jormungandr: refacto of error handling in realtime proxy

kinnou02:
Internally exception are used as a way of signaling an error, this allow
us to generate some statistics for each connectors that will help to
improve our monitoring

### DIFF
--- a/source/jormungandr/jormungandr/realtime_schedule/siri_lite.py
+++ b/source/jormungandr/jormungandr/realtime_schedule/siri_lite.py
@@ -28,7 +28,7 @@
 # IRC #navitia on freenode
 # https://groups.google.com/d/forum/navitia
 # www.navitia.io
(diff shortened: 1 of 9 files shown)


source/jormungandr/jormungandr/realtime_schedule/siri_lite.py
```diff
@@ -136,15 +135,12 @@ def _get_next_passage_for_route_point(self, route_point, count=None, from_dt=Non
         if not url:
             return None
         r = self._call(url)
-        if not r:
```
antoine-de, line 139: won't that crash here ?

kinnou02: We do not return `None` anymore

PR #2111, opened 2017-06-07
Doc: ntfs_fare_extention - multiple tickets on od

prhod:
Enabling multiple tickets on od_fares.csv


pbougue (commented):
Closing this as we were able to do without it, and we don't want people to use it like that even if it's working

PR #2112, opened 2017-06-07
Doc: better document datetime format in requests

TeXitoi:


documentation/slate/source/includes/objects.md
```diff
@@ -15,12 +15,18 @@ Lots of object are geographically localized :
 
 ### Iso-date-time
 
-Navitia manages every date or time as a UTC date-time. The web-service 
+Navitia
 
 -   exposes every date times as local times via an ISO 8601 "YYYYMMDDThhmmss" string
--   can be request using local times via an ISO 8601 "YYYYMMDDThhmmss" string
+-   can be request using local times via ISO 8601 as "YYYYMMDDThhmmss" or "YYYY-MM-DDThh:mm:ss"
```
antoine-de: can be request*ed* no ?

and maybe add a comment saying we consider it in the local time of the coverage ?

PR #2118, opened 2017-06-14
All: fix build on debian 9

### LINKED ISSUE #1858: make test failing for boost 1.62
`make test` is using ctest, which fails when using boost 1.62 (debian 9 version), probably because of this issue : https://svn.boost.org/trac/boost/ticket/12507

Launching tests manually works, though.
No specific dev intended on that for now, just reporting and filing it somewhere so that no-one searches on it a second time.

Workaround: use navitia-tested boost versions, like 1.55 and 1.61

kinnou02:
I had to disable unittest reports as they don't work with boost 1.62
fix #1858


pbougue (approved):

kinnou02:
jenkins, test this please

PR #2121, opened 2017-06-19
Jormungandr: Use sources to targets in scenario distributed

kadhikari:
Instead of api valhalla one_to_many, we use sources_to_targets to calcaulate fallback_durations for origin and destination. The use of this api calculates the durations from fallback points toward the destination (durations calculated from destination towards fallback points using one_to_many previously).

Ticket concerned: https://jira.canaltp.fr/browse/NAVP-651

### DIFF
--- a/source/jormungandr/jormungandr/scenarios/helper_classes/fallback_durations.py
+++ b/source/jormungandr/jormungandr/scenarios/helper_classes/fallback_durations.py
@@ -51,7 +51,7 @@ class FallbackDurations:
     The returned dict will look like {'stop_point:stopA': 360, 'stop_point:stopB': 180, 'stop_point:stopC': 60}
     """
     def __init__(self, future_manager, instance, requested_place_obj, mode, proximities_by_crowfly_pool, places_free_access,
-                 max_duration_to_pt, request, speed_switcher):
+                 max_duration_to_pt, request, speed_switcher, origin=True):
         """
         :param future_manager: a module that manages the future pool properly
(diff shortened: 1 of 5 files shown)


source/jormungandr/jormungandr/scenarios/helper_classes/fallback_durations.py
```diff
@@ -117,12 +117,22 @@ def _do_request(self):
...
             else:
                 return result
-        sn_routing_matrix = self._instance.get_street_network_routing_matrix([center_isochrone],
-                                                                             places_isochrone,
-                                                                             self._mode,
-                                                                             self._max_duration_to_pt,
-                                                                             self._request,
-                                                                             **self._speed_switcher)
+
+        if self._origin:
```
antoine-de: I think we can get this _origin another way, I don't remember how, come and we'll try to search for it

source/jormungandr/jormungandr/street_network/valhalla.py
```diff
@@ -264,9 +265,9 @@ def direct_path(self, mode, pt_object_origin, pt_object_destination, fallback_ex
     @classmethod
     def _get_matrix(cls, json_response, mode_park_cost):
         sn_routing_matrix = response_pb2.StreetNetworkRoutingMatrix()
-        for one_to_many in json_response['one_to_many']:
+        for souce_to_target in json_response['sources_to_targets']:
```
antoine-de: typo on sou**r**ce

antoine-de (approved):

PR #2128, opened 2017-06-21
Jormungandr: Ability to call HERE routing service to get realtime car routing data

antoine-de:
Add another streetnetwork connector calling [HERE](https://developer.here.com) service

the nice feature of this service is that there is realtime data on the streetnetwork.

One of the down side is that the routing matrix is limited to 100 elements, so it is usable only in a very sparse dataset (like the SNCF).

This is a first rough implementation.

some stuff needs to be implemented (but I'm not sure they are mandatory):
* search range (but the here api limit the range as a distance and we do it as a travel time)
* better directions (for the moment there is not the name of the streets and the direction to take)

what do cannot do:
* change the vehicle speed (only the walking speed can be set)
* the end fallback will have a false datetime as we cannot be precise (it depends on the public transport taken)

### DIFF
--- a/source/jormungandr/jormungandr/street_network/here.py
+++ b/source/jormungandr/jormungandr/street_network/here.py
@@ -1,0 +1,264 @@
+#  Copyright (c) 2001-2016, Canal TP and/or its affiliates. All rights reserved.
+#
+# This file is part of Navitia,
+#     the software to build cool stuff with public transport.
+#
+# Hope you'll enjoy and contribute to this project,
+#     powered by Canal TP (www.canaltp.fr).
+# Help us simplify mobility and open public transport:
+#     a non ending quest to the responsive locomotion way of traveling!
+#
+# LICENCE: This program is free software; you can redistribute it and/or modify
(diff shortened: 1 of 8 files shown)


source/jormungandr/jormungandr/street_network/here.py
```diff
@@ -0,0 +1,291 @@
...
+        self.breaker = pybreaker.CircuitBreaker(fail_max=app.config['CIRCUIT_BREAKER_MAX_HERE_FAIL'],
+                                                reset_timeout=app.config['CIRCUIT_BREAKER_HERE_TIMEOUT_S'])
+
+    def _call_here(self, url, params):
+        log = logging.getLogger(__name__)
+        log.debug('Here routing service, url: {}'.format(url))
+        try:
+            return self.breaker.call(requests.get, url, timeout=self.timeout, params=params)
+        except pybreaker.CircuitBreakerError as e:
+            log.error('Valhalla routing service dead (error: {})'.format(e))
```
kinnou02: please add an extra parameters with the id of the system

kinnou02 (approved):

PR #2132, opened 2017-07-03
jormungandr: choose streetnetwork at request

antoine-de:
add a hidden param `_street_network` to override the choosen streetnetwork

the id of the streetnetwork can be configured in the instance's configuration (and is defined by default : `kraken`, `valhalla`, `here` and `geovelo`)

### DIFF
--- a/source/jormungandr/jormungandr/instance.py
+++ b/source/jormungandr/jormungandr/instance.py
@@ -80,7 +80,8 @@ def _set_default_street_network_config(street_network_configs):
         street_network_configs = []
     default_sn_class = 'jormungandr.street_network.kraken.Kraken'
 
-    modes_in_configs = set(list(itertools.chain.from_iterable(config['modes'] for config in street_network_configs)))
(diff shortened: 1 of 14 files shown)


source/jormungandr/jormungandr/instance.py
```diff
@@ -489,8 +489,22 @@ def init(self):
...
+        overriden_sn_id = request.get('_street_network')
+        if overriden_sn_id:
+            def predicate(s):
+                return s.sn_system_id == overriden_sn_id
+        else:
+            def predicate(s):
+                return mode in s.modes
+        sn = next((s for s in self.street_network_services if predicate(s)), None)
+        if sn is None:
+            logging.getLogger(__name__).info('impossible to find streetnetwork module for {} ({})'.format(
```
kinnou02: what happen if we return `None` here?

kinnou02 (approved):

source/jormungandr/jormungandr/instance.py
antoine-de: it crashes and I forgot to push force the correction :wink: 

now a TechnicalError is raised

PR #2135, opened 2017-07-05
jormun: add a streetnetwok connector for asgard

kinnou02:
Asgard is based on libvalhalla but use zmq and protobuf like kraken.
it also includes a projection cache to speed up matrices

At this time asgard only handle matrices, so we still use valhalla to
compute direct path.

### DIFF
--- a/source/jormungandr/jormungandr/street_network/asgard.py
+++ b/source/jormungandr/jormungandr/street_network/asgard.py
@@ -1,0 +1,107 @@
+#  Copyright (c) 2001-2016, Canal TP and/or its affiliates. All rights reserved.
+#
+# This file is part of Navitia,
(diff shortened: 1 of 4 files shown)


source/jormungandr/jormungandr/street_network/asgard.py
```diff
@@ -0,0 +1,109 @@
...
+from jormungandr.street_network.valhalla import Valhalla
+
+from contextlib import contextmanager
+import queue
+from navitiacommon import response_pb2, request_pb2, type_pb2
+from zmq import green as zmq
+import six
+
+
+logging.debug('test')
```
antoine-de: debug I think

antoine-de (approved):

PR #2137, opened 2017-07-05
Kraken: better autocomplete

antoine-de:
there was a problem, the 'ligne b' was impossible to find because it all lines (with 'ligne' in their indexed label), and various starting with 'b' (like 'bus' :wink:)

So the autocomplete algorithm has been a bit reworked and now, after the scoring of all the prefix, we score all the candidates with:
* the absolute score of the object (like the size of the admin, of the stoppoints, ...)
* by trying to find the longest common substring between the string search and the indexed string
* the position of this substring in the indexed string

so when we search 'rer b':

if we get 3 candidates:

1 -> 'Busval d'Oise Bus 95-01 (Zone Aéroportuaire Aéroport Charles de 1 RER B - Luzarches Gare)'
2 -> 'Rosny Bois Perrier RER (Rosny-sous-Bois)'
3 -> 'rer b'

for 1 the longest common substring is 'rer b' and its position is 74
for 2 the longest common substring is 'rer' and its position is 22
for 3 the longest common substring is 'rer b' and its position is 5 (it's the end of the substring)

thus we sort in the order (3, 1, 2)

as a side effect we can remove the uggly hack to be able to find "rue de loire saint sebastien sur loire" :tada: 

Note: the lz4 for pdl grow from 293M to 302M (since we now store the full indexed label for each object) and the kraken RAM from 1450M to 1472M)

### DIFF
--- a/source/autocomplete/autocomplete.cpp
+++ b/source/autocomplete/autocomplete.cpp
@@ -175,7 +175,7 @@ Autocomplete<T>::compute_vec_quality(const std::string& str,
             quality.idx = i;
             quality.nb_found = word_quality_list.at(quality.idx).word_count;
             quality.word_len = wordLength;
-            quality.score = word_quality_list.at(quality.idx).score;
+            quality.scores = this->compute_result_scores(str, quality.idx);
             quality.quality = 100;
 
             // whole_way_name = road name + "(" +  city name + ")"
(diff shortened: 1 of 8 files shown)


source/autocomplete/autocomplete.cpp
```diff
@@ -155,65 +155,34 @@ void Autocomplete<T>::compute_score(type::PT_Data &pt_data, georef::GeoRef &geor
...
-
-            // whole_way_name = road name + "(" +  city name + ")"
-            auto whole_way_name = geo_ref.ways[i]->name;
-            for (const auto* admin: geo_ref.ways[i]->admin_list) {
-                if (admin && admin->level == 8) {
-                    whole_way_name += (" " + admin->name);
-                }
+    auto curr = std::vector<size_t>(str2.size());
+    auto prev = std::vector<size_t>(str2.size());
+    size_t maxSubstr = 0;
```
TeXitoi: That's not java!

TeXitoi (approved):

PR #2142, opened 2017-07-10
Doc: update transfer_type possible values

antoine-de:
we don't have neither `garanteed` nor `extension` transfer_type (I don't know where those value came from :wink: )

the transfer_type can either be `walking` or `stay_in` (also called vj extension, if you stay in the bus, but the bus change it's headsign)

@stifoon is that ok with you ?

Note: delete the branch afterward, I did not find in the github web interface how to push it on my fork


stifoon:
Probably historical values :)

PR #2144, opened 2017-07-11
Kraken: do not deepcopy of georef when clonning data

kinnou02:
It isn't really a big win, on fr-idf, at most we are down to 18sec from
21sec before


kinnou02:
this PR lack the reassociation of the stop_area define in `Admin`, so a post treatment would be required.

We don't gain enough to follow this path.

PR #2145, opened 2017-07-11
Jormungandr: add a feed publisher for bss

antoine-de:
now each time a bss provider is called, we add it's license in the
response (as a feed publisher)

by default the jcdecaux license has been defined as

```json
{
    "id": "jcdecaux",
    "name": "jcdecaux",
    "license": "Licence Ouverte / Open License",
    "url": "https://developer.jcdecaux.com/#/opendata/license"
}
```

and the atos has no feed publisher as I don't know what to put for it

### DIFF
--- a/source/jormungandr/jormungandr/parking_space_availability/bss/atos.py
+++ b/source/jormungandr/jormungandr/parking_space_availability/bss/atos.py
@@ -37,10 +37,7 @@ import pybreaker
 
 from jormungandr.ptref import FeedPublisher
 
-DEFAULT_ATOS_FEED_PUBLISHER = {
-    'id': 'atos',
-    'name': 'atos'
-}
+DEFAULT_ATOS_FEED_PUBLISHER = None
 
 
 class AtosProvider(BssProvider):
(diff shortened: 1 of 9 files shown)


kinnou02 (approved):

source/jormungandr/jormungandr/parking_space_availability/bss/atos.py
```diff
@@ -35,17 +35,26 @@
 import logging
 import pybreaker
 
+from jormungandr.ptref import FeedPublisher
+
+DEFAULT_ATOS_FEED_PUBLISHER = {
+    'id': 'atos',
+    'name': 'atos'
```
kinnou02: it should probably be something like "effia something" or cycleo

source/jormungandr/jormungandr/parking_space_availability/bss/bss_provider_manager.py
```diff
@@ -41,29 +41,36 @@ def __init__(self, bss_providers_configuration):
...
         for poi in get_from_to_pois_of_journeys(journeys):
-            self.handle_poi(poi)
-        return journeys
+            provider = self._handle_poi(poi)
+            if provider:
+                providers.add(provider)
+        return providers
 
-    def handle_poi(self, item):
+    def _handle_poi(self, item):
```
pbougue, line 62: maybe just comment that it returns provider if used

pbougue (approved):

PR #2149, opened 2017-07-19
Fare: Fix a crash

TeXitoi:
fix https://jira.canaltp.fr/browse/NAVITIAII-2258

### DIFF
--- a/source/fare/tests/fare_test.cpp
+++ b/source/fare/tests/fare_test.cpp
@@ -425,7 +425,7 @@ BOOST_FIXTURE_TEST_CASE(tram_with_od, fare_load_fixture) {
 BOOST_FIXTURE_TEST_CASE(exclusive_line, fare_load_fixture) {
     // On teste les lignes à tarif exclusif
     keys.clear();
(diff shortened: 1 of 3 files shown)


source/fare/tests/fare_test.cpp
```diff
@@ -425,7 +425,7 @@ BOOST_FIXTURE_TEST_CASE(tram_with_od, fare_load_fixture) {
 BOOST_FIXTURE_TEST_CASE(exclusive_line, fare_load_fixture) {
     // On teste les lignes à tarif exclusif
     keys.clear();
-    keys.push_back(";paris;098098001:1;areoport;2011|07|31;09|28;09|39;4;4;Bus");
```
antoine-de, line 428: can you add a comment explaining why it's bobette ?

antoine-de (approved):

source/fare/tests/fare_test.cpp
TeXitoi: Done

antoine-de (approved):

pbougue (approved):

PR #2150, opened 2017-07-19
Streetnetwork: remove allocs around dijkstra

pbougue:


TeXitoi (approved):

kinnou02:
Allocation during street network are fully removed,
On test request (a little bit extreme) we go down to 470ms from 1019ms.
Almost no time is spent in kernel space and call to clear_page have been eliminated

source/georef/dijkstra_shortest_paths_with_heap.h
```diff
@@ -0,0 +1,61 @@
...
+
+Stay tuned using
+twitter @navitia
+IRC #navitia on freenode
+https://groups.google.com/d/forum/navitia
+www.navitia.io
+*/
+
+#include <boost/graph/dijkstra_shortest_paths.hpp>
+
```
kinnou02, line 32: a little comment about what we did that will be nice I think
> :white_check_mark: Done

source/georef/street_network.h
```diff
@@ -168,20 +174,25 @@ struct PathFinder {
     template<class Visitor>
     void dijkstra(vertex_t start, Visitor visitor) {
         // Note: the predecessors have been updated in init
-        boost::two_bit_color_map<> color(boost::num_vertices(geo_ref.graph));
+
+        // Fill color map in white before dijkstra
+        std::fill(color.data.get(),
```
kinnou02, line 179: why don't we do this in the init?

kinnou02 (approved):

source/georef/street_network.h
pbougue: We do 2 dijkstra calls (one on each node of the edge receiving projection).

PR #2151, opened 2017-07-19
Jormungandr&Kraken: realtime proxy for stif

antoine-de:
Correction of the Sirilite connector used to get realtime schedules on fr-idf.

In the siri's realtime response we only got a destination to match a route, so we added a guessing function that try to find all the routes of a line that pass first by a stop and after by another stop (even if the 2nd stop is not the destination)

## Test

to test it, the sirilite codes in the data are needed.
if you have some you can test it on the gare de lyon

`/stop_areas/stop_area%3AOIF%3ASA%3A8768600/stop_schedules`
there should be realtime data on the line 1 and line 14

## new relic monitoring
To monitor a bit the service's response a new_relic event was added:
`realtime_proxy_additional_info`
with for moment the values:
 * `no_departure`
 * `no_destination`
 * `no_route_found`
 * `no_departure_time`

## deploy
Note: we need to remove `&MonitoringRef=` from the url in the deployment conf

### DIFF
--- a/source/jormungandr/jormungandr/realtime_schedule/siri_lite.py
+++ b/source/jormungandr/jormungandr/realtime_schedule/siri_lite.py
@@ -109,6 +109,7 @@ class SiriLite(RealtimeProxy):
                      for vj in d.get('MonitoredStopVisit', [])
                      if vj.get('MonitoredVehicleJourney', {}).get('LineRef', {}).get('value') == line_code]
         if not schedules:
+            self.record_additional_info('no_departure')
             return None
         next_passages = []
         for next_expected_st in schedules:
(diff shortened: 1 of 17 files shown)


antoine-de:
Jenkins, test this please

source/jormungandr/jormungandr/realtime_schedule/siri_lite.py
```diff
@@ -82,54 +82,77 @@ def _make_url(self, route_point):
...
-            return next_passages
-        else:
+        stop_deliveries = resp.get('Siri', {}).get('ServiceDelivery', {}).get('StopMonitoringDelivery', [])
+        schedules = [vj
+                     for d in stop_deliveries
+                     for vj in d.get('MonitoredStopVisit', [])
+                     if vj.get('MonitoredVehicleJourney', {}).get('LineRef', {}).get('value') == line_code]
+        if not schedules:
+            self.record_additional_info('no_departure')
             return None
```
kinnou02: Should it be `[]`?

TeXitoi (approved):

PR #2159, opened 2017-07-27
Kraken: clean useless vehicle journeys

antoine-de:
When receiving multiple realtime data, some realtime vehicle journeys might become useless (their validity pattern are always empty, so cannot be used)

Before, we used to keep them because of the fear of memory corruption, but thus the memory consumption was ever growing (and we are unfortunately not in an infinite memory world :cry: ) and the growing number of vehicle journeys was leading to timeouts on some APIs (`/traffic_reports` for example).

With this PR we now clean the useless vehicle journeys.
As kraken is written in C++ ( :heart: [rust](https://github.com/rust-lang/rust) ) it is quite dangerous and we'll need to be extra careful in the future not to add a backref to a vehicle journey without cleaning it :boom: 

Note: I still need to do some more tests again the production realtime flow 
Edit: tests done, seems OK

### DIFF
--- a/source/type/type.cpp
+++ b/source/type/type.cpp
@@ -44,6 +44,7 @@ www.navitia.io
 #include "third_party/eos_portable_archive/portable_oarchive.hpp"
 #include <boost/archive/binary_oarchive.hpp>
 #include <boost/archive/binary_iarchive.hpp>
+#include <boost/range/algorithm/find.hpp>
 
 namespace bt = boost::posix_time;
 
@@ -405,8 +406,8 @@ static bool is_useless(const nt::VehicleJourney& vj) {
 }
 
 template <typename C>
-static C::iterator erase_vj_from_list(const nt::VehicleJourney* vj, C& vjs) {
-    auto it = std::find(vjs.begin(), vjs.end(), vj);
+static typename C::iterator erase_vj_from_list(const nt::VehicleJourney* vj, C& vjs) {
+    auto it = boost::range::find(vjs, vj);
     if (it == vjs.end()) {
         LOG4CPLUS_WARN(log4cplus::Logger::getInstance("logger"),
                        "impossible to find the vj " << vj->uri << " in the list, something is strange");
--- a/source/type/type.cpp
+++ b/source/type/type.cpp
@@ -395,15 +395,6 @@ static bool concerns_base_at_period(const VehicleJourney& vj,
     return intersect;
 }
 
-namespace {
-template<typename VJ> std::vector<VJ*>& get_vjs(Route* r);
-template<> std::vector<DiscreteVehicleJourney*>& get_vjs(Route* r) {
-    return r->discrete_vehicle_journey_list;
-}
-template<> std::vector<FrequencyVehicleJourney*>& get_vjs(Route* r) {
-    return r->frequency_vehicle_journey_list;
-}
-}// anonymous namespace
 
 static bool is_useless(const nt::VehicleJourney& vj) {
     // a vj is useless if all it's validity pattern are empty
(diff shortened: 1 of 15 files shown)


antoine-de:
Jenkins, test this please

source/type/type.cpp
```diff
@@ -403,8 +423,64 @@ template<> std::vector<DiscreteVehicleJourney*>& get_vjs(Route* r) {
...
+    }
+    if (vj->physical_mode) {
+        erase_vj_from_list(vj, vj->physical_mode->vehicle_journey_list);
+    }
+    if (dynamic_cast<const nt::FrequencyVehicleJourney*>(vj)) {
+        erase_vj_from_list(vj, vj->route->frequency_vehicle_journey_list);
+    } else if (dynamic_cast<const nt::DiscreteVehicleJourney*>(vj)) {
+        erase_vj_from_list(vj, vj->route->discrete_vehicle_journey_list);
+    } else {
+        throw std::logic_error("c'est la merde"); // TODO remove this
```
kinnou02: This can probably be removed

source/type/type.cpp
```diff
@@ -418,7 +494,6 @@ VJ* MetaVehicleJourney::impl_create_vj(const std::string& uri,
     VJ* ret = vj_ptr.get();
     vj_ptr->meta_vj = this;
     vj_ptr->uri = uri;
-    vj_ptr->idx = pt_data.vehicle_journeys.size();
```
kinnou02, line 421: don't will we end up with vj without idx?

kinnou02 (commented):
Have you some insight about performance?

source/type/type.cpp
antoine-de: it's done below

antoine-de:
no significant impact on disruption's application can be seen (the data copying is way too long to see anything else :stuck_out_tongue_winking_eye:  )

source/type/type.cpp
```diff
@@ -395,6 +396,25 @@ static bool concerns_base_at_period(const VehicleJourney& vj,
...
+}
+
+template <typename C>
+static typename C::iterator erase_vj_from_list(const nt::VehicleJourney* vj, C& vjs) {
+    auto it = boost::range::find(vjs, vj);
+    if (it == vjs.end()) {
+        LOG4CPLUS_WARN(log4cplus::Logger::getInstance("logger"),
+                       "impossible to find the vj " << vj->uri << " in the list, something is strange");
+    }
+    return vjs.erase(it);
```
TeXitoi: should be in an else

TeXitoi (approved):
👍 approved after correction of the remark.

source/type/type.cpp
```diff
@@ -395,6 +396,26 @@ static bool concerns_base_at_period(const VehicleJourney& vj,
...
+
+template <typename C>
+static typename C::iterator erase_vj_from_list(const nt::VehicleJourney* vj, C& vjs) {
+    auto it = boost::range::find(vjs, vj);
+    if (it == vjs.end()) {
+        LOG4CPLUS_WARN(log4cplus::Logger::getInstance("logger"),
+                       "impossible to find the vj " << vj->uri << " in the list, something is strange");
+        return it;
+    }
+    return vjs.erase(it);
```
xlqian: hmm... dangerous....

..., erasing elements in positions other than the vector end causes the container to **RELOCATE** all the elements after the segment erased to their new positions. 


```
v.erase( std::remove( vjs.begin(), vjs.end(), vj ), vjs.end() ); 
```

source/type/type.cpp
```diff
@@ -403,8 +424,62 @@ template<> std::vector<DiscreteVehicleJourney*>& get_vjs(Route* r) {
...
+    for (const auto& level_and_vj_idx_to_remove: vj_idx_to_remove) {
+        auto rt_level = level_and_vj_idx_to_remove.first;
+        auto vj_idx = level_and_vj_idx_to_remove.second;
+        auto& vj = this->rtlevel_to_vjs_map[rt_level][vj_idx];
+
+        cleanup_useless_vj_link(vj.get(), pt_data);
+
+        // once all the links to the vj have been cleaned we can destroy the object
+        // (by removing the unique_ptr from the vector)
+        this->rtlevel_to_vjs_map[rt_level].erase(this->rtlevel_to_vjs_map[rt_level].begin() + vj_idx);
```
xlqian: use erase-remove idiom

PR #2153, opened 2017-07-25
Doc: Swagger journey

pbougue:
TODO: fields description (to be done after)

PR #2161, opened 2017-07-28
Doc: flask args described with "help" param

pbougue:
To be read and merged only after merge of #2153


kinnou02 (approved):

antoine-de (approved):
nice! 

some tests would have been nice though :stuck_out_tongue_winking_eye:

PR #2165, opened 2017-07-31
kraken: Fix error unknown origin point

kadhikari:
- A request with an object other than EntryPoint with or without a code returns an error message "Error: several errors occured: * .".

Concerned ticket: https://jira.canaltp.fr/browse/NAVITIAII-2263

### DIFF
--- a/source/jormungandr/tests/isochrone_tests.py
+++ b/source/jormungandr/tests/isochrone_tests.py
@@ -90,18 +90,18 @@ class TestIsochrone(AbstractTestFixture):
     #In the previous version, the result contained no_solution with error message : 'several errors occured: \n * '
     #with status code 200.
     #After this correction, since the EntryPoint is absent an error is raised.
-    def test_from_isochrone_sa(self):
+    def test_isochrone_from_unkown_sa(self):
         query = "v1/coverage/basic_routing_test/journeys?from={}&datetime={}"
-        query = query.format("stopA", "20120614T080000")
+        query = query.format("bob", "20120614T080000")
         normal_response, error_code = self.query_no_assert(query)
-        assert normal_response['error']['message'] == 'The entry point: stopA is not valid'
+        assert normal_response['error']['message'] == 'The entry point: bob is not valid'
         assert error_code == 404
 
-    def test_to_isochrone_sa(self):
+    def test_isochrone_to_unknown_sa(self):
         query = "v1/coverage/basic_routing_test/journeys?to={}&datetime={}&datetime_represents=arrival"
-        query = query.format("stopA", "20120614T080000")
+        query = query.format("bob", "20120614T080000")
         normal_response, error_code = self.query_no_assert(query)
-        assert normal_response['error']['message'] == 'The entry point: stopA is not valid'
+        assert normal_response['error']['message'] == 'The entry point: bob is not valid'
         assert error_code == 404
 
     def test_isochrone_from_point(self):
--- a/source/jormungandr/tests/isochrone_tests.py
+++ b/source/jormungandr/tests/isochrone_tests.py
@@ -90,12 +90,16 @@ class TestIsochrone(AbstractTestFixture):
     def test_from_isochrone_sa(self):
         query = "v1/coverage/basic_routing_test/journeys?from={}&datetime={}"
         query = query.format("stopA", "20120614T080000")
-        response = self.query(query)
+        normal_response, error_code = self.query_no_assert(query)
+        assert normal_response['error']['message'] == 'The entry point: stopA is not valid'
+        assert error_code == 404
 
     def test_to_isochrone_sa(self):
         query = "v1/coverage/basic_routing_test/journeys?to={}&datetime={}&datetime_represents=arrival"
         query = query.format("stopA", "20120614T080000")
-        response = self.query(query)
+        normal_response, error_code = self.query_no_assert(query)
+        assert normal_response['error']['message'] == 'The entry point: stopA is not valid'
+        assert error_code == 404
 
     def test_reverse_isochrone_coord(self):
         q = "v1/coverage/basic_routing_test/journeys?max_duration=100000" \
--- a/source/jormungandr/tests/isochrone_tests.py
+++ b/source/jormungandr/tests/isochrone_tests.py
@@ -87,6 +87,9 @@ class TestIsochrone(AbstractTestFixture):
         query = query.format(s_coord, "20120614T080000")
         response = self.query(query)
 
+    #In the previous version, the result contained no_solution with error message : 'several errors occured: \n * '
+    #with status code 200.
+    #After this correction, since the EntryPoint is absent an error is raised.
     def test_from_isochrone_sa(self):
         query = "v1/coverage/basic_routing_test/journeys?from={}&datetime={}"
         query = query.format("stopA", "20120614T080000")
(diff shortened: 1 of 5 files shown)


source/routing/raptor_api.cpp
```diff
@@ -1152,7 +1152,7 @@ void make_isochrone(navitia::PbCreator& pb_creator,
     auto departures = get_stop_points(origin, raptor.data, worker);
 
     if(departures.size() == 0){
-        pb_creator.set_response_type(pbnavitia::NO_ORIGIN_POINT);
+        pb_creator.fill_pb_error(pbnavitia::Error::unknown_object, "The entry point: " + origin.uri + " is not valid");
```
antoine-de: hum I'm not convinced this is the right place to change, for me you should change the `get_stop_points` message no ?

it's not the same to start from a far away location or from a VJ

kadhikari:
@antoine-de  i push with some modifications

source/jormungandr/tests/journey_common_tests.py
```diff
@@ -825,6 +825,16 @@ def test_call_kraken_boarding_alighting(self):
         assert journey['sections'][4]['departure_date_time'] == '20120614T233500'
         assert journey['sections'][4]['arrival_date_time'] == '20120614T233500'
 
+    def test_journey_from_non_valid_object_type(self):
+        """
+        When the departure is a non valid stop_area, the response status should be 404
```
antoine-de: hum it's not only for stop_area, we can start from any places

so I guess it's more `When the departure is a not a Place, the response status should be 404`

source/jormungandr/tests/isochrone_tests.py
```diff
@@ -90,12 +90,16 @@ def test_to_isochrone_coord(self):
     def test_from_isochrone_sa(self):
         query = "v1/coverage/basic_routing_test/journeys?from={}&datetime={}"
         query = query.format("stopA", "20120614T080000")
-        response = self.query(query)
+        normal_response, error_code = self.query_no_assert(query)
+        assert normal_response['error']['message'] == 'The entry point: stopA is not valid'
```
antoine-de: why stopA is not valid ?

source/jormungandr/tests/journey_common_tests.py
```diff
@@ -825,6 +825,16 @@ def test_call_kraken_boarding_alighting(self):
...
+        """
+        query = "vehicle_journeys/vehicle_journey:SNC/journeys?datetime={datetime}".format(datetime="20120614T080000")
+        response = self.query_region(query, check=False)
```
antoine-de: it's easier to deconstruct the pair immediatly:

```
response, status = self.query_region(query, check=False)
assert status ==  404
```

source/kraken/worker.cpp
```diff
@@ -1051,7 +1051,7 @@ void Worker::nearest_stop_points(const pbnavitia::NearestStopPointsRequest& requ
     street_network_worker->init(entry_point, {});
     //kraken don't handle reverse isochrone
     auto result = routing::get_stop_points(entry_point, *data, *street_network_worker, false);
```
antoine-de, line 1053: you should check if result is valid before no ?

source/routing/raptor_api.cpp
```diff
@@ -922,6 +922,9 @@ get_stop_points( const type::EntryPoint &ep, const type::Data& data,
             }
         }
         LOG4CPLUS_DEBUG(logger, result.size() << " sp found for admin");
+    } else {
+        LOG4CPLUS_WARN(logger, "invalid entry point object " << ep.uri);
```
antoine-de: with the tests below, I think we can give a better error message, saying that ep.type is not valid

source/routing/raptor_api.cpp
```diff
@@ -1150,9 +1153,8 @@ void make_isochrone(navitia::PbCreator& pb_creator,
     datetime = tmp_datetime.front();
     worker.init(origin);
     auto departures = get_stop_points(origin, raptor.data, worker);
-
-    if(departures.size() == 0){
-        pb_creator.set_response_type(pbnavitia::NO_ORIGIN_POINT);
+    if (departures == nullptr) {
```
antoine-de: you shouldn't check the equality to `nullptr`, only the boolean conversion
`if (departures) {`

source/routing/raptor_api.cpp
```diff
@@ -1053,7 +1056,7 @@ void make_response(navitia::PbCreator& pb_creator,
     auto destinations = get_stop_points(destination, raptor.data, worker, true);
     const auto direct_path = get_direct_path(worker, origin, destination);
 
-    if(departures.size() == 0 && destinations.size() == 0){
+    if(departures && (departures->size() == 0) && destinations && (destinations->size() == 0)){
```
antoine-de: I don't understand, you don't check `departures` and `arrivals` ?

source/jormungandr/tests/isochrone_tests.py
kadhikari: The response of the type "no_solution" does not raise error while using "response = self.query(query)" in the test. 

After this correction we have "unknown_object" in the response which is not supported by "response = self.query(query)" in the test.

source/kraken/worker.cpp
```diff
@@ -1051,11 +1051,13 @@ void Worker::nearest_stop_points(const pbnavitia::NearestStopPointsRequest& requ
...
-    for(const auto& item: *result){
-        auto* nsp = pb_creator.add_nearest_stop_points();
-        this->pb_creator.fill(planner->get_sp(item.first), nsp->mutable_stop_point(), 0);
-        nsp->set_access_duration(item.second.total_seconds());
-    }   
+    if (result){
```
antoine-de: don't you want an error if the result is false ?

source/jormungandr/tests/isochrone_tests.py
antoine-de: but I don't understand why stopA is not valid

source/routing/raptor_api.cpp
```diff
@@ -1278,18 +1291,19 @@ static bool fill_isochrone_common(IsochroneCommon& isochrone_common,
     worker.init(center);
     auto departures = get_stop_points(center, raptor.data, worker);
 
-    if (departures.empty()) {
```
antoine-de: you also want another error on `! departures` no?

kadhikari:
@antoine-de  could you verify my last modifications please?

source/jormungandr/tests/isochrone_tests.py
```diff
@@ -87,6 +87,9 @@ def test_to_isochrone_coord(self):
         query = query.format(s_coord, "20120614T080000")
         response = self.query(query)
 
+    #In the previous version, the result contained no_solution with error message : 'several errors occured: \n * '
+    #with status code 200.
+    #After this correction, since the EntryPoint is absent an error is raised.
     def test_from_isochrone_sa(self):
```
antoine-de: maybe rename the test function as `test_isochrone_from_unkown_sa`

source/kraken/worker.cpp
```diff
@@ -1051,7 +1051,10 @@ void Worker::nearest_stop_points(const pbnavitia::NearestStopPointsRequest& requ
     street_network_worker->init(entry_point, {});
     //kraken don't handle reverse isochrone
     auto result = routing::get_stop_points(entry_point, *data, *street_network_worker, false);
-    if (result){
+    if (!result){
+        this->pb_creator.fill_pb_error(pbnavitia::Error::unknown_object, "The entry point: " + entry_point.uri + " is not valid");
+        return;
+    } else {
```
antoine-de: since you return you don't need the `else`

source/kraken/worker.cpp
```diff
@@ -1051,7 +1051,10 @@ void Worker::nearest_stop_points(const pbnavitia::NearestStopPointsRequest& requ
     street_network_worker->init(entry_point, {});
     //kraken don't handle reverse isochrone
     auto result = routing::get_stop_points(entry_point, *data, *street_network_worker, false);
-    if (result){
+    if (!result){
+        this->pb_creator.fill_pb_error(pbnavitia::Error::unknown_object, "The entry point: " + entry_point.uri + " is not valid");
```
antoine-de: I think we should make the distinction between `not valid` and `unknown` no ?

kadhikari: The false value of the function "get_stop_points" means we don't know the type of EntryPoint.

antoine-de (approved):

PR #2170, opened 2017-08-01
journeys get_label(). return name before code

bjamet:
I propose this modification to change the "display_information" in a section type:"public_transport". 
The fields "code" and "label" were always valued with the code. 
With this modification, the name of the line is accessible in the display_information in the "label" field

### DIFF
--- a/source/jormungandr/jormungandr/interfaces/v1/fields.py
+++ b/source/jormungandr/jormungandr/interfaces/v1/fields.py
@@ -357,12 +357,12 @@ class stop_time_properties_links(fields.Raw):
 class get_label(fields.Raw):
 
     def output(self, key, obj):
-        if obj.code != '':
-            return obj.code
+        if obj.name != '':
+            return obj.name
         else:
-            if obj.name != '':
-                return obj.name
-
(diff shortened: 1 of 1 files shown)


buildbotctp:

buildbotctp:

kinnou02:
The label is here for facilitate the integration of navitia, in most case you can just display the label and you will have something a traveller will understand. 
As such we believe that most traveler known the code of their line and not necessarily its name.
In almost every dataset we have seen the name of a line is something like this: `{origin} - {destination}`
it doesn't really help most of the time.

If you need the line name we could add it.

bjamet:
Thank you for your answer.
On the network I'm working on, the code must be understand for everyday users, but maybe not by casual or tourists. That's why we need the line name on the result of the search. 
Since the code value seams duplicated in the label, I proposed this solution. If you can add the line name, feel free to reject this pull request. 
Thank you again

antoine-de:
I think we can close it since there is now the `name` in the display information (#2175)

I close this, but feel free to reopen it if you don't think it solves your problem.

Thanks for the modification!

PR #2173, opened 2017-08-02
 	Jormgundanr: add swagger definitions for schedules

antoine-de:
add swagger definition for `/stop_schedules`, `/route_schedules`, `/departures` and `/arrivals`

There is one gotcha, in `/stop_schedules` and `/route_schedules`, we display null field and this does not seems accepted by swagger 2, and swagger 3 codegen is not yet available :cry: 

so the swagger validator in the 2 is not happy, but the swagger 2 codegen seems to accept the null values...
so for the moment we let this as it is.

### DIFF
--- a/source/jormungandr/jormungandr/interfaces/v1/Schedules.py
+++ b/source/jormungandr/jormungandr/interfaces/v1/Schedules.py
@@ -43,7 +43,8 @@ from jormungandr.interfaces.v1.decorators import get_obj_serializer
 from jormungandr.interfaces.v1.serializer import api
 import datetime
 from jormungandr.interfaces.argument import ArgumentDoc
-from jormungandr.interfaces.parsers import DateTimeFormat, default_count_arg_type
+from jormungandr.interfaces.parsers import DateTimeFormat, default_count_arg_type, depth_argument, \
+    UnsignedInteger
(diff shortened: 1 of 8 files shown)


source/jormungandr/jormungandr/interfaces/v1/Schedules.py
```diff
@@ -73,14 +73,14 @@ def __init__(self, endpoint, *args, **kwargs):
...
+        parser_get.add_argument("depth", type=depth_argument, default=2)
         parser_get.add_argument("count", type=default_count_arg_type, default=10,
                                 help="Number of schedules per page")
         parser_get.add_argument("start_page", type=int, default=0,
                                 help="The current page")
-        parser_get.add_argument("max_date_times", type=natural,
+        parser_get.add_argument("max_date_times", type=UnsignedInteger(), deprecated=True,
                                 help="DEPRECATED, replaced by `items_per_schedule`")
-        parser_get.add_argument("forbidden_id[]", type=six.text_type,
+        parser_get.add_argument("forbidden_id[]", type=six.text_type,deprecated=True,
```
TeXitoi: lack space

TeXitoi (approved):

kinnou02 (approved):

PR #2175, opened 2017-08-04
Jormun: add the line name in display information

kinnou02:


antoine-de:
This should do what you want @bjamet 

I think we can thus close #2170

bjamet:
I think it will, thank you

PR #2181, opened 2017-08-17
Kraken: fix vehicle_journeys bad index

xlqian:


TeXitoi (approved):

source/type/type.cpp
```diff
@@ -445,9 +445,9 @@ void cleanup_useless_vj_link(const nt::VehicleJourney* vj, nt::PT_Data& pt_data)
     pt_data.headsign_handler.forget_vj(vj);
 
     // remove the vj from the global list/map
-    auto next_vj = erase_vj_from_list(vj, pt_data.vehicle_journeys);
+    erase_vj_from_list(vj, pt_data.vehicle_journeys);
```
pbougue, line 448: > then I guess this function shouldn't return anything as it's quite "unsafe" to use.

:white_check_mark: Done

pbougue (approved):
OK, this should be good now.
So v2.38.0 is bugged but no hotfix will be done, correction in 2.39.0 ( @eturck  )

Issue #2147, opened 2017-07-17
Fare zone id should be string


pbougue:
Following that conversation : https://github.com/CanalTP/navitia/commit/70093beaab61e8934bf76ed1810429b72c557e43#commitcomment-23134878

Problem: Fare zone ids being string does happen. ( @bjamet if you can mention your example we are curious about it, but if you can't no problem).

TODO: Requires some changes here and there (not big), don't forget to change the data version number (https://github.com/CanalTP/navitia/blob/a2b42d5aa3845559b6a00629561d2d1fac057d11/source/type/data.cpp#L67).
But for a newcomer most of the work would be on compiling the whole navitia project and run tests... 

Status: Anyway you can create a PR on that and we will gladly provide help. Otherwise we will probably create a PR on that so that it is solved in the next 2-3 versions (so in prod before 09/2017).
> Just let us know here if you start something so that we don't work in double (will do it if we start).

PR #2187, opened 2017-08-30
ed: fare zone id changed to string

kadhikari:
Ticket concerned: https://jira.canaltp.fr/browse/NAVITIAII-2273

Issue concerned: https://github.com/CanalTP/navitia/issues/2147

### DIFF
--- a/source/ed/types.h
+++ b/source/ed/types.h
@@ -300,7 +300,7 @@ struct StopPoint : public Header, Nameable, hasProperties{
     StopArea* stop_area;
     Network* network;
 
-    StopPoint(): fare_zone(), stop_area(NULL), network(NULL) {}
+    StopPoint(): fare_zone(), stop_area(nullptr), network(nullptr) {}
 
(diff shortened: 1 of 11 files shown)


pbougue (dismissed):
Need to change data version number
Also needs to migrate ed database field from int to string

source/ed/types.h
```diff
@@ -300,7 +300,7 @@ struct StopPoint : public Header, Nameable, hasProperties{
     StopArea* stop_area;
     Network* network;
 
-    StopPoint(): fare_zone(0), stop_area(NULL), network(NULL) {}
+    StopPoint(): fare_zone(), stop_area(NULL), network(NULL) {}
```
xlqian: ~NULL~ -> nullptr

pbougue (approved):
Checked that it worked before/after migration (fusio2ed, ed2nav, kraken on fr-idf with fare). So it looks good!

TeXitoi (approved):

PR #2191, opened 2017-09-04
jormun: add timeout and circuit breaker in stat_manager

kinnou02:


source/jormungandr/jormungandr/stat_manager.py
```diff
@@ -114,36 +116,35 @@ def tz_str_to_utc_timestamp(dt_str, timezone):
...
         """
         connection to rabbitmq and initialize queues
         """
-        try:
-            self.connection = kombu.Connection(self.broker_url)
-            exchange = kombu.Exchange(self.exchange_name, type="topic")
-            self.producer = self.connection.Producer(exchange=exchange)
-        except:
-            logging.getLogger(__name__).exception('Unable to activate the producer of stat')
+        self.connection = kombu.Connection(self.broker_url, connnect_timeout=self.connection_timeout)
```
pbougue: important typo: con`n`nect_timeout

source/jormungandr/jormungandr/stat_manager.py
```diff
@@ -114,36 +116,35 @@ def tz_str_to_utc_timestamp(dt_str, timezone):
...
+        self.broker_url = app.config.get('BROKER_URL', None)
+        self.exchange_name = app.config.get('EXCHANGE_NAME', None)
+        self.connection_timeout = app.config.get('STAT_CONNECTION_TIMEOUT', 1)
 
         if self.save_stat:
-            self._init_rabbitmq()
+            try:
+                self._init_rabbitmq()
+            except Exception:
+                logging.getLogger(__name__).exception('Unable to activate the producer of stat')
```
pbougue, line 131: Don't we want to log the exception also?
> OK, @antoine-de explained that exception is logged

source/jormungandr/jormungandr/stat_manager.py
kinnou02: good catch!

PR #2194, opened 2017-09-08
sql: fix on error while downgrading the fare_zone type from string to integer

kadhikari:
Linked to : https://github.com/CanalTP/navitia/pull/2187

### DIFF
--- a/source/sql/alembic/versions/46168c7abc89_change_fare_zone_id_to_string.py
+++ b/source/sql/alembic/versions/46168c7abc89_change_fare_zone_id_to_string.py
@@ -16,9 +16,9 @@ import geoalchemy2 as ga
 from sqlalchemy.dialects import postgresql
 
 def upgrade():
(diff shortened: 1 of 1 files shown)


source/sql/alembic/versions/46168c7abc89_change_fare_zone_id_to_string.py
```diff
@@ -16,9 +16,9 @@
 from sqlalchemy.dialects import postgresql
 
 def upgrade():
-    op.execute("ALTER TABLE navitia.stop_point ALTER COLUMN fare_zone TYPE text;")
+    op.execute("ALTER TABLE navitia.stop_point ALTER COLUMN fare_zone TYPE text using fare_zone::text;")
```
kadhikari, line 19: Downgrade  -1 followed  by Upgrade head works with/without "using fare_zone::text" added in the methode upgrade.

TeXitoi (approved):

PR #2207, opened 2017-09-21
Jormun: add possibility to configure max count value

sign0:
Added
- form jormungandr import app
- app.config['PARSER_MAX_COUNT']

### DIFF
--- a/source/jormungandr/jormungandr/interfaces/parsers.py
+++ b/source/jormungandr/jormungandr/interfaces/parsers.py
@@ -29,6 +29,7 @@
 from __future__ import absolute_import
 from dateutil import parser
 from jormungandr.interfaces.v1.serializer.jsonschema.fields import DateTimeType
+form jormungandr import app
 from navitiacommon import parser_args_type
 
 # TODO: to be moved completely into navitiacommon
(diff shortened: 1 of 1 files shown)


buildbotctp:

buildbotctp:

source/jormungandr/jormungandr/interfaces/parsers.py
```diff
@@ -38,7 +39,10 @@
 
 float_gt_0 = parser_args_type.PositiveFloat()
 
-default_count_arg_type = parser_args_type.IntervalValue(min_value=0, max_value=1000)
+parser_min_count = app.config['PARSER_MIN_COUNT']
```
antoine-de: I don't think it's worth having a `min_parameter`

(and it's not very important there, but It's better to create only one functionally-complete PR instead of 1 per file)

sign0: It makes sense... I confess I hesitated.

antoine-de (approved):

kinnou02:
ok to test

kinnou02 (approved):

PR #2210, opened 2017-09-25
Kraken: don't segfault when 2 different disruptions modify the same vj

TeXitoi:
If we have a disruption updating the stop times of a vj, and then another disruption on the same vj with another id, we have dangling pointers.  It should not arrive but better to secure that.

This PR fixes that in a simple way:
  * the disruption is kept, as some tests require it.
  * the disruption will be printed when it applies, even if the other disruption override it.
  * the impacted stops will be printed relative to UTC instead of local.

This is safe, but not the prettiest. As it should not happen, keep it simple.

### DIFF
--- a/source/type/message.cpp
+++ b/source/type/message.cpp
@@ -329,6 +329,16 @@ void DisruptionHolder::clean_weak_impacts(){
     clean_up_weak_ptr(weak_impacts);
 }
 
+void DisruptionHolder::forget_vj(const VehicleJourney* vj) {
+    for (const auto& impact: vj->meta_vj->get_impacts()) {
+        for (auto& stu: impact->aux_info.stop_times) {
+            if (stu.stop_time.vehicle_journey == vj) {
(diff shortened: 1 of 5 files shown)


source/type/message.cpp
```diff
@@ -329,6 +329,17 @@ void DisruptionHolder::clean_weak_impacts(){
     clean_up_weak_ptr(weak_impacts);
 }
 
+void DisruptionHolder::forget_vj(const VehicleJourney* vj) {
+    for (const auto& impact: vj->meta_vj->get_impacts()) {
+        for (auto& stu: impact->aux_info.stop_times) {
+            if (stu.stop_time.vehicle_journey == vj) {
+                std::cout << "problem with impact " << impact->uri << " and vj " << vj->uri << std::endl;
```
xlqian: :no_good: you may want a log here

TeXitoi: No I just need to remove this debug print.

kinnou02 (approved):

PR #2212, opened 2017-10-02
Jormungandr: get all bss stands in one call to jcdecaux api

kinnou02:
With this PR we only call the jcdecaux's api one time for all the
stations, this result is then cached so it be reuse for other request.
This will greatly improve performance on call like this one:
https://api.navitia.io/v1/coverage/fr-idf/poi_types/poi_type%3Aamenity%3Abicycle_rental/pois?count=500
In most case the hard timeout of 30sec was hit before navitia can
respond.

### DIFF
--- a/source/jormungandr/jormungandr/parking_space_availability/bss/jcdecaux.py
+++ b/source/jormungandr/jormungandr/parking_space_availability/bss/jcdecaux.py
@@ -95,3 +95,7 @@ class JcdecauxProvider(AbstractParkingPlacesProvider):
 
     def feed_publisher(self):
         return self._feed_publisher
+
+    def __repr__(self):
+        #TODO: make this shit python 3 compatible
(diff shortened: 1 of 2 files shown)


source/jormungandr/jormungandr/parking_space_availability/bss/jcdecaux.py
```diff
@@ -81,12 +84,18 @@ def _call_webservice(self, station_id):
...
+            return Stands(data[ref]['available_bike_stands'], data[ref]['available_bikes'])
 
     def status(self):
         return {'network': self.network, 'operators': self.operators, 'contract': self.contract}
 
     def feed_publisher(self):
         return self._feed_publisher
+
+    def __repr__(self):
+        #TODO: make this shit python 3 compatible
```
TeXitoi, line 100: Can't you do that now?

TeXitoi (approved):

pbougue (approved):

PR #2213, opened 2017-10-02
All: vj disruptions on stop_schedules

TeXitoi:
With this, we have access to the realtime disruptions in stop_schedules

### DIFF
--- a/source/jormungandr/tests/line_sections_tests.py
+++ b/source/jormungandr/tests/line_sections_tests.py
@@ -35,9 +35,9 @@ from .check_utils import *
 
 def impacted_ids(disrupts):
     # for the impacted obj, either get the id, or the headsign (for the display_information field)
-    return set(o.impacted_object.get('id',
-                                     o.impacted_object.get('headsign'))
-               for o in disrupts['line_section_on_line_1'])
+    return set(o.impacted_object.get('id', o.impacted_object.get('headsign'))
+               for o in disrupts['line_section_on_line_1']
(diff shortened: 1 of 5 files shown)


kinnou02 (approved):

source/type/pb_converter.cpp
```diff
@@ -1341,6 +1341,13 @@ void PbCreator::Filler::fill_pb_object(const StopTimeCalandar* stop_time_calenda
                                          true);
             rs_date_time->set_base_date_time(to_posix_timestamp(base_dt));
         }
+
+        const std::vector<const nt::StopTime*> stop_times = {stop_time_calendar->stop_time};
+        const auto meta_vj = stop_time_calendar->stop_time->vehicle_journey->meta_vj;
+        for (const auto& message: meta_vj->get_applicable_messages(pb_creator.now, pb_creator.action_period)) {
+            if (! message->is_relevant(stop_times)) { continue; }
```
pbougue: minor: looks better if you test is_relevant, then fill (avoid that `continue`).

source/jormungandr/tests/line_sections_tests.py
```diff
@@ -35,9 +35,9 @@
 
 def impacted_ids(disrupts):
     # for the impacted obj, either get the id, or the headsign (for the display_information field)
```
pbougue: minor: maybe more details on comment, but... "Looking for (identifiable) objects impacted by 'disruptions' inputed" (so then, rename parameter)

pbougue (approved):

PR #2220, opened 2017-10-09
PbConverter: fix output of StopTime

TeXitoi:


source/type/pb_converter.cpp
```diff
@@ -834,27 +834,33 @@ void PbCreator::Filler::fill_pb_object(const nt::StopPointConnection* c, pbnavit
...
 
-static uint32_t get_st_utc_offset(const nt::StopTime* st, const uint32_t time) {
+static uint32_t get_local_time(const nt::StopTime* st, const uint32_t time) {
     static const auto flag = std::numeric_limits<uint32_t>::max();
-    return time == flag || st->vehicle_journey == nullptr ? 0 : st->vehicle_journey->utc_to_local_offset();
+    uint32_t offset = 0;
+    if (time != flag && st->vehicle_journey != nullptr) {
+        // getting the positive offset to be safe about integer overflow
+        const int32_t day = DateTimeUtils::SECONDS_PER_DAY;
+        offset = (st->vehicle_journey->utc_to_local_offset() % day + day) % day;
```
xlqian: i think a little comment on this line will be very useful one day :) (or just a simple link of stackoverflow)

xlqian (approved):

source/type/pb_converter.cpp
xlqian:
```
If both operands are nonnegative then the remainder is nonnegative; if not, the sign of the remainder is implementation-defined
```
 :joy:

source/tests/timezone_cape_verde_test.cpp
```diff
@@ -0,0 +1,78 @@
...
+
+    const auto date = "20170101";
+    const auto begin = boost::gregorian::date_from_iso_string(date);
+    const boost::gregorian::date_period production_date = {begin, begin + boost::gregorian::years(1)};
+    navitia::type::TimeZoneHandler::dst_periods timezones = {{-3600, {production_date}}};
+    ed::builder b("20170101", "canal tp", "Atlantic/Cape_Verde", timezones);
+
+    b.vj("line:1", "000010")
+        .uri("vj:1:1")
+        ("A", "23:00"_t)
```
pbougue: Please add a comment to tell it's UTC when using helper :)

pbougue (approved):

PR #2226, opened 2017-10-13
jormun: add a few info on the autocomplete and streetnetwork backend in coverage's status

kinnou02:


xlqian (approved):

source/jormungandr/jormungandr/interfaces/v1/serializer/status.py
```diff
@@ -80,6 +80,28 @@ class TravelerProfilesSerializer(serpy.Serializer):
     walking_speed = Field(schema_type=float)
     wheelchair = Field(schema_type=bool)
 
+class AutocompleteSerializer(serpy.DictSerializer):
+    class_ = Field(schema_type=str, label='class', attr='class')
```
pbougue, line 84: Did you forget timeout? (probably display_none=False)

pbougue (approved):
Can you add serpy test for all that even if it will probably not test all configs (simple one like https://github.com/CanalTP/navitia/blob/dev/source/jormungandr/tests/schema_tests.py#L190) ?
Also: Don't you want to also add an overview of it directly in v1/status ? This can still be a good step as is though. But if coverages tend to disappear, can be nice to have a mimir status there (not sure about street network, but probably too)

kinnou02:
There is already a test for schema status: https://github.com/CanalTP/navitia/blob/dev/source/jormungandr/tests/schema_tests.py#L281

source/jormungandr/jormungandr/interfaces/v1/serializer/status.py
TeXitoi: done

source/jormungandr/jormungandr/interfaces/v1/serializer/status.py
```diff
@@ -80,6 +80,32 @@ class TravelerProfilesSerializer(serpy.Serializer):
...
+    current_state = Field(schema_type=str, display_none=True)
+    fail_counter = Field(schema_type=int, display_none=True)
+    reset_timeout = Field(schema_type=int, display_none=True)
+
+class StreetNetworkSerializer(serpy.DictSerializer):
+    class_ = Field(schema_type=str, label='class', attr='class')
+    id = Field(schema_type=str)
+    modes = StringListField(display_none=True)
+    timeout = MethodField(schema_type=float, display_none=False)
+    circuit_breaker = MethodField(schema_type=float, display_none=False)
```
TeXitoi: schema_type is wrong. And why do you use MethodField everywhere? That's the last ugly fallback.

kinnou02: hum, forgot to update it :(

I use method field because serpy cannot serialize an optional attr with a DictSerializer

TeXitoi (approved):

PR #2229, opened 2017-10-19
Jormungandr: disabling of sql connection pooling

kinnou02:
This PR give us the possibility to disable connection pooling to the postgresql database.
Currently each jormungrandr process keep it's sql connection open forever.
The number of connection a database can handle is limited and can't be increased without a restart.
Pooling is enabled by default because it increases the performance by not having to reopen a connection for each query, but in production we already use an external connection pooler (pg_pool) that allows us to share connections between multiple process or even box.

With this change we will be able to increase the number of jormungandr processes without impacting the database too much

### DIFF
--- a/source/jormungandr/jormungandr/default_settings.py
+++ b/source/jormungandr/jormungandr/default_settings.py
@@ -156,3 +156,7 @@ GREENLET_POOL_SIZE = int(os.getenv('JORMUNGANDR_GEVENT_POOL_SIZE', 10))
 USE_SERPY = boolean(os.getenv('JORMUNGANDR_USE_SERPY', False))
 
 PARSER_MAX_COUNT = int(os.getenv('JORMUNGANDR_PARSER_MAX_COUNT', 1000))
+
+if boolean(os.getenv('JORMUNGANDR_DISABLE_SQLPOOLING', False)):
+    from sqlalchemy.pool import NullPool
+    SQLALCHEMY_POOLCLASS = NullPool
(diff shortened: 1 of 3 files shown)


pbougue (approved):

source/navitiacommon/navitiacommon/sqlalchemy.py
```diff
@@ -0,0 +1,45 @@
+# Copyright (c) 2001-2014, Canal TP and/or its affiliates. All rights reserved.
```
xlqian, line 1: :D a very nonsense comment: 2001-2017

source/jormungandr/jormungandr/default_settings.py
```diff
@@ -156,3 +156,7 @@
 USE_SERPY = boolean(os.getenv('JORMUNGANDR_USE_SERPY', False))
 
 PARSER_MAX_COUNT = int(os.getenv('JORMUNGANDR_PARSER_MAX_COUNT', 1000))
+
+if boolean(os.getenv('JORMUNGANDR_DISABLE_SQLPOOLING', False)):
```
xlqian, line 160: hmmm, i don't know if it's a good idea to put a Class in a setting file, if I were you, i'm gonna keep the boolean in the setting file and use the real class in the implementation, it's up to you

xlqian (approved):

source/jormungandr/jormungandr/default_settings.py
kinnou02: default_settings as changed a lot since it's creation, it has become more a configuration parser from envvar and less a configuration file.
When we will update the deployment of jormungandr we won't create a "settings.py" anymore and configure everything by environment variables

PR #2231, opened 2017-10-20
jormungandr: attribut 'templated' deleted from args

kadhikari:
Before calling the function "create_external_link" with parameter "templated", we should delete this parameter in the parameter **kwargs if present.

### DIFF
--- a/source/jormungandr/jormungandr/interfaces/v1/Journeys.py
+++ b/source/jormungandr/jormungandr/interfaces/v1/Journeys.py
@@ -328,8 +328,10 @@ class add_journey_href(object):
                     args['is_journey_schedules'] = True
                     allowed_ids.update(args.get('allowed_id[]', []))
                     args['allowed_id[]'] = list(allowed_ids)
+                    args['_type'] = 'journeys'
+
                     journey['links'] = [
-                        create_external_link('v1.journeys', rel='same_journey_schedules', _type='journeys', **args)
+                        create_external_link('v1.journeys', rel='same_journey_schedules', **args)
                     ]
             return objects
         return wrapper
--- a/source/jormungandr/jormungandr/interfaces/v1/Journeys.py
+++ b/source/jormungandr/jormungandr/interfaces/v1/Journeys.py
@@ -308,7 +308,8 @@ class add_journey_href(object):
                         args['to'] = journey['to']['id']
                     if 'from' not in args:
                         args['from'] = journey['from']['id']
-                    journey['links'] = [create_external_link('v1.journeys', rel='journeys', **args)]
+                    args['rel'] = 'journeys'
+                    journey['links'] = [create_external_link('v1.journeys', **args)]
                 elif allowed_ids and 'public_transport' in (s['type'] for s in journey['sections']):
                     # exactly one first_section_mode
                     if any(s['type'].startswith('bss') for s in journey['sections'][:2]):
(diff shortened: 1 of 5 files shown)


source/jormungandr/jormungandr/interfaces/v1/Schedules.py
```diff
@@ -342,6 +342,8 @@ def wrapper(*args, **kwargs):
             delta = timedelta(seconds=1)
             dt = datetime.strptime(min_dt, "%Y%m%dT%H%M%S")
             if g.stat_interpreted_parameters.get('data_freshness') != 'realtime':
+                if "templated" in kwargs_links:
+                    del kwargs_links["kwargs"]
```
pbougue: you probably meant: del kwargs_links["`templated`"]

pbougue (commented):
I'd like some explainations on why it happens and why this is the clean way of doing it, because I don't understand much of it :)

kadhikari:
@pbougue when "create_external_link" is called with the parameter "templated" and with presence of attribut "templated": ".." in kwargs, we have an error : "exceptions:TypeError: create_external_link() got multiple values for keyword argument 'templated' "

When we add &templated=true/false in the url navitia, the same error is raised.

source/jormungandr/jormungandr/interfaces/v1/serializer/api.py
```diff
@@ -212,6 +212,8 @@ def get_links(self, obj):
                 else:
                      args[e.key] = e.values[0]
 
+            if "templated" in args:
```
xlqian: how about doing `args["templated"] = value.is_templated`, then remove `templated` at https://github.com/CanalTP/navitia/pull/2231/files#diff-4208927b849d6bd230bd4d6e04d9aa6cR220 ?

source/jormungandr/jormungandr/interfaces/v1/make_links.py
```diff
@@ -225,9 +227,11 @@ def wrapper(*args, **kwargs):
             if isinstance(data, dict) or isinstance(data, OrderedDict):
                 data = self.prepare_objetcs(objects, True)
                 kwargs = self.prepare_kwargs(kwargs, data)
+                if "templated" in kwargs:
+                    del kwargs["kwargs"]
```
xlqian: you can just overwrite `templated` and remove it from the arguments

source/jormungandr/jormungandr/interfaces/v1/Journeys.py
```diff
@@ -328,8 +328,10 @@ def wrapper(*args, **kwargs):
                     args['is_journey_schedules'] = True
                     allowed_ids.update(args.get('allowed_id[]', []))
                     args['allowed_id[]'] = list(allowed_ids)
+                    args['_type'] = 'journeys'
+
```
pbougue: extra blank line + why don't you do the same for `rel`? (this stands for the whole PR)

pbougue (approved):
Looks pretty good that way (let's discuss about `rel`)

source/jormungandr/jormungandr/interfaces/v1/Journeys.py
kadhikari: because rel is not an optional parameter

kadhikari:
jenkins test this please

source/jormungandr/jormungandr/interfaces/v1/make_links.py
```diff
@@ -36,7 +36,7 @@
 from flask.ext.restful.utils import unpack
 
 
-def create_external_link(url, rel, _type=None, templated=False, description=None, **kwargs):
+def create_external_link(url, rel=None, _type=None, templated=False, description=None, **kwargs):
```
pbougue: most probably, you can remove that None

pbougue (approved):

PR #2233, opened 2017-10-24
All: New chaos channel types added in navitia

kadhikari:
- New  channel types 'title' and 'beacon' added.
- Some tests added

### DIFF
--- a/source/kraken/make_disruption_from_chaos.cpp
+++ b/source/kraken/make_disruption_from_chaos.cpp
@@ -258,7 +258,9 @@ static std::set<nt::disruption::ChannelType> create_channel_types(const chaos::C
             res.insert(nt::disruption::ChannelType::beacon);
             break;
         default:
-            throw navitia::exception("Unhandled ChannelType value in Chaos.Proto");
+            res.insert(nt::disruption::ChannelType::unknown_type);
+            LOG4CPLUS_TRACE(log4cplus::Logger::getInstance("log"),
+                            "Unhandled ChannelType value in Chaos.Proto: unknown_type used");
         }
     }
     return res;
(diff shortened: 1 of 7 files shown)


source/kraken/make_disruption_from_chaos.cpp
```diff
@@ -251,6 +251,12 @@ static std::set<nt::disruption::ChannelType> create_channel_types(const chaos::C
...
             res.insert(nt::disruption::ChannelType::unknown_type);
             break;
+        case chaos::Channel_Type_title:
+            res.insert(nt::disruption::ChannelType::title);
+            break;
+        case chaos::Channel_Type_beacon:
+            res.insert(nt::disruption::ChannelType::beacon);
+            break;
         default:
             throw navitia::exception("Unhandled ChannelType value in Chaos.Proto");
```
kinnou02: we should apply the same rules between disruptions from the database or directly from chaos.
So the default case should put an `Channel_Type_unkown_type`.
Also a log when this happen will be useful

kinnou02 (changes requested):

source/kraken/make_disruption_from_chaos.cpp
```diff
@@ -251,8 +251,16 @@ static std::set<nt::disruption::ChannelType> create_channel_types(const chaos::C
...
+            res.insert(nt::disruption::ChannelType::title);
+            break;
+        case chaos::Channel_Type_beacon:
+            res.insert(nt::disruption::ChannelType::beacon);
+            break;
         default:
-            throw navitia::exception("Unhandled ChannelType value in Chaos.Proto");
+            res.insert(nt::disruption::ChannelType::unknown_type);
+            LOG4CPLUS_WARN(log4cplus::Logger::getInstance("log"),
+                            "Unhandled ChannelType value in Chaos.Proto: unknown_type used");
```
pbougue, line 263: minor: indent (1 extra space) + I'd rather have unknown always just before default (but it's minor)

pbougue (approved):

TeXitoi (approved):

kinnou02 (approved):

PR #2236, opened 2017-10-27
Tyr: fix nb of args in tyr

xlqian:
https://jira.canaltp.fr/projects/NAVIO/issues/NAVIO-717


pbougue:
We should also add a try somewhere around that (even more as this code is useless when noone listens to that rabbit-exchange... which is the case now that we unplugged fenrir).
And understand why the rabbit-connection is considered down in prod.

kinnou02 (approved):

PR #2238, opened 2017-10-31
Kraken: find best UTC offset for a validity pattern

TeXitoi:
Sometime, a validity pattern will not be a subset of a time shift, for example if we shift the validity pattern for technical reasons.

Fix vehicle_journeys API in the times of the stop_times, that can be sometimes badly shifted (and thus gives wrong times).

### DIFF
--- a/source/time_tables/tests/departure_board_test_data.h
+++ b/source/time_tables/tests/departure_board_test_data.h
@@ -283,7 +283,7 @@ struct calendar_fixture {
         b.data->pt_data->codes.add(b.get<nt::StopPoint>("StopR4"), "source", "Code-StopR4");
 
         beg = b.data->meta->production_date.begin();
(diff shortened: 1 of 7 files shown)


source/time_tables/tests/departure_board_test_data.h
```diff
@@ -283,7 +283,7 @@ struct calendar_fixture {
         b.data->pt_data->codes.add(b.get<nt::StopPoint>("StopR4"), "source", "Code-StopR4");
 
         beg = b.data->meta->production_date.begin();
-        end_of_year = beg + boost::gregorian::years(1) + boost::gregorian::days(1);
+        end_of_year = beg + boost::gregorian::years(1);
```
pbougue, line 286: OK: this was preventing it to be a subset

pbougue (approved):

PR #2241, opened 2017-11-02
kraken: contributor of trip.extension used in disruption for gtfs-rt

kadhikari:


pbougue (commented):
Looks good, but we have to check how it will behave for SNCF, see if it changes the contributor (and if it used to be meaningfull): avoid breaking integrations that would filter on that (see if the question is real)

pbougue:
Jenkins, test this please

kadhikari:
A contributor is always present in 'IRE' while using KirinModelBuilder and a disruption always have a contributor associated in gtfs_realtime_pb2.

We can also differenciate a contributor associated to a disruption and a contributor (rt_topic ?) used to canalize disruptions in RabbitMQ...

buildbotctp:
Build finished.

kinnou02 (approved):

TeXitoi (approved):

PR #2243, opened 2017-11-07
Build: sync pip versions between tyr and jormun + boost 1.62

TeXitoi:
for boost 1.62, see workaround at https://svn.boost.org/trac10/ticket/12507

yeld_fixture is deprecated and an alias of fixture. It doesn't work with older py.test versions using yeld_fixture because such fixture must have yeld in it...

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -29,6 +29,4 @@ debian/tmp
 debian/*.substvars
 debian/*.debhelper
 source/jormungandr/jormungandr/_version.py
-/source/jormungandr/jormungandr/.cache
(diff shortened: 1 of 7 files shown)


buildbotctp:
Build finished.

TeXitoi:
`--logger` is new in 1.62, will correct that.

pbougue (approved):
Looks good but that's probably gonna be decorrelated next time someone adds/upgrades a requirement

buildbotctp:
Build finished.

PR #2245, opened 2017-11-07
Kraken: correctly handle disruption output with teleportation vj

TeXitoi:
When the vehicle journey do a teleportation (can be possible with minute rounding in data), we were using an action period [datetime, datetime[, which is empty and thus cannot intersect anything.

Fix https://jira.canaltp.fr/browse/NAVITIAII-2297


buildbotctp:
Build finished.

pbougue (approved):

xlqian (approved):

PR #2248, opened 2017-11-15
Jormun: Fix reverse geocoding

azime:
TODO:

- [x] Test with bragi
- [x]  Docs

### DIFF
--- a/documentation/slate/source/includes/apis.md
+++ b/documentation/slate/source/includes/apis.md
@@ -165,8 +165,12 @@ Very simple service: you give Navitia some coordinates, it answers you
 
 | url | Result |
 |----------------------------------------------|---------------------------------------------------------------------|
+| `places/{lon;lat}`                           | Detailed address point                                              |
+| `/places/{id}`                               | Information about places                                            |
 | `coverage/{lon;lat}/places/{lon;lat}`        | Detailed address point, navitia guesses the region from coordinates |
+| `coverage/{lon;lat}/places/{id}`             | Detailed about places, navitia guesses the region from coordinates  |
 | `coverage/{region_id}/places/{lon;lat}`      | Detailed address point                                              |
+| `coverage/{region_id}/places/{id}`           | Detailed about places                                              |
 
 
 You can also combine `/coords` with other filter as :
--- a/documentation/slate/source/includes/apis.md
+++ b/documentation/slate/source/includes/apis.md
@@ -163,14 +163,14 @@ Very simple service: you give Navitia some coordinates, it answers you
 
 ### Accesses
 
-| url | Result |
+| url                                          | Result                                                              |
 |----------------------------------------------|---------------------------------------------------------------------|
 | `places/{lon;lat}`                           | Detailed address point                                              |
 | `/places/{id}`                               | Information about places                                            |
 | `coverage/{lon;lat}/places/{lon;lat}`        | Detailed address point, navitia guesses the region from coordinates |
 | `coverage/{lon;lat}/places/{id}`             | Detailed about places, navitia guesses the region from coordinates  |
 | `coverage/{region_id}/places/{lon;lat}`      | Detailed address point                                              |
-| `coverage/{region_id}/places/{id}`           | Detailed about places                                              |
+| `coverage/{region_id}/places/{id}`           | Detailed places                                                     |
 
 
 You can also combine `/coords` with other filter as :
--- a/documentation/slate/source/includes/apis.md
+++ b/documentation/slate/source/includes/apis.md
@@ -165,9 +165,8 @@ Very simple service: you give Navitia some coordinates, it answers you
 
 | url | Result |
 |----------------------------------------------|---------------------------------------------------------------------|
-| `coords/{lon;lat}`                           | Detailed address point                                              |
-| `coverage/{lon;lat}/coords/{lon;lat}`        | Detailed address point, navitia guesses the region from coordinates |
-| `coverage/{region_id}/coords/{lon;lat}`      | Detailed address point                                              |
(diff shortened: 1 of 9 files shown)


TeXitoi:
Can't we just say that coord endpoints just directly handled by Places? That would be just a big code remove, and much simpler.

documentation/slate/source/includes/apis.md
```diff
@@ -165,9 +165,8 @@ Very simple service: you give Navitia some coordinates, it answers you
 
 | url | Result |
 |----------------------------------------------|---------------------------------------------------------------------|
-| `coords/{lon;lat}`                           | Detailed address point                                              |
```
pbougue: We have to replace it with `places/{id}` and `places/{lon;lat}` because it's nowhere else in the doc, and we have to mention it.

documentation/slate/source/includes/apis.md
```diff
@@ -165,9 +165,8 @@ Very simple service: you give Navitia some coordinates, it answers you
...
-| `coverage/{lon;lat}/coords/{lon;lat}`        | Detailed address point, navitia guesses the region from coordinates |
-| `coverage/{region_id}/coords/{lon;lat}`      | Detailed address point                                              |
+| `coverage/{lon;lat}/places/{lon;lat}`        | Detailed address point, navitia guesses the region from coordinates |
```
pbougue: We have to mention `coverage/{lon;lat}/places/{id}` also (it was not documented before)

documentation/slate/source/includes/apis.md
```diff
@@ -165,9 +165,8 @@ Very simple service: you give Navitia some coordinates, it answers you
...
+| `coverage/{region_id}/places/{lon;lat}`      | Detailed address point                                              |
```
pbougue: Have to mention `coverage/{region_id}/places/{id}` too

source/jormungandr/tests/bragi_autocomplete_tests.py
```diff
@@ -430,6 +430,32 @@ def test_global_place_uri(self):
...
+        kwargs = {
+            'params': {
+                'pt_dataset': 'main_routing_test',
+                'lon': 3.282103,
+                'lat': 49.84758,
+            },
+            'timeout': 10
+        }
+
+        url += "/reverse?{}".format(urlencode(kwargs.get('params'), doseq=True))
```
pbougue: So I guess you can shorten it a bit (only define 'params').
Same on test below.

source/jormungandr/jormungandr/interfaces/v1/Coord.py
```diff
@@ -28,60 +28,72 @@
...
             self.region = r
             result.update(regions=[r])
-            pb_result = i_manager.dispatch(args, "places_nearby", instance_name=r)
-            if len(pb_result.places_nearby) > 0:
-                e_type = pb_result.places_nearby[0].embedded_type
-                if _NAVITIATYPE.values_by_name["ADDRESS"].number == e_type:
-                    new_address = marshal(pb_result.places_nearby[0].address,
-                                          address)
-                    result.update(address=new_address)
+            if all([lon, lat, is_coord(id)]):
```
pbougue, line 84: Are you sure it's restrictive enough? I wonder about an url being `v1/coords/{lon;lat}/places_nearby`.
Maybe this test already exists though (or would be nice to have it).

pbougue (commented):
To reply to @TeXitoi (now that I read it...) : this is mainly a welcomed refactor, and the "dispatching" part is actually quite small if I understood correctly.

documentation/slate/source/includes/apis.md
azime: done

source/jormungandr/jormungandr/interfaces/v1/Coord.py
azime: "places_nearby" is called in this case

documentation/slate/source/includes/apis.md
```diff
@@ -165,9 +165,12 @@ Very simple service: you give Navitia some coordinates, it answers you
 
 | url | Result |
 |----------------------------------------------|---------------------------------------------------------------------|
-| `coords/{lon;lat}`                           | Detailed address point                                              |
-| `coverage/{lon;lat}/coords/{lon;lat}`        | Detailed address point, navitia guesses the region from coordinates |
-| `coverage/{region_id}/coords/{lon;lat}`      | Detailed address point                                              |
+| `places/{lon;lat}`                           | Detailed address point                                              |
+| `/places/{id}`                               | Information about places                                            |
+| `coverage/{lon;lat}/places/{lon;lat}`        | Detailed address point, navitia guesses the region from coordinates |
+| `coverage/{lon;lat}/places/{id}`             | Detailed about places, navitia guesses the region from coordinates  |
```
pbougue: `Information` about or `Detailed places`

documentation/slate/source/includes/apis.md
```diff
@@ -165,9 +165,12 @@ Very simple service: you give Navitia some coordinates, it answers you
...
+| `coverage/{region_id}/places/{lon;lat}`      | Detailed address point                                              |
+| `coverage/{region_id}/places/{id}`           | Detailed about places                                              |
```
pbougue: Same + indent on last `|`

pbougue (commented):
You may also check (or add) a test on `v1/coords/{lon;lat}/places_nearby`
> ☑️ [See](https://github.com/CanalTP/navitia/blob/dev/source/jormungandr/tests/places_tests.py#L109)

documentation/slate/source/includes/apis.md
azime: done

pbougue (approved):

PR #2250, opened 2017-11-17
Kraken: fix typo on InformedEntitiesLinker

eturck:
The message was always printed since we were checking an empty set instead of the list of impacted vjs.


buildbotctp:

pbougue (approved):

Issue #2009, opened 2017-03-08
Realtime proxy: custom merging of kraken / proxy passages


eturck:
We have started to test realtime proxy for our real time passages. They are in an Inéo database which have all the passages, but only for the current day (from 3:30am on day D to 3:30am on D+1).
When we request next departures for the current day we want to replace all theoretical departures by the ones from the proxy, so the current behavior is great 96% of the time for us.

Another 2% of request are done for a specific date, and @TeXitoi told me there was a way to specifically handle this, which is done in the timeo proxy, return None if the requested date is not today.

But for the last 2% we want to get all departure times at a stop, for the next 2 to 15 days. So we want to keep the proxy passages for the current day (until 3:30am tomorrow), and keep the theoretical ones for the rest of the period). And I can't do this yet.
We talked about it in the past with at least @kinnou02 but I never really looked at the code. After looking around a bit I see multiple ways to do it : 
- Just pass the theoretical passage to next_passage_for_route_point. But since we have two structures for them (passages / departure_board) we would maybe need to convert them before to RealTimePassage ?
- Make _update_passages and _update_stop_schedule function of the realtime_proxy which can be overloaded by each proxy ?
- Instead of merging passages just make the proxy carry the criteria for deleting the theoretical passages in the protobuf, append the realtime passages and sort everything by datetime ?

That's my current ideas, you have maybe some much better ones :smile:.
Let me know what you think about this!

pbougue:
I like the idea to delegate to proxy the possibility to provide a datetime after which you won't rely on it (so the last idea you mentionned I guess).
So the idea would be to forgive everything the proxy gives you after that dt, and everything base-schedule provides before that dt (no sort required).
Can be usefull when:
* we know that a RT service is OK only in the next hour, not after
* RT is OK only in the current opening hours of the line (maybe consider the hour in the middle of the night in case there are delays), not next day.

This dt might be computed considering what the service provides (like: we have RT passages for today and tomorrow, so wipeout base_schedule for today and tomorrow, and if only today then wipeout only today)

This is not perfect, but will never be (no vj matching). And it's the easier to understand and to use I can think of.

kinnou02:
I like your second proposition, so this mechanism is fully comprised in the proxy, we don't even have to implement it by default.

PR #2253, opened 2017-11-24
Jormungandr: ability for the proxy to merge kraken / realtime proxy datas

eturck:
This PR is related to #2009.

For us the goal is to be able to respond to a request "What are the real time departures at this stop for the next seven days?" with realtime datas, but with a proxy having less than 7 days of datas.
At Tisséo our realtime database has datas for the current day, until 3:30am. What we want to do with this type of request is remove all departures from kraken before 3:30am, and replace them with departures from the proxy, but keep everything else.

I have made two changes in this PR : 
- First, I have added `duration` as a parameter of `next_passage_for_route_point`. Currently, if you want the next departures for the next 3600 seconds, the proxy doesn't have this information, and has no limits. The default value is the same as in stop_schedule and departures duration.
- Second. I moved `_update_passage` and `_update_stop_schedule` in `realtime_proxy.py`. I have also moved the function parameter of pb_del_if in another method. When the merging criteria is simple enough, you just have to overload those functions (It is what I did in `proxy_realtime_tests.py` and our proxy).

Since _update_stop_schedule can be overloaded completely, we could add some custom informations on each passage in notes too (like short / long train @kinnou02 ;)).
I have used this on our instance and it seems to work well. Let me know what you think and if you see anything that needs some changes.

### DIFF
--- a/source/jormungandr/jormungandr/realtime_schedule/cleverage.py
+++ b/source/jormungandr/jormungandr/realtime_schedule/cleverage.py
@@ -128,7 +128,7 @@ class Cleverage(RealtimeProxy):
         else:
             return None
 
-    def _get_next_passage_for_route_point(self, route_point, count=None, from_dt=None, current_dt=None):
+    def _get_next_passage_for_route_point(self, route_point, count=None, from_dt=None, current_dt=None, duration=None):
         url = self._make_url(route_point)
         if not url:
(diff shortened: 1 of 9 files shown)


buildbotctp:

kinnou02 (approved):

PR #2257, opened 2017-11-28
All: Disruptions properties from chaos to jormungandr

vpassama:
Hi,

The goal of this PR is just to make properties linked to disruptions from chaos visible in jormungandr.
Because I add some 'join' statements in the SQL request in fill_disruption_from_database, it can raise the number of duplicated rows for every disruptions leading to a really wrong number when logging this sentence : 

`[] fill_disruption_from_database.cpp:194 [INFO ] - Loading x disruptions`

So I thought adding a small request with the number of impacts would be a good way to get a correct information from the log. But it's no big deal and I'll delete this commit if you don't want that extra request. Or maybe you don't care about having the impacts number and I can change it to a simple count request.

The output in jormungandr is like this : 

```
{
    "properties": [
        {
            "type": "datasource",
            "key": "foo",
            "value": "bar"
        },
        {...}
    ]
}
```

 Thanks.

### DIFF
--- a/source/jormungandr/jormungandr/interfaces/v1/serializer/pt.py
+++ b/source/jormungandr/jormungandr/interfaces/v1/serializer/pt.py
@@ -84,6 +84,12 @@ class SeveritySerializer(PbNestedSerializer):
     priority = jsonschema.Field(schema_type=int)
 
 
+class DisruptionPropertySerializer(PbNestedSerializer):
+    key = jsonschema.Field(schema_type=str)
+    _type = jsonschema.Field(attr='type', label='type', schema_type=str)
+    value = jsonschema.Field(schema_type=str)
+
+
 class PtObjectSerializer(GenericSerializer):
     quality = jsonschema.Field(schema_type=int, required=False, display_none=True, deprecated=True)
     stop_area = jsonschema.MethodField(schema_type=lambda: StopAreaSerializer())
@@ -258,6 +264,7 @@ class DisruptionSerializer(PbNestedSerializer):
     uri = jsonschema.Field(schema_type=str, attr='uri')
     disruption_uri = jsonschema.Field(schema_type=str)
     contributor = jsonschema.Field(schema_type=str, display_none=True)
+    properties = DisruptionPropertySerializer(many=True, display_none=False)
 
 
 class PoiTypeSerializer(GenericSerializer):
--- a/source/kraken/fill_disruption_from_database.cpp
+++ b/source/kraken/fill_disruption_from_database.cpp
@@ -174,25 +174,29 @@ void fill_disruption_from_database(const std::string& connection_string,
     pqxx::result count;
     int impact_count = 0;
     std::string request = (boost::format(
-                               "SELECT DISTINCT ON (d.id) d.id, count(i)"
-                               "   FROM disruption d"
-                               "   LEFT JOIN impact i ON d.id = i.disruption_id"
-                               "   JOIN contributor AS co ON d.contributor_id = co.id"
-                               "   WHERE"
-                               "     (NOT (d.start_publication_date >= '%s' OR d.end_publication_date <= '%s')"
-                               "     OR (d.start_publication_date<='%s' and d.end_publication_date IS NULL))"
-                               "     AND co.contributor_code = ANY('{%s}')" // it's like a "IN" but won't crash if empty"
-                               "     AND d.status = 'published'"
-                               "     AND i.status = 'published'"
-                               "   GROUP BY d.id"
-                               " ;") %meta.production_date.end() % meta.production_date.begin() %meta.production_date.end()  % contributors_array
-                           ).str();
+       "SELECT DISTINCT ON (d.id) d.id, count(i)"
+       "   FROM disruption d"
+       "   LEFT JOIN impact i ON d.id = i.disruption_id"
+       "   JOIN contributor AS co ON d.contributor_id = co.id"
+       "   WHERE"
+       "     (NOT (d.start_publication_date >= '%s' OR d.end_publication_date <= '%s')"
+       "     OR (d.start_publication_date<='%s' and d.end_publication_date IS NULL))"
+       "     AND co.contributor_code = ANY('{%s}')" // it's like a "IN" but won't crash if empty"
+       "     AND d.status = 'published'"
+       "     AND i.status = 'published'"
+       "   GROUP BY d.id"
+       " ;") % meta.production_date.end() % meta.production_date.begin()
+             % meta.production_date.end() % contributors_array
+   ).str();
     count = work.exec(request);
     for (auto ct: count) {
         impact_count += ct[1].as<int>();
     }
 
-    LOG4CPLUS_INFO(log4cplus::Logger::getInstance("Logger"), "Loading " << count.size() << " disruption(s) with " << impact_count << " impact(s)");
+    LOG4CPLUS_INFO(
+        log4cplus::Logger::getInstance("Logger"),
+        "Loading " << count.size() << " disruption(s) with " << impact_count << " impact(s)"
+    );
     reader.finalize();
     LOG4CPLUS_INFO(log4cplus::Logger::getInstance("Logger"), count.size() << " disruption(s) loaded");
 }
--- a/source/kraken/fill_disruption_from_database.cpp
+++ b/source/kraken/fill_disruption_from_database.cpp
@@ -169,9 +169,32 @@ void fill_disruption_from_database(const std::string& connection_string,
         offset += result.size();
         LOG4CPLUS_TRACE(log4cplus::Logger::getInstance("sql"), request);
     }while(result.size() > 0);
-    LOG4CPLUS_INFO(log4cplus::Logger::getInstance("Logger"),"Loading " << offset << " disruptions");
+
+    // counting disruptions & impacts in order to get real numbers
+    pqxx::result count;
+    int impact_count = 0;
+    std::string request = (boost::format(
+                               "SELECT DISTINCT ON (d.id) d.id, count(i)"
+                               "   FROM disruption d"
+                               "   LEFT JOIN impact i ON d.id = i.disruption_id"
+                               "   JOIN contributor AS co ON d.contributor_id = co.id"
+                               "   WHERE"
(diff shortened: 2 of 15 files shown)


buildbotctp:

buildbotctp:

azime:
can you add "properties" in "v1/serialzer/api.py", DisruptionSerializer, for serpy

vpassama:
Done !

source/jormungandr/jormungandr/interfaces/v1/serializer/pt.py
```diff
@@ -84,6 +84,12 @@ class SeveritySerializer(PbNestedSerializer):
     priority = jsonschema.Field(schema_type=int)
 
 
+class DisruptionPropertySerializer(PbNestedSerializer):
+    key = jsonschema.Field(schema_type=str)
+    type = jsonschema.Field(schema_type=str)
```
xlqian: it'd be better avoid using key words as variable names, especially this magic serializer is using metaclass which uses `type` to do other magic things...  :)  i would suggest doing `_type = jsonschema.Field(attr='type', schema_type=str)`

source/kraken/make_disruption_from_chaos.cpp
```diff
@@ -36,6 +36,26 @@ namespace bt = boost::posix_time;
...
 
+static void update_property(nt::disruption::Property& property, const chaos::Property& chaos_property) {
+    property.key = chaos_property.key();
+    property.type = chaos_property.type();
+    property.value = chaos_property.value();
+}
+
+static boost::shared_ptr<nt::disruption::Property>
+make_property(const chaos::Property& chaos_property, nt::disruption::DisruptionHolder& holder) {
+    auto& weak_property = holder.properties[chaos_property.key() + chaos_property.type() + chaos_property.value()];
```
xlqian: please make a function to abstract the key's construction (`make_property_key(...)`)

xlqian:
Thank you for the contribution!

kinnou02:
Ok to test.

source/kraken/tests/disruption_reader_test.cpp
```diff
@@ -599,3 +634,57 @@ BOOST_AUTO_TEST_CASE(two_ptobjects) {
...
+    const_it.set_cause("1", "wording", "11", "22");
+    const_it.set_tag("1", "name", "11", "22");
+    const_it.set_impact("1", "11", "22");
+    const_it.set_severity("2", "wording", "22", "33",
+            "blocking", "color", "2");
+    const_it.set_application_period("0", "1", "2");
+    const_it.set_ptobject("id", "uri", "line", "1", "2");
+    const_it.set_property("foo", "bar", "31");
+    reader(const_it);
+    const_it.set_property("key", "type", "42");
```
TeXitoi: Using a concatenation of the 3 fields as key doesn't work:

```c++
const_it.set_property("keyty", "pe4", "2");// KABOOM
```

TeXitoi (changes requested):

kinnou02:
Jenkins, test this please.

source/kraken/tests/disruption_reader_test.cpp
vpassama: hmmm, I'm sad I didn't see this earlier. I have to use the property ID I think. If I do this, I'll have to make some changes in chaos-proto then.

TeXitoi: Can't you just put the object directly in the disruption?

vpassama: What about the uniqueness check then ? If the SQL request changes and makes properties appear multiple times isn't it risky ?

source/jormungandr/jormungandr/interfaces/v1/serializer/pt.py
vpassama: Done.

source/kraken/make_disruption_from_chaos.cpp
vpassama: Done.

source/kraken/tests/disruption_reader_test.cpp
TeXitoi: you can have something like a `std::set<Property>` in the `Disruption`.

vpassama: I rewrited everything without using holder and I replaced the Disruption properties with a set. Tell me if it's ok !

source/kraken/make_disruption_from_chaos.cpp
```diff
@@ -36,6 +36,20 @@ namespace bt = boost::posix_time;
...
 
+static void update_property(nt::disruption::Property& property, const chaos::Property& chaos_property) {
+    property.key = chaos_property.key();
+    property.type = chaos_property.type();
+    property.value = chaos_property.value();
+}
+
+static nt::disruption::Property
+make_property(const chaos::Property& chaos_property) {
+    nt::disruption::Property property;;
```
TeXitoi: double `;;`

source/kraken/tests/disruption_reader_test.cpp
```diff
@@ -93,7 +93,10 @@ struct Const_it {
             {"ptobject_created_at", Value()},
             {"ptobject_updated_at", Value()},
             {"ptobject_uri", Value()},
-            {"ptobject_type", Value()}
+            {"ptobject_type", Value()},
+            {"property_key", Value()},
+            {"property_type", Value()},
```
TeXitoi, line 98: `property_type` 2 times?

source/type/message.h
```diff
@@ -129,6 +129,21 @@ inline std::string to_string(ChannelType ct) {
     }
 }
 
+struct Property {
+    std::string key;
+    std::string type;
+    std::string value;
+
+    bool operator< (const Property &right) const {
+        return key < right.key || type < right.type || value < right.value;
```
TeXitoi: Doesn't work:

```
{A, B, C} < {B, A, C} -> true
{B, A, C} < {A, B, C} -> true
```

Lexicographical order works:
```c++
if (key != right.key) { return key < right.key; }
if (type != right.type) { return type < right.type; }
return value < right.value;
```

TeXitoi (changes requested):

source/kraken/tests/disruption_reader_test.cpp
vpassama: Are you sure ? I think you read property_type but it's pt_object_type because of the comma.

source/kraken/make_disruption_from_chaos.cpp
vpassama: I'll fix it.

source/type/message.h
vpassama: Ok my bad :)

source/kraken/make_disruption_from_chaos.cpp
```diff
@@ -36,6 +36,20 @@ namespace bt = boost::posix_time;
 
 namespace navitia {
 
+static void update_property(nt::disruption::Property& property, const chaos::Property& chaos_property) {
```
pbougue: Not sure it's worth it (requires a new PR on chaos-proto) : chaos::`Disruption`Property would be clearer (property is way too generic).

source/kraken/fill_disruption_from_database.h
```diff
@@ -66,11 +66,13 @@ namespace navitia {
         chaos::Message* message = nullptr;
         chaos::Channel* channel = nullptr;
         chaos::PtObject* pt_object = nullptr;
+        chaos::Property* property = nullptr;
```
pbougue: minor: `disruption_`property
> OK, it's obvious here, my mistake

source/kraken/fill_disruption_from_database.h
```diff
@@ -66,11 +66,13 @@ namespace navitia {
...
 
         std::string last_period_id = "",
                     last_channel_type_id = "";
 
         std::set<std::string> message_ids;
+        std::set<std::tuple<std::string, std::string, std::string>> properties;
```
pbougue, line 75: minor: `disruption_`properties
> OK, it's obvious here, my mistake

source/kraken/fill_disruption_from_database.cpp
```diff
@@ -164,9 +169,32 @@ void fill_disruption_from_database(const std::string& connection_string,
...
+                               "     AND i.status = 'published'"
+                               "   GROUP BY d.id"
+                               " ;") %meta.production_date.end() % meta.production_date.begin() %meta.production_date.end()  % contributors_array
+                           ).str();
+    count = work.exec(request);
+    for (auto ct: count) {
+        impact_count += ct[1].as<int>();
+    }
+
+    LOG4CPLUS_INFO(log4cplus::Logger::getInstance("Logger"), "Loading " << count.size() << " disruption(s) with " << impact_count << " impact(s)");
```
pbougue: minor: line length (max 113), above also

source/type/message.h
```diff
@@ -129,6 +129,21 @@ inline std::string to_string(ChannelType ct) {
     }
 }
 
+struct Property {
```
pbougue, line 132: I'd rather have `Disruption`Property

pbougue (commented):
Some naming (optional) comments, but looks good :)

source/kraken/tests/disruption_reader_test.cpp
TeXitoi: Yeah, I must have missread.

TeXitoi (approved):

pthegner:
Issue link : CanalTP/navitia#2261

source/type/message.h
vpassama: As discussed yesterday, we won't change this one.

source/kraken/make_disruption_from_chaos.cpp
vpassama: PR incoming

source/kraken/fill_disruption_from_database.cpp
vpassama: Will be fixed

pbougue (approved):

PR #2258, opened 2017-11-29
Ed: add parameter to fusio and gtfs2ed to define simplify tolerance

eturck:
Since at Tisséo we have only one kraken with a very limited dataset compare to what you have we would like to keep the full shape in the journey response (even if it is more expensive memory-wise). We have a dedicated person drawing lines so they can be quite precise on roundabouts or curved roads.

We talked about it with @kinnou02 and he was ok to add another parameters to the different importers. I added --simplify_tolerance (the term seemed to be used on multiple GIS forums).
I have added it to fusio2ed and gtfs2ed. I added an attribute to the data structure instead of passing the parameter to complete and other functions let me know if it was not the right way.

I have also noticed that for NTFS files, the shapes references for lines or routes are not simplified. We call boost::geometry::simplify only when building shape_from_prev. I don't know if it is by design or not, so I just left everything as it is.

### DIFF
--- a/source/ed/data.cpp
+++ b/source/ed/data.cpp
@@ -488,7 +488,8 @@ get_smallest_range(const LineStringIterPair& p1, const LineStringIterPair& p2) {
 nt::LineString
 create_shape(const nt::GeographicalCoord& from,
              const nt::GeographicalCoord& to,
-             const nt::LineString& shape)
+             const nt::LineString& shape,
+             const double simplify_tolerance)
 {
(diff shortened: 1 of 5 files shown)


buildbotctp:

kinnou02:
Ok to test.

kinnou02 (approved):

TeXitoi (approved):

xlqian (approved):

PR #2260, opened 2017-12-04
Doc: context

pbougue:
also refactors, and switch to timezone UTC on some endpoints' context

### DIFF
--- a/source/jormungandr/jormungandr/utils.py
+++ b/source/jormungandr/jormungandr/utils.py
@@ -491,8 +491,6 @@ def get_timezone_str(default='Africa/Abidjan'):
 
 
 def get_current_datetime_str(is_utc=False):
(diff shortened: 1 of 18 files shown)


source/jormungandr/jormungandr/utils.py
```diff
@@ -490,7 +490,9 @@ def get_timezone_str(default='Africa/Abidjan'):
         return timezone.zone if timezone else default
 
 
-def get_current_datetime_str():
+def get_current_datetime_str(is_utc=False):
     timezone = get_timezone_str()
+    if is_utc:
+        timezone = 'Africa/Abidjan'
```
azime: minor:
```python
timezone = 'Africa/Abidjan' if is_utc else get_timezone_str()
```

TeXitoi (approved):

PR #2262, opened 2017-12-05
Jormun: Migration /coord serpy

azime:


pbougue (approved):

source/jormungandr/jormungandr/modules/v1_routing/resources/Index.py
```diff
@@ -117,3 +131,6 @@ def get(self):
             response['regions'].append(resp_dict)
 
         return response
+
+    def options(self, **kwargs):
```
pbougue: I vote against swagger on that endpoint.
Also, we should add "beta-endpoint" on that to protect the use (probably also on coverage status).

pbougue (approved):

PR #2264, opened 2017-12-06
Doc: auto-check before releasing

pbougue:


script_release.py
```diff
@@ -26,17 +27,32 @@ def __init__(self, release_type, remote_name="canalTP"):
...
 
         #and we update dev and release branches
-        print "rebase dev and release"
+        print "rebasing dev and release..."
 
         #TODO quit on error
         self.git.rebase(remote_name + "/dev", "dev")
+        try:
+            self.git.checkout("release")
+        except:
```
xlqian:
```python
except Exception as e:
    print('Cannot checkout "release":{}, we're going to create one'.format(str(e)))
```

xlqian (approved):
:+1:

script_release.py
```diff
@@ -1,15 +1,19 @@
+
 # -*- coding: utf-8 -*-
+import os
+os.environ['LC_ALL'] = 'en_US'
+os.environ['GIT_PYTHON_TRACE'] = '1' #can be 0 (no trace), 1 (git commands) or full (git commands + git output)
```
xlqian, line 5: :+1:

PR #2265, opened 2017-12-07
Ed & Kraken: collections and stop_schedules ordering

TeXitoi:
https://jira.canaltp.fr/browse/NAVP-701

Order is almost as asked, except case insensitivity.

The order of the stop_schedules is changed after an online grooming to line > route > stop point

TODO:
  - [x] a stop_schedules test with line 13 and line 5 with 2 routes each, 2 stop areas with 2 stop points each
 - [x] update utils commit after https://github.com/CanalTP/utils/pull/52 is merged

### DIFF
--- a/source/time_tables/departure_boards.cpp
+++ b/source/time_tables/departure_boards.cpp
@@ -44,6 +44,8 @@ namespace pt = boost::posix_time;
 
 namespace navitia { namespace timetables {
 
+typedef std::pair<routing::RouteIdx, routing::SpIdx> RoutePointIdx;
+
 static bool is_last_stoptime(const nt::StopTime* stop_time, const nt::StopPoint* stp){
     return stop_time->vehicle_journey
             && ! stop_time->vehicle_journey->stop_time_list.empty()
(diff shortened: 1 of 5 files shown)


xlqian:
https://jira.canaltp.fr/browse/NAVP-701

source/type/type.cpp
```diff
@@ -924,11 +942,10 @@ std::string Line::get_label() const {
     return s.str();
 }
 
-bool Route::operator<(const Route & other) const{
-    if(this->uri != other.uri){
-        return this->uri < other.uri;
-    }
-    return this < &other;
+bool Route::operator<(const Route& other) const {
```
xlqian, line 946: :+1:

xlqian (approved):

source/time_tables/departure_boards.cpp
```diff
@@ -44,6 +44,8 @@ namespace pt = boost::posix_time;
 
 namespace navitia { namespace timetables {
 
+typedef std::pair<routing::RouteIdx, routing::SpIdx> RoutePointIdx;
```
xlqian: personally i will do `using RoutePointIdx = std::pair<routing::RouteIdx, routing::SpIdx>`, it's just peanuts....  :sunglasses:

xlqian (approved):

pbougue (approved):

PR #2266, opened 2017-12-08
Doc: Enhance Navitia architecture

stifoon:


pbougue (approved):
Just arrows update?
Would be nice to have the source of this diagram in github also.

stifoon:
Yes, it would be nice, but it needs to rewrite it in draw.io for example. Next PR?

pbougue:
Yes, I thought it was already some drawio.
To be done one day :)

PR #2267, opened 2017-12-11
Jormun: refactoring links on notes and exceptions

azime:


source/jormungandr/jormungandr/interfaces/v1/ResourceUri.py
```diff
@@ -173,31 +173,35 @@ def wrapper(*args, **kwargs):
                 return data
         return wrapper
 
+
 class complete_links(object):
     def __init__(self, resource):
         self.resource = resource
-    def complete(self, data, collect):
+        self.excepted_keys = set(['category', 'id', 'internal', 'rel', 'type'])
```
pbougue: shouldn't it be `expected`_keys?
Also, shouldn't it be static?

source/jormungandr/jormungandr/interfaces/v1/ResourceUri.py
```diff
@@ -173,31 +173,35 @@ def wrapper(*args, **kwargs):
...
-                        type_ = "Add" if elem['except_type'] == 0 else "Remove"
-                        result.append({"id": elem['id'], "date": elem['date'], "type": type_})
-
-                    list(map(elem.pop, del_types))
+                collect = elem.get('type')
+                if collect in result:
+                    link = self.make_and_get_link(elem, collect)
+                    if link.get('id') not in [l.get('id') for l in result[collect]]:
+                        result[collect].append(link)
+                    del_keys = set(elem.keys()).difference(self.excepted_keys)
```
pbougue: A comment would be nice here

source/jormungandr/jormungandr/interfaces/v1/ResourceUri.py
```diff
@@ -173,31 +173,35 @@ def wrapper(*args, **kwargs):
...
+
 class complete_links(object):
     def __init__(self, resource):
         self.resource = resource
-    def complete(self, data, collect):
+        self.excepted_keys = set(['category', 'id', 'internal', 'rel', 'type'])
+
+    def make_and_get_link(self, elem, collect):
+        if collect == "notes":
+            return {"id": elem['id'], "category": elem['category'], "value": elem['value'], "type": collect}
```
pbougue: I'm not sure : are the tests OK if you remove category here and in excepted_keys?
They should be, but if you didn't try it's worth the test (even if we can't let it in the code)

pbougue (approved):

PR #2270, opened 2017-12-14
kraken: Disruption fixes

### LINKED ISSUE #2225: Only NO_SERVICE disruptions on line_sections are linked on /lines requests
Hello!

We are using a lot of disruptions on line sections.
For example it is used for : 
- A deviation, there is a NO_SERVICE impact on a section,
- An elevator out of service, there is an OTHER_EFFECT impact on a section of the subway line on only one stop,
- A stop is temporarily moved a few meter because of road work, there is a STOP_MOVED impact on a line section,
- There is slight delays or minor problems on a section of line, but buses are still passing, we have another OTHER_EFFECT impact.

The problem we have is, when we are requesting /v1/coverage/tisseo/lines, some disruptions are not shown. And that's all the disruptions which are not with a "NO_SERVICE" effect.

We discuss this a bit on the phone with @kinnou02 and @antoine-de a few months ago, but we never made the issue. The problem appeared when you changed a bunch of stuff to be able to properly display disruptions on line_section, which is a good thing, but introduced this :).

I kind of know why it is happening, but don't really know how to properly fix it. 
It's in pb_converter : https://github.com/CanalTP/navitia/blob/dev/source/type/pb_converter.cpp#L620
And there is a pretty obvious comment about it. Here is the relevant part : 
```cpp
if (dump_message_options ==  DumpMessageOptions{DumpMessage::Yes, DumpLineSectionMessage::Yes} ) {
    /*
     * Here we dump the impacts which impact LineSection.
     * We could have link the LineSection impact with the line, but that would change the code and
     * the behavior too much.
     * */
    std::set<boost::shared_ptr<nt::disruption::Impact>> added_impact;
    auto fill_line_section_message = [&](const nt::VehicleJourney& vj) {
        for(const auto& impact: vj.meta_vj->impacted_by) {
            auto impact_ptr = impact.lock();
            if (! impact_ptr)
                continue;
            for (const auto& entity: impact_ptr->informed_entities()) {
                if (boost::get<nt::disruption::LineSection>(&entity) &&
                        added_impact.insert(impact_ptr).second ) {
                    fill_messages(vj.meta_vj, line);
                    return true;
                }
            }
        }
        return true;
    };
    for(const auto* route: l->route_list) {
        route->for_each_vehicle_journey(fill_line_section_message);
    }
}
```

Problem is, we loop on the vehicle_journeys to see if they are impacted, and we add the message to the line if that's the case. But in [apply_disruption](https://github.com/CanalTP/navitia/blob/dev/source/kraken/apply_disruption.cpp#L213) we have, at the start of the method applying a line section disruption : 
```cpp
if (impact->severity->effect != nt::disruption::Effect::NO_SERVICE) {
    LOG4CPLUS_DEBUG(log, "Unhandled action on " << uri);
    this->log_end_action(uri);
    return
}
```

The OTHER_EFFECT disruptions will not impact any vehicle_journey, therefore, we don't have it in /lines.
Without too much thinking I can see two ways to fix this, but I really don't know the side effect they would cause : 
- Linking the line_section disruption to the line, but with @antoine-de's comment I am afraid it is a bad idea :),
- Maybe still impacting the meta_vj but without creating an adapted vj? Would this break something/everything?

Let me know what you think!

TeXitoi:
Several disruption fixes:
 - don't remove stop point with SIGNIFICANT_DELAYS as reported in https://github.com/CanalTP/navitia/issues/2225#issuecomment-345217591
 - output all line sections on /lines API (fix #2225)
 - fix a bug when an impact has 2 line sections

### DIFF
--- a/source/tests/line_sections_test.cpp
+++ b/source/tests/line_sections_test.cpp
@@ -124,6 +124,15 @@ int main(int argc, const char* const argv[]) {
                               .get_disruption(),
                               *b.data->pt_data, *b.data->meta);
 
+    navitia::apply_disruption(b.impact(nt::RTLevel::Adapted, "line_section_on_line_1_other_effect")
+                              .severity(nt::disruption::Effect::OTHER_EFFECT)
+                              .application_periods(btp("20170101T000000"_dt, "20170105T000000"_dt))
+                              .publish(btp("20170101T000000"_dt, "20170110T000000"_dt))
+                              .on_line_section("line:1", "E", "F", {"route:line:1:1", "route:line:1:3"})
+                              .on_line_section("line:1", "F", "E", {"route:line:1:1", "route:line:1:3"})
+                              .get_disruption(),
+                              *b.data->pt_data, *b.data->meta);
+
     mock_kraken kraken(b, "line_sections_test", argc, argv);
 
     return 0;
--- a/source/type/message.cpp
+++ b/source/type/message.cpp
@@ -34,7 +34,7 @@ www.navitia.io
 
 #include <boost/format.hpp>
 #include <boost/algorithm/cxx11/all_of.hpp>
-#include <boost/algorithm/cxx11/one_of.hpp>
+#include <boost/algorithm/cxx11/any_of.hpp>
 
 namespace pt = boost::posix_time;
 namespace bg = boost::gregorian;
@@ -217,7 +217,7 @@ bool Impact::is_only_line_section() const {
 }
 
 bool Impact::is_line_section_of(const Line& line) const {
-    return boost::algorithm::one_of(informed_entities(), [&](const PtObj& entity) {
+    return boost::algorithm::any_of(informed_entities(), [&](const PtObj& entity) {
         const auto* line_section = boost::get<nt::disruption::LineSection>(&entity);
         return line_section && line_section->line && line_section->line->idx == line.idx;
     });
--- a/source/type/pb_converter.cpp
+++ b/source/type/pb_converter.cpp
@@ -620,18 +620,10 @@ void PbCreator::Filler::fill_pb_object(const nt::Line* l, pbnavitia::Line* line)
          * We could have link the LineSection impact with the line, but that would change the code and
          * the behavior too much.
          * */
-        std::set<boost::shared_ptr<nt::disruption::Impact>> added_impact;
         auto fill_line_section_message = [&](const nt::VehicleJourney& vj) {
-            for(const auto& impact: vj.meta_vj->impacted_by) {
-                auto impact_ptr = impact.lock();
-                if (! impact_ptr)
-                    continue;
-                for (const auto& entity: impact_ptr->informed_entities()) {
-                    if (boost::get<nt::disruption::LineSection>(&entity) &&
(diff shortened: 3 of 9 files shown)


TeXitoi:
@eturck can you review this please?

source/type/message.cpp
```diff
@@ -217,7 +217,7 @@ bool Impact::is_only_line_section() const {
 }
 
 bool Impact::is_line_section_of(const Line& line) const {
-    return boost::algorithm::one_of(informed_entities(), [&](const PtObj& entity) {
+    return boost::algorithm::any_of(informed_entities(), [&](const PtObj& entity) {
```
pbougue, line 220: nice catch ⚾️

source/type/pb_converter.cpp
```diff
@@ -620,18 +620,10 @@ void PbCreator::Filler::fill_pb_object(const nt::Line* l, pbnavitia::Line* line)
          * We could have link the LineSection impact with the line, but that would change the code and
          * the behavior too much.
          * */
-        std::set<boost::shared_ptr<nt::disruption::Impact>> added_impact;
         auto fill_line_section_message = [&](const nt::VehicleJourney& vj) {
-            for(const auto& impact: vj.meta_vj->impacted_by) {
```
pbougue, line 625: `impacted_by` should probably be renamed `impacted_and_modified_by` to avoid this kind of mistake

pbougue (approved):

source/tests/line_sections_test.cpp
```diff
@@ -124,6 +124,15 @@ int main(int argc, const char* const argv[]) {
                               .get_disruption(),
                               *b.data->pt_data, *b.data->meta);
 
+    navitia::apply_disruption(b.impact(nt::RTLevel::Adapted, "line_section_on_line_1_other_effect")
+                              .severity(nt::disruption::Effect::OTHER_EFFECT)
+                              .application_periods(btp("20170101T000000"_dt, "20170105T000000"_dt))
+                              .publish(btp("20170101T000000"_dt, "20170110T000000"_dt))
+                              .on_line_section("line:1", "E", "F", {"route:line:1:1", "route:line:1:3"})
+                              .on_line_section("line:1", "F", "E", {"route:line:1:1", "route:line:1:3"})
```
xlqian, line 132: i dont think this will change your test, but just looking at this line, it's weird for me that section "EF" and section "FE" have the same route

eturck (approved):
Great work :+1: 
Tested it with @vpassama on the problem we were having and it's fixed.

source/tests/line_sections_test.cpp
TeXitoi: That's all the routes. In practice, there is no vj that do F then E, but I need that to have 2 line sections on the impact to test the `one_of`.

source/type/pb_converter.cpp
TeXitoi: I've changed `impacted_by` by `modified_by`

source/tests/line_sections_test.cpp
xlqian: ok

PR #2274, opened 2017-12-22
Jormungandr: don't try to create isochrone links if it's disabled

eturck:
With the default settings isochrone is disabled and there is an error when opening the API.


buildbotctp:

pbougue (approved):
Oups!
Nice catch :)

PR #2280, opened 2018-01-04
Jormun: manage Cykleo station with id starting by "0"

pbougue:
Previously, if in OSM we had a station tagged "02" instead of "2"; Navitia couldn't find the station in Cykleo's feed.

JIRA: https://jira.canaltp.fr/browse/NAVITIAII-2330

### DIFF
--- a/source/jormungandr/jormungandr/parking_space_availability/bss/tests/cykleo_test.py
+++ b/source/jormungandr/jormungandr/parking_space_availability/bss/tests/cykleo_test.py
@@ -29,6 +29,7 @@
 # www.navitia.io
 
 from __future__ import absolute_import, print_function, unicode_literals, division
+from copy import deepcopy
 from jormungandr.parking_space_availability.bss.cykleo import CykleoProvider
 from jormungandr.parking_space_availability.bss.stands import Stands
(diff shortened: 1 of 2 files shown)


source/jormungandr/jormungandr/parking_space_availability/bss/tests/cykleo_test.py
```diff
@@ -61,6 +72,16 @@ def parking_space_availability_cykleo_support_poi_test():
     poi['properties']['operator'] = 'Bad_operator'
     assert not provider.support_poi(poi)
 
+    # same test with "02" in id, not "2"
+    assert provider.support_poi(poi_with_0)
+    poi_with_0['properties']['operator'] = 'Bad_operator'
```
xlqian: you are modifying a global var which is shared in other tests

prefer deep-copying before you modify it

(you're just lucky, in other tests, you don't test support_poi :D )

pbougue: Yep, just managed it the same way it was done before... I just changed that.

xlqian (approved):

PR #2285, opened 2018-01-08
Kraken: fix no solution found when forbidden_uri is in vj_extention

xlqian:
https://jira.canaltp.fr/browse/NAVITIAII-2288

when I passed a month on the bug and finally fixed it!!! 
jdixrit

### DIFF
--- a/source/routing/raptor.cpp
+++ b/source/routing/raptor.cpp
@@ -87,6 +87,7 @@ bool RAPTOR::apply_vj_extension(const Visitor& v,
             auto sp = st.stop_point;
             const auto sp_idx = SpIdx(*sp);
 
+            if (! valid_stop_points[sp_idx.val]) { continue; }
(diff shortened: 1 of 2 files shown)


TeXitoi (approved):

source/routing/raptor.cpp
```diff
@@ -87,6 +87,7 @@ bool RAPTOR::apply_vj_extension(const Visitor& v,
             auto sp = st.stop_point;
             const auto sp_idx = SpIdx(*sp);
 
+            if (! valid_stop_points[sp_idx.val]) { continue; }
```
pbougue, line 90: It would be nice to test (maybe it's already the case somewhere) that it works fine with pickup/dropoff forbidden

TeXitoi: 10 lines up

PR #2291, opened 2018-01-15
tyr : add API migrate_from_poi_to_osm

mbounabi:
https://jira.canaltp.fr/browse/NAVITIAII-2340
new ticket created afterwards:
https://jira.canaltp.fr/browse/NAVP-756

### DIFF
--- a/source/navitiacommon/navitiacommon/models.py
+++ b/source/navitiacommon/navitiacommon/models.py
@@ -408,11 +408,18 @@ class Instance(db.Model):
             .filter(DataSet.type == 'poi', Job.instance_id == self.id)\
             .all()
 
+        job_list = {}
         for dataset, job in result:
             # Cascade Delete not working so delete Metric associated manually
             db.session.query(Metric).filter(Metric.dataset_id == dataset.id).delete()
             db.session.delete(dataset)
-            if not (len(job.data_sets.all()) - 1):
+
+            if job.id not in job_list:
+                job_list[job.id] = 1
+            else:
+                job_list[job.id] += 1
+
+            if not (len(job.data_sets.all()) - job_list[job.id]):
                 db.session.delete(job)
 
         db.session.commit()
--- a/source/navitiacommon/navitiacommon/models.py
+++ b/source/navitiacommon/navitiacommon/models.py
@@ -273,10 +273,10 @@ class Instance(db.Model):
     # ============================================================
     # params for jormungandr
     # ============================================================
-    #the scenario used by jormungandr, by default we use the new default scenario (and not the default one...)
+    # the scenario used by jormungandr, by default we use the new default scenario (and not the default one...)
     scenario = db.Column(db.Text, nullable=False, default='new_default')
 
-    #order of the journey, this order is for clockwise request, else it is reversed
+    # order of the journey, this order is for clockwise request, else it is reversed
     journey_order = db.Column(db.Enum('arrival_time', 'departure_time', name='journey_order'),
                               default=default_values.journey_order, nullable=False)
 
@@ -401,6 +401,15 @@ class Instance(db.Model):
         else:
             raise Exception({'error': 'instance is required'}, 400)
 
+    def delete_dataset(self):
+        # db.session.query(Job).filter(Job.instance_id == self.id).delete()
+
+        jobs = db.session.query(Job).filter(Job.instance_id == self.id).all()
+        for j in jobs:
+            db.session.query(Metric).filter(Metric.job_id == j.id).delete()
+            db.session.query(DataSet).filter(DataSet.job_id == j.id).delete()
+            db.session.commit()
+
     def __repr__(self):
         return '<Instance %r>' % self.name
 
@@ -498,7 +507,7 @@ class Job(db.Model, TimestampMixin):
                             db.ForeignKey('instance.id'))
     autocomplete_params_id = db.Column(db.Integer,
                                        db.ForeignKey('autocomplete_parameter.id'))
-    #name is used for the ENUM name in postgreSQL
+    # name is used for the ENUM name in postgreSQL
     state = db.Column(db.Enum('pending', 'running', 'done', 'failed',
                               name='job_state'))
 
@@ -542,7 +551,7 @@ class DataSet(db.Model):
 
     uid = db.Column(UUID, unique=True)
 
-    job_id = db.Column(db.Integer, db.ForeignKey('job.id'))
+    job_id = db.Column(db.Integer, db.ForeignKey('job.id', ondelete='CASCADE'))
 
     def __init__(self):
         self.uid = str(uuid.uuid4())
@@ -550,7 +559,7 @@ class DataSet(db.Model):
     @classmethod
     def find_by_uid(cls, uid):
         if not uid:
-            #old dataset don't have uid, we don't want to get one of them
+            # old dataset don't have uid, we don't want to get one of them
             return None
         return cls.query.filter_by(uid=uid).first()
 
--- a/source/navitiacommon/navitiacommon/models.py
+++ b/source/navitiacommon/navitiacommon/models.py
@@ -402,12 +402,9 @@ class Instance(db.Model):
             raise Exception({'error': 'instance is required'}, 400)
 
     def delete_dataset(self):
-        # db.session.query(Job).filter(Job.instance_id == self.id).delete()
-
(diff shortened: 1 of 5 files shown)


source/tyr/tyr/api.py
```diff
@@ -78,6 +78,8 @@
 api.add_resource(resources.AutocompleteUpdateData,
                  '/v0/autocomplete_parameters/<ac_instance_name>/update_data')
 
+api.add_resource(resources.MigrateFromPoiToOsm, '/v0/instances/<string:instance_name>/:actions/migrate_from_poi_to_osm')
```
pbougue: consistency: In tartare URL is `/actions/<my_action>` https://github.com/CanalTP/tartare/blob/master/documentation/ressources/open_api.yaml#L182

source/tyr/tyr/resources.py
```diff
@@ -1081,3 +1081,18 @@ def post(self, ac_instance_name):
         _, job = import_autocomplete([filename], instance)
         job = models.db.session.merge(job) #reatache the object
         return marshal({'job': job}, one_job_fields), 200
+
+
+class MigrateFromPoiToOsm(flask_restful.Resource):
+    def delete(self, instance_name):
```
pbougue, line 1087: minor consistency: in Tartare, POST method is used https://github.com/CanalTP/tartare/blob/master/documentation/ressources/open_api.yaml#L183
But @kinnou02 is right, it's a delete, so...

source/navitiacommon/navitiacommon/models.py
```diff
@@ -542,15 +548,15 @@ class DataSet(db.Model):
 
     uid = db.Column(UUID, unique=True)
 
-    job_id = db.Column(db.Integer, db.ForeignKey('job.id'))
+    job_id = db.Column(db.Integer, db.ForeignKey('job.id', ondelete='CASCADE'))
```
kinnou02: This doesn't require a migration?

source/tyr/tyr/resources.py
```diff
@@ -1081,3 +1081,18 @@ def post(self, ac_instance_name):
...
+
+        instance = models.Instance.get_by_name(instance_name)
+        if instance:
+            instance.delete_dataset()
+            return_msg = "All datasets deleted for instance "+instance_name
```
kinnou02: Don't concatenate string, always use format like this:
```
"hello {}".format("world")
```
except with logger you should use the logging substitution that use `%s`

source/navitiacommon/navitiacommon/models.py
```diff
@@ -401,6 +401,12 @@ def get_from_id_or_name(cls, id=None, name=None):
         else:
             raise Exception({'error': 'instance is required'}, 400)
 
+    def delete_dataset(self):
+        jobs = db.session.query(Job).filter(Job.instance_id == self.id).all()
+        for j in jobs:
+            db.session.delete(j)
+            db.session.commit()
```
xlqian:
```python
for j in jobs:
    db.session.delete(j)
db.session.commit()
```

xlqian: you can probably do this with less code
Job.delete().where(Job.instance_id == self.id)
not sure, to be tested

source/tyr/tyr/api.py
mbounabi: changed to '/v0/instances/<string:instance_name>/actions/migrate_from_poi_to_osm'

source/navitiacommon/navitiacommon/models.py
mbounabi: No upgrade after migration. Now removed.

source/tyr/tyr/resources.py
mbounabi: Done

source/navitiacommon/navitiacommon/models.py
mbounabi: Done

TeXitoi (approved):

source/navitiacommon/navitiacommon/models.py
```diff
@@ -401,6 +401,19 @@ def get_from_id_or_name(cls, id=None, name=None):
...
             raise Exception({'error': 'instance is required'}, 400)
 
+    def delete_dataset(self):
+        jobs = db.session.query(Job).filter(Job.instance_id == self.id).all()
+        for j in jobs:
+            for dataset in j.data_sets.all():
+                if dataset.type == 'poi':
+                    db.session.query(Metric).filter(Metric.dataset_id == dataset.id).delete()
+                    db.session.delete(dataset)
+            db.session.commit()
```
kinnou02: you don't have to commit after every query, you can (and must) commit only at the end

source/navitiacommon/navitiacommon/models.py
```diff
@@ -401,6 +401,19 @@ def get_from_id_or_name(cls, id=None, name=None):
         else:
             raise Exception({'error': 'instance is required'}, 400)
 
+    def delete_dataset(self):
+        jobs = db.session.query(Job).filter(Job.instance_id == self.id).all()
+        for j in jobs:
+            for dataset in j.data_sets.all():
```
kinnou02: this isn't very efficient, you will get every job then iterate over all the datasets of all these job.
You could do a query to only get the dataset of type POI, then delete

source/navitiacommon/navitiacommon/models.py
```diff
@@ -401,6 +401,19 @@ def get_from_id_or_name(cls, id=None, name=None):
...
+                if dataset.type == 'poi':
+                    db.session.query(Metric).filter(Metric.dataset_id == dataset.id).delete()
+                    db.session.delete(dataset)
+            db.session.commit()
+            # if no more datasets related to job, delete it
+            if not j.data_sets.all():
```
kinnou02: does it re trigger a query every time?

kinnou02 (changes requested):

source/navitiacommon/navitiacommon/models.py
mbounabi: the new API specifies one instance in the URL. By querying all POI datasets, those from regions not specified in the query will be deleted too

kinnou02: and you can do `AND` in the query

mbounabi: Changed

source/navitiacommon/navitiacommon/models.py
mbounabi: Changed

source/navitiacommon/navitiacommon/models.py
mbounabi: according to the doc:
all()
    Return the results represented by this Query as a list.
    This results in an execution of the underlying query.

kinnou02:
BTW it' better to create the branch on your fork, so we don't have a lot of branches on this repo.
it's not bad for this one.

source/tyr/tests/integration/migrate_poi_to_osm_test.py
```diff
@@ -0,0 +1,130 @@
+from tests.check_utils import api_get, api_delete
```
xlqian, line 30: license

xlqian (approved):

source/navitiacommon/navitiacommon/models.py
```diff
@@ -401,6 +401,29 @@ def get_from_id_or_name(cls, id=None, name=None):
         else:
             raise Exception({'error': 'instance is required'}, 400)
 
+    def delete_dataset(self):
+        result = db.session.query(DataSet, Job) \
+            .join(Job) \
+            .join(Instance) \
```
kinnou02: do you really need to do this join? you don't use any fields from instances

mbounabi: 'self.id' in the filter. Need testing to check the results

source/tyr/tests/integration/migrate_poi_to_osm_test.py
mbounabi: Done. Only 1 file in the directory contains license header. Should be added in every files if needed

source/navitiacommon/navitiacommon/models.py
mbounabi: Results are the same. Removed

kinnou02 (approved):

PR #2293, opened 2018-01-16
Jormun: Generate ridesharing crowfly in new_default

pbougue:
Hand-tested locally using playground (and works so far ;) )

### DIFF
--- a/source/jormungandr/jormungandr/scenarios/new_default.py
+++ b/source/jormungandr/jormungandr/scenarios/new_default.py
@@ -34,6 +34,7 @@ import logging
 from flask.ext.restful import abort
 from flask import g
 from jormungandr.scenarios import simple, journey_filter, helpers
+from jormungandr.scenarios.ridesharing.ridesharing_helper import build_ridesharing_crowfly_journey
 from jormungandr.scenarios.utils import journey_sorter, change_ids, updated_request_with_default, \
     get_or_default, fill_uris, gen_all_combin, get_pseudo_duration, mode_weight
 from navitiacommon import type_pb2, response_pb2, request_pb2
@@ -42,7 +43,7 @@ from jormungandr.scenarios.qualifier import min_from_criteria, arrival_crit, dep
     has_no_bike, has_bike, has_no_bss, has_bss, non_pt_journey, has_walk, and_filters
 import numpy as np
 import collections
-from jormungandr.utils import date_to_timestamp
+from jormungandr.utils import date_to_timestamp, PeriodExtremity
 from jormungandr.scenarios.simple import get_pb_data_freshness
 import gevent, gevent.pool
 import flask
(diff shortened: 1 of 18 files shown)


source/jormungandr/jormungandr/scenarios/ridesharing/ridesharing_helper.py
```diff
@@ -0,0 +1,75 @@
...
+    start.lon = orig_coord.lon
+    start.lat = orig_coord.lat
+    end = ridesharing_section.shape.add()
+    end.lon = dest_coord.lon
+    end.lat = dest_coord.lat
+
+    distance = crowfly_distance_between(orig_coord, dest_coord)
+    ridesharing_section.length = int(distance)
+    # manhattan + 15min
+    ridesharing_section.duration = int(distance / (instance.car_speed / math.sqrt(2))) + 15*60
```
kinnou02: We willl probably need some kind of configuration here later, but in another PR.

source/jormungandr/jormungandr/scenarios/new_default.py
```diff
@@ -752,6 +753,11 @@ def fill_journeys(self, request_type, api_request, instance):
         min_asked_journeys = get_or_default(request, 'min_nb_journeys', 1)
         min_journeys_calls = get_or_default(request, '_min_journeys_calls', 1)
 
+        # building ridesharing request from "original" request
+        period_extremity = PeriodExtremity(request['datetime'], request['clockwise'])
```
kinnou02: Only do this if ridesharing is enabled for this request.

source/type/type.cpp
```diff
@@ -752,6 +752,9 @@ std::string static_data::captionByType(Type_e type){
 }
 
 Mode_e static_data::modeByCaption(const std::string & mode_str) {
+    if (mode_str == "ridesharing") {
```
pbougue: It might be smarter to test that it's one of `['walking', 'car', 'bss', 'bike']`
Any opinion @xlqian @kinnou02 @TeXitoi ?

source/jormungandr/tests/routing_tests_new_default.py
```diff
@@ -99,6 +99,57 @@ def test_first_walking_last_walking_section_mode(self):
...
+                "first_section_mode[]={first}&last_section_mode[]={last}"\
+                .format(first='ridesharing', last='walking')
+        response = self.query_region(query)
+        check_best(response)
+        self.is_valid_journey_response(response, query)
+        assert len(response["journeys"]) == 3
+        assert response["journeys"][0]["type"] == "best"
+        assert "walking" in response["journeys"][0]["tags"]
+        rs_journey = response["journeys"][2]
+        assert "ridesharing" in rs_journey["tags"]
```
pbougue, line 113: Some of these tests ought to be mutualized when managing ridesharing fallback... Now is too soon I guess

source/type/type.cpp
kinnou02: It is, we won't add new modes in kraken

kinnou02 (approved):

PR #2302, opened 2018-01-19
Kraken and Jormungandr : Add utc arrival/departure time field

benoit-bst:
1. Add 2 fields in StopTime
- utc_arrival_time
- utc_departure_time

2. Update navitia-proto

3. Add new test for vehicle_journeys in
schema_tests.py (The test did not exist).

### DIFF
--- a/source/jormungandr/jormungandr/interfaces/v1/serializer/pt.py
+++ b/source/jormungandr/jormungandr/interfaces/v1/serializer/pt.py
@@ -174,7 +174,7 @@ class CalendarExceptionSerializer(PbNestedSerializer):
 
 
 class CalendarSerializer(PbNestedSerializer):
-    id = jsonschema.MethodField(schema_type=str, display_none=True, description='Identifier of the object')
+    id = jsonschema.MethodField(schema_type=str, display_none=False, description='Identifier of the object')
     def get_id(self, obj):
         if obj.HasField(str('uri')) and obj.uri:
             return obj.uri
@@ -440,7 +440,9 @@ class JourneyPatternSerializer(GenericSerializer):
 
 class StopTimeSerializer(PbNestedSerializer):
     arrival_time = LocalTimeField()
+    utc_arrival_time = LocalTimeField()
     departure_time = LocalTimeField()
+    utc_departure_time = LocalTimeField()
     headsign = jsonschema.Field(schema_type=str)
     journey_pattern_point = JourneyPatternPointSerializer()
     stop_point = StopPointSerializer()
@@ -451,7 +453,7 @@ class VehicleJourneySerializer(GenericSerializer):
     stop_times = StopTimeSerializer(many=True)
     comments = CommentSerializer(many=True, display_none=False)
     comment = FirstCommentField(attr='comments', display_none=False)
-    codes = CodeSerializer(many=True)
+    codes = CodeSerializer(many=True, required=False)
     validity_pattern = ValidityPatternSerializer()
     calendars = CalendarSerializer(many=True)
     trip = TripSerializer()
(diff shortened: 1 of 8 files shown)


source/jormungandr/jormungandr/interfaces/v1/serializer/pt.py
```diff
@@ -440,7 +440,9 @@ class JourneyPatternSerializer(GenericSerializer):
 
 class StopTimeSerializer(PbNestedSerializer):
     arrival_time = LocalTimeField()
+    utc_arrival_time = LocalTimeField()
```
xlqian: `utc_` has a `LocalTimeFiled()` ? :D you may want to give that Field a more generic name

source/jormungandr/jormungandr/interfaces/v1/serializer/pt.py
```diff
@@ -451,7 +453,7 @@ class VehicleJourneySerializer(GenericSerializer):
     stop_times = StopTimeSerializer(many=True)
     comments = CommentSerializer(many=True, display_none=False)
     comment = FirstCommentField(attr='comments', display_none=False)
-    codes = CodeSerializer(many=True)
+    codes = CodeSerializer(many=True, required=False)
```
xlqian, line 456: why `codes` are involved?

source/jormungandr/tests/schema_tests.py
```diff
@@ -196,6 +196,9 @@ def test_journeys_schema(self):
                            'to=poi:station_1&'
                            'datetime=20120615T080000')
 
+    def test_vehicle_journeys(self):
```
xlqian, line 199: :+1:

source/jormungandr/jormungandr/interfaces/v1/serializer/pt.py
benoit-bst: Because of the schema check. We don't send "codes" for a vehicle journeys... The new tests highlight the problem.

source/jormungandr/jormungandr/interfaces/v1/serializer/pt.py
benoit-bst: It's true, we can change the name of the function "LocalTimeField" in "TimeField", because we use for other fields like (utc, base arrival, amended arrival, etc...).

source/jormungandr/jormungandr/interfaces/v1/serializer/pt.py
xlqian: ok! thx

kadhikari:
Could you add atleast one test around https://github.com/CanalTP/navitia/blob/dev/source/jormungandr/tests/ptref_tests.py#L89 so that we can compare values of "arrival_time/departure_time" and "utc_arrival_time/utc_depature_time"

benoit-bst:
@kadhikari, you have right, a simple test to check the utc_arrivel/departure_time value is a good thing. 
I add it.

kinnou02:
conflicts :(

source/jormungandr/jormungandr/interfaces/v1/serializer/pt.py
```diff
@@ -174,7 +174,7 @@ class CalendarExceptionSerializer(PbNestedSerializer):
 
 
 class CalendarSerializer(PbNestedSerializer):
-    id = jsonschema.MethodField(schema_type=str, display_none=True, description='Identifier of the object')
+    id = jsonschema.MethodField(schema_type=str, display_none=False, description='Identifier of the object')
```
pbougue, line 177: we may have to check what marshal does but it may be a bit dangerous for integrators, no?
(well integrators that use calendars 😄 )
If tests were breaking because of that "True" then it's OK, integrators were at risk before (well... if they ever used /v_j).

pbougue (approved):

kinnou02 (approved):

xlqian (approved):

mbounabi (approved):

pbougue:
Jenkins, test this please

PR #2303, opened 2018-01-19
Kraken: Add ridesharing into kraken

kinnou02:
This PR add a new fallback mode "car_no_park" into kraken, it use the
Car graph but doesn't require the passenger to park his car into a
parking.
This mode is used to compute an ETA for ridesharing journeys.


source/kraken/worker.cpp
```diff
@@ -912,6 +918,11 @@ type::EntryPoint make_sn_entry_point(const std::string& place,
             entry_point.streetnetwork_params.offset = data.geo_ref->offsets[mode_enum];
             entry_point.streetnetwork_params.speed_factor = speed / georef::default_speed[mode_enum];
             break;
+      /*  case type::Mode_e::Ridesharing:
+            entry_point.streetnetwork_params.offset = data.geo_ref->offsets[mode_enum];
+            entry_point.streetnetwork_params.speed_factor = speed / georef::default_speed[type::Mode_e::Car];
+            break;
+            */
```
xlqian: obsolete? TODO?

kinnou02:
So `car_no_park` in kraken and `ridesharing` in the interface with jormun doing the convertion between the two?

source/jormungandr/jormungandr/scenarios/new_default.py
```diff
@@ -85,11 +85,11 @@ def get_kraken_calls(request):
                            ('bike', 'bss'),
                            ('car', 'bss'),
                            ('bike', 'bike'),
-                           ('ridesharing', 'ridesharing'),
                            ('ridesharing', 'walking'),
                            ('walking', 'ridesharing'),
                            ('ridesharing', 'bss'),
                            ('bss', 'ridesharing')]
+    # We don't want to do ridesharing - ridesharing journeys
```
pbougue, line 92: I added this to allow calls `first=last=[ridesharing]&max_duration_to_pt=0` for direct ridesharing only...
But I agree on the functionnal need of no ridesharing-(maybe PT)-ridesharing.
The 2 "natural" integrations for direct ridesharing only are:
* `first=[ridesharing]&max_duration_to_pt=0` (which would work anyway)
* `first=last=[ridesharing]&max_duration_to_pt=0` (which I think should work, but we probably can think of other workaround, or decide to not care)

source/jormungandr/jormungandr/scenarios/new_default.py
```diff
@@ -801,7 +800,7 @@ def fill_journeys(self, request_type, api_request, instance):
         journey_filter.final_filter_journeys(responses, instance, api_request)
         pb_resp = merge_responses(responses)
 
-        if 'ridesharing' in ridesharing_req['origin_mode'] and instance.ridesharing_services:
+        if 'ridesharing' in ridesharing_req['origin_mode'] and False:#TODO: update with kraken stuff
```
pbougue: You will want to keep checking that instance has a ridesharing_service configured, no?

source/kraken/worker.cpp
```diff
@@ -999,7 +1006,8 @@ void Worker::direct_path(const pbnavitia::Request& request) {
                              origin,
                              destination,
                              {bt::from_time_t(dp_request.datetime())},
-                             dp_request.clockwise());
+                             dp_request.clockwise(),
+                             street_network_worker->direct_path_finder.mode);
```
pbougue: not sure this will be needed if kraken is "ridesharing-blind"

source/type/pb_converter.cpp
```diff
@@ -1644,7 +1645,7 @@ void PbCreator::fill_street_sections(const type::EntryPoint& ori_dest, const geo
     georef::PathItem last_item;
 
     //we create 1 section by mean of transport
-    for (auto item : path.path_items) {
+    for (const auto& item : path.path_items) {
```
pbougue, line 1648: 👍

source/tyr/migrations/versions/3acdde6329db_settings_for_ridesharing.py
```diff
@@ -0,0 +1,24 @@
...
+# revision identifiers, used by Alembic.
+revision = '3acdde6329db'
+down_revision = '4e80856d6e89'
+
+from alembic import op
+import sqlalchemy as sa
+
+
+def upgrade():
+    op.add_column('instance', sa.Column('max_ridesharing_duration_to_pt', sa.Integer(), nullable=False, server_default='1800'))
```
pbougue: Don't you also need to upgrade the possible enum values for first/last_section_mode?
And I had other trailing migrations, probably not super important:
```
def upgrade():
    ### commands auto generated by Alembic - please adjust! ###
    op.alter_column('instance', 'bss_provider',
               existing_type=sa.BOOLEAN(),
               nullable=False,
               existing_server_default='true')
    op.alter_column('poi_type_json', 'poi_types_json',
               existing_type=sa.TEXT(),
               nullable=True)
    op.alter_column('user', 'billing_plan_id',
               existing_type=sa.INTEGER(),
               nullable=False)
    op.alter_column('user', 'end_point_id',
               existing_type=sa.INTEGER(),
               nullable=False)
    ### end Alembic commands ###


def downgrade():
    ### commands auto generated by Alembic - please adjust! ###
    op.alter_column('user', 'end_point_id',
               existing_type=sa.INTEGER(),
               nullable=True)
    op.alter_column('user', 'billing_plan_id',
               existing_type=sa.INTEGER(),
               nullable=True)
    op.alter_column('poi_type_json', 'poi_types_json',
               existing_type=sa.TEXT(),
               nullable=False)
    op.alter_column('instance', 'bss_provider',
               existing_type=sa.BOOLEAN(),
               nullable=True,
               existing_server_default='true')
    ### end Alembic commands ###
```

pbougue (commented):
I vote for `car_no_park`, yes :)

source/tyr/migrations/versions/3acdde6329db_settings_for_ridesharing.py
kinnou02: The fallback mode of the travelers profile are deprecated since at least two years.
As for the missing migration it isn't in the scope of this PR.

source/kraken/worker.cpp
kinnou02: it isn't since everything will have to be done in jormungandr where it's slower :)

source/jormungandr/jormungandr/scenarios/new_default.py
kinnou02: I didn't understand everything here.
I have seen with @mmartinet that we don't ever want to display journeys that use ridesharing at the begin and at the end.

pbougue: Agreed.
So: we forbid that ridesharing-ridesharing call here, and we add/replace walking to last section if first=last=[ridesharing]

source/tyr/migrations/versions/3acdde6329db_settings_for_ridesharing.py
pbougue: Ok, deserves a separate job for missing migrations and maybe check that traveler's fallback are really not plugged (we should probably also do the same job for max_duration(s)_to_pt which is a pain to maintain.

xlqian (approved):

source/type/pb_converter.cpp
```diff
@@ -1896,6 +1897,9 @@ void PbCreator::fill_crowfly_section(const type::EntryPoint& origin, const type:
     case type::Mode_e::Bss:
         section->mutable_street_network()->set_mode(pbnavitia::Bss);
         break;
+    case type::Mode_e::CarNoPark:
+        section->mutable_street_network()->set_mode(pbnavitia::CarNoPark);//TODO: use ridesharing in pb
```
pbougue: I don't understand this TODO

pbougue (approved):
This looks good 🎉

source/type/pb_converter.cpp
kinnou02: It's an old todo, I have removed it

PR #2305, opened 2018-01-23
Jormun: Add status in InstantSystem

xlqian:


source/jormungandr/jormungandr/interfaces/v1/serializer/status.py
```diff
@@ -97,19 +97,26 @@ class CircuitBreakerSerializer(serpy.DictSerializer):
...
 
     def get_timeout(self, obj):
         return obj.get('timeout', None)
 
-    def get_circuit_breaker(self, obj):
-        o = obj.get('circuit_breaker', None)
-        return CircuitBreakerSerializer(o, display_none=False).data if o else None
+
+class RidesharingServicesSerializer(OutsideServiceCommon):
+    service_url = Field(schema_type=str, display_none=False)
```
kinnou02: it's better to not expose that kind of things, sometime we have to deal with services that doesn't any kind of authentication, so we can't publicly show their url.

kinnou02 (changes requested):

source/jormungandr/jormungandr/interfaces/v1/serializer/status.py
xlqian: ok

xlqian: good point

kadhikari (approved):

kinnou02 (approved):

source/jormungandr/jormungandr/scenarios/ridesharing/instant_system.py
```diff
@@ -62,6 +62,15 @@ def __init__(self, instance, service_url, api_key, network, rating_scale_min=Non
         self.breaker = pybreaker.CircuitBreaker(fail_max=app.config['CIRCUIT_BREAKER_MAX_INSTANT_SYSTEM_FAIL'],
                                                 reset_timeout=app.config['CIRCUIT_BREAKER_INSTANT_SYSTEM_TIMEOUT_S'])
 
+    def status(self):
+        return {'id': self.system_id,
+                'class': self.__class__.__name__,
+                'service_url': self.service_url,
```
pbougue: So you don't need it anymore, no?

xlqian: oops

Issue #2304, opened 2018-01-19
(swagger) Coord def seems weird


jbcrestot:
Coord is def with lon and lat as optionnal whereas the Coord property itself is always optionnal.

I think it's unlikely that you have Coord with only lat or lon.

Issue #2298, opened 2018-01-18
(swagger) StopDateTime.additional_informations def


jbcrestot:
I found that StopDateTime have a property additional_informations that is typed as array of string, it would be nice to have an enum

mbounabi:
https://github.com/CanalTP/navitia/pull/2306

Issue #2297, opened 2018-01-18
(swagger) Network def


jbcrestot:
It might have a pb with network definition as property lines is not def as array of line

`lines: {
$ref: “#/definitions/Line”
},`

PR #2306, opened 2018-01-24
Jormungandr : Swagger issues

### LINKED ISSUE #2295: (swagger) Link Def
I found that Links can have a href property that is not referenced

mbounabi:
fix: https://github.com/CanalTP/navitia/issues/2295 https://github.com/CanalTP/navitia/issues/2297 https://github.com/CanalTP/navitia/issues/2298 https://github.com/CanalTP/navitia/issues/2304

### DIFF
--- a/source/jormungandr/jormungandr/interfaces/v1/serializer/fields.py
+++ b/source/jormungandr/jormungandr/interfaces/v1/serializer/fields.py
@@ -100,8 +100,8 @@ class ErrorSerializer(PbNestedSerializer):
 
 
 class CoordSerializer(serpy.Serializer):
-    lon = DoubleToStringField(display_none=True)
-    lat = DoubleToStringField(display_none=True)
+    lon = DoubleToStringField(display_none=True, required=True)
+    lat = DoubleToStringField(display_none=True, required=True)
 
 
 class CodeSerializer(serpy.Serializer):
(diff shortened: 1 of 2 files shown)


source/jormungandr/jormungandr/interfaces/v1/serializer/fields.py
```diff
@@ -100,8 +100,9 @@ class ErrorSerializer(PbNestedSerializer):
 
 
 class CoordSerializer(serpy.Serializer):
-    lon = DoubleToStringField()
-    lat = DoubleToStringField()
+    lon = DoubleToStringField(display_none=True)
```
kinnou02: `required=True` no?

mbounabi: if 'display_none' isn't set, it will be False by default (jsonschema/fields) and 'required' will always be set to False (serpy/fields.py) if 'display_none' is False.

pbougue: yep, but here (and below) it's required=True also (or I didn't understood your response)

pbougue (commented):
* proto will need a rebase (and update this PR once merged)
* it would be great to add tests breaking without this PR, and OK with it (TDD-style!)
* this will need a rebase as it has conflicts (probably submodule 💩 )

source/jormungandr/jormungandr/interfaces/v1/serializer/fields.py
mbounabi: I forgot to mention that 'required' is set to True by default (serpy/fields.py). So, as long as 'display_none' is set to True, 'required' is also set True

kinnou02:
You need to rebase

kinnou02 (approved):

Issue #2279, opened 2018-01-04
Problem on stop_schedules with duration of multiple days for routes active only on certain days


eturck:
Hello!

We are currently working on our stop_schedules API with kraken as the backend. It was pretty simple but while testing I was missing a few departures in my response. I tried using the Navitia API to see if I made a mistake but I have the same problems with it.

One of our line (line 21) is running Monday to Saturday with a different terminus on the week-end.
If I ask, on a Thursday,  the stop_schedules for one of the stop_area on the line, with a duration of two weeks for example (duration=86400*14) I will have : 
- On the week route, all the departures for Thursday and Friday,
- On the week-end route, nothing.
What I think we should have (and what I want :D) : 
- On the week route, all the departures for Thursday and Friday + all next week Monday through Friday + the week after Monday to Wednesday,
- On the week-end route, departures of the next two Saturdays.

I took a quick look at the code and it's pretty clear why it's doing this. In [next_valid_discrete ](https://github.com/CanalTP/navitia/blob/dev/source/routing/next_stop_time.cpp#L118)(in next_stop_time.cpp) :
```cpp
    auto date = DateTimeUtils::date(dt);
    for (const auto* st: dataRaptor.next_stop_time_data.stop_time_range_after(jpp_idx, dt, stop_event)) {
        (...)
    }

    //if none was found, we try again the next day
    date++;
    for (const auto* st: dataRaptor.next_stop_time_data.stop_time_range_forward(jpp_idx, stop_event)) {
        (...)
    }

    // if nothing found, return max
    return {nullptr, DateTimeUtils::inf};
```

We search for the next valid stop time only for today and tomorrow, nothing else.
So in get_stop_times, when [we initialize each journey pattern point with the first departure ](https://github.com/CanalTP/navitia/blob/dev/source/routing/get_stop_times.cpp#L60)we only look for those two days, and same thing [while we loop until max_dt](https://github.com/CanalTP/navitia/blob/dev/source/routing/get_stop_times.cpp#L87) (which his from_datetime + duration), when we are on Friday at the last departure of this day, we only look for the next one Friday and Saturday. Since there is no departure, the next one is Monday, we never output the departures for next week.

As always I have a few questions :) : 
- Is it intended? Since it's in routing I was wondering if with RAPTOR we explicitly don't want to look for a departure  further than the next day.
[...]

PR #2309, opened 2018-01-25
Kraken: fix schedules for lines not circulating on two consecutive days or more

eturck:
This is following issue #2279.

This problem is occurring on stop/route_schedules with a duration of multiple days. Our use case is that we want to know which lines have a service on the next seven days at a certain stop. So we ask stop_schedules the next 2 departures in the next 7 days.
Some of our lines have routes only active on Saturdays or Sundays, so when we test it at the beginning of the week, we have no results.

After @TeXitoi's answer I took the safe approach, not wanting to impact anything performance-wise. I used the already existing bound for next/previous_valid_discrete (I added it to the frequency functions).
It's not just used to stop searching after a certain bound now, I also use it to determine how far into the future I need to go. For example next_valid_discrete will search on the current day and increment the date until we pass the bound as long as we have not found the next stop time.

If the bound is not set, I set it to DateTime(dt.date + 2, 0). It's kind of "random", but it's to emulate the current behavior. I don't know if there is a more elegant way to do it. I've added a bunch of tests too.

Let me know if that's seems an OK way to fix it, if I miss something or if I need to change / add anything.

### DIFF
--- a/source/routing/next_stop_time.cpp
+++ b/source/routing/next_stop_time.cpp
@@ -123,20 +123,10 @@ next_valid_discrete(const StopEvent stop_event,
         const type::VehicleProperties& vehicle_props,
         const DateTime bound) {
     auto date = DateTimeUtils::date(dt);
-    for (const auto* st: dataRaptor.next_stop_time_data.stop_time_range_after(jpp_idx, dt, stop_event)) {
-        assert(dataRaptor.jp_container.get_jpp(*st) == jpp_idx);
-        const uint32_t hour = (stop_event == StopEvent::pick_up) ? st->boarding_time : st->alighting_time;
-        const DateTime cur_dt = DateTimeUtils::set(date, DateTimeUtils::hour(hour));
-        if (bound < cur_dt) { return {nullptr, DateTimeUtils::inf}; }
-        if (is_valid(st, date, true, rt_level, vehicle_props)) {
-            return {st, cur_dt};
-        }
-    }
+    // On the first iteration we only check the stop_times after dt, we init the range with this
+    auto st_range = dataRaptor.next_stop_time_data.stop_time_range_after(jpp_idx, dt, stop_event);
 
-    const auto st_range = dataRaptor.next_stop_time_data.stop_time_range_forward(jpp_idx, stop_event);
-    //if none was found, we try again until we crossed the bound
-    while(DateTimeUtils::date(bound) > date) {
-        date++;
+    while(DateTimeUtils::date(bound) >= date) {
         for (const auto* st: st_range) {
             assert(dataRaptor.jp_container.get_jpp(*st) == jpp_idx);
             const uint32_t hour = (stop_event == StopEvent::pick_up) ? st->boarding_time : st->alighting_time;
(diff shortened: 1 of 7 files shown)


buildbotctp:

kinnou02:
ok to test.

source/routing/next_stop_time.cpp
```diff
@@ -133,15 +133,18 @@ next_valid_discrete(const StopEvent stop_event,
...
-        const DateTime cur_dt = DateTimeUtils::set(date, DateTimeUtils::hour(hour));
-        if (bound < cur_dt) { return {nullptr, DateTimeUtils::inf}; }
-        if (is_valid(st, date, true, rt_level, vehicle_props)) {
-            return {st, cur_dt};
+    const auto st_range = dataRaptor.next_stop_time_data.stop_time_range_forward(jpp_idx, stop_event);
+    //if none was found, we try again until we crossed the bound
+    while(DateTimeUtils::date(bound) > date) {
+        date++;
+        for (const auto* st: st_range) {
+            assert(dataRaptor.jp_container.get_jpp(*st) == jpp_idx);
```
xlqian: this is a 100% repeated code 🐱 as the L127 can you refactorize that if possbile?

source/routing/next_stop_time.cpp
```diff
@@ -153,98 +156,71 @@ static std::pair<const type::StopTime*, DateTime>
 next_valid_frequency(const StopEvent stop_event,
         const dataRAPTOR& dataRaptor,
         const JppIdx jpp_idx,
-        const DateTime dt,
+        DateTime dt,
```
pbougue: It will probably imply code a bit away of your PR but I'd rather add `const`s than remove'em.
But it's int underlying here, so copied...

pbougue (approved):
Looks good
I'm just wondering if we might hit some "DOS" issues with people going a bit too far in time with big counts, or things like that.

source/routing/next_stop_time.cpp
eturck: I can keep the `const` and start by copying dt in another `DateTime` if it's better. That's not really a problem I just took the easy way without much thinking ;).

eturck:
@pbougue it could probably cause some issue with people getting a stop_schedules for a year or something. Maybe like there is a max_count in the config you could have a max_duration too? It's not perfect if everybody can request something like two weeks but it's limited compare to the end production date as a limit.

source/routing/next_stop_time.cpp
pbougue: Or even... propagate it 😝

eturck: I'm not really sure to fully understand sorry :worried:. Do want me to just keep the `const` and work on a copy of the `DateTime` and also add a `const` to the dt of `get_next_stop_time`? Or something else/more?

pbougue: Probably work on a copy for the needed part (as it is mutable, which I didn't remark at first sight).
And also at first my idea was to add `const`s to function params you are using (`DateTimeUtils::date()` and `get_next_stop_time()`): it's legacy, you can do it if you want but it's not your responsibility at all :)

eturck:
I changed next/previous_valid_discrete to reduce repeated code and added some const :).

xlqian (approved):

PR #2311, opened 2018-01-26
Jormun : [Siri Lite] Add new tests

benoit-bst:
New tests :
- Real time passage with no departure time
- Real time passage with departure time in the past

New check :
- If possible route is not empty

### DIFF
--- a/source/jormungandr/tests/proxy_realtime_sirilite_integration_tests.py
+++ b/source/jormungandr/tests/proxy_realtime_sirilite_integration_tests.py
@@ -138,6 +138,38 @@ class TestSiriLite(AbstractTestFixture):
                             "MonitoringRef": {"value": "syn_stoppoint21"},
                             "RecordedAtTime": "2017-07-19T13:01:54.727Z"
                         },
+                        # we add a passage with no departure time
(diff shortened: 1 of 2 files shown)


source/jormungandr/tests/proxy_realtime_sirilite_integration_tests.py
```diff
@@ -138,6 +138,38 @@ def test_stop_schedule(self):
...
+                                "MonitoredCall": {
+                                    "ExpectedDepartureTime": "",
+                                    "StopPointName": [{"value": "the stop"}],
+                                    "VehicleAtStop": False
+                                }
+                            },
+                            "MonitoringRef": {"value": "syn_stoppoint21"},
+                            "RecordedAtTime": "2017-07-19T13:01:54.727Z"
+                        },
+                        # we add a passage with departure time in the past
```
pbougue: could be nice to add `(we should not consider it)` (same above)

pbougue (approved):

benoit-bst:
https://jira.canaltp.fr/browse/NAVITIAII-2350

PR #2317, opened 2018-01-31
jormungandr: test on ridesharing with a public transport

kadhikari:


source/jormungandr/tests/instant_system_new_default_routing_tests.py
```diff
@@ -205,3 +206,61 @@ def test_basic_ride_sharing(self):
...
+
+    def test_ride_sharing_with_pt(self):
+        """
+        test ridesharing_jouneys details
+        """
+        q = journey_basic_query + \
+            "&last_section_mode[]=walking" + \
+            "&first_section_mode[]=ridesharing" + \
+            "&ridesharing_speed=2.5" + \
+            "&debug=true"
```
pbougue: you should not use this debug and still have the result, we have to remove it for tests.
But you will have to provide a mock response for the ridesharing to prevent it being filtered

source/jormungandr/tests/instant_system_new_default_routing_tests.py
```diff
@@ -205,3 +206,61 @@ def test_basic_ride_sharing(self):
...
+        assert rs_section.get('duration') == 7
+
+        #second section is of public transport
+        pt_section = sections[1]
+        assert pt_section.get('type') == 'public_transport'
+        assert pt_section.get('from').get('id') == 'stop_point:stopB'
+        assert pt_section.get('to').get('id') == 'stop_point:stopA'
+        assert pt_section.get('duration') == 2
+
+        #tird section is of walking
```
pbougue: typo: t`h`ird

source/jormungandr/tests/instant_system_new_default_routing_tests.py
```diff
@@ -205,3 +206,61 @@ def test_basic_ride_sharing(self):
...
+            "&first_section_mode[]=ridesharing" + \
+            "&ridesharing_speed=2.5" + \
+            "&debug=true"
+        response = self.query_region(q)
+        self.is_valid_journey_response(response, q, check_journey_links=False)
+
+        journeys = get_not_null(response, 'journeys')
+        assert len(journeys) == 2
+
+        #The first journey is direct ridesharing
```
pbougue, line 224: if this journey is not present without the `debug=true` it's OK, but we have to check it's there (and it's the only one) if we request with `max_ridesharing_duration_to_pt=0`

pbougue (changes requested):

xlqian (approved):

kinnou02 (approved):

pbougue (approved):

PR #2318, opened 2018-01-31
Jormungandr : Use gevent patch ssl

mbounabi:
https://jira.canaltp.fr/browse/NAVP-779

Résultats:
https://jira.canaltp.fr/browse/NAVP-780
https://jira.canaltp.fr/browse/NAVP-781
https://jira.canaltp.fr/browse/NAVP-782

### DIFF
--- a/source/jormungandr/jormungandr/api.py
+++ b/source/jormungandr/jormungandr/api.py
@@ -44,9 +44,9 @@ from jormungandr.authentication import get_user, get_token, get_app_name, get_us
 import six
 
 if rest_api.app.config.get('PATCH_WITH_GEVENT_SOCKET', False):
-    logger = logging.getLogger('jormungandr.patch_gevent_socket')
+    logger = logging.getLogger('jormungandr.patch_gevent')
     logger.info("Attention! You'are patching requests.packages.urllib3.connection.connection.socket with gevent.socket,"
-                "parallel http calling by requests is activated")
+                "parallel http/https calling by requests is activated")
     # This line replaces the gevent.monkey.patch_socket()
     # Note that "monkey_patch" only patches on http request because we want asynchronisation on that,
     # but we don't want that for reddis because it may cause performance regression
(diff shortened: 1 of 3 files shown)


xlqian (approved):
We may want to enable patch_ssl and patch_socket in the integration test https://github.com/CanalTP/navitia/blob/dev/source/jormungandr/jormungandr/test_settings.py @kinnou02 @pbougue @TeXitoi

source/jormungandr/jormungandr/api.py
```diff
@@ -48,15 +48,18 @@
...
-    # the reason why we don't use patch_socket() at the very beginning of jormungandr is
-    # that it caused a mysterious performance regression for certain requests, thus we patch
-    # only at places where asynchronisation is needed
     # Note that "monkey_patch" only patches on http request because we want asynchronisation on that,
     # but we don't want that for reddis because it may cause performance regression
     import requests
     import gevent.socket
     requests.packages.urllib3.connection.connection.socket = gevent.socket
 
+if rest_api.app.config.get('PATCH_GEVENT_SSL', False):
```
kinnou02: I don't see a case where we want to patch the socket but not the ssl since it will break things

mbounabi: Not pertinent indeed. Changed!

kinnou02 (approved):

TeXitoi:
@xlqian we can but it will not test so much as we don't call http services in the tests.

kinnou02 (approved):

xlqian:
@TeXitoi yes, i'm aware of that, I mean, just in case those patches have some side effects(that we haven't detected yet) on the code, the integration tests will **_roughly_** protect us

kinnou02:
You can change it for the tests and also for every environement except the production (You can also create the PR for prod already but with donotmerge tag), it will help us see if there is any problems.

mbounabi:
jenkins, test this please!

benoit-bst (approved):

pbougue (approved):
OK for me.
As Alex said, yo can open a PR on deployment conf, co that every platform is like dev, except for prod : https://github.com/CanalTP/navitia_deployment_conf/blob/master/dev.py#L187
And also a "do_not_merge"-PR for prod aside.

mbounabi:
PRs on navitia_deployment_conf have been created:
https://github.com/CanalTP/navitia_deployment_conf/pull/368
https://github.com/CanalTP/navitia_deployment_conf/pull/369

PR #2326, opened 2018-02-06
autocomplete: bug on longest common substring

eturck:
Hello.

A small thing I came across while playing with places.
When I was inputing "Jean Jaurès Toulouse", which is the exact name of one of our main stop_area I had a bad longest common substring (I looked at scores in the pbf).
This was caused by two things:
- When building the stop_area autocomplete an extra space was added. I had a better lcs when inputing "Jean Jaurès  Toulouse" (2 spaces) because of this,
- The lcs was comparing the raw input to the indexed string which is the name of the stop_area in lowercase without any accents. I transform the raw input with the same method before getting the lcs now.

### DIFF
--- a/source/autocomplete/autocomplete.h
+++ b/source/autocomplete/autocomplete.h
@@ -331,7 +331,7 @@ struct Autocomplete
         auto global_score = word_quality_list.at(position).score;
 
         const auto& indexed_str = indexed_string.at(position);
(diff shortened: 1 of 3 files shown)


buildbotctp:

kinnou02 (approved):

TeXitoi (approved):

PR #2328, opened 2018-02-09
Jormungandr : Stat Instant System ridesharing

benoit-bst:


benoit-bst:
JIRA :
https://jira.canaltp.fr/browse/NAVP-789

pbougue (approved):

kinnou02:
You should use a custom logger as it will be easier to find it later. Also we now have json log, so you should set the number of journey found as an extra data so it will be in a field of it's own so we won't have to parse the string manually.

PR #2330, opened 2018-02-14
Jormungandr: Parallel calls to real-time proxies

mbounabi:


source/jormungandr/jormungandr/default_settings.py
```diff
@@ -158,6 +158,9 @@
 
 GREENLET_POOL_SIZE = int(os.getenv('JORMUNGANDR_GEVENT_POOL_SIZE', 10))
 
+# Number of greenlets simultaneously used for Real-Time proxies calls
+GREENLET_RT_POOL_SIZE = int(os.getenv('JORMUNGANDR_GEVENT_RT_POOL_SIZE', 10))
```
kinnou02: It also need to be configurable by rt proxy instances

source/jormungandr/jormungandr/schedule.py
```diff
@@ -256,11 +261,21 @@ def next_departures(self, request):
...
+
+        def worker(rt_proxy, route_point, template, request, resp):
+            return resp, rt_proxy, route_point, template, self._get_next_realtime_passages(rt_proxy, route_point, request)
+
         for route_point, template in route_points.items():
             rt_proxy = self._get_realtime_proxy(route_point)
             if rt_proxy:
-                next_rt_passages = self._get_next_realtime_passages(rt_proxy, route_point, request)
-                rt_proxy._update_passages(resp.next_departures, route_point, template, next_rt_passages)
+                futures.append(pool.spawn(worker, rt_proxy, route_point, template, request, resp))
```
kinnou02, line 278: You will probably have to pass the `request_id` too

kinnou02 (changes requested):
Did you try without async nor monkeypatch on request to check that everything still work?

source/jormungandr/jormungandr/schedule.py
mbounabi: As discussed, the request id can't be used in greenlet as is

source/jormungandr/jormungandr/default_settings.py
mbounabi: As discussed, the number of greenlets is now configurable for each instance in the db

source/jormungandr/jormungandr/instance.py
```diff
@@ -398,13 +403,8 @@ def _send_and_receive(self,
                          timeout=app.config.get('INSTANCE_TIMEOUT', 10000),
                          quiet=False,
                          **kwargs):
+
         with self.socket(self.context) as socket:
-            try:
-                request.request_id = flask.request.id
```
kinnou02: I still want to set the (flask) request id to the (pb) request

source/jormungandr/jormungandr/schedule.py
kinnou02: The point was to find a way to have it in the greenlet, I won't merge it without it. It's far too valuable for debuging purpose.

kinnou02 (changes requested):

source/tyr/migrations/versions/242138b08c92_realtime_pool_size.py
```diff
@@ -0,0 +1,54 @@
...
+down_revision = '3acdde6329db'
+
+from alembic import op
+import sqlalchemy as sa
+
+
+def upgrade():
+    ### commands auto generated by Alembic - please adjust! ###
+    op.add_column('instance', sa.Column('realtime_pool_size', sa.Integer(), nullable=True))
+    op.alter_column('instance', 'bss_provider',
```
kinnou02: This isn't related to this PR

source/jormungandr/jormungandr/schedule.py
```diff
@@ -181,16 +184,24 @@ def _get_realtime_proxy(self, route_point):
...
         return rt_system
 
-    def _get_next_realtime_passages(self, rt_system, route_point, request):
+    def _get_next_realtime_passages(self, rt_system, route_point, request, **kwargs):
         log = logging.getLogger(__name__)
-
         next_rt_passages = None
+
+        try:
+            request.request_id = flask.request.id
```
kinnou02: This won't probably work since request is a dict here, and even if it doesn't crash it won't do what we want.
In fact we dont't do any call to kraken in this greenlet so there is no point to set the `request_id` of a pbrequest.

What I wanted is here: http://flask.pocoo.org/docs/0.12/api/#flask.copy_current_request_context

mbounabi:
As the decorator provided by flask (@copy_current_request_context) will generate an assertion error with multiple greenlets, new functions are created to use flask request context in greenlet

source/jormungandr/jormungandr/schedule.py
```diff
@@ -256,11 +261,25 @@ def next_departures(self, request):
                             for rp in resp.route_points)
 
         rt_proxy = None
+        futures = []
+        pool = gevent.pool.Pool(self.instance.realtime_pool_size)
+
+        # Copy the current request context to be used in greenlet
+        reqctx=utils.copy_flask_request_context()
```
kinnou02: spacing

source/jormungandr/jormungandr/utils.py
```diff
@@ -524,3 +525,41 @@ def get_house_number(housenumber):
...
+    # Copy flask request context to be used in greenlet
+    top = flask._request_ctx_stack.top
+    if top is None:
+        raise RuntimeError('This function can only be used at local scopes '
+                            'when a request context is on the stack.  For instance within '
+                            'view functions.')
+    return top.copy()
+
+@contextmanager
+def copy_context_in_greenlet_stack(request_context):
```
xlqian, line 547: https://stackoverflow.com/questions/48814122/unable-to-copy-current-flask-request-context-to-multiple-gevent-threads/48929494#48929494

xlqian (approved):

kinnou02 (approved):

source/jormungandr/jormungandr/schedule.py
mbounabi: fixed

kinnou02:
Jenkins, test this please.

PR #2331, opened 2018-02-14
Jormun: handle crowfly in ridesharing

xlqian:
https://jira.canaltp.fr/browse/NAVP-792

### DIFF
--- a/source/jormungandr/jormungandr/scenarios/new_default.py
+++ b/source/jormungandr/jormungandr/scenarios/new_default.py
@@ -556,7 +556,16 @@ def _is_fake_car_section(section):
 
 
 def _switch_back_to_ridesharing(response, is_first_section):
+    """
+    
+    :param response: a pb_response returned by kraken
+    :param is_first_section: a bool indicates that if the first_section or last_section is a ridesharing section
+                             True if the first_section is, False if the last_section is
+    :return: 
+    """
     for journey in response.journeys:
+        if len(journey.sections) == 0:
+            continue
         section_idx = 0 if is_first_section else -1
         section = journey.sections[section_idx]
         if _is_fake_car_section(section):
--- a/source/jormungandr/jormungandr/scenarios/new_default.py
+++ b/source/jormungandr/jormungandr/scenarios/new_default.py
@@ -548,20 +548,17 @@ def _tag_by_mode(responses):
 
 
 def _is_fake_car(i, section, sections):
-    if section.type == response_pb2.STREET_NETWORK \
-            and section.street_network.mode == response_pb2.Car:
-        if (len(sections) > i + 1 and sections[i + 1].type != response_pb2.PARK) or \
-                (i > 0 and sections[i - 1].type != response_pb2.LEAVE_PARKING) \
-                or len(sections) == 1:
-                    return True
-    return False
+    return section.type == response_pb2.STREET_NETWORK \
+            and section.street_network.mode == response_pb2.Car and\
+            ((len(sections) > i + 1 and sections[i + 1].type != response_pb2.PARK) or
+             (i > 0 and sections[i - 1].type != response_pb2.LEAVE_PARKING)
+             or len(sections) == 1)
 
 
 def _is_fake_crowfly(i, section, sections, dep_mode, arr_mode):
-    if (dep_mode == "ridesharing" and i == 0 and section.type == response_pb2.CROW_FLY) or \
-            (arr_mode == "ridesharing" and i == (len(sections) - 1) and section.type == response_pb2.CROW_FLY):
-                return True
-    return False
+    return (dep_mode == "ridesharing" and i == 0 and section.type == response_pb2.CROW_FLY) or \
+            (arr_mode == "ridesharing" and i == (len(sections) - 1) and section.type == response_pb2.CROW_FLY)
+
 
 
 def _switch_back_to_ridesharing(response, dep_mode, arr_mode):
--- a/source/jormungandr/jormungandr/scenarios/new_default.py
+++ b/source/jormungandr/jormungandr/scenarios/new_default.py
@@ -555,16 +555,16 @@ def _is_fake_car_section(section):
            section.street_network.mode == response_pb2.Car
 
 
-def _switch_back_to_ridesharing(response, section_index):
-
+def _switch_back_to_ridesharing(response, sections_ind):
     for journey in response.journeys:
-        section = journey.sections[section_index]
-        if _is_fake_car_section(section):
-            section.street_network.mode = response_pb2.Ridesharing
-            journey.durations.ridesharing += section.duration
-            journey.durations.car -= section.duration
-            journey.distances.ridesharing += section.length
-            journey.distances.car -= section.length
+        for i in sections_ind:
+            section = journey.sections[i]
+            if _is_fake_car_section(section):
+                section.street_network.mode = response_pb2.Ridesharing
+                journey.durations.ridesharing += section.duration
+                journey.durations.car -= section.duration
+                journey.distances.ridesharing += section.length
+                journey.distances.car -= section.length
 
 def nb_journeys(responses):
     return sum(1 for r in responses for j in r.journeys if not journey_filter.to_be_deleted(j))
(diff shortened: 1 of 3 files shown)


source/jormungandr/jormungandr/scenarios/new_default.py
```diff
@@ -546,20 +546,34 @@ def _tag_by_mode(responses):
...
-                        (i > 0 and journey.sections[i-1].type != response_pb2.LEAVE_PARKING) \
-                        or len(journey.sections) == 1:
-                    #TODO: handle crowfly :(
-                    section.street_network.mode = response_pb2.Ridesharing
-                    journey.durations.ridesharing += section.duration
-                    journey.durations.car -= section.duration
-                    journey.distances.ridesharing += section.length
-                    journey.distances.car -= section.length
+            if _is_fake_car(i, section, journey.sections) or \
+                    _is_fake_crowfly(i, section, journey.sections, dep_mode, arr_mode):
```
kinnou02: Since we look at `dep_mode` and `arr_mode` it's probably easier to always do it this way and no more guessing with the absence of park section

kinnou02 (dismissed):

pbougue (commented):
Would be nice to add test if it's easily doable.

xlqian:
@pbougue that's why i added the do_not_merge tag while i was working on the test

source/jormungandr/jormungandr/scenarios/new_default.py
```diff
@@ -546,20 +546,26 @@ def _tag_by_mode(responses):
         for j in r.journeys:
             _tag_journey_by_mode(j)
 
-def _switch_back_to_ridesharing(response):
+
+def _is_fake_car(i, section, sections, dep_mode, arr_mode):
```
pbougue: please rename that `i` param 🙏  (if you keep the code that way)
also it's redundant to have `index and sections` on one side, and `section` on the other side

source/jormungandr/jormungandr/scenarios/new_default.py
```diff
@@ -546,20 +546,26 @@ def _tag_by_mode(responses):
...
-                if (len(journey.sections) > i+1 and journey.sections[i+1].type != response_pb2.PARK) or \
-                        (i > 0 and journey.sections[i-1].type != response_pb2.LEAVE_PARKING) \
-                        or len(journey.sections) == 1:
-                    #TODO: handle crowfly :(
-                    section.street_network.mode = response_pb2.Ridesharing
-                    journey.durations.ridesharing += section.duration
-                    journey.durations.car -= section.duration
-                    journey.distances.ridesharing += section.length
-                    journey.distances.car -= section.length
+            if _is_fake_car(i, section, journey.sections, dep_mode, arr_mode):
```
pbougue: It looks a bit weird to loop over all sections when only the first and the last are of interest... To me a function for that is obfuscating the code for no gain (half of the params are useless depending on the case, if not all when on PT section).
Maybe a function with only `(section, requested_mode)` params could be nicer.

xlqian: you're right, I hesitated A LOT and finally i left it that way :D, since you spot that, i'll make it in your way

source/jormungandr/jormungandr/scenarios/new_default.py
```diff
@@ -869,7 +873,9 @@ def worker(dep_mode, arr_mode, instance, request, request_id):
             for idx, j in enumerate(local_resp.journeys):
                 j.internal_id = "{resp}-{j}".format(resp=self.nb_kraken_calls, j=idx)
             if 'ridesharing' in dep_mode or 'ridesharing' in arr_mode:
-                _switch_back_to_ridesharing(local_resp)
+                # dep_mode and arr_mode cannot be ridesharing at the same time
```
TeXitoi: False, it can if you ask for only ridesharing in last and first section mode.

pbougue (approved):

source/jormungandr/jormungandr/scenarios/new_default.py
```diff
@@ -546,21 +546,25 @@ def _tag_by_mode(responses):
...
-                    journey.durations.car -= section.duration
-                    journey.distances.ridesharing += section.length
-                    journey.distances.car -= section.length
 
+def _is_fake_car_section(section):
+    """
+    This function test if the section is a fake car section
+    """
+    return (section.type == response_pb2.STREET_NETWORK or section.type == response_pb2.CROW_FLY) and \
+           section.street_network.mode == response_pb2.Car
```
TeXitoi: If you do a journey with `first_section_mode[]=car&first_section_mode[]=ridesharing&first_section_mode[]=walking` it will not work.

TeXitoi (changes requested):

source/jormungandr/jormungandr/scenarios/new_default.py
xlqian: @TeXitoi  `then it's impossible to achieve this PR's objective in a simple way... since when the OSM is missing, i can't tell if the journey is for car or ridesharing.... ` <-- not true...

source/jormungandr/jormungandr/scenarios/new_default.py
xlqian: ok

source/jormungandr/jormungandr/scenarios/new_default.py
pbougue: At that point we only have one `dep_mode` and one `arr_mode` (already dispatched, and tested earlier), no?
To be tested but it looks OK.

source/jormungandr/jormungandr/scenarios/new_default.py
pbougue: OK with @TeXitoi we should manage it separatly, then (first_section, then last_section).

source/jormungandr/jormungandr/scenarios/new_default.py
TeXitoi: If a test shows that it's working, OK.

xlqian:
Adding more tests...

xlqian:
Jenkins, test this please!

source/jormungandr/jormungandr/scenarios/new_default.py
```diff
@@ -546,22 +546,26 @@ def _tag_by_mode(responses):
...
+    This function test if the section is a fake car section
+    """
+    return (section.type == response_pb2.STREET_NETWORK or section.type == response_pb2.CROW_FLY) and \
+            section.street_network.mode == response_pb2.Car
 
 
+def _switch_back_to_ridesharing(response, is_first_section):
+    for journey in response.journeys:
+        section_idx = 0 if is_first_section else -1
+        section = journey.sections[section_idx]
```
pbougue: could be nice to test that you have a section

source/jormungandr/jormungandr/scenarios/new_default.py
```diff
@@ -546,22 +546,26 @@ def _tag_by_mode(responses):
...
+
+def _is_fake_car_section(section):
+    """
+    This function test if the section is a fake car section
+    """
+    return (section.type == response_pb2.STREET_NETWORK or section.type == response_pb2.CROW_FLY) and \
+            section.street_network.mode == response_pb2.Car
 
 
+def _switch_back_to_ridesharing(response, is_first_section):
```
pbougue: minor: nice to have a comment on is_first_section use

source/jormungandr/tests/instant_system_new_default_routing_tests.py
```diff
@@ -292,6 +292,33 @@ def test_start_ridesharing_with_pt(self):
         assert 'ridesharing' in journeys[0].get('tags')
         assert journeys[0].get('durations').get('ridesharing') == 37
 
+        q = journey_basic_query + \
+            ("&first_section_mode[]=ridesharing"
+             "&first_section_mode[]=car"
+             "&last_section_mode[]=walking"
+             "&ridesharing_speed=2.5")
```
pbougue: If you add `car_speed=2.5` you should be able to test that you get a fourth journey car+PT
Also would be nice to test section types on each journey (and a comment to explain direct car, fallback ridesharing, etc.)

benoit-bst (approved):

pbougue (approved):

TeXitoi (approved):

PR #2333, opened 2018-02-15
Jormun: fix bug when requesting ridesharing only on last section 

pbougue:


TeXitoi (approved):

benoit-bst (approved):

source/jormungandr/jormungandr/scenarios/new_default.py
```diff
@@ -822,7 +822,8 @@ def fill_journeys(self, request_type, api_request, instance):
         compute_car_co2_emission(pb_resp, api_request, instance)
         tag_journeys(pb_resp)
 
-        if 'ridesharing' in ridesharing_req['origin_mode'] and instance.ridesharing_services:
+        if 'ridesharing' in ridesharing_req['origin_mode'] + ridesharing_req['destination_mode'] \
```
xlqian: (just a technical detail) i'd prefer `if 'ridesharing' in (ridesharing_req['origin_mode'], ridesharing_req['destination_mode'])`, since the `+` operator will do a useless copy(in this case)

xlqian (approved):

source/jormungandr/jormungandr/scenarios/new_default.py
xlqian: OTAN pour moi

mbounabi (approved):

kinnou02 (approved):

PR #2335, opened 2018-02-20
Jormun: fix tag ridesharing

xlqian:
fix a bug in the PR https://github.com/CanalTP/navitia/pull/2331


source/jormungandr/jormungandr/scenarios/new_default.py
```diff
@@ -531,8 +531,9 @@ def _tag_journey_by_mode(journey):
         elif section.type == response_pb2.STREET_NETWORK \
```
pbougue: We should add it here also

pbougue (approved):

PR #2338, opened 2018-02-22
Jormun: Fix address in poi in api global autocomplete 

kadhikari:
Ticket: https://jira.canaltp.fr/browse/NAVITIAII-2378

### DIFF
--- a/source/jormungandr/jormungandr/autocomplete/geocodejson.py
+++ b/source/jormungandr/jormungandr/autocomplete/geocodejson.py
@@ -133,24 +133,28 @@ def get_lon_lat(obj):
     return lon, lat
 
 
-def create_address_field(geocoding, poi_lat, poi_lon):
+def create_address_field(geocoding, poi_lat=None, poi_lon=None):
     if not geocoding:
         return None
     coord = geocoding.get('coord', {})
     lat = str(coord.get('lat')) if coord and coord.get('lat') else poi_lat
     lon = str(coord.get('lon')) if coord and coord.get('lon') else poi_lon
-
-    return {
-        "id": geocoding.get('id'),
+    address_id = '{lon};{lat}'.format(lon=lon, lat=lat)
+    resp = {
+        "id": address_id,
         "label": geocoding.get('label'),
         "name": geocoding.get('name'),
         "coord": {
             "lat": lat,
             "lon": lon
         },
-        "house_number": get_house_number(geocoding.get('housenumber')),
-        "administrative_regions": create_administrative_regions_field(geocoding) or create_admin_field(geocoding)
+        "house_number": get_house_number(geocoding.get('housenumber'))
     }
+
+    admins = create_administrative_regions_field(geocoding) or create_admin_field(geocoding)
+    if admins:
+        resp['administrative_regions'] = admins
+    return resp
 
 
 class AdministrativeRegionField(fields.Raw):
@@ -212,7 +216,6 @@ class PoiField(fields.Raw):
         geocoding = obj.get('properties', {}).get('geocoding', {})
         poi_types = geocoding.get('poi_types', [])
 
-        # TODO add address, properties attributes
         res = {
             "id": geocoding.get('id'),
             "coord": {
@@ -337,6 +340,8 @@ class GeocodeJson(AbstractAutocomplete):
     TYPE_HOUSE = "house"
     TYPE_STREET = "street"
 
+    TYPE_LIST = [TYPE_STOP_AREA, TYPE_CITY, TYPE_POI, TYPE_HOUSE, TYPE_STREET]
+
     def __init__(self, **kwargs):
         self.host = kwargs.get('host')
         self.timeout = kwargs.get('timeout', 10)
@@ -362,7 +367,18 @@ class GeocodeJson(AbstractAutocomplete):
             raise TechnicalError('error in autocomplete request')
 
     @classmethod
-    def _clean_response(cls, response, depth=3):
+    def _clean_response(cls, response, depth=1):
+
+        def is_deleteable(_key, _value, _depth):
+            if _depth > -1:
+                return False
+            else:
+                if _key == 'administrative_regions':
+                    return True
+                elif isinstance(_value, dict) and _value.get('type') in cls.TYPE_LIST:
+                    return True
+                else:
+                    return False
 
         def _clear_object(obj):
             if isinstance(obj, list):
@@ -371,19 +387,12 @@ class GeocodeJson(AbstractAutocomplete):
                 obj.clear()
 
         def _manage_depth(_key, _value, _depth):
-            if _depth == -1:
+
+            if is_deleteable(_key, _value, _depth):
                 _clear_object(_value)
-            #We should not delete any element of admins of parent node
-            elif _key == 'administrative_regions':
-                if depth == -1:
-                    _clear_object(_value)
-            elif isinstance(_value, list):
-                for obj in _value:
-                    for k, v in obj.items():
-                        _manage_depth(k, v, _depth-1)
             elif isinstance(_value, dict):
                 for k, v in _value.items():
-                    if _depth == -1:
+                    if is_deleteable(k, v, _depth):
                         _clear_object(v)
                     else:
                         _manage_depth(k, v, _depth-1)
@@ -400,7 +409,7 @@ class GeocodeJson(AbstractAutocomplete):
         return response
 
     @classmethod
-    def response_marshaler(cls, response_bragi, uri=None, depth=3):
+    def response_marshaler(cls, response_bragi, uri=None, depth=1):
         cls._check_response(response_bragi, uri)
         json_response = response_bragi.json()
         #Clean dict objects depending on depth passed in request parameter.
@@ -459,7 +468,9 @@ class GeocodeJson(AbstractAutocomplete):
 
         if request.get("from"):
             params["lon"], params["lat"] = self.get_coords(request["from"])
-        if request.get("depth"):
+
+        #valid values of depth 0 to N
+        if 'depth' in request:
             params["depth"] = request["depth"]
         return params
 
@@ -476,7 +487,7 @@ class GeocodeJson(AbstractAutocomplete):
             method = requests.post
 
         raw_response = self.call_bragi(url, method, **kwargs)
-        depth = request.get('depth', 3)
+        depth = request.get('depth', 1)
 
         return self.response_marshaler(raw_response, None, depth)
 
--- a/source/jormungandr/jormungandr/autocomplete/geocodejson.py
+++ b/source/jormungandr/jormungandr/autocomplete/geocodejson.py
@@ -93,7 +93,7 @@ def create_administrative_regions_field(geocoding):
         zip_codes = admin.get('zip_codes', [])
         response.append({
             "insee": admin.get('insee'),
-            "name": admin.get('label'),
+            "name": admin.get('name'),
             "level":
                 int(admin.get('level')) if admin.get('level') else None,
             "coord": {
@@ -224,13 +224,15 @@ class PoiField(fields.Raw):
             },
             "label": geocoding.get('label'),
             "name": geocoding.get('name'),
-            "administrative_regions":
-                create_administrative_regions_field(geocoding) or create_admin_field(geocoding),
             "properties": {p.get('key'): p.get('value') for p in geocoding.get("properties", [])},
             "address": create_address_field(geocoding.get("address"), poi_lat=lat, poi_lon=lon)
         }
         if isinstance(poi_types, list) and poi_types:
             res['poi_type'] = poi_types[0]
+
+        admins = create_administrative_regions_field(geocoding) or create_admin_field(geocoding)
+        if admins:
+            res['administrative_regions'] = admins
         return res
 
 
@@ -469,9 +471,6 @@ class GeocodeJson(AbstractAutocomplete):
         if request.get("from"):
             params["lon"], params["lat"] = self.get_coords(request["from"])
 
-        #valid values of depth 0 to N
-        if 'depth' in request:
-            params["depth"] = request["depth"]
         return params
 
     def get(self, request, instances):
--- a/source/jormungandr/jormungandr/autocomplete/geocodejson.py
+++ b/source/jormungandr/jormungandr/autocomplete/geocodejson.py
@@ -193,7 +193,7 @@ class AddressField(fields.Raw):
         lon, lat = get_lon_lat(obj)
         geocoding = obj.get('properties', {}).get('geocoding', {})
 
-        return {
+        res = {
             "id": '{};{}'.format(lon, lat),
             "coord": {
                 "lon": lon,
@@ -201,11 +201,13 @@ class AddressField(fields.Raw):
             },
             "house_number": get_house_number(geocoding.get('housenumber')),
             "label": geocoding.get('label'),
-            "name": geocoding.get('name'),
-            "administrative_regions":
-                create_administrative_regions_field(geocoding) or create_admin_field(geocoding) ,
+            "name": geocoding.get('name')
         }
 
+        admins = create_administrative_regions_field(geocoding) or create_admin_field(geocoding)
+        if admins:
+            res['administrative_regions'] = admins
+        return res
 
 class PoiField(fields.Raw):
     def output(self, key, obj):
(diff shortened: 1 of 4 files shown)


kinnou02:
Jenkins, test this please.

source/jormungandr/jormungandr/autocomplete/geocodejson.py
```diff
@@ -127,12 +127,12 @@ def get_lon_lat(obj):
...
 
 
-def create_address_field(geocoding):
+def create_address_field(geocoding, poi_lat, poi_lon):
     if not geocoding:
         return None
     coord = geocoding.get('coord', {})
-    lat = str(coord.get('lat')) if coord and coord.get('lat') else None
-    lon = str(coord.get('lon')) if coord and coord.get('lon') else None
+    lat = str(coord.get('lat')) if coord and coord.get('lat') else poi_lon
```
kinnou02: You mix lat and lon here

source/jormungandr/jormungandr/autocomplete/geocodejson.py
```diff
@@ -217,7 +218,7 @@ def output(self, key, obj):
             "administrative_regions":
                 create_administrative_regions_field(geocoding) or create_admin_field(geocoding),
             "properties": {p.get('key'): p.get('value') for p in geocoding.get("properties", [])},
-            "address": create_address_field(geocoding.get("address"))
+            "address": create_address_field(geocoding.get("address"), lat, lon)
```
kinnou02: I might be paranoiac, but I will prefer that we use named parameter for lat and lon, as: 
```
create_address_field(geocoding.get("address"), lat=lat, lon=lon)
```

source/jormungandr/jormungandr/autocomplete/geocodejson.py
```diff
@@ -349,10 +350,47 @@ def _check_response(cls, response, uri):
             raise TechnicalError('error in autocomplete request')
 
     @classmethod
-    def response_marshaler(cls, response_bragi, uri=None):
+    def _clean_response(cls, response, depth=3):
+
+        def _clear_object(obj, dep):
+            if dep == -1:
```
kinnou02: Since we are already checking the `depth` before calling this method we don't have to do it another time

source/jormungandr/jormungandr/autocomplete/geocodejson.py
```diff
@@ -349,10 +350,47 @@ def _check_response(cls, response, uri):
...
+                if isinstance(obj, list):
+                    del obj[:]
+                elif isinstance(obj, dict):
+                    obj.clear()
+
+        def _manage_depth(_key, _value, _depth):
+
+            if _depth == -1:
+                _clear_object(_value, _depth)
+            elif _key == 'administrative_regions':
```
kinnou02: Why? a comment might help

source/jormungandr/jormungandr/autocomplete/geocodejson.py
```diff
@@ -349,10 +350,47 @@ def _check_response(cls, response, uri):
...
+                    if _depth == -1:
+                        _clear_object(v, _depth)
+                    else:
+                        _manage_depth(k, v, _depth-1)
+
+        features = response.get('features')
+        if features:
+            for feature in features:
+                key = 'geocoding'
+                value = feature.get('properties').get('geocoding')
```
kinnou02: `properties` or `geocoding` might be None

kinnou02 (changes requested):

kadhikari:
@kinnou02 could you have a look please?

source/jormungandr/jormungandr/autocomplete/geocodejson.py
```diff
@@ -127,12 +127,12 @@ def get_lon_lat(obj):
     return lon, lat
 
 
-def create_address_field(geocoding):
+def create_address_field(geocoding, poi_lat, poi_lon):
```
pbougue: Why not populate `geocoding.address.coord`before (if empty) instead of adding params?
If we keep those params, default them to None might be a bit better.

source/jormungandr/jormungandr/autocomplete/geocodejson.py
```diff
@@ -418,8 +460,9 @@ def get(self, request, instances):
             method = requests.post
 
         raw_response = self.call_bragi(url, method, **kwargs)
+        depth = request.get('depth', 3)
```
pbougue: Default depth is 1 (so I guess this should be the default value everywhere)

source/jormungandr/jormungandr/autocomplete/geocodejson.py
```diff
@@ -349,10 +350,49 @@ def _check_response(cls, response, uri):
...
+                key = 'geocoding'
+                value = feature.get('properties', {}).get('geocoding')
+                if not value:
+                    continue
+                _manage_depth(key, value, depth)
+
+        return response
+
+    @classmethod
+    def response_marshaler(cls, response_bragi, uri=None, depth=3):
```
pbougue: depth=1 by default

source/jormungandr/jormungandr/autocomplete/geocodejson.py
```diff
@@ -349,10 +350,49 @@ def _check_response(cls, response, uri):
             raise TechnicalError('error in autocomplete request')
 
     @classmethod
-    def response_marshaler(cls, response_bragi, uri=None):
+    def _clean_response(cls, response, depth=3):
```
pbougue: depth=1 by default

source/jormungandr/jormungandr/autocomplete/geocodejson.py
```diff
@@ -349,10 +350,49 @@ def _check_response(cls, response, uri):
...
+
+        def _clear_object(obj):
```
pbougue: Minor: might even be in utils

source/jormungandr/jormungandr/autocomplete/geocodejson.py
```diff
@@ -349,10 +350,49 @@ def _check_response(cls, response, uri):
...
+            if isinstance(obj, list):
+                del obj[:]
+            elif isinstance(obj, dict):
+                obj.clear()
+
+        def _manage_depth(_key, _value, _depth):
```
pbougue, line 395: This feels too generic, more than it is in kraken...
This code feels to clean :stuck_out_tongue_closed_eyes: 
Let's just compare results obtained in jormun using bragi vs. those using kraken (that's what you're supposed to do in tests)

source/jormungandr/jormungandr/interfaces/v1/serializer/geocode_json.py
```diff
@@ -169,7 +169,8 @@ def get_address(self, obj):
         address = obj.get('properties', {}).get('geocoding', {}).get('address', None)
         if not address:
             return None
-        return create_address_field(address)
+        poi_lon, poi_lat = get_lon_lat(obj)
+        return create_address_field(address, poi_lat, poi_lon)
```
TeXitoi: Same here, something like `return create_address_field(address, lat=poi_lat, lon=poi_lon)`?

source/jormungandr/tests/bragi_autocomplete_tests.py
```diff
@@ -279,7 +399,8 @@ def test_autocomplete_call(self):
             'q': u'bob',
             'type[]': [u'public_transport:stop_area', u'street', u'house', u'poi', u'city'],
             'limit': 10,
-            'pt_dataset': 'main_routing_test'
+            'pt_dataset': 'main_routing_test',
+            'depth': 1
```
pbougue: This should not be necessary as it's default value

source/jormungandr/tests/bragi_autocomplete_tests.py
```diff
@@ -323,7 +444,8 @@ def test_autocomplete_call_override(self):
...
```
pbougue: This should not be necessary as it's default value

source/jormungandr/tests/bragi_autocomplete_tests.py
```diff
@@ -509,6 +631,149 @@ def test_stop_area_without_modes(self):
...
             assert 'physical_modes' not in r[0]['stop_area']
 
+    def test_autocomplete_call_with_depth_1(self):
+        url = 'https://host_of_bragi/autocomplete'
+        params = {
+            'q': u'bobette',
+            'type[]': [u'public_transport:stop_area', u'street', u'house', u'poi', u'city'],
+            'limit': 10,
+            'pt_dataset': 'main_routing_test',
+            'depth': 1
```
pbougue: Nice to fix it in one test, though (please let that one ;) )

source/jormungandr/tests/bragi_autocomplete_tests.py
```diff
@@ -674,131 +940,20 @@ def test_journey_with_external_uri_from_bragi(self):
...
-                    "type": "Feature"
-                }
-            ]
-        }
-
         args = {
             u'pt_dataset': 'main_routing_test',
             u'type[]': [u'public_transport:stop_area', u'street', u'house', u'poi', u'city'],
             u'limit': 10,
+            u'depth': 1,
```
pbougue: This should not be necessary as it's default value

source/jormungandr/tests/bragi_autocomplete_tests.py
```diff
@@ -825,11 +980,13 @@ def get_autocomplete(query):
                 assert response_from['poi']['properties']["amenity"] == "bicycle_rental"
                 assert response_from['poi']['properties']["capacity"] == "20"
                 assert response_from['poi']['properties']["ref"] == "12"
-                assert response_from['poi']['address']['id'] == "179077655"
+                assert response_from['poi']['address']['id'] == "addr:8.9809147;42.561667"
```
pbougue: This is not a correct address id: you don't get a prefix from kraken autocomplete
This was also part of the JIRA comment (the minor part of the 3)

source/jormungandr/tests/bragi_autocomplete_tests.py
```diff
@@ -509,6 +631,149 @@ def test_stop_area_without_modes(self):
...
+            assert poi_admins[0]['id'] == "admin:fr:02000"
+            assert poi_admins[0]['insee'] == "02000"
+            assert poi_admins[0]['label'] == "Bobtown (02000)"
+            assert poi_admins[0]['coord']['lat'] == "48.8396154"
+            assert poi_admins[0]['coord']['lon'] == "2.3957517"
+
+            address = poi['address']
+            assert address['coord']['lat'] == '8.98312e-05'
+            assert address['coord']['lon'] == '8.98312e-05'
+            assert address['id'] == "addr:8.9809147;42.561667"
```
pbougue: wrong id

source/jormungandr/tests/bragi_autocomplete_tests.py
```diff
@@ -509,6 +631,149 @@ def test_stop_area_without_modes(self):
...
```
pbougue: wrong id

source/jormungandr/tests/bragi_autocomplete_tests.py
```diff
@@ -509,6 +631,149 @@ def test_stop_area_without_modes(self):
...
```
pbougue: wrong id

pbougue (changes requested):

source/jormungandr/tests/bragi_autocomplete_tests.py
kadhikari: This parameter is needed for the Mock.

source/jormungandr/jormungandr/autocomplete/geocodejson.py
kadhikari: I think we parse and modify two times if i modify the code as you suggest?

kadhikari:
@pbougue i pushed a version with a complete verification of queries with depth = 0, 1, 2 and 3

source/jormungandr/tests/bragi_autocomplete_tests.py
```diff
@@ -509,6 +633,183 @@ def test_stop_area_without_modes(self):
...
+            poi_admins = poi['administrative_regions']
+            assert len(poi_admins) == 1
+            assert poi_admins[0]['id'] == "admin:fr:02000"
+            assert poi_admins[0]['insee'] == "02000"
+            assert poi_admins[0]['label'] == "Bobtown (02000)"
+            assert poi_admins[0]['coord']['lat'] == "48.8396154"
+            assert poi_admins[0]['coord']['lon'] == "2.3957517"
+
+            address = poi['address']
+            assert address['coord']['lat'] == '8.98312e-05'
```
pbougue: Please have different values in lat/lon so we know if we mixed them (same everywhere)

source/jormungandr/tests/bragi_autocomplete_tests.py
```diff
@@ -509,6 +633,183 @@ def test_stop_area_without_modes(self):
...
+            assert address['coord']['lon'] == '8.98312e-05'
+            assert address['id'] == "8.98312e-05;8.98312e-05"
```
pbougue: same: lat != lon

pbougue (commented):
This also has to be hand-tested (I can give you my conf)

kinnou02 (approved):

source/jormungandr/jormungandr/autocomplete/geocodejson.py
```diff
@@ -93,7 +93,7 @@ def create_administrative_regions_field(geocoding):
         zip_codes = admin.get('zip_codes', [])
         response.append({
             "insee": admin.get('insee'),
-            "name": admin.get('label'),
+            "name": admin.get('name'),
```
pbougue, line 96: :+1:

source/jormungandr/jormungandr/interfaces/v1/test/elastic_search_tests.py
```diff
@@ -63,6 +63,7 @@ def bragi_house_jaures_feature():
                         "insee": "02000",
                         "level": 8,
                         "label": "Saint-Quentin",
```
pbougue: It would be nicer to have realistic labels on the ones tested (on all of them would be even nicer).
If not realistic, at least different from name, as mixing them was a bug before.

source/jormungandr/jormungandr/autocomplete/geocodejson.py
```diff
@@ -349,10 +357,53 @@ def _check_response(cls, response, uri):
...
+            elif isinstance(obj, dict):
+                obj.clear()
+
+        def _manage_depth(_key, _value, _depth):
+
+            if is_deleteable(_key, _value, _depth):
+                _clear_object(_value)
+            elif isinstance(_value, dict):
+                for k, v in _value.items():
+                    if is_deleteable(k, v, _depth):
```
pbougue: Isn't that part already managed 2 lines below (as you call it with depth-1) ?

pbougue (approved):
This looks good.
What worries me is that the cleaning impacts every objects, not just poi.address.administrative_regions...So we need to test it for all depth, and all other features (stops, address, admin, streets), at least manually.

TeXitoi (approved):

source/jormungandr/jormungandr/autocomplete/geocodejson.py
```diff
@@ -127,15 +127,15 @@ def get_lon_lat(obj):
...
-    lat = str(coord.get('lat')) if coord and coord.get('lat') else None
-    lon = str(coord.get('lon')) if coord and coord.get('lon') else None
-
-    return {
-        "id": geocoding.get('id'),
+    lat = str(coord.get('lat')) if coord and coord.get('lat') else poi_lat
+    lon = str(coord.get('lon')) if coord and coord.get('lon') else poi_lon
+    address_id = '{lon};{lat}'.format(lon=lon, lat=lat)
+    resp = {
+        "id": address_id,
```
benoit-bst, line 144: Why the Poi address ID is now long;lat ? It doesn't exist an associated ID ?

benoit-bst (approved):

source/jormungandr/jormungandr/autocomplete/geocodejson.py
TeXitoi: address ids are coords. If we don't have reverse geocoding, an empty address can be generated with such an id.

benoit-bst: :+1:

PR #2343, opened 2018-02-28
ed/gtfs2ed : Add safety in row check (carriage return case)

benoit-bst:


benoit-bst:
https://jira.canaltp.fr/browse/NAVITIAII-2375

source/ed/connectors/gtfs_parser.cpp
```diff
@@ -639,8 +639,6 @@ void CalendarGtfsHandler::finish(Data& data) {
 }
 
 void CalendarGtfsHandler::handle_line(Data& data, const csv_row& row, bool) {
-    if (row.empty())
```
TeXitoi, line 642: There is another bunch of `row.empty()` in the file that would need some cleanups.

TeXitoi (changes requested):

source/ed/connectors/gtfs_parser.cpp
benoit-bst: :+1:

TeXitoi (approved):

TeXitoi (changes requested):
Please add a test:
 - [x]  add empty lines in a fixture/ed/gtfs???
 - [x]  check that it fails in docker_test
 - [x]  check that it's fixed in docker_test with the fix.

fixtures/ed/gtfs/calendar_dates.txt
```diff
@@ -1,2 +1,4 @@
-service_id,date,exception_type
-FULLW,20110101,2
+service_id,date,exception_type
+
```
benoit-bst, line 2: We don't see but this is \r\n

source/ed/tests/gtfsparser_test.cpp
```diff
@@ -98,6 +98,17 @@ BOOST_AUTO_TEST_CASE(parse_agencies) {
...
 }
 
+BOOST_AUTO_TEST_CASE(parse_gtfs_file_with_empty_line_carriage_return) {
+    using file_parser = ed::connectors::FileParser<ed::connectors::CalendarDatesGtfsHandler>;
+    {
+        // in fixtures/ed/gtfs, calendar_dates.txt contains empty lines and carriage return (\r\n)
+        // to test different CSV parser case.
+        ed::Data data;
+        ed::connectors::GtfsParser parser(std::string(navitia::config::fixtures_dir) + gtfs_path);
+        file_parser(parser.gtfs_data, std::string(navitia::config::fixtures_dir) + gtfs_path + "/calendar_dates.txt").fill(data);
```
pbougue: line length

pbougue (commented):
tests! :tada: 
Looks good but needs a rebase
Also a test on ntfs would be nice (search one that is actually read by docker_test)

pbougue (approved):

TeXitoi (approved):

PR #2344, opened 2018-03-01
Ed : executables crashe when run with no parameter

nonifier:
Fixes osm2ed when run without parameter


nonifier:
Currently looking at the other executables that might have the same issue.

source/ed/tests/osm2ed_test.cpp
```diff
@@ -0,0 +1,47 @@
...
+struct logger_initialized {
+    logger_initialized()   { init_logger(); }
+};
+
+BOOST_GLOBAL_FIXTURE( logger_initialized )
+
+BOOST_AUTO_TEST_CASE(osm2ed_with_no_param_should_start_without_throwing) 
+{
+    const char* argv [] = { "osm2ed_test" };
+    BOOST_CHECK_NO_THROW(ed::connectors::osm2ed(1, argv));
```
xlqian, line 46: maybe test the returned value?

source/ed/osm2ed_main.cpp
```diff
@@ -0,0 +1,35 @@
...
+IRC #navitia on freenode
+https://groups.google.com/d/forum/navitia
+www.navitia.io
+*/
+
+#include "osm2ed.h"
+
+int main(int argc, const char** argv) {
+    return ed::connectors::osm2ed(argc, argv);
+}
```
xlqian: newline plz :)

source/ed/tests/osm2ed_test.cpp
nonifier: The behavior I intent to fix here, is the throw of a `boost::throw_exception<boost::program_options::required_option>` that triggers a signal abort from `po::notify(vm);`, hence the `BOOST_CHECK_NO_THROW`.

The return value could be tested indeed, but it depends on the command parameters passed in and varies from one another. I'm not sure I want to enforce those behaviors here as it's not relevant to the fix.

source/ed/osm2ed_main.cpp
nonifier: Added in cc7a068320fa0493468adccc2f634394c667a370 :)
But are we not using C++11 feature set ? I thought new lines were not a requirement anymore for end of files.

kinnou02 (approved):

source/ed/osm2ed_main.cpp
pbougue: We are, but it's more elegant when github doesn't warn :)

pbougue (commented):
It's worth having the other exe managed in the same PR

source/ed/osm2ed_main.cpp
nonifier: got it :+1:

nonifier:
Alright! I've justed pushed similar changes for `fare2ed` and `ed2nav` :)
@xlqian & @pbougue : would you mind having a final look at the new changes pliz ?

source/ed/tests/osm2ed_test.cpp
xlqian: ok!

xlqian (approved):

source/ed/ed2nav.cpp
```diff
@@ -273,3 +274,5 @@ int main(int argc, char * argv[])
 
     return 0;
 }
+
+} // namespace ed
```
pbougue: minor: newline

source/ed/fare2ed.cpp
```diff
@@ -114,3 +113,6 @@ int main(int argc, char * argv[])
 
     return 0;
 }
+
+} // namespace ed
+
```
pbougue: minor: isn't there an extra_line? :)

pbougue (approved):

benoit-bst (approved):

PR #2353, opened 2018-03-06
kraken: Exclude way without edge from the autocomplete

kinnou02:
We may have some Way that don't have any edges because we removed small unconnected connexes component.
When we return such way their coordinates are (0,0) and they don't have id

### DIFF
--- a/source/ed/build_helper.h
+++ b/source/ed/build_helper.h
@@ -226,6 +226,7 @@ struct builder {
     std::unique_ptr<navitia::type::Data> data = std::make_unique<navitia::type::Data>();
     navitia::georef::GeoRef street_network;
 
+    //lazy init of vertexes used when creating ways with add_way
     boost::optional<navitia::georef::vertex_t> vertex_a;
     boost::optional<navitia::georef::vertex_t> vertex_b;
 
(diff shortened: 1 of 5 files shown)


source/ed/build_helper.h
```diff
@@ -224,6 +226,9 @@ struct builder {
     std::unique_ptr<navitia::type::Data> data = std::make_unique<navitia::type::Data>();
     navitia::georef::GeoRef street_network;
 
+    boost::optional<navitia::georef::vertex_t> vertex_a;
```
pbougue: please comment the use of it (and if it's too much of a pain, remove it ;) )

source/ed/build_helper.h
```diff
@@ -302,6 +307,8 @@ struct builder {
     void manage_admin();
     void build_autocomplete();
     void fill_missing_destinations();
+
+    navitia::georef::Way* add_way(const std::string& name, const std::string& way_type);
```
pbougue, line 312: :+1: :tada: nice!

pbougue (commented):
Nice cleanup, you unlocked the @mmartinet - :medal_sports: !
But please, can you add an actual test of what you did? :stuck_out_tongue_closed_eyes:

TeXitoi (approved):

pbougue (approved):

PR #2361, opened 2018-03-13
tyr: create a new job to manage data injection towards mimir

kadhikari:
- Ticket: https://jira.canaltp.fr/browse/NAVP-813

- For a data_set with family_type = 'pt' and an instance with  export to mimir activated, we create a new job and a data_set with family_type = 'mimir', and name = file path of source job (job with 'pt'). 
    
- In the function last_datasets the data_set with family_type = 'mimir' is excluded for backward compatibility of other apis. The function last_datasets is used many times.

- in the api "jobs" all the jobs and data_sets are listed with status of each job.

TODO:
- [x]  wait for dev deploy to be up

### DIFF
--- a/source/tyr/tests/integration/dataset_test.py
+++ b/source/tyr/tests/integration/dataset_test.py
@@ -36,7 +36,6 @@ def create_basic_job_with_data_sets():
         models.db.session.commit()
 
 
-#@pytest.fixture
 def add_job_with_data_set_mimir(create_basic_job_with_data_sets):
     with app.app_context():
         # we also create 1 job with a dataset for mimir
--- a/source/tyr/tests/integration/dataset_test.py
+++ b/source/tyr/tests/integration/dataset_test.py
@@ -1,0 +1,132 @@
+from __future__ import absolute_import, print_function, unicode_literals, division
+from tests.check_utils import api_get
+import pytest
+from navitiacommon import models
+from tyr import app
+
+@pytest.fixture
+def create_basic_job_with_data_sets():
+    with app.app_context():
+        instance = models.Instance('fr')
+        models.db.session.add(instance)
+        models.db.session.commit()
+
+        job = models.Job()
+        job.instance = instance
+
+        # we also create 2 datasets, one for fusio, one for synonym
+        for i, dset_type in enumerate(['fusio', 'synonym']):
+            dataset = models.DataSet()
+            dataset.type = dset_type
+            dataset.family_type = dataset.type
+            if dataset.type == 'fusio':
+                dataset.family_type = 'pt'
+            dataset.name = '/path/to/dataset_{}'.format(i)
+            models.db.session.add(dataset)
+
+            job.data_sets.append(dataset)
+
+        job.state = 'done'
+        models.db.session.add(job)
+        models.db.session.commit()
+
+
+@pytest.fixture
+def create_jobs_with_three_data_sets():
+    with app.app_context():
+        instance = models.Instance('fr')
+        models.db.session.add(instance)
+        models.db.session.commit()
+
+        job = models.Job()
+        job.instance = instance
+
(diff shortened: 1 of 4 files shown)


source/tyr/tyr/binarisation.py
```diff
@@ -606,6 +606,8 @@ def stops2mimir(self, instance_config, input, job_id=None, dataset_uid=None):
             # stops2mimir have to be non-blocking.
             # @TODO : Find a way to raise error without breaking celery tasks chain
             logger.error('stops2mimir failed')
+            if job_id:
+                job.state = 'failed'
```
kinnou02, line 610: change will be rollbacked if you do not commit them.

> :heavy_check_mark: Done

kinnou02 (changes requested):

source/tyr/tests/integration/dataset_test.py
```diff
@@ -0,0 +1,132 @@
...
+
+            job.data_sets.append(dataset)
+
+        job.state = 'done'
+        models.db.session.add(job)
+        models.db.session.commit()
+
+
+@pytest.fixture
+def create_jobs_with_three_data_sets():
```
nonifier: There is a lot of duplicated code from the test above. Could this be re-factored please ?

source/tyr/tests/integration/dataset_test.py
```diff
@@ -0,0 +1,132 @@
...
+
+    """
+    we query the loaded datasets of fr
+    we loaded 3 datasets, 1 in a job with two data_sets (pt and synonym)
+    another job with a data_set having family_type 'mimir'
+    In last_datasets we return a data_set per family_type except 'mimir'.
+    """
+    resp = api_get('/v0/instances/fr/last_datasets')
+
+    assert len(resp) == 2
```
nonifier: Same here regarding code duplication. Most of the asserts are already done above and should not be needed here.

source/tyr/tests/integration/dataset_test.py
```diff
@@ -0,0 +1,132 @@
...
+        models.db.session.commit()
+
+
+def test_basic_datasets(create_basic_job_with_data_sets):
+    """
+    we query the loaded datasets of fr
+    we loaded 2 datasets, one for pt, one for synonym
+    """
+    resp = api_get('/v0/instances/fr/last_datasets')
+
```
mbounabi, line 64: What is the use of this test? There are already tests about last datasets.
Also, as mentionned by @nonifier , the next test does almost the same thing at the beginning

source/tyr/tyr/tasks.py
```diff
@@ -145,6 +144,52 @@ def import_data(files, instance, backup_file, async=True, reload=True, custom_ou
...
+    #currently the name of a dataset is the path to it
+    dataset.name = filename
+    models.db.session.add(dataset)
+    job.data_sets.append(dataset)
+
+    models.db.session.add(job)
+    models.db.session.commit()
+
+    # We pass the job id to each tasks, but job need to be commited for having an id
+    for action in actions:
```
mbounabi: At this point, the list 'actions' is empty, isn't it?

source/tyr/tests/integration/dataset_test.py
kadhikari: As already discussed with @mbounabi, here we test the result once with one job having two data_sets and then after adding a new job with a data_set with family_type 'mimir'. We should have the same result as data_set with family_type 'mimir' is filtered in this pull request.

kadhikari:
jenkins test this please

mbounabi (approved):

nonifier (approved):
Looking good :)

pbougue (approved):

kinnou02 (approved):

PR #2362, opened 2018-03-13
Ed: Poi id in Kraken as in Mimir

nonifier:
POI's ID names were different from Kraken to Mimir when using auto-completion. This PR is meant to fix it.


source/ed/tests/osm2ed_test.cpp
```diff
@@ -40,8 +40,34 @@ struct logger_initialized {
...
-    BOOST_CHECK_NO_THROW(ed::connectors::osm2ed(1, argv));
+    BOOST_CHECK_NO_THROW(osm2ed(1, argv));
+}
+
+BOOST_AUTO_TEST_CASE(poi_id_node_should_match_mimirs_one) 
+{
+    BOOST_CHECK(
+        boost::contains(
+            construct_poi_id(OsmObjectType::Node, 123456), 
+            "poi:osm:node:"));
```
TeXitoi: You should test the exact result

source/ed/tests/osm2ed_test.cpp
```diff
@@ -40,8 +40,34 @@ struct logger_initialized {
...
+}
+
+BOOST_AUTO_TEST_CASE(poi_id_way_should_match_mimirs_one) 
+{
+    BOOST_CHECK(
+        boost::contains(
+            construct_poi_id(OsmObjectType::Way, 123456), 
+            "poi:osm:way:"));
```
TeXitoi: same

source/ed/tests/osm2ed_test.cpp
```diff
@@ -40,8 +40,34 @@ struct logger_initialized {
...
+}
+
+BOOST_AUTO_TEST_CASE(poi_id_relation_should_match_mimirs_one) 
+{
+    BOOST_CHECK(
+        boost::contains(
+            construct_poi_id(OsmObjectType::Relation, 123456), 
+            "poi:osm:relation:"));
```
TeXitoi: same

TeXitoi (changes requested):
Great!

source/ed/ed_persistor.cpp
```diff
@@ -251,11 +251,17 @@ void EdPersistor::insert_pois(const Georef &data) {
...
-                std::to_string(itm.second.weight),
-                this->to_geographic_point(itm.second.coord),
-                itm.second.name, "poi:" + itm.first, poi_type, std::to_string(itm.second.visible),
-                itm.second.address_number, itm.second.address_name});
+        this->lotus.insert({
+                std::to_string(itm.second.id), // id
+                std::to_string(itm.second.weight), // weight
+                this->to_geographic_point(itm.second.coord), // coord
+                itm.second.name, // name 
+                itm.first, // uri
```
kinnou02: Poi imported with poi2ed won't be prefixed by `poi:` anymore.

kinnou02 (changes requested):

source/ed/ed_persistor.cpp
nonifier: Now fixed in 6543acd. 
Do you know of any other dependence by any chance? I can't see any other. 
Also, do you know of any `poi.txt` file I can write tests against for `poi2ed` ?

source/ed/docker_tests/ed_integration_tests.cpp
```diff
@@ -388,3 +388,24 @@ BOOST_FIXTURE_TEST_CASE(ntfs_v5_test, ArgsFixture) {
...
+
+    for(const auto & poi : pois) {
+        BOOST_CHECK(
+            boost::starts_with(poi->uri, "poi:osm:")
+        );
+        BOOST_CHECK(
+            boost::contains(poi->uri, "node") 
+            || boost::contains(poi->uri, "way") 
+            || boost::contains(poi->uri, "relation") 
+        );
```
xlqian: minor: we won't be able to identify which one goes wrong when the check fails :)

source/ed/osm2ed.h
```diff
@@ -372,18 +372,34 @@ struct Rect{
...
-    case OsmObjectType::Way: return "w";
+    case OsmObjectType::Node : return "node";
+    case OsmObjectType::Way : return "way";
+    case OsmObjectType::Relation : return "relation";
     default:
         throw navitia::exception("not implemented");
     }
 }
 
+class OsmPoi : public ed::types::Poi
```
xlqian, line 389: if i understand it right, you use the class only for constructing the id... maybe a free function can do the job and it'll be simpler

TeXitoi (approved):

kinnou02 (approved):
I agree with @xlqian that it's a bit complex to create uris

source/ed/docker_tests/ed_integration_tests.cpp
```diff
@@ -388,3 +388,16 @@ BOOST_FIXTURE_TEST_CASE(ntfs_v5_test, ArgsFixture) {
 
     check_ntfs(data);
 }
+
+BOOST_FIXTURE_TEST_CASE(osm_pois_id_sould_match_mimir_naming, ArgsFixture)
```
pbougue: typo : s`h`ould

source/ed/osm2ed.cpp
```diff
@@ -856,7 +855,7 @@ const OSMWay* PoiHouseNumberVisitor::find_way(const CanalTP::Tags& tags, const d
                 min_distance = distance;
                 candidate_way = &way;
             }
-        }
+        }
```
pbougue: useless whitespace

source/ed/types.h
```diff
@@ -437,8 +437,13 @@ struct Poi {
     std::map<std::string, std::string> properties;
     std::string address_number = "";
     std::string address_name = "";
+    std::string uri = "poi:";
```
pbougue, line 440: minor: This prefix is a bit hidden elsewhere (OsmPoi, builder)... If you have an idea (naming, refactor, comment) to avoid futur problems (understanding, correcting for one bug, but breaking the other usecases...) it could be nice, but it's minor.

source/ed/osm2ed.h
pbougue: Maybe then use the constructor from parent class?
Although, the typing can help not messing OSM and custom POI use/creation...

pbougue (commented):
Not sure it's tested that nothing else changed in custom POI naming.

xlqian (approved):

source/ed/osm2ed.h
nonifier: @xlqian - I started with a free function, but we have 2 different behaviors with different parameter types for `Poi` and `OsmPoi`. The problem is that we also have some common logic between the 2 objects (both `uri`s start with `"poi:"`, hence the class design.
@pbougue - I do use the parent constructor. When you create an `OsmPoi`, `Poi::uri` is being set through the default constructor (eg `Poi::Poi() = default;`).

nonifier:
Just to let you guys know, integration tests for poi2ed will follow in another PR not to block this one.

nonifier:
Oupsy.... I just closed the PR by mistake :sob: - Now reopened !

benoit-bst (approved):

TeXitoi (approved):

pbougue (approved):

PR #2368, opened 2018-03-20
Jormungandr: add speed in asgard

TeXitoi:


kinnou02 (approved):

source/jormungandr/jormungandr/street_network/asgard.py
```diff
@@ -48,11 +48,12 @@ def __init__(self, instance, service_url, asgard_socket, modes=[], id='asgard',
         self._sockets = queue.Queue()
 
     def get_street_network_routing_matrix(self, origins, destinations, mode, max_duration, request, **kwargs):
-        if len(origins) > 1:
```
pbougue, line 51: I don't see the point of deleting this code : we can handle n*m and we are comfortable letting responsibility to the caller?

xlqian (approved):

source/jormungandr/jormungandr/street_network/asgard.py
TeXitoi: Yes, Asgard can do NxM as Valhalla. Valhalla doesn't do this check, only Kraken do as it can't do NxM.

nonifier: Just to educate an ignorant, what's the difference between Asgard and Valhalla ? (apart from the fact that Valhalla is a hall located in Asgard :D )

TeXitoi: Valhalla is a lib + some services to do routing. Asgard is a service that use the valhalla libs to produce an optimized service for our usecase.

TeXitoi: Asgard is a POC for the moment.

nonifier: got it :+1:

benoit-bst (approved):

pbougue (approved):

PR #2371, opened 2018-03-22
ptref: add a link between PhysicalMode and JourneyPattern in the graph

TeXitoi:


pbougue (approved):

nonifier:
It would be really nice to have a test for this once the situation calms down :)

TeXitoi:
@nonifier as @pbougue said on slack, no functional change, and the functional impacted by this already exists.
> only perf change, and we hope that functionnal regressions that might appear are tested elsewhere, because that's linked to other functionnalities. We miss perf tests...

kinnou02 (approved):

PR #2372, opened 2018-03-22
Ed: Poi2ed integration test

nonifier:


source/CMakeLists.txt
```diff
@@ -69,12 +69,30 @@ include_directories("${CMAKE_SOURCE_DIR}"
 )
 link_directories(${Boost_LIBRARY_DIRS})
 
-MACRO(ADD_BOOST_TEST EXE_NAME)
+SET(TEST_DEBUG_OUTPUT OFF CACHE BOOL "Shoot test output to stdin")
+
+MACRO(BUILD_BOOST_TEST_ARGS EXE_NAME)
```
benoit-bst, line 74: Just an idea : Maybe it's cleaner to move cmake macro functions in a special directory like, **root_project/cmake** ?
We can use this directory for others functions as well.

benoit-bst (approved):

source/CMakeLists.txt
nonifier: That's a very good point! If we keep on adding macros here, it would make sense to refactor it your way ;) Let's keep that in mind.

benoit-bst: :+1:

pbougue, line 74: :+1:

source/CMakeLists.txt
```diff
@@ -69,12 +69,30 @@ include_directories("${CMAKE_SOURCE_DIR}"
 )
 link_directories(${Boost_LIBRARY_DIRS})
 
-MACRO(ADD_BOOST_TEST EXE_NAME)
+SET(TEST_DEBUG_OUTPUT OFF CACHE BOOL "Shoot test output to stdin")
```
pbougue, line 72: :+1:

pbougue (commented):
It looks like it breaks docker tests while reading osm from sao-tome on my laptop.
Dev's docker_test is broken also (maybe a problem on my side?)... Checking it for release.
Appart from that, seems all good.

pbougue (approved):

PR #2374, opened 2018-03-27
Jormun: slight modification to timeo to make it easy for gomeo

kinnou02:
Timeo use a non standard way of formatting parameters: `StopDescription`
is an "object" that contains params between the `?` and `;`. As you can
guess there is not a single implementation of something like this.
Standard implementation will consider that the value of
`StopDescription` end after the next `&`, so we "loose" the first
params.
This PR put `StopTimeType` in first position since it's the less useful
parameter of all, since for us it is obvious that we call timeo to get
realtime data. Doing so make it very easy for [goemo](https://github.com/kinnou02/gomeo)
to parse the query.

### DIFF
--- a/source/jormungandr/jormungandr/realtime_schedule/timeo.py
+++ b/source/jormungandr/jormungandr/realtime_schedule/timeo.py
@@ -208,12 +208,14 @@ class Timeo(RealtimeProxy):
             .format(dt=self._timestamp_to_date(from_dt).strftime('%Y-%m-%dT%H:%M:%S')) \
             if from_dt else ''
 
+        #We want to have StopTimeType as it make parsing of the request way easier
+        #for alternative implementation of timeo since w<e can ignore this params
         stop_id_url = ("StopDescription=?"
-                       "StopTimeoCode={stop}"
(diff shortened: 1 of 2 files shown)


source/jormungandr/jormungandr/realtime_schedule/timeo.py
```diff
@@ -208,12 +208,14 @@ def _make_url(self, route_point, count=None, from_dt=None):
             .format(dt=self._timestamp_to_date(from_dt).strftime('%Y-%m-%dT%H:%M:%S')) \
             if from_dt else ''
 
+        #We want to have StopTimeType as it make parsing of the request way easier
+        #for alternative implementation of timeo since w<e can ignore this params
```
TeXitoi: typo on "we"

kinnou02:
Tests will fail, there is a mock that use the generated URL as key, I change this asap

kinnou02:
Jenkins, test this please.

TeXitoi (approved):

pbougue (approved):

PR #2380, opened 2018-03-30
Jormungandr : Remove Marshall

mbounabi:
Il y a des commentaires 'TODO : [NAVP-805]' avec des questions.
La plupart concerne l'utilisation de paramètres 'collection(s)' qui n'est pas très claire.


source/jormungandr/jormungandr/interfaces/v1/decorators.py
```diff
@@ -26,17 +26,12 @@
...
-from collections import OrderedDict
+from jormungandr.interfaces.v1.serializer import serialize_with
 
 
-def get_serializer(serpy, marshall):
-    if jormungandr.USE_SERPY:
-        return serialize_with(serpy)
-    else:
-        return marshal_with(OrderedDict(marshall), display_null=False)
+def get_serializer(serpy):
```
xlqian, line 32: it seems this function has no more utilities...

source/jormungandr/jormungandr/interfaces/v1/Calendars.py
```diff
@@ -119,8 +74,9 @@ def __init__(self):
                                 help='The datetime considered as "now". Used for debug, default is '
                                      'the moment of the request. It will mainly change the output '
                                      'of the disruptions.')
-        self.collection = 'calendars'
-        self.collections = calendars
+        # TODO : [NAVP-805] Collection(s) : useful or useless ?
+        # self.collection = 'calendars'
+        # self.collections = api.CalendarsSerializer
```
nonifier, line 79: If it's commented out, I would think this is useless, and I would rather remove that code.
If it's useful, that will come back to us anyway ;)

nonifier:
This PR is approved by this guy

Mr Propre

Issue #2367, opened 2018-03-16
[ed fusio parser] contributor_id removed in NTFS 0.7


prhod:
In the latest version of the NTFS, the trips.txt file doesn't contain a contributor_id field anymore.

https://github.com/CanalTP/navitia/blob/dev/source/ed/connectors/fusio_parser.cpp#L601
There is no read of the dataset_id if there is no contributor_id.
New NTFS files could end in error or with a wrong association.

PR #2382, opened 2018-04-03
ed : dataset_id is now required in trips.txt

benoit-bst:
Issue : https://github.com/CanalTP/navitia/issues/2367
Jira : https://jira.canaltp.fr/browse/NAVITIAII-2413

### DIFF
--- a/source/ed/docker_tests/ed_integration_tests.cpp
+++ b/source/ed/docker_tests/ed_integration_tests.cpp
@@ -378,7 +378,9 @@ BOOST_FIXTURE_TEST_CASE(ntfs_v5_test, ArgsFixture) {
             boost::gregorian::date_period("20150826"_d, "20150926"_d));
     BOOST_CHECK_EQUAL(data.pt_data->datasets[0]->realtime_level == nt::RTLevel::Base, true);
     BOOST_CHECK_EQUAL(data.pt_data->datasets[0]->system, "obiti");
-    BOOST_CHECK(data.pt_data->vehicle_journeys[0]->dataset == nullptr);
+    
+    // accepted side-effect (no link) as ntfs_v5 fixture does not contain datasets.txt, which is now required
+    BOOST_CHECK(data.pt_data->vehicle_journeys[0]->dataset == nullptr); 
 
     BOOST_REQUIRE_EQUAL(data.pt_data->networks.size(), 1);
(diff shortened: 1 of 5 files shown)


nonifier (approved):
Much nicer :D

source/ed/docker_tests/ed_integration_tests.cpp
```diff
@@ -378,7 +378,7 @@ BOOST_FIXTURE_TEST_CASE(ntfs_v5_test, ArgsFixture) {
             boost::gregorian::date_period("20150826"_d, "20150926"_d));
     BOOST_CHECK_EQUAL(data.pt_data->datasets[0]->realtime_level == nt::RTLevel::Base, true);
     BOOST_CHECK_EQUAL(data.pt_data->datasets[0]->system, "obiti");
-    BOOST_CHECK_EQUAL(data.pt_data->vehicle_journeys[0]->dataset->uri, "d1");
+    BOOST_CHECK(data.pt_data->vehicle_journeys[0]->dataset == nullptr);
```
pbougue: Can you add a comment to that, so we know that it's not an enforced behavior, it's a side-effect but never supposed to happen.
Something like:
`// as ntfs_v5 fixture does not contain datasets.txt which is now required, we shouldn't have a dataset linked to vj`
That way, if someone implements a fallback he will have no remorse changing the test.

source/ed/tests/fusioparser_test.cpp
```diff
@@ -423,7 +424,7 @@ BOOST_AUTO_TEST_CASE(sync_ntfs) {
     BOOST_CHECK_EQUAL(data.datasets[0]->validation_period, boost::gregorian::date_period("20150826"_d, "20150926"_d));
     BOOST_CHECK_EQUAL(data.datasets[0]->realtime_level == nt::RTLevel::Base, true);
     BOOST_CHECK_EQUAL(data.datasets[0]->system, "obiti");
-    BOOST_CHECK_EQUAL(data.vehicle_journeys[0]->dataset->uri, "d1");
+    BOOST_CHECK(data.vehicle_journeys[0]->dataset == nullptr);
```
pbougue: Same here, see comment in `source/ed/docker_tests/ed_integration_tests.cpp`

pbougue (approved):

TeXitoi (approved):

PR #2383, opened 2018-04-04
Tyr : Create command 'purge_old_jobs' + optim

mbounabi:
Ticket Jira : https://jira.canaltp.fr/browse/NAVITIAII-2425

### DIFF
--- a/source/navitiacommon/navitiacommon/models.py
+++ b/source/navitiacommon/navitiacommon/models.py
@@ -442,8 +442,8 @@ class Instance(db.Model):
         dataset_to_keep = self.last_datasets()
         dataset_file_to_keep = [f.name for f in dataset_to_keep]
 
-	# Keep the jobs associated        
-	jobs_to_keep = set()
+        # Keep the jobs associated
+        jobs_to_keep = set()
         for dataset in dataset_to_keep:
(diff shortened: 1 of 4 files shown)


nonifier (approved):

source/navitiacommon/navitiacommon/models.py
```diff
@@ -445,21 +445,21 @@ def delete_old_jobs_and_list_datasets(self, time_limit):
         # Keep the jobs associated
         jobs_to_keep = []
         for dataset in dataset_to_keep:
-            jobs_to_keep.append(db.session.query(Job).filter(Job.data_sets.contains(dataset)).first())
+            job_associated = db.session.query(Job).filter(Job.data_sets.contains(dataset)).first()
+            jobs_to_keep.append(job_associated) if job_associated not in jobs_to_keep else None
```
pbougue: Not sure if this is working the way you expect.
Wouldn't it be simpler to just do:
```python
if job_associated not in jobs_to_keep:
    jobs_to_keep.append(job_associated)
```

kinnou02 (approved):

source/navitiacommon/navitiacommon/models.py
mbounabi: It's the same thing in one line instead of two.

pbougue (approved):

PR #2384, opened 2018-04-04
ed: correction for a pattern with dst

kadhikari:
For a pattern with period intersecting two dsts, if the pattern is
active only one day and shift <> 0 then do not shift.

Concerned ticket: https://jira.canaltp.fr/browse/NAVITIAII-2411

Some infomations in the tickets


source/type/type.cpp
```diff
@@ -566,13 +566,16 @@ VJ* MetaVehicleJourney::impl_create_vj(const std::string& uri,
 
     // Desactivating the other vjs. The last creation has priority on
     // all the already existing vjs.
+    // We don't modify validity pattern active only once and shift other than 0
     const auto mask = ~canceled_vp.days;
     for_all_vjs([&] (VehicleJourney& vj) {
+        if (!(vj.validity_patterns[0]->days.count() == 1) && (vj.shift != 0)) {
```
pbougue: Please use enum `base_schedule` explicitly.

pbougue: What has to be implemented (for this patch) is just: don't modify a vj that has a valid base_schedule vp.
You are testing too much things here.

pbougue (commented):
As discussed, and to be explicit here: this needs testing.
> :heavy_check_mark: done

pbougue (approved):

source/ed/docker_tests/ed_integration_tests.cpp
```diff
@@ -426,3 +426,41 @@ BOOST_FIXTURE_TEST_CASE(poi2ed_pois_uri_should_match_mimir_naming, ArgsFixture)
...
+    bool failed = false;
+    try {
+        data.load_nav(input_file);
+    } catch(const navitia::data::data_loading_error&) {
+        failed = true;
+    }
+    data.build_raptor();
+    BOOST_REQUIRE_EQUAL(failed, false);
+
+    BOOST_REQUIRE_EQUAL(data.pt_data->lines.size(), 1);
```
nonifier: Why do you require `lines` and `routes` to have a size of 1 when you don't access them ? Maybe they are `BOOST_CHECK_*` instead ?

source/ed/docker_tests/ed_integration_tests.cpp
```diff
@@ -426,3 +426,41 @@ BOOST_FIXTURE_TEST_CASE(poi2ed_pois_uri_should_match_mimir_naming, ArgsFixture)
...
+    BOOST_REQUIRE_EQUAL(data.pt_data->datasets.size(), 1);
+    BOOST_REQUIRE_EQUAL(data.pt_data->contributors.size(), 1);
+    BOOST_REQUIRE_EQUAL(data.pt_data->contributors[0], data.pt_data->datasets[0]->contributor);
+    BOOST_CHECK_EQUAL(data.pt_data->datasets[0]->desc, "centre-sncf");
+    BOOST_CHECK_EQUAL(data.pt_data->datasets[0]->uri, "SCF:23");
+    BOOST_CHECK_EQUAL(data.pt_data->datasets[0]->validation_period,
+            boost::gregorian::date_period("20180319"_d, "20180617"_d));
+    BOOST_CHECK_EQUAL(data.pt_data->datasets[0]->realtime_level == nt::RTLevel::Base, true);
+    BOOST_CHECK_EQUAL(data.pt_data->datasets[0]->system, "ChouetteV2");
+    BOOST_CHECK_EQUAL(data.pt_data->vehicle_journeys[0]->dataset->uri, "SCF:23");
```
nonifier: You might need a `BOOST_REQUIRE_GT(data.pt_data->lines.size(), 1);` before this check. Otherwise, your test will crash if you have no VJ.

nonifier (changes requested):

nonifier (approved):

PR #2387, opened 2018-04-06
Kraken: Unicode parse bugfix

pbougue:
Only test for now (and a WIP commit to remember it).

The problem is the same as the one when calling:
v1/coverage/toto/stop_points?filter=stop_point.has_code(KisioDigitalis%C3%A9%2C%20KisioDigitalis_C%3AS43)
:arrow_right: this fails for parsing while the same call without accent success (or fails because no object is found)

Have to use quotes for this kind of query

### DIFF
--- a/source/jormungandr/jormungandr/instance.py
+++ b/source/jormungandr/jormungandr/instance.py
@@ -409,7 +409,6 @@ class Instance(object):
                 request.request_id = flask.request.id
             except RuntimeError:
                 # we aren't in a flask context, so there is no request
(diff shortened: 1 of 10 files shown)


nonifier:
It's a bit tricky to test Unicode encoding using `é` as this character is part of latin-1 (ISO-8859-1) and could be encoded over 8 bits (similar to ASCII). 

Maybe you could use a character encoded over 4 bytes instead, to make sure you hit an UTF-* representation. I can suggest `U+1F984` :smiley:

PR #2388, opened 2018-04-06
Jormung: Admin type not city as admin region

nonifier:
Admin with an `Administrative_region` returned by Bragi are now treated as `geocode_admin`


pbougue (approved):

benoit-bst (approved):

nonifier:
jenkins, test this please.

source/jormungandr/tests/bragi_autocomplete_tests.py
```diff
@@ -1236,21 +1205,32 @@ def test_autocomplete_for_admin_depth_zero(self):
...
+            assert r[0]['administrative_region']['label'] == 'Créteil (94000)'
+            assert r[0]['administrative_region']['id'] == 'admin:fr:941'
+            assert r[0]['administrative_region']['zip_code'] == '94000'
+
+    def test_autocomplete_for_administrative_region_with_wrong_type(self):
+        mock_requests = mock_bragi_autocomplete_call(BRAGI_MOCK_ADMINISTRATIVE_REGION_WITH_WRONG_TYPE)
+        with mock.patch('requests.get', mock_requests.get):
+            response = self.query_region("places?q=bob")
+            r = response.get('places')
+            assert (r is None) or len(r) == 0
```
pbougue: So depending on the serialiser, you either have no 'places' or an empty one? Not sure it's what we want... We probably want the serpy to serialise places even if empty, if it's the case for marshall.

nonifier: I do agree with you. I've changed the `display_none` for `GeocodePlacesSerializer` so that serpy mimics marshal's behavior (cf 005c899).

PR #2389, opened 2018-04-06
Doc: Document autocompletion setup for Bragi & Jormung

nonifier:


source/jormungandr/readme.md
```diff
@@ -80,6 +80,21 @@ From `navitia/source/jormungandr/jormungandr` run :
...
+# Option
+## Autocomplete
+
+Setup Jormung for using either `Bragi` or `Kraken` for autocompletion
+
+```sh
+JORMUNGANDR_AUTOCOMPLETE_SYSTEMS='{"kraken": { "class": "jormungandr.autocomplete.kraken.Kraken" }, "bragi": {"class": "jormungandr.autocomplete.geocodejson.GeocodeJson", "args": {"host": "http://127.0.0.1:4000", "timeout": 2}}}'
+```
+
+You can select your autocomplete engine based on the `_autocomplete` parameter when querying Jormung:
```
pbougue: I don't know if we want to expose this in doc, but if we do we should probably tell it's a `hidden debug-only` parameter (same for all those `_`-prefixed params).

source/jormungandr/readme.md
```diff
@@ -80,6 +80,21 @@ From `navitia/source/jormungandr/jormungandr` run :
 PYTHONPATH=..:../../navitiacommon/ JORMUNGANDR_INSTANCES_DIR=~/jormung_conf/ python manage.py runserver
 ```
 
+# Option
+## Autocomplete
+
+Setup Jormung for using either `Bragi` or `Kraken` for autocompletion
```
pbougue: The short for Jörmungandr is `Jormun`, no `g` :stuck_out_tongue_closed_eyes:

source/jormungandr/readme.md
nonifier: It's not that I wanna expose that doc so much, but I'd rather keep this somewhere for reference.

I'm happy to move this anywhere more relevant, but as far as the doc goes, this is the only place I know that gather knowledge on how to use Jormun.

source/jormungandr/readme.md
nonifier: Goddammit ! :sob:

benoit-bst (approved):
:+1:

source/jormungandr/readme.md
```diff
@@ -80,6 +80,21 @@ From `navitia/source/jormungandr/jormungandr` run :
...
+# Option
+## Autocomplete
+
+Setup Jormungandr for using either `Bragi` or `Kraken` for autocompletion
+
+```sh
+JORMUNGANDR_AUTOCOMPLETE_SYSTEMS='{"kraken": { "class": "jormungandr.autocomplete.kraken.Kraken" }, "bragi": {"class": "jormungandr.autocomplete.geocodejson.GeocodeJson", "args": {"host": "http://127.0.0.1:4000", "timeout": 2}}}'
+```
+
+You can select your autocomplete engine based on the `_autocomplete` parameter when querying Jormungandr:
```
pbougue: Please just add: `hidden debug-only` before parameter (unless someone has a better idea)

pbougue (approved):

PR #2393, opened 2018-04-16
kraken: correction on disruption reader from database for channel type

kadhikari:
Ticket concerned: https://jira.canaltp.fr/browse/NAVITIAII-2421

### DIFF
--- a/source/kraken/fill_disruption_from_database.cpp
+++ b/source/kraken/fill_disruption_from_database.cpp
@@ -154,8 +154,8 @@ void fill_disruption_from_database(const std::string& connection_string,
                "     AND co.contributor_code = ANY('{%s}')" // it's like a "IN" but won't crash if empty"
                "     AND d.status = 'published'"
                "     AND i.status = 'published'"
-               // Attension: we should not change order from "d.id" to "cht.id"
-               "     ORDER BY d.id, c.id, t.id, i.id, m.id, ch.id, cht.id, a.id, p.id"
+               // Warning : Any change in this order may produce error while charging in fill_disruption_from_database.h/DisruptionDatabaseReader"
+               "     ORDER BY d.id, c.id, t.id, i.id, m.id, ch.id, cht.id"
                "     LIMIT %i OFFSET %i"
                " ;") %meta.production_date.end() % meta.production_date.begin() %meta.production_date.end()  % contributors_array % items_per_request % offset).str();
         result = work.exec(request);
--- a/source/kraken/fill_disruption_from_database.cpp
+++ b/source/kraken/fill_disruption_from_database.cpp
@@ -154,7 +154,8 @@ void fill_disruption_from_database(const std::string& connection_string,
                "     AND co.contributor_code = ANY('{%s}')" // it's like a "IN" but won't crash if empty"
                "     AND d.status = 'published'"
                "     AND i.status = 'published'"
-               // Warning : Any change in this order may produce error while charging in fill_disruption_from_database.h/DisruptionDatabaseReader"
+               // Warning : Any change in this order may produce error while charging in
+               // fill_disruption_from_database.h/DisruptionDatabaseReader"
                "     ORDER BY d.id, c.id, t.id, i.id, m.id, ch.id, cht.id"
                "     LIMIT %i OFFSET %i"
                " ;") %meta.production_date.end() % meta.production_date.begin() %meta.production_date.end()  % contributors_array % items_per_request % offset).str();
--- a/source/kraken/fill_disruption_from_database.cpp
+++ b/source/kraken/fill_disruption_from_database.cpp
@@ -154,7 +154,8 @@ void fill_disruption_from_database(const std::string& connection_string,
                "     AND co.contributor_code = ANY('{%s}')" // it's like a "IN" but won't crash if empty"
                "     AND d.status = 'published'"
                "     AND i.status = 'published'"
-               "     ORDER BY d.id, c.id, t.id, i.id, a.id, p.id, m.id, ch.id, cht.id"
+               // Attension: we should not change order from "d.id" to "cht.id"
+               "     ORDER BY d.id, c.id, t.id, i.id, m.id, ch.id, cht.id, a.id, p.id"
                "     LIMIT %i OFFSET %i"
                " ;") %meta.production_date.end() % meta.production_date.begin() %meta.production_date.end()  % contributors_array % items_per_request % offset).str();
         result = work.exec(request);
--- a/source/kraken/fill_disruption_from_database.h
+++ b/source/kraken/fill_disruption_from_database.h
@@ -81,7 +81,7 @@ namespace navitia {
         // This function and all others below are templated so they can be tested
         template<typename T>
         void operator() (T const_it) {
-            // This code is strongly related to the database query (and it's order): fill_disruption_from_database.
+            // This code is strongly related to the database query (and it's order): fill_disruption_from_database.cpp
             if (!disruption || disruption->id() !=
                     const_it["disruption_id"].template as<std::string>()) {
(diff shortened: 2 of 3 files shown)


kadhikari:
jenkins test this please

kadhikari:
jenkins test this please

source/kraken/fill_disruption_from_database.cpp
```diff
@@ -154,7 +154,8 @@ void fill_disruption_from_database(const std::string& connection_string,
                "     AND co.contributor_code = ANY('{%s}')" // it's like a "IN" but won't crash if empty"
                "     AND d.status = 'published'"
                "     AND i.status = 'published'"
-               "     ORDER BY d.id, c.id, t.id, i.id, a.id, p.id, m.id, ch.id, cht.id"
+               // Attension: we should not change order from "d.id" to "cht.id"
```
pbougue: Is that a warning? I don't understand what it means, let's discuss it orally.

source/kraken/fill_disruption_from_database.h
```diff
@@ -124,8 +124,10 @@ namespace navitia {
                 }
             }
 
-            if (impact && (last_period_id != const_it["application_id"].template as<std::string>())) {
+            if (impact && !const_it["application_id"].is_null() &&
```
pbougue, line 128: This code is hard to read, if you can add a comment, it would be very welcome.

source/kraken/tests/disruption_reader_test.cpp
```diff
@@ -690,3 +732,262 @@ BOOST_AUTO_TEST_CASE(two_properties) {
     BOOST_CHECK_EQUAL(property.value(), "31");
 }
 
+BOOST_AUTO_TEST_CASE(message) {
```
pbougue: A comment on what you test would be nice (so describe what used to crash)

source/kraken/tests/disruption_reader_test.cpp
```diff
@@ -690,3 +732,262 @@ BOOST_AUTO_TEST_CASE(two_properties) {
...
+    navitia::type::PT_Data pt_data;
+    navitia::type::MetaData meta;
+    navitia::DisruptionDatabaseReader reader(pt_data, meta);
+    reader.disruption = std::make_unique<chaos::Disruption>();
+
+    // Add an impact
+    Const_it const_it;
+    const_it.set_message("m_id_1", "message_text_web", "1", "2");
+    reader.impact = new chaos::Impact();
+    // Add a message with channel web
```
pbougue: That comment is too early, we do that 20 lines below, no? Or it's a general comment, and it's with "add an impact", maybe?

source/kraken/tests/disruption_reader_test.cpp
```diff
@@ -690,3 +732,262 @@ BOOST_AUTO_TEST_CASE(two_properties) {
...
+    const_it.set_channel("channel_id_3", "channel_name_notification", "type","400" ,"1", "2");
+    channel = message->mutable_channel();
+    reader.fill_channel(const_it, channel);
+    const_it.set_channel_type("id_3", "notification");
+    reader.fill_channel_type(const_it, channel);
+    BOOST_REQUIRE_EQUAL(channel->types_size(), 1);
+}
+
+/* a badly sorted disruption record example
+ * Disruption record sorted in this way does not work for channel_type but works well for application_periods
```
pbougue: What does "work for channel_type" or application_periods means?

source/kraken/tests/disruption_reader_test.cpp
```diff
@@ -690,3 +732,262 @@ BOOST_AUTO_TEST_CASE(two_properties) {
...
+ * ..., message_1 , channel_1, channel_type_1, application_period_1 ...
+ * ..., message_2 , channel_2, channel_type_2, application_period_2 ...
+ * ..., message_3 , channel_3, channel_type_3, ...
+ * ..., message_1 , channel_1, channel_type_1, application_period_1 ...
+ * ..., message_2 , channel_2, channel_type_2, application_period_2 ...
+ * ..., message_3 , channel_3, channel_type_3, ...
+ */
+BOOST_AUTO_TEST_CASE(disruption_badly_sorted_informations) {
```
pbougue: This deserves more comment to explain what case you are testing, but are you testing here that it doesn't behave correctly? Do we want to enforce that behaviour?
To me it's interesting to see that you reproduiced the issue, but we cannot add it to tests.
Instead we should add a test that is using the correctly ordered db request, and that would crash if the order changes, because that's what you corrected.

source/kraken/tests/disruption_reader_test.cpp
```diff
@@ -690,3 +732,262 @@ BOOST_AUTO_TEST_CASE(two_properties) {
...
+    channel_type = channel.types(0);
+    BOOST_CHECK_EQUAL(channel_type, chaos::Channel_Type_mobile);
+
+    // Message notification
+    message = impact.messages(2);
+    BOOST_CHECK_EQUAL(message.text(), "message_text_notification");
+    channel = message.channel();
+    BOOST_CHECK_EQUAL(channel.name(), "channel_name_notification");
+    // Here we have 4 channel_types instead of 1. This is an error
+    BOOST_REQUIRE_EQUAL(channel.types_size(), 4);
```
pbougue: We don't want to enforce that behaviour : if we get a code that corrects it, it's OK, but then this test would fail and we wouldn't know if it's a problem.

source/kraken/tests/disruption_reader_test.cpp
```diff
@@ -690,3 +732,262 @@ BOOST_AUTO_TEST_CASE(two_properties) {
...
+    channel_type = channel.types(0);
+    BOOST_CHECK_EQUAL(channel_type, chaos::Channel_Type_notification);
+}
+
+/* a well sorted disruption record example
```
pbougue: If I understood correctly, you didn't change the behaviour of what you are testing here, so not sure all this test is important.
But I really might have understood it wrong as I'm a bit lost in that part of Navitia.

pbougue (commented):
Not sure: Shouldn't we also improve the code from DisruptionDatabaseReader so the order from db query doesn't matter?
> It's paginated for perf... to big issue for not so much reward :)

source/kraken/fill_disruption_from_database.h
```diff
@@ -68,19 +68,20 @@ namespace navitia {
...
         std::set<std::string> pt_object_ids;
+        std::set<std::string> appplication_periods_ids;
         std::set<std::string> associate_objects_ids;
         type::PT_Data& pt_data;
         const type::MetaData& meta;
 
         // This function and all others below are templated so they can be tested
         template<typename T>
         void operator() (T const_it) {
+            // This code is strongly related to the database query (and it's order): fill_disruption_from_database.
```
pbougue: minor: you can either remove file name, or complete it (with .cpp)

source/kraken/fill_disruption_from_database.cpp
```diff
@@ -154,7 +154,8 @@ void fill_disruption_from_database(const std::string& connection_string,
                "     AND co.contributor_code = ANY('{%s}')" // it's like a "IN" but won't crash if empty"
                "     AND d.status = 'published'"
                "     AND i.status = 'published'"
-               "     ORDER BY d.id, c.id, t.id, i.id, a.id, p.id, m.id, ch.id, cht.id"
+               // Warning : Any change in this order may produce error while charging in fill_disruption_from_database.h/DisruptionDatabaseReader"
```
pbougue: minor: line length

pbougue (commented):
As discussed with @Msaglier and @kadhikari:
 - [x] OK to merge, once @Msaglier is ready for testing (but we cannot merge before)

PR #2395, opened 2018-04-17
tyr: no job to be created if no action for mimir on dataset

kadhikari:
ticket: https://jira.canaltp.fr/browse/NAVP-855

### DIFF
--- a/source/tyr/tyr/tasks.py
+++ b/source/tyr/tyr/tasks.py
@@ -153,37 +153,38 @@ def send_to_mimir(instance, filename):
     returns action list
     """
     # This test is to avoid creating a new job if there is no action on mimir.
-    if instance.import_ntfs_in_mimir or instance.import_stops_in_mimir:
-        actions = []
-        job = models.Job()
-        instance_config = load_instance_config(instance.name)
-        job.instance = instance
-        job.state = 'pending'
+    if not (instance.import_ntfs_in_mimir or instance.import_stops_in_mimir):
+        return []
 
-        dataset = models.DataSet()
-        dataset.family_type = 'mimir'
-        dataset.type = 'fusio'
+    actions = []
+    job = models.Job()
+    instance_config = load_instance_config(instance.name)
+    job.instance = instance
+    job.state = 'pending'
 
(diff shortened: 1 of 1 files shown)


kadhikari:
jenkins test this please

mbounabi (approved):

source/tyr/tyr/tasks.py
```diff
@@ -130,9 +130,11 @@ def import_data(files, instance, backup_file, async=True, reload=True, custom_ou
         if reload:
             actions.append(reload_data.si(instance_config, job.id))
 
-        for dataset in job.data_sets:
-            if dataset.family_type == 'pt':
-                actions.extend(send_to_mimir(instance, dataset.name))
+        # Test parameters for instance for send_to_mimir
+        if instance.import_ntfs_in_mimir or instance.import_stops_in_mimir:
```
pbougue: To me it would be better to group it in send_to_mimir with all the tests on those params (just return empty actions in that case).
The code would be less scattered, easier to maintain.

pbougue (commented):
Can you add a test on that (probably in source/tyr/tests/integration/dataset_test.py)?

source/tyr/tyr/tasks.py
```diff
@@ -152,36 +152,38 @@ def send_to_mimir(instance, filename):
...
-    instance_config = load_instance_config(instance.name)
-    job.instance = instance
-    job.state = 'pending'
-
-    dataset = models.DataSet()
-    dataset.family_type = 'mimir'
-    dataset.type = 'fusio'
+    # This test is to avoid creating a new job if there is no action on mimir.
+    if instance.import_ntfs_in_mimir or instance.import_stops_in_mimir:
+        actions = []
```
pbougue: You could probably just do it here more compact:
```python
if not (instance.import_ntfs_in_mimir or instance.import_stops_in_mimir) :
    return actions  # or return []
```

pbougue (approved):
That's fine by me, if you want to change it, as you wish.

xlqian (approved):

PR #2398, opened 2018-04-19
Jormungandr:  parking DIVIA real-time

mbounabi:
https://jira.kisio.org/browse/NAVP-844

### DIFF
--- a/source/jormungandr/jormungandr/default_settings.py
+++ b/source/jormungandr/jormungandr/default_settings.py
@@ -129,6 +129,9 @@ CIRCUIT_BREAKER_JCDECAUX_TIMEOUT_S = 60  # the circuit breaker retries after thi
 CIRCUIT_BREAKER_MAX_STAR_FAIL = 4  # max instance call failures before stopping attempt
 CIRCUIT_BREAKER_STAR_TIMEOUT_S = 60  # the circuit breaker retries after this timeout (in seconds)
 
+CIRCUIT_BREAKER_MAX_DIVIA_FAIL = 4  # max instance call failures before stopping attempt
+CIRCUIT_BREAKER_DIVIA_TIMEOUT_S = 60  # the circuit breaker retries after this timeout (in seconds)
+
 CIRCUIT_BREAKER_MAX_CLEVERAGE_FAIL = 4  # max instance call failures before stopping attempt
 CIRCUIT_BREAKER_CLEVERAGE_TIMEOUT_S = 60  # the circuit breaker retries after this timeout (in seconds)
 
--- a/source/jormungandr/jormungandr/default_settings.py
+++ b/source/jormungandr/jormungandr/default_settings.py
@@ -126,11 +126,8 @@ CIRCUIT_BREAKER_SYNTHESE_TIMEOUT_S = 60  # the circuit breaker retries after thi
 CIRCUIT_BREAKER_MAX_JCDECAUX_FAIL = 4  # max instance call failures before stopping attempt
 CIRCUIT_BREAKER_JCDECAUX_TIMEOUT_S = 60  # the circuit breaker retries after this timeout (in seconds)
 
-CIRCUIT_BREAKER_MAX_STAR_FAIL = 4  # max instance call failures before stopping attempt
-CIRCUIT_BREAKER_STAR_TIMEOUT_S = 60  # the circuit breaker retries after this timeout (in seconds)
-
-CIRCUIT_BREAKER_MAX_DIVIA_FAIL = 4  # max instance call failures before stopping attempt
-CIRCUIT_BREAKER_DIVIA_TIMEOUT_S = 60  # the circuit breaker retries after this timeout (in seconds)
+CIRCUIT_BREAKER_MAX_CAR_PARK_FAIL = 4  # max instance call failures before stopping attempt
+CIRCUIT_BREAKER_CAR_PARK_TIMEOUT_S = 60  # the circuit breaker retries after this timeout (in seconds)
 
 CIRCUIT_BREAKER_MAX_CLEVERAGE_FAIL = 4  # max instance call failures before stopping attempt
 CIRCUIT_BREAKER_CLEVERAGE_TIMEOUT_S = 60  # the circuit breaker retries after this timeout (in seconds)
--- a/source/jormungandr/jormungandr/parking_space_availability/car/car_park_provider_manager.py
+++ b/source/jormungandr/jormungandr/parking_space_availability/car/car_park_provider_manager.py
@@ -26,10 +26,11 @@
 # IRC #navitia on freenode
 # https://groups.google.com/d/forum/navitia
 # www.navitia.io
+
 from __future__ import absolute_import, print_function, unicode_literals, division
 from jormungandr.parking_space_availability.abstract_provider_manager import AbstractProviderManager
 
-POI_TYPE_ID = 'poi_type:amenity:parking'
+POI_TYPE_IDS = ('poi_type:amenity:parking', 'poi_type:Airesdestationnement')
 
 
 class CarParkingProviderManager(AbstractProviderManager):
@@ -42,7 +43,7 @@ class CarParkingProviderManager(AbstractProviderManager):
             self.car_park_providers.append(self._init_class(configuration['class'], arguments))
 
     def _handle_poi(self, item):
-        if 'poi_type' in item and item['poi_type']['id'] == POI_TYPE_ID:
+        if 'poi_type' in item and item['poi_type']['id'] in POI_TYPE_IDS:
             provider = self._find_provider(item)
             if provider:
                 item['car_park'] = provider.get_informations(item)
--- a/source/jormungandr/jormungandr/parking_space_availability/car/car_park_provider_manager.py
+++ b/source/jormungandr/jormungandr/parking_space_availability/car/car_park_provider_manager.py
@@ -30,7 +30,7 @@
 from __future__ import absolute_import, print_function, unicode_literals, division
 from jormungandr.parking_space_availability.abstract_provider_manager import AbstractProviderManager
 
-POI_TYPE_IDS = ('poi_type:amenity:parking', 'poi_type:public_parking')
+POI_TYPE_ID = 'poi_type:amenity:parking'
 
 
 class CarParkingProviderManager(AbstractProviderManager):
@@ -43,7 +43,7 @@ class CarParkingProviderManager(AbstractProviderManager):
             self.car_park_providers.append(self._init_class(configuration['class'], arguments))
 
     def _handle_poi(self, item):
-        if 'poi_type' in item and item['poi_type']['id'] in POI_TYPE_IDS:
+        if 'poi_type' in item and item['poi_type']['id'] in POI_TYPE_ID:
             provider = self._find_provider(item)
             if provider:
                 item['car_park'] = provider.get_informations(item)
--- a/source/jormungandr/jormungandr/parking_space_availability/car/car_park_provider_manager.py
+++ b/source/jormungandr/jormungandr/parking_space_availability/car/car_park_provider_manager.py
@@ -43,7 +43,7 @@ class CarParkingProviderManager(AbstractProviderManager):
             self.car_park_providers.append(self._init_class(configuration['class'], arguments))
 
     def _handle_poi(self, item):
-        if 'poi_type' in item and item['poi_type']['id'] in POI_TYPE_ID:
+        if 'poi_type' in item and item['poi_type']['id'] == POI_TYPE_ID:
             provider = self._find_provider(item)
             if provider:
                 item['car_park'] = provider.get_informations(item)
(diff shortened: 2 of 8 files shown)


xlqian:
just in case, don't forget deployment_conf

source/jormungandr/jormungandr/parking_space_availability/car/common_car_park_provider.py
```diff
@@ -0,0 +1,103 @@
...
+        pass
+
+    def get_informations(self, poi):
+        return self._get_information(poi)
+
+    def support_poi(self, poi):
+        properties = poi.get('properties', {})
+        return properties.get('operator', '').lower() in self.operators
+
+    @cache.memoize(app.config['CACHE_CONFIGURATION'].get('TIMEOUT_CAR_PARK', 30))
```
xlqian, line 75: minor: i think it may be useful if we can set the timeout for different proxies...

source/jormungandr/jormungandr/parking_space_availability/car/star.py
```diff
@@ -27,49 +27,36 @@
...
 
 DEFAULT_STAR_FEED_PUBLISHER = None
 
 
-class StarProvider(AbstractParkingPlacesProvider):
+class StarProvider(CommonCarParkProvider):
 
     def __init__(self, url, operators, dataset, timeout=1, feed_publisher=DEFAULT_STAR_FEED_PUBLISHER, **kwargs):
 
         self.ws_service_template = url + '/?dataset={}&refine.idparc={}'
```
xlqian: just for the record, as you said, we can probably call the service just once, get the complete list, cache it and use it later, to be tested

source/jormungandr/jormungandr/parking_space_availability/car/divia.py
```diff
@@ -0,0 +1,67 @@
...
+        if kwargs.get('api_key'):
+            self.api_key = kwargs.get('api_key')
+
+    def _get_information(self, poi):
+        data = self._call_webservice(self.ws_service_template.format(self.dataset))
+
+        if not data:
+            return
+
+        park = jmespath.search('records[?fields.numero_parking==\'{}\']|[0]'.format(poi['properties']['ref'].zfill(2)), data)
```
xlqian: for other reviewers who are interested in the syntax: http://jmespath.org/tutorial.html#filter-projections

xlqian: can you try this: `records[?fields.numero_parking==\'{}\'] | [0].{available: fields.nombre_places_libres, occupied: fields.nombre_places} `

source/jormungandr/jormungandr/parking_space_availability/car/tests/divia_test.py
```diff
@@ -0,0 +1,110 @@
...
+    {
+        "records":[
+            {
+                "fields": {
+                }
+            }
+        ]
+    }
+    """
+
```
xlqian, line 99: can you add a test with

```python
    divia_response = """
    {
        "records":[
            {
                "fields": {
                      "numero_parking": "42"
                }
            }
        ]
    }
    """
```

source/jormungandr/jormungandr/parking_space_availability/car/divia.py
nonifier: Interesting :) I was wondering What The F#%$£ !

source/jormungandr/jormungandr/parking_space_availability/car/star.py
mbounabi: Sure. To be discussed

source/jormungandr/jormungandr/parking_space_availability/car/divia.py
mbounabi: @xlqian the search request doesn't work.

source/jormungandr/jormungandr/parking_space_availability/car/tests/divia_test.py
mbounabi: Done

source/jormungandr/jormungandr/parking_space_availability/car/car_park_provider_manager.py
```diff
@@ -26,10 +26,11 @@
 # IRC #navitia on freenode
 # https://groups.google.com/d/forum/navitia
 # www.navitia.io
+
 from __future__ import absolute_import, print_function, unicode_literals, division
 from jormungandr.parking_space_availability.abstract_provider_manager import AbstractProviderManager
 
-POI_TYPE_ID = 'poi_type:amenity:parking'
+POI_TYPE_IDS = ('poi_type:amenity:parking', 'poi_type:Airesdestationnement')
```
TeXitoi: We don't want that in the code.

source/jormungandr/jormungandr/parking_space_availability/car/divia.py
xlqian: @mbounabi :sob:

xlqian (approved):

source/jormungandr/jormungandr/default_settings.py
```diff
@@ -129,6 +129,9 @@
 CIRCUIT_BREAKER_MAX_STAR_FAIL = 4  # max instance call failures before stopping attempt
 CIRCUIT_BREAKER_STAR_TIMEOUT_S = 60  # the circuit breaker retries after this timeout (in seconds)
 
+CIRCUIT_BREAKER_MAX_DIVIA_FAIL = 4  # max instance call failures before stopping attempt
```
pbougue: shouldn't we merge it with star, as it's the same provider and have a param "ODS_API" ?

source/jormungandr/jormungandr/parking_space_availability/car/car_park_provider_manager.py
pbougue: We won't merge it right now, so it's OK for testing, but no way to go with that crap French naming :)
We will ask for a data change : I suggest `poi_type:public_parking`.
And if needed we add conf for crappy naming, but not in code :nauseated_face:

source/jormungandr/jormungandr/parking_space_availability/car/common_car_park_provider.py
```diff
@@ -0,0 +1,102 @@
...
+import logging
+import pybreaker
+import requests as requests
+
+from jormungandr import cache, app, utils, new_relic
+from jormungandr.parking_space_availability import AbstractParkingPlacesProvider
+from abc import abstractmethod
+
+
+class CommonCarParkProvider(AbstractParkingPlacesProvider):
```
pbougue, line 42: Shouldn't it be OpenDataSoftCarParkProvider? And maybe just a conf to tell which field goes where, URL, key, provider name...

pbougue (changes requested):
poi_type naming needs a change
also quite big refactor if possible

source/jormungandr/jormungandr/default_settings.py
mbounabi: let's make an unanimous decision on this one!

source/jormungandr/jormungandr/parking_space_availability/car/car_park_provider_manager.py
mbounabi: Fine for me! But we need to make sure it is changed in the data

source/jormungandr/jormungandr/parking_space_availability/car/common_car_park_provider.py
mbounabi: After discussion, seems that this idea is off the table

source/jormungandr/jormungandr/parking_space_availability/car/car_park_provider_manager.py
mbounabi: Same comment as below:
Fine for me! But we need to make sure it is changed in the data

source/jormungandr/jormungandr/parking_space_availability/car/common_car_park_provider.py
```diff
@@ -0,0 +1,115 @@
...
+            self.api_key = kwargs.get('api_key')
+
+    @abstractmethod
+    def process_data(self, data, poi):
+        pass
+
+    def get_informations(self, poi):
+        data = self._call_webservice(self.ws_service_template.format(self.dataset))
+
+        if not data or not poi.get('properties', {}).get('ref'):
```
pbougue: We should test properties[ref] before calling webservice: this can save precious time.

source/jormungandr/jormungandr/parking_space_availability/car/divia.py
```diff
@@ -0,0 +1,59 @@
...
+        self.provider_name = 'DIVIA'
+
+        super(DiviaProvider, self).__init__(url, operators, dataset, timeout, **kwargs)
+
+    def process_data(self, data, poi):
+        park = jmespath.search('records[?fields.numero_parking==\'{}\']|[0]'.format(poi['properties']['ref'].zfill(2)), data)
+        if park:
+            available = jmespath.search('fields.nombre_places_libres', park)
+            nb_places = jmespath.search('fields.nombre_places', park)
+            if available is not None and nb_places and nb_places > available:
```
pbougue: Here, I think we should test for "... and nb_places `is not None` and nb_places >`=` available.
No place on a parking is dumb but it may happen while construction works, and parking can definitely be empty, I think.

source/jormungandr/jormungandr/parking_space_availability/car/parking_places.py
```diff
@@ -41,7 +41,7 @@ def __init__(self, available=None, occupied=None, available_PRM=None, occupied_P
         if occupied_PRM is not None:
             self.occupied_PRM = occupied_PRM
 
-        if any(True for n in [available, occupied, available_PRM, occupied_PRM] if n is not None):
+        if any(n is not None for n in [available, occupied, available_PRM, occupied_PRM]):
             self.total_places = (available or 0) + (occupied or 0) + (available_PRM or 0) + (occupied_PRM or 0)
```
pbougue, line 45: Not sure it's the scope of this PR: to me, None is "no information" while 0 is no place.
But it was there before, so unless someone else thinks this can be a problem (those `or 0`), let's keep it that way.

mbounabi: As discussed, this does what is expected. 0 (no place) will be returned.

source/jormungandr/jormungandr/parking_space_availability/car/common_car_park_provider.py
mbounabi: As discussed, this is a default value and it is better to keep it common to every car park providers

pbougue (approved):
So now we wait until we know more about the use-case of Divia (fallback on public_parking or not), and then about the data we'll get.

TODO:
 - [x] ensure use and data are OK (`do_not_merge` label is here for that)

source/jormungandr/jormungandr/parking_space_availability/car/car_park_provider_manager.py
```diff
@@ -42,7 +43,7 @@ def __init__(self, car_park_providers_configurations):
             self.car_park_providers.append(self._init_class(configuration['class'], arguments))
 
     def _handle_poi(self, item):
-        if 'poi_type' in item and item['poi_type']['id'] == POI_TYPE_ID:
+        if 'poi_type' in item and item['poi_type']['id'] in POI_TYPE_ID:
```
pbougue: Here, without an array of ids I think we can have weird bugs like `poi_type:amenity:park` would be valid, as we check that a string is in another one, no?

mbounabi: True

source/jormungandr/jormungandr/parking_space_availability/car/common_car_park_provider.py
```diff
@@ -0,0 +1,116 @@
...
+            if self.api_key:
+                headers = {'Authorization': 'apiKey {}'.format(self.api_key)}
+            else:
+                headers = None
+            data = self.breaker.call(requests.get, url=request_url, headers=headers, timeout=self.timeout)
+            # record in newrelic
+            self.record_call("OK")
+            return data.json()
+        except pybreaker.CircuitBreakerError as e:
+            msg = '{} service dead (error: {})'.format(self.provider_name, e)
```
nonifier, line 87: I can see the same exception handling code being copied over 3 times...
I would rather extract out a function that records failure to Newrelic (and that could potentially be used elsewhere)

nonifier (approved):

source/jormungandr/jormungandr/parking_space_availability/car/common_car_park_provider.py
mbounabi: After investigation, the message handling and new relic event recording can be factorized a bit. I suugest that this can be handled in a technical debt ticket because it goes beyond this PR.

PR #2402, opened 2018-04-25
tyr: Modify migrate_from_poi_to_osm to delete_dataset

kadhikari:
1. delete_dataset deletes all data_sets of a type.
2. New end point migrate_from_poi_to_osm added to activate parameter "parse_pois_from_osm" in the table navitia.paameters.

3. Concerned ticket: https://jira.kisio.org/browse/NAVITIAII-2390

### DIFF
--- a/source/ed/readme.md
+++ b/source/ed/readme.md
@@ -57,7 +57,7 @@ Note: fares can also be loaded with fusio2ed if the fare files are in the fusio 
 Component that loads POI (Point Of Interest) into `ed` from csv files.
 
 You can use poi2ed to get POI in addition of osm2ed (that mainly extract street network data) ; If you do so, osm2ed will not extract any POI from the OSM dataset and use the poi2ed dataset instead.
-If you want to switch your POI source back to OSM, you will need to set the `parse_pois_from_osm` boolean to `true` by using an api tyr `PUT $HOST/v0/instances/$INSTANCE_NAME/actions/migrate_from_poi_to_osm` and launch again an osm2ed data integration job.
+If you want to switch your POI source back to OSM, you will need to set the `parse_pois_from_osm` boolean to `true` by using an api tyr `PUT $HOST/v0/instances/$INSTANCE_NAME/actions/migrate_from_poi_to_osm` followed by `DELETE $HOST/v0/instances/$INSTANCE_NAME/actions/delete_dataset/poi` and launch again an osm2ed data integration job.
 
 ## geopal2ed
 Component that loads geopal street network data into `ed`.
--- a/source/ed/readme.md
+++ b/source/ed/readme.md
@@ -57,7 +57,7 @@ Note: fares can also be loaded with fusio2ed if the fare files are in the fusio 
 Component that loads POI (Point Of Interest) into `ed` from csv files.
 
 You can use poi2ed to get POI in addition of osm2ed (that mainly extract street network data) ; If you do so, osm2ed will not extract any POI from the OSM dataset and use the poi2ed dataset instead.
-If you want to switch your POI source back to OSM, you will need to set the `parse_pois_from_osm` boolean to `true` by using an api tyr `DELETE $HOST/v0/instances/$INSTANCE_NAME/actions/migrate_from_poi_to_osm` and launch again an osm2ed data integration job.
+If you want to switch your POI source back to OSM, you will need to set the `parse_pois_from_osm` boolean to `true` by using an api tyr `PUT $HOST/v0/instances/$INSTANCE_NAME/actions/migrate_from_poi_to_osm` and launch again an osm2ed data integration job.
 
 ## geopal2ed
 Component that loads geopal street network data into `ed`.
--- a/source/ed/readme.md
+++ b/source/ed/readme.md
@@ -57,7 +57,7 @@ Note: fares can also be loaded with fusio2ed if the fare files are in the fusio 
 Component that loads POI (Point Of Interest) into `ed` from csv files.
 
 You can use poi2ed to get POI in addition of osm2ed (that mainly extract street network data) ; If you do so, osm2ed will not extract any POI from the OSM dataset and use the poi2ed dataset instead.
-If you want to switch your POI source back to OSM, you will need to set the `parse_pois_from_osm` boolean to `true` in the `ed` database (table `navitia.parameters`) and launch again an osm2ed data integration job.
+If you want to switch your POI source back to OSM, you will need to set the `parse_pois_from_osm` boolean to `true` by using an api tyr `DELETE $HOST/v0/instances/$INSTANCE_NAME/actions/migrate_from_poi_to_osm` and launch again an osm2ed data integration job.
 
 ## geopal2ed
 Component that loads geopal street network data into `ed`.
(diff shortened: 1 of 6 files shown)


source/ed/readme.md
```diff
@@ -57,7 +57,7 @@ Note: fares can also be loaded with fusio2ed if the fare files are in the fusio
 Component that loads POI (Point Of Interest) into `ed` from csv files.
 
 You can use poi2ed to get POI in addition of osm2ed (that mainly extract street network data) ; If you do so, osm2ed will not extract any POI from the OSM dataset and use the poi2ed dataset instead.
-If you want to switch your POI source back to OSM, you will need to set the `parse_pois_from_osm` boolean to `true` in the `ed` database (table `navitia.parameters`) and launch again an osm2ed data integration job.
+If you want to switch your POI source back to OSM, you will need to set the `parse_pois_from_osm` boolean to `true` by using an api tyr `DELETE $HOST/v0/instances/$INSTANCE_NAME/actions/migrate_from_poi_to_osm` and launch again an osm2ed data integration job.
```
mbounabi: A PUT request is needed for 'migrate_from_poi_to_osm', not DELETE

source/tyr/tests/integration/dataset_delete_by_type_test.py
```diff
@@ -110,31 +110,54 @@ def create_tyr_update_jobs():
         return instance.name
 
 
-def test_migrate_poi_to_osm_wrong_instance(create_instance):
+def test_delete_for_wrong_instance(create_instance):
 
-    resp, status_code = api_delete('/v0/instances/wrong_instance/actions/migrate_from_poi_to_osm', check=False)
+    resp, status_code = api_delete('/v0/instances/wrong_instance/actions/delete_dataset/toto', check=False)
```
mbounabi: Two tests might be useful here:
- wrong instance
- wrong POI type

mbounabi (approved):

source/navitiacommon/navitiacommon/models.py
```diff
@@ -410,18 +410,21 @@ def get_from_id_or_name(cls, id=None, name=None):
...
-    def delete_dataset(self):
+    def delete_dataset(self, _type):
         result = db.session.query(DataSet, Job) \
             .join(Job) \
-            .filter(DataSet.type == 'poi', Job.instance_id == self.id)\
+            .filter(DataSet.type == _type, Job.instance_id == self.id)\
             .all()
 
+        if not result:
+            return None
```
nonifier: Shall you return `0` instead ? to be consistent with `return len(result)`

source/tyr/tyr/resources.py
```diff
@@ -1095,13 +1095,40 @@ def post(self, ac_instance_name):
...
+        return {'action': return_msg}, return_status
+
+
 class MigrateFromPoiToOsm(flask_restful.Resource):
-    def delete(self, instance_name):
 
+    def put(self, instance_name):
         instance = models.Instance.get_by_name(instance_name)
         if instance:
-            instance.delete_dataset()
```
kinnou02, line 1103: You still need to delete the dataset ot type poi

source/tyr/tyr/resources.py
```diff
@@ -1095,13 +1095,40 @@ def post(self, ac_instance_name):
...
-            return_msg = 'All POI datasets deleted for instance {}'.format(instance_name)
+            instance_conf = load_instance_config(instance_name)
+            connection_string = "postgres://{u}:{pw}@{h}:{port}/{db}"\
+                .format(u=instance_conf.pg_username, pw=instance_conf.pg_password,
+                        h=instance_conf.pg_host, db=instance_conf.pg_dbname, port=instance_conf.pg_port)
+            engine = sqlalchemy.create_engine(connection_string)
+
+            # update the line, simplified to approx 100m
```
kinnou02: This comment is false

source/tyr/tyr/resources.py
```diff
@@ -1095,13 +1095,40 @@ def post(self, ac_instance_name):
         return marshal({'job': job}, one_job_fields), 200
 
 
+class DeleteDataset(flask_restful.Resource):
+
+    def delete(self, instance_name, type='poi'):
```
kinnou02: Not a fan of the default value, I prefer that kind of api to be explicit

kinnou02 (changes requested):

source/tyr/tyr/resources.py
mbounabi: @kinnou02  Does it have to be in the same request? I think it will be better to maintain the separation betwwen the two endpoints and make two requests to delete the POI datasets first and then set the boolean.
The doc (ed/readme.md) will have to be explicit about it if we agree to do that.

kinnou02: It's error prone, if we keep two separate request the documentation have to be very clear.
remember that theses actions should be performed by members of the data team, that aren't developers

kadhikari: I agree with @mbounabi  and propose to add "...an api tyr `PUT $HOST/v0/instances/$INSTANCE_NAME/actions/migrate_from_poi_to_osm` followed by DELETE $HOST/v0/instances/$INSTANCE_NAME/actions/delete_dataset/$TYPE  ..." in the file /ed/readme.md

kinnou02 (approved):

kadhikari:
jenkins test this please

kadhikari:
jenkins, test this please

source/tyr/readme.md
```diff
@@ -714,13 +714,27 @@ Finally if you want to delete a poi_type you just have to use the DELETE action:
 
 #### Migrate from POI to OSM
 
-Remove all POI datasets stored in database.
+Load POI from OSM.
 
-    DELETE $HOST/v0/instances/$INSTANCE_NAME/actions/migrate_from_poi_to_osm
+    PUT $HOST/v0/instances/$INSTANCE_NAME/actions/migrate_from_poi_to_osm
```
pbougue, line 719: We should add that it must be followed by a delet_dataset/poi (otherwise we can have a problem on reload, no?)
In that case we may even add it to the json response.

source/tyr/readme.md
```diff
@@ -714,13 +714,27 @@ Finally if you want to delete a poi_type you just have to use the DELETE action:
...
+    "action": "Parameter parse_pois_from_osm activated"
+}
+```
+
+
+#### Delete data_sets by type
+
+Delete all the data_sets by type stored in database.
+
+    DELETE $HOST/v0/instances/$INSTANCE_NAME/actions/delete_dataset/$TYPE
```
pbougue, line 733: We must add the type available : `poi, fusio, osm, geopal, synonym, shape`?
Maybe we should retrain them, so that we don't get any error when requesting/dropping part of the db?

source/tyr/readme.md
pbougue: Also a comment on how adding a poi file switches automatically to poi (and so why this command is useful) would be nice.

pbougue (commented):
Doc changes request (to be done in another PR, then)

PR #2403, opened 2018-04-26
tyr: delete old jobs of discarded instance

kadhikari:
Ticket : https://jira.kisio.org/browse/NAVITIAII-2442

Note: old jobs with state <> 'done' or without any data_sets are cleaned with existing version

### DIFF
--- a/source/tyr/tests/integration/purge_test.py
+++ b/source/tyr/tests/integration/purge_test.py
@@ -71,14 +71,6 @@ def create_job(creation_date, dataset_type, backup_dir):
     return job
 
 
-def add_data_set(job, creation_date, dataset_type, backup_dir):
-    dataset_backup_dir = tempfile.mkdtemp(dir=backup_dir)
-    dataset, metric = create_dataset(dataset_type, dataset_backup_dir)
-    job.data_sets.append(dataset)
-    job.metrics.append(metric)
-    return job
-
-
 def create_jobs_with_same_datasets(name, backup_dir):
     with app.app_context():
         job_list = []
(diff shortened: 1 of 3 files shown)


source/tyr/tests/integration/purge_test.py
```diff
@@ -207,3 +216,34 @@ def test_purge_old_jobs_same_dataset():
...
+    instances_resp = api_delete('/v0/instances/fr-discarded')
+    assert instances_resp['discarded'] is True
+
+    jobs_resp = api_get('/v0/jobs/fr-discarded')
+    assert len(jobs_resp['jobs']) == 3
+    jobs_resp = api_get('/v0/jobs/fr')
+    assert len(jobs_resp['jobs']) == 3
+    tasks.purge_jobs()
+    jobs_resp = api_get('/v0/jobs/fr-discarded')
+    assert len(jobs_resp['jobs']) == 1
```
mbounabi, line 239: Does this mean that for a discarded instance, the last dataset is still kept?

source/tyr/tests/integration/purge_test.py
```diff
@@ -71,6 +71,14 @@ def create_job(creation_date, dataset_type, backup_dir):
     return job
 
 
+def add_data_set(job, creation_date, dataset_type, backup_dir):
```
mbounabi: Where is this function used?

source/tyr/tests/integration/purge_test.py
kadhikari: Exactly. We will have only fewer datasets of discarded instance.

mbounabi (approved):

kinnou02 (approved):

PR #2405, opened 2018-04-27
kraken: min_nb_journeys feature

benoit-bst:
This PR contains all logics to iterate on several Raptor call with "min_nb_journeys" parameters.

JIRA :
https://jira.kisio.org/browse/NAVP-863


source/routing/raptor_api.cpp
```diff
@@ -1167,74 +1164,126 @@ void make_response(navitia::PbCreator& pb_creator,
                     destination,
                     datetimes,
                     clockwise);
-        if (pb_creator.empty_journeys()) {
-            pb_creator.fill_pb_error(pbnavitia::Error::no_destination,
-                                     pbnavitia::NO_DESTINATION_POINT,
-                                     "no destination point");
+
+        if (pb_creator.empty_journeys())
```
benoit-bst: :+1:

source/jormungandr/jormungandr/scenarios/new_default.py
```diff
@@ -802,16 +808,31 @@ def fill_journeys(self, request_type, api_request, instance):
...
+        # Return the possible couples combinations (origin_mode and destination_mode)
         krakens_call = get_kraken_calls(api_request)
 
+        # min_nb_journeys option
+        # we take into account the option only if we have one
+        # origin_mode and destination_mode couple.
+        if api_request['min_nb_journeys']:
+            min_nb_journeys = api_request['min_nb_journeys']
+            if min_nb_journeys > 1 and len(krakens_call) > 1:
+                api_request['min_nb_journeys'] = '1'
```
xlqian: why '1'? why string?

source/jormungandr/jormungandr/scenarios/new_default.py
```diff
@@ -802,16 +808,31 @@ def fill_journeys(self, request_type, api_request, instance):
...
         request = deepcopy(api_request)
-        min_asked_journeys = get_or_default(request, 'min_nb_journeys', 1)
+
         min_journeys_calls = get_or_default(request, '_min_journeys_calls', 1)
 
         responses = []
         nb_try = 0
         while request is not None and \
-                ((nb_journeys(responses) < min_asked_journeys and nb_try < min_asked_journeys)
+                ((nb_journeys(responses) < min_nb_journeys and nb_try < min_nb_journeys)
```
xlqian: ~nb_try < min_nb_journeys~
```
 nb_try < min_journeys_calls
``` 
no? i think it was a copy-paste evil

source/jormungandr/jormungandr/scenarios/new_default.py
```diff
@@ -802,16 +808,31 @@ def fill_journeys(self, request_type, api_request, instance):
...
+        # min_nb_journeys option
+        # we take into account the option only if we have one
+        # origin_mode and destination_mode couple.
+        if api_request['min_nb_journeys']:
+            min_nb_journeys = api_request['min_nb_journeys']
+            if min_nb_journeys > 1 and len(krakens_call) > 1:
+                api_request['min_nb_journeys'] = '1'
+                min_nb_journeys = 1
+        else:
+            api_request['min_nb_journeys'] = '1'
```
xlqian: same here

source/jormungandr/jormungandr/scenarios/new_default.py
```diff
@@ -932,12 +953,14 @@ def create_next_kraken_request(self, request, responses):
         to do that we find ask the next (resp previous) query datetime
         """
         vjs = [j for r in responses for j in r.journeys if not journey_filter.to_be_deleted(j)]
```
xlqian: this line seems to be useless

xlqian: the unit tests for this functions fail

source/routing/journey.h
```diff
@@ -0,0 +1,85 @@
...
+    bool operator==(const Journey & rhs) const;
+    friend std::ostream& operator<<(std::ostream& os, const Journey& j);
+
+    std::vector<Section> sections;// the pt sections, with transfer between them
+    navitia::time_duration sn_dur = 0_s;// street network duration
+    navitia::time_duration transfer_dur = 0_s;// walking duration during transfer
+    navitia::time_duration min_waiting_dur = 0_s;// minimal waiting duration on every transfers
+    DateTime departure_dt = 0;// the departure dt of the journey, including sn
+    DateTime arrival_dt = 0;// the arrival dt of the journey, including sn
+    uint8_t nb_vj_extentions = 0;// number of vehicle journey extentions (I love useless comments!)
```
xlqian, line 73: vj extensions?

source/routing/raptor_api.cpp
```diff
@@ -1167,74 +1166,131 @@ void make_response(navitia::PbCreator& pb_creator,
...
-            std::vector<Path> tmp = raptor.compute_all(
-                *departures, *destinations, init_dt, rt_level, transfer_penalty, bound, max_transfers,
+        }
+
+        // Call raptor
+        // note : Loop is for min_nb_journeys options
+        //
+        // min_nb_journeys options :
+        // Compute several loop until the number of journeys >= min_nb_journeys
+        // For each step, we fing best pathes,
```
xlqian: we find(?) best paths

source/routing/raptor_api.cpp
```diff
@@ -1167,74 +1166,131 @@ void make_response(navitia::PbCreator& pb_creator,
...
+        // start to the lastest departure + 1(clockwise) or the lastest arrival - 1.
```
xlqian: the ~~lastest~~ latest departure + 1(clockwise) or the ~~lastest~~ earliest arrival - 1.

source/routing/raptor_api.cpp
```diff
@@ -980,6 +979,28 @@ void free_radius_filter(routing::map_stop_point_duration& sp_list,
     }
 }
 
+DateTime prepare_next_call_for_raptor(const std::list<Journey>& journeys, const bool clockwise)
```
xlqian: tests in python may be migrated into cpp

source/jormungandr/jormungandr/scenarios/new_default.py
```diff
@@ -173,6 +173,12 @@ def create_pb_request(requested_type, request, dep_mode, arr_mode):
     if request["free_radius_to"]:
         req.journeys.free_radius_to = request["free_radius_to"]
 
+    if request["min_nb_journeys"]:
```
mbounabi, line 176: Doesn't this parameter clash with one that has the same name in https://github.com/CanalTP/navitia/pull/2413 ?

source/jormungandr/jormungandr/scenarios/new_default.py
```diff
@@ -803,33 +809,48 @@ def fill_journeys(self, request_type, api_request, instance):
...
+        # we take into account the option only if we have one
+        # origin_mode and destination_mode couple.
+        if api_request['min_nb_journeys']:
+            min_nb_journeys = api_request['min_nb_journeys']
+            if min_nb_journeys > 1 and len(krakens_call) > 1:
+                api_request['min_nb_journeys'] = 1
+                min_nb_journeys = 1
+        else:
+            api_request['min_nb_journeys'] = 1
+            min_nb_journeys = 1
```
```
if not api_request['min_nb_journeys']:
    api_request['min_nb_journeys'] = 1
min_nb_journeys = api_request['min_nb_journeys']
min_nb_journeys  = api_request['min_nb_journeys'] = 1 if min_nb_journeys>1 and len(kraken_call)>1 else min_nb_journeys
```
(not sure if it's better)

source/jormungandr/jormungandr/scenarios/new_default.py
```diff
@@ -803,33 +809,48 @@ def fill_journeys(self, request_type, api_request, instance):
...
 
+        # Return the possible couples combinations (origin_mode and destination_mode)
         krakens_call = get_kraken_calls(api_request)
 
+        # min_nb_journeys option
+        # we take into account the option only if we have one
+        # origin_mode and destination_mode couple.
+        if api_request['min_nb_journeys']:
+            min_nb_journeys = api_request['min_nb_journeys']
+            if min_nb_journeys > 1 and len(krakens_call) > 1:
```
xlqian: why `min_nb_journeys > 1` is needed? 

shouldn't `len(krakens_call) == 1` be sufficient?

like 

```
if not len(kraken_call) == 1:
     api_request['min_nb_journeys'] = 1
``` 
?

source/jormungandr/jormungandr/scenarios/new_default.py
```diff
@@ -959,19 +980,27 @@ def isochrone(self, request, instance):
 
         return resp
 
-    def create_next_kraken_request(self, request, responses):
+    def create_next_kraken_request(self, request, responses, min_nb_journeys = 0):
```
xlqian: it'd make more sense if we default min_nb_journeys to 1

xlqian: i'm confused,  `min_nb_journeys` is never used in this function, why keep it in the arguments?

source/routing/journey.h
nonifier: I don't really know what it's used for. It's been moved from `source/routing/raptor_solution_reader.h` (see below)

nonifier:
jenkins please test this.

source/jormungandr/jormungandr/scenarios/new_default.py
```diff
@@ -803,19 +806,34 @@ def fill_journeys(self, request_type, api_request, instance):
...
         krakens_call = get_kraken_calls(api_request)
 
+        # min_nb_journeys option
+        # we take into account the option only if we have one
+        # origin_mode and destination_mode couple.
+        if api_request['min_nb_journeys']:
+            min_nb_journeys = api_request['min_nb_journeys']
+            if min_nb_journeys > 1 and len(krakens_call) > 1:
+                api_request['min_nb_journeys'] = 1
+                min_nb_journeys = 1
```
kinnou02: It look like you do not use the min_nb_journeys if there is more than one call to kraken but we still want to use it, we just can't do it in kraken as it will generate too much journeys, that we do not want to display

source/routing/journey.h
kinnou02: Also known as "stay in" or  "prolongement de service" in french.
there is some explanation in the documentation: http://doc.navitia.io/#journeys

source/routing/raptor_api.cpp
```diff
@@ -980,12 +979,34 @@ void free_radius_filter(routing::map_stop_point_duration& sp_list,
     }
 }
 
-static std::vector<bt::ptime>
+DateTime prepare_next_call_for_raptor(const std::list<Journey>& journeys, const bool clockwise)
```
kinnou02: could be`Journeys`

source/routing/journey.cpp
```diff
@@ -0,0 +1,118 @@
...
+
+    boost::hash_combine(seed, s.get_in_st->order());
+    boost::hash_combine(seed, s.get_in_dt);
+    boost::hash_combine(seed, s.get_out_st->order());
+    boost::hash_combine(seed, s.get_out_dt);
+
+    return seed;
+}
+
+size_t JourneyHash::operator()(const Journey& j) const {
```
kinnou02, line 110: What do you want to do here? And by that I mean what do you consider a "similar" journey.
You only compare in and out datetime and the relative position of the stoptime in the VJ without checking that it is the same vj.

source/routing/raptor_api.cpp
```diff
@@ -1167,74 +1167,135 @@ void make_response(navitia::PbCreator& pb_creator,
...
+            request_date_secs = prepare_next_call_for_raptor(raptor_journeys, clockwise);
+
+            // filter the similar journeys
+            for(const auto & journey : raptor_journeys) {
+                journeys.insert(journey);
+            }
+
+            nb_try++;
+
+        } while (( journeys.size() < min_nb_journeys) && (nb_try < MAX_NB_RAPTOR_CALL));
```
kinnou02: 100 try seems to be too much, in jormun  `MAX_NB_RAPTOR_CALL` is equal to `min_nb_journeys`, it's not that bad but I think we need to set lower and upper limit if we choose to do it this way.

source/jormungandr/jormungandr/scenarios/new_default.py
kinnou02: What do you min by that? This isn't a new parameter, it exist since a few years

nonifier:
jenkins please test this

source/jormungandr/jormungandr/scenarios/new_default.py
benoit-bst: If I understood correctly, we said to use the Min_nb_journeys option only with one couple of origin / destination.
@nonifier ?

source/jormungandr/jormungandr/scenarios/new_default.py
benoit-bst: There is no collision with min_nb_journeys parameters as Alex said, it exist since a few years

source/routing/raptor_api.cpp
benoit-bst: MAX_NB_RAPTOR_CALL = 100 is temporary.
I think is in the next JIRA ticket to add a flexible limit : https://jira.kisio.org/secure/RapidBoard.jspa?rapidView=111&projectKey=NAVP&view=planning&selectedIssue=NAVP-859

xlqian:
jenkins, test this please!

source/routing/journey.cpp
```diff
@@ -0,0 +1,118 @@
...
+bool Journey::Section::operator==(const Section & rhs) const {
+
+    return get_in_st->order() == rhs.get_in_st->order()
+        && get_in_dt == rhs.get_in_dt
+        && get_out_st->order() == rhs.get_out_st->order()
+        && get_out_dt == rhs.get_out_dt;
+}
+
+bool Journey::is_pt() const {
+    if (sections.empty()) {
```
xlqian: return ! sections.empty(); 

no?

source/routing/raptor_api.cpp
```diff
@@ -1167,73 +1237,144 @@ void make_response(navitia::PbCreator& pb_creator,
                     destination,
                     datetimes,
                     clockwise);
-        if (pb_creator.empty_journeys()) {
-            pb_creator.fill_pb_error(pbnavitia::Error::no_destination,
-                                     pbnavitia::NO_DESTINATION_POINT,
-                                     "no destination point");
+
+        if (pb_creator.empty_journeys())
+        {
```
kinnou02: it's a matter of taste but the `{` should be on the same line than the `if`, `for`, etc... and they are required.

source/jormungandr/jormungandr/scenarios/new_default.py
benoit-bst: :white_check_mark:

source/jormungandr/jormungandr/scenarios/new_default.py
benoit-bst: It doesn't work. We set min_nb_journeys to "1" only, if min_nb_journeys is active and if the couple origin/dest is greater than 1.

source/routing/raptor_api.cpp

source/routing/journey.cpp
benoit-bst: :white_check_mark:

source/routing/raptor_api.cpp

source/jormungandr/jormungandr/scenarios/new_default.py
benoit-bst: :white_check_mark:

source/jormungandr/jormungandr/scenarios/new_default.py
benoit-bst: :white_check_mark:

source/jormungandr/jormungandr/scenarios/new_default.py
nonifier: @kinnou02 : are you saying that, when we have multiple calls to kraken, we should not use it within Kraken but still use it in Jormung ? 

In this case we should remove this line ?

TeXitoi: In all cases, we must do the min_nb_journeys on jormungandr as before. In the case of one first_section_mode and one last_section_mode, we give the min_nb_journeys to kraken as an optimization, but still continue do manage it in jormungandr as jormungandr might filter journeys.

source/routing/journey.cpp
```diff
@@ -0,0 +1,114 @@
...
+    get_out_st(&out),
+    get_out_dt(out_dt)
+{}
+
+bool Journey::Section::operator==(const Section & rhs) const {
+
+    return get_in_st->order() == rhs.get_in_st->order()
+        && get_in_dt == rhs.get_in_dt
+        && get_out_st->order() == rhs.get_out_st->order()
+        && get_out_dt == rhs.get_out_dt;
```
TeXitoi: That's not enough, you must also check get_in_st->vj and get_out_dt->vj, or directly get_in_st and get_out_dt

source/routing/raptor.cpp
```diff
@@ -389,6 +389,37 @@ RAPTOR::compute_all(const map_stop_point_duration& departures,
                     bool clockwise,
                     const boost::optional<navitia::time_duration>& direct_path_dur,
                     const size_t max_extra_second_pass) {
+
+    const auto & journeys = compute_all_journeys(departures,
+                                         destinations,
```
TeXitoi: strange indentation

benoit-bst:
Jenkins, test this please !

source/routing/raptor_api.cpp
```diff
@@ -264,7 +264,7 @@ void add_direct_path(PbCreator& pb_creator,
                      const georef::Path& path,
                      const type::EntryPoint& origin,
                      const type::EntryPoint& destination,
-                     const std::vector<bt::ptime>& datetimes,
+                     const std::vector<std::pair<bt::ptime, uint64_t>>& datetimes,
```
TeXitoi: Why? Looks like whatever this is for, we can do it in another clearer way. Reading at this prototype, I can't say what this uint64_t is for.

source/routing/raptor_api.cpp
```diff
@@ -1082,6 +1102,77 @@ void make_pt_response(navitia::PbCreator& pb_creator,
...
+        if(way_later(j1, j2, params)) {
+            late_journeys.push_back(journey_pair.first);
+        }
+        else if(way_later(j2, j1, params)) {
+            late_journeys.push_back(journey_pair.second);
+        }
+    }
+
+    for(auto& late_journey : late_journeys) {
+        journeys.erase(late_journey);
```
TeXitoi: Kaboom!

source/routing/tests/journey_test.cpp
```diff
@@ -0,0 +1,122 @@
...
+
+BOOST_AUTO_TEST_CASE(similar_sections_should_be_equal) {
+
+    ed::builder b("20180101");
+
+    auto& vj1 = b.vj("VJ1")("StopPoint1", "09:42:00"_t)("StopPoint2", "10:00:00"_t);
+    auto& vj2 = b.vj("VJ2")("StopPoint1", "09:37:00"_t)("StopPoint2", "09:55:00"_t);
+
+    type::StopTime st1(1, 2, b.sps["StopPoint1"]);
+    type::StopTime st2(3, 4, b.sps["StopPoint2"]);
```
TeXitoi: Doing that, order() will do shitty results.

TeXitoi (changes requested):

source/jormungandr/jormungandr/scenarios/new_default.py
nonifier: Ok, we now always keep `min_nb_journeys` in Jormun with a separate logic for Kraken that disable it with multiple modes, or set it to the number of journeys that remains.

source/jormungandr/jormungandr/scenarios/new_default.py
```diff
@@ -803,26 +809,43 @@ def fill_journeys(self, request_type, api_request, instance):
...
 
         responses = []
         nb_try = 0
-        nb_qualified_journeys = nb_journeys(responses)
+        nb_qualified_journeys = 0
 
         while request is not None and \
-                ((nb_qualified_journeys < min_asked_journeys and nb_try < min_asked_journeys)
-                 or nb_try < min_journeys_calls):
+                ((nb_qualified_journeys < min_nb_journeys and nb_try < min_nb_journeys) \
```
xlqian, line 833: shouldn't ~~nb_try < min_nb_journeys~~ rather be   `nb_try < min_journeys_calls`? (it's not your code though...)

source/jormungandr/jormungandr/scenarios/new_default.py
```diff
@@ -965,13 +988,21 @@ def create_next_kraken_request(self, request, responses):
 
         to do that we find ask the next (resp previous) query datetime
         """
-        vjs = journey_filter.get_qualified_journeys(responses)
-        if request["clockwise"]:
-            request['datetime'] = self.next_journey_datetime(vjs, request["clockwise"])
+
+        # If Kraken send a new request date time, we use it
+        # for the next call to skip current Journeys
+        if len(responses) > 0 and responses[0].HasField('next_request_date_time'):
```
xlqian: minor: if responses  and responses[0] ...

source/routing/raptor_api.cpp
```diff
@@ -994,11 +1017,11 @@ parse_datetimes(RAPTOR& raptor,
                                      pbnavitia::DATE_OUT_OF_BOUNDS,
                                      "date is not in data production period");
         }
-        datetimes.push_back(ptime);
+        datetimes.push_back(std::make_pair(ptime, datetime));
     }
     if(clockwise)
         std::sort(datetimes.begin(), datetimes.end(),
-                  [](bt::ptime dt1, bt::ptime dt2){return dt1 > dt2;});
+                  [](std::pair<bt::ptime, uint64_t> dt1, std::pair<bt::ptime, uint64_t> dt2){return dt1.first > dt2.first;});
```
xlqian: minor: `[](const std::pair<bt::ptime, uint64_t>& dt1, const std::pair<bt::ptime, uint64_t>& dt2)` (when can we have c++14? :p )

source/routing/raptor_api.cpp
```diff
@@ -1082,6 +1103,85 @@ void make_pt_response(navitia::PbCreator& pb_creator,
...
+{
+    if(journeys.size() == 0)
+        return;
+
+    auto journeys_pairs_gen = utils::make_pairs_generator(journeys);
+    std::vector<RAPTOR::Journeys::const_iterator> late_journeys;
+
+    for(auto journey_pair : journeys_pairs_gen) {
+
+        auto is_in_pair = [&](const RAPTOR::Journeys::const_iterator& it) {
```
xlqian: we need a test to keep this safe....

source/routing/journey.cpp
nonifier: That's exactly what I asked you about last time :cry: :cry: :cry: 
So `->order()` isn't enough. I will also add it to the hash function.

source/routing/journey.cpp
nonifier: Yes, I understood that similar journeys could be done like this, I understood wrong... I'll had `boost::hash_combine(seed, s.get_in_st->vj);` and `boost::hash_combine(seed, s.get_out_st->vj);` with the same logic in the comparison operator.

source/routing/raptor_api.cpp
```diff
@@ -1082,6 +1103,85 @@ void make_pt_response(navitia::PbCreator& pb_creator,
...
+void filter_late_journeys(RAPTOR::Journeys & journeys,
+                          const NightBusFilter::Params & params)
+{
+    if(journeys.size() == 0)
+        return;
+
+    auto journeys_pairs_gen = utils::make_pairs_generator(journeys);
+    std::vector<RAPTOR::Journeys::const_iterator> late_journeys;
+
+    for(auto journey_pair : journeys_pairs_gen) {
```
xlqian: JUST FOR FUN & TL;DR 

I think what you need here is a Permutation( your generator gives a Combination actually, you did a Permutation later by swapping the pair in the loop, this saves you from detecting self-repeating pair I think :smile: ), the permutation can be done by a cartesian product without repeat, a little search gave me this: https://stackoverflow.com/a/16688622/1614576, this solution handles the problem nicely and generically.
 (I prefer your implementation here which is much simpler to read :smile: )

xlqian: it doesn't handle the self repeating cases...

source/jormungandr/jormungandr/scenarios/new_default.py
benoit-bst: :white_check_mark:

source/routing/raptor_api.cpp
benoit-bst: I'll  change that (Using to_posix_timestamp function)
:white_check_mark:

source/routing/raptor_api.cpp
benoit-bst: Fixed by @nonifier 
:white_check_mark:

source/routing/raptor_api.cpp
benoit-bst: :smiley: 
:white_check_mark:

source/routing/tests/journey_test.cpp
nonifier: what's the issue with `order()` here ? How am I supposed to create the Stop Times then ?

TeXitoi: order want that the stop time be in the vj https://github.com/CanalTP/navitia/blob/dev/source/type/type.h#L710-L718

With the builder of the vj?

source/routing/raptor_api.cpp
nonifier: Yes you're right, I've chosen Combination over Permutation to avoid evaluating the swapped case when the first case succeed. 

For fun, it would be nice to generalise `pairs_generator` for a combination of n elements and submit it to boost :p

source/routing/raptor_api.cpp
nonifier: done ! :)

benoit-bst:
Jenkins, test this please !

benoit-bst:
Jenkins, test this please !

nonifier:
Jenkins, test this please !

source/jormungandr/jormungandr/scenarios/new_default.py
```diff
@@ -804,26 +810,45 @@ def fill_journeys(self, request_type, api_request, instance):
...
 
         while request is not None and \
-                ((nb_qualified_journeys < min_asked_journeys and nb_try < min_asked_journeys)
-                 or nb_try < min_journeys_calls):
+                ((nb_qualified_journeys < min_nb_journeys and nb_try < min_nb_journeys) \
+                or nb_try < min_journeys_calls):
+
             nb_try = nb_try + 1
 
+            # we take into account the option only if we have one origin_mode and destination_mode couple.
```
pbougue: for clarity, I'd change comment to "...if we have `a single` origin_mode..."

source/jormungandr/jormungandr/scenarios/new_default.py
```diff
@@ -804,26 +810,45 @@ def fill_journeys(self, request_type, api_request, instance):
...
+            if len(krakens_call) > 1:
+                request['min_nb_journeys'] = 0
+            else:
+                min_nb_journeys_left = min_nb_journeys - nb_qualified_journeys
+                request['min_nb_journeys'] = max(0, min_nb_journeys_left)
```
pbougue, line 843: Is that max really necessary? (proto is an int32, don't know for the rest of kraken)

source/jormungandr/jormungandr/scenarios/new_default.py
```diff
@@ -971,13 +996,21 @@ def create_next_kraken_request(self, request, responses):
...
         to do that we find ask the next (resp previous) query datetime
         """
-        vjs = journey_filter.get_qualified_journeys(responses)
-        if request["clockwise"]:
-            request['datetime'] = self.next_journey_datetime(vjs, request["clockwise"])
+
+        # If Kraken send a new request date time, we use it
+        # for the next call to skip current Journeys
+        if responses and responses[0].HasField('next_request_date_time'):
+            request['datetime'] = responses[0].next_request_date_time
```
pbougue, line 1003: Shouldn't we update the "next" link also with that?

source/routing/journey.cpp
pbougue: :heavy_check_mark: Done

source/jormungandr/jormungandr/scenarios/new_default.py
pbougue: (You didn't change a thing, but there might be a worst case optim)
If we are about to send the same request to kraken than the previous one, we should probably stop  even if we didn't reach those stop-conditions.
What do you think about adding that @TeXitoi ?

source/routing/raptor_api.cpp
```diff
@@ -980,14 +980,36 @@ void free_radius_filter(routing::map_stop_point_duration& sp_list,
...
+{
+    DateTime lastest_arrival    = DateTimeUtils::min;
+    DateTime earliest_departure = DateTimeUtils::inf; // clockwise
+
+    for (const auto & journey : journeys) {
+        if (clockwise) {
+            earliest_departure = std::min( earliest_departure, journey.departure_dt);
+        }
+        else {
+            lastest_arrival = std::max( lastest_arrival, journey.arrival_dt);
```
pbougue: minor: spacing

source/routing/raptor_api.cpp
```diff
@@ -980,14 +980,36 @@ void free_radius_filter(routing::map_stop_point_duration& sp_list,
     }
 }
 
+DateTime prepare_next_call_for_raptor(const RAPTOR::Journeys& journeys, const bool clockwise)
+{
+    DateTime lastest_arrival    = DateTimeUtils::min;
+    DateTime earliest_departure = DateTimeUtils::inf; // clockwise
```
pbougue, line 986: minor: this code is a bit weird to declare 2 variables and use only one...
Maybe a single `datetime_next` variable, or always update the 2 variables and use only one at the end?

source/routing/raptor_api.cpp
```diff
@@ -1167,74 +1227,143 @@ void make_response(navitia::PbCreator& pb_creator,
...
-                *departures, *destinations, init_dt, rt_level, transfer_penalty, bound, max_transfers,
+        }
+
+        // Call raptor
+        // note : Loop is for min_nb_journeys options
+        //
+        // min_nb_journeys options :
+        // Compute several loop until the number of journeys >= min_nb_journeys
+        // For each step, we find best pathes,
+        // start to the latest departure + 1 (clockwise) or the latest arrival - 1.
```
pbougue: would be nice to precise "departure/arrival `time` +/- 1`s`"

source/routing/raptor_api.cpp
```diff
@@ -1167,74 +1227,143 @@ void make_response(navitia::PbCreator& pb_creator,
...
+        // If Raptor does not return anything, we stop.
+        // If the number of Path finds is greater or equal than min_nb_journeys, we stop.
```
pbougue: english "If the number of pathes found"

source/routing/raptor_api.cpp
```diff
@@ -1167,74 +1227,143 @@ void make_response(navitia::PbCreator& pb_creator,
...
-                // Lorsqu'on demande plusieurs horaires, on garde que l'arrivée au plus tôt / départ au plus tard
-                tmp.back().request_time = datetime;
-                pathes.push_back(tmp.back());
-                bound = to_datetime(tmp.back().items.back().arrival, raptor.data);
-            } else // Lorsqu'on demande plusieurs horaires, et qu'il n'y a pas de résultat, on retourne un itinéraire vide
-                pathes.push_back(Path());
+            // Remove direct path
+            filter_direct_path(raptor_journeys);
+
+            // filter joureys that are too late.....with the magic formula...
```
pbougue: typo: "jour`n`eys"

pbougue (commented):
Did not read test so far.
Looks good, mostly minor (optional) comments.

source/routing/tests/journey_test.cpp
nonifier: I'm now using StopTimes from the builder.

source/jormungandr/jormungandr/scenarios/new_default.py
mbounabi: +1. Why 'nb_try < min_nb_journeys' ?

source/jormungandr/jormungandr/scenarios/new_default.py
nonifier: The `max` is necessary. If kraken returns more itineraries than expected, `min_nb_journeys_left` will be negative because of the difference and we having a negative `min_nb_journey` doesn't make sense.

mbounabi (commented):
Looks good!
(I trust you for the raptor part...)

source/routing/tests/routing_api_test.cpp
```diff
@@ -800,8 +850,8 @@ BOOST_AUTO_TEST_CASE(journey_array){
     auto st2 = section.stop_date_times(1);
     BOOST_CHECK_EQUAL(st1.stop_point().uri(), "stop_area:stop1");
     BOOST_CHECK_EQUAL(st2.stop_point().uri(), "stop_area:stop2");
-    BOOST_CHECK_EQUAL(st1.departure_date_time(), navitia::test::to_posix_timestamp("20120614T081100"));
-    BOOST_CHECK_EQUAL(st2.arrival_date_time(), navitia::test::to_posix_timestamp("20120614T082000"));
+    BOOST_CHECK_EQUAL(st1.departure_date_time(), navitia::test::to_posix_timestamp("20120614T091100"));
```
pbougue: Looks like it deserves a sort before more than this change

source/routing/tests/routing_api_test.cpp
```diff
@@ -2631,29 +2685,31 @@ BOOST_AUTO_TEST_CASE(journeys_with_free_radius_filter) {
...
 
+    journeys = sort_journeys(resp);
+
     // Journey 1
-    journey = resp.journeys(0);
+    journey = journeys[0];
     BOOST_REQUIRE_EQUAL(journey.sections_size(), 3);
     section = journey.sections(0);
-    BOOST_CHECK_EQUAL(section.duration(), 0);
+    BOOST_CHECK_EQUAL(section.duration(), 55);
```
pbougue, line 2696: minor: So journeys where not sorted before, here?
@TeXitoi if it looks weird to you, raise a hand ;)

source/routing/tests/routing_api_test.cpp
```diff
@@ -2744,3 +2800,585 @@ BOOST_AUTO_TEST_CASE(section_geometry_with_shapes) {
...
+            stl2[1], navitia::DateTimeUtils::set(0, "23:00"_t)
+        ));
+        j2.departure_dt = navitia::DateTimeUtils::set(0, "21:55"_t);
+        j2.arrival_dt = navitia::DateTimeUtils::set(0, "23:10"_t);
+
+        return std::make_tuple(j1, j2);
+    }
+}
+
+BOOST_AUTO_TEST_CASE(night_bus_should_treated_as_way_later)
```
pbougue: typo: should_`be_`treated (same below)

source/routing/tests/routing_api_test.cpp
```diff
@@ -2744,3 +2800,585 @@ BOOST_AUTO_TEST_CASE(section_geometry_with_shapes) {
...
+
+    nr::NightBusFilter::Params filter_params = get_default_filter_params();
+    filter_params.requested_datetime = navitia::DateTimeUtils::set(0, "23:30"_t);
+    filter_params.clockwise = false;
+
+    BOOST_CHECK_EQUAL(nr::way_later(j1, j2, filter_params), true);
+    BOOST_CHECK_EQUAL(nr::way_later(j2, j1, filter_params), false);
+}
+
+BOOST_AUTO_TEST_CASE(should_filter_night_bus_when_too_late)
```
pbougue: The test is more `night_bus_filter_is_order_agnostic` ?

source/jormungandr/jormungandr/scenarios/new_default.py
nonifier: We already talked about this during the refinement, and we've decided not to cover it in this PR.

I've created NAVP-891 to track the issue.

source/jormungandr/jormungandr/scenarios/new_default.py
pbougue: @mbounabi it's the same as before, just a little refactor on naming

source/routing/tests/routing_api_test.cpp
nonifier: :heavy_check_mark:

source/routing/tests/routing_api_test.cpp
nonifier: :heavy_check_mark:

source/routing/raptor_api.cpp
nonifier: :heavy_check_mark:

source/routing/raptor_api.cpp
nonifier: :heavy_check_mark:

source/routing/raptor_api.cpp
nonifier: :heavy_check_mark:

source/routing/raptor_api.cpp
nonifier: :heavy_check_mark:

source/routing/raptor_api.cpp
nonifier: :heavy_check_mark:

source/jormungandr/jormungandr/scenarios/new_default.py
nonifier: :heavy_check_mark:

source/jormungandr/jormungandr/scenarios/new_default.py
TeXitoi: You will never send the same request to kraken. Or you break the loop because there is no solution (or only a non-pt solution), or you have a solution and it will change the requested datetime.

source/routing/tests/routing_api_test.cpp
TeXitoi: I think they were in the order of found by the algorithm.

mbounabi (approved):

nonifier (approved):
Just because I can :)

source/jormungandr/jormungandr/scenarios/new_default.py
pbougue: You don't break the loop if there is no solution, that's my point : it looks like you don't stop until `nb_try == min_nb_journeys`

TeXitoi: https://github.com/CanalTP/navitia/blob/dev/source/jormungandr/jormungandr/scenarios/new_default.py#L831

source/routing/tests/routing_api_test.cpp
nonifier: :heavy_check_mark:

source/routing/raptor_api.h
```diff
@@ -35,8 +35,11 @@ www.navitia.io
 #include "utils/flat_enum_map.h"
 #include "type/rt_level.h"
 #include <limits>
+#include "raptor.h"
 #include "routing/routing.h"
 
+#define MAX_NB_RAPTOR_CALL 100
```
xlqian, line 41: minor: since it's only used in .cpp, for a better encapsulation, shouldn't this be in .cpp? and const static int ?

source/jormungandr/jormungandr/scenarios/new_default.py
pbougue: :heavy_check_mark: yep anyway that's fine :)

pbougue (approved):

xlqian (approved):

TeXitoi (approved):

nonifier:
And the proto changes that come with it - https://github.com/CanalTP/navitia-proto/pull/114

WilliamFoster (approved):

PR #2408, opened 2018-05-03
jomungandr: Fix attributs in swagger

kadhikari:
**Tickets:** 
https://jira.kisio.org/browse/NAVITIAII-2415

**Attributs:**

- embedded_type
- gender
- transfer_type
- data_freshness
- mode=StreetNetworkMode
- type=SectionType

### DIFF
--- a/source/jormungandr/jormungandr/interfaces/v1/fields.py
+++ b/source/jormungandr/jormungandr/interfaces/v1/fields.py
@@ -702,7 +702,7 @@ line_group["comments"] = NonNullList(NonNullNested(comment))
 
 route = deepcopy(generic_type)
 route["links"] = DisruptionLinks()
-route["is_frequence"] = fields.String
+route["is_frequence"] = fields.Boolean()
 route["line"] = PbField(line)
 route["direction_type"] = fields.String
 route["stop_points"] = NonNullList(NonNullNested(stop_point))
--- a/source/jormungandr/jormungandr/interfaces/v1/fields.py
+++ b/source/jormungandr/jormungandr/interfaces/v1/fields.py
@@ -702,7 +702,7 @@ line_group["comments"] = NonNullList(NonNullNested(comment))
 
 route = deepcopy(generic_type)
 route["links"] = DisruptionLinks()
-route["is_frequence"] = fields.Boolean()
+route["is_frequence"] = fields.String
 route["line"] = PbField(line)
 route["direction_type"] = fields.String
 route["stop_points"] = NonNullList(NonNullNested(stop_point))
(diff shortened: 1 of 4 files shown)


kadhikari:
Is_frequece modified to boolean

source/jormungandr/jormungandr/interfaces/v1/serializer/base.py
```diff
@@ -131,7 +131,7 @@ def to_value(self, value):
 
     @staticmethod
     def _get_all_possible_values(pb_type):
-        return [v.name for v in pb_type.DESCRIPTOR.values]
+        return [v.name.lower() for v in pb_type.DESCRIPTOR.values]
```
kinnou02, line 134: I think that we keep a few enum in upper case.

source/jormungandr/jormungandr/interfaces/v1/fields.py
```diff
@@ -702,7 +702,7 @@ def output(self, key, value):
 
 route = deepcopy(generic_type)
 route["links"] = DisruptionLinks()
-route["is_frequence"] = fields.String
+route["is_frequence"] = fields.Boolean()
```
kinnou02: This look like a non backward compatible change

kinnou02 (changes requested):

mbounabi (approved):

source/jormungandr/jormungandr/interfaces/v1/serializer/base.py
kadhikari: As we discussed yesterday, among all the enum  types, we have some in uppercase where all the attributs are lowercase (forced in the code). What should i do? manage some in uppercase and some in lowercase, all in lowercase or all in uppercase?

source/jormungandr/jormungandr/interfaces/v1/fields.py
kadhikari: Test swagger fails without this change.

kinnou02: In production we return a string, not a bool:
```
"is_frequence": "False",
```
it's ugly but we can't change it without breaking thing.

source/jormungandr/jormungandr/interfaces/v1/serializer/base.py
kinnou02: https://github.com/CanalTP/navitia/pull/2408/files#diff-29d0a12a94cc6560194088f98fcd72f3R130
so it's ok to lower every values

nonifier (approved):

xlqian (approved):

kinnou02 (approved):

PR #2409, opened 2018-05-14
Jormun: Cache timezone

xlqian:
a "POTENTIAL" way to cache the timezone and timestamp str...

this boosts the /route_schedule api by about 30%, 
since this api is migrating to Gormun, this PR is not necessarily useful.

But it may open a way to improve the perf of other API's endpoints.

Note that I put the timezone in the `request` context because it makes more sense, then I added a request_id to the get_timezone function so it can be cached correctly

### DIFF
--- a/source/jormungandr/jormungandr/scenarios/simple.py
+++ b/source/jormungandr/jormungandr/scenarios/simple.py
@@ -360,7 +360,6 @@ class Scenario(object):
     def _add_prev_link(self, resp, params, clockwise):
         prev_dt = self.previous_journey_datetime(resp.journeys, clockwise)
         if prev_dt is not None:
-            from flask import request
             params['datetime'] = timestamp_to_str(prev_dt, request.id)
             params['datetime_represents'] = 'arrival'
             add_link(resp, rel='prev', **params)
--- a/source/jormungandr/jormungandr/scenarios/simple.py
+++ b/source/jormungandr/jormungandr/scenarios/simple.py
@@ -360,31 +360,37 @@ class Scenario(object):
     def _add_prev_link(self, resp, params, clockwise):
         prev_dt = self.previous_journey_datetime(resp.journeys, clockwise)
         if prev_dt is not None:
-            params['datetime'] = timestamp_to_str(prev_dt)
+            from flask import request
+            params['datetime'] = timestamp_to_str(prev_dt, request.id)
             params['datetime_represents'] = 'arrival'
             add_link(resp, rel='prev', **params)
 
     def _add_next_link(self, resp, params, clockwise):
         next_dt = self.next_journey_datetime(resp.journeys, clockwise)
         if next_dt is not None:
-            params['datetime'] = timestamp_to_str(next_dt)
+            from flask import request
+            params['datetime'] = timestamp_to_str(next_dt, request.id)
             params['datetime_represents'] = 'departure'
             add_link(resp, rel='next', **params)
 
     def _add_first_last_links(self, resp, params):
         soonest_departure_ts = min(j.departure_date_time for j in resp.journeys)
-        soonest_departure = timestamp_to_datetime(soonest_departure_ts)
+        from flask import request
+        soonest_departure = timestamp_to_datetime(soonest_departure_ts, request.id)
         if soonest_departure:
             soonest_departure = soonest_departure.replace(hour=0, minute=0, second=0)
-            params['datetime'] = dt_to_str(soonest_departure)
+            from flask import request
+            params['datetime'] = dt_to_str(soonest_departure, request.id)
             params['datetime_represents'] = 'departure'
             add_link(resp, rel='first', **params)
 
         tardiest_arrival_ts = max(j.arrival_date_time for j in resp.journeys)
-        tardiest_arrival = timestamp_to_datetime(tardiest_arrival_ts)
+        from flask import request
+        tardiest_arrival = timestamp_to_datetime(tardiest_arrival_ts, request.id)
         if tardiest_arrival:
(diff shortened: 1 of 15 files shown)


source/jormungandr/jormungandr/scenarios/simple.py
```diff
@@ -360,31 +360,37 @@ def isochrone(self, request, instance):
     def _add_prev_link(self, resp, params, clockwise):
         prev_dt = self.previous_journey_datetime(resp.journeys, clockwise)
         if prev_dt is not None:
-            params['datetime'] = timestamp_to_str(prev_dt)
+            from flask import request
```
nonifier: Why don't you import request once at the top of the file ?

source/jormungandr/jormungandr/utils.py
```diff
@@ -150,21 +152,22 @@ def timestamp_to_datetime(timestamp, tz=None):
...
 
-def dt_to_str(dt):
+@functools32.lru_cache(2048)
+def dt_to_str(dt, req_id):
     return dt.strftime(DATETIME_FORMAT)
 
-
-def timestamp_to_str(timestamp):
-    dt = timestamp_to_datetime(timestamp)
+@functools32.lru_cache(2048)
```
nonifier: :+1: 

But :) how did you come up with 2048 ? Do you have any info about cache miss/hit rate ?

source/jormungandr/jormungandr/timezone.py
```diff
@@ -50,7 +51,7 @@ def set_request_instance_timezone(instance):
     if not instance.timezone:
         logger.warn("region {} has no timezone".format(instance.name))
         try:
-            g.timezone = None
+            request.timezone = None
```
kinnou02, line 54: since request is also a thread local storage does this change anything?

source/jormungandr/jormungandr/timezone.py
```diff
@@ -61,20 +62,22 @@ def set_request_instance_timezone(instance):
...
-        g.timezone = tz
+        request.timezone = tz
+
     except RuntimeError:
         pass#working outside of an application context...
 
 
-def get_timezone():
+@functools32.lru_cache(1)
+def get_timezone(req_id):
```
kinnou02, line 72: I expect this to not work well with https://github.com/CanalTP/navitia/blob/dev/source/jormungandr/jormungandr/interfaces/v1/Journeys.py#L563

source/jormungandr/jormungandr/utils.py
```diff
@@ -48,6 +48,7 @@
 import re
 import flask
 from contextlib import contextmanager
+import functools32
```
kinnou02: I suppose that you need to add it to the requirements

kinnou02 (changes requested):

pbougue:
TODO: Check how request is managed with greenlet

source/jormungandr/jormungandr/utils.py
xlqian: it comes totally from nowhere :D I'll do more test on it

xlqian: About the hit/miss, it's about ~400000 hits and ~2000 misses with the benchmark request

source/jormungandr/jormungandr/timezone.py
xlqian: you're right, It doesn't change anything, i just found it' would be more meaningful to put it in `request`

xlqian:
(This PR is closed because it was too messy to be achieved.)
 
# Two major ideas were tested this PR:

1. Cache the time zone (Fetching the timezone via flask's global takes about 20% - 30% !), that caused a lot of technical difficulties in fixing the unit tests, because of using the `flask.request.id` need a flask context, I tried mocking the module and Kaboom! 

2. Cache the function `make_properties_links`(takes ~15%), since protobuf is not hashable and not modifiable, I used a hack class to cache that function. What kept me from using this solution is, during processing the request, jormun modifies the cached objects, so the next same request crashed because of the previous modification. I tried a deepcopy when returning the cached object, it slows down a lot, and i tried a copy-on-write pattern which is too complicated.    

# In terms of performance:

### Benchmark requests

`http://127.0.0.1:5000/v1/coverage/stif/stop_areas/stop_area%3A0%3ASA%3A8738221/route_schedules?from_datetime=20180605T110259&count=100&` (100 route schedules at LaDefense)

on dev (3fc5264): **19.4s**
with timezone cache: **16.1s**
with timezone cache and make_properties_cache: **14.5s**

---------------------------------------------------------------------------------------------------------------

### Benchmark request:

`http://127.0.0.1:5000/v1/coverage/stif/stop_areas/stop_area%3A0%3ASA%3A8768600/route_schedules?from_datetime=20180605T110259&count=100&`(100 route schedules at Gare de lyon)

on dev (3fc5264): **21.2s**
with timezone cache: **16.5s**
with timezone cache and make_properties_cache: **14.9s**

---------------------------------------------------------------------------------------------------------------


### Benchmark request:

`http://127.0.0.1:5000/v1/coverage/stif/stop_areas/stop_area%3A0%3ASA%3A59566/route_schedules?from_datetime=20180605T110259&count=100&`(100 route schedules at Chatelet)

on dev (3fc5264): **41.2s**
with timezone cache: **34.15s**
with timezone cache and make_properties_cache: **30.5s**


# Possible future works

1. Use the hacker class to cache serializer functions (i.e. stop_area, stop_point)
2. We pass ~16.7% in the function `get_objets`(`make_links.py`) and ~18% in the function `get_links`(`ResourceUri.py`) due to the slowness of recursive walking through in the python dict. We can save easily ~20% if we can find an efficient way to do the recursive work. 
3. Can we avoid walking through the whole dict? Can you generate the links during the serialization?

PR #2413, opened 2018-05-17
jormungandr: Add scenario stif in tyr

kadhikari:
https://jira.kisio.org/browse/NAVP-866


source/navitiacommon/navitiacommon/default_values.py
```diff
@@ -116,6 +116,22 @@
...
 realtime_pool_size = 3
 
+# Minimum number of different suggested journeys
+min_nb_journeys = 0
+
+# Minimum number of calls to kraken
+min_journeys_calls = 1
+
+# Maximum number of successive physical modes for an itinerary
+max_successive_physical_mode = 100
```
xlqian: 100?

kadhikari: I am looking for a default value infinity. Do you have any suggestion?

source/jormungandr/jormungandr/instance.py
```diff
@@ -366,6 +366,31 @@ def realtime_pool_size(self):
         instance_db = self.get_models()
         return get_value_or_default('realtime_pool_size', instance_db, self.name)
 
+    @property
+    def min_nb_journeys(self):
```
kinnou02, line 370: if we add `min_nb_journeys` we should also add `max_nb_journeys`

source/jormungandr/jormungandr/interfaces/v1/Journeys.py
```diff
@@ -535,7 +535,16 @@ def _set_specific_params(mod):
                 args['_night_bus_filter_max_factor'] = mod.night_bus_filter_max_factor
             if args.get('_max_additional_connections') is None:
                 args['_max_additional_connections'] = mod.max_additional_connections
-
+            if args.get('min_nb_journeys') is None:
```
kinnou02, line 538: Min_nb_journeys has a default value of 0 in https://github.com/CanalTP/navitia/blob/dev/source/jormungandr/jormungandr/interfaces/v1/journey_common.py#L172 so the database value won't be use

source/jormungandr/jormungandr/instance.py
kadhikari: @kinnou02 What do you propose de default values for max_nb_journeys and max_successive_physical_mode?

source/jormungandr/jormungandr/interfaces/v1/Journeys.py
kadhikari: @kinnou02  min_nb_journeys is not initialized in journey_common.py. Your link above points to the parameter "min_nb_transfers" but not "min_nb_journeys".

kinnou02: Look like I doesn't know how to read ><

source/jormungandr/jormungandr/instance.py
kinnou02: I will put them at `None`, there is also `float("inf")` but is will be harder to store in database.
you can mix both by storing null in the database and returning `float("inf")` in that case

kadhikari: done

kadhikari:
jenkins test this please

kadhikari:
As discussed with @kinnou02  i changed the default values of two parameters: 

- max_nb_journeys = 99999

- max_successive_physical_mode = 99

xlqian (approved):

source/tyr/tyr/resources.py
```diff
@@ -298,6 +297,24 @@ def put(self, id=None, name=None):
...
+                            help='minimum number of calls to kraken', location=('json', 'values'),
+                            default=instance.min_journeys_calls)
+        parser.add_argument('max_successive_physical_mode', type=int, required=False,
+                            help='maximum number of successive physical modes in an itinerary',
+                            location=('json', 'values'))
+        parser.add_argument('final_line_filter', type=inputs.boolean,
+                            help='filter on vj using same lines and same stops', location=('json', 'values'),
+                            default=instance.final_line_filter)
+        parser.add_argument('max_extra_second_pass', type=int,
+                            help='maximum number of second pass to get more itinararies',
```
mbounabi: typo: itineraries

source/tyr/tyr/resources.py
```diff
@@ -298,6 +297,24 @@ def put(self, id=None, name=None):
         parser.add_argument('import_ntfs_in_mimir', type=inputs.boolean,
                             help='import ntfs data in global autocomplete',
                             location=('json', 'values'), default=instance.import_ntfs_in_mimir)
+        parser.add_argument('min_nb_journeys', type=int,
+                            help='minimum number of different suggested journeys', location=('json', 'values'),
+                            default=instance.min_nb_journeys)
+        parser.add_argument('max_nb_journeys', type=int, required=False,
+                            help='maximum number of different suggested journeys', location=('json', 'values'))
+        parser.add_argument('min_journeys_calls', type=int,
+                            help='minimum number of calls to kraken', location=('json', 'values'),
```
mbounabi, line 306: Why is there a parameter to set this? By setting the 'min_nb_journeys', the number of calls to kraken doesn't really matter, does it?

source/navitiacommon/navitiacommon/default_values.py
```diff
@@ -116,6 +116,25 @@
...
+# Maximum number of successive physical modes for an itinerary
+max_successive_physical_mode = 99
+
+# Minimum number of calls to kraken
+min_journeys_calls = 1
+
+# Filter on vj using same lines and same stops
+final_line_filter = False
+
+# Maximum number of second pass to get more itinararies
```
mbounabi: typo: itineraries

mbounabi (approved):

nonifier (approved):

source/tyr/tyr/resources.py
kadhikari: We don't send min_nb_journeys to kraken but is used to control the number of calls to kraken

source/jormungandr/jormungandr/scenarios/new_default.py
```diff
@@ -417,7 +417,7 @@ def culling_journeys(resp, request):
     logger = logging.getLogger(__name__)
 
     if not request["max_nb_journeys"] or request["max_nb_journeys"] >= len(resp.journeys):
-        logger.debug('No need to cull journeys')
+        logger.debug('No need to call journeys')
```
kinnou02: culling was good

source/tyr/tyr/resources.py
kinnou02: it's more a max_extra_journey_call as we want to force more than one call to kraken, except that we don't want to call kraken a third time if we have called it two times. 
Hope it helps understanding the feature...

PR #2414, opened 2018-05-17
Jormun: durations of car and bike for distributed

mbounabi:
This PR will fix latest Artemis builds failures.

### DIFF
--- a/source/jormungandr/jormungandr/scenarios/helper_classes/helper_utils.py
+++ b/source/jormungandr/jormungandr/scenarios/helper_classes/helper_utils.py
@@ -26,6 +26,7 @@
 # IRC #navitia on freenode
 # https://groups.google.com/d/forum/navitia
 # www.navitia.io
+
 from __future__ import absolute_import
 import math
 from jormungandr.street_network.street_network import StreetNetworkPathType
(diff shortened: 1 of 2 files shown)


xlqian (approved):

source/jormungandr/jormungandr/scenarios/helper_classes/helper_utils.py
```diff
@@ -120,6 +121,9 @@ def _extend_journey(pt_journey, fallback_dp, fallback_extremity):
     pt_journey.duration += aligned_fallback.journeys[0].duration
     pt_journey.durations.total = pt_journey.duration
     pt_journey.durations.walking += aligned_fallback.journeys[0].durations.walking
+    pt_journey.durations.bike += aligned_fallback.journeys[0].durations.bike
+    pt_journey.durations.car += aligned_fallback.journeys[0].durations.car
+
```
kadhikari, line 126: What about crowfly? Is it also necessary in https://github.com/CanalTP/navitia/blob/43e3fd6a1e4247b311954125fdb5d059932097de/source/jormungandr/jormungandr/scenarios/helper_classes/helper_utils.py#L53

mbounabi: In the specific case where the crowfly has a duration, it is associated to walking. Bike and car duration aren't involved.

kadhikari (approved):

pbougue (commented):
Looks good, but you may check:
* the distance (should work pretty much the same as duration)
* if artemis sees this difference in output on distributed scenario

kinnou02 (approved):

mbounabi:
@pbougue, the only case 'distances' is updated for a journey is for ridesharing for new_default scenario. Otherwise, it is never set and left to '0'. I think that another ticket is necessary to fill these fields.
I will however check if it doesn't cause any problem with Artemis.

mbounabi:
Artemis reference needs to be updated with this PR:
https://github.com/CanalTP/artemis_references/pull/123

PR #2419, opened 2018-05-25
kraken: reject non-publishable disruption while charging from RMQ

kadhikari:
https://jira.kisio.org/browse/NAVITIAII-2454

### DIFF
--- a/source/kraken/make_disruption_from_chaos.cpp
+++ b/source/kraken/make_disruption_from_chaos.cpp
@@ -321,6 +321,30 @@ make_impact(const chaos::Impact& chaos_impact, nt::PT_Data& pt_data,
     return impact;
 }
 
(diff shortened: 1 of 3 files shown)


source/kraken/make_disruption_from_chaos.cpp
```diff
@@ -321,6 +321,30 @@ make_impact(const chaos::Impact& chaos_impact, nt::PT_Data& pt_data,
     return impact;
 }
 
+bool is_publishable(transit_realtime::TimeRange publication_period,
```
kinnou02, line 324: It will be very easy to add some unittest on this function additionally to the integration tests

kadhikari: done

xlqian (approved):

kinnou02 (approved):

PR #2423, opened 2018-05-30
Doc: debug schema inconsistent on impacted_section.routes

pbougue:
`impacted_section.routes` is a list (and should be) in the API, but was a simple object in swagger.


source/jormungandr/jormungandr/interfaces/v1/serializer/pt.py
```diff
@@ -214,7 +214,7 @@ def get_stop_point(self, obj):
 class ImpactedSectionSerializer(PbNestedSerializer):
     f = PtObjectSerializer(label='from', attr='from')
     to = PtObjectSerializer()
-    routes = jsonschema.MethodField(schema_type=lambda: RouteSerializer(Many=True))
+    routes = jsonschema.MethodField(schema_type=lambda: RouteSerializer(many=True))
```
nonifier, line 217: :duck:

PR #2425, opened 2018-06-01
Kraken: Add min nb journeys for distributed

benoit-bst:
It's the continuation of the min_nb_journeys PR (https://github.com/CanalTP/navitia/pull/2405).
Adding mutual logic for both new_default and distributed scenario.

JIRA : https://jira.kisio.org/browse/NAVP-877


source/kraken/worker.cpp
```diff
@@ -742,13 +742,23 @@ void Worker::journeys(const pbnavitia::JourneysRequest &request, pbnavitia::API
...
+                                        request.clockwise(),
+                                        arg.accessibilite_params,
+                                        arg.forbidden,
+                                        arg.allowed,
+                                        arg.rt_level,
+                                        seconds{request.walking_transfer_penalty()},
+                                        request.max_duration(),
+                                        request.max_transfers(),
+                                        request.max_extra_second_pass(),
+                                        request.has_direct_path_duration() ?
```
pbougue, line 759: minor: It was there before, but it's worth a specific var here :)

source/routing/raptor_api.cpp
```diff
@@ -1193,8 +1316,8 @@ void make_response(navitia::PbCreator& pb_creator,
     worker.init(origin, {destination});
 
     // Get stop points for departure and destination
-    auto departures = get_stop_points(origin, raptor.data, worker, free_radius_from);
-    auto destinations = get_stop_points(destination, raptor.data, worker, free_radius_to, true);
+    const auto departures = get_stop_points(origin, raptor.data, worker, free_radius_from);
```
pbougue, line 1325: :+1:

source/routing/raptor_api.h
```diff
@@ -113,6 +116,9 @@ void make_isochrone(navitia::PbCreator& pb_creator,
                     int max_duration = 3600,
                     uint32_t max_transfers=std::numeric_limits<uint32_t>::max());
 
+/**
+ * @brief Used for Pt with distributed mode
+ */
 void make_pt_response(navitia::PbCreator& pb_creator,
```
pbougue, line 120: minor minor minor: You can remove the "s" from "timestamp`s`" parameter

source/routing/raptor_api.cpp
```diff
@@ -1021,6 +1021,123 @@ parse_datetimes(RAPTOR& raptor,
...
+            nb_try++;
+
+        } while (( journeys.size() + nb_direct_path < min_nb_journeys) && (nb_try < MAX_NB_RAPTOR_CALL));
+
+
+        // create date time for next
+        pb_creator.set_next_request_date_time(to_posix_timestamp(request_date_secs, raptor.data));
+
+        auto tmp_pathes = raptor.from_journeys_to_path(journeys);
+         LOG4CPLUS_DEBUG(logger, "raptor made " << tmp_pathes.size() << " Path(es)");
```
pbougue: indent

source/routing/raptor_api.cpp
```diff
@@ -1021,6 +1021,123 @@ parse_datetimes(RAPTOR& raptor,
...
+             const uint32_t max_duration,
+             const uint32_t max_transfers,
+             const size_t max_extra_second_pass,
+             const double night_bus_filter_max_factor,
+             const int32_t night_bus_filter_base_factor)
+{
+    log4cplus::Logger logger = log4cplus::Logger::getInstance(LOG4CPLUS_TEXT("logger"));
+    std::vector<Path> pathes;
+
+        // For each date time
```
pbougue: indent

pbougue (commented):
Looks good, did not check much the smarter things though (remove free_radius and generic call_raptor)
It's probably missing tests (depending on how easy it is to add them):
* On distributed scenario
* On new_default before this PR, there probably was a bug on night bus filter introduced by #2405 and it did not pop in tests then so it could be nice to add a test that would fail without this current #2425 PR

source/routing/raptor_api.cpp
```diff
@@ -1021,6 +1021,123 @@ parse_datetimes(RAPTOR& raptor,
     return datetimes;
 }
 
+std::vector<Path>
+_call_raptor(navitia::PbCreator& pb_creator,
```
xlqian: it seems that this function may be static

source/routing/raptor_api.cpp
```diff
@@ -1066,33 +1188,35 @@ void make_pt_response(navitia::PbCreator& pb_creator,
...
-            max_extra_second_pass);
-
-    for(auto & path : pathes) {
-        path.request_time = datetime;
-    }
-    LOG4CPLUS_DEBUG(logger, "raptor found " << pathes.size() << " solutions");
-    if(clockwise){
-        std::reverse(pathes.begin(), pathes.end());
-    }
+    // Call Raptor loop
```
xlqian, line 1197: :+1:

xlqian (approved):

source/routing/raptor_api.cpp
benoit-bst: static functions are functions that are only visible to other functions in the same file (more precisely the same translation unit). But we have namespace for this function :smile:

source/routing/raptor_api.cpp

source/routing/raptor_api.cpp

source/routing/raptor_api.h
benoit-bst: :white_check_mark:

source/routing/raptor_api.cpp

source/jormungandr/jormungandr/scenarios/new_default.py
```diff
@@ -847,8 +847,6 @@ def fill_journeys(self, request_type, api_request, instance):
             _tag_direct_path(new_resp)
             _tag_bike_in_pt(new_resp)
 
-            journey_filter._filter_too_late_journeys(new_resp, request)
```
nonifier, line 850: woohoo

nonifier (approved):

benoit-bst:
@pbougue, I tested before and after the suppression of "**filter_late_journeys**" in jormun.
This is what you said, if you have 2 levels of filter (in Jormun and in Kraken), we erase more than we want.
If we have only one filter in Kraken, The process is good.

So **the PR is good**, and the actual test show this :smile:

pbougue (approved):
Just waiting for the verdict on weird passing tests, but it seems fine :)
> :heavy_check_mark: Tests to be added in another PR to // work, but no problem detected on testing itself.

source/routing/raptor_api.cpp
```diff
@@ -49,6 +49,129 @@ www.navitia.io
...
+            const size_t max_extra_second_pass,
+            const double night_bus_filter_max_factor,
+            const int32_t night_bus_filter_base_factor)
+{
+    log4cplus::Logger logger = log4cplus::Logger::getInstance(LOG4CPLUS_TEXT("logger"));
+    std::vector<Path> pathes;
+
+    // For each date time
+    DateTime bound = clockwise ? DateTimeUtils::inf : DateTimeUtils::min;
+    for(const auto & datetime : datetimes) {
```
xlqian, line 84: minor: Can you just add a TODO on this line telling that it's a deprecated feature and the loop is to be removed? thx

xlqian: we can do that later...

PR #2433, opened 2018-06-11
doc: add RFC on distributed

kinnou02:


documentation/rfc/distributed.md
```diff
@@ -0,0 +1,36 @@
...
+
+
+This scenario allow us to call multiple engines for computing fallbacks. These engines can be:
+
+  - kraken
+  - geovelo
+  - here
+  - valhalla
+
+The orchestration is a bit complex, as such this documentation will try to explain what is done, and it what order.
```
stifoon: **So** this documentation will try to explain what is done, and **in** what order // or //will try to explain the sequence of what is done

documentation/rfc/distributed.md
```diff
@@ -0,0 +1,36 @@
...
+Each fallback mode can use a specific engine that is configured at the instance level.
```
stifoon: instance ? coverage ?

documentation/rfc/distributed.md
```diff
@@ -0,0 +1,36 @@
...
+
+First we search for the origin and destination object with kraken.
```
stifoon: objetcs

documentation/rfc/distributed.md
```diff
@@ -0,0 +1,36 @@
...
+
+Then we compute for each departure fallback modes the direct path between the origin and the destination with the
+appropriate engine. This is done asynchronously.
+If the public transport is disabled (max_duration=0) we return these results and stop.
```
stifoon: we := Navitia return these results

documentation/rfc/distributed.md
```diff
@@ -0,0 +1,36 @@
...
+
+We then start computing the stop points that can be accessed for each mode at both extremity.
```
stifoon: We:=Navitia
accessed:=reached

documentation/rfc/distributed.md
```diff
@@ -0,0 +1,36 @@
...
+First we search for the origin and destination object with kraken.
+
+Then we compute for each departure fallback modes the direct path between the origin and the destination with the
+appropriate engine. This is done asynchronously.
+If the public transport is disabled (max_duration=0) we return these results and stop.
+
+We then start computing the stop points that can be accessed for each mode at both extremity. 
+This is done by using a crowfly at constant speed that is computed by kraken.
+
+While this is computing we look for stop point that can be accessed in zero sec, typically the stop_point
```
stifoon: While computing, Navitia looks for stop points that can be reached in zero sec (typically stop points within the stop area)

documentation/rfc/distributed.md
```diff
@@ -0,0 +1,36 @@
...
+
+We then start computing the stop points that can be accessed for each mode at both extremity. 
+This is done by using a crowfly at constant speed that is computed by kraken.
+
+While this is computing we look for stop point that can be accessed in zero sec, typically the stop_point 
+of the stop_area. This might require a call to kraken in some case.
+
+At this point we have to wait for all of the previous task to finish, when this is done we can create a matrix of
+stop_points accessible per mode, that we update with "real" duration by using the configured engine for each mode.
+This is where we use the matrix capabilities of the streetnetwork engines.
```
stifoon: we:=navitia

documentation/rfc/distributed.md
```diff
@@ -0,0 +1,36 @@
...
+
+After this we have for each modes and at the origin and the destination a matrix of stop_points with the time required to access them.
```
stifoon: After this :

* for each modes at the origin 
* and for each modes at the destination

Navitia get a matrix of stop_points with the time required to access them.

documentation/rfc/distributed.md
```diff
@@ -0,0 +1,36 @@
...
+
+This matrix is used to start the Public Transport part of the algorithm.
+Kraken return journeys that we will need to complete by computing a streetnetwork journey
+from the origin to the departure stop_point of the first section and another journeys from de destination
```
stifoon: Too complex: maybe something like ?
Kraken returns public transport journeys between these matrix. Navitia just has to glue them with a streenetwork journey from the origin to the...

stifoon: 2 things:
* I wrote Navitia, maybe "Jormungandr" is better
* I have some diagrams to help the explanation. Do you want them?

stifoon (dismissed):

documentation/rfc/distributed.md
kinnou02: In the jormungandr routing, aka urls it is call coverage, in some endpoint it is called region, and the jormungandr source code, the object is called `Instance`...

Region is almost never used and is just kept for retro compatibility purpose at the API level.
But we freely use instance and coverage to designate it.

documentation/rfc/distributed.md
kinnou02: This is the first step to understand how distributed work, what service it call and when do it need the response of a request. id doesn't aim to be "simple".

Also this isn't a user documentation, as such I prefer the "we" because that what us, developers, get or do.
If it's the diagrams I think of, they are too generic.

stifoon: from de destination > from the destination
And then, it will be OK

documentation/rfc/distributed.md
```diff
@@ -0,0 +1,36 @@
+Scenario Distributed
+====================
+
+
+This scenario allow us to call multiple engines for computing fallbacks. These engines can be:
```
nonifier: `allowS`

documentation/rfc/distributed.md
```diff
@@ -0,0 +1,36 @@
...
+  - valhalla
+
+The orchestration is a bit complex, so this documentation will try to explain what is done, and in what order.
+Each fallback mode can use a specific engine that is configured at the instance level.
+
+First we search for the origin and destination objects with kraken.
+
+Then we compute for each departure fallback modes the direct path between the origin and the destination with the
+appropriate engine. This is done asynchronously.
+If the public transport is disabled (max_duration=0) jormungandr return these results and stop.
```
nonifier: if ~~the~~ public transport is disabled, jormungandr `returnS` [..] and `stopS`

documentation/rfc/distributed.md
```diff
@@ -0,0 +1,36 @@
...
+
+Then we start computing the stop points that can be reached for each mode at both extremity. 
+This is done by using a crowfly at constant speed that is computed by kraken.
+
+While this is computing we look for stop point that can be accessed in zero sec (typically the stop_point
+of the stop_area). This might require a call to kraken in some case.
```
nonifier: in some `caseS`

documentation/rfc/distributed.md
```diff
@@ -0,0 +1,36 @@
...
+
+At this point we have to wait for all of the previous task to finish, when this is done we can create a matrix of
```
nonifier: all the previous `taskS`

documentation/rfc/distributed.md
```diff
@@ -0,0 +1,36 @@
...
+stop_points accessible per mode, that we update with "real" duration by using the configured engine for each mode.
+This is where we use the matrix capabilities of the streetnetwork engines.
+
+After this we have for each modes at the origin and the destination a matrix of stop_points with the time required to access them.
+
+This matrix is used to start the Public Transport part of the algorithm.
+Kraken return journeys that we will need to complete by computing a streetnetwork journey
```
nonifier: Kraken `returnS`

nonifier (approved):

kadhikari (approved):

documentation/rfc/distributed.md
```diff
@@ -0,0 +1,36 @@
...
+
+Then we start computing the stop points that can be reached for each mode at both extremity. 
+This is done by using a crowfly at constant speed that is computed by kraken.
+
+While this is computing we look for stop point that can be accessed in zero sec (typically the stop_point
+of the stop_area). This might require a call to kraken in some case.
+
+At this point we have to wait for all of the previous task to finish, when this is done we can create a matrix of
+stop_points accessible per mode, that we update with "real" duration by using the configured engine for each mode.
+This is where we use the matrix capabilities of the streetnetwork engines.
```
TeXitoi: That's not true, there is no join at this point. When the 2 needed streetnetwork and the direct path for a kraken request is ready, the pt computation is launched.

documentation/rfc/distributed.md
```diff
@@ -0,0 +1,36 @@
...
+
+After this we have for each modes at the origin and the destination a matrix of stop_points with the time required to access them.
+
+This matrix is used to start the Public Transport part of the algorithm.
+Kraken return journeys that we will need to complete by computing a streetnetwork journey
+from the origin to the departure stop_point of the first section and another journeys from the destination
+stop_point of the last section to the destination.
```
TeXitoi: That's not clear that it will be done in parallel, and that common detailed fallback are computed only once.

TeXitoi (changes requested):
I think a graph of the different tasks as node and the depedencies as edges on an example will improve the readability of this.

For example, https://api.navitia.io/v1/coverage/fr-idf/journeys?from=stop_area%3AOIF%3ASA%3A59238&to=2.33071%3B48.83890&first_section_mode%5B%5D=bike&first_section_mode%5B%5D=walking&datetime=20180614T113500&last_section_mode%5B%5D=walking& looks great.

documentation/rfc/distributed.md
kinnou02: that's what I meant to say, but not what wrote :/

documentation/rfc/distributed.md
```diff
@@ -0,0 +1,41 @@
...
+First we search for the origin and destination objects with kraken.
+
+Then we compute for each departure fallback modes the direct path between the origin and the destination with the
+appropriate engine. This is done asynchronously.
+If public transport is disabled (max_duration=0) jormungandr returns these results and stops.
+
+Then we start computing the stop points that can be reached for each mode at both extremity.
+This is done by using a crowfly at constant speed that is computed by kraken.
+
+While this is computing we look for stop point that can be accessed in zero sec (typically the stop_point
```
mbounabi, line 24: stop point**s**, twice

documentation/rfc/distributed.md
```diff
@@ -0,0 +1,41 @@
...
+of the stop_area). This might require a call to kraken in some cases.
+
+The previous step will let us build a matrix of stop_points accessible per mode that will be updated with "real"
+duration by using the configured streetnetwork engine and it's matrix API.
```
mbounabi, line 28: **its**

documentation/rfc/distributed.md
```diff
@@ -0,0 +1,41 @@
...
+
+At this point we can launch the public transport part of the algorithm, we will launch one for each valid combination
+of departure mode and arrival mode. To start it we need:
+  - the origin matrix
+  - the destination matrix
+  - the direct path
+As soon as we have these three elements the PT journey for a combination will be launched.
+
+Kraken returns journeys that we will need to complete by computing a streetnetwork journey
+from the origin to the departure stop_point of the first section and another journeys from the destination
```
mbounabi, line 38: journey (no 's')

mbounabi (approved):

TeXitoi (approved):

PR #2434, opened 2018-06-11
jomungandr: synchronize fallback section objects with pt section

kadhikari:
Ticket : https://jira.kisio.org/browse/NAVP-880

### DIFF
--- a/source/jormungandr/jormungandr/scenarios/helper_classes/helper_utils.py
+++ b/source/jormungandr/jormungandr/scenarios/helper_classes/helper_utils.py
@@ -114,7 +114,7 @@ def _extend_pt_sections_with_direct_path(pt_journey, dp_journey):
         pt_journey.sections.extend(dp_journey.journeys[0].sections)
 
 
-def _extend_journey(pt_journey, fallback_dp, fallback_extremity):
+def _extend_journey(pt_journey, fallback_dp, fallback_extremity, is_start_fallback):
     """
     :param fallback_extremity: is a PeriodExtremity (a datetime and it's meaning on the fallback period)
     """
@@ -130,12 +130,12 @@ def _extend_journey(pt_journey, fallback_dp, fallback_extremity):
     pt_journey.distances.bike += aligned_fallback.journeys[0].distances.bike
     pt_journey.distances.car += aligned_fallback.journeys[0].distances.car
 
-    #Copy pt_section.origin to fallback_section.destination and pt_section.destination to fallback_section.origin
-    _, represents_start_fallback = fallback_extremity
-    if represents_start_fallback:
-        aligned_fallback.journeys[0].sections[0].origin.CopyFrom(pt_journey.sections[-1].destination)
+    # For start fallback section copy pt_section.origin to last fallback_section.destination
+    # where as for end fallback section copy last pt_section.destination to fallback_section.origin
+    if is_start_fallback:
+        aligned_fallback.journeys[0].sections[-1].destination.CopyFrom(pt_journey.sections[0].origin)
     else:
-        aligned_fallback.journeys[0].sections[0].destination.CopyFrom(pt_journey.sections[0].origin)
+        aligned_fallback.journeys[0].sections[0].origin.CopyFrom(pt_journey.sections[-1].destination)
 
     _extend_pt_sections_with_direct_path(pt_journey, aligned_fallback)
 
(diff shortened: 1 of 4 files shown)


source/jormungandr/jormungandr/scenarios/helper_classes/helper_utils.py
```diff
@@ -129,6 +130,13 @@ def _extend_journey(pt_journey, fallback_dp, fallback_extremity):
     pt_journey.distances.bike += aligned_fallback.journeys[0].distances.bike
     pt_journey.distances.car += aligned_fallback.journeys[0].distances.car
 
+    #Copy pt_section.origin to fallback_section.destination and pt_section.destination to fallback_section.origin
+    _, represents_start_fallback = fallback_extremity
+    if represents_start_fallback:
+        aligned_fallback.journeys[0].sections[0].origin.CopyFrom(pt_journey.sections[-1].destination)
+    else:
+        aligned_fallback.journeys[0].sections[0].destination.CopyFrom(pt_journey.sections[0].origin)
```
kinnou02: There is probably something that I don't understand here, because this seems to be the inverse of what we should do.

Also the `aligned_fallback` can have multiple sections, typically when using bss.
And since it look like that [we insert the section at the then sort it later](https://github.com/kadhikari/navitia/blob/f255a334999fec696dc345f1897e5efc3d7dd32f/source/jormungandr/jormungandr/scenarios/helper_classes/helper_utils.py#L202-L204), we have to be sure that the last section is the correct one.

nonifier (approved):

source/jormungandr/jormungandr/scenarios/helper_classes/helper_utils.py
```diff
@@ -129,6 +130,13 @@ def _extend_journey(pt_journey, fallback_dp, fallback_extremity):
     pt_journey.distances.bike += aligned_fallback.journeys[0].distances.bike
     pt_journey.distances.car += aligned_fallback.journeys[0].distances.car
 
+    # For start fallback section copy pt_section.origin to last fallback_section.destination
+    # where as for end fallback section copy last pt_section.destination to fallback_section.origin
+    if is_start_fallback:
```
nonifier, line 135: It would have been nice to have a test that covers this.

kadhikari: done

mbounabi (approved):

source/jormungandr/jormungandr/scenarios/helper_classes/helper_utils.py
kinnou02: I would prefer unittest

source/jormungandr/jormungandr/scenarios/helper_classes/tests/helper_classes_test.py
```diff
@@ -0,0 +1,159 @@
...
+from jormungandr.scenarios.distributed import Scenario
+from jormungandr.utils import str_to_time_stamp
+from jormungandr.street_network.tests.streetnetwork_test_utils import make_pt_object
+from jormungandr.utils import PeriodExtremity, SectionSorter
+from jormungandr.scenarios.helper_classes.helper_utils import _extend_journey
+
+
+# This function creates a jouney with 4 sections from a stop_point to stop_point.
+# Sections PUBLIC_TRANSPORT, TRANSFER, WAITING, PUBLIC_TRANSPORT
+def create_pt_journeys_with_pt():
```
mbounabi: Can't this function name just be 'create_pt_journeys' ? Or 'create_journeys_with_pt' ?

kinnou02 (approved):

source/jormungandr/jormungandr/scenarios/helper_classes/helper_utils.py
```diff
@@ -113,7 +114,7 @@ def _extend_pt_sections_with_direct_path(pt_journey, dp_journey):
         pt_journey.sections.extend(dp_journey.journeys[0].sections)
 
 
-def _extend_journey(pt_journey, fallback_dp, fallback_extremity):
+def _extend_journey(pt_journey, fallback_dp, fallback_extremity, is_start_fallback):
```
pbougue, line 117: :sob: I'm so sad (maybe coz' my code was unclear, or is it because of python?)
This param is redundant with part of `fallback_extremity`, which is a PeriodExtremity https://github.com/CanalTP/navitia/blob/8cd17a0de9b61aad7ccc0056ac6856e96b09c8b0/source/jormungandr/jormungandr/utils.py#L446
Maybe it's not clear enough, though. But then we should refactor the code or document it to clarify it.

pbougue (commented):
Late review...

source/jormungandr/jormungandr/scenarios/helper_classes/helper_utils.py
mbounabi: Indeed. :+1:

PR #2436, opened 2018-06-11
Jormungandr: Condition to stop Kraken calls

mbounabi:
https://jira.kisio.org/browse/NAVP-904

### DIFF
--- a/source/jormungandr/jormungandr/scenarios/new_default.py
+++ b/source/jormungandr/jormungandr/scenarios/new_default.py
@@ -829,7 +829,7 @@ class Scenario(simple.Scenario):
         nb_previously_qualified_journeys = 0
         last_chance_retry = False
 
-        min_journeys_calls = get_or_default(request, '_min_journeys_calls', 1)
+        min_journeys_calls = request.get('_min_journeys_calls', 1)
         max_journeys_calls = app.config.get('MAX_JOURNEYS_CALLS', 20)
         max_nb_calls = min(min_nb_journeys, max_journeys_calls)
 
@@ -895,7 +895,7 @@ class Scenario(simple.Scenario):
             nb_qualified_journeys = nb_journeys(responses)
 
             if nb_previously_qualified_journeys == nb_qualified_journeys:
-                # If there are no qualified journey in the kraken response,
+                # If there is no additional qualified journey in the kraken response,
                 # another request is sent to try to find more journeys, just in case...
                 if last_chance_retry:
                     break
--- a/source/jormungandr/jormungandr/scenarios/new_default.py
+++ b/source/jormungandr/jormungandr/scenarios/new_default.py
@@ -823,22 +823,26 @@ class Scenario(simple.Scenario):
         # is modified by create_next_kraken_request function.
         request = deepcopy(api_request)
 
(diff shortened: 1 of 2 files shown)


nonifier:
much tidier !

nonifier (approved):

source/jormungandr/jormungandr/scenarios/new_default.py
```diff
@@ -883,22 +888,28 @@ def fill_journeys(self, request_type, api_request, instance):
...
-                journey_filter.filter_shared_sections_journeys(pool2, api_request)
+            journey_filter.filter_similar_vj_journeys(journeys_pool, api_request)
 
             responses.extend(new_resp)  # we keep the error for building the response
 
             nb_qualified_journeys = nb_journeys(responses)
 
-            # We allow one more call to kraken if there is no valid journey.
+            if nb_previously_qualified_journeys == nb_qualified_journeys:
+                # If there are no qualified journey in the kraken response,
```
kadhikari: If there is no additional qualified journey in the last kraken response

kadhikari (approved):

source/jormungandr/jormungandr/scenarios/new_default.py
```diff
@@ -823,21 +823,26 @@ def fill_journeys(self, request_type, api_request, instance):
...
 
-        min_journeys_calls = get_or_default(request, '_min_journeys_calls', 1)
-
         responses = []
         nb_try = 0
         nb_qualified_journeys = 0
+        nb_previously_qualified_journeys = 0
+        last_chance_retry = False
 
+        min_journeys_calls = get_or_default(request, '_min_journeys_calls', 1)
```
kinnou02: `request.get('_min_journeys_calls', 1)` doesn't work?

kinnou02 (approved):

source/jormungandr/jormungandr/scenarios/new_default.py
mbounabi: Not tested. This line was just moved here, it isn't related to this PR.

source/jormungandr/jormungandr/scenarios/new_default.py
```diff
@@ -883,22 +888,28 @@ def fill_journeys(self, request_type, api_request, instance):
...
+            nb_previously_qualified_journeys = nb_qualified_journeys
+
             if nb_qualified_journeys == 0:
                 min_journeys_calls = max(min_journeys_calls, 2)
-
+                
         logger.debug('nb of call kraken: %i', nb_try)
 
+        if api_request['no_shared_section']:
+            journey_filter.filter_shared_sections_journeys(journey_filter.get_qualified_journeys(responses), api_request)
```
pbougue, line 911: depends if you are in debug mode, should probably be handled in final filter

mbounabi: The final filter has its purpose : to delete journeys with the same lines sequence. So this filter has nothing to do with it. I don't see the point.
Also, the journeys filtered here will be handled as every other journeys in every other filters, (mark_as_dead in the function '_filter_similar_journeys')

PR #2439, opened 2018-06-13
Kraken: Fix protofile dependencies & Log night bus filtered itinaries

nonifier:
Say goodbye to `make protobuf_files && make`, say hello to naked `make` :)

### DIFF
--- a/source/routing/raptor_api.cpp
+++ b/source/routing/raptor_api.cpp
@@ -125,6 +125,8 @@ call_raptor(navitia::PbCreator& pb_creator,
             };
             filter_late_journeys(raptor_journeys, params);
 
+            LOG4CPLUS_DEBUG(logger, "after filtering late journeys: " << raptor_journeys.size() << " solution(s) left");
+
             if (raptor_journeys.empty())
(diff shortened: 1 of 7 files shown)


benoit-bst (approved):

kinnou02 (approved):

source/routing/raptor_api.cpp
```diff
@@ -125,6 +125,8 @@ call_raptor(navitia::PbCreator& pb_creator,
             };
             filter_late_journeys(raptor_journeys, params);
 
+            LOG4CPLUS_DEBUG(logger, "after filtering late journeys: " << raptor_journeys.size() << " solution(s) left");
```
pbougue, line 128: This is not the way I understood the need for log (though it's good to see how many are filtered in a single loop).
I thought of a log describing the filtered journeys (summary).
Alternative being to tag them `to_delete` and `deleted_because_too_late_journey` when debug=True (but still send them to jormun, and then the user)...
The goal being to understand more easily why a certain journey doesn't appear in "normal" mode.

PR #2440, opened 2018-06-15
Jormun: Expose configuration file pattern for selecting json to be loaded at runtime

nonifier:
This is useful to only load a specific set of configuration files, when running Jormun with an INSTANCES_DIR that contains multiple json files.


kinnou02 (approved):
It is also possible to define instances with envvar, that what I do to add and remove instances.

mbounabi (approved):

benoit-bst (approved):

PR #2442, opened 2018-06-19
Jormun & Kraken: add timeframe_duration for journeys

xlqian:
JIRA : https://jira.kisio.org/browse/NAVP-859


benoit-bst:
Jenkins, test this please !

source/jormungandr/jormungandr/interfaces/v1/Journeys.py
```diff
@@ -500,6 +501,9 @@ def __init__(self):
         parser_get.add_argument("_no_shared_section", type=BooleanType(), default=False, hidden=True,
                                 dest="no_shared_section",
                                 help="Shared section journeys aren't returned as a separate journey")
+        parser_get.add_argument("timeframe_duration", type=int,
+                                dest="timeframe_duration",
+                                help="timeframe_duration") # more explanation!!!
```
nonifier: More explanation !!! :)

source/jormungandr/jormungandr/planner.py
```diff
@@ -33,7 +33,10 @@ def __init__(self,
                  walking_transfer_penalty=120,
                  direct_path_duration=None,
                  night_bus_filter_max_factor=None,
-                 night_bus_filter_base_factor=None):
+                 night_bus_filter_base_factor=None,
+                 min_nb_journeys=None,
```
nonifier, line 37: 😳 thanks for fixing `min_nb_journeys` with experimental

source/jormungandr/jormungandr/scenarios/new_default.py
```diff
@@ -799,6 +804,11 @@ def __init__(self):
     def fill_journeys(self, request_type, api_request, instance):
         logger = logging.getLogger(__name__)
 
+        # early return
+        if api_request['timeframe_duration'] == api_request['min_nb_journeys'] == 0:
```
nonifier: I understand that we could do an early return if `timeframe_duration` is null, but having a `min_nb_jouney=0` means that we could potentially have results still. 

Having an early return with `api_request['max_nb_journeys'] == 0` or `api_request['count'] == 0` would make more sense to me.

source/jormungandr/tests/journey_common_tests.py
```diff
@@ -1589,3 +1589,190 @@ def test_night_bus_filter_parameters_with_maximum_values_3(self):
...
         self.is_valid_journey_response(response, query)
         assert len(response['journeys']) == 2
+
+
+@dataset({"min_nb_journeys_test": {}})
+class JourneysTimeFrameDuration():
+
+    def test_timeframe_duration_simple_case(self):
+        """
+        The data contains 20 journeys (each 10 min) + 1 journeys 24H after the first.
```
nonifier: `each 10 min` -> `every 10 min`

nonifier: > +1 journey~~S~~

source/jormungandr/tests/journey_common_tests.py
```diff
@@ -1589,3 +1589,190 @@ def test_night_bus_filter_parameters_with_maximum_values_3(self):
...
+        The first journeys is 20180315T080000.
```
nonifier: > The first journey~~S~~

Same applies later in the PR ;)

source/routing/journey.cpp
```diff
@@ -119,3 +119,15 @@ size_t JourneyHash::operator()(const Journey& j) const {
 
     return seed;
 };
+
+bool JourneyCmp::operator()(const Journey& l, const Journey& r) const {
```
nonifier: It would be nice to have some unit tests in `Journey_test.cpp` for this operator.
> :heavy_check_mark: code will be removed, not needed anymore

benoit-bst:
Jenkins, test this please !

source/jormungandr/jormungandr/interfaces/v1/Journeys.py
benoit-bst: :white_check_mark:

source/jormungandr/jormungandr/planner.py
benoit-bst: :rabbit:

source/jormungandr/jormungandr/scenarios/new_default.py
benoit-bst: In fact, the *early return* concern only the timeframe_duration.
it's better like that:

```python
# early return
if (api_request['timeframe_duration'] == 0):
    logger.debug("timeframe_duration option have to be greater than 0)")
    return response_pb2.Response()
```
I'll change that

benoit-bst: I added a test

source/jormungandr/tests/journey_common_tests.py
benoit-bst: :white_check_mark:

benoit-bst: :white_check_mark:

source/jormungandr/tests/journey_common_tests.py
benoit-bst: :white_check_mark:

source/jormungandr/tests/journey_common_tests.py
```diff
@@ -1589,3 +1601,210 @@ def test_night_bus_filter_parameters_with_maximum_values_3(self):
...
+        """
+
+        # Time frame to catch only the first journeys, timeframe_duration = 10 min (60*10=600).
+        query = ('journeys?from={_from}&'
+                'to={to}&'
+                'datetime={datetime}&'
+                'timeframe_duration={timeframe_duration}&').format( _from='stop_area:sa1',
+                                                                    to='stop_area:sa3',
+                                                                    datetime="20180315T080000",
```
kinnou02: minor but I don't like quote mixing, choose between `'` and `"`.
Also, I have chosen for you: `'` :)

source/jormungandr/tests/journey_common_tests.py
```diff
@@ -1589,3 +1601,210 @@ def test_night_bus_filter_parameters_with_maximum_values_3(self):
...
+                                                                    timeframe_duration=600)
+        response = self.query_region(query)
+        assert 1 == len(response['journeys'])
+
+        # Time frame to catch only the first journeys, timeframe_duration = 1 H (60*60=3600).
```
kinnou02: comment isn't clear, it the first 6 journeys

source/jormungandr/tests/journey_common_tests.py
```diff
@@ -1589,3 +1601,210 @@ def test_night_bus_filter_parameters_with_maximum_values_3(self):
...
+                'to={to}&'
+                'datetime={datetime}&'
+                'min_nb_journeys={min_nb_journeys}&'
+                'timeframe_duration={timeframe_duration}&').format( _from='stop_area:sa1',
+                                                                    to='stop_area:sa3',
+                                                                    datetime="20180315T080000",
+                                                                    min_nb_journeys=8,
+                                                                    timeframe_duration=3600)
+        response = self.query_region(query)
+        assert 8 <= len(response['journeys'])
```
kinnou02: But we should get at least 6

source/routing/raptor_api.cpp
```diff
@@ -128,11 +249,10 @@ call_raptor(navitia::PbCreator& pb_creator,
 
             LOG4CPLUS_DEBUG(logger, "after filtering late journeys: " << raptor_journeys.size() << " solution(s) left");
 
-            if (raptor_journeys.empty())
+            // Break if empty
+            if (raptor_journeys.empty()) {
```
kinnou02: :heart:

source/routing/raptor_api.cpp
```diff
@@ -1263,7 +1402,7 @@ bool is_way_later(const Journey & j1, const Journey & j2,
 
     auto get_pseudo_duration = [&](const Journey & j) {
         auto dt = clockwise ? j.arrival_dt : j.departure_dt;
-        return std::abs(double(dt) - double(requested_dt));
+        return std::abs((double)dt - requested_dt);
```
kinnou02: old style cast

source/routing/raptor_api.cpp
```diff
@@ -51,6 +51,126 @@ namespace navitia { namespace routing {
...
+            // Case 2: we continue searching journeys until min_nb_journeys is reached
+            return true;
+        }
+        // Case 3: we have already enough journeys, but the search datetime is still inside of the time frame
+        //         we keep searching until the end of time frame
+        return is_inside(request_date_secs, timeframe_duration.get());
+    }
+    if (! timeframe_duration) {
+        // Case 4: Neither time duration nor min_nb_journeys are given, we return all raptor can find
+        return (total_nb_journeys <= 0) && (nb_try <= max_nb_raptor_call);
```
kinnou02: If `timeframe_duration` and `min_nb_journeys` aren't set we only return what has been found on the first call to raptor

kinnou02 (commented):
If jormun call kraken multiple time to get next journeys I don't think the time_frame_duration will be updated

source/jormungandr/tests/journey_common_tests.py
benoit-bst: Haha, okay accepted

source/routing/raptor_api.cpp
benoit-bst: @nonifier, ￼it will never end, haha. ￼
I change that

source/jormungandr/tests/journey_common_tests.py
benoit-bst: In 1H the **timeframe_duration** contains 6 journeys, but min_nb_journeys is active.
If min_nb_journeys is not reached, we continue until to find the minimum of journeys. It becomes the max criteria.

source/jormungandr/tests/journey_common_tests.py
benoit-bst: my mistake

source/routing/raptor_api.cpp
benoit-bst: Very good note :+1: . I'll change that like :

```c++
if (! timeframe_duration) {
    // Case 4: Neither time duration nor min_nb_journeys are given, we call only once Raptor
    return false;
```

Slightly more simple :smile:

source/routing/raptor_api.cpp
```diff
@@ -51,6 +51,126 @@ namespace navitia { namespace routing {
...
+
+        auto is_inside = [clockwise](const routing::Journey& j, DateTime dt){
+            return clockwise ? j.departure_dt < dt : j.arrival_dt > dt;
+        };
+
+        uint32_t count = 0;
+        boost::for_each(sorted, [&](const routing::Journey& j) {
+
+            // Erase journey if
+            // mim_nb_journeys is deactivated and journey exceed the time frame limit
```
mbounabi: mi**n**

source/routing/raptor_api.cpp
```diff
@@ -51,6 +51,126 @@ namespace navitia { namespace routing {
...
+            // or
+            // mim_nb_journeys exist and Nb journeys is greater than min_nb_journeys criteria 
+            // and journey exceed the time frame limit
```
mbounabi: exceed**s**

source/routing/raptor_api.cpp
```diff
@@ -51,6 +51,126 @@ namespace navitia { namespace routing {
...
+            return clockwise ? j.departure_dt < dt : j.arrival_dt > dt;
+        };
+
+        uint32_t count = 0;
+        boost::for_each(sorted, [&](const routing::Journey& j) {
+
+            // Erase journey if
+            // mim_nb_journeys is deactivated and journey exceed the time frame limit
+            // or
+            // mim_nb_journeys exist and Nb journeys is greater than min_nb_journeys criteria
```
mbounabi: mi**n**

source/routing/tests/routing_api_test.cpp
```diff
@@ -3378,3 +3379,297 @@ BOOST_FIXTURE_TEST_CASE(should_filter_late_journeys_with_empty_list, Night_bus_f
...
+    //-----------------------------------
+    // Case 3 :
+    // clockwise = false !!
+    // timeframe_duration = 10*60 (10 min)
+    // timeframe_max_datetime = 86400 (24H)
+    //
+    // We have a time frame duration = 10 min related to the first Journeys (08:30:00).
+    // The response must contains 1 journeys.
+    min_nb_journeys = 0;
+    timeframe_duration = 10*60*2;    // 10 min
```
mbounabi: *2 ?

source/routing/tests/routing_api_test.cpp
```diff
@@ -3378,3 +3379,297 @@ BOOST_FIXTURE_TEST_CASE(should_filter_late_journeys_with_empty_list, Night_bus_f
...
+
+    // Journey 1
+    journey = journeys[0];
+    BOOST_REQUIRE_EQUAL(journey.sections_size(), 1);
+    section = journey.sections(0);
+    BOOST_CHECK_EQUAL(section.type(), pbnavitia::SectionType::PUBLIC_TRANSPORT);
+    BOOST_CHECK_EQUAL(section.begin_date_time(), "20180309T082000"_pts);
+    BOOST_CHECK_EQUAL(section.end_date_time(), "20180309T082500"_pts);
+
+    // Journey 1
```
mbounabi: Journey 1 again?

source/jormungandr/jormungandr/interfaces/v1/Journeys.py
```diff
@@ -522,6 +526,9 @@ def get(self, region=None, lon=None, lat=None, uri=None):
             args['min_nb_journeys'] = args['count']
             args['max_nb_journeys'] = args['count']
 
+        if args.get('timeframe_duration'):
+            args['timeframe_duration'] = min(args['timeframe_duration'], timedelta(days=1).total_seconds())
```
mbounabi: Why use timedelta? The default value 'max_duration' is available. 
Plus, timedelta returns a float and the argument an integer.

source/jormungandr/jormungandr/scenarios/new_default.py
```diff
@@ -901,6 +910,10 @@ def fill_journeys(self, request_type, api_request, instance):
 
             nb_qualified_journeys = nb_journeys(responses)
 
+            if (api_request['timeframe_duration'] and api_request['min_nb_journeys'] and (nb_qualified_journeys < min_nb_journeys)):
+                # we break because the max time frame duration is reached
+                break
```
mbounabi: Does the condition match the comment?

source/jormungandr/jormungandr/interfaces/v1/Journeys.py
benoit-bst: It's is usefull to limit the *timeframe_duration* to **24h**.
*timedelta(days=1).total_seconds()* return an integer (24*60*60 = 86400 = 24H)

benoit-bst: I'll change that

source/routing/tests/routing_api_test.cpp
benoit-bst: :white_check_mark:

source/routing/tests/routing_api_test.cpp
benoit-bst: :white_check_mark:

source/routing/raptor_api.cpp

source/routing/raptor_api.cpp

source/routing/raptor_api.cpp

source/jormungandr/jormungandr/scenarios/new_default.py
benoit-bst: Yes, the comment is not exact.
:white_check_mark:

source/jormungandr/jormungandr/interfaces/v1/Journeys.py
benoit-bst: :white_check_mark:

kinnou02 (approved):

mbounabi (approved):

source/routing/raptor_api.cpp
```diff
@@ -51,6 +51,126 @@ namespace navitia { namespace routing {
...
+            return clockwise ? j.departure_dt < dt : j.arrival_dt > dt;
+        };
+
+        uint32_t count = 0;
+        boost::for_each(sorted, [&](const routing::Journey& j) {
+
+            // Erase journey if
+            // min_nb_journeys is deactivated and journey exceed the time frame limit
+            // or
+            // min_nb_journeys exist and Nb journeys is greater than min_nb_journeys criteria
```
nonifier: Shouldn't we do the culling when we exceed the `max_nb_journeys` (rather than the `min_nb_journey`) ?

source/routing/raptor_api.cpp
nonifier: @kinnou02 : I'm not sure whether you're in love with the comment or the curly braces ? :p

source/routing/raptor_api.cpp
```diff
@@ -1152,19 +1288,20 @@ void make_pt_response(navitia::PbCreator& pb_creator,
...
                       const std::vector<std::string>& forbidden,
                       const std::vector<std::string>& allowed,
                       const type::RTLevel rt_level,
                       const navitia::time_duration& transfer_penalty,
-                      uint32_t max_duration,
-                      uint32_t max_transfers,
-                      uint32_t max_extra_second_pass,
+                      const uint32_t max_duration,
+                      const uint32_t max_transfers,
+                      const uint32_t max_extra_second_pass,
```
nonifier, line 1253: Constness for real ! :+1:

source/jormungandr/jormungandr/interfaces/v1/Journeys.py
```diff
@@ -500,6 +500,9 @@ def __init__(self):
         parser_get.add_argument("_no_shared_section", type=BooleanType(), default=False, hidden=True,
                                 dest="no_shared_section",
                                 help="Shared section journeys aren't returned as a separate journey")
+        parser_get.add_argument("timeframe_duration", type=int,
+                                dest="timeframe_duration",
+                                help="time frame limit to search journeys")
```
pbougue: Here I think we need to elaborate a bit more, I propose:
"Minimum timeframe to search journeys. For example 'timeframe_duration=3600' will search for all interesting journeys departing within the next hour. Nota: 'max_nb_journeys' has priority over 'timeframe_duration' parameter."

source/jormungandr/tests/journey_common_tests.py
```diff
@@ -1589,3 +1601,210 @@ def test_night_bus_filter_parameters_with_maximum_values_3(self):
...
+        assert 6 == len(response['journeys'])
+
+        # Time frame to catch only the first two journeys before the date time, because clockwise is active.
+        query = ('journeys?from={_from}&'
+                'to={to}&'
+                'datetime={datetime}&'
+                'datetime_represents={datetime_represents}&'
+                'timeframe_duration={timeframe_duration}&').format( _from='stop_area:sa1',
+                                                                    to='stop_area:sa3',
+                                                                    datetime="20180315T083500",
```
pbougue: same on `'` here (ctrl+F `"` on this PR's file :wink: )

source/jormungandr/tests/routing_tests_experimental.py
```diff
@@ -68,8 +68,6 @@ def test_best_filtering(self):
         This feature is no longer supported"""
         pass
 
-    def test_datetime_represents_arrival(self):
-        super(TestJourneysDistributed, self).test_datetime_represents_arrival()
```
pbougue, line 73: I don't understand the original purpose of this... Useless, so you removed it, am I right ?
> :heavy_check_mark: that's it

source/routing/journey.h
```diff
@@ -85,6 +85,12 @@ struct SectionHash {
 
 typedef std::unordered_set<Journey, JourneyHash> JourneySet;
 
+
+struct JourneyCmp {
+    bool operator()(const Journey& l, const Journey& r) const;
+    bool clockwise;
```
pbougue: (if this code stays, depends on :meat_on_bone: butcher behaviour)
> Can you default-init this? Afraid of random bugs here.
> Doesn't prevent bug, only their randomness :smile: 

:heavy_check_mark: code to be removed, not used as we won't cull

source/routing/journey.cpp
```diff
@@ -119,3 +119,15 @@ size_t JourneyHash::operator()(const Journey& j) const {
...
+    if (clockwise) {
+        if (l.departure_dt != r.departure_dt) { return l.departure_dt <= r.departure_dt; }
```
pbougue: (if this code stays, depends on :meat_on_bone: butcher behaviour)
> We already have this kind of comparison for RAPTOR's pareto front (to see which journey's are better on "arrival-time" as it's the main criterion): https://github.com/CanalTP/navitia/blob/dev/source/routing/journey.cpp#L60.
> So I see 2 possible drawbacks here:
> * there should be only one Comparator if possible (depends on the desired culling behaviour, but as we compare using this everywhere...)
> * I don't see why they would be different and they are: first we compare **arrival**, then departure, then **transfers**, then waiting

:heavy_check_mark: code to be removed, not used as we won't cull

source/routing/raptor_api.cpp
pbougue: Yep, I agree (at least in part) with @nonifier.

JIRA ticket says that Kraken stops as soon as it can but returns everything it found (remember "the butcher who leaves you the extra 30 grams" :meat_on_bone: )

@kinnou02 @TeXitoi Is that decision still valid? Because we'll end up with journeys on the next day in response if we don't cull. That's a bit frustrating to compute more for nothing, but I'm OK with culling. If we cull, let's amend JIRA's ticket !

I propose:
* cull only considering timeframe, not min_nb_journeys (if both timeframe and min_nb_journey are exceeded)
* cull on max_nb_journeys (no matter the other params, probably in jormun, but apply magic-algorithm only if only one raptor call was done)

What about "next"? We might have trouble from jormun or for the integrator.

source/routing/raptor_api.cpp
```diff
@@ -51,6 +51,126 @@ namespace navitia { namespace routing {
 // Usefull to protect raptor loop
 static const uint max_nb_raptor_call = 100;
```
pbougue: I'd like UPPER_CASE on that one :pray:

source/routing/raptor_api.cpp
```diff
@@ -51,6 +51,126 @@ namespace navitia { namespace routing {
...
+    }
+    // Case 5: no min_nb_journeys is given, we find all journeys inside of the time frame
+    return is_inside(request_date_secs, timeframe_duration.get());
+}
+
+/**
+ * @brief Culling excessive journeys if we exceed limits
+ */
+static void culling_excessive_journeys(const boost::optional<uint32_t>& min_nb_journeys,
+                                       const boost::optional<uint32_t>& timeframe_duration,
```
pbougue: It's a timeframe_datetime_limit, not a duration?

source/routing/raptor_api.cpp
```diff
@@ -51,6 +51,126 @@ namespace navitia { namespace routing {
...
+ *                       \ Y
+ *      total_nb_journeys < min_nb_journeys?
+ *                   /        \
+ *                  /Y         \N
+ *              Case 2       Case 3
+ *
+ *
+ *
+ */
+static bool keep_going(const uint32_t total_nb_journeys,
```
pbougue: We have to consider max_nb_journeys here also (we absolutely want to respect it, so there's no use to keep going if we are over it already)

source/routing/raptor_api.cpp
```diff
@@ -51,6 +51,126 @@ namespace navitia { namespace routing {
...
+                       const uint32_t nb_try,
+                       const bool clockwise,
+                       const DateTime request_date_secs,
+                       const boost::optional<uint32_t>& min_nb_journeys,
+                       const boost::optional<uint32_t>& timeframe_duration) {
```
pbougue: OK, this is super confusing (and this is the case quite some times in this PR): This is a datetime, no? Absolutely not a duration.
It's the same type underneath, but at least we need to name it correctly (we should definitely type it).

source/routing/raptor_api.cpp
```diff
@@ -51,6 +51,126 @@ namespace navitia { namespace routing {
...
+        LOG4CPLUS_DEBUG(logger, "after culling excessive journeys: " << journeys.size() << " solution(s) left");
+    }
+}
+
+/**
+ * @brief Update timeframe_duration in raptor referential
+ */
+static void update_timeframe_parameters(const uint32_t request_date_secs,
+                                        const bool clockwise,
+                                        boost::optional<uint32_t>& timeframe_duration)
```
pbougue: From what I understand (naming is hard here) this looks ugly: input is a duration, output is a datetime...
We are changing :carrot: in :potato: , and I don't like :stew:

source/routing/raptor_api.cpp
```diff
@@ -51,6 +51,126 @@ namespace navitia { namespace routing {
...
+{
+    if (timeframe_duration) {
+        if (clockwise) {
+            *timeframe_duration += request_date_secs;
```
pbougue: Why not testing on overflow also here? (it's "underflow" we test below, if I'm correct)

> it's from the beginning of validity_period, it's usefull, no change to be done.

source/jormungandr/jormungandr/scenarios/new_default.py
```diff
@@ -799,6 +803,11 @@ def __init__(self):
     def fill_journeys(self, request_type, api_request, instance):
         logger = logging.getLogger(__name__)
 
+        # early return
+        if (api_request['timeframe_duration'] == 0):
```
pbougue: this test should probably be `<=`
and we should error, no?

source/jormungandr/jormungandr/scenarios/new_default.py
```diff
@@ -901,6 +910,11 @@ def fill_journeys(self, request_type, api_request, instance):
 
             nb_qualified_journeys = nb_journeys(responses)
 
+            if api_request['timeframe_duration']:
+                # If timeframe_duration is active, it is useless to recall Kraken,
+                # it has already sent back that he could
```
pbougue: typo: `w`hat he could

source/jormungandr/jormungandr/scenarios/utils.py
```diff
@@ -280,13 +280,13 @@ def change_ids(new_journeys, journey_count):
         for i in range(len(ticket.section_id)):
             ticket.section_id[i] = ticket.section_id[i] + '_' + six.text_type(journey_count)
 
-    for new_journey in new_journeys.journeys:
+    for count, new_journey in enumerate(new_journeys.journeys):
         for i in range(len(new_journey.fare.ticket_id)):
             new_journey.fare.ticket_id[i] = new_journey.fare.ticket_id[i] \
-                                            + '_' + six.text_type(journey_count)
+                                            + '_' + six.text_type(journey_count+count)
```
pbougue: If I understand correctly, you're correcting an unrelated bug here (but I'm a bit lost).
Isn't the problem inside kraken, more?
Why add instead of concatenate?

pbougue (changes requested):
At least the naming is too confusing, and we still have some functional (butcher :meat_on_bone: ) issues to answer... :disappointed: 
> Will review tests afterward.

source/tests/min_nb_journeys_test.cpp
```diff
@@ -38,25 +38,41 @@ www.navitia.io
...
- *   Start x        - 8:03                - 8:22        x Stop
- *         |        - 8:04                - 8:23        |
- *         |                                            |
- *         |          We add an other way to test       |
- *         \         similar journeys filtering         /
- *          \        (Faster but with more walking)    /
- *           \       sp1                   sp2        /
- *            \-------x---------------------x--------/
- *                 - 8:05                - 8:15
+ *                   sp1    (A Line & C line)   sp2
```
pbougue: it's `sa1:sp1` and `sa3:sp1`

source/tests/min_nb_journeys_test.cpp
```diff
@@ -38,25 +38,41 @@ www.navitia.io
...
+ *            /-------x--------------------------x--------\
+ *           /                                             \
+ *          /                                               \
+ *   Start x                                                 x Stop
+ *          \                                               /
+ *           \       sp1                        sp2        /
```
pbougue: it's `sa1:sp2` and `sa3:sp2`

source/tests/min_nb_journeys_test.cpp
```diff
@@ -82,13 +98,35 @@ int main(int argc, const char* const argv[]) {
...
+	// All those vj start their service on 20180315
+    auto dep_time = "08:00:00"_t;
+    auto arr_time = "08:05:00"_t;
+    for (int nb = 0; nb < 20; ++nb) {
+        b.vj("C", "1000000", "", false, "vjC_" + std::to_string(nb))
+                ("stop_point:sa1:s1", dep_time + nb * "00:10:00"_t)
+                ("stop_point:sa3:s1", arr_time + nb * "00:10:00"_t);
+    }
+    // Add a VJ 24H after to test the time frame duration max limit (24H)
+    b.vj("C", "10000000", "", false, "vjC_out_of_24_bound")
```
pbougue, line 123: the validity_pattern should be moved of 1 day, no?
> OK, it's actually 1 day after (sorry)

source/tests/min_nb_journeys_test.cpp
```diff
@@ -82,13 +98,35 @@ int main(int argc, const char* const argv[]) {
 	b.lines["A"]->text_color = "FFFFFF";
 
     // journey 2
+    // We add an other way to test similar journeys filtering
+    // (Faster but with more walking)
     b.vj("B", "111111", "", false, "vj5")("stop_point:sa1:s2", "8:00"_t, "8:05"_t)("stop_point:sa3:s2", "8:15"_t, "8:16"_t);
 
     // Line B
```
pbougue: this comment can be moved above

source/tests/min_nb_journeys_test.cpp
```diff
@@ -82,13 +98,35 @@ int main(int argc, const char* const argv[]) {
...
+        b.vj("C", "1000000", "", false, "vjC_" + std::to_string(nb))
+                ("stop_point:sa1:s1", dep_time + nb * "00:10:00"_t)
+                ("stop_point:sa3:s1", arr_time + nb * "00:10:00"_t);
+    }
+    // Add a VJ 24H after to test the time frame duration max limit (24H)
+    b.vj("C", "10000000", "", false, "vjC_out_of_24_bound")
+                ("stop_point:sa1:s1", "08:00:00"_t)
+                ("stop_point:sa3:s1", "08:05:00"_t);
+
+    // Line B
```
pbougue: typo: line `C`
and this can be moved above

source/routing/tests/routing_api_test.cpp
```diff
@@ -3378,3 +3379,297 @@ BOOST_FIXTURE_TEST_CASE(should_filter_late_journeys_with_empty_list, Night_bus_f
...
+
+/**
+ * @brief This test aims to highlight the timeframe_duration option on a journey request
+ *
+ * The data for testing :
+ *
+ * Vehicle journey with multiple stops times
+ * date : 2018 03 09
+ *
+ *                   sp1                   sp2
```
pbougue: `sa1` and `sa3`

source/routing/tests/routing_api_test.cpp
```diff
@@ -3378,3 +3379,297 @@ BOOST_FIXTURE_TEST_CASE(should_filter_late_journeys_with_empty_list, Night_bus_f
...
+
+    std::vector<std::string> forbidden;
+    ng::StreetNetwork sn_worker(*b.data->geo_ref);
+
+    auto * data_ptr = b.data.get();
+
+    //-----------------------------------
+    // Case 1 :
+    // clockwise = true
+    // timeframe_duration = 0
```
pbougue, line 3480: minor: not super-realistic as it is early-returned

source/jormungandr/jormungandr/scenarios/new_default.py
pbougue: Ah we may consider it as just "no contraint on timeframe"...
But what if we also have a min_nb_journeys ? Early return is weird if we don't error.
And by error, I mean explicit "your param is wrong".

source/routing/tests/routing_api_test.cpp
```diff
@@ -3378,3 +3379,297 @@ BOOST_FIXTURE_TEST_CASE(should_filter_late_journeys_with_empty_list, Night_bus_f
...
+
+    //-----------------------------------
+    // Case 2 :
+    // clockwise = true
+    // min_nb_journeys = none
+    // timeframe_duration = 10*60 (10 min)
+    // timeframe_max_datetime = 86400 (24H)
+    //
+    // We have a time frame duration = 10 min related to the first Journeys (08:00:00).
+    // The response must contains 1 journeys
```
pbougue: typo: contain (no `s`)

source/routing/tests/routing_api_test.cpp
```diff
@@ -3378,3 +3379,297 @@ BOOST_FIXTURE_TEST_CASE(should_filter_late_journeys_with_empty_list, Night_bus_f
...
+    // clockwise = true
+    // timeframe_duration = 0
+    // timeframe_max_datetime = 86400 (24H)
+    //
+    // In this case, the time frame duration is equal to 0.
+    // In the response, we want 0 journeys, all is filtered.
+    bool clockwise = true;
+
+    // send request
+    navitia::PbCreator pb_creator2(data_ptr, "20180309T080000"_dt, null_time_period);
```
pbougue: minor: naming pb_creator`_case_1` would be clearer (same below)

source/routing/raptor_api.cpp
```diff
@@ -51,6 +51,129 @@ namespace navitia { namespace routing {
...
+                       const uint32_t nb_try,
+                       const bool clockwise,
+                       const DateTime request_date_secs,
+                       const boost::optional<uint32_t>& min_nb_journeys,
+                       const boost::optional<uint32_t>& timeframe_duration) {
+    if(nb_try > max_nb_raptor_call) {
+        return false;
+    }
+
+    auto is_inside = [clockwise](DateTime lhs, DateTime rhs){
```
pbougue: can be a clockwise`&`

source/routing/tests/routing_api_test.cpp
```diff
@@ -3378,3 +3379,297 @@ BOOST_FIXTURE_TEST_CASE(should_filter_late_journeys_with_empty_list, Night_bus_f
...
+                  0,
+                  boost::none,
+                  1.5,
+                  900,
+                  timeframe_duration);
+
+    // get the response
+    resp = pb_creator3.get_response();
+    BOOST_REQUIRE_EQUAL(resp.response_type(), pbnavitia::ITINERARY_FOUND);
+    BOOST_REQUIRE_EQUAL(resp.journeys_size(), 1);
```
pbougue: This is a edge case, so timeframe is [a; b[ , right?
If it's the case, it should be documented a bit, not only tested.
And I think most of the tests shouldn't be on edge, as it is now the case.

source/routing/tests/routing_api_test.cpp
```diff
@@ -3378,3 +3379,297 @@ BOOST_FIXTURE_TEST_CASE(should_filter_late_journeys_with_empty_list, Night_bus_f
...
+    // get the response
+    auto resp = pb_creator2.get_response();
+    BOOST_REQUIRE_EQUAL(resp.response_type(), pbnavitia::ITINERARY_FOUND);
+    BOOST_REQUIRE_EQUAL(resp.journeys_size(), 1);
+
+    //-----------------------------------
+    // Case 2 :
+    // clockwise = true
+    // min_nb_journeys = none
+    // timeframe_duration = 10*60 (10 min)
```
pbougue: remove it, or move it next to the other comment...

source/routing/tests/routing_api_test.cpp
```diff
@@ -3378,3 +3379,297 @@ BOOST_FIXTURE_TEST_CASE(should_filter_late_journeys_with_empty_list, Night_bus_f
...
+    BOOST_REQUIRE_EQUAL(journey.sections_size(), 1);
+    auto section = journey.sections(0);
+    BOOST_CHECK_EQUAL(section.type(), pbnavitia::SectionType::PUBLIC_TRANSPORT);
+    BOOST_CHECK_EQUAL(section.begin_date_time(), "20180309T080000"_pts);
+    BOOST_CHECK_EQUAL(section.end_date_time(), "20180309T080500"_pts);
+
+    //-----------------------------------
+    // Case 3 :
+    // clockwise = false !!
+    // timeframe_duration = 10*60 (10 min)
```
pbougue: wrong comment

source/routing/tests/routing_api_test.cpp
```diff
@@ -3378,3 +3379,297 @@ BOOST_FIXTURE_TEST_CASE(should_filter_late_journeys_with_empty_list, Night_bus_f
...
+    BOOST_CHECK_EQUAL(section.begin_date_time(), "20180309T083000"_pts);
+    BOOST_CHECK_EQUAL(section.end_date_time(), "20180309T083500"_pts);
+
+    //-----------------------------------
+    // Case 4 :
+    // clockwise = true
+    // timeframe_duration = 60*60*1 (1H)
+    // timeframe_max_datetime = 86400   (24H)
+    //
+    // We have a time frame duration = 30 min related to the first Journeys (08:30:00) and
```
pbougue: wrong comment

source/routing/tests/routing_api_test.cpp
```diff
@@ -3378,3 +3379,297 @@ BOOST_FIXTURE_TEST_CASE(should_filter_late_journeys_with_empty_list, Night_bus_f
...
+    // a min_nb_journeys = 20.
+    // We don't have 20 journeys in 30 min, so we continue until 20 journeys (min_nb_journeys is the main criteria).
```
pbougue: same: 1H

source/routing/tests/routing_api_test.cpp
```diff
@@ -3378,3 +3379,297 @@ BOOST_FIXTURE_TEST_CASE(should_filter_late_journeys_with_empty_list, Night_bus_f
...
+                  destination,
+                  {ntest::to_posix_timestamp("20180309T080000")},
+                  clockwise,
+                  navitia::type::AccessibiliteParams(),
+                  forbidden,
+                  {},
+                  sn_worker,
+                  nt::RTLevel::Base,
+                  2_min,
+                  60*60,
```
pbougue: would be nice to name it

source/routing/tests/routing_api_test.cpp
```diff
@@ -3378,3 +3379,297 @@ BOOST_FIXTURE_TEST_CASE(should_filter_late_journeys_with_empty_list, Night_bus_f
...
+    // to
+    navitia::type::EntryPoint destination(navitia::type::Type_e::StopPoint, "stop_point:sa3:s1");
+
+    std::vector<std::string> forbidden;
+    ng::StreetNetwork sn_worker(*b.data->geo_ref);
+
+    auto * data_ptr = b.data.get();
+
+    //-----------------------------------
+    // Case 1 :
```
pbougue, line 3478: It would be cool to add a test `min_nb_journey=2&timeframe_duration=40*60` also

source/jormungandr/tests/journey_common_tests.py
```diff
@@ -1486,6 +1486,18 @@ class JourneysMinNbJourneys():
     Test min_nb_journeys and late journey filter
     """
 
+    def test_min_nb_journeys_options_with_minmum_value(self):
```
pbougue: typo: min`i`mum

source/jormungandr/tests/journey_common_tests.py
```diff
@@ -1589,3 +1601,210 @@ def test_night_bus_filter_parameters_with_maximum_values_3(self):
...
+
+        If timeframe_duration and min_nb_journeys are set to 0, the response have to return 0 journeys.
+        """
+        query = ('journeys?from={_from}&'
+                'to={to}&'
+                'datetime={datetime}&'
+                'timeframe_duration={timeframe_duration}&').format( _from='stop_area:sa1',
+                                                                    to='stop_area:sa3',
+                                                                    datetime="20180315T080000",
+                                                                    min_nb_journeys=0,
```
pbougue: we should also test `min_nb_journeys=2&timeframe_duration=0` once we agree on the behaviour

source/jormungandr/tests/journey_common_tests.py
pbougue: Agree with @benoit-bst. As we intended it in the :skull:-refinement `timeframe_duration` is actually a "`min_`timeframe_duration"... and we have to satisfy both min as they are not contradictory.
Only contradiction is with max_nb_journeys, and max wins.

BTW I will submit a renaming of the parameter to "min_timeframe_duration" to be more explicit in my upcoming PR on this PR

> No rename todo, doc will explain it's a minimum and that's it

source/jormungandr/tests/journey_common_tests.py
```diff
@@ -1589,3 +1601,210 @@ def test_night_bus_filter_parameters_with_maximum_values_3(self):
...
+                 'timeframe_duration={timeframe_duration}&').format(_from='stop_area:sa1',
+                                                                    to='stop_area:sa3',
+                                                                    datetime="20180315T080000",
+                                                                    min_nb_journeys=8,
+                                                                    timeframe_duration=87300)
+        response = self.query_region(query)
+        assert 20 == len(response['journeys'])
+
+        # min_nb_journeys = 2 and timeframe_duration = 1H (60*60 = 3600)
+        # The response have to contains 20 journeys because min_nb_journeys is verified.
```
pbougue: wrong comment `6` journeys

source/jormungandr/tests/journey_common_tests.py
```diff
@@ -1589,3 +1601,210 @@ def test_night_bus_filter_parameters_with_maximum_values_3(self):
...
+                                                                    datetime="20180315T080000",
+                                                                    min_nb_journeys="20",
+                                                                    timeframe_duration=14400)
+        response = self.query_region(query)
+        assert 20 == len(response['journeys'])
+
+        # min_nb_journeys = 21 and timeframe_duration = 4H (60*60*4 = 14400)
+        # The response have to contains 20 journeys because min_nb_journeys is not verified.
+        # Criteria is not verified, because we don't have 21 journeys in the time frame duration.
+        # We continue to search out of the bound (during 24h if the min_nb_journeys is always not verified)
```
pbougue: add `and we reach that 24h-max limit before finding 21st journey`

pbougue (commented):
tests review, to be (at least part of it) processed in a coming PR on this PR

source/routing/raptor_api.cpp
nonifier: I think there is a major problem with culling the journeys in Raptor. As long as Jormun is doing most of the filtering, Raptor is removing potential valid journeys. 

Also, Jormun is already culling the 'max_nb_journey' AFTER having filtered the solution.

nonifier: IMHO, Kraken should just return whatever Raptor has computed (with the extra journeys). Jormun should then do the filtering before the culling.

source/jormungandr/jormungandr/scenarios/new_default.py
TeXitoi: As said, we remove this early return
> :heavy_check_mark: Done in side-PR

source/routing/raptor_api.cpp
pbougue: As discussed, no culling will be done: no strong functionnal opinion on that, but it's less code.
And if it needs to be done, it will probably be in jormun.

source/jormungandr/jormungandr/scenarios/utils.py
xlqian: yes, accepted

source/routing/raptor_api.cpp
```diff
@@ -49,7 +49,129 @@ www.navitia.io
...
+ *                    /               \
+ *                   / Y               \ N
+ *          timeframe is set?        timeframe is set?
+ *            /       \                  /     \
+ *           / Y       \ N              / Y      \ N
+ *       Case 1      Case 2          Case 3     Case 4
+ *
+ *
+ */
+static bool keep_going(const uint32_t total_nb_journeys,
```
pbougue, line 77: We have to consider max_nb_journeys here also (we absolutely want to respect it, so there's no use to keep going if we are over it already)

source/routing/tests/routing_api_test.cpp
pbougue: No early return anymore. So, nice to have that one !

source/jormungandr/tests/journey_common_tests.py
```diff
@@ -1589,3 +1601,211 @@ def test_night_bus_filter_parameters_with_maximum_values_3(self):
...
+        The data contains 20 journeys (every 10 min) + 1 journey 24H after the first.
+        The first journeys is 20180315T080000.
+
+        If timeframe_duration and min_nb_journeys are set to 0, the response should return 1 journey,
+        as there is simply no contraint
+        """
+        query = ('journeys?from={_from}&'
+                'to={to}&'
+                'datetime={datetime}&'
+                'timeframe_duration={timeframe_duration}&').format( _from='stop_area:sa1',
```
pbougue: we should also test `min_nb_journeys=2&timeframe_duration=0` once we agree on the behaviour

source/routing/tests/routing_api_test.cpp
```diff
@@ -3378,3 +3379,299 @@ BOOST_FIXTURE_TEST_CASE(should_filter_late_journeys_with_empty_list, Night_bus_f
...
+                  0,
+                  boost::none,
+                  1.5,
+                  900,
+                  timeframe_duration);
+
+    // get the response
+    resp = pb_creator_case2.get_response();
+    BOOST_REQUIRE_EQUAL(resp.response_type(), pbnavitia::ITINERARY_FOUND);
+    BOOST_REQUIRE_EQUAL(resp.journeys_size(), 1);
```
pbougue: Lot of edge cases: a bit is nice, but not all, no?

pbougue (commented):
Add comments that disapeared but weren't processed.

benoit-bst (approved):

source/routing/raptor_api.cpp
xlqian: max_nb_journeys will be handled probably in jormun and MAYBE in another PR

source/jormungandr/tests/journey_common_tests.py
xlqian: done, test added at https://github.com/CanalTP/navitia/pull/2442/files#diff-0832703183fb3233e60448ac96ec726bR1697

source/routing/tests/routing_api_test.cpp
xlqian: done test added https://github.com/CanalTP/navitia/pull/2442/files#diff-d7f228dace402afdab826f03fcda090cR3689

source/jormungandr/tests/journey_common_tests.py
```diff
@@ -1589,3 +1601,228 @@ def test_night_bus_filter_parameters_with_maximum_values_3(self):
...
+class JourneysTimeFrameDuration():
+
+    def test_timeframe_duration_simple_case(self):
+        """
+        The data contains 20 journeys (every 10 min) + 1 journey 24H after the first.
+        The first journeys is 20180315T080000.
+        """
+
+        # Time frame to catch only the first journeys, timeframe_duration = 10 min (60*10=600).
+        # Even though a journey's departure is later thant the timeframe, we still keep it
```
pbougue: minor typo (in case you have to correct something else) : than`t` (one extra :tea: is nice only for British people :smile:)

source/jormungandr/tests/journey_common_tests.py
```diff
@@ -1589,3 +1601,228 @@ def test_night_bus_filter_parameters_with_maximum_values_3(self):
...
+        query = ('journeys?from={_from}&'
+                'to={to}&'
+                'datetime={datetime}&'
+                'timeframe_duration={timeframe_duration}&').format( _from='stop_area:sa1',
+                                                                    to='stop_area:sa3',
+                                                                    datetime='20180315T080000',
+                                                                    timeframe_duration=600)
+        response = self.query_region(query)
+        assert 2 == len(response['journeys'])
```
pbougue: here, shouldn't we just assert >= 1?

source/jormungandr/tests/journey_common_tests.py
```diff
@@ -1589,3 +1601,228 @@ def test_night_bus_filter_parameters_with_maximum_values_3(self):
...
+        # Time frame to catch journeys in the first hour, timeframe_duration = 1 H (60*60=3600).
+        query = ('journeys?from={_from}&'
+                'to={to}&'
+                'datetime={datetime}&'
+                'timeframe_duration={timeframe_duration}&').format( _from='stop_area:sa1',
+                                                                    to='stop_area:sa3',
+                                                                    datetime='20180315T080000',
+                                                                    timeframe_duration=3600)
+        response = self.query_region(query)
+        assert 7 == len(response['journeys'])
```
pbougue: same >= 6 ?

source/jormungandr/tests/journey_common_tests.py
```diff
@@ -1589,3 +1601,228 @@ def test_night_bus_filter_parameters_with_maximum_values_3(self):
...
+                'to={to}&'
+                'datetime={datetime}&'
+                'datetime_represents={datetime_represents}&'
+                'timeframe_duration={timeframe_duration}&').format( _from='stop_area:sa1',
+                                                                    to='stop_area:sa3',
+                                                                    datetime='20180315T083500',
+                                                                    datetime_represents='arrival',
+                                                                    timeframe_duration=1200)
+        response = self.query_region(query)
+        assert 3 == len(response['journeys'])
```
pbougue: same >= 2?

source/jormungandr/tests/journey_common_tests.py
```diff
@@ -1589,3 +1601,228 @@ def test_night_bus_filter_parameters_with_maximum_values_3(self):
...
+                'to={to}&'
+                'datetime={datetime}&'
+                'timeframe_duration={timeframe_duration}&'
+                 'min_nb_journeys={min_nb_journeys}').format( _from='stop_area:sa1',
+                                                              to='stop_area:sa3',
+                                                              datetime='20180315T080000',
+                                                              min_nb_journeys=0,
+                                                              timeframe_duration=0)
+        response = self.query_region(query)
+        assert 1 <= len(response['journeys'])
```
pbougue, line 1695: minor: weird to read an `<=` where reference value is left

source/jormungandr/tests/journey_common_tests.py
```diff
@@ -1589,3 +1601,228 @@ def test_night_bus_filter_parameters_with_maximum_values_3(self):
...
+                'to={to}&'
+                'datetime={datetime}&'
+                'min_nb_journeys={min_nb_journeys}&'
+                'timeframe_duration={timeframe_duration}&').format( _from='stop_area:sa1',
+                                                                    to='stop_area:sa3',
+                                                                    datetime='20180315T080000',
+                                                                    min_nb_journeys=2,
+                                                                    timeframe_duration=3600)
+        response = self.query_region(query)
+        assert 7 == len(response['journeys'])
```
pbougue: same >= 6?

source/routing/journey.cpp
```diff
@@ -119,3 +119,4 @@ size_t JourneyHash::operator()(const Journey& j) const {
 
     return seed;
 };
+
```
pbougue: minor: you can remove these blank lines

source/routing/journey.h
```diff
@@ -85,6 +85,8 @@ struct SectionHash {
 
 typedef std::unordered_set<Journey, JourneyHash> JourneySet;
 
+
```
pbougue: minor: you can remove these blank lines

source/routing/tests/routing_api_test.cpp
```diff
@@ -3378,3 +3379,351 @@ BOOST_FIXTURE_TEST_CASE(should_filter_late_journeys_with_empty_list, Night_bus_f
...
+
+    auto * data_ptr = b.data.get();
+
+    //-----------------------------------
+    // Case 1 :
+    // clockwise = true
+    // timeframe_duration = 0
+    // timeframe_max_bound = 24H (very late search limit)
+    //
+    // In the response, we want 0 journeys, all is filtered.
```
pbougue: outdated comment

source/routing/tests/routing_api_test.cpp
```diff
@@ -3378,3 +3379,351 @@ BOOST_FIXTURE_TEST_CASE(should_filter_late_journeys_with_empty_list, Night_bus_f
...
+                  24*60*60,
+                  10,
+                  0,
+                  0,
+                  0);
+
+    // get the response
+    auto resp = pb_creator_case1.get_response();
+    BOOST_REQUIRE_EQUAL(resp.response_type(), pbnavitia::ITINERARY_FOUND);
+    BOOST_REQUIRE_EQUAL(resp.journeys_size(), 1);
```
pbougue: I guess it's >=1

source/routing/tests/routing_api_test.cpp
```diff
@@ -3378,3 +3379,351 @@ BOOST_FIXTURE_TEST_CASE(should_filter_late_journeys_with_empty_list, Night_bus_f
...
+                  0,
+                  boost::none,
+                  1.5,
+                  900,
+                  timeframe_duration);
+
+    // get the response
+    resp = pb_creator_case2.get_response();
+    BOOST_REQUIRE_EQUAL(resp.response_type(), pbnavitia::ITINERARY_FOUND);
+    BOOST_REQUIRE_EQUAL(resp.journeys_size(), 2);
```
pbougue: `>= 1` ?

source/routing/tests/routing_api_test.cpp
```diff
@@ -3378,3 +3379,351 @@ BOOST_FIXTURE_TEST_CASE(should_filter_late_journeys_with_empty_list, Night_bus_f
...
+    //-----------------------------------
+    // Case 3 :
+    // clockwise = false !!
+    // timeframe_max_bound = 24H (very late search limit)
+    //
+    // timeframe_duration = 20*60 (20 min)
+    // We have a timeframe duration = 20 min related to the first
+    // journeys (08:30:00), as explained above, even though some journeys' arrivals are
+    // not in the interval, we still keep them
+    // The response must contain 3 journeys.
```
pbougue: to update if we update test below

source/routing/tests/routing_api_test.cpp
```diff
@@ -3378,3 +3379,351 @@ BOOST_FIXTURE_TEST_CASE(should_filter_late_journeys_with_empty_list, Night_bus_f
...
+                  0,
+                  boost::none,
+                  1.5,
+                  900,
+                  timeframe_duration);
+
+    // get the response
+    resp = pb_creator_case3.get_response();
+    BOOST_REQUIRE_EQUAL(resp.response_type(), pbnavitia::ITINERARY_FOUND);
+    BOOST_REQUIRE_EQUAL(resp.journeys_size(), 3);
```
pbougue: `>= 2` ?

source/routing/tests/routing_api_test.cpp
```diff
@@ -3378,3 +3379,351 @@ BOOST_FIXTURE_TEST_CASE(should_filter_late_journeys_with_empty_list, Night_bus_f
...
+
+    journeys = sort_journeys_by(resp, JourneySectionCompare());
+
+    // Journey 1
```
pbougue: remove that one (test with reverse indexes...)?

source/routing/tests/routing_api_test.cpp
```diff
@@ -3378,3 +3379,351 @@ BOOST_FIXTURE_TEST_CASE(should_filter_late_journeys_with_empty_list, Night_bus_f
...
+    //-----------------------------------
+    // Case 6 :
+    // clockwise = true
+    // min_nb_journeys = 2
+    // timeframe_duration = 40*60 (40min)
+    // timeframe_max_bound = 1H
+    //
+    // the reponse should contain at least 2 journeys
+    min_nb_journeys = 2;
+    timeframe_duration = 4*60; // 40 min
```
pbougue: `40`*60

source/routing/tests/routing_api_test.cpp
```diff
@@ -3378,3 +3379,351 @@ BOOST_FIXTURE_TEST_CASE(should_filter_late_journeys_with_empty_list, Night_bus_f
...
+    // timeframe_duration = 4*60*60 (4H)
+    // timeframe_max_bound = 1H (very late search limit)
+    //
+    // the limit max is less than timeframe_duration.
+    // We compute until the max limit
+    // The response must contain only 6 journeys.
+    min_nb_journeys = 0;
+    timeframe_duration = 60*60*4; // 4H
+    clockwise = true;
+    uint32_t timeframe_max_duration = 60*60;
```
pbougue, line 3684: we should test with `timeframe_max_duration < timeframe_duration`

pbougue (commented):
My point on testing `>=` is about testing what we really want functionnaly, no more.
If the behavior evolves (for example we choose to cull) we shouldn't have to maintain these tests.

source/routing/tests/routing_api_test.cpp
```diff
@@ -3378,3 +3379,434 @@ BOOST_FIXTURE_TEST_CASE(should_filter_late_journeys_with_empty_list, Night_bus_f
...
+    //-----------------------------------
+    // Case 3 :
+    // clockwise = false !!
+    // timeframe_max_bound = 24H (very late search limit)
+    //
+    // timeframe_duration = 20*60 (20 min)
+    // We have a timeframe duration = 20 min related to the first
+    // journeys (08:30:00), as explained above, even though some journeys' arrivals are
+    // not in the interval, we still keep them
+    // The response must contain at least 3 journeys.
```
pbougue: at least `2`

source/routing/tests/routing_api_test.cpp
```diff
@@ -3378,3 +3379,434 @@ BOOST_FIXTURE_TEST_CASE(should_filter_late_journeys_with_empty_list, Night_bus_f
...
+                  0,
+                  boost::none,
+                  1.5,
+                  900,
+                  timeframe_duration);
+
+    // get the response
+    resp = pb_creator_case3.get_response();
+    BOOST_REQUIRE_EQUAL(resp.response_type(), pbnavitia::ITINERARY_FOUND);
+    BOOST_REQUIRE_GE(resp.journeys_size(), 1);
```
pbougue: should be `2` here

source/routing/tests/routing_api_test.cpp
pbougue: still outdated, nothing is filtered or culled

source/routing/tests/routing_api_test.cpp
```diff
@@ -3378,3 +3379,434 @@ BOOST_FIXTURE_TEST_CASE(should_filter_late_journeys_with_empty_list, Night_bus_f
...
+    BOOST_REQUIRE_EQUAL(resp.journeys_size(), 6);
+
+    //-----------------------------------
+    // Case 6 :
+    // clockwise = true
+    // min_nb_journeys = 2
+    // timeframe_duration = 40*60 (40min)
+    // timeframe_max_bound = 1H
+    //
+    // the reponse should contain at least 2 journeys
```
pbougue: at least `4`

source/routing/tests/routing_api_test.cpp
```diff
@@ -3378,3 +3379,434 @@ BOOST_FIXTURE_TEST_CASE(should_filter_late_journeys_with_empty_list, Night_bus_f
...
+                  0,
+                  boost::none,
+                  1.5,
+                  900,
+                  timeframe_duration);
+
+    // get the response
+    resp = pb_creator_case6.get_response();
+    BOOST_REQUIRE_EQUAL(resp.response_type(), pbnavitia::ITINERARY_FOUND);
+    BOOST_REQUIRE_GE(resp.journeys_size(), 2);
```
pbougue: should be `4` here (in 40min)

source/routing/tests/routing_api_test.cpp
```diff
@@ -3378,3 +3379,434 @@ BOOST_FIXTURE_TEST_CASE(should_filter_late_journeys_with_empty_list, Night_bus_f
...
+    // get the response
+    resp = pb_creator_case7.get_response();
+    BOOST_REQUIRE_EQUAL(resp.response_type(), pbnavitia::ITINERARY_FOUND);
+    BOOST_REQUIRE_EQUAL(resp.journeys_size(), 1);
+
+    //-----------------------------------
+    // Case 8 :
+    // clockwise = true
+    // min_nb_journeys = 5
+    // timeframe_duration = 1 day
```
pbougue, line 3805: this line is describing the same param than 2 lines below but saying different things

source/routing/tests/routing_api_test.cpp
```diff
@@ -3378,3 +3379,434 @@ BOOST_FIXTURE_TEST_CASE(should_filter_late_journeys_with_empty_list, Night_bus_f
...
+    // timeframe_max_bound = 1H
+    // max_duration = 10s
+    //
+    // max_duration is too small, no solution is found
+    min_nb_journeys = 5;
+    timeframe_duration = 24*60*60; // 1 day
+    clockwise = true;
+    timeframe_max_duration = 10; // 10 second
```
pbougue: typo: second`s`

nonifier (approved):

pbougue (approved):

PR #2449, opened 2018-06-26
Ed: match stop_area and poi by using admin boundary if possible

kinnou02:
When OSM data is used we get the boundary of each admin, in that case
we want to use these boundaries to associate stop_points to an admin.

This doesn't use any kind of spacial indexes, so it might be slower than
before.

If we do not find any admin by using the boundary we still keep the old
approach as a fallback, so if there are no boundaries it works as before.

|coverage|before|before with cities|after|after with cities|
|---|-------------|---------------------|---------------------|----------------------|
|bordeaux| 7042ms| 6974ms | 417ms | 287ms |
|fr-idf| 70811ms | 50539ms | 153190ms| 118166ms|

Bordeaux has only a few admins, so the after result is very fast, also, no call to cities is done, so it has no effect on perf.
So funnily cities make the admin search faster...

### DIFF
--- a/source/ed/ed_reader.cpp
+++ b/source/ed/ed_reader.cpp
@@ -132,7 +132,7 @@ void EdReader::fill(navitia::type::Data& data, const double min_non_connected_gr
 
 void EdReader::fill_admins(navitia::type::Data& nav_data, pqxx::work& work){
     std::string request = "SELECT id, name, uri, comment, insee, level, ST_X(coord::geometry) as lon, "
-        "ST_Y(coord::geometry) as lat "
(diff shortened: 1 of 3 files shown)


TeXitoi:
The "only in cities" case is much better with cities:
 - without cities, each times, search all admins, then old fashion admin search
 - withi cities, first time search all admins, then old fashion admin search, then cities, and the others, direct found in all admins.

This might explain the performance boost.

TeXitoi (approved):

nonifier (approved):

PR #2454, opened 2018-07-02
Jormungandr: Condition for feed publishers

mbounabi:
https://jira.kisio.org/browse/NAVP-677

### DIFF
--- a/source/jormungandr/jormungandr/scenarios/new_default.py
+++ b/source/jormungandr/jormungandr/scenarios/new_default.py
@@ -732,6 +732,10 @@ def merge_responses(responses, debug):
         for fp in merged_response.feed_publishers:
             initial_feed_publishers[fp.id] = fp
 
+        # Add feed publishers from the qualified journeys only
+        # Note : For BSS, it can happen that one journey in the response has returned a walking fallback.
+        # If all other journeys in the response are to delete, the feed publisher will still be added
+        # TODO: link feed publisher to a journey instead of a response with several journeys
         merged_response.feed_publishers.extend(fp for fp in r.feed_publishers
                                                if fp.id not in initial_feed_publishers
                                                and (debug or all('to_delete' not in j.tags for j in r.journeys)))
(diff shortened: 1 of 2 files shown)


source/jormungandr/jormungandr/scenarios/new_default.py
```diff
@@ -731,7 +731,10 @@ def merge_responses(responses):
         initial_feed_publishers = {}
         for fp in merged_response.feed_publishers:
             initial_feed_publishers[fp.id] = fp
-        merged_response.feed_publishers.extend((fp for fp in r.feed_publishers if fp.id not in initial_feed_publishers))
+
+        merged_response.feed_publishers.extend(fp for fp in r.feed_publishers
```
xlqian: will it be easy to add some tests on this?

TeXitoi (commented):
OK with a unit test on merge_responses

source/jormungandr/jormungandr/scenarios/new_default.py
mbounabi: Unit test added

TeXitoi (approved):

xlqian (approved):

source/jormungandr/jormungandr/scenarios/new_default.py
```diff
@@ -731,7 +731,10 @@ def merge_responses(responses):
...
+                                               if fp.id not in initial_feed_publishers
+                                               and (debug or all('to_delete' not in j.tags for j in r.journeys)))
```
kinnou02: If a response has multiple journeys, some deleted, some not, I don't think we will get what we want if journeys of a response have different feed_publisher.
I don't think this case happen for walking, bike and car. For BSS it could happen as a bss request can return a walking solution, but since we do not use any external system for bss it doesn't really matter.

Maybe a comment to explain that it can happen is enough as we do not have the problem right now and maybe never will.

mbounabi: Then it may be a design flaw. The feed publisher should be linked to one journey if we really want to prevent this case.

kinnou02 (approved):

PR #2456, opened 2018-07-04
traffic_reports: add lines in the tf section

vpassama:
Hello,

This is a small PR where we add line objects directly in traffic_report API even for line_section disruptions.

Requesting impacted line by line_section `/lines/line:id/traffic_reports` will now display the impacted line in traffic_report.

There is a small change for tests about stop_point / stop_area objects as the disruption will now appear even if the object is not directly impacted by the line_section. The link is still on the line object and not on the stop_point / stop_area in this case.

Thank you.

### DIFF
--- a/source/jormungandr/tests/line_sections_tests.py
+++ b/source/jormungandr/tests/line_sections_tests.py
@@ -69,6 +69,41 @@ class TestLineSections(AbstractTestFixture):
         r = self.default_query('{col}/{uri}'.format(col=object_get.collection, uri=object_get.uri))
         return has_disruption(r, object_get, disruption_uri)
 
+    def has_tf_disruption(self, q, disruption):
+        """
+        checks a disruption is present in traffic report response to
+        the request provided
+        """
+        r = self.default_query(q)
+
+        return any(
+            disruption == d['disruption_id'] for d in r.get('disruptions', [])
+        )
+
+    def has_tf_linked_disruption(self, q, disruption, object_get):
+        """
+        checks a disruption is present in traffic report response to
+        the request provided
+        also check that the disruption is linked to the specified object
+        """
+        r = self.default_query(q)
+
+        if not (r.get('disruptions')
+                and r.get('traffic_reports')
+                and r['traffic_reports'][0].get(object_get.collection)):
+            return False
+
+        if disruption not in [d['disruption_id'] for d in r['disruptions']]:
+            return False
+
+        for obj in r['traffic_reports'][0][object_get.collection]:
+            if obj['id'] == object_get.uri:
+                return any(
+                    disruption == link['id'] for link in obj.get('links', [])
+                )
+
+        return False
+
     def test_line_section_structure(self):
         r = self.default_query('stop_points/C_1')
 
@@ -108,16 +143,48 @@ class TestLineSections(AbstractTestFixture):
         """
         we should be able to find the related line section disruption with /traffic_report
         """
-        def has_dis(q):
-            r = self.default_query(q)
-            return 'line_section_on_line_1' in (d['disruption_id'] for d in r['disruptions'])
 
-        assert not has_dis('stop_areas/A/traffic_reports')
-        assert not has_dis('stop_areas/B/traffic_reports')
-        assert has_dis('stop_areas/C/traffic_reports')
-        assert has_dis('stop_areas/D/traffic_reports')
-        assert has_dis('stop_areas/E/traffic_reports')
-        assert not has_dis('stop_areas/F/traffic_reports')
+        # line_section_on_line_1
+        scenario = {
+            'A': False,
+            'B': False,
+            'C': True,
+            'D': True,
+            'E': True,
+            'F': False,
+        }
+
+        for sa, result in scenario.iteritems():
+            assert result == self.has_tf_disruption(
+                'stop_areas/{}/traffic_reports'.format(sa),
+                'line_section_on_line_1',
+            )
+            assert result == self.has_tf_linked_disruption(
+                'stop_areas/{}/traffic_reports'.format(sa),
+                'line_section_on_line_1',
+                ObjGetter('stop_areas', sa)
+            )
+
+        # line_section_on_line_1_other_effect
+        scenario = {
+            'A': False,
+            'B': False,
+            'C': False,
+            'D': False,
+            'E': True,
+            'F': True,
+        }
+
+        for sa, result in scenario.iteritems():
+            assert result == self.has_tf_disruption(
+                'stop_areas/{}/traffic_reports'.format(sa),
+                'line_section_on_line_1_other_effect',
+            )
+            assert result == self.has_tf_linked_disruption(
+                'stop_areas/{}/traffic_reports'.format(sa),
+                'line_section_on_line_1_other_effect',
+                ObjGetter('stop_areas', sa)
+            )
 
     def test_traffic_reports_on_networks(self):
         def has_dis(q):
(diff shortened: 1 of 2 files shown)


buildbotctp:

kinnou02:
Jenkins, test this please.

kinnou02 (approved):

nonifier (approved):

source/jormungandr/tests/line_sections_tests.py
```diff
@@ -69,6 +69,29 @@ def has_disruption(self, object_get, disruption_uri):
         r = self.default_query('{col}/{uri}'.format(col=object_get.collection, uri=object_get.uri))
         return has_disruption(r, object_get, disruption_uri)
 
+    def has_tf_disruption(self, q, disruption, pt_obj_type):
```
pbougue: minor: has_`tr`_disruption looks better to me

source/jormungandr/tests/line_sections_tests.py
```diff
@@ -69,6 +69,29 @@ def has_disruption(self, object_get, disruption_uri):
...
+        """
+        check a disruption is present in traffic report requests
```
pbougue: in traffic_reports' response to the request provided

source/jormungandr/tests/line_sections_tests.py
```diff
@@ -69,6 +69,29 @@ def has_disruption(self, object_get, disruption_uri):
...
+        check also that the disruption is linked to a certain object type
```
pbougue: this seems difficult for integrator that want to filter and print only disruptions affecting stop_point/area

pbougue (commented):
OK, so the way I understand it, when we only have line_section's disruptions, we used to have:
```json
"traffic_reports": [
  {
    "network": {...},
    "stop_areas": [...]
  }
]
```

And now we will have lines:
```json
"traffic_reports": [
  {
    "network": {...},
    "stop_areas": [...],
    "lines": [...]
  }
]
```
Am I correct?

I guess the behavior is pretty much the same on line_reports, so we have to pay attention I think (at least it should probably match).

Also, I think we should update the RFC on that if the functional of this PR doesn't change, especially on the part of stop_areas out of line_section: https://github.com/CanalTP/navitia/blame/dev/documentation/rfc/line_sections.md#L235 .

But I'm not super-fan of having that info on stops that are not affected by line_section disruptions... Shouldn't we filter a bit more on stops that are in the line_section? poke @TeXitoi 

@vpassama if the "strict" functional need you have is on `coverage/tisseo/line/line1/traffic_reports` and `coverage/tisseo/traffic_reports` then it's OK, and I think your feature is nice. But I vote it would be nicer to filter more when requesting on stops (don't know if it's so hard).

source/jormungandr/tests/line_sections_tests.py
pbougue: looks like I'm the only one liking more `tr`, so nevermind :smiley:

vpassama:
Hi @pbougue, thank you for the review.

You're right on what is expected in this PR, we want the field `lines` to be displayed for line_sections disruptions. The need is only on `lines/line/traffic_reports` and `/traffic_reports`.

Do you need to discuss more about the way you want to filter the stop_areas for an other PR or should I do it directly in this one ?

pbougue:
So, as discussed on the phone:
As we already had the disruptions we wanted to output, let's just try to parse them after we selected them from stops.
Then from this parsing, add lines if they are line_section disruptions.

If the code is too complicated or dirty, we will try to have a look at it (and we will handle ourself the "cleanness" we want).

The final (but probably acceptable anyway) option would be to let the code as is (so don't loose it, please in case we want to have a look at it :mag: ) and just update the RFC.

source/jormungandr/tests/line_sections_tests.py
```diff
@@ -69,6 +69,29 @@ def has_disruption(self, object_get, disruption_uri):
...
+                and r.get('traffic_reports')
+                and r['traffic_reports'][0].get(pt_obj_type)):
+            return False
+
+        if disruption not in [d['disruption_id'] for d in r['disruptions']]:
+            return False
+
+        pt_obj = r['traffic_reports'][0][pt_obj_type][0]
+
+        for link in pt_obj.get('links', []):
```
xlqian: return any(disruption == link['id'] for link in pt_obj.get('links', []))

nonifier:
@vpassama, @pbougue : Hey guys, just to follow-up on this PR as it's been here for quite some time... What was the outcome? are they any actions undertaken? How do we progress from here ? :)

buildbotctp:

mbounabi:
+1 regarding @nonifier comment: @vpassama, @pbougue, what are the next steps regarding this PR?

vpassama:
Sorry I did not make any progress on this PR since the discussion I had with @pbougue (holidays + other work to do). So, as he said, we discussed about this and he told me to try some modifications with the code. I think I'll do it next week and I'll let you know if it works.

vpassama:
Hi guys, @pbougue I made the modifications (quick & dirty), it seems all the tests pass. You said you would like a dedicated function maybe, I made this as a first step and if you want something better don't hesitate to tell me.

I used two functions to test the disruptions in traffic_reports, the first is like the old one, it only checks the disruption is in the response. The second is like the one I made in my first commit, it also checks that the specified object is linked to the disruption in the traffic_reports section of the response.

Doing so, I saw that the `/lines/line:2/traffic_reports` is linked to a disruption but the [RFC](https://github.com/CanalTP/navitia/blob/dev/documentation/rfc/line_sections.md#ptref-with-traffic_reports) says it shouldn't. Anyway, this happens without this PR.

nonifier:
@pbougue : in all likelihood, you are in the best position to merge this PR if you think there is no more work to be done :)

source/jormungandr/tests/line_sections_tests.py
```diff
@@ -136,12 +203,48 @@ def test_traffic_reports_on_lines(self):
...
+            'lines/line:1/traffic_reports',
+            'line_section_on_line_1_other_effect',
+        )
+        assert self.has_tf_linked_disruption(
+            'lines/line:1/traffic_reports',
+            'line_section_on_line_1_other_effect',
+            ObjGetter('lines', 'line:1')
+        )
+
+        assert self.has_tf_disruption(
```
pbougue: I think we should `assert not` and comment code here (with a note saying it doesn't currently work).
To me adding a test on something that's not supposed to happen just enforces it, and if that behaviour changes, we don't want the maintainer to revert it's code because of this test (it would actually be nice not to have that disruption).

But I don't understand why we get this disruption and not the other one...

source/jormungandr/tests/line_sections_tests.py
```diff
@@ -69,6 +69,41 @@ def has_disruption(self, object_get, disruption_uri):
         r = self.default_query('{col}/{uri}'.format(col=object_get.collection, uri=object_get.uri))
         return has_disruption(r, object_get, disruption_uri)
 
+    def has_tf_disruption(self, q, disruption):
+        """
+        check a disruption is present in traffic report response to
```
pbougue: minor: "check`s`" or "check `that`"

source/jormungandr/tests/line_sections_tests.py
```diff
@@ -69,6 +69,41 @@ def has_disruption(self, object_get, disruption_uri):
...
+        """
+        r = self.default_query(q)
+
+        return any(
+            disruption == d['disruption_id'] for d in r.get('disruptions', [])
+        )
+
+    def has_tf_linked_disruption(self, q, disruption, object_get):
+        """
+        check a disruption is present in traffic report response to
```
pbougue: minor: "check`s`" or "check `that`"

pbougue (commented):
Looks good to me except the one where we test that the behaviour is wrong.
(other minor remarks are very optional to process)

We may probably do some things super smart now with the new ptref filter from @TeXitoi ... But not in this PR!

vpassama:
Ok @pbougue, I made the last changes, thank you for the review.

pbougue (approved):

benoit-bst:
Jenkins, test this please !

PR #2457, opened 2018-07-04
Jormun & Kraken: Remove unused filter

mbounabi:
https://jira.kisio.org/browse/NAVP-902

- Remove unused filter in Jormun (moved in Kraken)
- Remove now useless associated tests
- Add a missing use case in Kraken


source/routing/tests/routing_api_test.cpp
```diff
@@ -46,7 +46,7 @@ www.navitia.io
 struct logger_initialized {
     logger_initialized()   { init_logger(); }
 };
-BOOST_GLOBAL_FIXTURE( logger_initialized );
+BOOST_GLOBAL_FIXTURE( logger_initialized )
```
nonifier: The semicolon generates a warning, but it is necessary when compiling with an older version of BOOST. It's ugly but it will prevent you from breaking the debian 7 build ;)

source/routing/tests/routing_api_test.cpp
```diff
@@ -3370,6 +3370,38 @@ BOOST_FIXTURE_TEST_CASE(night_bus_filter_should_be_order_agnostic, Night_bus_fix
     }
 }
 
+BOOST_FIXTURE_TEST_CASE(night_bus_filter_change_parameters, Night_bus_fixture)
```
nonifier, line 3373: Nice one :)

nonifier (changes requested):

kinnou02:
There is currently a regression on this filter caused by the previous PR

nonifier (approved):

kinnou02 (approved):

PR #2459, opened 2018-07-05
jormun: Round second of datetime before calling realtime proxy to improve cachability

kinnou02:
The cache key used with realtime proxy contains the url called, so it
does contains the datetime with the seconds, it makes cache only working
for one second, that's not very good...
This PR round the second to a multiple of 30, giving us 30 seconds of
cache, this is also configurable with the `step` parameter.

We may want to rename this params as it's not very clear what it does
based on it's name, so if you have an idea I take it.

### DIFF
--- a/source/jormungandr/jormungandr/realtime_schedule/siri.py
+++ b/source/jormungandr/jormungandr/realtime_schedule/siri.py
@@ -34,7 +34,7 @@ from flask import logging
 import pybreaker
 import requests as requests
 from jormungandr import cache, app
-from jormungandr.realtime_schedule.realtime_proxy import RealtimeProxy, RealtimeProxyError
+from jormungandr.realtime_schedule.realtime_proxy import RealtimeProxy, RealtimeProxyError, floor_datetime
 from jormungandr.schedule import RealTimePassage
 import xml.etree.ElementTree as et
 import aniso8601
@@ -108,6 +108,7 @@ class Siri(RealtimeProxy):
         self.instance = instance
         self.breaker = pybreaker.CircuitBreaker(fail_max=app.config.get('CIRCUIT_BREAKER_MAX_SIRI_FAIL', 5),
                                                 reset_timeout=app.config.get('CIRCUIT_BREAKER_SIRI_TIMEOUT_S', 60))
+        self.step = kwargs.get('step', 30)
 
     def __repr__(self):
         """
@@ -223,7 +224,7 @@ class Siri(RealtimeProxy):
             </GetStopMonitoring>
           </x:Body>
         </x:Envelope>
-        """.format(dt=datetime.utcfromtimestamp(dt).isoformat(),
+        """.format(dt=floor_datetime(datetime.utcfromtimestamp(dt), self.step).isoformat(),
                    count=count,
                    RequestorRef=self.requestor_ref,
                    MessageIdentifier=message_identifier,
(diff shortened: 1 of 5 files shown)


mbounabi (approved):

nonifier (approved):

source/jormungandr/jormungandr/realtime_schedule/timeo.py
```diff
@@ -217,9 +218,9 @@ def _make_url(self, route_point, count=None, from_dt=None):
         # timeo can only handle items_per_schedule if it's < 5
         count = min(count or 5, 5)# if no value defined we ask for 5 passages
 
-        # if a custom datetime is provided we give it to timeo
+        # if a custom datetime is provided we give it to timeo but we round it to improve cachability
         dt_param = '&NextStopReferenceTime={dt}'\
-            .format(dt=self._timestamp_to_date(from_dt).strftime('%Y-%m-%dT%H:%M:%S')) \
+            .format(dt=floor_datetime(self._timestamp_to_date(from_dt), self.step).strftime('%Y-%m-%dT%H:%M:%S')) \
```
pbougue: Then we should also filter afterward, removing those in interval `[floor(from_dt), from_dt[` no?

source/jormungandr/jormungandr/realtime_schedule/timeo.py
```diff
@@ -217,9 +218,9 @@ def _make_url(self, route_point, count=None, from_dt=None):
...
             if from_dt else ''
```
pbougue: So we don't have to modify this as it is cached on url (and it never happens)... But shouldn't we enforce that "step" and cache-duration match?

source/jormungandr/jormungandr/realtime_schedule/timeo.py
```diff
@@ -73,6 +73,7 @@ def __init__(self, id, service_url, service_args, timezone,
         fail_max = kwargs.get('circuit_breaker_max_fail', app.config['CIRCUIT_BREAKER_MAX_TIMEO_FAIL'])
         reset_timeout = kwargs.get('circuit_breaker_reset_timeout', app.config['CIRCUIT_BREAKER_TIMEO_TIMEOUT_S'])
         self.breaker = pybreaker.CircuitBreaker(fail_max=fail_max, reset_timeout=reset_timeout)
+        self.step = kwargs.get('step', 30)
```
pbougue: same here, a comment would be nice.

source/jormungandr/jormungandr/realtime_schedule/siri.py
```diff
@@ -108,6 +108,7 @@ def __init__(self, id, service_url, requestor_ref,
         self.instance = instance
         self.breaker = pybreaker.CircuitBreaker(fail_max=app.config.get('CIRCUIT_BREAKER_MAX_SIRI_FAIL', 5),
                                                 reset_timeout=app.config.get('CIRCUIT_BREAKER_SIRI_TIMEOUT_S', 60))
+        self.step = kwargs.get('step', 30)
```
pbougue: I'd name it `from_datetime_step` but it's still not super sexy...
at least a comment is welcome: "A step is applied on from_datetime to discretize calls and allow caching them"

source/jormungandr/jormungandr/realtime_schedule/timeo.py
kinnou02: We could, but we didn't do it before when we were thinking that the cache was working

source/jormungandr/jormungandr/realtime_schedule/timeo.py
kinnou02: true, I will initialize `step` with `app.config['CACHE_CONFIGURATION'].get('TIMEOUT_TIMEO', 60)`

pbougue (approved):
Still not super fan of returning something before that param... I don't see any way an integration would fail on this, though, and as it's a small duration, it's probably fine.

PR #2461, opened 2018-07-05
Kraken: Fix get_best_journey when datetime represents arrival

nonifier:


benoit-bst (approved):

mbounabi (approved):
Approved with the explanation :+1:

source/routing/tests/journey_test.cpp
```diff
@@ -79,6 +81,8 @@ class JourneyTest
         Journey::Section section = make_section(vj_name, departure_time, arrival_time);
 
         j.sections.push_back(section);
+        j.departure_dt = duration_from_string(departure_time).total_seconds();
+        j.arrival_dt = duration_from_string(arrival_time).total_seconds();
```
TeXitoi, line 85: So, all the tests was useless?

TeXitoi (approved):

source/routing/tests/journey_test.cpp
nonifier: Yes !

kinnou02 (approved):

PR #2462, opened 2018-07-09
Jormun: Update ProtoResponse from geovelo

mbounabi:


source/jormungandr/jormungandr/street_network/tests/geovelo_test.py
```diff
@@ -434,3 +434,24 @@ def isochrone_geovelo_test():
...
+    geovelo = Geovelo(instance=instance,
+                      service_url='http://bob.com')
+    resp_json = direct_path_response_valid()
+
+    origin = make_pt_object(type_pb2.ADDRESS, lon=2, lat=48.2, uri='refStart1')
+    destination = make_pt_object(type_pb2.ADDRESS, lon=3, lat=48.3, uri='refEnd1')
+    fallback_extremity = PeriodExtremity(str_to_time_stamp('20161010T152000'), True)
+
+    proto_resp = geovelo._get_response(resp_json, origin, destination, fallback_extremity)
+    assert proto_resp.journeys[0].durations.total
```
kinnou02: I don't understand theses lines

kinnou02 (approved):

benoit-bst (approved):

source/jormungandr/jormungandr/street_network/tests/geovelo_test.py
pbougue: Yep, I think the next line is enough in python.
You can probably just add 3 lines in `direct_path_geovelo_test` and `direct_path_geovelo_zero_test` to test that, and put real values (not only `!=0`) so that the response consistency is checked (length of section vs. journey, etc.)

pbougue (commented):
Did you (manually?) check that this works for fallbacks: bike-PT-bike (everything is correctly added and all)?
But this can be another JIRA ticket...

source/jormungandr/jormungandr/street_network/tests/geovelo_test.py
mbounabi: In the ProtoResponse, a field is not present at all if it isn't set. So the 2 lines are there to highlight 2 different problems : is the field set and does it have a value? 
-> After discussion, seems useless. WIll change it
A small change in the data (for future tests for ex.) will force us to calculate it again. 
-> A warning in the mock data will be added to prevent changes in the data without changes in the test assert.

mbounabi:
@pbougue Of course, it was tested for first and last section. Not sure however if I had a case with both fallbacks in bike. But there will be QA testing once it is merged.

source/jormungandr/jormungandr/street_network/tests/geovelo_test.py
```diff
@@ -434,3 +436,21 @@ def isochrone_geovelo_test():
...
+    geovelo = Geovelo(instance=instance,
+                      service_url='http://bob.com')
+    resp_json = direct_path_response_valid()
+
+    origin = make_pt_object(type_pb2.ADDRESS, lon=2, lat=48.2, uri='refStart1')
+    destination = make_pt_object(type_pb2.ADDRESS, lon=3, lat=48.3, uri='refEnd1')
+    fallback_extremity = PeriodExtremity(str_to_time_stamp('20161010T152000'), True)
+
+    proto_resp = geovelo._get_response(resp_json, origin, destination, fallback_extremity)
+    assert proto_resp.journeys[0].durations.total == 3155
```
pbougue, line 454: minor-minor: Don't you want to move those up with the other tests?

pbougue (approved):

source/jormungandr/jormungandr/street_network/tests/geovelo_test.py
mbounabi: Not really : new improvement -> new test. But it can be discussed

PR #2464, opened 2018-07-10
add example for performance tests based on gatling

kinnou02:
https://gatling.io

### DIFF
--- a/gatling/readme.md
+++ b/gatling/readme.md
@@ -1,0 +1,8 @@
+How to run
+===========
+
+gatling must be run from its directory
+
+```
+./bin/gatling.sh -df <PATHTO>/navitia/gatling/data -sf <PATHTO>/navitia/gatling/simulations
+```
--- a/gatling/simulations/navitia/distributed.scala
+++ b/gatling/simulations/navitia/distributed.scala
@@ -1,0 +1,32 @@
+package navitia
+import io.gatling.core.Predef._
+import io.gatling.http.Predef._
+
+
+object Base {
+
+  val httpConf = http
+        .baseURL("http://localhost:5000/v1/coverage") //jormun
+        .acceptEncodingHeader("gzip, deflate")
+
+  def query(scn: String) = {
+    val requests = csv("journeys_fr-idf.csv").circular
+      exec().feed(requests).exec(
+        http("${type}")
+          .get("/${coverage}/journeys?${other_args}")
+          .queryParam("from", "${from}")
+          .queryParam("to", "${to}")
+          .queryParam("datetime", "${datetime}")
+          .queryParam("datetime_represents", "${datetime_represents}")
+          .queryParam("_override_scenario", scn)
+          .check(status.is(200)))
+  }
+}
+
+class Distributed extends Simulation {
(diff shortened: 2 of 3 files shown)


gatling/readme.md
```diff
@@ -0,0 +1,8 @@
+How to run
+===========
+
+gatling must be run from its directory
+
+```
+./bin/gatling.sh -df <PATHTO>/gatling_navitia/data -sf <PATHTO>/gatling_navitia/simulations
```
pbougue: Maybe add a bit on how to install gatling (from this line I'd say one should download it, not install it from packages).
Also "<PATHTO>/`navitia/gatling/`... (twice) is closer from how you've stored it.

gatling/simulations/navitia/distributed.scala
```diff
@@ -0,0 +1,32 @@
...
+          .queryParam("to", "${to}")
+          .queryParam("datetime", "${datetime}")
+          .queryParam("datetime_represents", "${datetime_represents}")
+          .queryParam("_override_scenario", scn)
+          .check(status.is(200)))
+  }
+}
+
+class Distributed extends Simulation {
+    setUp(scenario("distributed").exec(Base.query("distributed")).inject(constantUsersPerSec(1) during(300) randomized).protocols(Base.httpConf))
```
pbougue, line 27: OK, I guess that you used this: https://gatling.io/docs/current/general/simulation_setup/#simulation-setup-throttling
But I'm pretty sure I miss something here: You're injecting 300(1 per second) users during 300s, at random moments, is that it?
Then is that random part repeatable and the same for both simulations?

pbougue (commented):
Youhou, one new language to navitia, and one from the JVM :yum:
Looks quite easy and neat... We may have to pay attention if the needs and uses grow around it, though.

gatling/simulations/navitia/distributed.scala
kinnou02: that's random without being random, it doesn't inject a user at the start of each second, so the charge isn't perfectly uniform.
It's not required for our use case but I like to add it when doing real load testing.

gatling/readme.md
kinnou02: oops, I updated the wrong file :(
it should be good now

nonifier:
Pure awesomeness : 

```java
setUp(
  scn.inject(
    nothingFor(4 seconds), // 1
    atOnceUsers(10), // 2
    rampUsers(10) over (5 seconds), // 3
    constantUsersPerSec(20) during (15 seconds), // 4
    constantUsersPerSec(20) during (15 seconds) randomized, // 5
    rampUsersPerSec(10) to 20 during (10 minutes), // 6
    rampUsersPerSec(10) to 20 during (10 minutes) randomized, // 7
    splitUsers(1000) into (rampUsers(10) over (10 seconds)) separatedBy (10 seconds), // 8
    splitUsers(1000) into (rampUsers(10) over (10 seconds)) separatedBy atOnceUsers(30), // 9
    heavisideUsers(1000) over (20 seconds) // 10
  ).protocols(httpConf)
)
```

nonifier (approved):

kinnou02:
skip ci

pbougue (approved):

mbounabi (approved):

xlqian (approved):

PR #2466, opened 2018-07-12
Ed: add rtree to speedup stop_point association to admins

kinnou02:
it's faster :)

less than 400ms to associate the stop_points from fr-idf

### DIFF
--- a/source/georef/georef.h
+++ b/source/georef/georef.h
@@ -300,9 +300,8 @@ struct GeoRef {
 
     /// Recherche d'une adresse avec un numéro en utilisant Autocomplete
     std::vector<nf::Autocomplete<nt::idx_t>::fl_quality> find_ways(const std::string & str, const int nbmax, const int search_type,std::function<bool(nt::idx_t)> keep_element, const std::set<std::string>& ghostwords) const;
-
-
     std::vector<Admin*> find_admins(const type::GeographicalCoord&) const;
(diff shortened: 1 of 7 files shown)


nonifier (approved):

source/georef/georef.h
```diff
@@ -302,7 +305,8 @@ struct GeoRef {
     std::vector<nf::Autocomplete<nt::idx_t>::fl_quality> find_ways(const std::string & str, const int nbmax, const int search_type,std::function<bool(nt::idx_t)> keep_element, const std::set<std::string>& ghostwords) const;
 
 
-    std::vector<Admin*> find_admins(const type::GeographicalCoord&) const;
+    std::vector<Admin*> find_admins(const type::GeographicalCoord&);
```
TeXitoi: I don't like such a method not const on GeoRef, that can break everything in the context of multithreading.

I'd prefer to not have this method, as it should only be used during binarisation.

kinnou02:
`find_admins` is back to `const`. I had to keep a `find_admins` without rtree as it's used by raptor here: https://github.com/CanalTP/navitia/blob/dev/source/routing/raptor_api.cpp#L1049-L1051
I'am pretty sure that we can drop compatibility with ntfs 0.1

TeXitoi (approved):

PR #2469, opened 2018-07-19
Jormungandr: make min/max_nb_journeys > 0

xlqian:
make min/max_nb_journeys > 0

### DIFF
--- a/documentation/slate/source/includes/apis.md
+++ b/documentation/slate/source/includes/apis.md
@@ -994,7 +994,7 @@ disrupted, it doesn't prevent the journey from beeing executed but modifies it. 
 | nop     | bike_speed           | float   | Biking speed for the fallback<br>Speed unit must be in meter/seconds | 4.1 m/s<br>(14.7 km/h)   |
 | nop     | bss_speed            | float   | Speed while using a bike from a bike sharing system for the fallback sections<br>Speed unit must be in meter/seconds | 4.1 m/s<br>(14.7 km/h)    |
 | nop     | car_speed            | float   | Driving speed for the fallback sections<br>Speed unit must be in meter/seconds         | 16.8 m/s<br>(60 km/h)   |
-| nop     | min_nb_journeys      | non-negtive int | Minimum number of different suggested journeys<br>More in multiple_journeys  |             |
+| nop     | min_nb_journeys      | non-negative int | Minimum number of different suggested journeys<br>More in multiple_journeys  |             |
 | nop     | max_nb_journeys      | positive int | Maximum number of different suggested journeys<br>More in multiple_journeys  |             |
 | nop     | count                | int     | Fixed number of different journeys<br>More in multiple_journeys  |             |
 | nop     | max_nb_tranfers      | int     | Maximum number of transfers in each journey  | 10          |
--- a/documentation/slate/source/includes/apis.md
+++ b/documentation/slate/source/includes/apis.md
@@ -994,8 +994,8 @@ disrupted, it doesn't prevent the journey from beeing executed but modifies it. 
 | nop     | bike_speed           | float   | Biking speed for the fallback<br>Speed unit must be in meter/seconds | 4.1 m/s<br>(14.7 km/h)   |
 | nop     | bss_speed            | float   | Speed while using a bike from a bike sharing system for the fallback sections<br>Speed unit must be in meter/seconds | 4.1 m/s<br>(14.7 km/h)    |
 | nop     | car_speed            | float   | Driving speed for the fallback sections<br>Speed unit must be in meter/seconds         | 16.8 m/s<br>(60 km/h)   |
-| nop     | min_nb_journeys      | int     | Minimum number of different suggested journeys<br>More in multiple_journeys  |             |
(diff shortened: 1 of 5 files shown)


source/jormungandr/jormungandr/interfaces/v1/Journeys.py
```diff
@@ -470,7 +470,7 @@ def __init__(self):
                                      "it'll override some specific parameters")
         parser_get.add_argument("min_nb_journeys", type=int,
                                 help='Minimum number of different suggested journeys')
-        parser_get.add_argument("max_nb_journeys", type=int,
+        parser_get.add_argument("max_nb_journeys", type=inputs.positive,
```
nonifier: would be cool to have the same `type=inputs.positive` for `min_nb_journeys`

source/jormungandr/jormungandr/scenarios/new_default.py
```diff
@@ -778,6 +778,9 @@ def __init__(self):
     def fill_journeys(self, request_type, api_request, instance):
         logger = logging.getLogger(__name__)
 
+        if api_request['max_nb_journeys'] is not None and api_request['max_nb_journeys'] <= 0:
```
nonifier, line 781: if `max_nb_journeys` if of type `inputs.positive`, you might not need to check if it's below zero init' ?.

source/jormungandr/tests/journey_common_tests.py
```diff
@@ -942,6 +942,19 @@ def test_section_fare_zone(self):
...
 
+    def test_max_nb_journeys_equals_0(self):
+        """
+        if the user ask for 0 journey, we return 0 journey
+        """
+        query = "journeys?from={from_sa}&to={to_sa}&datetime={datetime}&max_nb_journeys={max_nb_journeys}"\
+                .format(from_sa="stopA",
+                        to_sa="stopB",
+                        datetime="20120614T223000",
+                        max_nb_journeys=0)
```
nonifier: Out of curiosity, what happened if you do a `max_nb_journeys=-13` ? do you have any error or the param gets converted into a unsigned int ?

source/jormungandr/jormungandr/scenarios/new_default.py
pbougue, line 781: why would max be <= 0 at that point?

xlqian: it's a safe check, just in case someone overwrite it like this:
https://github.com/CanalTP/navitia/blob/dev/source/jormungandr/jormungandr/interfaces/v1/Journeys.py#L532

pbougue (commented):
We need to do something about "count" also, no?
https://github.com/CanalTP/navitia/blob/dev/source/jormungandr/jormungandr/interfaces/v1/Journeys.py#L532
I'd probably change `default_count_arg_type` to [1; whatever[ but I don't know if we should do this just for journeys (safe option) or for every request (logical option)...

source/jormungandr/tests/journey_common_tests.py
xlqian: there is no built-in unsinged int in python, you can make it happen with a third-party lib,

source/jormungandr/jormungandr/scenarios/new_default.py
xlqian: it's a safe check, just in case someone overwrite it like this:
https://github.com/CanalTP/navitia/blob/dev/source/jormungandr/jormungandr/interfaces/v1/Journeys.py#L532

xlqian:
@pbougue I'm aware of the behavior of count=0, i won't do it in this PR,

pbougue (commented):
As there is no consensus on `min > 0` I'll simply vote against enforcing it, as we might break some integration and it's blocking nothing functionally (as opposed to max).
Anyway, we should precise the `>0` of concerned params in doc.navitia.io.

documentation/slate/source/includes/apis.md
```diff
@@ -994,8 +994,8 @@ disrupted, it doesn't prevent the journey from beeing executed but modifies it.
 | nop     | bike_speed           | float   | Biking speed for the fallback<br>Speed unit must be in meter/seconds | 4.1 m/s<br>(14.7 km/h)   |
 | nop     | bss_speed            | float   | Speed while using a bike from a bike sharing system for the fallback sections<br>Speed unit must be in meter/seconds | 4.1 m/s<br>(14.7 km/h)    |
 | nop     | car_speed            | float   | Driving speed for the fallback sections<br>Speed unit must be in meter/seconds         | 16.8 m/s<br>(60 km/h)   |
-| nop     | min_nb_journeys      | int     | Minimum number of different suggested journeys<br>More in multiple_journeys  |             |
-| nop     | max_nb_journeys      | int     | Maximum number of different suggested journeys<br>More in multiple_journeys  |             |
+| nop     | min_nb_journeys      | non-negtive int | Minimum number of different suggested journeys<br>More in multiple_journeys  |             |
```
mbounabi: typo: neg**a**tive

nonifier (approved):

mbounabi (approved):

benoit-bst (approved):

TeXitoi (approved):

PR #2476, opened 2018-07-25
Kraken: Set valid jp and jpp - follow up

nonifier:
The main `set_valid_jp_and_jpp` was forgotten to be removed... My bad !


xlqian (approved):

benoit-bst (approved):

pbougue (commented):
Looks better :-)
Don't know why `rt_level` is needed in boucleRAPTOR (nice naming), but it feels like the cache is blocking some possible optims here also, as we consider and use rt_level and accessibilite_params (nice naming also) at each iteration.
poke @TeXitoi

PR #2480, opened 2018-07-27
Doc: add Eitri as an alternative to .nav generation

nonifier:


xlqian:
btw you can update the Readme of eitri...

xlqian:
add `#!/usr/bin/env python` at the very beginning of eitri.py then change the launch command in eitri Readme to ./eitri.py

nonifier:
done !

PR #2483, opened 2018-07-30
All: Miscellaneous fixes

pbougue:
Mostly warning removals, also late PR reviews follow-ups

A bit easier to read by commit.

TODO:

- [x] update utils commit when https://github.com/CanalTP/utils/pull/67 is merged
- [x] check if some other warnings may be treated as well


source/georef/georef.cpp
```diff
@@ -632,7 +632,7 @@ std::pair<GeoRef::ProjectionByMode, bool> GeoRef::project_stop_point(const type:
         nt::Mode_e::Car // CarNoPark -> Car
     }}};
 
-    for (auto const &mode_layer: mode_to_layer) {
+    for (auto const mode_layer: mode_to_layer) {
```
nonifier, line 635: did ya remove that on purpose ? did you had a warning on this ?

mbounabi:
Can be discussed before. Not even mentionned once in stand-up meetings!

source/georef/georef.cpp
pbougue: Yep, clang warns on extra refs and missing refs (it's the same warning : either all of them or none. I chose to keep all).
Here we always get a copy, so no point having a ref.

pbougue (commented):
Yep, we can discuss a bit.
Was mentionned on standup when I said I setted up my new laptop and I would probably open a PR on all those warnings from the new clang version I got.
I added follow-ups on PR I had no time to review but found some things that could be improved (but were not bugged). I can make separate PRs for those if you want.

source/routing/isochrone.cpp
```diff
@@ -73,7 +73,7 @@ type::GeographicalCoord project_in_direction(const type::GeographicalCoord& cent
     } else {
         double delta_lat;
         double delta_lon;
-        if (fmod(direction, 180) == 0) {
+        if (fabs(fmod(direction, 180)) < GeographicalCoord::coord_epsilon) {
```
nonifier, line 76: Why do you need `fabs` here ? I'm surprised a remainder of a division could negative :confused:

source/cmake_modules/CompilationFlags.cmake
```diff
@@ -24,7 +24,7 @@ endif(CMAKE_COMPILER_IS_GNUCXX)
 if("${CMAKE_CXX_COMPILER_ID}" MATCHES "Clang")
     # it could be great to remove these warnings, but there is so much :
     # -Wno-shorten-64-to-32 -Wno-sign-conversion -Wno-conversion
-    set(CMAKE_CLANG_WARN_FLAGS "-Weverything -Wno-c++98-compat -Wno-c++98-compat-pedantic -Wno-padded -Wno-global-constructors -Wno-exit-time-destructors -Wno-documentation -Wno-shadow -Wno-covered-switch-default -Wno-switch-enum -Wno-missing-noreturn -Wno-disabled-macro-expansion -Wno-shorten-64-to-32 -Wno-sign-conversion -Wno-conversion")
+    set(CMAKE_CLANG_WARN_FLAGS "-Weverything -Wno-c++98-compat -Wno-c++98-compat-pedantic -Wno-padded -Wno-global-constructors -Wno-exit-time-destructors -Wno-documentation -Wno-shadow -Wno-covered-switch-default -Wno-switch-enum -Wno-missing-noreturn -Wno-disabled-macro-expansion -Wno-shorten-64-to-32 -Wno-sign-conversion -Wno-conversion -Wno-double-promotion")
```
nonifier: As long as we don't support any 32b machine, I'm okay with this.

source/georef/georef.cpp
nonifier: I guess that's a general comment for all the references being removed in this PR

source/cmake_modules/CompilationFlags.cmake
pbougue: I can't support 32b platform :stuck_out_tongue_closed_eyes: (OK: sounds better in French)
the alternative being to promoted everything to double... :-/

source/georef/georef.cpp
pbougue: Yup!

source/routing/isochrone.cpp
pbougue: `fmod(-5, 2) = -1` (just tested it) ... But I must admit I didn't think much of it before your question, just added safety that could have been a bit too much :stuck_out_tongue:

source/cmake_modules/CompilationFlags.cmake
TeXitoi: Can't we just fix the warnings? We should never use float in navitia AFAIK.

TeXitoi (approved):
OK for me except the no-double-promotion.

source/cmake_modules/CompilationFlags.cmake
TeXitoi: We have double precision FPU since 386DX, so we are safe around float to double upscaling even on 32 bits architecture.

source/routing/isochrone.cpp
nonifier: With a negative dividend, that actually make sense. Never thought of it :tired_face:

source/cmake_modules/CompilationFlags.cmake
nonifier: I agree with @TeXitoi, I would rather fix the warning in the code.

source/georef/street_network.cpp
```diff
@@ -43,7 +43,7 @@ namespace navitia { namespace georef {
 const auto source_e = ProjectionData::Direction::Source;
 const auto target_e = ProjectionData::Direction::Target;
 
-navitia::time_duration PathFinder::crow_fly_duration(const double distance) const {
+navitia::time_duration PathFinder::crow_fly_duration(const float distance) const {
```
TeXitoi: This one is not needed, but we don't really care.

TeXitoi (approved):

nonifier (approved):

pbougue:
Jenkins, test this please

pbougue:
Jenkins, test this please

PR #2485, opened 2018-08-01
Eitri, Tyr & docker tests: Launch even if docker can't reach server

pbougue:
Allows to launch eitri or docker_test when not connected to internet.

This was hand-tested: launched `eitri` and `make docker_test` while offline.

TODO:

- [x] Check if we can even do smarter checks

### DIFF
--- a/source/eitri/docker_wrapper.py
+++ b/source/eitri/docker_wrapper.py
@@ -1,3 +1,4 @@
+# coding=utf-8
 # Copyright (c) 2001-2015, Canal TP and/or its affiliates. All rights reserved.
 #
 # This file is part of Navitia,
@@ -34,7 +35,7 @@ import psycopg2
 
 # postgres/postgis image
 POSTGIS_IMAGE = 'github.com/CanalTP/docker-postgis.git'
-POSTGIS_CONTAINER_NAME = 'postgis:2.1'
+POSTGIS_IMAGE_NAME = 'postgis:2.1'
 
 
 class DbParams(object):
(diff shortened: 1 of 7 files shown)


pbougue:
GitHub hooks are still messed-up, actual build : https://ci.navitia.io/job/navitia_pull_request_build/5182/consoleFull

xlqian (approved):

source/eitri/docker_wrapper.py
```diff
@@ -65,15 +65,25 @@ def old_school_cnx_string(self):
...
     def __init__(self):
         log = logging.getLogger(__name__)
         self.docker = docker.Client(base_url='unix://var/run/docker.sock')
 
-        log.info("building the temporary docker image")
-        for build_output in self.docker.build(POSTGIS_IMAGE, tag=POSTGIS_CONTAINER_NAME, rm=True):
-            log.debug(build_output)
+        log.info("Trying to build/update the docker image")
+        try:
+            for build_output in self.docker.build(POSTGIS_IMAGE, tag=POSTGIS_CONTAINER_NAME, rm=True):
```
nonifier: Do we have to build the image every time ? Would it not be better to pull the image instead ?

pbougue:
Please read by commit.
Changed docker lib from https://pypi.org/project/docker-py/ to https://pypi.org/project/docker/ (it's the evolution of the lib, only it's maintained).

Warning: those 2 libs are not compatible, so you will have to remove `docker-py` from your virtualenv (the simplest is just to dump the venv).
It should not affect anything/anybody else than developpers, as it is only used by eitri, docker_test and tyr tests (none are launched on other places than dev's laptops, so far)

nonifier (approved):

source/eitri/docker_wrapper.py
pbougue: Then we have to have a job building automatically the image, pushing it to a docker repo, and pull it in tests, if I'm not missing something.
It seems too big of a change to me.

pbougue:
Jenkins, test this please

PR #2486, opened 2018-08-02
Kraken: Teleportation crowfly = walking

mbounabi:
https://jira.kisio.org/browse/NAVITIAII-2482

### DIFF
--- a/source/type/pb_converter.cpp
+++ b/source/type/pb_converter.cpp
@@ -1898,22 +1898,6 @@ void PbCreator::fill_crowfly_section(const type::EntryPoint& origin, const type:
     fill(&origin, section->mutable_origin(), 2);
     fill(&destination, section->mutable_destination(), 2);
 
-    section->set_begin_date_time(navitia::to_posix_timestamp(origin_time));
-    section->set_duration(crow_fly_duration.total_seconds());
-    if (crow_fly_duration.total_seconds() > 0) {
-        section->set_length(origin.coordinates.distance_to(destination.coordinates));
-        auto* new_coord = section->add_shape();
-        new_coord->set_lon(origin.coordinates.lon());
-        new_coord->set_lat(origin.coordinates.lat());
-        new_coord = section->add_shape();
-        new_coord->set_lon(destination.coordinates.lon());
-        new_coord->set_lat(destination.coordinates.lat());
-    } else {
-        section->set_length(0);
-    }
-    section->set_end_date_time(navitia::to_posix_timestamp(origin_time + crow_fly_duration.to_posix()));
-    section->set_type(pbnavitia::SectionType::CROW_FLY);
-
     //we want to store the transportation mode used
     switch (mode) {
     case type::Mode_e::Walking:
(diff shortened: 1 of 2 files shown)


source/type/pb_converter.cpp
```diff
@@ -1932,6 +1916,26 @@ void PbCreator::fill_crowfly_section(const type::EntryPoint& origin, const type:
...
+        auto* new_coord = section->add_shape();
+        new_coord->set_lon(origin.coordinates.lon());
+        new_coord->set_lat(origin.coordinates.lat());
+        new_coord = section->add_shape();
+        new_coord->set_lon(destination.coordinates.lon());
+        new_coord->set_lat(destination.coordinates.lat());
+    } else {
+        section->set_length(0);
+        // For teleportation crow_fly (duration=0), the mode is always 'walking'
+        section->mutable_street_network()->set_mode(pbnavitia::Walking);
```
pbougue, line 1933: This will do the job.
Just an idea (maybe it can raise warnings in some others mind): We sometimes get 0 durations while still using a real fallback... Probably very rare and probably not a big problem to have walking in those case?
I don't know if we can check easily something else like the "is free" matrix (but then it should be tested at a higher level I guess).

source/type/tests/fill_pb_object_tests.cpp
```diff
@@ -273,3 +273,30 @@ BOOST_AUTO_TEST_CASE(pb_convertor_ptref) {
...
+    boost::gregorian::date d1(2014,06,14);
+    boost::posix_time::ptime t1(d1, boost::posix_time::seconds(10));
+    boost::posix_time::ptime t2(d1, boost::posix_time::hours(10));
+    boost::posix_time::time_period period(t1, t2);
+
+    auto * data_ptr = b.data.get();
+    navitia::PbCreator pb_creator(data_ptr, pt::not_a_date_time, period); 
+
+    pb_creator.fill_crowfly_section(EntryPoint(Type_e::StopPoint, "Totorigin"), 
+                                    EntryPoint(Type_e::StopPoint, "Tatastination"),
```
pbougue, line 294: :smile:

pbougue (commented):
OK for new_default scenario.

I wonder how Navitia behaves on distributed scenario (also when we don't use kraken beneath, like for geovelo).
We should probably handle that right away if it's easy, or add an distributed-isofunctionnality JIRA ticket?
Maybe we can mutualize code between scenario?
> JIRA for that : https://jira.kisio.org/browse/NAVP-941

To be QA-tested on sncf instances (we don't have street-network, always crowfly, and we shouldn't break behavior there)

xlqian (approved):

benoit-bst (approved):

nonifier (approved):

source/type/pb_converter.cpp
nonifier: I'll merge the PR as your point isn't blocking. But please, raise a PR if you think this would be usefull :)

PR #2487, opened 2018-08-02
jormungandr: logs on bss services in newrelic

kadhikari:
Ticket: https://jira.kisio.org/browse/NAVITIAII-2488
Note: Please see the comments in the ticket.

### DIFF
--- a/source/jormungandr/jormungandr/parking_space_availability/bss/atos.py
+++ b/source/jormungandr/jormungandr/parking_space_availability/bss/atos.py
@@ -88,7 +88,7 @@ class AtosProvider(CommonBssProvider):
         with self._get_client() as client:
             all_stands = client.service.getSummaryInformationTerminals(self.id_ao)
             return {stands.libelle: Stands(stands.nbPlacesDispo, stands.nbVelosDispo,
-                                           StandsStatus.open if stands.etatConnexion == 'CONNECTEE' else 'UNAVAILABLE')
+                                           StandsStatus.open if stands.etatConnexion == 'CONNECTEE' else StandsStatus.unavailable)
                     for stands in all_stands}
 
     @contextmanager
--- a/source/jormungandr/jormungandr/parking_space_availability/bss/atos.py
+++ b/source/jormungandr/jormungandr/parking_space_availability/bss/atos.py
@@ -79,8 +79,7 @@ class AtosProvider(CommonBssProvider):
 
                 return stands
         except:
-            msg = 'transport error during call to {} bss provider'.format(self.network)
-            logging.getLogger(__name__).exception(msg)
+            logging.getLogger(__name__).exception('transport error during call to %s bss provider', self.id_ao)
 
         return Stands(0, 0, StandsStatus.unavailable)
 
--- a/source/jormungandr/jormungandr/parking_space_availability/bss/atos.py
+++ b/source/jormungandr/jormungandr/parking_space_availability/bss/atos.py
@@ -34,15 +34,15 @@ import logging
 import pybreaker
 import zeep
 
-from jormungandr import cache, app, utils, new_relic
-from jormungandr.parking_space_availability import AbstractParkingPlacesProvider
+from jormungandr import cache, app
+from jormungandr.parking_space_availability.bss.common_bss_provider import CommonBssProvider, BssProxyError
 from jormungandr.parking_space_availability.bss.stands import Stands, StandsStatus
 from jormungandr.ptref import FeedPublisher
 
 DEFAULT_ATOS_FEED_PUBLISHER = None
 
 
-class AtosProvider(AbstractParkingPlacesProvider):
+class AtosProvider(CommonBssProvider):
 
     def __init__(self, id_ao, network, url, operators={'keolis'}, timeout=5,
                  feed_publisher=DEFAULT_ATOS_FEED_PUBLISHER, **kwargs):
@@ -63,13 +63,12 @@ class AtosProvider(AbstractParkingPlacesProvider):
         return properties.get('operator', '').lower() in self.operators and \
                properties.get('network', '').lower() == self.network
 
-    def get_informations(self, poi):
+    def _get_informations(self, poi):
         logging.debug('building stands')
         try:
             all_stands = self.breaker.call(self._get_all_stands)
             ref = poi.get('properties', {}).get('ref')
             if not ref:
-                utils.record_internal_failure('unavailable', 'bss', self.network)
                 return Stands(0, 0, StandsStatus.unavailable)
             stands = all_stands.get(ref.lstrip('0'))
             if stands:
@@ -78,15 +77,11 @@ class AtosProvider(AbstractParkingPlacesProvider):
                     stands.available_places = 0
                     stands.total_stands = 0
 
-                utils.record_call('ok', 'bss', self.network)
                 return stands
         except:
             msg = 'transport error during call to {} bss provider'.format(self.network)
             logging.getLogger(__name__).exception(msg)
-            #record in newrelic
-            utils.record_external_failure(msg, 'bss', self.network)
 
-        utils.record_internal_failure('unavailable', 'bss', self.network)
         return Stands(0, 0, StandsStatus.unavailable)
 
     @cache.memoize(app.config['CACHE_CONFIGURATION'].get('TIMEOUT_ATOS', 30))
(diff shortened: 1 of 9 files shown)


source/jormungandr/jormungandr/parking_space_availability/bss/jcdecaux.py
```diff
@@ -88,9 +94,11 @@ def get_informations(self, poi):
         data = self._call_webservice()
         if data and 'status' in data.get(ref, {}):
             if data[ref]['status'] == 'OPEN':
+                utils.record_call('ok', 'bss', self.network)
                 return Stands(data[ref].get('available_bike_stands', 0), data[ref].get('available_bikes', 0), 'OPEN')
             elif data[ref]['status'] == 'CLOSED':
                 return Stands(0, 0, 'CLOSED')
```
pbougue: missing one here, no?

source/jormungandr/jormungandr/parking_space_availability/bss/cykleo.py
```diff
@@ -145,18 +154,22 @@ def get_informations(self, poi):
...
             return Stands(0, 0, 'UNAVAILABLE')
 
         if obj_station.get('station', {}).get('status') == 'IN_SERVICE' and 'availableDockCount' in obj_station:
+            utils.record_call('ok', 'bss', self.network)
             return Stands(obj_station.get('availableDockCount', 0),
                           obj_station.get('availableClassicBikeCount', 0) +
                           obj_station.get('availableElectricBikeCount', 0),
                           'OPEN')
         elif obj_station.get('station', {}).get('status') in ('OUT_OF_SERVICE', 'DECOMMISSIONED'):
             return Stands(0, 0, 'CLOSED')
```
pbougue: missing one here?

pbougue (commented):
This will conflict with my PR changing the enum (one of the 2 PR will need a rebase).
Did not thoroughly read it, but it looks like we have no choice but record before each return (and before raising also) and some are missing (not sure I didn't miss any).
I think @kinnou02 should definitely review it for the format of records.

kinnou02:
We should abstract thing a little bit like it has been done in schedules so we do not have to handle newrelic in every connector: https://github.com/CanalTP/navitia/blob/dev/source/jormungandr/jormungandr/realtime_schedule/realtime_proxy.py#L122-L131

nonifier:
There is a lot of duplicated code around the logging, like :
```python
            msg = 'JCDecaux error'
            logging.getLogger(__name__).exception(msg)
            utils.record_external_failure(msg, 'bss', self.network)
```
Would be nice to wrap those repetitions in a function :)

source/jormungandr/jormungandr/parking_space_availability/bss/atos.py
```diff
@@ -87,7 +88,7 @@ def _get_all_stands(self):
         with self._get_client() as client:
             all_stands = client.service.getSummaryInformationTerminals(self.id_ao)
             return {stands.libelle: Stands(stands.nbPlacesDispo, stands.nbVelosDispo,
-                                           StandsStatus.open if stands.etatConnexion == 'CONNECTEE' else StandsStatus.unavailable)
+                                           StandsStatus.open if stands.etatConnexion == 'CONNECTEE' else 'UNAVAILABLE')
```
kinnou02: why?

source/jormungandr/jormungandr/parking_space_availability/bss/common_bss_provider.py
```diff
@@ -0,0 +1,59 @@
...
+        pass
+
+    def get_informations(self, poi):
+        try:
+            stand = self._get_informations(poi)
+            self.record_call('ok')
+            return stand
+
+        except BssProxyError as e:
+            self.record_call('failure', reason=str(e))
```
kinnou02, line 52: something is missing here, you need a `raise` or a `return`

source/jormungandr/jormungandr/parking_space_availability/bss/common_bss_provider.py
```diff
@@ -0,0 +1,59 @@
...
+from jormungandr import new_relic
+from jormungandr.parking_space_availability import AbstractParkingPlacesProvider
+from abc import abstractmethod
+
+
+class BssProxyError(RuntimeError):
+    pass
+
+
+class CommonBssProvider(AbstractParkingPlacesProvider):
```
kinnou02, line 39: why create a new class and not put it directly in `AbstractParkingPlacesProvider`

kadhikari: I did as for CommonCarParkProvider. I tried by moving the code from CommonCarParkProvider to AbstractParkingPlacesProvider but didn't work...

source/jormungandr/jormungandr/parking_space_availability/bss/common_bss_provider.py
kadhikari: done

source/jormungandr/jormungandr/parking_space_availability/bss/cykleo.py
```diff
@@ -75,14 +75,18 @@ def service_caller(self, method, url, headers, data=None, params=None):
             if not response or response.status_code != 200:
                 logging.getLogger(__name__).error('cykleo, Invalid response, status_code: {}'.format(
                     response.status_code))
+                raise BssProxyError('non 200 response')
                 return None
```
kinnou02: dead code to remove

source/jormungandr/jormungandr/parking_space_availability/bss/cykleo.py
```diff
@@ -75,14 +75,18 @@ def service_caller(self, method, url, headers, data=None, params=None):
...
         except pybreaker.CircuitBreakerError as e:
             logging.getLogger(__name__).error('cykleo service dead (error: {})'.format(e))
+            raise BssProxyError('circuit breaker open')
         except requests.Timeout as t:
             logging.getLogger(__name__).error('cykleo service timeout (error: {})'.format(t))
+            raise BssProxyError('timeout')
         except Exception as e:
             logging.getLogger(__name__).exception('cykleo error : {}'.format(str(e)))
+            raise BssProxyError(str(e))
         return None
```
kinnou02: dead code to remove

source/jormungandr/jormungandr/parking_space_availability/bss/jcdecaux.py
```diff
@@ -76,19 +76,23 @@ def _call_webservice(self):
...
+            raise BssProxyError('circuit breaker open')
         except requests.Timeout as t:
             logging.getLogger(__name__).error('JCDecaux service timeout (error: {})'.format(t))
-        except:
-            logging.getLogger(__name__).exception('JCDecaux error')
+            raise BssProxyError('timeout')
+        except Exception as e:
+            logging.getLogger(__name__).exception('JCDecaux error : {}'.format(str(e)))
+            raise BssProxyError(str(e))
         return None
```
kinnou02: dead code to remove

source/jormungandr/jormungandr/parking_space_availability/car/common_car_park_provider.py
```diff
@@ -84,21 +84,14 @@ def _call_webservice(self, request_url):
             self.record_call("OK")
             return data.json()
```
kinnou02: You should call `data.json()` before doing the `record_call` as this method can raise an exception: http://docs.python-requests.org/en/master/user/quickstart/#json-response-content

source/jormungandr/jormungandr/parking_space_availability/car/common_car_park_provider.py
```diff
@@ -80,25 +80,18 @@ def _call_webservice(self, request_url):
             else:
                 headers = None
             data = self.breaker.call(requests.get, url=request_url, headers=headers, timeout=self.timeout)
-            # record in newrelic
-            self.record_call("OK")
             return data.json()
+            self.record_call("OK")
```
mbounabi: This will never be called as it returns just before, right?

source/jormungandr/jormungandr/realtime_schedule/cleverage.py
```diff
@@ -29,7 +29,7 @@
 # https://groups.google.com/d/forum/navitia
 # www.navitia.io
 from __future__ import absolute_import, print_function, division
-from jormungandr.realtime_schedule.realtime_proxy import RealtimeProxy
+from jormungandr.realtime_schedule.realtime_proxy import RealtimeProxy, RealtimeProxyError
```
mbounabi, line 32: How was it working previously if it wasn't imported? Why is it necessary now?

kadhikari: Forgotten only in this file. All the oher files have this import : https://github.com/CanalTP/navitia/blob/b51ee3c1a5d166623f2a08f7e9243d2cb0cc9004/source/jormungandr/jormungandr/realtime_schedule/synthese.py#L34

source/jormungandr/jormungandr/parking_space_availability/bss/common_bss_provider.py
kinnou02: I think we said it should be a stands with Unavailable status?

kadhikari: It's my bad. Done

mbounabi (approved):

kinnou02 (approved):

PR #2492, opened 2018-08-08
Kraken: Fix adapted validity pattern with disruption

nonifier:
We were not using the right disruption freshness level when trying to match a disruption with a VJ.

### DIFF
--- a/source/kraken/tests/apply_disruption_test.cpp
+++ b/source/kraken/tests/apply_disruption_test.cpp
@@ -2312,6 +2312,20 @@ BOOST_AUTO_TEST_CASE(test_disruption_on_line_then_stop_point) {
                               .get_disruption(),
                               *b.data->pt_data, *b.data->meta);
 
+    navitia::apply_disruption(b.impact(nt::RTLevel::RealTime, "Y'a plus qu'a attendre ton train")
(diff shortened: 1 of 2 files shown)


source/kraken/tests/apply_disruption_test.cpp
```diff
@@ -2312,6 +2312,20 @@ BOOST_AUTO_TEST_CASE(test_disruption_on_line_then_stop_point) {
                               .get_disruption(),
                               *b.data->pt_data, *b.data->meta);
 
+    navitia::apply_disruption(b.impact(nt::RTLevel::RealTime, "Y'a plus qu'a attendre ton train")
```
xlqian: nothing keeps us from doing so, but in reality, it doesn't make sense when you apply `SIGNIFICANT_DELAYS` without new datetimes :)

xlqian: you may take a look in one of the previous tests how to handle SIGNIFICANT_DELAYS and mix those two realtime level...

nonifier: I've removed the SIGNIFICANT_DELAYS as they were not affecting the test.

TeXitoi (approved):

TeXitoi:
@xlqian can you merge if it's OK for you?

PR #2495, opened 2018-08-10
jormungandr: Add param tags[] to filter disruptions

kadhikari:
Ticket: https://jira.kisio.org/browse/NAVP-945
Blocked by https://jira.kisio.org/browse/NAVP-944

### DIFF
--- a/source/jormungandr/jormungandr/interfaces/v1/Disruptions.py
+++ b/source/jormungandr/jormungandr/interfaces/v1/Disruptions.py
@@ -96,6 +96,9 @@ class TrafficReport(ResourceUri):
                                 help="Distance range of the query. Used only if a coord is in the query")
         parser_get.add_argument("disable_geojson", type=BooleanType(), default=False,
                                 help="remove geojson from the response")
+        self.parsers["get"].add_argument("tags[]", type=six.text_type, action="append",
+                                         help="If filled, will restrained the search within "
+                                              "the given disruption tags")
         self.collection = 'traffic_reports'
         self.collections = traffic_reports
         self.get_decorators.insert(0, ManageError())
--- a/source/jormungandr/jormungandr/interfaces/v1/Disruptions.py
+++ b/source/jormungandr/jormungandr/interfaces/v1/Disruptions.py
@@ -96,9 +96,8 @@ class TrafficReport(ResourceUri):
                                 help="Distance range of the query. Used only if a coord is in the query")
         parser_get.add_argument("disable_geojson", type=BooleanType(), default=False,
                                 help="remove geojson from the response")
-        self.parsers["get"].add_argument("tags[]", type=six.text_type, action="append",
-                                         help="If filled, will restrained the search within "
-                                              "the given disruption tags")
+        parser_get.add_argument("tags[]", type=six.text_type, action="append",
+                                help="If filled, will restrained the search within the given disruption tags")
         self.collection = 'traffic_reports'
         self.collections = traffic_reports
         self.get_decorators.insert(0, ManageError())
@@ -119,10 +118,12 @@ class TrafficReport(ResourceUri):
         for forbid_id in args['__temporary_forbidden_id[]']:
             args['forbidden_uris[]'].append(forbid_id)
 
+        uris = []
         if uri:
             if uri[-1] == "/":
                 uri = uri[:-1]
             uris = uri.split("/")
+        if uris or args.get('tags[]', []):
             args["filter"] = self.get_filter(uris, args)
         else:
             args["filter"] = ""
--- a/source/jormungandr/jormungandr/interfaces/v1/Disruptions.py
+++ b/source/jormungandr/jormungandr/interfaces/v1/Disruptions.py
@@ -123,10 +123,7 @@ class TrafficReport(ResourceUri):
             if uri[-1] == "/":
                 uri = uri[:-1]
             uris = uri.split("/")
-        if uris or args.get('tags[]', []):
-            args["filter"] = self.get_filter(uris, args)
-        else:
-            args["filter"] = ""
+        args["filter"] = self.get_filter(uris, args)
 
         response = i_manager.dispatch(args, "traffic_reports", instance_name=self.region)
 
--- a/source/jormungandr/jormungandr/interfaces/v1/Disruptions.py
+++ b/source/jormungandr/jormungandr/interfaces/v1/Disruptions.py
@@ -97,7 +97,7 @@ class TrafficReport(ResourceUri):
         parser_get.add_argument("disable_geojson", type=BooleanType(), default=False,
                                 help="remove geojson from the response")
         parser_get.add_argument("tags[]", type=six.text_type, action="append",
-                                help="If filled, will restrained the search within the given disruption tags")
+                                help="If filled, will restrain the search within the given disruption tags")
         self.collection = 'traffic_reports'
         self.collections = traffic_reports
         self.get_decorators.insert(0, ManageError())
--- a/source/jormungandr/jormungandr/interfaces/v1/LineReports.py
+++ b/source/jormungandr/jormungandr/interfaces/v1/LineReports.py
@@ -92,9 +92,8 @@ class LineReports(ResourceUri, ResourceUtc):
                                 help="use disruptions valid after this date")
         parser_get.add_argument("until", type=DateTimeFormat(),
                                 help="use disruptions valid before this date")
-        self.parsers["get"].add_argument("tags[]", type=six.text_type, action="append",
-                                         help="If filled, will restrained the search within "
-                                              "the given disruption tags")
+        parser_get.add_argument("tags[]", type=six.text_type, action="append",
+                                help="If filled, will restrained the search within the given disruption tags")
 
         self.collection = 'line_reports'
         self.collections = line_reports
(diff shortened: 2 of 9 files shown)


source/jormungandr/jormungandr/interfaces/v1/LineReports.py
```diff
@@ -92,6 +92,8 @@ def __init__(self):
                                 help="use disruptions valid after this date")
         parser_get.add_argument("until", type=DateTimeFormat(),
                                 help="use disruptions valid before this date")
+        parser_get.add_argument("tags[]", type=six.text_type, action="append",
+                                help="If filled, will restrained the search within the given disruption tags")
```
xlqian: i'd better precise it's a "OR" condition

kadhikari: When we use a parameter with multiple values, it's obviously 'or' condition non?

nonifier (approved):

source/jormungandr/jormungandr/interfaces/v1/Uri.py
```diff
@@ -119,7 +121,7 @@ def get(self, region=None, lon=None, lat=None, uri=None, id=None):
         if args.get("headsign"):
             f = u"vehicle_journey.has_headsign({})".format(protect(args["headsign"]))
             if args.get("filter"):
-                args["filter"] += " and " + f
+                args["filter"] = '({})'.format(args["filter"]) + " and " + f
```
kinnou02: Only one format can be used here:
```python
'({}) and {}'.format(args["filter"]) + " and " + f)
```

source/jormungandr/jormungandr/interfaces/v1/ResourceUri.py
```diff
@@ -96,6 +96,14 @@ def get_filter(self, items, args):
...
                     filter_list.append(type_ + ".uri=" + protect(item))
                 type_ = None
+        # handle tags
+        tags = args.get("tags[]", [])
+        if tags:
+            tag_list = []
+            for tag in tags:
+                tag_list.append("disruption.tag({})".format(protect(tag)))
+            tag_filter = '({})'.format(" or ".join(tag_list))
+            filter_list.append(tag_filter)
```
kinnou02: since #2498 include a `tags()` synthax it's possible to make it more readable:
```python
tags = args.get("tags[]", [])
￼if tags:
    filter_list.append('disruptions.tags({})'.format(' ,'.join([protect(t) for t in tags])))
```

source/jormungandr/jormungandr/interfaces/v1/Uri.py
```diff
@@ -163,17 +165,19 @@ def get(self, region=None, lon=None, lat=None, uri=None, id=None):
...
         if not self.region:
             return {"error": "No region"}, 404
+        uris = []
         if uri:
             if uri[-1] == "/":
                 uri = uri[:-1]
             uris = uri.split("/")
             if collection is None:
                 collection = uris[-1] if len(uris) % 2 != 0 else uris[-2]
+        if uris or args.get('tags[]', []):
```
kinnou02: why?

source/jormungandr/jormungandr/interfaces/v1/ResourceUri.py
```diff
@@ -96,6 +96,14 @@ def get_filter(self, items, args):
...
         return " and ".join(filter_list)
```
pbougue: parenthesis everywhere would be safer in my opinion:
```python
return '({})'.format(') and ('.join(filter_list))
```

This way you protect anything and don't need to change l.69

source/jormungandr/jormungandr/interfaces/v1/Uri.py
pbougue: I vote for parenthesis around `(f)`

source/jormungandr/jormungandr/interfaces/v1/Uri.py
pbougue: (Comment to be managed on all occurences in this PR if needed)
Shouldn't we always do it, no condition and manage args/uri it in `get_filter` directly?
Would avoid simple/stupid bugs, I guess.

source/jormungandr/jormungandr/interfaces/v1/Uri.py
```diff
@@ -163,17 +165,19 @@ def get(self, region=None, lon=None, lat=None, uri=None, id=None):
...
             args["filter"] = self.get_filter(uris, args)
         if collection and id:
             f = u'{o}.uri={v}'.format(o=collections_to_resource_type[collection], v=protect(id))
             if args.get("filter"):
-                args["filter"] += " and " + f
+                args["filter"] = '({})'.format(args["filter"]) + " and " + f
```
pbougue: same: I vote for parenthesis around `(f)`
(same above)

source/jormungandr/tests/chaos_disruptions_tests.py
```diff
@@ -89,6 +90,9 @@ def test_disruption_on_stop_area_b(self):
         assert any(d['disruption_id'] == 'bob_the_disruption' for d in disruptions)
         assert disruptions[0]['cause'] == 'CauseTest'
         assert 'category' not in disruptions[0]
+        assert len(disruptions[0]['tags']) == 2
+        assert 'metro' in disruptions[0]['tags']
```
pbougue, line 94: You only test that tags are outputed, not the filtering if I understand correctly.
Can you add a test on filtering as it's the main feature of this PR?

source/jormungandr/jormungandr/interfaces/v1/Uri.py
kadhikari: Filter statement contains condition for each uri as well as tags used in the query.

pbougue (approved):
Fine for me, I let you decide for the 2 suggestions left in comment (and discussed on slack)

xlqian (approved):

source/jormungandr/jormungandr/interfaces/v1/Uri.py
```diff
@@ -163,17 +165,21 @@ def get(self, region=None, lon=None, lat=None, uri=None, id=None):
...
         if uri:
             if uri[-1] == "/":
                 uri = uri[:-1]
             uris = uri.split("/")
             if collection is None:
                 collection = uris[-1] if len(uris) % 2 != 0 else uris[-2]
-            args["filter"] = self.get_filter(uris, args)
+        param_filter = self.get_filter(uris, args)
+        if param_filter:
+            args["filter"] = param_filter
```
pbougue: Minor: you can directly do that and remove the 2 lines above (check below shouldn't be affected)

source/jormungandr/jormungandr/interfaces/v1/Disruptions.py
```diff
@@ -96,6 +96,8 @@ def __init__(self):
                                 help="Distance range of the query. Used only if a coord is in the query")
         parser_get.add_argument("disable_geojson", type=BooleanType(), default=False,
                                 help="remove geojson from the response")
+        parser_get.add_argument("tags[]", type=six.text_type, action="append",
+                                help="If filled, will restrained the search within the given disruption tags")
```
pbougue: Minor: "will restrain(no `ed`)"
Same on other occurrences

source/jormungandr/tests/chaos_disruptions_tests.py
pbougue: :heavy_check_mark:  Done

pbougue (dismissed):

source/jormungandr/jormungandr/interfaces/v1/Uri.py
kadhikari: No, it wont work. If we have a parameter filter (stop_area.id = 'toto' or stop_area.id = 'titi'), the code as you suggested will delete this filter. I had a test failed.

pbougue: You're right, sorry!
Nullifying filters is bad :-/

source/jormungandr/jormungandr/interfaces/v1/Uri.py
```diff
@@ -119,7 +121,7 @@ def get(self, region=None, lon=None, lat=None, uri=None, id=None):
         if args.get("headsign"):
             f = u"vehicle_journey.has_headsign({})".format(protect(args["headsign"]))
             if args.get("filter"):
-                args["filter"] += " and " + f
+                args["filter"] = '({}) and {}'.format(args["filter"], f)
```
pbougue: Minor: This "headsign" param should probably be grouped and handled in get_filters.

source/jormungandr/jormungandr/interfaces/v1/ResourceUri.py
```diff
@@ -65,8 +65,10 @@ def __init__(self, authentication=True, links=True, *args, **kwargs):
             self.get_decorators.append(authentication_required)
 
     def get_filter(self, items, args):
+        if not (items or args.get('tags[]', [])):
+            return ""
```
pbougue: In that case we should return `args["filter"]` unchanged (and it's what would be done without this early-return).
So I think early-return here is not a good idea, as we never know what params will be managed and added below.

That way you should be able to use it safely (and removing those 2 lines from Uri.py shouldn't break tests).

pbougue (commented):
"get_filter" changes his behaviour with the early-return and even if you managed it externally (so it doesn't break anything), I think the behaviour should stay the same.

pbougue (approved):
This looks nice :)
Should even solve https://jira.kisio.org/browse/NAVP-1002 :tada:

PR #2496, opened 2018-08-12
A few optimizations in jormun's hot path

kinnou02:
string formating with "%" is faster than format
`is_walk_after_parking` can only be call on street_network section

On a journeys with 217 journeys taking arround 1500ms these change
improve response time by 50/100ms. So not great, but better than nothing.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,7 +11,7 @@ documentation/build
 *.swp
 source/CMakeLists.txt.user
 tags
-*.pyc
+*.py[co]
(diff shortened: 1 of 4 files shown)


xlqian (approved):

xlqian:
interesting...

kadhikari (approved):

nonifier (approved):

benoit-bst (approved):

PR #2498, opened 2018-08-13
Ptref disruption by tags

nonifier:
A pt_ref request can now contain : `disruption.tag=my_tag` that returns a list of indexes referring to the disruptions with the specific tag.

I'm suggesting to add a syntax for querying multiple tags at once with `disruption.tags(my_tag1, my_tag2, <...>)`. With a similar behaviour to `disruption.tag=my_tag1 OR disruption.tag=my_tag2 OR <...>)` but with the benefits of parsing the disruptions only once with no AST parsing overhead.

## Limitation
There is a limitation though. We can only retrieve disruptions through the impacts in the model with: `make_query_ng(Type_e::Impact, "disruption.tag(TAG)")`. Links are missing in the data graph to resolve pt_objects from an impact. Thus, retrieving the lines affected by a disruption like `make_query_ng(Type_e::Line, "disruption.tag(TAG)")` will not work :cry:

### DIFF
--- a/source/ptreferential/tests/ptref_ng_test.cpp
+++ b/source/ptreferential/tests/ptref_ng_test.cpp
@@ -1,28 +1,28 @@
 /* Copyright © 2001-2014, Canal TP and/or its affiliates. All rights reserved.
-  
+
 This file is part of Navitia,
     the software to build cool stuff with public transport.
- 
+
 Hope you'll enjoy and contribute to this project,
     powered by Canal TP (www.canaltp.fr).
 Help us simplify mobility and open public transport:
     a non ending quest to the responsive locomotion way of traveling!
-  
+
 LICENCE: This program is free software; you can redistribute it and/or modify
 it under the terms of the GNU Affero General Public License as published by
 the Free Software Foundation, either version 3 of the License, or
 (at your option) any later version.
-   
+
 This program is distributed in the hope that it will be useful,
 but WITHOUT ANY WARRANTY; without even the implied warranty of
 MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
 GNU Affero General Public License for more details.
-   
+
 You should have received a copy of the GNU Affero General Public License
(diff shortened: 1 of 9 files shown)


source/ptreferential/tests/ptref_ng_test.cpp
```diff
@@ -300,3 +314,71 @@ BOOST_AUTO_TEST_CASE(get_connection) {
...
+
+    std::vector<std::string> disruption_names;
+    boost::range::transform(indexes, std::back_inserter(disruption_names),
+                            [&b](const navitia::idx_t id) {
+        return get_disruption_uri(id, *b.data);
+    });
+
+    auto expected_disruption_names = {"disrupt_0", "disrupt_2"};
+    BOOST_CHECK_EQUAL_RANGE(disruption_names, expected_disruption_names);
+
```
kadhikari, line 383: Is it possible to add a query to filter with tag_0 and tag_s another way (disruption.tag=tag_0 or disruption.tag=tag_2) ?

nonifier: Yes, it's possible indeed. This case is already covered at the grammar level (see : https://github.com/CanalTP/navitia/blob/17ae23c10cde3da41493f940efc4bb205d2d08c9/source/ptreferential/ptreferential_ng.cpp#L89). When an `or` is encountered in the request, the 2 sides of the expression are evaluated with the vistor and the result concatenated (cf. https://github.com/CanalTP/navitia/blob/17ae23c10cde3da41493f940efc4bb205d2d08c9/source/ptreferential/ptreferential_ng.cpp#L260).

In my example `disruption.tags(T1, T2, ... )` is just a quicker way to get to the same result :)

kadhikari: :+1:

xlqian (approved):

kinnou02 (approved):

benoit-bst (approved):

PR #2500, opened 2018-08-14
Jormun: divia connector P+R

xlqian:
Divia's api returns jsons with different attribute mappings depending on the requested dataset.
 
I made a class factory to manage those cases...

in jormungandr's settings.py, the settings will be something like:

```
        'class': 'jormungandr.parking_space_availability.car.divia.DiviaPRParkProvider',
        'args':
            {
                'url': 'https://data.explore.divia.fr/api/records/1.0/search/',
                'operators': {'DIVIA'},
                'api_key': 'XXXX',
                'dataset': 'xxxxx',
                'fail_max': 5,
                'timeout': 1
            }
```


source/jormungandr/jormungandr/parking_space_availability/car/divia.py
```diff
@@ -40,18 +40,26 @@
...
+                 attr_mapping=None, **kwargs):
 
         self._feed_publisher = FeedPublisher(**feed_publisher) if feed_publisher else None
         self.provider_name = 'DIVIA'
+        if attr_mapping is None:
+            attr_mapping = ['numero_parking', 'nombre_places_libres', 'nombre_places']
+
+        self.park = attr_mapping[0]
+        self.available = attr_mapping[1]
+        self.nb_places = attr_mapping[2]
```
kinnou02: I don't really like the idea of mapping as configuration
I would prefer that "kind of config" to be done at "build time", something like a factory of class: 
```
DiviaPRParking = ODSParking("numero_parking", "nombre_places_libres", "nombre_places")
DiviaParkingOther = ODSParking("nb_parking", "nb_places_libres", "nb_places")
```
And use this generated class in the configuration

xlqian: @kinnou02 I did think about your solution and I hesitated a lot...  my argument to do that in conf would be that if Divia provides a new dataset with a new format(hope they won't), we don't need to do dev but only updated the deployment conf...

kinnou02 (approved):

mbounabi (approved):

PR #2504, opened 2018-08-21
Jormun & Tyr: add max_nb_crowfly_by_mode

xlqian:
https://jira.kisio.org/browse/NAVP-951


source/jormungandr/jormungandr/interfaces/v1/Journeys.py
```diff
@@ -508,6 +508,18 @@ def __init__(self):
                                      "actually a minimum.\n"
                                      "Nota 2: 'max_nb_journeys' parameter has priority over "
                                      "'timeframe_duration' parameter.")
+        parser_get.add_argument("max_nb_crowfly_by_walking", type=int,
```
kinnou02: I think it should be a private parameter

source/jormungandr/jormungandr/interfaces/v1/Journeys.py
```diff
@@ -574,6 +586,13 @@ def _set_specific_params(mod):
                 args['_final_line_filter'] = mod.final_line_filter
             if args.get('_max_extra_second_pass') is None:
                 args['_max_extra_second_pass'] = mod.max_extra_second_pass
+
+            args['max_nb_crowfly_by_mode'] = mod.max_nb_crowfly_by_mode # it's a dict of str vs int
+            for mode in args['max_nb_crowfly_by_mode']:
```
pbougue: for mode`, _` in
would be clearer to me

source/jormungandr/jormungandr/interfaces/v1/Journeys.py
```diff
@@ -508,6 +508,18 @@ def __init__(self):
                                      "actually a minimum.\n"
                                      "Nota 2: 'max_nb_journeys' parameter has priority over "
                                      "'timeframe_duration' parameter.")
+        parser_get.add_argument("max_nb_crowfly_by_walking", type=int,
+                                help="limit nb of stop points accesible by crowfly, "
+                                     "used especially in distributed scenario")
+        parser_get.add_argument("max_nb_crowfly_by_car", type=int,
+                                help="limit nb of stop points accesible by crowfly, "
+                                     "used especially in distributed scenario")
+        parser_get.add_argument("max_nb_crowfly_by_bike", type=int,
```
pbougue: I did not think it is usefull to add it to /journeys (the JIRA is not clear on that, and we each thought of different things during refinement it seems :-/ ).
What you did could be easier to check impacts and negociate with partners, but it would not be permanent code (only debug/test) and I think it would be better to have a param only in tyr so noone can reactivate bike calls if we "deactivate" it in tyr.

Same as @kinnou02 at least, these should not be public params (`_` and `hidden=True`)

source/tyr/tyr/resources.py
```diff
@@ -315,6 +315,10 @@ def put(self, id=None, name=None):
                             help='maximum number of second pass to get more itineraries',
                             location=('json', 'values'), default=instance.max_extra_second_pass)
 
+        parser.add_argument('max_nb_crowfly_by_mode', type=dict,
+                            help='maximum number of second pass to get more itineraries',
```
pbougue: copy/paste leftover :wink:

source/jormungandr/jormungandr/interfaces/v1/Journeys.py
pbougue: So if the config is broken, we are never gonna be able to change a mode that's not in config, no ? It's less the case with what you did now in tyr config, I guess ?

pbougue (commented):
I let you guys decide if you keep the /journeys params, I wrote my pros/cons but no strong opposition from me.

source/jormungandr/jormungandr/interfaces/v1/Journeys.py
xlqian: @pbougue i added the params only in tyr at first place, and then i realized that it may be useful to expose the params in /journeys for debug purposes(for finding the appropriate limit nb which allows to find a better solution for example, it'll be annoying if we want to do that without a journey param..).....

source/jormungandr/jormungandr/interfaces/v1/Journeys.py
xlqian: @pbougue i'll do it in another way

xlqian: @pbougue you see the difference between:
`for mode, _ in args['max_nb_crowfly_by_mode'].items()`
and
`for mode in args['max_nb_crowfly_by_mode']`  
:wink:

source/tyr/migrations/versions/483639f1f00_add_max_nb_crowfly_by_mode.py
```diff
@@ -0,0 +1,23 @@
...
+from alembic import op
+import sqlalchemy as sa
+
+
+def upgrade():
+    import json
+    op.add_column('instance', sa.Column('max_nb_crowfly_by_mode', sa.PickleType(pickler=json), nullable=True))
+
+
+def downgrade():
```
nonifier, line 21: out of curiosity, I'm not quite familiar with this... Is this generated ? if so, how does this work ?

nonifier (approved):

source/tyr/migrations/versions/483639f1f00_add_max_nb_crowfly_by_mode.py
xlqian: @nonifier yep, it's generated by [alembic](http://alembic.zzzcomputing.com/en/latest/index.html), you've got a command(using alembic) in tyr which allows you to migrate/update your db with the given [models](https://github.com/CanalTP/navitia/blob/dev/source/navitiacommon/navitiacommon/models.py)/[schemas](https://github.com/CanalTP/navitia/tree/dev/source/tyr/migrations/versions), quite auto-magic though

kadhikari (approved):

source/tyr/migrations/versions/483639f1f00_add_max_nb_crowfly_by_mode.py
```diff
@@ -0,0 +1,23 @@
+"""empty message
```
kinnou02: It shouldn't be empty

kinnou02 (dismissed):

source/jormungandr/tests/routing_tests_experimental.py
```diff
@@ -64,6 +64,26 @@ def test_journey_with_different_fallback_modes(self):
         assert 'car_direct_path' in context
         assert 'co2_emission' in context['car_direct_path']
 
+    def test_journey_with_limited_nb_crowfly(self):
+        """
+        Test when max_nb_crowfly_by_car=0, we cannot fallback with car..
```
TeXitoi, line 69: Please also check that it's working with 1

TeXitoi (changes requested):

TeXitoi (approved):

PR #2508, opened 2018-08-26
Python code formatting with black

kinnou02:
This is a proposition to format python code using [black](https://black.readthedocs.io/en/stable/)
Obviously this change a lot lot of files, and will conflict with any PR, as such if this is approved and merged, it should be done when there is no PR open on python code.

Some of the formatting isn't really to my liking, but at least won't have to think about it anymore.
I have set the line length to 113, that's the size of the github window for diff.
I also disabled the string normalization as we use mostly simple quotes while black want to use double quotes, this will change every strings, but this is open to discussion.

Every tests (including tyr) still run :)

If this is approved I will look to do the same thing on c++ codes with clang-format, hopefully with less changes


pbougue (commented):
Looks fine to me.
If we go that way, we should probably add a CONTRIBUTING.md file, that would regroup what's in 'How to contribute', 'Want to dev in Navitia?' and we need to have a word on the commands that need to be run (and probably add something to jenkins checking that formating is OK)

TeXitoi (approved):

nonifier:
I'm vastly in favour of having such tool !

I'm not gonna to fussy about the coding style, especially as long people with better Python experience are fine with it and if we don't spend ages tuning it.

The questions I'm would like to address are:
- Line size: 113 seems somewhat large to me. The code isn't only displayed on Github's diff, and we should also consider other mediums like: `grep`ing the code in your console, compiler outputs in CI, IDE, diffing in other tools etc... In term of UX, it's also proven that long lines are hard and painful to read. And finally, longer lines are more likely to conflict. I would rather be in favour line size around ~80-90. 
- Hook: I don't know when `black` is being triggered, but a pre-commit integration seems good to avoid meaningless commits.


+1 for having this in C++ as well :)

kinnou02:
I already have a setup for a pre-commit hook (using https://pre-commit.com/) 
We choose 113 a long time ago, we can change it, if there some tools doing the refacto in don't really care

kinnou02:
I have added a `CONTRIBUTING.md` with instruction to run Black and setup the pre-commit hooks

benoit-bst (approved):

nonifier (approved):

xlqian (approved):

TeXitoi:
There are conflicts...

pbougue:
I only add a do_not_merge on it because I have some modifications to be applied before re-apply then merge this PR.
And we have to be synced to merge it, otherwise someone with an ongoing python work could really sweat...

source/tyr/tyr/binarisation.py
```diff
@@ -81,7 +81,7 @@ def move_to_backupdirectory(filename, working_directory):
     now = datetime.datetime.now()
     working_directory += "/" + now.strftime("%Y%m%d-%H%M%S%f")
     # this works even if the intermediate 'backup' dir does not exists
-    os.makedirs(working_directory, 0755)
+    os.makedirs(working_directory, 755)
```
TeXitoi: Nop, need `0o755`

TeXitoi (changes requested):

TeXitoi (approved):

PR #2509, opened 2018-08-27
Jormun & Kraken: first/last datetime

xlqian:
https://jira.kisio.org/browse/NAVP-934

### DIFF
--- a/source/cmake_modules/CompilationFlags.cmake
+++ b/source/cmake_modules/CompilationFlags.cmake
@@ -14,7 +14,7 @@ ENDIF(NOT CMAKE_BUILD_TYPE)
 # Gcc Release flags
 if(CMAKE_COMPILER_IS_GNUCXX)
     set(CMAKE_GNUCXX_WARN_FLAGS "-Wall -Wextra -Woverloaded-virtual -Wundef -pedantic")
-    set(CMAKE_GNUCXX_COMMON_FLAGS "-std=c++0x -rdynamic -g")
+    set(CMAKE_GNUCXX_COMMON_FLAGS "-std=c++0x -rdynamic -g  -fno-omit-frame-pointer")
     set(CMAKE_CXX_FLAGS "${CMAKE_GNUCXX_COMMON_FLAGS} ${CMAKE_GNUCXX_WARN_FLAGS}")
     set(CMAKE_CXX_FLAGS_PROFILE "${CMAKE_CXX_FLAG} --coverage -g")
     set(CMAKE_EXE_LINKER_FLAGS_PROFILE "${CMAKE_EXE_LINKER_FLAGS} --coverage")
--- a/source/cmake_modules/CompilationFlags.cmake
+++ b/source/cmake_modules/CompilationFlags.cmake
@@ -14,7 +14,7 @@ ENDIF(NOT CMAKE_BUILD_TYPE)
 # Gcc Release flags
 if(CMAKE_COMPILER_IS_GNUCXX)
     set(CMAKE_GNUCXX_WARN_FLAGS "-Wall -Wextra -Woverloaded-virtual -Wundef -pedantic")
-    set(CMAKE_GNUCXX_COMMON_FLAGS "-std=c++0x -rdynamic -g  -fno-omit-frame-pointer")
+    set(CMAKE_GNUCXX_COMMON_FLAGS "-std=c++0x -rdynamic -g")
     set(CMAKE_CXX_FLAGS "${CMAKE_GNUCXX_COMMON_FLAGS} ${CMAKE_GNUCXX_WARN_FLAGS}")
     set(CMAKE_CXX_FLAGS_PROFILE "${CMAKE_CXX_FLAG} --coverage -g")
     set(CMAKE_EXE_LINKER_FLAGS_PROFILE "${CMAKE_EXE_LINKER_FLAGS} --coverage")
--- a/source/jormungandr/jormungandr/interfaces/v1/serializer/schedule.py
+++ b/source/jormungandr/jormungandr/interfaces/v1/serializer/schedule.py
@@ -89,6 +89,8 @@ class StopScheduleSerializer(PbNestedSerializer):
     display_informations = pt.RouteDisplayInformationSerializer(attr='pt_display_informations')
     date_times = DateTimeTypeSerializer(many=True, display_none=True)
     links = jsonschema.MethodField(schema_type=LinkSchema(many=True))
+    first_datetime = DateTimeTypeSerializer(display_none=True)
+    last_datetime = DateTimeTypeSerializer(display_none=True)
 
     def get_links(self, obj):
         return _get_links(obj)
--- a/source/jormungandr/jormungandr/interfaces/v1/serializer/schedule.py
+++ b/source/jormungandr/jormungandr/interfaces/v1/serializer/schedule.py
@@ -89,8 +89,8 @@ class StopScheduleSerializer(PbNestedSerializer):
     display_informations = pt.RouteDisplayInformationSerializer(attr='pt_display_informations')
     date_times = DateTimeTypeSerializer(many=True, display_none=True)
     links = jsonschema.MethodField(schema_type=LinkSchema(many=True))
-    first_datetime = DateTimeTypeSerializer(display_none=True)
-    last_datetime = DateTimeTypeSerializer(display_none=True)
+    first_datetime = DateTimeTypeSerializer(display_none=False)
+    last_datetime = DateTimeTypeSerializer(display_none=False)
 
     def get_links(self, obj):
         return _get_links(obj)
--- a/source/jormungandr/tests/departure_board_tests.py
+++ b/source/jormungandr/tests/departure_board_tests.py
@@ -1107,31 +1107,39 @@ class TestSchedules(AbstractTestFixture):
 
 @dataset({"timezone_cape_verde_test": {}})
 class TestFirstLastDatetimeWithNegativeTimezone(AbstractTestFixture):
-    def test_first_last_datetime(self):
-        """
-        test first/last datetime with a timezone whose UTC offset is **negative**
 
-        * The opening time should be the earliest departure of all vjs of the line
-        * The closing time should be the latest arrival of all vjs of the line
+    """
+    test first/last datetime with a timezone whose UTC offset is **negative**
 
-        The closing time of line X is 02:45(local time) and the opening time is 8:00(local time)
+    * The opening time should be the earliest departure of all vjs of the line
+    * The closing time should be the latest arrival of all vjs of the line
 
-        Datetimes shown in the figure are all **LOCAL** datetimes
+    The closing time of line X is 02:45(local time) and the opening time is 8:00(local time)
 
-        [2017T0101, 2017T0103] and [2017T0106, 2017T0108]
-        X_s1 ------------ X_S2 ------------ X_S3
-        07:00             15:00             23:30    X:vj1
-        08:00             16:00             24:30    X:vj2
-        09:00             17:00             25:30    X:vj3
+    Datetimes shown in the figure are all **LOCAL** datetimes
 
-        [2017T0104, 2017T0105]
-        X_s1 ------------ X_S2 ------------ X_S3
-        07:05             15:05             23:45    X:vj4
-        08:05             16:05             24:45    X:vj5
-        09:05             17:05             25:45    X:vj6
+    [2017T0101, 2017T0103] and [2017T0106, 2017T0108]
+    X_s1 ------------ X_S2 ------------ X_S3
+    07:00             15:00             23:30    X:vj1    route_1
+    08:00             16:00             24:30    X:vj2    route_1
+    09:00             17:00             25:30    X:vj3    route_1
 
-        """
+    [2017T0104, 2017T0105]
+    X_s1 ------------ X_S2 ------------ X_S3
+    07:05             15:05             23:45    X:vj4    route_1
+    08:05             16:05             24:45    X:vj5    route_1
+    09:05             17:05             25:45    X:vj6    route_1
 
+
+
+    [2017T0101, 2017T0108]
+    X_s3 ------------ X_S2 ------------ X_S1
+    08:05             16:05             24:45    X:vj7    route_2
+    09:05             17:05             25:45    X:vj8    route_2
+
+    """
+
+    def test_first_last_datetime_same_day(self):
         if not app.config['USE_SERPY']:
             return
 
@@ -1144,13 +1152,28 @@ class TestFirstLastDatetimeWithNegativeTimezone(AbstractTestFixture):
          X_s1 ------------ X_S2 ------------ X_S3
          
         """
-        response = self.query_region("stop_points/X_S2/stop_schedules?from_datetime=20170102T1630"
-                                     "&data_freshness=base_schedule")
+        from_datetime = "20170102T1630"
+        response = self.query_region("stop_points/X_S2/stop_schedules?from_datetime={}"
+                                     "&data_freshness=base_schedule".format(from_datetime))
 
         stop_schedules = response['stop_schedules']
-        assert stop_schedules
+        assert len(stop_schedules) == 2
+
         assert stop_schedules[0]['first_datetime']['date_time'] == "20170102T150000"
         assert stop_schedules[0]['last_datetime']['date_time'] == "20170102T170000"
+        assert stop_schedules[1]['first_datetime']['date_time'] == "20170102T160500"
+        assert stop_schedules[1]['last_datetime']['date_time'] == "20170102T170500"
+
+        response = self.query_region("stop_points/X_S1/stop_schedules?from_datetime={}"
+                                     "&data_freshness=base_schedule".format(from_datetime))
+
+        stop_schedules = response['stop_schedules']
+        assert len(stop_schedules) == 2
+        assert stop_schedules[0]['first_datetime']['date_time'] == "20170103T070000"
+        assert stop_schedules[0]['last_datetime']['date_time'] == "20170102T090000"
+        # since it's the last stop_point, we cannot get on the bus....
+        assert 'first_datetime' not in stop_schedules[1]
+        assert 'last_datetime' not in stop_schedules[1]
 
         """              20170102
                                from_datetime 20170103T0001
@@ -1161,14 +1184,31 @@ class TestFirstLastDatetimeWithNegativeTimezone(AbstractTestFixture):
          X_s1 ------------ X_S2 ------------ X_S3
 
         """
-        response = self.query_region("stop_points/X_S2/stop_schedules?from_datetime=20170103T0001"
-                                     "&data_freshness=base_schedule")
+        from_datetime = "20170103T0001"
+        response = self.query_region("stop_points/X_S2/stop_schedules?from_datetime={}"
+                                     "&data_freshness=base_schedule".format(from_datetime))
 
         stop_schedules = response['stop_schedules']
-        assert stop_schedules
+        assert len(stop_schedules) == 2
         assert stop_schedules[0]['first_datetime']['date_time'] == "20170103T150000"
         assert stop_schedules[0]['last_datetime']['date_time'] == "20170103T170000"
+        assert stop_schedules[1]['first_datetime']['date_time'] == "20170103T160500"
+        assert stop_schedules[1]['last_datetime']['date_time'] == "20170103T170500"
 
+        response = self.query_region("stop_points/X_S1/stop_schedules?from_datetime={}"
+                                     "&data_freshness=base_schedule".format(from_datetime))
+
+        stop_schedules = response['stop_schedules']
+        assert len(stop_schedules) == 2
+        assert stop_schedules[0]['first_datetime']['date_time'] == "20170103T070000"
+        assert stop_schedules[0]['last_datetime']['date_time'] == "20170102T090000"
+        # since it's the last stop_point, we cannot get on the bus....
+        assert 'first_datetime' not in stop_schedules[1]
+        assert 'last_datetime' not in stop_schedules[1]
+
+    def test_first_last_datetime_diff_validity(self):
+        if not app.config['USE_SERPY']:
+            return
         """              20170102                                                20170103
                                               from_datetime 20170103T0300       
                                                          |  
@@ -1178,13 +1218,27 @@ class TestFirstLastDatetimeWithNegativeTimezone(AbstractTestFixture):
          X_s1 ------------ X_S2 ------------ X_S3                X_s1 ------------ X_S2 ------------ X_S3
 
         """
-        response = self.query_region("stop_points/X_S2/stop_schedules?from_datetime=20170103T0300"
-                                     "&data_freshness=base_schedule")
+        from_datetime = "20170103T0300"
+        response = self.query_region("stop_points/X_S2/stop_schedules?from_datetime={}"
+                                     "&data_freshness=base_schedule".format(from_datetime))
 
         stop_schedules = response['stop_schedules']
         assert stop_schedules
         assert stop_schedules[0]['first_datetime']['date_time'] == "20170103T150000"
         assert stop_schedules[0]['last_datetime']['date_time'] == "20170103T170000"
+        assert stop_schedules[1]['first_datetime']['date_time'] == "20170103T160500"
+        assert stop_schedules[1]['last_datetime']['date_time'] == "20170103T170500"
+
+        response = self.query_region("stop_points/X_S1/stop_schedules?from_datetime={}"
+                                     "&data_freshness=base_schedule".format(from_datetime))
+
+        stop_schedules = response['stop_schedules']
+        assert len(stop_schedules) == 2
+        assert stop_schedules[0]['first_datetime']['date_time'] == "20170103T070000"
+        assert stop_schedules[0]['last_datetime']['date_time'] == "20170102T090000"
+        # since it's the last stop_point, we cannot get on the bus....
+        assert 'first_datetime' not in stop_schedules[1]
+        assert 'last_datetime' not in stop_schedules[1]
 
         """ 
                         20170103                                                  20170104    
@@ -1196,14 +1250,31 @@ class TestFirstLastDatetimeWithNegativeTimezone(AbstractTestFixture):
          X_s1 ------------ X_S2 ------------ X_S3                X_s1 ------------ X_S2 ------------ X_S3
 
         """
-        response = self.query_region("stop_points/X_S2/stop_schedules?from_datetime=20170104T0300"
-                                     "&data_freshness=base_schedule")
+        from_datetime = "20170104T0300"
+        response = self.query_region("stop_points/X_S2/stop_schedules?from_datetime={}"
+                                     "&data_freshness=base_schedule".format(from_datetime))
 
         stop_schedules = response['stop_schedules']
         assert stop_schedules
         assert stop_schedules[0]['first_datetime']['date_time'] == "20170104T150500"
         assert stop_schedules[0]['last_datetime']['date_time'] == "20170104T170500"
+        assert stop_schedules[1]['first_datetime']['date_time'] == "20170104T160500"
+        assert stop_schedules[1]['last_datetime']['date_time'] == "20170104T170500"
 
+        response = self.query_region("stop_points/X_S1/stop_schedules?from_datetime={}"
+                                     "&data_freshness=base_schedule".format(from_datetime))
+
+        stop_schedules = response['stop_schedules']
+        assert len(stop_schedules) == 2
+        assert stop_schedules[0]['first_datetime']['date_time'] == "20170104T070500"
+        assert stop_schedules[0]['last_datetime']['date_time'] == "20170104T090500"
+        # since it's the last stop_point, we cannot get on the bus....
+        assert 'first_datetime' not in stop_schedules[1]
+        assert 'last_datetime' not in stop_schedules[1]
+
+    def test_first_last_datetime_another_validity(self):
+        if not app.config['USE_SERPY']:
+            return
         """              20170104
                                from_datetime 20170105T0030       
                                           |  
@@ -1213,40 +1284,59 @@ class TestFirstLastDatetimeWithNegativeTimezone(AbstractTestFixture):
          X_s1 ------------ X_S2 ------------ X_S3
 
         """
-        response = self.query_region("stop_points/X_S2/stop_schedules?from_datetime=20170105T0030"
-                                     "&data_freshness=base_schedule")
+        from_datetime = "20170105T0030"
+        response = self.query_region("stop_points/X_S2/stop_schedules?from_datetime={}"
+                                     "&data_freshness=base_schedule".format(from_datetime))
 
         stop_schedules = response['stop_schedules']
         assert stop_schedules
         assert stop_schedules[0]['first_datetime']['date_time'] == "20170105T150500"
         assert stop_schedules[0]['last_datetime']['date_time'] == "20170105T170500"
+        assert stop_schedules[1]['first_datetime']['date_time'] == "20170105T160500"
+        assert stop_schedules[1]['last_datetime']['date_time'] == "20170105T170500"
+
+        response = self.query_region("stop_points/X_S1/stop_schedules?from_datetime={}"
+                                     "&data_freshness=base_schedule".format(from_datetime))
+
+        stop_schedules = response['stop_schedules']
+        assert len(stop_schedules) == 2
+        assert stop_schedules[0]['first_datetime']['date_time'] == "20170105T070500"
+        assert stop_schedules[0]['last_datetime']['date_time'] == "20170105T090500"
+        # since it's the last stop_point, we cannot get on the bus....
+        assert 'first_datetime' not in stop_schedules[1]
+        assert 'last_datetime' not in stop_schedules[1]
 
 
 @dataset({"timezone_hong_kong_test": {}})
 class TestFirstLastDatetimeWithPositiveTimezone(AbstractTestFixture):
-    def test_first_last_datetime(self):
-        """
-        test first/last datetime with a timezone whose UTC offset is **positive**
-        (opening_time - utc_offset < 0)
+    """
+    test first/last datetime with a timezone whose UTC offset is **positive**
+    (opening_time - utc_offset < 0)
 
-        The closing time of line X is 07:25(local time) and the opening time is 6:00(local time)
+    The closing time of line X is 07:25(local time) and the opening time is 6:00(local time)
 
-        Datetimes shown in the figure are all **UTC** datetimes
+    Datetimes shown in the figure are all **UTC** datetimes
 
-        [2017T0101, 2017T0103] and [2017T0106, 2017T0108]
-        X_s1 ------------ X_S2 ------------ X_S3
-        06:00             06:30             07:00    X:vj1
-        06:10             06:40             07:10    X:vj2
-        06:20             06:50             07:20    X:vj3
+    [2017T0101, 2017T0103] and [2017T0106, 2017T0108]
+    X_s1 ------------ X_S2 ------------ X_S3
+    06:00             06:30             07:00    X:vj1    route_1
+    06:10             06:40             07:10    X:vj2    route_1
+    06:20             06:50             07:20    X:vj3    route_1
 
-        [2017T0104, 2017T0105]
-        X_s1 ------------ X_S2 ------------ X_S3
-        06:05             06:35             07:05    X:vj4
-        06:05             06:45             07:15    X:vj5
-        06:05             06:55             07:25    X:vj6
+    [2017T0104, 2017T0105]
+    X_s1 ------------ X_S2 ------------ X_S3
+    06:05             06:35             07:05    X:vj4    route_1
+    06:05             06:45             07:15    X:vj5    route_1
+    06:05             06:55             07:25    X:vj6    route_1
 
-        """
 
+    [2017T0101, 2017T0108]
+    X_s3 ------------ X_S2 ------------ X_S1
+    06:20             06:50             07:20    X:vj7    route_2
+    06:25             06:55             07:25    X:vj8    route_2
+    """
+
+    def test_first_last_datetime_same_day(self):
         if not app.config['USE_SERPY']:
             return
 
@@ -1259,13 +1349,27 @@ class TestFirstLastDatetimeWithPositiveTimezone(AbstractTestFixture):
          X_s1 ------------ X_S2 ------------ X_S3
 
         """
-        response = self.query_region("stop_points/X_S2/stop_schedules?from_datetime=20170102T0631"
-                                     "&data_freshness=base_schedule")
+        from_datetime = "20170102T0631"
+        response = self.query_region("stop_points/X_S2/stop_schedules?from_datetime={}"
+                                     "&data_freshness=base_schedule".format(from_datetime))
 
         stop_schedules = response['stop_schedules']
-        assert stop_schedules
+        assert len(stop_schedules) == 2
         assert stop_schedules[0]['first_datetime']['date_time'] == "20170102T063000"
         assert stop_schedules[0]['last_datetime']['date_time'] == "20170102T065000"
+        assert stop_schedules[1]['first_datetime']['date_time'] == "20170102T065000"
+        assert stop_schedules[1]['last_datetime']['date_time'] == "20170102T065500"
+
+        response = self.query_region("stop_points/X_S1/stop_schedules?from_datetime={}"
+                                     "&data_freshness=base_schedule".format(from_datetime))
+
+        stop_schedules = response['stop_schedules']
+        assert len(stop_schedules) == 2
+        assert stop_schedules[0]['first_datetime']['date_time'] == "20170103T060000"
+        assert stop_schedules[0]['last_datetime']['date_time'] == "20170102T062000"
+        # since it's the last stop_point, we cannot get on the bus....
+        assert 'first_datetime' not in stop_schedules[1]
+        assert 'last_datetime' not in stop_schedules[1]
 
         """              20170102
                                from_datetime 20170102T0700
@@ -1276,14 +1380,31 @@ class TestFirstLastDatetimeWithPositiveTimezone(AbstractTestFixture):
          X_s1 ------------ X_S2 ------------ X_S3
 
         """
-        response = self.query_region("stop_points/X_S2/stop_schedules?from_datetime=20170103T0001"
-                                     "&data_freshness=base_schedule")
+        from_datetime = "20170103T0001"
+        response = self.query_region("stop_points/X_S2/stop_schedules?from_datetime={}"
+                                     "&data_freshness=base_schedule".format(from_datetime))
 
         stop_schedules = response['stop_schedules']
-        assert stop_schedules
+        assert len(stop_schedules) == 2
         assert stop_schedules[0]['first_datetime']['date_time'] == "20170103T063000"
         assert stop_schedules[0]['last_datetime']['date_time'] == "20170103T065000"
+        assert stop_schedules[1]['first_datetime']['date_time'] == "20170103T065000"
+        assert stop_schedules[1]['last_datetime']['date_time'] == "20170103T065500"
 
+        response = self.query_region("stop_points/X_S1/stop_schedules?from_datetime={}"
+                                     "&data_freshness=base_schedule".format(from_datetime))
+
+        stop_schedules = response['stop_schedules']
+        assert len(stop_schedules) == 2
+        assert stop_schedules[0]['first_datetime']['date_time'] == "20170103T060000"
+        assert stop_schedules[0]['last_datetime']['date_time'] == "20170102T062000"
+        # since it's the last stop_point, we cannot get on the bus....
+        assert 'first_datetime' not in stop_schedules[1]
+        assert 'last_datetime' not in stop_schedules[1]
+
+    def test_first_last_datetime_diff_validity(self):
+        if not app.config['USE_SERPY']:
+            return
         """              20170102                                                20170103
                                               from_datetime 20170103T0300       
                                                          |  
@@ -1293,13 +1414,27 @@ class TestFirstLastDatetimeWithPositiveTimezone(AbstractTestFixture):
          X_s1 ------------ X_S2 ------------ X_S3                X_s1 ------------ X_S2 ------------ X_S3
 
         """
-        response = self.query_region("stop_points/X_S2/stop_schedules?from_datetime=20170103T0300"
-                                     "&data_freshness=base_schedule")
+        from_datetime = "20170103T0300"
+        response = self.query_region("stop_points/X_S2/stop_schedules?from_datetime={}"
+                                     "&data_freshness=base_schedule".format(from_datetime))
 
         stop_schedules = response['stop_schedules']
-        assert stop_schedules
+        assert len(stop_schedules) == 2
         assert stop_schedules[0]['first_datetime']['date_time'] == "20170103T063000"
         assert stop_schedules[0]['last_datetime']['date_time'] == "20170103T065000"
+        assert stop_schedules[1]['first_datetime']['date_time'] == "20170103T065000"
+        assert stop_schedules[1]['last_datetime']['date_time'] == "20170103T065500"
+
+        response = self.query_region("stop_points/X_S1/stop_schedules?from_datetime={}"
+                                     "&data_freshness=base_schedule".format(from_datetime))
+
+        stop_schedules = response['stop_schedules']
+        assert len(stop_schedules) == 2
+        assert stop_schedules[0]['first_datetime']['date_time'] == "20170103T060000"
+        assert stop_schedules[0]['last_datetime']['date_time'] == "20170102T062000"
+        # since it's the last stop_point, we cannot get on the bus....
+        assert 'first_datetime' not in stop_schedules[1]
+        assert 'last_datetime' not in stop_schedules[1]
 
         """ 
                         20170104                                                  20170105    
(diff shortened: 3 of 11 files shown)


source/jormungandr/tests/departure_board_tests.py
```diff
@@ -1104,3 +1103,107 @@ def test_routes_schedule_with_invalid_forbidden_uri(self):
...
         assert code == 404
+
+
+@dataset({"timezone_cape_verde_test": {}})
+class TestFirstLastDatetime(AbstractTestFixture):
+    def test_first_last_datetime(self):
+        """
+        test first/last datetime
+
+        * The opening time should be the earliest departure of all vjs of the line
```
pbougue: Those conditions are not mandatory, no? So you're not testing it here, it just in the data provided.
In **data**, we should have `latest_arrival >= opening >= closing >= first_departure` (and ideally your comment is true)

source/jormungandr/tests/departure_board_tests.py
```diff
@@ -1104,3 +1103,107 @@ def test_routes_schedule_with_invalid_forbidden_uri(self):
...
+        08:00             15:00             26:45  
+         X_s1 ------------ X_S2 ------------ X_S3
+         
+        """
+        response = self.query_region("stop_points/X_S2/stop_schedules?from_datetime=20170102T1630"
+                                     "&data_freshness=base_schedule")
+
+        stop_schedules = response['stop_schedules']
+        assert stop_schedules
+        assert stop_schedules[0]['first_datetime']['date_time'] == "20171002T080000"
```
pbougue: shouldn't it be 16:00 and 18:00 one way and same the other way, as we are at X_S2 ?

source/jormungandr/tests/departure_board_tests.py
```diff
@@ -1104,3 +1103,107 @@ def test_routes_schedule_with_invalid_forbidden_uri(self):
...
+        08:00             15:00             26:45  
+         X_s1 ------------ X_S2 ------------ X_S3
+
+        """
+        response = self.query_region("stop_points/X_S2/stop_schedules?from_datetime=20170103T000000"
+                                     "&data_freshness=base_schedule")
+
+        stop_schedules = response['stop_schedules']
+        assert stop_schedules
+        assert stop_schedules[0]['first_datetime']['date_time'] == "20171002T080000"
```
pbougue: most of these should probably tested on all stop_points, and for both routes (would be nice to have 2 routes in data if it's not the case)

source/jormungandr/tests/departure_board_tests.py
```diff
@@ -1104,3 +1103,107 @@ def test_routes_schedule_with_invalid_forbidden_uri(self):
...
+                                     "&data_freshness=base_schedule")
+
+        stop_schedules = response['stop_schedules']
+        assert stop_schedules
+        assert stop_schedules[0]['first_datetime']['date_time'] == "20171003T080000"
+        assert stop_schedules[0]['last_datetime']['date_time'] == "20171003T023000"
+
+        """ 
+                        20170103                                                  20170104    
+                                             from_datetime 20170103T0300
```
pbougue: Here, from the rest of your comment, it's the 4th you are testing, not the 3rd of january.

source/jormungandr/tests/departure_board_tests.py
```diff
@@ -1104,3 +1103,107 @@ def test_routes_schedule_with_invalid_forbidden_uri(self):
...
+                        20170103                                                  20170104    
+                                             from_datetime 20170103T0300        
+                                                        |  
+                                                        v
+        (Open)                             (Close)              (Open)                             (Close)
+        08:00             15:00             26:45      03:00    08:00             15:00             26:45  
+         X_s1 ------------ X_S2 ------------ X_S3                X_s1 ------------ X_S2 ------------ X_S3
+
+        """
+        response = self.query_region("stop_points/X_S2/stop_schedules?from_datetime=20170103T0300"
```
pbougue: 201701`04`, no?

source/jormungandr/tests/departure_board_tests.py
```diff
@@ -1104,3 +1103,107 @@ def test_routes_schedule_with_invalid_forbidden_uri(self):
...
+                                     "&data_freshness=base_schedule")
+
+        stop_schedules = response['stop_schedules']
+        assert stop_schedules
+        assert stop_schedules[0]['first_datetime']['date_time'] == "20170104T080500"
+        assert stop_schedules[0]['last_datetime']['date_time'] == "20170104T024500"
+
+        """              20170104
+                               from_datetime 20170105T0030
```
pbougue: - [x] Should be nice to test also at 2:40 local (one way open the other closed) between 2 working days.
- [x] Would be nice to test also that nothing breaks after the closing of the last VJ (or before opening of the first) : 20170105T040000, 20170107T000000 and 20170101T040000
- [ ] Can't we anticipate/test bad behavior having in data local closing time between 2am and 3am... Probably not nice on DST change? And probably more frequent to have those hours for open/close than for real stop_times.

source/tests/timezone_cape_verde_test.cpp
```diff
@@ -72,6 +72,22 @@ int main(int argc, const char* const argv[]) {
...
+    b.vj("X", "11100111").uri("X:vj1")("X_S1", "09:00"_t)("X_S2", "17:00"_t)("X_S3", "25:30"_t).make();
+    b.vj("X", "11100111").uri("X:vj1")("X_S1", "10:00"_t)("X_S2", "18:00"_t)("X_S3", "26:30"_t).make();
+
+    b.vj("X", "00011000").uri("X:vj2")("X_S1", "08:05"_t)("X_S2", "16:00"_t)("X_S3", "24:45"_t).make();
+    b.vj("X", "00011000").uri("X:vj2")("X_S1", "09:05"_t)("X_S2", "17:00"_t)("X_S3", "25:45"_t).make();
+    b.vj("X", "00011000").uri("X:vj2")("X_S1", "10:05"_t)("X_S2", "18:00"_t)("X_S3", "26:45"_t).make();
+    auto line_X = b.lines.find("X")->second;
+
+    // opening time should be the earliest departure of all vjs of the line
+    line_X->opening_time = boost::make_optional(boost::posix_time::duration_from_string("08:00"));
```
pbougue: This may deserve tests on another TZ to check it works fine (with DST? :-/ ).

source/time_tables/tests/departure_boards_test.cpp
```diff
@@ -205,6 +211,108 @@ BOOST_AUTO_TEST_CASE(departureboard_test1) {
...
+        auto * data_ptr = b.data.get();
+        navitia::PbCreator pb_creator1(data_ptr, bt::second_clock::universal_time(), null_time_period);
+        departure_board(pb_creator1, "stop_point.uri=stop1", {}, {}, d("20150615T094500"), 43200, 0,
+                        10, 0, nt::RTLevel::Base, std::numeric_limits<size_t>::max());
+
+        resp = pb_creator1.get_response();
+        BOOST_REQUIRE_EQUAL(resp.stop_schedules_size(), 2);
+        BOOST_REQUIRE_EQUAL(resp.stop_schedules(0).has_first_datetime(),true);
+        BOOST_REQUIRE_EQUAL(resp.stop_schedules(0).has_last_datetime(), true);
+        BOOST_REQUIRE_EQUAL(resp.stop_schedules(0).first_datetime().time(), "06:00"_t);
```
pbougue: Isn't that test redundant? Better testing whole dt than just time, no?
Same for all those tests

source/time_tables/tests/departure_boards_test.cpp
```diff
@@ -205,6 +211,108 @@ BOOST_AUTO_TEST_CASE(departureboard_test1) {
...
+
+        // Case 2 :
+        //
+        // Note
+        // - Classical test with a closing date time the same day.
+        //
+        // Input
+        // - opening date time : 05:30:00
+        // - closing date time : 09:00:00
+        // - request date time : 09:45:00
```
pbougue: So we are on the no-service hole, here no?

source/time_tables/tests/departure_boards_test.cpp
```diff
@@ -205,6 +211,108 @@ BOOST_AUTO_TEST_CASE(departureboard_test1) {
...
+        departure_board(pb_creator3, "stop_point.uri=stop1", {}, {}, d("20150615T020000"), 86400, 0,
+                        10, 0, nt::RTLevel::Base, std::numeric_limits<size_t>::max());
+
+        resp = pb_creator3.get_response();
+        BOOST_REQUIRE_EQUAL(resp.stop_schedules_size(), 2);
+        BOOST_REQUIRE_EQUAL(resp.stop_schedules(0).has_first_datetime(),true);
+        BOOST_REQUIRE_EQUAL(resp.stop_schedules(0).has_last_datetime(), true);
+        BOOST_REQUIRE_EQUAL(resp.stop_schedules(0).first_datetime().time(), "06:00"_t);
+        BOOST_REQUIRE_EQUAL(resp.stop_schedules(0).first_datetime().base_date_time(), "20150615T060000"_pts);
+        BOOST_REQUIRE_EQUAL(resp.stop_schedules(0).last_datetime().time(), "10:00"_t);
```
pbougue, line 341: Looks important to test it's on the good day also, not only the time

source/time_tables/departure_boards.cpp
```diff
@@ -270,11 +374,80 @@ void departure_board(PbCreator& pb_creator, const std::string& request,
...
+    if (type == datetime_type::closing) {
+        max_dt = 0; // clockwise == false
+    }
+    std::vector<routing::datetime_stop_time> stop_times;
+    DateTime new_current_time = convert_duration_into_datetime(type,
+                                                               current_datetime,
+                                                               int(opening_time.total_seconds() - utc_offset),
+                                                               int(closing_time.total_seconds() - utc_offset));
+    if (calendar_id) {
+
```
pbougue: minor: useless blank line  (same below)

source/time_tables/departure_boards.cpp
```diff
@@ -270,11 +374,80 @@ void departure_board(PbCreator& pb_creator, const std::string& request,
...
+        } else {
+            return DateTimeUtils::date(date)*DateTimeUtils::SECONDS_PER_DAY + closing_time;
+        }
+    }
 }
+
+boost::optional<routing::datetime_stop_time>
+get_one_stop_time(const datetime_type type,
+                  const boost::optional<const std::string>& calendar_id,
+                  const DateTime current_datetime,
```
pbougue: is it the current_dt or the requested_dt?

source/time_tables/departure_boards.cpp
```diff
@@ -270,11 +374,80 @@ void departure_board(PbCreator& pb_creator, const std::string& request,
...
+           map_route_last_point,
+           handler.date_time,
+           handler.max_datetime,
+           calendar_id, depth);
 
     pb_creator.make_paginate(total_result, start_page, count, std::max(pb_creator.departure_boards_size(),
                                                                        pb_creator.stop_schedules_size()));
 }
+
+DateTime convert_duration_into_datetime(const datetime_type type,
```
pbougue: * The signature of this method is a bit weird.
Can't we return a struct as a response that contains both first and last, no ambiguity?

* Also, does it work for a Noctilien line? open=1am, close=5am, resquested=11pm > we want last_dt from that day, but first_dt to be the day after I guess (from other tests).
You probably want to use a complete requested_datetime, not just the date (as you want last_dt and first_dt to be always surrounding your requested_dt, first_dt and last_dt can be reverted)

source/time_tables/departure_boards.cpp
```diff
@@ -222,15 +252,55 @@ void departure_board(PbCreator& pb_creator, const std::string& request,
     // since we want to display the departures grouped by route
     // the route being a loose commercial direction
     for (const auto& route_point: route_points) {
+
         const type::StopPoint* stop_point = pb_creator.data->pt_data->stop_points[route_point.second.val];
         const type::Route* route = pb_creator.data->pt_data->routes[route_point.first.val];
+
         const auto routepoint_jpps = get_jpp_from_route_point(route_point, *pb_creator.data->dataRaptor);
+        std::cout << "route jpps " << routepoint_jpps.size() << std::endl;
```
pbougue: debug leftover

source/time_tables/departure_boards.h
```diff
@@ -59,5 +70,29 @@ bool line_closed (const time_duration& duration,
                   const time_duration& opening,
                   const time_duration& closing,
                   const pt::ptime& date );
+/**
+ * @brief Convert a posix_time::time_duration in DateTime (uint32_t).
+ * Usefull for opening and closing date time.
+ */
+DateTime convert_duration_into_datetime(const datetime_type type,
+                                        const DateTime date,
+                                        const int opening_time,
```
pbougue: Why not a time_duration type here (and line just below)?

source/time_tables/departure_boards.h
```diff
@@ -59,5 +70,29 @@ bool line_closed (const time_duration& duration,
...
+get_one_stop_time(const datetime_type type,
+                  const boost::optional<const std::string>& calendar_id,
+                  const DateTime current_datetime,
+                  const pt::time_duration& opening_time,
+                  const pt::time_duration& closing_time,
+                  const std::vector<routing::JppIdx>& journey_pattern_points,
+                  const DateTime max_datetime,
+                  const type::Data& data,
+                  const type::RTLevel rt_level,
+                  const size_t utc_offset);
```
pbougue: Is it gonna work around DST change? Shouldn't you use TZ directly instead?

pbougue (commented):
Not sure I get all the spirit/archi of this PR, so maybe all remarks are not accurate...

source/jormungandr/tests/departure_board_tests.py
xlqian: that's the general definition of the closing time and the opening time, confirmed by @patochectp

source/cmake_modules/CompilationFlags.cmake
```diff
@@ -14,7 +14,7 @@ ENDIF(NOT CMAKE_BUILD_TYPE)
 # Gcc Release flags
 if(CMAKE_COMPILER_IS_GNUCXX)
     set(CMAKE_GNUCXX_WARN_FLAGS "-Wall -Wextra -Woverloaded-virtual -Wundef -pedantic")
-    set(CMAKE_GNUCXX_COMMON_FLAGS "-std=c++0x -rdynamic -g")
+    set(CMAKE_GNUCXX_COMMON_FLAGS "-std=c++0x -rdynamic -g  -fno-omit-frame-pointer")
```
kinnou02: Maybe not in release/production

source/jormungandr/jormungandr/interfaces/v1/serializer/schedule.py
```diff
@@ -89,6 +89,8 @@ class StopScheduleSerializer(PbNestedSerializer):
     display_informations = pt.RouteDisplayInformationSerializer(attr='pt_display_informations')
     date_times = DateTimeTypeSerializer(many=True, display_none=True)
     links = jsonschema.MethodField(schema_type=LinkSchema(many=True))
+    first_datetime = DateTimeTypeSerializer(display_none=True)
```
kinnou02: really?
Why bloat the response if there is information to display

source/time_tables/departure_boards.h
```diff
@@ -1,53 +1,64 @@
...
 */
-
 #pragma once
+
 #include "type/pb_converter.h"
 #include "routing/routing.h"
 #include "routing/get_stop_times.h"
+#include "routing/raptor.h"
+
+namespace pt = boost::posix_time;
```
kinnou02: namespace alias in headers isn't really nice, even if we already have a lot of them

source/time_tables/departure_boards.cpp
```diff
@@ -270,11 +338,65 @@ void departure_board(PbCreator& pb_creator, const std::string& request,
...
                                                                        pb_creator.stop_schedules_size()));
 }
+
+DateTime first_or_last_request_datetime(const datetime_type type,
+                                        const routing::datetime_stop_time& first_stop_time,
+                                        const int opening_time)
+{
+    if (type == datetime_type::first) {
+        return DateTimeUtils::date(first_stop_time.first)*DateTimeUtils::SECONDS_PER_DAY + opening_time - 1;
+    } else {
```
kinnou02: No need for an else after a return.
Also a bit of explanation about what is being done will really help

source/time_tables/departure_boards.cpp
```diff
@@ -270,11 +338,65 @@ void departure_board(PbCreator& pb_creator, const std::string& request,
...
+        if (first_stop_time.second->departure_time <= opening_time) {
+            return (DateTimeUtils::date(first_stop_time.first))*DateTimeUtils::SECONDS_PER_DAY + opening_time - 1;
+        } else {
+            return (DateTimeUtils::date(first_stop_time.first) + 1)*DateTimeUtils::SECONDS_PER_DAY + opening_time - 1;
+        }
+    }
 }
+
+boost::optional<routing::datetime_stop_time>
+get_one_stop_time(const datetime_type type,
```
kinnou02: Not sure about this `datetime_type` it doesn't help to read the code

source/time_tables/departure_boards.cpp
```diff
@@ -270,11 +338,65 @@ void departure_board(PbCreator& pb_creator, const std::string& request,
...
+                  const pt::time_duration& opening_time,
+                  const std::vector<routing::JppIdx>& journey_pattern_points,
+                  const DateTime max_datetime,
+                  const type::Data& data,
+                  const type::RTLevel rt_level,
+                  const size_t utc_offset)
+{
+    DateTime max_dt = max_datetime;
+    if (type == datetime_type::last) {
+        max_dt = 0; // clockwise == false
```
kinnou02: Not sure about this, but max_datetime can be smaller than one day, it can only be a few hours and not include the end of the service, doesn't this change the result?

source/time_tables/departure_boards.cpp
```diff
@@ -42,7 +42,17 @@ www.navitia.io
 
 namespace pt = boost::posix_time;
 
-namespace navitia { namespace timetables {
+namespace {
+    // mathematical modulus
+    // math_mod(1, 5) == 1
+    // math_mod(-1, 5) == 4
+    int math_mod(int x, int m){
+        return (x%m + m)%m;
```
nonifier: Jesus Christ.... I never realised the modulus function was symmetrical around zero :sob:  
It would probably make more sense to have this in `utils` or somewhere this could be reused

source/time_tables/departure_boards.cpp
```diff
@@ -113,11 +127,29 @@ render(PbCreator& pb_creator,
                 }
             }
         }
+
+        // add first datetime
+        const auto& first_it = map_route_first_point.find(id_vec.first);
```
nonifier: I see duplicated code here !

source/time_tables/departure_boards.cpp
```diff
@@ -270,11 +346,75 @@ void departure_board(PbCreator& pb_creator, const std::string& request,
...
+           calendar_id, depth);
 
     pb_creator.make_paginate(total_result, start_page, count, std::max(pb_creator.departure_boards_size(),
                                                                        pb_creator.stop_schedules_size()));
 }
+
+DateTime first_or_last_request_datetime(const datetime_type type,
+                                        const routing::datetime_stop_time& first_stop_time,
+                                        const uint32_t opening_time)
+{
```
nonifier: After spending time with @xlqian, it seems that `first_stop_time` actually is a `next_stop_time` :)

Also, returning the 2 values for opening and closing datetimes at the same time would simplify the readability. Maybe having something like:

```c++
std::pair<> get_opening_and_closing_datetime(
	const DateTime next_stoptime_datetime,
	const uint32_t opening_time)
{
	
	auto opening_datetime = DateTimeUtils::date(next_stoptime_datetime);
	auto opening_datetime = closing_datetime;

	if(next_stoptime_datetime > opening_time) {
	    opening_datetime += 1;
	}

	return make_pair( 
				DateTimeUtils::set(opening_datetime, opening_time)-1, 
				DateTimeUtils::set(opening_datetime, opening_time)-1);
}
```

source/time_tables/departure_boards.cpp
```diff
@@ -270,11 +346,75 @@ void departure_board(PbCreator& pb_creator, const std::string& request,
...
+                                         journey_pattern_points,
+                                         new_current_time,
+                                         max_dt,
+                                         1,
+                                         data,
+                                         rt_level);
+    if (!stop_times.empty()) {
+        return stop_times[0];
+    } else {
+        return boost::optional<routing::datetime_stop_time>{};
```
nonifier: An empty optional can simply be returned with: `boost::none`

nonifier (changes requested):

source/time_tables/departure_boards.cpp
kinnou02: And we would be able to test it too :)

source/time_tables/departure_boards.cpp
benoit-bst: :+1:

source/time_tables/departure_boards.cpp
benoit-bst: Yes it is, because we have 2 maps, one for `first` and one for `last`, and the key is a particular route point. Thus, we have same code for both, but not on the same map.
We can find a way to reduce code, `map_route_first_point` and `map_route_last_point` can become a map with a pair like : 

```
map<pair<first,last>>
```
I'll change that

source/time_tables/departure_boards.cpp
benoit-bst: This is not so easy, but you gave me a good way to refactor this part. In any case, the `std::pair` in return and *variable name* is a good idea.

source/jormungandr/tests/departure_board_tests.py
```diff
@@ -1104,3 +1103,402 @@ def test_routes_schedule_with_invalid_forbidden_uri(self):
...
+        """              20170102
+                               from_datetime 20170102T1630       
+                                    |  
+                                    v
+        (Open)                             (Close)
+        07:00                               25:45  
+         X_s1 ------------ X_S2 ------------ X_S3
+         
+        """
+        from_datetime = "20170102T1630"
```
kinnou02: The format should be `20170102T163000`
And can you add a test with a very small duration plz

source/time_tables/departure_boards.h
xlqian: we don't handle dst

source/tests/timezone_cape_verde_test.cpp
xlqian: we won't handle DST for this feature, it's kind of hard to do(both implementation and test) and the benefit is minor. I'd prefer wait for the bug that will be reported by the client with a concrete example

source/jormungandr/tests/departure_board_tests.py
xlqian: ok

xlqian:
@pbougue I changed the integration tests a lot... by adding another route and another dataset with positive utc offset... your comments are partially taken into account... i don't understand all, we can talk about them if you want

source/tests/timezone_cape_verde_test.cpp
```diff
@@ -72,7 +72,41 @@ int main(int argc, const char* const argv[]) {
...
+    		.uri("X:vj8")("X_S4", "09:55"_t)("X_S3", "10:05"_t)("X_S2", "18:05"_t)("X_S1", "26:45"_t).make();
+
+    auto line_X = b.lines.find("X")->second;
+
+    // the opening time should be the earliest departure of all vjs of the line and it's stored as local time
+    line_X->opening_time = boost::make_optional(boost::posix_time::duration_from_string("07:00"));
+    // the closing time should be the latest arrival of all vjs of the line and it's stored as local time
+    line_X->closing_time = boost::make_optional(boost::posix_time::duration_from_string("26:55"));
+
+    b.generate_dummy_basis();
```
nonifier, line 119: minor: there is now a `void builder::make();` function that builds the all things for you ;)

source/tests/timezone_hong_kong_test.cpp
```diff
@@ -0,0 +1,98 @@
...
+ * DST.
+ */
+int main(int argc, const char* const argv[]) {
+    navitia::init_app();
+
+    const auto date = "20170101";
+    const auto begin = boost::gregorian::date_from_iso_string(date);
+    const boost::gregorian::date_period production_date = {begin, begin + boost::gregorian::years(1)};
+    navitia::type::TimeZoneHandler::dst_periods timezones = {{"08:00"_t, {production_date}}};
+    ed::builder b("20170101", "canal tp", "Asia/Hong_Kong", timezones);
```
nonifier, line 51: Hehe we're now ready to ship Navitia to Asia :)

nonifier (approved):

benoit-bst (approved):

source/time_tables/tests/departure_boards_test.cpp
```diff
@@ -207,6 +212,172 @@ BOOST_AUTO_TEST_CASE(departureboard_test1) {
     }
 }
 
+BOOST_AUTO_TEST_CASE(first_last_test1) {
+    ed::builder b("20150615");
+    b.vj("A", "11111111111111", "", true, "vj1", "")
+        ("stop1", "06:00"_t, "06:00"_t)
+        ("stop1", "07:00"_t, "07:00"_t)
+        ("stop1", "10:00"_t, "10:00"_t)
```
xlqian, line 220: your bus have stayed on the same stop(stop1) for 4 hours... :D

TeXitoi: That's a touristic bus :wink:

source/time_tables/departure_boards.h
```diff
@@ -58,6 +63,17 @@ time_duration length_of_time(const time_duration& duration_1,
 bool line_closed (const time_duration& duration,
                   const time_duration& opening,
                   const time_duration& closing,
-                  const pt::ptime& date );
+                  const boost::posix_time::ptime& date );
+
+std::pair<DateTime, DateTime> shift_next_stop_time_to_opening_time(const routing::datetime_stop_time& next_stop_time,
+                                                                   const uint32_t opening_time);
```
pbougue: Should this be a `pt::time_duration` ?

source/time_tables/departure_boards.h
```diff
@@ -58,6 +63,17 @@ time_duration length_of_time(const time_duration& duration_1,
...
+first_and_last_stop_time get_first_and_last_stop_time(const routing::datetime_stop_time& next_stop_time,
+                                                      const pt::time_duration& opening_time,
```
pbougue: So we don't need closing_time for that method because it's kind of redundant with opening time and may raise consistency issue like stop-times after closing time but before opening time ? Am I right ?
minor: If it's that, it could be in a comment.

source/time_tables/departure_boards.cpp
```diff
@@ -222,15 +258,34 @@ void departure_board(PbCreator& pb_creator, const std::string& request,
...
+            if (route->line->opening_time && !stop_times.empty()) {
+
+                // retrieve utc offset
+                utc_offset = stop_times[0].second->vehicle_journey->utc_to_local_offset();
+
+                // first and last Date time
+                map_route_first_and_last_point[route_point] = get_first_and_last_stop_time(stop_times[0],
+                                                                                           *route->line->opening_time,
+                                                                                           routepoint_jpps,
+                                                                                           handler.date_time + DateTimeUtils::SECONDS_PER_DAY,
```
pbougue: minor: line length

source/time_tables/departure_boards.cpp
```diff
@@ -222,15 +258,34 @@ void departure_board(PbCreator& pb_creator, const std::string& request,
...
                     handler.max_datetime, items_per_route_point, *pb_creator.data, rt_level);
             std::sort(stop_times.begin(), stop_times.end(), sort_predicate);
+
+            if (route->line->opening_time && !stop_times.empty()) {
+
+                // retrieve utc offset
+                utc_offset = stop_times[0].second->vehicle_journey->utc_to_local_offset();
+
+                // first and last Date time
+                map_route_first_and_last_point[route_point] = get_first_and_last_stop_time(stop_times[0],
```
pbougue: naming: "map_route`_point`_first`_last_st`" is maybe easier to understand when used elsewhere.

source/time_tables/departure_boards.cpp
```diff
@@ -70,14 +74,17 @@ static void
 render(PbCreator& pb_creator,
        const std::map<RoutePointIdx, pbnavitia::ResponseStatus>& response_status,
        const std::map<RoutePointIdx, vector_dt_st>& map_route_stop_point,
-       DateTime datetime,
-       DateTime max_datetime,
-       boost::optional<const std::string> calendar_id,
-       uint32_t depth) {
+       const std::map<RoutePointIdx, first_and_last_stop_time>& map_route_first_and_last_point,
```
pbougue: naming: suggesting "map_route`_point`_first`_last_st`"

source/time_tables/departure_boards.h
```diff
@@ -58,6 +63,17 @@ time_duration length_of_time(const time_duration& duration_1,
...
+                                                      const std::vector<routing::JppIdx>& journey_pattern_points,
+                                                      const DateTime max_datetime,
+                                                      const type::Data& data,
+                                                      const type::RTLevel rt_level,
+                                                      const int utc_offset);
```
pbougue, line 98: Should this be a `pt::time_duration` also ?

source/time_tables/departure_boards.cpp
```diff
@@ -270,11 +325,82 @@ void departure_board(PbCreator& pb_creator, const std::string& request,
...
+
+first_and_last_stop_time get_first_and_last_stop_time(const routing::datetime_stop_time& next_stop_time,
+                                                      const pt::time_duration& opening_time,
+                                                      const std::vector<routing::JppIdx>& journey_pattern_points,
+                                                      const DateTime max_datetime,
+                                                      const type::Data& data,
+                                                      const type::RTLevel rt_level,
+                                                      const int utc_offset)
+{
+    // if the opening_time is 03:00(local_time) and the utc_offset is 5 hours, we want 22:00(UTC) as opening time
```
pbougue, line 365: minor-minor (it's how GTFS work): and the utc_offset is 5 hours `at noon (local time) that day`,

source/time_tables/departure_boards.cpp
```diff
@@ -270,11 +325,82 @@ void departure_board(PbCreator& pb_creator, const std::string& request,
...
+    int local_opening_time = math_mod(opening_time.total_seconds() - utc_offset, DateTimeUtils::SECONDS_PER_DAY);
+    assert(local_opening_time > 0);
+
+    // Shift the first stop time finded to local opening time. In this way, we can find the first or the last stop time,
```
pbougue: :gb: minor: first stop time `found` [...] . `(no 'in')` This way

source/time_tables/departure_boards.cpp
```diff
@@ -270,11 +325,82 @@ void departure_board(PbCreator& pb_creator, const std::string& request,
...
+    // it depends of the direction (clockwise or not) to get stop time.
```
pbougue: :gb: minor: it depends `on` the direction

source/time_tables/departure_boards.cpp
```diff
@@ -270,11 +325,82 @@ void departure_board(PbCreator& pb_creator, const std::string& request,
...
+    auto new_request_date_times = shift_next_stop_time_to_opening_time(next_stop_time,
+                                                                       static_cast<uint32_t>(local_opening_time));
```
pbougue: :nauseated_face: static_cast

source/time_tables/departure_boards.cpp
```diff
@@ -270,11 +325,82 @@ void departure_board(PbCreator& pb_creator, const std::string& request,
...
+           map_route_first_and_last_point,
+           handler.date_time,
+           handler.max_datetime,
+           calendar_id, depth);
 
     pb_creator.make_paginate(total_result, start_page, count, std::max(pb_creator.departure_boards_size(),
                                                                        pb_creator.stop_schedules_size()));
 }
+
+std::pair<DateTime, DateTime> shift_next_stop_time_to_opening_time(const routing::datetime_stop_time& next_stop_time,
```
pbougue: This part of the code is quite difficult to understand. I see some by order of preference ideas to refactor it (not sure all/any are good):
* split the method in 2 parts: one for first st, one for last st
* use a structure to name the returned values
* rename method to `get_daily_opening_time_bounds`
* add comment (which is what, what we present to traveler depending on the cases and why it's computed that way > maybe where we actually compute first/last)

source/time_tables/departure_boards.cpp
```diff
@@ -270,11 +325,82 @@ void departure_board(PbCreator& pb_creator, const std::string& request,
...
+                                                                   const uint32_t opening_time)
+{
+    uint nb_days = DateTimeUtils::date(next_stop_time.first); // Nb days from the data set origin until the next stop time.
+
+    // sometimes a departure_time(stop_time) may exceed 24H and we suppose it doesn't exceed 48H  
+    uint32_t dep_time = math_mod(next_stop_time.second->departure_time, DateTimeUtils::SECONDS_PER_DAY);
+    if (dep_time < opening_time) {
+        return std::make_pair((nb_days - 1)*DateTimeUtils::SECONDS_PER_DAY + opening_time - 1,
```
pbougue: Does it work fine if `nb_days == 0` ? Same below for `nb_day == 365` ?

source/time_tables/departure_boards.cpp
```diff
@@ -270,11 +325,82 @@ void departure_board(PbCreator& pb_creator, const std::string& request,
...
+                              nb_days*DateTimeUtils::SECONDS_PER_DAY + opening_time - 1);
```
pbougue: shouldn't it be a `+`1 at the end here as we go anticlockwise when using it? same below

source/time_tables/departure_boards.cpp
```diff
@@ -270,11 +325,82 @@ void departure_board(PbCreator& pb_creator, const std::string& request,
...
+                                               new_request_date_times.first,
+                                               max_datetime,
+                                               1,
+                                               data,
+                                               rt_level);
+    // get last stop time
+    auto last_times = routing::get_stop_times(routing::StopEvent::pick_up,
+                                              journey_pattern_points,
+                                              new_request_date_times.second,
+                                              0, // clockwise = false
```
pbougue, line 384: That's a tricky one (but the interface of the method is to blame, if any).
Could be nice to have an actual `Datetime(0)` here and not an implicit cast.

source/time_tables/departure_boards.cpp
```diff
@@ -113,11 +120,34 @@ render(PbCreator& pb_creator,
                 }
             }
         }
+
+        // add first and last datetime
+        const auto& first_it = map_route_first_and_last_point.find(id_vec.first);
```
pbougue: naming: first`_last`_it seems clearer to me (weird to use first_it for the managemant of last, below)

pbougue (commented):
:+1: ++const

Didn't read test so far.

Remarks on naming are here to improve the speed when having to understand the code (maintain, review), and detect mistakes quicker when the spirit doesn't match the code.
Minors on :gb: can be skipped as far as I'm concerned.

source/jormungandr/tests/departure_board_tests.py
```diff
@@ -1104,3 +1103,559 @@ def test_routes_schedule_with_invalid_forbidden_uri(self):
...
         assert code == 404
+
+
+@dataset({"timezone_cape_verde_test": {}})
+class TestFirstLastDatetimeWithNegativeTimezone(AbstractTestFixture):
+
+    """
+    test first/last datetime with a timezone whose UTC offset is **negative**
+
+    * The opening time should be the earliest departure of all vjs of the line
```
pbougue, line 1114: minor: You can mention that opening time is read from the data (we don't want to test it here, and it's a bit ambiguous)

source/tests/timezone_cape_verde_test.cpp
```diff
@@ -72,7 +70,57 @@ int main(int argc, const char* const argv[]) {
...
+        	.uri("X:vj7")
+			("X_S4", "08:55"_t)("X_S3", "09:05"_t)("X_S2", "17:05"_t)("X_S1", "25:45"_t).make();
+    b.vj("X", "11111111")
+    		.route("route_2")
+    		.uri("X:vj8")
+			("X_S4", "09:55"_t)("X_S3", "10:05"_t)("X_S2", "18:05"_t)("X_S1", "26:45"_t).make();
+
+    auto line_X = b.lines.find("X")->second;
+
+    // the opening time should be the earliest departure of all vjs of the line and it's stored as local time
```
pbougue, line 114: minor: "`is supposed to be` the earliest" is less ambiguous

source/tests/timezone_hong_kong_test.cpp
```diff
@@ -0,0 +1,113 @@
...
+    		.uri("X:vj8")
+			("X_S4", "22:05"_t)("X_S3", "22:45"_t)("X_S2", "39:45"_t)("X_S1", "40:25"_t).make();
+
+
+    auto line_X = b.lines.find("X")->second;
+
+    // opening time should be the earliest departure of all vjs of the line and it's stored as local time
+    line_X->opening_time = boost::make_optional(boost::posix_time::duration_from_string("06:00"));
+    // closing time should be the latest arrival of all vjs of the line and it's stored as local time
+    line_X->closing_time = boost::make_optional(boost::posix_time::duration_from_string("00:55"));
```
pbougue, line 100: :+1: have both case (closing > opening and closing < opening)

source/tests/timezone_hong_kong_test.cpp
```diff
@@ -0,0 +1,113 @@
...
+
+    b.generate_dummy_basis();
```
pbougue, line 102: same minor: you can use make()

source/time_tables/tests/departure_boards_test.cpp
```diff
@@ -62,8 +62,13 @@ static pt::ptime d(std::string str) {
 
 BOOST_AUTO_TEST_CASE(departureboard_test1) {
     ed::builder b("20150615");
-    b.vj("A", "110011000001", "", true, "vj1", "")("stop1", 10*3600, 10*3600)("stop2", 10*3600 + 30*60,10*3600 + 30*60);
-    b.vj("B", "110000001111", "", true, "vj2", "")("stop1", 10*3600 + 10*60, 10*3600 + 10*60)("stop2", 10*3600 + 40*60,10*3600 + 40*60)("stop3", 10*3600 + 50*60,10*3600 + 50*60);
+    b.vj("A", "110011000001", "", true, "vj1", "")
+        ("stop1", "10:00"_t, "10:00"_t)
```
pbougue, line 66: :+1:

source/time_tables/tests/departure_boards_test.cpp
pbougue: That's a weird case to test indeed (but it may break other things, so nice to have it)

source/time_tables/departure_boards.cpp
benoit-bst: I think you are talking about **DateTime**, but **DateTime** is not a structure just a `typedef = uint32_t`

source/time_tables/tests/departure_boards_test.cpp
```diff
@@ -207,6 +212,172 @@ BOOST_AUTO_TEST_CASE(departureboard_test1) {
...
+    ed::builder b("20150615");
+    b.vj("A", "11111111111111", "", true, "vj1", "")
+        ("stop1", "06:00"_t, "06:00"_t)
+        ("stop1", "07:00"_t, "07:00"_t)
+        ("stop1", "10:00"_t, "10:00"_t)
+        ("stop2", "10:30"_t, "10:30"_t);
+
+    const auto it1 = b.sas.find("stop2");
+    b.data->pt_data->routes.front()->destination= it1->second; // Route A
+    b.finish();
```
pbougue, line 225: same minor: can use make()

source/time_tables/departure_boards.cpp
benoit-bst: Nice catch ! I'll change that

source/time_tables/tests/departure_boards_test.cpp
```diff
@@ -207,6 +212,172 @@ BOOST_AUTO_TEST_CASE(departureboard_test1) {
...
+        BOOST_REQUIRE_EQUAL(resp.stop_schedules(0).has_last_datetime(), true);
+        BOOST_REQUIRE_EQUAL(resp.stop_schedules(0).first_datetime().time(), "06:00"_t);
+        BOOST_REQUIRE_EQUAL(resp.stop_schedules(0).first_datetime().base_date_time(), "20150616T060000"_pts);
+        BOOST_REQUIRE_EQUAL(resp.stop_schedules(0).last_datetime().time(), "10:00"_t);
+        BOOST_REQUIRE_EQUAL(resp.stop_schedules(0).last_datetime().base_date_time(), "20150616T100000"_pts);
+
+        // Case 4 :
+        //
+        // Note
+        // - opening date time is equal to the next stop time.
```
pbougue: :+1:

source/time_tables/departure_boards.h
benoit-bst: pt = posix::time but in this header file pt is not declared.

As said @kinnou02  https://github.com/CanalTP/navitia/pull/2509#discussion_r214627329

source/time_tables/tests/departure_boards_test.cpp
```diff
@@ -207,6 +212,172 @@ BOOST_AUTO_TEST_CASE(departureboard_test1) {
...
+        BOOST_REQUIRE_EQUAL(resp.stop_schedules(0).has_last_datetime(), true);
+        BOOST_REQUIRE_EQUAL(resp.stop_schedules(0).first_datetime().time(), "06:00"_t);
+        BOOST_REQUIRE_EQUAL(resp.stop_schedules(0).first_datetime().base_date_time(), "20150615T060000"_pts);
+        BOOST_REQUIRE_EQUAL(resp.stop_schedules(0).last_datetime().time(), "10:00"_t);
+        BOOST_REQUIRE_EQUAL(resp.stop_schedules(0).last_datetime().base_date_time(), "20150615T100000"_pts);
+
+        // Case 3 :
+        //
+        // Note
+        // - Request date time is on the service hole.
```
pbougue, line 319: Can you add a test on hole, but before closing date time, like `request date time : 11:00:00` ?

source/jormungandr/tests/departure_board_tests.py
```diff
@@ -1104,3 +1103,559 @@ def test_routes_schedule_with_invalid_forbidden_uri(self):
...
+        stop_schedules = response['stop_schedules']
+        assert len(stop_schedules) == 2
+        assert stop_schedules[0]['first_datetime']['date_time'] == "20170103T070000"
+        assert stop_schedules[0]['last_datetime']['date_time'] == "20170103T090000"
+        # since it's the last stop_point, we cannot get on the bus....
+        assert 'first_datetime' not in stop_schedules[1]
+        assert 'last_datetime' not in stop_schedules[1]
+
+        """              20170102
+                       from_datetime 20170103T0001
```
pbougue, line 1179: Could be nice to have that test also 1h earlier (same on HK)

source/time_tables/departure_boards.cpp
xlqian: i have a test for `nb_days == 0`: https://github.com/CanalTP/navitia/pull/2509/files#diff-b4158732da04c6a27c4538fc599e1ce1R1577
for `nb_day == 365` since every datetime is relative to the beginning of the production, I think it's ok

source/time_tables/departure_boards.cpp
xlqian: we don't have `anticlockwise` in `/stop_schedule`, we always looked for  stop_times with `PICK_UP` event  with `clockwise`

source/jormungandr/tests/departure_board_tests.py
```diff
@@ -1104,3 +1103,559 @@ def test_routes_schedule_with_invalid_forbidden_uri(self):
...
+        assert stop_schedules[0]['last_datetime']['date_time'] == "20170104T090500"
+        # since it's the last stop_point, we cannot get on the bus....
+        assert 'first_datetime' not in stop_schedules[1]
+        assert 'last_datetime' not in stop_schedules[1]
+
+    def test_first_last_datetime_another_validity(self):
+        if not app.config['USE_SERPY']:
+            return
+        """              20170104
+                           from_datetime 20170105T0030
```
pbougue, line 1342: Can you test it on 20170105T2230 and check we get 15-17 on the 6th? (same on HK)

source/time_tables/departure_boards.h
xlqian: utc_offset may be negative

benoit-bst: Moreover, we handle **DateTime** in this function, so it is preferable to have offset as an **int**.

source/jormungandr/tests/departure_board_tests.py
```diff
@@ -1104,3 +1103,559 @@ def test_routes_schedule_with_invalid_forbidden_uri(self):
...
+
+        """
+        from_datetime = "20170102T000001"
+        response = self.query_region("stop_points/X_S3/stop_schedules?from_datetime={}"
+                                     "&data_freshness=base_schedule".format(from_datetime))
+
+        stop_schedules = response['stop_schedules']
+        assert stop_schedules
+        assert stop_schedules[0]['first_datetime']['date_time'] == "20170102T235000"
+        assert stop_schedules[0]['last_datetime']['date_time'] == "20170103T001000"
```
pbougue, line 1656: You could test on stop_schedules[1] also, no? You would have 20170102T064000 - 20170102T064500 I guess (same below).

pbougue (commented):
Tests looks good, only requested little more tests
and minor comments to be handled or not :)

OK to avoid handling DST, as long as it doesn't crash (not so important to have the right result here)

source/time_tables/departure_boards.h
benoit-bst: Yes it's true, closing is not used in this function.
The algorithm is only based on *opening time*. 
We just change the orientation of the request (clokwise or not), and adding a simple logic to change the day of the request.
That's why we don't have closing time parameters. This is not clear for you when you read the code ?

source/time_tables/departure_boards.cpp
benoit-bst: This is in contradiction with the other comment (https://github.com/CanalTP/navitia/pull/2509#discussion_r214945688) to split in half, i think.
But find an other function name, and add few comments, why not :+1:

source/time_tables/departure_boards.cpp
benoit-bst: why not

source/time_tables/tests/departure_boards_test.cpp
benoit-bst: **Closing time** is useless for the algorithm.

Maybe we have an other option to be less confused. 
- Suppress the closing date time in all tests. What do you think of that @pbougue ?

source/time_tables/departure_boards.h
pbougue: The point from kinnou is to avoid aliases in header, not to avoid using the type, in my understanding.
So to me it would be nicer to have a posix::time::time_duration.

source/time_tables/departure_boards.h
pbougue: The whole logic was not clear to me (cf. what I said on renaming the method).
The logic of not using closing is "external" to the code and it took me some time to realise it makes sense :sweat_drops:

source/time_tables/departure_boards.h
pbougue: From what I read in boost doc, "A duration can be either positive or negative."
And as we add this offset to a DateTime it makes more sense to me to have a Duration, no?
But I'm all for hard-typing, :crab: style :smile:

source/time_tables/departure_boards.cpp
pbougue: Here, it's an anticlockwise call: https://github.com/CanalTP/navitia/pull/2509#discussion-diff-215946896R383 no?

source/time_tables/departure_boards.cpp
pbougue: Yep, but it really makes more sense when you read the code (especially with the count below that is an int), and if we ever change the type (OK, never gonna happen :smile: )

source/time_tables/departure_boards.h
benoit-bst: Okay, so some comments above of the function is necessary

source/time_tables/departure_boards.h
benoit-bst: I'll try to change that if you want for more reading confort.

source/time_tables/tests/departure_boards_test.cpp
pbougue: That's fine to remove it, yes.
As for the test itself, I also wanted to challenge the shift of day (first and last being the day after the requested one).

kinnou02 (approved):

source/jormungandr/tests/departure_board_tests.py
xlqian: done

source/jormungandr/tests/departure_board_tests.py
xlqian: hmmm, the case you're talking about (switch of validity pattern between two days) is tested here: https://github.com/CanalTP/navitia/pull/2509/files/2191f7fe7d417780366e4918b7b0d7153862553e#diff-b4158732da04c6a27c4538fc599e1ce1R1209 

i will add it anyway :)

source/jormungandr/tests/departure_board_tests.py
xlqian: done

benoit-bst:
Okay @pbougue, I took your remarks into account. 
I hope it's good for you now.

pbougue (approved):

PR #2512, opened 2018-08-28
Jormun: Fix missing tickets

xlqian:
https://jira.kisio.org/browse/NAVITIAII-2549

more tests will be added into Artemis

### DIFF
--- a/source/jormungandr/jormungandr/scenarios/helper_classes/helper_utils.py
+++ b/source/jormungandr/jormungandr/scenarios/helper_classes/helper_utils.py
@@ -30,14 +30,12 @@
 from __future__ import absolute_import
 import math
 from jormungandr.street_network.street_network import StreetNetworkPathType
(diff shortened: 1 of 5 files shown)


source/jormungandr/jormungandr/scenarios/helper_classes/helper_utils.py
```diff
@@ -37,6 +37,7 @@
 import copy
 import logging
 import six
+import shortuuid
```
nonifier: maybe the dependence to `uuid` could be removed. Even from the pip requirements if that's used anywhere else.

nonifier (approved):

TeXitoi (approved):

xlqian:
In this PR, shortuuid is used for all internal ids (section_id, ticket_id...)

kadhikari (approved):

PR #2513, opened 2018-08-28
Doc: various fixes

pbougue:
minor corrections, mostly on doc (code and slate)

### DIFF
--- a/documentation/rfc/autocomplete.md
+++ b/documentation/rfc/autocomplete.md
@@ -20,7 +20,7 @@ Every result has four properties used for the "scoring":
         terms' query should have a score of 30. (In reality it's more complicated than that).
   - the *global score*: computed at the "binarisation" is the *importance* of a object, each type has it's own formula:
     - admin: based on the number of *ways* of the admin or the number of *stop_points* associated to it, if it doesn't have any way. This score is normalized with the formula: `score = log(score+2)*10`.
-    - stop_area: the number of *stop_points* associated to the stop_area, normalized as a integer between 0 and 100,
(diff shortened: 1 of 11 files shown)


TeXitoi (approved):

documentation/rfc/autocomplete.md
```diff
@@ -55,6 +55,6 @@ and finally by the quality.
 If a request returns a lot of admins, the response will be filled with the ten most important admins 
 and won't have any stop_area or way.
 
-Most of the times the complain we got on the autocompletion is that a stop_area can't be found.
+Most of the times the complains we got on the autocompletion is that a stop_area can't be found.
```
nonifier: Either: 'Most of the time~s~' (because uncountable) or simply `Most of the complains we've got are [...]` (you need to use present perfect as there is no time reference)

source/georef/georef.cpp
```diff
@@ -615,7 +615,7 @@ std::vector<Admin*> GeoRef::find_admins(const type::GeographicalCoord& coord, Ad
     if(!result.empty()){
         return result;
     }
-    //we didn't found any results with boundary, as a fallback we search for the admin of the closest way
+    //we didn't find any result with boundary, as a fallback we search for the admin of the closest way
```
nonifier: we didn't find any result with**in the** boundary, as a fallback we search for the **nearest** admin.

nonifier (changes requested):

nonifier (approved):

TeXitoi (approved):

PR #2516, opened 2018-08-30
jormun: do not return a 500 on bragi timeout for journeys

kinnou02:


source/jormungandr/jormungandr/autocomplete/abstract_autocomplete.py
```diff
@@ -78,6 +78,8 @@ def get_object_by_uri(self, uri, instances=None, current_datetime=None):
             details = self.get_by_uri(uri, instances=instances, current_datetime=current_datetime)
         except TechnicalError as e:
             log_exception(sender=None, exception=e)
+        except RuntimeError as e:
```
xlqian, line 81: very minor 

```python
except (TechnicalError, RuntimeError) as e:
```

change it or not, as you wish

xlqian (approved):

source/jormungandr/jormungandr/autocomplete/abstract_autocomplete.py
kinnou02: I don't really like this synthax, I found it harder to read, but if everyone think it's better I can change it

PR #2524, opened 2018-09-06
Jormun: improve cache in realtimeproxy

kinnou02:
We want to be able to cache some realtime proxy more than 60 second, so
we need to round more than just the seconds of the datetime.
Every datetime is rounded from midnight of the day to reduce surprises.


pbougue (approved):

TeXitoi:
I want tests with timezone!

kinnou02:
We use "naive" datetime everywhere in jormun, so timezone doesn't have impact

nonifier (approved):

xlqian (approved):

benoit-bst (approved):

PR #2525, opened 2018-09-06
Kraken: Do not display any disruption if parameter disable_disruption used

kadhikari:
Ticket: https://jira.kisio.org/browse/NAVP-853

### DIFF
--- a/source/jormungandr/tests/ptref_tests.py
+++ b/source/jormungandr/tests/ptref_tests.py
@@ -642,6 +642,55 @@ class TestPtRef(AbstractTestFixture):
 
         assert len([r for r in i.ptref.get_objs(type_pb2.ROUTE)]) == 3
 
+    def test_ptref_on_stop_areas_with_disable_disruption(self):
+        """
+        stop_area:stop1 with and without disable_disruption
+        """
+        utc_datetime = '20140115T235959'
+        response = self.query_region("stop_areas/stop_area:stop1?_current_datetime={}".format(utc_datetime))
+
+        disruptions = get_not_null(response, 'disruptions')
+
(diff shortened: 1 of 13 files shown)


source/jormungandr/tests/ptref_tests.py
```diff
@@ -642,6 +642,55 @@ def test_pt_ref_internal_method(self):
 
         assert len([r for r in i.ptref.get_objs(type_pb2.ROUTE)]) == 3
 
+    def test_ptref_on_stop_areas_with_disable_disruption(self):
+        """
+        stop_area:stop1 with and without disable_disruption
+        """
+        utc_datetime = '20140115T235959'
```
xlqian: (you can add "Z" at the end of datetime to make it UTC )

xlqian (approved):

source/type/pb_converter.cpp
```diff
@@ -1017,7 +1017,9 @@ void PbCreator::Filler::fill_message(const boost::shared_ptr<nd::Impact>& impact
     if (none_of(pb_object->impact_uris(), [&](const std::string& uri) { return impact->uri == uri; })) {
         *pb_object->add_impact_uris() = impact->uri;
     }
-    pb_creator.impacts.insert(impact);
+    if (!pb_creator.disable_disruption) {
```
kinnou02: It's better to shortcircuit at the start of the function.
In line 1018 we add the uri of the impact to the object, with this change we do not put the disruption be still have the link to it.

kinnou02 (changes requested):

kinnou02 (approved):

PR #2532, opened 2018-09-11
Jormun: refacto initialization to add newrelic and gevent

kinnou02:
This moves some of the crap in __init__ to functions in init.py
It also gives us the ability to initialize newrelic directly from jormungandr.
Finally, we are now monkeypatching all sockets with gevent and not just request, this shouldn't have any negative performance impact as jormungandr is now executed with only one thread, so every request has a dedicated process and do not have to fight with another request for some CPU time inside a python process

### DIFF
--- a/source/jormungandr/jormungandr/init.py
+++ b/source/jormungandr/jormungandr/init.py
@@ -1,0 +1,62 @@
+#!/usr/bin/env python
+# coding=utf-8
+
+# Copyright (c) 2001-2014, Canal TP and/or its affiliates. All rights reserved.
+#
+# This file is part of Navitia,
+#     the software to build cool stuff with public transport.
+#
+# Hope you'll enjoy and contribute to this project,
+#     powered by Canal TP (www.canaltp.fr).
+# Help us simplify mobility and open public transport:
+#     a non ending quest to the responsive locomotion way of traveling!
+#
+# LICENCE: This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU Affero General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+# GNU Affero General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with this program. If not, see <http://www.gnu.org/licenses/>.
+#
+# Stay tuned using
+# twitter @navitia
+# IRC #navitia on freenode
+# https://groups.google.com/d/forum/navitia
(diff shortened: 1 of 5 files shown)


source/jormungandr/jormungandr/new_relic.py
```diff
@@ -39,6 +40,16 @@
     agent = None
 
 
+def init(config):
+    if not agent or not config:
+        return
+    if os.path.exists(config):
+        agent.initialize(config)
+    else:
+        logging.error('%s doesn\'t exist, newrelic disabled', config)
```
nonifier: you might want use the logger from `logger = logging.getLogger(__name__)` instead ?

source/jormungandr/jormungandr/init.py
```diff
@@ -0,0 +1,62 @@
...
+        logging.config.dictConfig(app.config['LOGGER'])
+    else:  # Default is std out
+        handler = logging.StreamHandler(stream=sys.stdout)
+        app.logger.addHandler(handler)
+        app.logger.setLevel('INFO')
+
+def patch_http():
+    logger = logging.getLogger(__name__)
+    logger.info(
+        "Attention! You'are patching socket with gevent parallel http/https calling by requests is activated"
```
nonifier: :fr: `Attention` >> :gb: `Warning`

nonifier (approved):

kinnou02:
Jenkins, test this please.

source/jormungandr/jormungandr/init.py
```diff
@@ -0,0 +1,62 @@
...
+
+def patch_http():
+    logger = logging.getLogger(__name__)
+    logger.info(
+        "Warning! You'are patching socket with gevent, parallel http/https calling by requests is activated"
+    )
+
+    from gevent import monkey
+    monkey.patch_ssl()
+    monkey.patch_socket()
```
xlqian: this function has caused mysterious performance regression when calling redis.

prefer using: 
```python
    import requests
    import gevent.socket
    requests.packages.urllib3.connection.connection.socket = gevent.socket
```

xlqian:
" so every request has a decicated processus" how:question:  i'am curious about this...

xlqian:
when hanging around on the stackoverflow, i found this: https://github.com/kennethreitz/grequests or https://github.com/ross/requests-futures

TeXitoi (approved):

kinnou02:
> " so every request has a decicated processus" how i'am curious about this...

https://github.com/CanalTP/navitia_deployment_conf/blob/master/prod.py#L76-L77

grequest and request-futures are interesting, but doesn't really solve the same thing, they will require some work to be used in place of request as they do not have the same interface.

kinnou02:
Jenkins, test this please.

xlqian (approved):

kinnou02:
Jenkins, test this please.

kinnou02:
Jenkins, test this please.

PR #2536, opened 2018-09-17
doc: tell mkvirtualenv to use python 2

nonifier:
By default, `mkvirtualenv` now uses python 3 when setting up an environment. It doesn't work so well with Jormun....


pbougue (approved):

antoine-de:
you should move to [pipenv](https://pipenv.readthedocs.io/en/latest/) it's way better than mkvirtualenv (well it won't change the need to define the python version  :smile:

benoit-bst (approved):

mbounabi (approved):

PR #2537, opened 2018-09-18
tests: rewrite line section tests in order to check all disruptions

vpassama:
Hi,

This is a small PR I would like to make in order to add all missing checks in existing functions of line_sections_test in jormungandr.
I also rewrited some details for pep8 compliance and I used black tool in order to be sure that the code style is OK.

The final goal behind that is in a second PR I'm about to make where I remove DumpLineSectionMessage (yes...) So I need clearer tests in order to see all the changes and discuss them with you.

Thank you.

### DIFF
--- a/source/jormungandr/tests/disruptions_tests.py
+++ b/source/jormungandr/tests/disruptions_tests.py
@@ -29,7 +29,17 @@
 
 from __future__ import absolute_import, print_function, unicode_literals, division
 from .tests_mechanism import AbstractTestFixture, dataset
-from .check_utils import *
+from .check_utils import (
+    get_not_null,
+    get_all_element_disruptions,
+    get_disruptions,
+    is_valid_line,
+    is_valid_disruption,
+    is_valid_network,
+    is_valid_stop_area,
+    is_valid_line_report,
+    s_coord,
+)
 import jmespath
 
 
@@ -676,9 +686,9 @@ class TestDisruptionsLineSections(AbstractTestFixture):
     def test_line_reports(self):
         response = self.query_region("line_reports?_current_datetime=20170103T120000")
         disruptions = get_not_null(response, 'disruptions')
-        assert len(disruptions) == 2
+        assert len(disruptions) == 3
         line_reports = get_not_null(response, 'line_reports')
-        assert len(line_reports) == 1
+        assert len(line_reports) == 2
         is_valid_line_report(line_reports[0])
         assert line_reports[0]['line']['id'] == 'line:1'
         assert len(line_reports[0]['pt_objects']) == 5
@@ -689,15 +699,18 @@ class TestDisruptionsLineSections(AbstractTestFixture):
         assert line_reports[0]['pt_objects'][4]['id'] == 'F_1'
         for pt_object in line_reports[0]['pt_objects']:
             assert pt_object['embedded_type'] == 'stop_point'
-            assert len(pt_object['stop_point']['links']) == (2 if pt_object['id'] == 'E_1' else 1)
+            assert len(pt_object['stop_point']['links']) == (
+                2 if pt_object['id'] in ['A_1', 'E_1', 'F_1'] else 1
+            )
             if pt_object['id'] != 'F_1':
                 assert pt_object['stop_point']['links'][0]['id'] == 'line_section_on_line_1'
         assert (
             line_reports[0]['pt_objects'][3]['stop_point']['links'][1]['id']
             == 'line_section_on_line_1_other_effect'
         )
+        assert line_reports[0]['pt_objects'][4]['stop_point']['links'][0]['id'] == 'line_section_on_line_2'
         assert (
-            line_reports[0]['pt_objects'][4]['stop_point']['links'][0]['id']
+            line_reports[0]['pt_objects'][4]['stop_point']['links'][1]['id']
             == 'line_section_on_line_1_other_effect'
         )
 
@@ -712,9 +725,9 @@ class TestDisruptionsLineSections(AbstractTestFixture):
             "line_reports?_current_datetime=20170101T120000" "&since=20170104T130000&until=20170106T000000"
         )
         disruptions = get_not_null(response, 'disruptions')
-        assert len(disruptions) == 2
+        assert len(disruptions) == 3
         line_reports = get_not_null(response, 'line_reports')
-        assert len(line_reports) == 1
+        assert len(line_reports) == 2
         is_valid_line_report(line_reports[0])
         assert line_reports[0]['line']['id'] == 'line:1'
         assert len(line_reports[0]['pt_objects']) == 5
@@ -725,24 +738,27 @@ class TestDisruptionsLineSections(AbstractTestFixture):
         assert line_reports[0]['pt_objects'][4]['id'] == 'F_1'
         for pt_object in line_reports[0]['pt_objects']:
             assert pt_object['embedded_type'] == 'stop_point'
-            assert len(pt_object['stop_point']['links']) == (2 if pt_object['id'] == 'E_1' else 1)
+            assert len(pt_object['stop_point']['links']) == (
+                2 if pt_object['id'] in ['A_1', 'E_1', 'F_1'] else 1
+            )
             if pt_object['id'] != 'F_1':
                 assert pt_object['stop_point']['links'][0]['id'] == 'line_section_on_line_1'
         assert (
             line_reports[0]['pt_objects'][3]['stop_point']['links'][1]['id']
             == 'line_section_on_line_1_other_effect'
         )
+        assert line_reports[0]['pt_objects'][4]['stop_point']['links'][0]['id'] == 'line_section_on_line_2'
         assert (
-            line_reports[0]['pt_objects'][4]['stop_point']['links'][0]['id']
+            line_reports[0]['pt_objects'][4]['stop_point']['links'][1]['id']
             == 'line_section_on_line_1_other_effect'
         )
 
     def test_line_reports_with_since_intersects_application_period(self):
         response = self.query_region("line_reports?_current_datetime=20170101T120000&since=20170104T130000")
         disruptions = get_not_null(response, 'disruptions')
-        assert len(disruptions) == 2
+        assert len(disruptions) == 3
         line_reports = get_not_null(response, 'line_reports')
-        assert len(line_reports) == 1
+        assert len(line_reports) == 2
         is_valid_line_report(line_reports[0])
         assert line_reports[0]['line']['id'] == 'line:1'
         assert len(line_reports[0]['pt_objects']) == 5
@@ -753,15 +769,18 @@ class TestDisruptionsLineSections(AbstractTestFixture):
         assert line_reports[0]['pt_objects'][4]['id'] == 'F_1'
         for pt_object in line_reports[0]['pt_objects']:
             assert pt_object['embedded_type'] == 'stop_point'
-            assert len(pt_object['stop_point']['links']) == (2 if pt_object['id'] == 'E_1' else 1)
+            assert len(pt_object['stop_point']['links']) == (
+                2 if pt_object['id'] in ['A_1', 'E_1', 'F_1'] else 1
+            )
             if pt_object['id'] != 'F_1':
                 assert pt_object['stop_point']['links'][0]['id'] == 'line_section_on_line_1'
         assert (
             line_reports[0]['pt_objects'][3]['stop_point']['links'][1]['id']
             == 'line_section_on_line_1_other_effect'
         )
+        assert line_reports[0]['pt_objects'][4]['stop_point']['links'][0]['id'] == 'line_section_on_line_2'
         assert (
-            line_reports[0]['pt_objects'][4]['stop_point']['links'][0]['id']
+            line_reports[0]['pt_objects'][4]['stop_point']['links'][1]['id']
             == 'line_section_on_line_1_other_effect'
         )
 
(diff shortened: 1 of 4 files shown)


buildbotctp:

pbougue:
OK to test.

source/jormungandr/tests/line_sections_tests.py
```diff
@@ -312,11 +518,12 @@ def test_journeys_use_vj_impacted_by_line_section(self):
...
         period of the impact
         """
-        date = '20170102T080000'  # with this departure time we should take 'vj:1:1' that is impacted
+        # with this departure time we should take 'vj:1:1' that is impacted
+        date = '20170102T080000'
         current = '_current_datetime=20170101T080000'
 
         def journey_query(_from, to):
-            return 'journeys?from={fr}&to={to}&datetime={dt}&{cur}&data_freshness=base_schedule'.format(
+            return 'journeys?from={fr}&to={to}&datetime={dt}&{cur}' '&data_freshness=base_schedule'.format(
```
vpassama: I just saw this, I think black rewrited 2 lines in one and made this :3 I would like to change this before a merge.

source/jormungandr/tests/line_sections_tests.py
```diff
@@ -373,11 +628,12 @@ def test_journeys_use_vj_not_impacted_by_line_section(self):
...
         impacted route) during the application period of the impact
         """
-        date = '20170102T090000'  # with this departure time we should take 'vj:1:2' that is impacted
+        # with this departure time we should take 'vj:1:2' that is impacted
+        date = '20170102T090000'
         current = '_current_datetime=20170101T080000'
 
         def journey_query(_from, to):
-            return 'journeys?from={fr}&to={to}&datetime={dt}&{cur}&data_freshness=base_schedule'.format(
+            return 'journeys?from={fr}&to={to}&datetime={dt}&{cur}' '&data_freshness=base_schedule'.format(
```
vpassama: same here.

vpassama (commented):
2 mistakes I made with black tool I guess.

source/jormungandr/tests/disruptions_tests.py
```diff
@@ -725,24 +738,27 @@ def test_line_reports_with_since_until_intersects_application_period(self):
...
+            assert len(pt_object['stop_point']['links']) == (
+                2 if pt_object['id'] in ['A_1', 'E_1', 'F_1'] else 1
+            )
             if pt_object['id'] != 'F_1':
                 assert pt_object['stop_point']['links'][0]['id'] == 'line_section_on_line_1'
         assert (
             line_reports[0]['pt_objects'][3]['stop_point']['links'][1]['id']
             == 'line_section_on_line_1_other_effect'
         )
+        assert line_reports[0]['pt_objects'][4]['stop_point']['links'][0]['id'] == 'line_section_on_line_2'
```
pbougue: It may be better (to avoid later maintaining this) to write it like (untested):
```python
links = line_reports[0]['pt_objects'][4]['stop_point']['links']
assert 'line_section_on_line_2' in [ l['id'] for l in links ]
```
(same for other tests on links in this PR)

source/jormungandr/tests/line_sections_tests.py
```diff
@@ -50,28 +55,31 @@ def get_id(obj):
             id = obj.impacted_object['links'][0]['id']
         return id
 
-    return set(get_id(o) for o in disruptions['line_section_on_line_1'])
+    ids = set()
+    for label, d in disruptions.iteritems():
```
pbougue: So you want all the ids of all disruptions if I understand correctly.
Also, you don't use `label`, clearer to change it to `for _, d in disruptions.iteritems():`

source/jormungandr/tests/line_sections_tests.py
```diff
@@ -165,90 +248,181 @@ def test_traffic_reports_on_stop_areas(self):
...
-            assert len(sa['links']) == (2 if sa['id'] == 'E' else 1)
+            assert len(sa['links']) == (2 if sa['id'] in ['A', 'E', 'F'] else 1)
 
     def test_traffic_reports_on_lines(self):
         """
         we should be able to find the related line section disruption with /traffic_report
         """
 
+        # The commented results are not working for now, in those cases, the disruption affects at least 1 stop
+        # of the line whereas the line itself is not concerned (the disrupted routes are linked to other lines)
```
pbougue: :see_no_evil: good to note it, may be worth an issue if we want to fix it.

source/jormungandr/tests/line_sections_tests.py
```diff
@@ -165,90 +248,181 @@ def test_traffic_reports_on_stop_areas(self):
...
-            return 'line_section_on_line_1' in (d['disruption_id'] for d in r['disruptions'])
+        # line_section_on_line_1
+        # route 3 has been impacted by the line section but it has no stoppoint
+        # in the line section so in this case we do not display the disruption
+        scenario = {
+            'route:line:1:1': True,
+            'route:line:1:2': False,
+            'route:line:1:3': False,
+            'route:line:2:1': False,
+            'route:line:3:1': True,
```
pbougue: Shouldn't it be False here ? (commented if we follow the what's already done elsewhere)

source/jormungandr/tests/line_sections_tests.py
```diff
@@ -165,90 +248,181 @@ def test_traffic_reports_on_stop_areas(self):
...
+
+        for route, result in scenario.iteritems():
+            assert result == self.has_dis('routes/{}/traffic_reports'.format(route), 'line_section_on_line_1')
+
+        # line_section_on_line_1_other_effect
+        scenario = {
+            'route:line:1:1': True,
+            'route:line:1:2': False,
+            'route:line:1:3': True,
+            'route:line:2:1': True,
```
pbougue: Shouldn't it be False here ? (commented if we follow the what's already done elsewhere)

source/jormungandr/tests/line_sections_tests.py
```diff
@@ -165,90 +248,181 @@ def test_traffic_reports_on_stop_areas(self):
...
+        for route, result in scenario.iteritems():
+            assert result == self.has_dis(
+                'routes/{}/traffic_reports'.format(route), 'line_section_on_line_1_other_effect'
+            )
+
+        # line_section_on_line_2
+        scenario = {
+            'route:line:1:1': True,
+            'route:line:1:2': False,
+            'route:line:1:3': True,
```
pbougue: Shouldn't it be False here ? (commented if we follow the what's already done elsewhere)

source/jormungandr/tests/line_sections_tests.py
```diff
@@ -165,90 +248,181 @@ def test_traffic_reports_on_stop_areas(self):
...
+        }
+
+        for route, result in scenario.iteritems():
+            assert result == self.has_dis(
+                'routes/{}/traffic_reports'.format(route), 'line_section_on_line_1_other_effect'
+            )
+
+        # line_section_on_line_2
+        scenario = {
+            'route:line:1:1': True,
```
pbougue: Shouldn't it be False here ? (commented if we follow the what's already done elsewhere)

source/jormungandr/tests/line_sections_tests.py
```diff
@@ -165,90 +248,181 @@ def test_traffic_reports_on_stop_areas(self):
...
+        for route, result in scenario.iteritems():
+            assert result == self.has_dis('routes/{}/traffic_reports'.format(route), 'line_section_on_line_2')
 
     def test_traffic_reports_on_vjs(self):
         """
         for /traffic_reports on vjs it's a bit the same as the lines
         we display a line section disruption if it impacts the stops of the vj
         """
+        # line_section_on_line_1
+        scenario = {'vj:1:1': True, 'vj:1:2': False, 'vj:1:3': False, 'vj:2': False, 'vj:3': True}
```
pbougue: Shouldn't it be false for 'vj:3' ?

source/jormungandr/tests/line_sections_tests.py
```diff
@@ -165,90 +248,181 @@ def test_traffic_reports_on_stop_areas(self):
...
+
+        for vj, result in scenario.iteritems():
+            assert result == self.has_dis(
+                'vehicle_journeys/{}/traffic_reports'.format(vj), 'line_section_on_line_1'
+            )
+
+        # line_section_on_line_1_other_effect
+        scenario = {'vj:1:1': True, 'vj:1:2': False, 'vj:1:3': True, 'vj:2': True, 'vj:3': False}
```
pbougue: Shouldn't it be false for 'vj:2' ?

source/jormungandr/tests/line_sections_tests.py
```diff
@@ -165,90 +248,181 @@ def test_traffic_reports_on_stop_areas(self):
...
+        for vj, result in scenario.iteritems():
+            assert result == self.has_dis(
+                'vehicle_journeys/{}/traffic_reports'.format(vj), 'line_section_on_line_1_other_effect'
+            )
 
-        def has_dis(q):
-            r = self.default_query(q)
-            return 'line_section_on_line_1' in (d['disruption_id'] for d in r['disruptions'])
+        # line_section_on_line_2
+        scenario = {'vj:1:1': True, 'vj:1:2': False, 'vj:1:3': True, 'vj:2': True, 'vj:3': False}
```
pbougue: Shouldn't it be false for 'vj:1:1' and 'vj:1:3' ?

source/tests/line_sections_test.cpp
```diff
@@ -120,6 +120,14 @@ int main(int argc, const char* const argv[]) {
                               .get_disruption(),
                               *b.data->pt_data, *b.data->meta);
 
+    navitia::apply_disruption(b.impact(nt::RTLevel::Adapted, "line_section_on_line_2")
```
pbougue: minor: Can be moved below to group disruptions by line

source/jormungandr/tests/line_sections_tests.py
```diff
@@ -410,32 +703,80 @@ def test_stop_schedule_impacted_by_line_section(self):
...
+        r = self.query_region(query.format(sa='E', cur=cur, d=dt, f=fresh))
+        d = get_all_element_disruptions(r['stop_schedules'], r)
+        for disruption, result in scenario.iteritems():
+            assert result == (disruption in d)
+        assert impacted_ids(d) == {'vj:1:1'}
+
+        # F
+        scenario = {
+            'line_section_on_line_1': False,
+            'line_section_on_line_1_other_effect': False,
```
pbougue: OK, so it's False for `line_section_on_line_1_other_effect` and `line_section_on_line_2` just because stop_schedule is empty in concerned directions, here I guess.

source/jormungandr/tests/line_sections_tests.py
```diff
@@ -444,41 +785,138 @@ def test_departures_impacted_by_line_section(self):
         cur = '_current_datetime=20170101T100000'
         dt = 'from_datetime=20170101T080000'
         fresh = 'data_freshness=base_schedule'
-        r = self.query_region('stop_areas/A/departures?{cur}&{d}&{f}'.format(cur=cur, d=dt, f=fresh))
-        assert 'line_section_on_line_1' not in get_all_element_disruptions(r['departures'], r)
+        query = 'stop_areas/{sa}/departures?{cur}&{d}&{f}'
```
pbougue: You could merge scenario with stop_schedules, they should be identical.

pbougue (commented):
Great testing, covers a lot of cases!
* Most of the questions are regarding the fact that we may be enforcing a bad behavior here and there.
* For `minor` feel free to not care :wink: 
* Also, for review purpose, can you push new commits for your corrections (no push force of corrected commit), so that we know what changed and what did not? At the merge we can squash all commits in one if you want to (let us know).

kinnou02 (approved):

source/jormungandr/tests/line_sections_tests.py
vpassama: I'm not really sure if it should be False because of the note below the function : 

for routes since we display the impacts on all the stops (but we do not display a route object) we display the disruption even if the route has not been directly impacted

But I don't mind commenting the lines if you think it is not the right behavior anyway.

source/jormungandr/tests/line_sections_tests.py
vpassama: The note below the function says:

we display a line section disruption if it impacts the stops of the vj

So, like the route tests, I'm not sure for the vj:3 as the stops are impacted.

source/jormungandr/tests/line_sections_tests.py
pbougue: As discussed:
* the current behavior si OK for now (at least for Kisio) but would be improved if routes where managed.
* let's try to write tests for the targeted behavior (route-point being managed) but add a clear xfail for them
  * If it's heavy, let's switch back to targeted behavior but commented tests
  * If possible let's make those xfail strict so that we maintain them (like we should do for artemis I guess :sweat_smile:  )

source/jormungandr/tests/line_sections_tests.py
```diff
@@ -165,90 +249,211 @@ def test_traffic_reports_on_stop_areas(self):
...
 
         # line_section_on_line_1
-        assert self.has_tf_disruption('lines/line:1/traffic_reports', 'line_section_on_line_1')
-        assert self.has_tf_linked_disruption(
-            'lines/line:1/traffic_reports', 'line_section_on_line_1', ObjGetter('lines', 'line:1')
-        )
+        scenario = {
+            'line:1': True,
+            'line:2': False,
+            'line:3': False  # fail
```
pbougue: What I fear now is we won't realize line:1 or line:2 test failed as it's an xfail test as a whole.
But then it becomes more painfull to use xfail if that prevent testcode-mutualization...

pbougue (approved):
Looks great!

PR #2538, opened 2018-09-19
Manage bss configuration with Tyr

kinnou02:
This PR aims to manage bss connector's configuration in tyr to replace the static configuration in the `settings.py` of jormungandr.

This is the first step toward removing all "functionnal" configuration of jormungandr from fabric to reduce our dependency to the devops team.

It's probably easier to read this PR by commit

The last commit add the execution of tyr's tests with travis


source/navitiacommon/navitiacommon/models/bss_configuration.py
```diff
@@ -0,0 +1,70 @@
+# encoding: utf-8
+
+#  Copyright (c) 2001-2014, Canal TP and/or its affiliates. All rights reserved.
```
pbougue: minor: 2018

source/tyr/tests/integration/bss_configuration_test.py
```diff
@@ -0,0 +1,150 @@
+# coding: utf-8
+# Copyright (c) 2001-2018, Canal TP and/or its affiliates. All rights reserved.
+#
+# This file is part of Navitia,
+#     the software to build cool stuff with public transport.
+#
+#     powered by Canal TP (www.canaltp.fr).
```
pbougue, line 7: minor: missing line `# Hope you'll enjoy and contribute to this project,` (don't know why, it's super frequent)

source/navitiacommon/navitiacommon/models/bss_configuration.py
```diff
@@ -0,0 +1,70 @@
...
+
+from __future__ import absolute_import
+from sqlalchemy.dialects.postgresql.json import JSONB
+from navitiacommon.models import db, TimestampMixin
+
+
+class BssConfiguration(db.Model, TimestampMixin):
+    id = db.Column(db.Text, primary_key=True)
+    network = db.Column(db.Text, unique=False, nullable=False)
+    klass = db.Column(db.Text, unique=False, nullable=False)
```
pbougue: Curious about why this "klass" with a  `k`.

source/navitiacommon/navitiacommon/models/bss_configuration.py
```diff
@@ -0,0 +1,70 @@
...
+    timeout = db.Column(db.Float, nullable=False, default=2)
+    discarded = db.Column(db.Boolean, nullable=False, default=False)
```
pbougue: Why that `discarded`? To deactivate temporarily but keep the config? In that case 'deactivate' or 'disable' looks clearer to me.

pbougue (commented):
Did not read all, but on the general purpose I agree.
I see 2 potential issues here:
* Can we merge car-park providers with that? If so, naming should change. Maybe we can merge with some other things (ridesharing is another use case I guess).
* No versioning and no easy save/restore, also no right-management, so hard to track accesses and changes, a bit dangerous. But the way to do that I imagine is not super clean, like continuously tracking file changes from tyr, those files being updated from a config git repo.

source/navitiacommon/navitiacommon/models/bss_configuration.py
TeXitoi: `class` is a keyword

kinnou02: and `cls` isn't really reserved, but almost

source/navitiacommon/navitiacommon/models/bss_configuration.py
kinnou02: `discarded` is used for the instance, it's not pretty, but a bit consistent

source/navitiacommon/navitiacommon/models/bss_configuration.py
nonifier: `qulasse` maybe ? :fr: :D

source/navitiacommon/navitiacommon/models/bss_configuration.py
```diff
@@ -0,0 +1,70 @@
...
+from __future__ import absolute_import
+from sqlalchemy.dialects.postgresql.json import JSONB
+from navitiacommon.models import db, TimestampMixin
+
+
+class BssConfiguration(db.Model, TimestampMixin):
+    id = db.Column(db.Text, primary_key=True)
+    network = db.Column(db.Text, unique=False, nullable=False)
+    klass = db.Column(db.Text, unique=False, nullable=False)
+    timeout = db.Column(db.Float, nullable=False, default=2)
```
nonifier: `Float` is a bit misleading here as it obfuscates the type (second ?, ms? etc...). using a `db.Time` would resolve ambiguities.

kinnou02: Fair point, I will look if it's possible to do something without having to convert everything manually ten times

kinnou02:
> - Can we merge car-park providers with that? If so, naming should change. Maybe we can merge with some other things (ridesharing is another use case I guess).

I don't think we should, typing is a great thing

> - No versioning and no easy save/restore, also no right-management, so hard to track accesses and changes, a bit dangerous. But the way to do that I imagine is not super clean, like continuously tracking file changes from tyr, those files being updated from a config git repo.

I will probably create a small script that update tyr from some config stored in git, this way we keep the versioning.
As for the lack of right-management, if it needs to be done it's another subject that will impact the whole api of tyr.

kinnou02:
the naming in jormun is nice `bss_provider` I will probably rename thing in tyr something like this

nonifier (dismissed):

benoit-bst (dismissed):

nonifier (approved):

benoit-bst (approved):

kinnou02:
Do not merge this just before the release.

PR #2539, opened 2018-09-19
CMake: Clean for cmake prefix install

benoit-bst:
**Topic** : Clean the CMake:

- Using _CMAKE_INSTALL_PREFIX_ (_1st commit_)
   There were a lot of deprecated CMake functions used (install_targets, install_files).
   These old functions do not work with the _CMAKE_INSTALL_PREFIX_

- Remove compatibility for Debian 7 because it is now deprecated for navitia (_2nd commit_)
   Prometheus stub is now useless (This simplifies the code).

I tested the package building on:
- `Debian 8 (OK)`
- `Ubuntu 16.4 (OK)`

**Extra** : I am sad because Gcc version on debian 8 is 4.9.2... Must wait debian 9 for C++14 full support within Gcc.

### DIFF
--- a/source/CMakeLists.txt
+++ b/source/CMakeLists.txt
@@ -24,7 +24,7 @@ project(Kraken)
 
 # List of directories to search for CMake modules
 set(CMAKE_MODULE_PATH "${CMAKE_SOURCE_DIR}/cmake_modules")
-set(CMAKE_INSTALL_PREFIX "/usr/bin/")
+set(CMAKE_INSTALL_PREFIX "/usr")
 
 # project version,
 # get git revision as the version number
--- a/source/CMakeLists.txt
+++ b/source/CMakeLists.txt
@@ -24,6 +24,7 @@ project(Kraken)
 
 # List of directories to search for CMake modules
 set(CMAKE_MODULE_PATH "${CMAKE_SOURCE_DIR}/cmake_modules")
+set(CMAKE_INSTALL_PREFIX "/usr/bin/")
 
 # project version,
 # get git revision as the version number
--- a/source/cities/CMakeLists.txt
+++ b/source/cities/CMakeLists.txt
@@ -4,12 +4,11 @@ SET(BOOST_LIBS ${Boost_PROGRAM_OPTIONS_LIBRARY} ${Boost_FILESYSTEM_LIBRARY}
 add_executable(cities cities.cpp)
 target_link_libraries(cities transportation_data_import connectors types ${PQXX_LIB} ${OSMPBF} pb_lib utils
     ${BOOST_LIBS} log4cplus z protobuf)
-INSTALL_TARGETS(/usr/bin/ cities)
-
-INSTALL_FILES(/usr/share/navitia/cities/alembic FILES alembic.ini alembic/env.py)
+install(TARGETS cities DESTINATION ${CMAKE_INSTALL_PREFIX})
+install(FILES alembic.ini alembic/env.py DESTINATION /usr/share/navitia/cities/alembic)
 
 FILE(GLOB
     alembic_cities
     alembic/versions/*.py
     )
-INSTALL_FILES(/usr/share/navitia/cities/alembic/versions FILES ${alembic_cities})
+install(FILES ${alembic_cities} DESTINATION /usr/share/navitia/cities/alembic/versions)
--- a/source/cities/CMakeLists.txt
+++ b/source/cities/CMakeLists.txt
@@ -4,11 +4,11 @@ SET(BOOST_LIBS ${Boost_PROGRAM_OPTIONS_LIBRARY} ${Boost_FILESYSTEM_LIBRARY}
 add_executable(cities cities.cpp)
 target_link_libraries(cities transportation_data_import connectors types ${PQXX_LIB} ${OSMPBF} pb_lib utils
     ${BOOST_LIBS} log4cplus z protobuf)
-install(TARGETS cities DESTINATION ${CMAKE_INSTALL_PREFIX})
-install(FILES alembic.ini alembic/env.py DESTINATION /usr/share/navitia/cities/alembic)
+install(TARGETS cities DESTINATION ${CMAKE_INSTALL_PREFIX}/bin)
+install(FILES alembic.ini alembic/env.py DESTINATION ${CMAKE_INSTALL_PREFIX}/share/navitia/cities/alembic)
 
 FILE(GLOB
     alembic_cities
     alembic/versions/*.py
     )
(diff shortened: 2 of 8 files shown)


source/CMakeLists.txt
```diff
@@ -24,6 +24,7 @@ project(Kraken)
 
 # List of directories to search for CMake modules
 set(CMAKE_MODULE_PATH "${CMAKE_SOURCE_DIR}/cmake_modules")
+set(CMAKE_INSTALL_PREFIX "/usr/bin/")
```
TeXitoi: `INSTALL_PREFIX` should be `/usr/local`

source/cities/CMakeLists.txt
```diff
@@ -4,12 +4,11 @@ SET(BOOST_LIBS ${Boost_PROGRAM_OPTIONS_LIBRARY} ${Boost_FILESYSTEM_LIBRARY}
 add_executable(cities cities.cpp)
 target_link_libraries(cities transportation_data_import connectors types ${PQXX_LIB} ${OSMPBF} pb_lib utils
     ${BOOST_LIBS} log4cplus z protobuf)
-INSTALL_TARGETS(/usr/bin/ cities)
-
-INSTALL_FILES(/usr/share/navitia/cities/alembic FILES alembic.ini alembic/env.py)
+install(TARGETS cities DESTINATION ${CMAKE_INSTALL_PREFIX})
+install(FILES alembic.ini alembic/env.py DESTINATION /usr/share/navitia/cities/alembic)
```
TeXitoi: Should be `$INSTALL_PREFIX/share/navitia/cities/alembic`

source/cities/CMakeLists.txt
```diff
@@ -4,12 +4,11 @@ SET(BOOST_LIBS ${Boost_PROGRAM_OPTIONS_LIBRARY} ${Boost_FILESYSTEM_LIBRARY}
 add_executable(cities cities.cpp)
 target_link_libraries(cities transportation_data_import connectors types ${PQXX_LIB} ${OSMPBF} pb_lib utils
     ${BOOST_LIBS} log4cplus z protobuf)
-INSTALL_TARGETS(/usr/bin/ cities)
-
-INSTALL_FILES(/usr/share/navitia/cities/alembic FILES alembic.ini alembic/env.py)
+install(TARGETS cities DESTINATION ${CMAKE_INSTALL_PREFIX})
```
TeXitoi: Should be `$INSTALL_PREFIX/bin`

source/cities/CMakeLists.txt
```diff
@@ -4,12 +4,11 @@ SET(BOOST_LIBS ${Boost_PROGRAM_OPTIONS_LIBRARY} ${Boost_FILESYSTEM_LIBRARY}
...
+install(FILES alembic.ini alembic/env.py DESTINATION /usr/share/navitia/cities/alembic)
 
 FILE(GLOB
     alembic_cities
     alembic/versions/*.py
     )
-INSTALL_FILES(/usr/share/navitia/cities/alembic/versions FILES ${alembic_cities})
+install(FILES ${alembic_cities} DESTINATION /usr/share/navitia/cities/alembic/versions)
```
TeXitoi: Should use INSTALL_PREFIX

source/cmake_modules/ThirdParty.cmake
```diff
@@ -35,32 +34,12 @@ include_directories(SYSTEM "${CMAKE_SOURCE_DIR}/third_party/")
 # prometheus-cpp
 #
 # We want to deactivate prometheus lib if CMake version < 2.8.12.2
```
TeXitoi: comment unsync

source/kraken/metrics.h
```diff
@@ -48,25 +47,17 @@ class Registry;
...
-        Metrics(const boost::optional<std::string>&, const std::string&) {}
-        void observe_api(pbnavitia::API, double) const {}
-#endif
+protected:
+    std::unique_ptr<prometheus::Exposer> exposer;
+    std::shared_ptr<prometheus::Registry> registry;
+    std::map<pbnavitia::API, prometheus::Histogram*> request_histogram;
+public:
+    Metrics(const boost::optional<std::string>& endpoint, const std::string& coverage);
+    void observe_api(pbnavitia::API api, double duration) const;
```
TeXitoi, line 60: :+1:

TeXitoi (changes requested):

source/CMakeLists.txt
nonifier: In this case, it should rather be `/usr/local/bin`, 
edit: actually, the default value is `/usr/local` according to the doc. What was the reasoning behind using `/usr/bin` ?

source/cities/CMakeLists.txt
nonifier: `$CMAKE_INSTALL_PREFIX` should not be used as it's automatically added by cmake when the destination is a relative path. 
I would rather have `install(TARGETS cities DESTINATION bin)` to have more flexibility when installing with `DESTDIR` like `make DESTDIR=/tmp install`

source/cmake_modules/ThirdParty.cmake
```diff
@@ -35,32 +34,12 @@ include_directories(SYSTEM "${CMAKE_SOURCE_DIR}/third_party/")
...
-if("${CMAKE_VERSION}" VERSION_GREATER 2.8.12.1)
```
nonifier: Oh Yeah ! :pray:

nonifier:
Thanks for the clean up :)

nonifier (approved):

kinnou02:
poke @eturck: let's talk about that this evening/tomorrow.

source/CMakeLists.txt
benoit-bst: I know it is better to use `/usr/local`, but @kinnou02 thinks it is a tricky change because of the software ecosystem around navitia.

source/cities/CMakeLists.txt
benoit-bst: Same has above. Currently is `/usr/share`

source/CMakeLists.txt
TeXitoi: My point is that INSTALL_PREFIX doesn't end with bin, it should be /usr or /usr/local or /opt. Binaries are then installed in the bin directory.

benoit-bst:
@TeXitoi, @kinnou02  I changed the `CMAKE_INSTALL_PREFIX` (_= /usr_ for the moment). 
It is more confortable now to change the target binary location with `CMAKE_INSTALL_PREFIX =/usr/local` if we want.

benoit-bst:
Jenkins, test this please !

TeXitoi (approved):

PR #2540, opened 2018-09-19
Jormun: Newrelic improve monitoring for distributed scenario

nonifier:
Add general parameter like `version` and `scenario` to discriminate transactions.
Add a NewRelic's custom_event called `distributed` that store the following data:

Column      |   Description
------      | -------
service     | "name of the service (eg.Kraken, Vahlalalala, Here...)",
call        | "function name",
group       | "name that groups function (eg. places)",
status      | "ok" \| "failed"
duration    | duration of the request in second
[exception] | optional: the exception been caught from the external service

To query the average time spent, with number of calls by service in the last 30 min: 
```sql
SELECT average(duration), count(*) FROM distributed FACET call, service SINCE 30 minutes ago
```

To get the number of errors:
````sql
FROM distributed SELECT count(status) WHERE status LIKE 'failed'
````

### DIFF
--- a/source/jormungandr/jormungandr/new_relic.py
+++ b/source/jormungandr/jormungandr/new_relic.py
@@ -97,31 +97,37 @@ def ignore():
             logger.exception('failure while ignoring transaction')
 
 
-class DistributedEvent:
+def distributedEvent(call_name, group_name):
     """
     Custom event that we publish to New Relic for distributed scenario
     """
 
-    def __init__(self, service, call_name, group_name):
-        self.event_params = {
-            "service": type(service).__name__,
-            "call": call_name,
-            "group": group_name,
-            "status": "ok",
-        }
+    def wrap(func):
+        def wrapper(obj, service, *args, **kwargs):
+            event_params = {
+                "service": type(service).__name__,
+                "call": call_name,
+                "group": group_name,
+                "status": "ok",
+            }
 
-    def time_function(self, function, *args, **kwargs):
-        start_time = timeit.default_timer()
-        result = None
-        try:
-            result = function(*args, **kwargs)
-        except Exception as e:
-            self.event_params["status"] = "failed"
-            self.event_params.update({"exception": e})
+            start_time = timeit.default_timer()
+            result = None
+            try:
+                result = func(obj, *args, **kwargs)
+            except Exception as e:
+                event_params["status"] = "failed"
+                event_params.update({"exception": e})
+                raise
 
-        duration = timeit.default_timer() - start_time
-        self.event_params.update({"duration": duration})
+            duration = timeit.default_timer() - start_time
+            event_params.update({"duration": duration})
 
-        # Send the custom event to newrelic !
-        record_custom_event("distributed", self.event_params)
-        return result
+            # Send the custom event to newrelic !
+            record_custom_event("distributed", event_params)
+
+            return result
+
+        return wrapper
+
+    return wrap
--- a/source/jormungandr/jormungandr/new_relic.py
+++ b/source/jormungandr/jormungandr/new_relic.py
@@ -32,6 +32,7 @@ import logging
 import flask
 import os
 import timeit
+import functools
 
 try:
     from newrelic import agent
@@ -103,6 +104,7 @@ def distributedEvent(call_name, group_name):
     """
 
     def wrap(func):
+        @functools.wraps(func)
         def wrapper(obj, service, *args, **kwargs):
             event_params = {
                 "service": type(service).__name__,
--- a/source/jormungandr/jormungandr/new_relic.py
+++ b/source/jormungandr/jormungandr/new_relic.py
@@ -40,6 +40,7 @@ except ImportError:
     logger.exception('failure while importing newrelic')
     agent = None
 
+
 def init(config):
     if not agent or not config:
         return
@@ -100,14 +101,14 @@ class DistributedEvent:
     """
     Custom event that we publish to New Relic for distributed scenario
     """
+
     def __init__(self, service, call_name, group_name):
         self.event_params = {
-            "service" : type(service).__name__,
-            "call" : call_name,
-            "group" : group_name,
-            "status" : "ok",
+            "service": type(service).__name__,
+            "call": call_name,
+            "group": group_name,
+            "status": "ok",
         }
-
 
     def time_function(self, function, *args, **kwargs):
         start_time = timeit.default_timer()
@@ -116,12 +117,11 @@ class DistributedEvent:
             result = function(*args, **kwargs)
         except Exception as e:
             self.event_params["status"] = "failed"
-            self.event_params.update({"exception" : e})
+            self.event_params.update({"exception": e})
 
         duration = timeit.default_timer() - start_time
-        self.event_params.update({"duration" : duration})
+        self.event_params.update({"duration": duration})
 
         # Send the custom event to newrelic !
         record_custom_event("distributed", self.event_params)
         return result
-
--- a/source/jormungandr/jormungandr/new_relic.py
+++ b/source/jormungandr/jormungandr/new_relic.py
@@ -78,7 +78,7 @@ def record_custom_event(event_type, params):
         except RuntimeError:
             pass  # we are outside of a flask context :(
         try:
-            agent.record_custom_event(event_type, params)
+            agent.record_custom_event(event_type, params, agent.application())
         except:
             logger = logging.getLogger(__name__)
             logger.exception('failure while reporting to newrelic')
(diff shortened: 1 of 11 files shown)

### CODE REVIEW

File: source/jormungandr/jormungandr/new_relic.py [hunk 1]
```diff
@@ -94,3 +95,33 @@ def ignore():
...
+            "group": group_name,
+            "status": "ok",
+        }
+
+    def time_function(self, function, *args, **kwargs):
+        start_time = timeit.default_timer()
+        result = None
+        try:
+            result = function(*args, **kwargs)
+        except Exception as e:
```
kinnou02: You probably want to reraise this exception

source/jormungandr/jormungandr/scenarios/default.py
```diff
@@ -167,6 +168,7 @@ def parse_journey_request(self, requested_type, request):
         return req
 
     def call_kraken(self, req, instance, tag=None):
+        record_custom_parameter('scenario', "simple")
```
kinnou02: It should be default, not simple

source/jormungandr/jormungandr/new_relic.py
```diff
@@ -94,3 +95,33 @@ def ignore():
...
+
+    def __init__(self, service, call_name, group_name):
+        self.event_params = {
+            "service": type(service).__name__,
+            "call": call_name,
+            "group": group_name,
+            "status": "ok",
+        }
+
+    def time_function(self, function, *args, **kwargs):
```
kinnou02: You could use a decorator to do this without having to update every location where it's used

source/jormungandr/jormungandr/new_relic.py
```diff
@@ -94,3 +95,33 @@ def ignore():
...
+        start_time = timeit.default_timer()
+        result = None
+        try:
+            result = function(*args, **kwargs)
+        except Exception as e:
+            self.event_params["status"] = "failed"
+            self.event_params.update({"exception": e})
```
kinnou02: it's more `str(e)`

benoit-bst (approved):
I don't know new Relic

s-l300

source/jormungandr/jormungandr/new_relic.py
pbougue: or even `e.encode('utf-8', 'backslashreplace')` if you want to be supersafe.

source/jormungandr/jormungandr/new_relic.py
```diff
@@ -94,3 +95,33 @@ def ignore():
...
+    def __init__(self, service, call_name, group_name):
+        self.event_params = {
+            "service": type(service).__name__,
+            "call": call_name,
+            "group": group_name,
+            "status": "ok",
+        }
+
+    def time_function(self, function, *args, **kwargs):
+        start_time = timeit.default_timer()
```
pbougue: Why not the "classic" datetime.utcnow() ?

source/jormungandr/jormungandr/scenarios/default.py
pbougue: minor (same below): consistency about the simple quote would be nice if possible, so `'simple'`

source/jormungandr/jormungandr/scenarios/helper_classes/fallback_durations.py
```diff
@@ -152,8 +153,18 @@ def _do_request(self):
         else:
             origins = places_isochrone
             destinations = [center_isochrone]
-        sn_routing_matrix = self._instance.get_street_network_routing_matrix(
-            origins, destinations, self._mode, self._max_duration_to_pt, self._request, **self._speed_switcher
+
+        streetnetwork_service = self._instance.get_street_network(self._mode, self._request)
+        custom_event = new_relic.DistributedEvent(streetnetwork_service, "routing_matrix", "street_network")
```
pbougue: Don't know if decorators will do the job but yes, it's a bit heavy like this (but I like the functionnality).

source/jormungandr/jormungandr/scenarios/new_default.py
```diff
@@ -489,13 +490,13 @@ def culling_journeys(resp, request):
...
     Why aggregating journeys before culling journeys?
-    We have encountered severe slowness when combining max_nb_journeys(ex: 20) and a big timeframe_duration(ex: 86400s). 
-    It turned out that, with this configuration, kraken will return a lot of journeys(ex: 100 journeys) and the 
+    We have encountered severe slowness when combining max_nb_journeys(ex: 20) and a big timeframe_duration(ex: 86400s).
+    It turned out that, with this configuration, kraken will return a lot of journeys(ex: 100 journeys) and the
     algorithm was trying to figure out the best solution over 5.35E+20 possible combinations
     ( 5.35E+20=Combination(100,20) )!!
-     
+
     aggregated_journeys will group journeys that are similar('similar' is defined by 'Journeys that have the same sequence
```
pbougue, line 498: minor: would be nice to shorten this line on the way

source/jormungandr/jormungandr/new_relic.py
```diff
@@ -94,3 +95,33 @@ def ignore():
...
+
+
+class DistributedEvent:
+    """
+    Custom event that we publish to New Relic for distributed scenario
+    """
+
+    def __init__(self, service, call_name, group_name):
+        self.event_params = {
+            "service": type(service).__name__,
```
pbougue: minor (same): simple quotes if possible

I let you look (ctrl+F) along this PR if you want to change it.

pbougue (commented):
Nice that we get that, although it's a but heavy to use!

source/jormungandr/jormungandr/scenarios/new_default.py
nonifier: if it's fine with Black, it's fine with me :) :black_heart:

pbougue: Then we have a problem with black, it looks like a 123-character comment line is OK...
Which I understand as it's hard to auto-format, but then I think we should be warned and try to pay attention to that uncovered point.
Black should ease formatting, but it's not because we have something that we should stop questioning and improving it, although sometimes we can't (also true for "it's historical" :wink: )

source/jormungandr/jormungandr/new_relic.py
```diff
@@ -94,3 +95,39 @@ def ignore():
...
             logger.exception('failure while ignoring transaction')
+
+
+def distributedEvent(call_name, group_name):
+    """
+    Custom event that we publish to New Relic for distributed scenario
+    """
+
+    def wrap(func):
+        def wrapper(obj, service, *args, **kwargs):
```
kinnou02: https://docs.python.org/2/library/functools.html#functools.wraps

source/jormungandr/jormungandr/scenarios/helper_classes/fallback_durations.py
```diff
@@ -100,6 +101,12 @@ def _get_duration(self, resp, place):
         }
         return map_response[resp.routing_status]
 
+    @new_relic.distributedEvent("routing_matrix", "street_network")
+    def _get_street_network_routing_matrix(self, origins, destinations):
+        return self._instance.get_street_network_routing_matrix(
```
pbougue, line 106: Can't you decorate instance method (here `get_street_network_routing_matrix`) directly?

pbougue (commented):
Looks better but I'd like the approval of @kinnou02 or @xlqian on that (especially decorator)

source/jormungandr/jormungandr/scenarios/helper_classes/fallback_durations.py
nonifier: Nope, because it's a function that depends on the `_instance`. I don't want to decorate all the functions from every instances that implement `get_street_network_routing_matrix`. Even more, if I decorate every `get_street_network_routing_matrix` we will have to remember doing it the next instances that will be developed later... 

It's related to the open/close principle - https://en.wikipedia.org/wiki/Open%E2%80%93closed_principle

source/jormungandr/jormungandr/scenarios/new_default.py
nonifier: I've never said that we should not be questioning and improving our process, I do think we clearly should, and your first comment never was about our process.

My point was that your comment is irrelevant in this PR as I'm not going to make cosmetic changes when it is the linter's job. The point of having black is to avoid this kind of comment.

It's fine to disagree with the linter's policy, but if yo do, please, fill a change in Black's rules so we can discuss it.

kinnou02 (approved):

PR #2542, opened 2018-09-24
jormungandr: Avoid unnecessary call for street_networks

kadhikari:
**Test case:**
- service pt: kraken
- service bike: geovelo

call parameters: 
- first_section_mode[]=bike, walking, last_section_mode[]=bike, walking, min_nb_journeys=10

Before improvement: 
- Call to geovelo for direct path = 3
- Call to Geovelo for fallback durations from departure = 3
- Call to Geovelo for fallback durations from arrival = 3
- Call to Geovelo for street_network from one point to another (one streetnetwork on each next) = 3
- Total call to geovelo = 12 (4 calls to geovelo on each iteration in distributed)

After improvement: 
- Call to geovelo for direct path = 1
- Call to Geovelo for fallback durations from departure = 1
- Call to Geovelo for fallback durations from arrival = 1
- Call to Geovelo for street_network from one point to another (one streetnetwork on each next) = 1
- Total call to geovelo = 4 (4 calls to geovelo only once)

Ticket: https://jira.kisio.org/browse/NAVP-901

Test "TestJourneysDistributedWithMock" added in "routing_tests_experimental" to show the difference of call count to street_network direct_path and routing_matrix

**Side bug**
This work highlighted a bug introduced since `min_nb_journeys` feature.
There was a difference between results because of `next_request_datetime` (when _next_ is processed)

It is visible when using `distributed / new_default`, `min_nb_journeys`, `several mode combination ((walk, walk), (bike, walk), etc)`

Now for the next request date time we use minimum value among "next_request_date_time" in the response kraken.

Corrected by : https://github.com/CanalTP/navitia/pull/2542/files#diff-ff96ef579b6f85489dec9c8a3ad480f6R224
(previously we had a `next_request_dt == request_dt` in case of no journey)

### DIFF
--- a/source/jormungandr/jormungandr/scenarios/distributed.py
+++ b/source/jormungandr/jormungandr/scenarios/distributed.py
@@ -34,6 +34,7 @@ from jormungandr.utils import PeriodExtremity
 from jormungandr.street_network.street_network import StreetNetworkPathType
 from jormungandr.scenarios.helper_classes import *
 from jormungandr.scenarios.utils import fill_uris, switch_back_to_ridesharing
+from flask import g
 from jormungandr.new_relic import record_custom_parameter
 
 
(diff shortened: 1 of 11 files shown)


source/jormungandr/jormungandr/scenarios/distributed.py
```diff
@@ -57,139 +58,141 @@ def _compute_all(future_manager, request, instance, krakens_call):
...
-        for mode in requested_dep_modes:
-            streetnetwork_path_pool.add_async_request(
-                requested_orig_obj=requested_orig_obj,
-                requested_dest_obj=requested_dest_obj,
-                mode=mode,
-                period_extremity=period_extremity,
-                request=request,
-                streetnetwork_path_type=StreetNetworkPathType.DIRECT,
-            )
+        if not g.already_done:
```
kinnou02: the name is a bit too generic, there is a risk of collision

kinnou02 (commented):
This is putting a lot of things in g (that is quite costly to access).
I would prefer a solution that passes a new argument to `compute_all` that is used as context, this means that `new_default` has to be updated too but it makes thing more explicit and you are already doing stuff specific to distributed in new_default

source/jormungandr/jormungandr/scenarios/new_default.py
```diff
@@ -1094,16 +1099,12 @@ def create_next_kraken_request(self, request, responses):
         to do that we find ask the next (resp previous) query datetime
         """
 
-        # If Kraken send a new request date time, we use it
-        # for the next call to skip current Journeys
-        if responses and responses[0].HasField('next_request_date_time'):
-            request['datetime'] = responses[0].next_request_date_time
+        # We always calculate the next_request_datetime from current journeys
```
benoit-bst: Here, the focal point (Bug)

drawing

source/jormungandr/jormungandr/scenarios/distributed.py
```diff
@@ -34,21 +34,22 @@
...
 from jormungandr.scenarios.utils import fill_uris, switch_back_to_ridesharing
+from flask import g
 from jormungandr.new_relic import record_custom_parameter
+from jormungandr.scenarios.simple import Context
 
 
 class Scenario(new_default.Scenario):
     def __init__(self):
         super(Scenario, self).__init__()
+        self._context = Context()
```
kinnou02: scenario are shared between request and threads, the context will be used by multiple request.
Also context is a quite unclear name and since it's dedicated to distributed there is no reason to put in simple.

source/jormungandr/jormungandr/scenarios/new_default.py
kinnou02: the log a few lines later doesn't match what is being done (https://github.com/CanalTP/navitia/pull/2542/files?utf8=%E2%9C%93&diff=unified&w=1#diff-f050c03e2b24acf6b927fade7d249a87R1111), and should not be an error.
From what I can see `next_request_date_time` and `get_next_datetime` aren't used anymore, so they could be remove (take care of breaking change between jormun and kraken).

I don't see any tests added for this bug fix.

source/jormungandr/jormungandr/scenarios/simple.py
```diff
@@ -88,6 +88,16 @@ def isochrone_common(isochrone, request, instance, journey_req):
     journey_req.streetnetwork_params.destination_mode = isochrone.destination_modes[0]
 
 
+class Context:
+    requested_orig = None
```
kinnou02: Use python new class by inheriting of `object`

source/jormungandr/jormungandr/scenarios/distributed.py
```diff
@@ -57,139 +58,142 @@ def _compute_all(future_manager, request, instance, krakens_call):
...
         res.extend((j for j in completed_pt_journeys if j))
 
-        check_final_results_or_raise(res, orig_fallback_durations_pool, dest_fallback_durations_pool)
+        check_final_results_or_raise(res, self._context.orig_fallback_durations_pool,
+                                     self._context.dest_fallback_durations_pool)
 
         for r in res:
             fill_uris(r)
+
+        g.exist_distributed_streetnetworks = True
```
kinnou02: Once we have a proper `Context` there is no point to use `g` anymore

kinnou02 (dismissed):

source/jormungandr/jormungandr/scenarios/new_default.py
kadhikari: We have lots of tests with min_nb_journeys requiring many next calls in journey_common_tests.py
This test is also used by new_default and distributed. I don't have any idea how we can test the modification ('next_request_date_time ' not used to calculate new datetime using best journey from the response).

https://github.com/CanalTP/navitia/blob/dev/source/jormungandr/tests/journey_common_tests.py#L1830

source/jormungandr/jormungandr/scenarios/distributed.py
```diff
@@ -34,21 +34,35 @@
 from jormungandr.street_network.street_network import StreetNetworkPathType
 from jormungandr.scenarios.helper_classes import *
 from jormungandr.scenarios.utils import fill_uris, switch_back_to_ridesharing
+from flask import g
 from jormungandr.new_relic import record_custom_parameter
 
 
+class Context:
```
kinnou02: Still need to inherit from `object`

source/jormungandr/jormungandr/scenarios/distributed.py
```diff
@@ -34,21 +34,35 @@
 from jormungandr.street_network.street_network import StreetNetworkPathType
 from jormungandr.scenarios.helper_classes import *
 from jormungandr.scenarios.utils import fill_uris, switch_back_to_ridesharing
+from flask import g
```
kinnou02: `g` isn't used anymore

source/jormungandr/jormungandr/scenarios/new_default.py
```diff
@@ -1094,20 +1100,16 @@ def create_next_kraken_request(self, request, responses):
...
-            if request["clockwise"]:
-                request['datetime'] = self.next_journey_datetime(vjs, request["clockwise"])
-            else:
-                request['datetime'] = self.previous_journey_datetime(vjs, request["clockwise"])
+            request['datetime'] = self.previous_journey_datetime(vjs, request["clockwise"])
 
         if request['datetime'] is None:
             logger = logging.getLogger(__name__)
-            logger.error("In response next_request_date_time does not exist")
+            logger.error("Impossible to calculate new date_time from journeys in response for next call")
```
kinnou02: This isn't an error, at best a warning

source/jormungandr/jormungandr/scenarios/distributed.py
```diff
@@ -34,21 +34,35 @@
...
 
+class Context:
+    requested_orig = None
+    requested_dest = None
+    requested_orig_obj = None
+    requested_dest_obj = None
+    streetnetwork_path_pool = None
+    orig_fallback_durations_pool = None
+    dest_fallback_durations_pool = None
+    is_initialized = False
```
kinnou02: `Context` and `is_initialized` are very generic names

kinnou02 (commented):
This PR doesn't cover the "cache" of the fallback computation, they are still being done for every journeys.
Also Every thing is being done at the scenario level, so if we want to handle realtime journey planning in the future we will have to change everything.

source/jormungandr/jormungandr/scenarios/new_default.py
benoit-bst: We can explain you at the next *stand up*

benoit-bst (approved):

source/jormungandr/jormungandr/scenarios/new_default.py
```diff
@@ -1094,20 +1100,16 @@ def create_next_kraken_request(self, request, responses):
         to do that we find ask the next (resp previous) query datetime
         """
 
-        # If Kraken send a new request date time, we use it
-        # for the next call to skip current Journeys
-        if responses and responses[0].HasField('next_request_date_time'):
-            request['datetime'] = responses[0].next_request_date_time
```
xlqian, line 1100: (if this line can be removed, there may be some dead codes in kraken... can be handled later)

xlqian (approved):
:+1:

nonifier (approved):

source/jormungandr/jormungandr/scenarios/new_default.py
TeXitoi: You can't remove that, it will break the min_nb_journeys passed to kraken optimization.

Please add a test to check we don't break this feature.

TeXitoi (changes requested):

kinnou02:
Also, there isn't a single test that check that we don't recompute everything each time.

kadhikari:
jenkins test this please

kadhikari:
jenkins test this please

kadhikari:
jenkins test this please

kinnou02:
There is not a single test for the main subject of this PR: optimizing the number of calls to the streetnetwork engine. At best we are looking for a side effect of the optimization.
There is no guarantee that future changes won't break what has been done here. 
I won't accept nor block this PR.

source/jormungandr/tests/routing_tests_experimental.py
```diff
@@ -38,6 +38,55 @@
...
+        check_best(response)
+        self.is_valid_journey_response(response, query)
+
+        # Without optimization (context.partial_response_is_empty = True in distributed._compute_all()
+        # journey count = 18 / direct_path_call_count = 88 / routing_matrix_call_count = 76
+        # get_directpath_count_by_mode(response, 'walking') == 5
+        # get_directpath_count_by_mode(response, 'bike') == 5
+        assert len(response["journeys"]) == 10
+        assert sn_service.direct_path_call_count == 24
+        assert sn_service.routing_matrix_call_count == 20
```
kinnou02: This is huge, we should have 4 matrices: 
  - walking at departure
  - bike at departure
  - walking at arrival
  - bike at arrival

It seems way better than before but I don't understand the numbers, I might have forgot something though

kinnou02: I did understand!
`self.is_valid_journey_response(response, query)` line 78 call jormun to check the next/prev/etc... link, so we compute more journeys with more fallback. 
If we check that after the call count we get the expected values.

kinnou02:
I made the change and also rebased the PR from dev

kinnou02 (approved):

source/jormungandr/jormungandr/scenarios/distributed.py
```diff
@@ -57,142 +70,145 @@ def _compute_all(future_manager, request, instance, krakens_call):
...
-                if mode == "ridesharing":
-                    switch_back_to_ridesharing(dp, True)
-                res.append(dp)
+        if context.partial_response_is_empty:
+            for mode in requested_dep_modes:
+                dp = context.direct_paths_by_mode.get(mode).wait_and_get()
+                if getattr(dp, "journeys", None):
+                    if mode == "ridesharing":
+                        switch_back_to_ridesharing(dp, True)
+                    res.append(dp)
```
TeXitoi, line 197: First Kraken:
 - journey A to B, we compute from-A and B-to

Second Kraken:
 - journey C to D, we need from-C and D-to, but it seems that we will not compute fallbacks on the second run.

TeXitoi (dismissed):

source/jormungandr/jormungandr/scenarios/new_default.py
```diff
@@ -1162,3 +1160,10 @@ def get_entrypoint_detail(self, entrypoint, instance):
                 return bragi.get_object_by_uri(entrypoint, instances=[instance])
 
         return None
+
+    def get_next_datetime(self, responses):
+        request_datetime_list = []
+        for r in responses:
+            if r.HasField('next_request_date_time'):
+                request_datetime_list.append(r.next_request_date_time)
+        return min(request_datetime_list) if request_datetime_list else None
```
pbougue, line 1169: I maybe a bit lost... But how does this work properly if kraken does multiple inner loops? Jormun will restart at the second one, no?
The way I see it, we should first compute the max by SN-couple (max for walk-walk, max for walk-bike) to work the same as previsouly when only one couple, then the min over those max, no?
I don't know what's done about separating computations by SN-couple, but it might be the best to avoid both the bug you had and computing useless raptors.

pbougue (commented):
The issue is quite complicated, as this interwines kraken loops, jormun loops, multi-modes through the use of min_nb_journeys... This should probably be tested high-level (probably artemis) to check that both scenario work the same, but correct test design on that is not easy.
Maybe also some tests that we still call kraken with the good params (requested_dt) when using min_nb_journeys, looping into kraken and then needing a second jormun loop (after removal of some journeys by jormun), for both scenarios if needed.
This PR feels a bit dangerous.

source/jormungandr/jormungandr/scenarios/new_default.py
TeXitoi: next_request_datetime is by request, i.e. for each sn couple, kraken give the next request dt (which is the starting dt + 1s of the best journey of the last raptor call).

So it does basically what you're saying.

TeXitoi (approved):

PR #2543, opened 2018-09-24
Jormun : Fix min/max nb journey schema's type

nonifier:
`min_nb_journey` and `max_nb_journey`'s types are exposed in the schema as `type: <function natural at 0x7f3b670b7ed8>` which is not ideal for the front...

This PR fixes the type using `UnsignedInteger` & `PositiveInteger` to define them as `type: int`.

### DIFF
--- a/source/jormungandr/jormungandr/interfaces/parsers.py
+++ b/source/jormungandr/jormungandr/interfaces/parsers.py
@@ -70,18 +70,3 @@ class DateTimeFormat(CustomSchemaType):
 
     def schema(self):
         return TypeSchema(type=str, metadata={'format': 'date-time'})
-
-
-class UnsignedInteger(CustomSchemaType):
-    def __call__(self, value):
-        try:
-            d = int(value)
-            if d < 0:
-                raise ValueError('invalid positive int')
-
-            return d
-        except ValueError as e:
-            raise ValueError("Unable to evaluate, {}".format(e))
-
-    def schema(self):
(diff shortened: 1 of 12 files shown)


pbougue (commented):
It looks good, I have 2 questions non-blocking but hopefully worthy:
* Why not moving DateTimeFormat to navitiacommon also (even all the file, following the TODO written there)?
* Not sure there are tests on swagger's input's descriptions but if it's quite easy it would be nice to have some (and so check your change).

nonifier:
I agree that it would have been nice to move `DateTimeFormat`, but that would also mean moving the other types as well. Which is a much bigger change to make... I wanted to keep the changes to a minimum and² would rather address your change in a different PR as cosmetic changes.

Regarding the tests, I looked for some, but couldn't find any :( I'm happy to add some if someone with more Python experience could give me pointers on the best way to do it.

nonifier:
After talking to @xlqian, turns out that we probably can re-use the Swagger integration tests to test this. I'll try to do something with this.

nonifier:
@pbougue : I've added a test for the schema structure not to regress in the future. It's only effective on the `journeys` API so far, but once merged, we could use the same test for the other endpoints.

kinnou02 (approved):

pbougue (approved):
Nice tests!
Don't know what crashed on Jenkins, but it looks good.

And I'm OK, it's not the scope of this PR to do cleanup of the complete file.
Although I'm afraid we'll never do the cleanups if we don't do it when we work around on something.

PR #2545, opened 2018-09-25
Jormun: Schema structure testing

nonifier:
We now ensure that an endpoint referenced in the schema has its parameters defined with: 
- A name
- A description
- A type (that can be: 'integer', 'number', 'float', 'string', 'unicode', 'boolean' or 'array') 

I'm only testing endpoints with no or just the `{region}` parameter (eg. `/v1/coverage/{lon}:{lat}/journeys` isn't tested but `/v1/coverage/{region}/journeys` is ! ).

### DIFF
--- a/source/jormungandr/tests/schema_tests.py
+++ b/source/jormungandr/tests/schema_tests.py
@@ -108,11 +108,11 @@ class SchemaChecker:
 
         params = get_params(response)
 
-        assert len(params) > 1
-
         for name, param in params.iteritems():
             assert param.has_key('name')
-            assert param.has_key('description')
+            assert param.has_key('description'), (
+                "API parameter '" + param['name'] + "' should have a description in the schema!"
+            )
             assert param.has_key('type')
 
             assert type(param['name']) is unicode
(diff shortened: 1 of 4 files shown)


source/jormungandr/tests/schema_tests.py
```diff
@@ -323,8 +323,29 @@ def test_geo_status(self):
...
+        def endpoints_with_no_param(path):
+            return '{' not in path
+
+        # Filter endpoints with only a {region} parameter (ie. /v1/coverage/{region}/journeys)
+        def endpoints_with_only_region_param(path):
+            other_param = any(elem in path for elem in ['{id}', '{uri}'])
+            return '{region}' in path and not other_param
+
+        urls = list(filter(endpoints_with_no_param, paths))
+        urls.extend(list(filter(endpoints_with_only_region_param, paths)))
```
xlqian: urls = chain(filter(endpoints_with_no_param, paths), filter(endpoints_with_only_region_param, paths))

xlqian (approved):

TeXitoi (approved):

source/jormungandr/tests/schema_tests.py
nonifier: :+1:

PR #2546, opened 2018-09-26
Add dateutil dependency to navitiacommon, and link it from tyr

nonifier:
Proposal for splitting dependency requirements. 
As Tyr pulls navitiacommon, I want Tyr to relies on navitiacommon's dependencies; hence the new `requirements.txt` in navitiacommon.

This is a first step, but I would love to see all navitiacommon's requirements in the right files from now on :)

### DIFF
--- a/source/navitiacommon/requirements.txt
+++ b/source/navitiacommon/requirements.txt
@@ -1,1 +1,0 @@
-python-dateutil==2.5.3
--- a/source/navitiacommon/requirements.txt
+++ b/source/navitiacommon/requirements.txt
@@ -1,0 +1,1 @@
+python-dateutil==2.5.3
--- a/source/tyr/requirements.txt
+++ b/source/tyr/requirements.txt
@@ -1,3 +1,4 @@
+-r ../navitiacommon/requirements.txt
 Flask==0.12.4
 Flask-Migrate==1.1.1
 git+https://github.com/CanalTP/flask-restful.git@76346f65ca12a8edb9b32688cdc192ad391fa686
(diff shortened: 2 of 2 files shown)


pbougue (commented):
That's a nice idea ! Totally makes sense.
We should pass over all dependencies that are common to tyr, eitri and jormun to check if they should be in that navitiacommon's requirement.
Also, shouldn't you attach navitiacommon's requirement to eitri and jormun's requirements? (that's why I don't approve it now, but I'm not sure)

nonifier:
Thanks PE, I agree, there are way more dependencies that need to be reflected, but this PR mainly aims at fixing https://ci.navitia.io/job/deploy_navitia_on_dev/1216/.

I wanted this change to be minimal to be quickly reviewed and merged ;) But, yes, more work to be done in the future :construction_worker_man:

benoit-bst (approved):

kinnou02 (changes requested):
This requirements file are packaged and deployed on the production server, as such reference to others file won't work and prevent deployment

benoit-bst:
So, maybe it's time to update deployment scripts or it's too expensive?

nonifier:
@kinnou02 : How does deployment work then? Do we not install the dependencies via pip in production?

nonifier:
Ok, I've removed the requirement file... because deployment = pile of mud ^^

benoit-bst:
yoda
Fighting you can not always win...

benoit-bst (approved):

kinnou02 (approved):

PR #2547, opened 2018-09-26
Remove specific filter for line sections

vpassama:
Hi,

As discussed since some time now with @eturck and @kinnou02, we would like to suggest some changes about how the line sections are displayed in navitia. The main goal for us is to be able to see them in a request on `/lines/line:xx/stop_schedules` but in order to achieve this, we have to deal with the option "DumpLineSectionMessage". As you already know, doing this adds a lot of links between impacts messages and pb objects and make disruptions appear a lot more in services like journeys or stop schedules, departures...

We know that the discussion is still in progress for you but in order to make a first step, here is a PR where we try to do something :3 

To summarize what is in the changes:

- Remove DumpLineSectionMessage option (we may replace this by keeping the option and activating it with a parameter in the requests or whatever if we cannot agree on the changes)
- Try to fill the impacts messages for route points instead of stop points by adding the route in the fill_message function. This is useful to limit the links in the output to what is really impacted in a journey or in stop schedules
- Also add the route in fill_messages called for crow fly sections, for the same reason

The current work may be incomplete as I didn't manage very well with the crowfly sections and I think the tests lack some journeys with tranfers and multiple impacts with multiple lines. I'll work on this after the first comments.

### DIFF
--- a/source/jormungandr/tests/line_sections_tests.py
+++ b/source/jormungandr/tests/line_sections_tests.py
@@ -617,11 +617,11 @@ class TestLineSections(AbstractTestFixture):
     def test_journeys_use_vj_impacted_by_line_section(self):
         """
         for /journeys, we should display a line section disruption only if we use an impacted section
-
         We do all the calls as base_schedule to see the disruption but not be really impacted by them
-
-        In this test we leave at 08:00 so we'll use vj:1:1 (that is impacted) during the application
-        period of the impact
+        In this test we leave at:
+            - 08:00 using vj:1:1
+            - 10:00 using vj:1:3
+            - 11:00 using vj:2
         """
         # with this departure time we should take 'vj:1:1' that is impacted
         date = '20170102T080000'
@@ -638,21 +638,23 @@ class TestLineSections(AbstractTestFixture):
             self.is_valid_journey_response(r, q)
             return r
 
+        # If the journey contains stop points or lines impacted in a line section,
+        # the disruption is displayed and linked to them even if this
+        # line section is not used in the path
+
         # A -> B
         scenario = {
             'line_section_on_line_1': False,
             'line_section_on_line_1_other_effect': False,
             'line_section_on_line_2': False,
         }
-
-        # if we do not use an impacted section we shouldn't see the impact
         r = journeys(_from='A', to='B')
         # we check the used vj to be certain we took an impacted vj
         assert get_used_vj(r) == [['vj:1:1']]
         d = get_all_element_disruptions(r['journeys'], r)
-        assert not impacted_ids(d)
         for disruption, result in scenario.iteritems():
             assert result == (disruption in d)
+        assert not impacted_ids(d)
 
         # A -> C
         scenario = {
@@ -665,9 +667,9 @@ class TestLineSections(AbstractTestFixture):
         r = journeys(_from='A', to='C')
         assert get_used_vj(r) == [['vj:1:1']]
         d = get_all_element_disruptions(r['journeys'], r)
-        assert impacted_ids(d) == {'vj:1:1'}
         for disruption, result in scenario.iteritems():
             assert result == (disruption in d)
+        assert impacted_ids(d) == {'vj:1:1', 'C_1'}
 
         # B -> C
         scenario = {
@@ -679,9 +681,9 @@ class TestLineSections(AbstractTestFixture):
         r = journeys(_from='B', to='C')
         assert get_used_vj(r) == [['vj:1:1']]
         d = get_all_element_disruptions(r['journeys'], r)
-        assert impacted_ids(d) == {'vj:1:1'}
         for disruption, result in scenario.iteritems():
             assert result == (disruption in d)
+        assert impacted_ids(d) == {'vj:1:1', 'C_1'}
 
         # C -> D
         scenario = {
@@ -692,9 +694,9 @@ class TestLineSections(AbstractTestFixture):
         r = journeys(_from='C', to='D')
         assert get_used_vj(r) == [['vj:1:1']]
         d = get_all_element_disruptions(r['journeys'], r)
-        assert impacted_ids(d) == {'vj:1:1'}
         for disruption, result in scenario.iteritems():
             assert result == (disruption in d)
+        assert impacted_ids(d) == {'vj:1:1', 'C_1', 'D_1'}
 
         # D -> F
         scenario = {
@@ -705,13 +707,13 @@ class TestLineSections(AbstractTestFixture):
         r = journeys(_from='D', to='F')
         assert get_used_vj(r) == [['vj:1:1']]
         d = get_all_element_disruptions(r['journeys'], r)
-        assert impacted_ids(d) == {'vj:1:1'}
         for disruption, result in scenario.iteritems():
             assert result == (disruption in d)
+        assert impacted_ids(d) == {'vj:1:1', 'D_3', 'E_1', 'F_1'}
 
         # A -> F
         scenario = {
-            'line_section_on_line_1': False,
+            'line_section_on_line_1': True,
             'line_section_on_line_1_other_effect': True,
             'line_section_on_line_2': False,
         }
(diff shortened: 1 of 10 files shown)


buildbotctp:

source/type/pb_converter.cpp
```diff
@@ -1296,6 +1321,32 @@ void PbCreator::Filler::fill_messages(const VjStopTimes* vj_stoptimes,
...
+            fill_message(message, pb_obj);
+        }
+        return;
+    }
+
+    for (const auto& message: sp->get_applicable_messages(pb_creator.now, pb_creator.action_period)) {
+        bool has_link = false;
+        for (const auto& ptobj : message->informed_entities()) {
+            const auto line_section = boost::get<nt::disruption::LineSection>(&ptobj);
+            if (line_section == nullptr || line_section->line == route->line) {
```
kinnou02, line 1340: Why check for the line and not for the route?

source/routing/raptor_api.cpp
```diff
@@ -800,12 +800,32 @@ void make_pathes(PbCreator& pb_creator,
...
                 auto departure_time = path.items.front().departures.front() - seconds_to_departure;
-                pb_creator.fill_crowfly_section(origin,
-                                                destination_tmp,
-                                                crow_fly_duration,
-                                                worker.departure_path_finder.mode,
-                                                departure_time,
-                                                pb_journey);
+
+                // Getting section route in order to filter messages for route point
+                const auto& section = path.items.front();
```
kinnou02, line 805: We may also ignore disruptions in crowfly and streetnetwork section. @CanalTP/coreteam any opinion on this?
Or we could always take the stop_point object of the transport section to have to be too smart here, like what we do in the  distributed scenario.

source/jormungandr/tests/line_sections_tests.py
```diff
@@ -705,13 +707,13 @@ def journeys(_from, to):
...
         d = get_all_element_disruptions(r['journeys'], r)
-        assert impacted_ids(d) == {'vj:1:1'}
         for disruption, result in scenario.iteritems():
             assert result == (disruption in d)
+        assert impacted_ids(d) == {'vj:1:1', 'D_3', 'E_1', 'F_1'}
 
         # A -> F
         scenario = {
-            'line_section_on_line_1': False,
+            'line_section_on_line_1': True,
```
kinnou02, line 716: This look like an unwanted change, we do not want this disruption to "impact" our journey has it's still possible to do this trip even with the disruption.

source/jormungandr/tests/line_sections_tests.py
```diff
@@ -705,13 +707,13 @@ def journeys(_from, to):
         r = journeys(_from='D', to='F')
         assert get_used_vj(r) == [['vj:1:1']]
         d = get_all_element_disruptions(r['journeys'], r)
-        assert impacted_ids(d) == {'vj:1:1'}
         for disruption, result in scenario.iteritems():
             assert result == (disruption in d)
+        assert impacted_ids(d) == {'vj:1:1', 'D_3', 'E_1', 'F_1'}
```
kinnou02, line 712: `F_1` should only be impacted by `line_section_on_line_1_other_effect`
Since we now impact a lot more of objects it might be useful to check what disruption impact them.

source/jormungandr/tests/line_sections_tests.py
```diff
@@ -813,52 +930,52 @@ def test_stops_schedules_and_departures_impacted_by_line_section(self):
...
+                'line_section_on_line_1': True,
+                'line_section_on_line_1_other_effect': True,
+                'line_section_on_line_2': True,
             }
             r = self.query_region(query.format(sa='A', cur=cur, d=dt, f=fresh, q=q))
             d = get_all_element_disruptions(r[q], r)
             for disruption, result in scenario.iteritems():
                 assert result == (disruption in d)
-            assert not impacted_ids(d)
+            assert impacted_ids(d) == {'line:1', 'line:2'}
```
kinnou02, line 941: Here the `vj:1:2` shouldn't be impacted, so the stop_schedule for his route shouldn't too.
We will probably need to check thing for each vj/route_point.

Also in the response the disruption can be on multiple place, not sure there is a logic on that currently, but it might be something we have to look: with your change we return disruption that doesn't impact your stop_point (you will be pick up by the bus) but impact the vj on another part of the path (you might not be able to drop off at your desired stop).
Should we also return disruption that impact the vj but before your stop?

source/jormungandr/tests/line_sections_tests.py
```diff
@@ -870,19 +987,32 @@ def test_stops_schedules_and_departures_impacted_by_line_section(self):
...
+            'line_section_on_line_1_other_effect': False,
+            'line_section_on_line_2': False,
+        }
+        r = self.query_region(query.format(sa='F', cur=cur, d=dt, f=fresh))
+        d = get_all_element_disruptions(r['departures'], r)
+        for disruption, result in scenario.iteritems():
+            assert result == (disruption in d)
+        assert not impacted_ids(d)
+
+        query = 'stop_areas/{sa}/stop_schedules?{cur}&{d}&{f}'
```
kinnou02, line 1005: I don't understand why the result is different between stop_schedules and departures

kinnou02 (commented):
This is a first review that add more answer that anything else :)

Let's talk about it this afternoon!

source/type/pb_converter.h
```diff
@@ -387,9 +380,7 @@ struct PbCreator {
         void fill_messages(const nt::HasMessages* nav_obj, P* pb_obj){
             if (nav_obj == nullptr) {return ;}
             if (dump_message_options.dump_message == DumpMessage::No) { return; }
-            const bool dump_line_sections = dump_message_options.dump_line_section == DumpLineSectionMessage::Yes;
             for (const auto& message: nav_obj->get_applicable_messages(pb_creator.now, pb_creator.action_period)) {
-                if (!dump_line_sections && message->is_only_line_section()) { continue; }
```
pbougue, line 392: There might be a subject on removing `is_only_line_section` (so check other uses).
But maybe in another PR (or issue).

source/routing/raptor_api.cpp
pbougue: Currently, I see no point checking disruptions on virtual-street-network crowfly nor on actual streetnetwork sections.
We can think of elevators disruptions in the future, not sure it's gonna be handled the same way though.

I don't get your second idea, though @kinnou02 : Do you suggest adding disruption to every stop_point used (even streetnetwork sections) ?

source/type/pb_converter.h
```diff
@@ -252,9 +266,15 @@ struct PbCreator {
...
-                              const time_duration& crow_fly_duration, type::Mode_e mode,
-                              pt::ptime origin_time, pbnavitia::Journey* pb_journey);
+    void fill_crowfly_section(const type::EntryPoint& origin,
+                              const type::EntryPoint& destination,
+                              const time_duration& crow_fly_duration,
+                              type::Mode_e mode,
+                              pt::ptime origin_time,
+                              pbnavitia::Journey* pb_journey,
+                              const nt::Route* route = nullptr,
+                              const bool way = false);
```
pbougue, line 276: I don't understand that `route_switch` or `way` bool parameter and its use: Is it possible to have unique name, and maybe comment it?
Maybe not worth being done until we discuss the crowfly's disruptions, though.

source/jormungandr/tests/line_sections_tests.py
```diff
@@ -638,21 +638,23 @@ def journeys(_from, to):
             self.is_valid_journey_response(r, q)
             return r
 
+        # If the journey contains stop points or lines impacted in a line section,
+        # the disruption is displayed and linked to them even if this
+        # line section is not used in the path
```
pbougue, line 643: I don't understanding the usecase for the traveler here... Isn't it gonna be confusing (for integrator or for traveler) to output disruptions impacting unused part of the journey (the same line but elsewhere) ?

pbougue (commented):
Didn't read all the tests though.
Thanks for commit split, it's easier to read (at least for me) :smiley:

vpassama:
Hi, @eturck told me what was going on after the discussion with @kinnou02. So, we agree on the changes but as we need to work on other parts of kraken in order to release it in our platforms, and as we don't know the amount of time it will take to finish this one, this PR will be delayed. It will let you some time to think about details, and then we'll be able to finish it later. Thank you for your time on the review, I'll answer the comments when I work on it again.

eturck:
We had a different approach going into this problem. On our side we were thinking of each object in the API as a separate context. For example having a line nested in another object would still just be the line, so if there was a disruption on it, we would display it, independently of any context around.
It has one advantage, you have a lot of information, but on the other end, it's quite messy and could lead to some confusion on what disruption has really an effect on the journey / departure you're getting. Since it's not mandatory for us, and we can easily work around if we change our approach.

That doesn't mean there is nothing to do. There is still a bit of work to display line section disruptions where there should be, mainly in the next departures APIs.
To summarize what we said yesterday, but need to be discussed further :
- We will only display line_section disruption directly related to the stop_time in departures / stop_schedules. 
  - Directly related needs to be exactly defined. Even if the stop you are getting in is not disrupted, the stop you want to get off might be. Do we take any disruption on the VJ? Only if the disruption is after the current route_stop? Something else?
  - In stop_schedules the link to the disruption would be in the date_time.
  - In departures it could be on the links of the display_information (similar to journeys), or the stop_date_time (similar to departures). @kinnou02 was in favor of the first one but wanted to talk about it with everyone, I don't really have an opinion myself.
- In journey ideally not changing anything, except maybe displaying disruptions on intermediate stop_times (in stop_date_times of public_transport section). We would only display them in the links of each stop_date_time, and it would not affect the status of the journey. That's something we are interested in for the future, some people want to display impacted stops on journeys.

@kinnou02, if I forgot anything feel free to add it.
As @vpassama said we are quite busy right now, and really late on some projects... It might take a bit of time to finish this since we are going to work around the problem we have right now another way.

pbougue:
Hi,

To react (lately) to "where should we add link to disruption in /departures endpoint?":
I checked how we behave, and we already have a link into `display_informations` currently (checked it on our private Sherbrooke instance). So I would vote like @kinnou02 to keep everything there.

I'm also coming back on that PR to possibly close it, as no work seems to be planned on those subjects in a close future (workaround done for Tisseo, working on other kinds of disruptions for Kisio).
If that's OK with you @vpassama and @eturck , I would close it. And when the time comes to work again on that subject, we could still reopen it if it's the easiest way to move on.

PR #2548, opened 2018-09-28
Jormungandr: Fix min bound datetime

benoit-bst:
Currently, when a request is created with `datetime=19700101T001900`, an exception is raised :
```
Stack trace
exceptions:ValueError: Value out of range: -2460
```
A conversion function doesn't like the `19700101`...

I think it is the most tiny PR of the world :earth_africa:

### DIFF
--- a/source/navitiacommon/navitiacommon/parser_args_type.py
+++ b/source/navitiacommon/navitiacommon/parser_args_type.py
@@ -235,7 +235,7 @@ class DateTimeFormat(CustomSchemaType):
         """
         try:
             d = _parse_input_date(value)
-            if d.year < 1970:
+            if d.year <= 1970:
                 raise ValueError('date is too early!')
 
             return d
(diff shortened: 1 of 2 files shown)


source/navitiacommon/navitiacommon/parser_args_type.py
```diff
@@ -235,7 +235,7 @@ def __call__(self, value):
         """
         try:
             d = _parse_input_date(value)
-            if d.year < 1970:
+            if d.year <= 1970:
```
nonifier, line 238: :ok_hand:

nonifier (approved):

xlqian (approved):

pbougue (commented):
So it works for 19710101T000000 ?
Is it possible to add a test on 19700101 and 19710101 (and hand-check that the 1970's test failed previously?
It's maybe just on the first day of alltimes (so 19700101...) that we get issues on datetime conversion with TZ (that's the only "logical" issue I imagine, but my imagination is limited).
Anyway, it's an OK solution to me.

TeXitoi:
I suspect the bug is with pre 1970-01-01T00:00:00Z datetime, which the example is with timezone. Cutting out all of 1970 is OK for me, we don't care on such datetime.

benoit-bst:
I added a test, `datetime > 1970y`. 
It is the only restriction to protect early datetime.

TeXitoi (approved):

pbougue (approved):

PR #2549, opened 2018-10-01
Tyr: Add /cities endpoint

mbounabi:


benoit-bst (approved):

kadhikari (commented):
Is it possible to add at least one test?

mbounabi:
@kadhikari adding a test will imply having a real OSM file (73MB) sent and stored somewhere and also a database cities, up-to-date to store the data.

kadhikari (approved):

source/tyr/tyr/api.py
```diff
@@ -78,6 +78,8 @@
     resources.DeleteDataset, '/v0/instances/<string:instance_name>/actions/delete_dataset/<string:type>'
 )
 
+api.add_resource(resources.Cities, '/v0/cities/', '/v0/cities/status')
```
pbougue: Not sure (and not sure it matters much): That way we can post on cities/status and get a status on /cities, no?
Shouldn't we have 2 classes, one with only a get on /cities/status, and one with only a post on /cities?

pbougue (approved):
:+1: Finally a cities endpoint :tada:

source/tyr/tyr/resources.py
```diff
@@ -1634,3 +1636,43 @@ def put(self, instance_name):
...
+                return {'message': 'No file provided'}, 400
+
+            osm_file = args['file']
+            osm_file_path = str(os.path.join(os.path.abspath(current_app.config['CITIES_OSM_FILE_PATH']), "file"))
+            osm_file.save(osm_file_path)
+
+            if cities.delay(osm_file_path) != 0:
+                return {'message': 'cities failed!'}, 400
+
+            return {'message': 'OK'}, 200
```
pbougue: Not sure either: As cities import can be long, sender's side should timeout before getting that message.
Shouldn't we manage it like import_data and use finish_job? That way sender can check tyr's /jobs endpoint and see if it's job is failed, pending or done...
Letting @kinnou02 react on that, but as it does the job, it's at least a good first step to me.

source/tyr/tyr/resources.py
```diff
@@ -1634,3 +1636,43 @@ def put(self, instance_name):
             return_status = 404
 
         return {'action': return_msg}, return_status
+
+
+class Cities(flask_restful.Resource):
+    def check_db(self):
+        cities_db = sqlalchemy.create_engine(current_app.config['CITIES_DATABASE_URI'])
```
kinnou02: You could start by checking that `CITIES_DATABASE_URI` isn't `None`

source/tyr/tyr/resources.py
```diff
@@ -1634,3 +1636,43 @@ def put(self, instance_name):
...
+        try:
+            cities_db.connect()
+            result = cities_db.execute("SELECT version_num FROM alembic_version")
+            for row in result:
+                return row['version_num']
+        except Exception as e:
+            logging.exception("cities db not created : {}".format(e.message))
```
kinnou02: `Return None`

source/tyr/tyr/resources.py
kinnou02: [`delay()`](http://docs.celeryproject.org/en/latest/reference/celery.app.task.html#celery.app.task.Task.delay) return a future, so it won't wait if  we don't call `wait()`, bu it's a bit strange that comparing it to `0` doesn't fail

source/tyr/tyr/tasks.py
```diff
@@ -455,8 +455,9 @@ def cities(osm_path):
         )
         if res != 0:
             logging.error('cities failed')
-    except:
-        logging.exception('')
+    except Exception as e:
+        logging.exception('cities exception : {}'.format(e.message))
```
kinnou02, line 459: `logging.exception` already log the exception

source/tyr/tyr/default_settings.py
```diff
@@ -2,13 +2,14 @@
...
 from celery import schedules
+import os
 
 # URL for the brokker, by default it's the local rabbitmq
 # For amqp (rabbitMQ) the syntax is:
 # amqp://<user>:<password>@<host>:<port>/<vhost>
 # the default vhost is "/" so the URL end with *two* slash
 # http://docs.celeryproject.org/en/latest/configuration.html#std:setting-BROKER_URL
-CELERY_BROKER_URL = 'amqp://guest:guest@localhost:5672//'
+CELERY_BROKER_URL = os.getenv('CELERY_BROKER_URL', 'amqp://guest:guest@localhost:5672//')
```
nonifier:
```diff
- CELERY_BROKER_URL
+ TYR_CELERY_BROKER_URL
```
To be more consistent with the other options and to avoid clashing with other env var.

nonifier (approved):

nonifier (approved):

source/tyr/tyr/resources.py
mbounabi: :heavy_check_mark:

source/tyr/tyr/resources.py
mbounabi: it returns None by default.

source/tyr/tyr/tasks.py
mbounabi: log without e.message:
[2018-10-03 11:32:03,037] [ERROR] [21584] [root] cities db not created
log with e.message:
[2018-10-03 11:32:38,547] [ERROR] [21637] [root] cities db not created (psycopg2.OperationalError) FATAL:  database "cities" does not exist

source/tyr/tyr/api.py
mbounabi: Correct, this behavior isn't the best.

source/tyr/tyr/default_settings.py
mbounabi: :heavy_check_mark:

benoit-bst (approved):

source/tyr/tyr/resources.py
mbounabi: the comparison to 0 was put when the task was called without the delay. It was a mistake not to remove it when the delay() was used. However, you're right, I don't get how it doesn't fail because the id of the celery AsyncResult returned isn't 0.
???

source/tyr/tyr/resources.py
```diff
@@ -1637,6 +1639,51 @@ def put(self, instance_name):
...
+        cities_db.connect()
+        result = cities_db.execute("SELECT version_num FROM alembic_version")
+        for row in result:
+            return row['version_num']
+    except Exception as e:
+        logging.exception("cities db not created : {}".format(e.message))
+        return None
+
+
+class CitiesGET(flask_restful.Resource):
```
pbougue: Minor: here `CitiesStatus` looks clearer to me.
And below `Cities` as it might handle other methods (but we can change it when needs be).

pbougue (approved):

PR #2551, opened 2018-10-02
Jormun: use a stack in place of a queue to store opened sockets

kinnou02:
This change should reduce the number of connection opened to kraken as
unused connection will be closed by haproxy.

### DIFF
--- a/source/jormungandr/jormungandr/instance.py
+++ b/source/jormungandr/jormungandr/instance.py
@@ -105,7 +105,7 @@ class Instance(object):
         autocomplete_type,
     ):
         self.geom = None
-        self._sockets = queue.Queue()
+        self._sockets = queue.LifoQueue()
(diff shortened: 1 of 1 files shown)


benoit-bst:
If you are sure that a LIFO can solve the problem, I am agree.
I imagine you have tested it.

benoit-bst (approved):

pbougue (approved):
Don't see how it might hurt, especially if tests are OK on CI.
More queues should expire so it's good.
But we also won't renew some as often as we did, and if there are some aging problems it's good to know about it (hope we won't learn it the hard way :sweat_smile: ).

kinnou02:
with a queue we use every socket created, even if we don't need more than one at the same time. With a stack unneeded socket won't be used and there connection will be dropped.
But it's not enough, zmq reestablish the connection when i's dropped :(

PR #2555, opened 2018-10-03
Jormun: Distributed refacto before breaking the whole thingy

nonifier:
I've just prepared the codebase for the upcoming changes, where the journeys will be attached a crowfly before being filtered. Only then, fallbacks will be computed.

The only change in the business logic is that we are now waiting on the pt_journeys in the `call_kraken` **before** computing the fallbacks.

There is a new function in the scenario (`Scenario.finalise_journeys()`) that will be moved after the filtering.

### DIFF
--- a/source/jormungandr/jormungandr/scenarios/distributed.py
+++ b/source/jormungandr/jormungandr/scenarios/distributed.py
@@ -33,44 +33,58 @@ from jormungandr.scenarios import new_default
 from jormungandr.utils import PeriodExtremity
 from jormungandr.street_network.street_network import StreetNetworkPathType
 from jormungandr.scenarios.helper_classes import *
+from jormungandr.scenarios.helper_classes.complete_pt_journey import wait_and_get_pt_journeys
 from jormungandr.scenarios.utils import fill_uris, switch_back_to_ridesharing
 from jormungandr.new_relic import record_custom_parameter
 
 
-class Scenario(new_default.Scenario):
-    def __init__(self):
-        super(Scenario, self).__init__()
+class Distributed:
+    def __init__(self, future_manager, request, instance, krakens_call):
+        self.future_manager = future_manager
+        self.request = request
+        self.instance = instance
+        self.krakens_call = krakens_call
+        self.requested_orig_obj = None
+        self.requested_dest_obj = None
+        self.streetnetwork_path_pool = None
+        self.orig_fallback_durations_pool = None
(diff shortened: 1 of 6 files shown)


nonifier:
Jenkins, retest this please

nonifier:
Jenkins, test this please

source/jormungandr/jormungandr/scenarios/distributed.py
```diff
@@ -206,8 +237,8 @@ def call_kraken(self, request_type, request, instance, krakens_call):
         """
         try:
             with FutureManager() as future_manager:
-                res = self._compute_all(future_manager, request, instance, krakens_call)
-                return res
+                self._scenario = Distributed(future_manager, request, instance, krakens_call)
```
xlqian, line 240: `return Distributed(future_manager, request, instance, krakens_call)._compute_all()` and remove `self._scenario`?

source/jormungandr/jormungandr/scenarios/distributed.py
```diff
@@ -33,44 +33,58 @@
...
+        self.dest_places_free_access = None
+        self.pt_journey_pool = None
+
+    def _compute_all(self):
         """
         For all krakens_call, call the kraken and aggregate the responses
 
         return the list of all responses
         """
+        future_manager = self.future_manager
```
xlqian, line 62: I foresee a lot of conflicts....... be prepared :D

xlqian (commented):
:+1:

source/jormungandr/jormungandr/scenarios/distributed.py
```diff
@@ -186,12 +193,36 @@ def _compute_all(future_manager, request, instance, krakens_call):
...
+            pt_journeys=self.pt_journeys,
+        )
+
+        return completed_pt_journeys
+
+
+class Scenario(new_default.Scenario):
+    def __init__(self):
+        super(Scenario, self).__init__()
+        self._scenario = None
```
kinnou02, line 224: This class is used in a multithreaded context, so you should never use self to store thing.

kinnou02 (changes requested):

source/jormungandr/jormungandr/scenarios/distributed.py
nonifier: :ok_hand:  I've now using the shared context class created by @kadhikari in his last PR.

PR #2556, opened 2018-10-03
Jormun: add a memory cache layer and migration to flask_caching

kinnou02:
Flask_cache is unmaintained since 2014, it has been forked and maintained has flask_caching.
This PR also add a new cache instance named `memory_cache` that aims to
provide a first layer of cache that is faster (but local) than the
traditional `cache` that is (most of the time) backed by a redis.

The `memory_cache` is only used to reduce load on the database and as
such isn't used for realtime proxies. This is mostly due to the fact
that the priviligied backend for this cache will be the [uwsgi
cache](https://uwsgi-docs.readthedocs.io/en/latest/Caching.html) that
require to be configured with a maximum number of item and doesn't
provide both LRU and time based expiration at the same time.

### DIFF
--- a/source/jormungandr/jormungandr/authentication.py
+++ b/source/jormungandr/jormungandr/authentication.py
@@ -39,7 +39,7 @@ from jormungandr.exceptions import RegionNotFound
 import datetime
 import base64
 from navitiacommon.models import User, Instance, Key
-from jormungandr import cache, app as current_app
+from jormungandr import cache, memory_cache, app as current_app
 
 
 def authentication_required(func):
@@ -105,6 +105,7 @@ def get_token():
         return auth
 
 
+@memory_cache.memoize(current_app.config['MEMORY_CACHE_CONFIGURATION'].get('TIMEOUT_AUTHENTICATION', 30))
 @cache.memoize(current_app.config['CACHE_CONFIGURATION'].get('TIMEOUT_AUTHENTICATION', 300))
 def has_access(region, api, abort, user):
     """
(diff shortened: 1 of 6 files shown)


source/jormungandr/jormungandr/authentication.py
```diff
@@ -105,6 +105,7 @@ def get_token():
         return auth
 
 
+@memory_cache.memoize(current_app.config['MEMORY_CACHE_CONFIGURATION'].get('TIMEOUT_AUTHENTICATION', 30))
 @cache.memoize(current_app.config['CACHE_CONFIGURATION'].get('TIMEOUT_AUTHENTICATION', 300))
```
xlqian, line 109: why we need to keep the old cache for this function? with the new docorator cache, the old cache is short-circuited, right?

xlqian (approved):

source/jormungandr/jormungandr/authentication.py
kinnou02: there is two layer of cache, in memory, and redis, the local cache has a very short timeout to prevent desynchronisation between jormungandr's instances, while the redis cache allow us to not depend too much of the database.
And by default the `memory_cache` doesn't do anything, configuration will have to be updated, and it's always slow to do...

xlqian: I mean... it seems that the result is always returned by memory_cache...

kinnou02: we did talk about it directly, but for everyone else: these two layer have different cache duration, so an object can be in redis but not in memory

pbougue (approved):

xlqian:
Jenkins, test this please!

TeXitoi (approved):

PR #2557, opened 2018-10-03
Swagger: Use additional properties to test schema consistency

xlqian:
https://jira.kisio.org/browse/NAVITIAII-2576

use addtionalProperties to test schema consistency


xlqian:
Jenkins, test this please!

source/jormungandr/tests/tests_mechanism.py
```diff
@@ -495,3 +502,61 @@ def deco(cls):
...
+        return FeedPublisher(id='mock_car_park', name=self.name, license='license to kill', url='MI.6')
+
+
+def mock_car_park_providers(pois_supported):
+    providers = [MockCarParkProvider(pois_supported=pois_supported)]
+    return mock.patch(
+        'jormungandr.parking_space_availability.car.CarParkingProviderManager._get_providers',
+        new_callable=mock.PropertyMock,
+        return_value=lambda: providers,
+    )
```
pbougue: minor: missing newline

source/jormungandr/tests/schema_tests.py
```diff
@@ -74,8 +73,12 @@ def get_schema(self):
             self._schema = self.query('v1/schema')
         return self._schema
 
-    def _check_schema(self, url, hard_check=True):
+    def _check_schema(self, url, hard_check=True, additional_properties=False):
         schema = self.get_schema()
+        # 'additional_properties=False' won't allow fields that are not specifeid in schema to appear
```
pbougue: typo: specif`ie`d

source/jormungandr/tests/schema_tests.py
```diff
@@ -74,8 +73,12 @@ def get_schema(self):
             self._schema = self.query('v1/schema')
         return self._schema
 
-    def _check_schema(self, url, hard_check=True):
+    def _check_schema(self, url, hard_check=True, additional_properties=False):
```
pbougue, line 76: Not sure: From the use, `=None` would be more pythonic, no?
Otherwise, if you end up setting `schema['definitions'][key]['additionalProperties'] = True` everywhere I'd just put it like this, using additional_properties only for the condition, not for the value.

source/jormungandr/tests/tests_mechanism.py
```diff
@@ -495,3 +502,61 @@ def deco(cls):
...
+        return ParkingPlaces(available=41, occupied=42, available_PRM=43, occupied_PRM=44)
+
+    def status(self):
+        return {}
+
+    def feed_publisher(self):
+        return FeedPublisher(id='mock_car_park', name=self.name, license='license to kill', url='MI.6')
+
+
+def mock_car_park_providers(pois_supported):
```
pbougue, line 556: Is it on purpose that you don't use it anywhere?

xlqian: https://github.com/CanalTP/navitia/pull/2557/files#diff-9885c601624667a3777b344c72cab0dfR394

source/jormungandr/tests/schema_tests.py
xlqian: updated

pbougue: Wouldn't you end up erasing those (used for OSM tags in this case) : https://github.com/CanalTP/navitia/blob/0508146082bbb5dac4bae85beb26fe0a98614e7e/source/jormungandr/jormungandr/interfaces/v1/serializer/pt.py#L330 ?

xlqian: they're two different things.
the first circle is what I added in this PR
the second is in the old code
image

xlqian:
Jenkins, test this please!

source/jormungandr/tests/schema_tests.py
```diff
@@ -374,3 +376,21 @@ def test_stop_schedules(self):
...
         # we check that the response is not empty
         assert any((o.get('date_times') for o in obj.get('stop_schedules', [])))
+
+
+@dataset({"main_routing_test": {}})
+class TestSwaggerSchemaBssStands(AbstractTestFixture, SchemaChecker):
+    def test_pois_with_stands_on_first_poi(self):
+        with mock_bss_providers(pois_supported=[]):
+            self._check_schema(
+                '/v1/coverage/main_routing_test/pois/poi:station_1'
```
nonifier, line 386: Do you think this could be generalised to all (most) endpoints ? 
Maybe based on what we did here: https://github.com/CanalTP/navitia/blob/55a22f778cd88b5d0633549105a403cb8317670e/source/jormungandr/tests/schema_tests.py#L325-L346.

nonifier (approved):

xlqian:
Jenkins, test this please!

PR #2558, opened 2018-10-04
Jormun: allow pt_objects search for stop_points

kinnou02:


pbougue (approved):
Looks better like this.
2 questions though:
* So we had a bug previously on Sherbrooke's app no?
> (self-reply) No, doesn't use this endpoint.
* Did you hand-test the behavior with bragi not handling stop_points, if I remember correctly ?
> (self-reply) Same, not plugged to that endpoint currently.

benoit-bst (approved):

PR #2561, opened 2018-10-08
Run black on each commit with travis

kinnou02:
[skip ci]

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,12 +16,20 @@ before_install:
   - sudo apt update && sudo apt install -y protobuf-compiler
   - sed -i 's,git\@github.com:\([^/]*\)/\(.*\).git,https://github.com/\1/\2,' .gitmodules
   - git submodule update --init --recursive
-  - if [[ $TRAVIS_PYTHON_VERSION == '2.7' ]]; then pip install -r source/tyr/requirements_dev.txt; fi
-  - if [[ $TRAVIS_PYTHON_VERSION == '3.6' ]]; then python3.6 -m pip install black==18.9b0; fi
+
+before_script:
+  - pip install -r source/tyr/requirements_dev.txt
   - bash source/scripts/build_protobuf.sh
 
 script:
   # Make sure submodules references are merged in !
   - bash source/scripts/check_submodules.sh
-  - if [[ $TRAVIS_PYTHON_VERSION == '2.7' ]]; then pushd source/tyr && PYTHONPATH=.:../navitiacommon/ py.test && popd; fi
-  - if [[ $TRAVIS_PYTHON_VERSION == '3.6' ]]; then black --check --diff .; fi
+  - pushd source/tyr && PYTHONPATH=.:../navitiacommon/ py.test && popd
+
+matrix:
+  include:
+    - python: "3.6"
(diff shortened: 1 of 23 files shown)


nonifier (approved):
:black_heart:

.travis.yml
```diff
@@ -10,13 +10,15 @@ git:
   submodules: false
 
 before_install:
-  - sudo apt update && sudo apt install -y python-pip python-dev protobuf-compiler
+  - sudo apt update && sudo apt install -y python-pip python-dev protobuf-compiler python3-pip
   - sed -i 's,git\@github.com:\([^/]*\)/\(.*\).git,https://github.com/\1/\2,' .gitmodules
   - git submodule update --init --recursive
   - sudo pip install -r source/tyr/requirements_dev.txt
+  - pip3 install black
```
nonifier: from what I remember, pip3 uses python 3.5 when Black requires 3.6 :/ 
You need to do something like: `python3.6 -m pip install black`

pyproject.toml
```diff
@@ -11,5 +11,6 @@ exclude = '''
   | debug
   | debian
   | source/third_party
-)/
+  | source/navitiacommon/navitiacommon/.*_pb2\.py
```
pbougue: You also have chaos proto

pbougue (commented):
Weird black behavior, not sure I get the same for:
* `pre-commit run black --all` vs `black .`
* on my laptop vs yours (cf my https://github.com/CanalTP/navitia/pull/2562)

kinnou02:
Jenkins, test this please.

pyproject.toml
kinnou02: I did fix the version used for black in this PR, some of us didn't have the same version.

.travis.yml
```diff
@@ -10,13 +13,15 @@ git:
   submodules: false
 
 before_install:
-  - sudo apt update && sudo apt install -y python-pip python-dev protobuf-compiler
+  - sudo apt update && sudo apt install -y protobuf-compiler
   - sed -i 's,git\@github.com:\([^/]*\)/\(.*\).git,https://github.com/\1/\2,' .gitmodules
   - git submodule update --init --recursive
-  - sudo pip install -r source/tyr/requirements_dev.txt
+  - if [[ $TRAVIS_PYTHON_VERSION == '2.7' ]]; then pip install -r source/tyr/requirements_dev.txt; fi
+  - if [[ $TRAVIS_PYTHON_VERSION == '3.6' ]]; then python3.6 -m pip install black==18.9b0; fi
```
nonifier: Out of curiosity, how do you tell Travis to use either 2.7 or 3.6 ?

source/jormungandr/jormungandr/interfaces/v1/serializer/api.py
```diff
@@ -1,4 +1,3 @@
-
```
pbougue, line 1: Here in py PR I prefered adding `coding=utf-8` and keep that line... Anyway I'll rebase mine and it should work fine.

pbougue (approved):
Anyway it's fine for now, and we'll adjust if we have so problems between `pre-commit` and `black .`

PR #2563, opened 2018-10-09
tyr: current jobs(state='running') with data_sets should not be deleted

kadhikari:
Please see: https://jira.kisio.org/browse/NAVP-973 for details

### DIFF
--- a/source/navitiacommon/navitiacommon/models/__init__.py
+++ b/source/navitiacommon/navitiacommon/models/__init__.py
@@ -527,7 +527,8 @@ class Instance(db.Model):
             jobs_to_keep.add(job_associated)
 
         # Retrieve all jobs created before the time limit
-        old_jobs = db.session.query(Job).filter(Job.instance_id == self.id, Job.created_at < time_limit).all()
+        old_jobs = db.session.query(Job).filter(Job.instance_id == self.id, Job.created_at < time_limit,
+                                                Job.state != 'running').all()
 
         # List all jobs that can be deleted
         to_delete = list(set(old_jobs) - jobs_to_keep)
--- a/source/navitiacommon/navitiacommon/models/__init__.py
+++ b/source/navitiacommon/navitiacommon/models/__init__.py
@@ -460,6 +460,20 @@ class Instance(db.Model):
             result += data_sets
         return result
 
+    def running_datasets(self):
+        """
+        return all datasets with job state = 'running' for this instance
+        """
+        data_sets = (
+            db.session.query(DataSet)
+            .join(Job)
+            .join(Instance)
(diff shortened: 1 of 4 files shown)


mbounabi (approved):

kinnou02 (commented):
The purge run everyday use the `last_datasets()` method: https://github.com/CanalTP/navitia/blob/dev/source/tyr/tyr/tasks.py#L344

kinnou02 (changes requested):
Since this method isn't used it won't solve the problem

source/navitiacommon/navitiacommon/models/__init__.py
```diff
@@ -460,6 +460,20 @@ def last_datasets(self, nb_dataset=1, family_type=None):
...
+    def running_datasets(self):
+        """
+        return all datasets with job state = 'running' for this instance
+        """
+        data_sets = (
+            db.session.query(DataSet)
+            .join(Job)
+            .join(Instance)
+            .filter(Instance.id == self.id, Job.state == 'running')
+            .order_by(Job.created_at.desc())
```
pbougue, line 472: Why do you need that order?
If it's only for logging we can probably save it, or order it ascending as the more interesting (problematic) are the oldest ones.

source/tyr/tests/integration/purge_test.py
```diff
@@ -241,3 +257,65 @@ def test_purge_old_jobs_of_multi_instance():
     assert len(jobs_resp['jobs']) == 1
     jobs_resp = api_get('/v0/jobs/fr')
     assert len(jobs_resp['jobs']) == 1
+
+
+@pytest.mark.usefixtures("init_instances_dir")
+def test_purge_old_jobs_with_diff_job_states():
+    """
+    we shouldn't delete jobs with status 'running'. This is to avoid
+    deletinng jobs and theirs files on running state.
```
pbougue: typo: deleting (one `n`) jobs and their (no `s`) `while` running.

source/tyr/tyr/tasks.py
```diff
@@ -344,7 +344,10 @@ def purge_instance(instance_id, nb_to_keep):
         os.path.realpath(os.path.dirname(dataset.name)) for dataset in instance.last_datasets(nb_to_keep)
     )
     logger.info('loaded  data are: %s', loaded)
-    to_remove = [os.path.join(instance_config.backup_directory, f) for f in backups - loaded]
+
+    running = set(os.path.realpath(os.path.dirname(dataset.name)) for dataset in instance.running_datasets())
+    logger.info('running  data are: %s', running)
```
pbougue: running `bina` are

pbougue (approved):

kinnou02 (approved):

PR #2565, opened 2018-10-11
Distributed: use request speed in place of the instance speed

kinnou02:
By default request["*_speed"] will contains the instance speed if it
hasn't been overloaded at the request level.
This speed define  the search radius for stop_points.

I don't see any easy way to tests for this change as we only have a
microscopic streetnetwork in the tests


xlqian (approved):

TeXitoi (approved):

TeXitoi:
Do we want a test on that?

xlqian:
setup params in an instance, do a request, set params with the request, then do another request, does that check a little bit?

TeXitoi:
@xlqian just checking that we found a journey that need more than the fallback duration with default speed when we augment the speed.

benoit-bst (approved):

PR #2568, opened 2018-10-12
Jormun: fix missing feed publisher

xlqian:
https://jira.kisio.org/browse/NAVITIAII-2570

The feed_publisher is missing when requesting journeys.

According to the old code, we accept the feed_publisher returned by kraken when ALL journeys are NOT tagged as `to_delete`(tagged/filtered by jormun).

Now consider the following request(i suggest to do it twice, with and without debug=true):

http://api.navitia.io/v1/journeys?from=3.05719%3B50.63871&to=3.04797%3B50.6211&datetime=20180827T080000&datetime_represents=departure&forbidden_uris%5B%5D=physical_mode%3ALongDistanceTrain&forbidden_uris%5B%5D=physical_mode%3ATrain&forbidden_uris%5B%5D=physical_mode%3ATaxi&forbidden_uris%5B%5D=physical_mode%3ABoat&forbidden_uris%5B%5D=physical_mode%3AFerry&forbidden_uris%5B%5D=physical_mode%3AAir&forbidden_uris%5B%5D=physical_mode%3ACar&forbidden_uris%5B%5D=physical_mode%3ATrolleybus

With debug, the feed_publishers appear! 

The tricky thing here is that kraken has returned 4 journeys, one of these are filtered(a.k.a tagged as `to_delete`), the M1 journey is deleted because it looks like the another very much(kraken returned that journey because from the point of view front pareto, it makes sense). When dealing with the feed_publisher, we just omit them because there is one journey among 4 is tagged as `to_delete`.

So is it ok that replace `all` by `any` in this case?

### DIFF
--- a/source/jormungandr/jormungandr/scenarios/new_default.py
+++ b/source/jormungandr/jormungandr/scenarios/new_default.py
@@ -771,7 +771,7 @@ def merge_responses(responses, debug):
             fp
             for fp in r.feed_publishers
             if fp.id not in initial_feed_publishers
-            and (debug or all('to_delete' not in j.tags for j in r.journeys))
+            and (debug or any('to_delete' not in j.tags for j in r.journeys))
         )
 
         # handle impacts
--- a/source/jormungandr/jormungandr/scenarios/new_default.py
+++ b/source/jormungandr/jormungandr/scenarios/new_default.py
@@ -771,7 +771,7 @@ def merge_responses(responses, debug):
             fp
             for fp in r.feed_publishers
             if fp.id not in initial_feed_publishers
-            and (debug or any('to_delete' not in j.tags for j in r.journeys))
+            and (debug or len(responses) == 1 or all('to_delete' not in j.tags for j in r.journeys))
         )
 
         # handle impacts
--- a/source/jormungandr/jormungandr/scenarios/new_default.py
+++ b/source/jormungandr/jormungandr/scenarios/new_default.py
@@ -771,7 +771,7 @@ def merge_responses(responses, debug):
             fp
             for fp in r.feed_publishers
             if fp.id not in initial_feed_publishers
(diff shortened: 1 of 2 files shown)


xlqian:
i'm adding tests.

xlqian:
the test for the old logic: 
https://github.com/CanalTP/navitia/blob/dev/source/jormungandr/jormungandr/scenarios/tests/new_default_tests.py#L297

source/jormungandr/jormungandr/scenarios/new_default.py
```diff
@@ -771,7 +771,7 @@ def merge_responses(responses, debug):
             fp
             for fp in r.feed_publishers
             if fp.id not in initial_feed_publishers
-            and (debug or all('to_delete' not in j.tags for j in r.journeys))
+            and (debug or len(responses) == 1 or all('to_delete' not in j.tags for j in r.journeys))
```
xlqian: this won't work when every proto_reponse returned by kraken contain a journey tagged as 'to_delete'... 

I'm stuck...

TeXitoi: there is no `any` in python as `any_of` in C++?

xlqian: In my first commit I did this:

`(debug or any('to_delete' not in j.tags for j in r.journeys))`

which broke this test:  https://github.com/CanalTP/navitia/blob/dev/source/jormungandr/jormungandr/scenarios/tests/new_default_tests.py#L316

TeXitoi (changes requested):
Please a test for "if all the pt journeys are deleted, no feed_publisher"

benoit-bst (approved):

pbougue (approved):

TeXitoi (approved):

PR #2571, opened 2018-10-17
jormun: Sytral realtime based on sytral-rt

kinnou02:
Realtime departures for sytral, this use sytral-rt: https://github.com/CanalTP/sytralrt you should review that too before merging this PR. Sytral realtime data are provided as csv file over ftp (it's also the case for parking and equipments), It would be difficult to integrate this directly on jormungandr as it doesn't have any kind of background job and each process (around 1000 in production) is totally independent.

This connector is based on the cleverage one as we got mostly the same
information. It has the same drawback of not being able to distinguish
between the departure of two route of the same line at the same
stop_point.

The second commit isn't directly related to the sytral connector, I suggest reading each commit separately.

### DIFF
--- a/source/jormungandr/jormungandr/realtime_schedule/sytral.py
+++ b/source/jormungandr/jormungandr/realtime_schedule/sytral.py
@@ -159,10 +159,8 @@ class Sytral(RealtimeProxy):
         if not url:
             return None
         r = self._call(url)
-        if not r:
-            return None
 
-        if r.status_code != 200:
+        if r.status_code != requests.codes.ok:
             logging.getLogger(__name__).error(
                 'sytralrt RT service unavailable, impossible to query : {}'.format(r.url),
                 extra={'rt_system_id': unicode(self.rt_system_id)},
--- a/source/jormungandr/jormungandr/realtime_schedule/sytral.py
+++ b/source/jormungandr/jormungandr/realtime_schedule/sytral.py
@@ -1,0 +1,183 @@
+# coding=utf-8
+
+# Copyright (c) 2001-2016, Canal TP and/or its affiliates. All rights reserved.
+#
+# This file is part of Navitia,
+#     the software to build cool stuff with public transport.
+#
+# Hope you'll enjoy and contribute to this project,
+#     powered by Canal TP (www.canaltp.fr).
+# Help us simplify mobility and open public transport:
+#     a non ending quest to the responsive locomotion way of traveling!
+#
+# LICENCE: This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU Affero General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+# GNU Affero General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with this program. If not, see <http://www.gnu.org/licenses/>.
+#
+# Stay tuned using
+# twitter @navitia
+# IRC #navitia on freenode
+# https://groups.google.com/d/forum/navitia
+# www.navitia.io
+from __future__ import absolute_import, print_function, division
+from jormungandr.realtime_schedule.realtime_proxy import RealtimeProxy, RealtimeProxyError
+from flask import logging
+import pybreaker
+import pytz
+import requests as requests
+from jormungandr import cache, app
+from jormungandr.schedule import RealTimePassage
+import aniso8601
+
+
+class Sytral(RealtimeProxy):
+    """
+    class managing calls to sytral-rt service providing real-time next passages
+
+    """
+
+    def __init__(
+        self,
+        id,
+        service_url,
+        object_id_tag="source",
+        destination_id_tag="source",
+        instance=None,
+        timeout=2,
+        **kwargs
+    ):
+        self.service_url = service_url
+        self.timeout = timeout  # timeout in seconds
+        self.rt_system_id = id
+        self.object_id_tag = object_id_tag
+        self.destination_id_tag = destination_id_tag
+        self.instance = instance
+        self.breaker = pybreaker.CircuitBreaker(
+            fail_max=app.config['CIRCUIT_BREAKER_MAX_SYTRAL_FAIL'],
+            reset_timeout=app.config['CIRCUIT_BREAKER_SYTRAL_TIMEOUT_S'],
+        )
+        # TODO create a logger (with rt_system_id) and use it everywhere
+
+    def __repr__(self):
+        """
+         used as the cache key. we use the rt_system_id to share the cache between servers in production
+        """
+        try:
+            return self.rt_system_id.encode('utf-8', 'backslashreplace')
+        except:
+            return self.rt_system_id
+
+    @cache.memoize(app.config['CACHE_CONFIGURATION'].get('TIMEOUT_SYTRAL', 30))
+    def _call(self, url):
+        """
+        http call to sytralRT
+        """
+        logging.getLogger(__name__).debug(
+            'systralRT RT service , call url : {}'.format(url),
+            extra={'rt_system_id': unicode(self.rt_system_id)},
+        )
+        try:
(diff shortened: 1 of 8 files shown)


source/jormungandr/jormungandr/realtime_schedule/sytral.py
```diff
@@ -0,0 +1,183 @@
...
+
+    def _get_next_passage_for_route_point(
+        self, route_point, count=None, from_dt=None, current_dt=None, duration=None
+    ):
+        url = self._make_url(route_point)
+        if not url:
+            return None
+        r = self._call(url)
+        if not r:
+            return None
```
nonifier: If `r` is null, something surely went wrong. Maybe you want to log this ?

source/jormungandr/jormungandr/realtime_schedule/sytral.py
```diff
@@ -0,0 +1,183 @@
...
+
+        if r.status_code != 200:
```
nonifier: `200` is very strict and any `2xx` status code should potentially be accepted. 
Maybe you could use [`r.raise_for_status()`](http://docs.python-requests.org/en/master/_modules/requests/models/#Response.raise_for_status) instead.

source/jormungandr/jormungandr/realtime_schedule/tests/realtime_proxy_test.py
```diff
@@ -136,3 +137,28 @@ def filter_filter_dt_duration_test(mocker):
...
+    c.value = "foo"
+    c.type = "source"
+
+    c = r.codes.add()
+    c.value = "bar"
+    c.type = "extcode"
+
+    assert RoutePoint._get_all_codes(r, "source") == ["foo"]
+    assert RoutePoint._get_all_codes(r, "extcode") == ["bar"]
+    # add a duplicate, this happen in real life...
```
nonifier: :gb: 3rd person -> happen**S**

source/jormungandr/jormungandr/realtime_schedule/tests/sytral_test.py
```diff
@@ -0,0 +1,307 @@
...
+import mock
+from jormungandr.realtime_schedule.sytral import Sytral
+import validators
+import datetime
+import pytz
+import pytest
+
+
+def make_url_test():
+    sytral = Sytral(id='tata', service_url='http://bob.com/')
```
nonifier, line 40: I had to go to `http://bob.com` and sadly, nothing came up :disappointed:

nonifier (approved):

source/jormungandr/jormungandr/realtime_schedule/sytral.py
kinnou02: good catch, in fact `r` can't be null :)

source/jormungandr/jormungandr/realtime_schedule/sytral.py
kinnou02: sytral rt only return 200, 400, 404 and 500.
but I will use that:
```
r.status_code == requests.codes.ok
```

source/jormungandr/jormungandr/realtime_schedule/sytral.py
```diff
@@ -0,0 +1,181 @@
...
+        self.timeout = timeout  # timeout in seconds
+        self.rt_system_id = id
+        self.object_id_tag = object_id_tag
+        self.destination_id_tag = destination_id_tag
+        self.instance = instance
+        self.breaker = pybreaker.CircuitBreaker(
+            fail_max=app.config['CIRCUIT_BREAKER_MAX_SYTRAL_FAIL'],
+            reset_timeout=app.config['CIRCUIT_BREAKER_SYTRAL_TIMEOUT_S'],
+        )
+        # TODO create a logger (with rt_system_id) and use it everywhere
```
pbougue: minor poke on that TODO: in case you wanted to do it now.

source/jormungandr/jormungandr/realtime_schedule/sytral.py
```diff
@@ -0,0 +1,181 @@
...
+import requests as requests
+from jormungandr import cache, app
+from jormungandr.schedule import RealTimePassage
+import aniso8601
+
+
+class Sytral(RealtimeProxy):
+    """
+    class managing calls to sytral-rt service providing real-time next passages
+
```
pbougue, line 45: It would be nice to have a bit of debugging doc here like https://github.com/CanalTP/navitia/blob/dev/source/jormungandr/jormungandr/realtime_schedule/timeo.py#L50-L61
Maybe it should be also on sytralRT for the initial feed.

source/jormungandr/jormungandr/realtime_schedule/sytral.py
```diff
@@ -0,0 +1,181 @@
...
+    def _make_url(self, route_point):
+        """
+        The url returns something like a departure on a stop point
+        """
+
+        stop_id = route_point.fetch_stop_id(self.object_id_tag)
+
+        if not stop_id:
+            # one a the id is missing, we'll not find any realtime
```
pbougue: typo: the comment is probably a bit messed-up

source/jormungandr/jormungandr/realtime_schedule/sytral.py
```diff
@@ -0,0 +1,181 @@
...
+        self, route_point, count=None, from_dt=None, current_dt=None, duration=None
+    ):
+        url = self._make_url(route_point)
+        if not url:
+            return None
+        r = self._call(url)
+
+        if r.status_code != requests.codes.ok:
+            logging.getLogger(__name__).error(
+                'sytralrt RT service unavailable, impossible to query : {}'.format(r.url),
```
pbougue: maybe typo: sytral`rt RT` service

source/jormungandr/tests/proxy_realtime_sytral_integration_tests.py
```diff
@@ -0,0 +1,184 @@
...
+            "timezone": "UTC",
+            "service_url": "http://XXXX",
+            "timeout": 15,
+        },
+    }
+]
+
+
+def _get_departure(dep, sp_uri, line_code):
+    """ small helper that extract the information from a route point departures """
```
pbougue: Helpers might be shared in a common pool.

kadhikari (approved):

source/jormungandr/jormungandr/realtime_schedule/sytral.py
```diff
@@ -0,0 +1,181 @@
...
+        **kwargs
+    ):
+        self.service_url = service_url
+        self.timeout = timeout  # timeout in seconds
+        self.rt_system_id = id
+        self.object_id_tag = object_id_tag
+        self.destination_id_tag = destination_id_tag
+        self.instance = instance
+        self.breaker = pybreaker.CircuitBreaker(
+            fail_max=app.config['CIRCUIT_BREAKER_MAX_SYTRAL_FAIL'],
```
mbounabi: Minor: maybe a default value in case of missing parameter in config file

source/jormungandr/jormungandr/realtime_schedule/sytral.py
```diff
@@ -0,0 +1,181 @@
...
+        if r.status_code != requests.codes.ok:
+            logging.getLogger(__name__).error(
+                'sytralrt RT service unavailable, impossible to query : {}'.format(r.url),
+                extra={'rt_system_id': unicode(self.rt_system_id)},
+            )
+            raise RealtimeProxyError('non 200 response')
+
+        return self._get_passages(route_point, r.json())
+
+    def status(self):
```
mbounabi, line 176: Not specific to this PR but the status() function is the same for every RT providers. It can be worth moving in the RealtimeProxy class which is the parent of them all

mbounabi (approved):

source/jormungandr/jormungandr/realtime_schedule/sytral.py
kinnou02: True, but it's a bit more work as it also require to move `self.timeout`, `self.breaker` and `self.rt_system_id` to the parent class by adding a `__init__` to the parent class.
I would prefer to do that kind of refactoring in another PR.

kinnou02:
Jenkins, test this please.

pbougue (approved):
Should we merge it now?

PR #2573, opened 2018-10-18
Jormun: add tests on endpoint /vj with timezones

pbougue:
"Fix" part of https://github.com/CanalTP/kirin/issues/157

It ensures that Navitia correctly handles requests from Kirin that now sometimes use timezones in the form +0130 (for a timezone that is 1h30min "east of" UTC)

### DIFF
--- a/source/jormungandr/jormungandr/tests/authentication_test.py
+++ b/source/jormungandr/jormungandr/tests/authentication_test.py
@@ -57,7 +57,7 @@ def get_token_basic_auth_unicode_test():
     base64 doesn't seems to like unicode, it doesn't matter since our token are always in ascii (uuid).
     But we want a clean error, not a 500
     """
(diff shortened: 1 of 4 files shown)


pbougue:
This PR is failing on black on a line that I didn't change...
No obvious change to correct that from the few I searched.
I'm not planning any action about that as I already spent too much time setting black up, so I'm OK if you'd rather close this PR because of that.

pbougue (commented):
The point was to test it with str, not unicode it seems, we have to check if anyone has a better idea...

And we must check why this wasn't managed by black on the global black-init. But if I do something about black it's probably gonna be removing it :-/

nonifier (approved):
Nice one !

xlqian (approved):

kadhikari (approved):

PR #2577, opened 2018-10-19
Tyr & Jormun: update tests so they run doctests

pbougue:


nonifier:
No idea why it failed, but it's indeed a good idea the run the doctest !

kinnou02 (approved):

nonifier (approved):

PR #2579, opened 2018-10-19
Kraken: take creation of stop_time into account from FIRST

xlqian:
https://jira.kisio.org/browse/NAVP-1047

PR #2581, opened 2018-10-25
Kraken: Manage adding of StopTime in realtime

pbougue:
* Manage adding of stop time in realtime (https://jira.kisio.org/browse/NAVP-1047)
* New conf param to ignore realtime add (https://jira.kisio.org/browse/NAVP-1038)

TODO:

- [x] adapt tests so they test with realtime add allowed (first commit tests nothing is broken when it's forbidden)
- [x] deployment PR (fabric + conf)
- [x] add tests on that ? Maybe following the tests on realtime add (from #2579 )

### DIFF
--- a/source/kraken/realtime.cpp
+++ b/source/kraken/realtime.cpp
@@ -195,10 +195,16 @@ get_relationship(const transit_realtime::TripUpdate_StopTimeEvent& event,
 static nt::disruption::StopTimeUpdate::Status
 get_status(const transit_realtime::TripUpdate_StopTimeEvent& event,
            const transit_realtime::TripUpdate_StopTimeUpdate& st) {
-    if (get_relationship(event, st) ==
-            transit_realtime::TripUpdate_StopTimeUpdate_ScheduleRelationship_SKIPPED) {
+    auto transit_realtime_status = get_relationship(event, st);
+    switch (transit_realtime_status) {
+    case transit_realtime::TripUpdate_StopTimeUpdate_ScheduleRelationship_SKIPPED:
         return nt::disruption::StopTimeUpdate::Status::DELETED;
-    } else if (! event.has_delay() // for retrocompatibility since the old kirin version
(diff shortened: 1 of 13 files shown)


source/kraken/realtime.cpp
```diff
@@ -205,14 +207,44 @@ static nt::disruption::StopTimeUpdate::Status get_status(const transit_realtime:
     }
 }
 
+static bool is_realtime_add(const transit_realtime::TripUpdate& trip_update) {
```
nonifier: I'm not sure I understand the second part of the function. 
Also, the naming seems a bit confusing `is_added_service()` would probably be more meaningful. Would it make sense to have that function as part of the class `transit_realtime::TripUpdate` ?

pbougue: OK for the naming and I add some little comments on the second part (hoping they'll help).
`transit_realtime::TripUpdate` is generated from gtfs-rt proto https://github.com/CanalTP/chaos-proto/blob/master/gtfs-realtime.proto#L123 so it's not (easily) doable to have that function part of it :wink:

nonifier (approved):

kadhikari (commented):
Can't we leave some tests "navitia::handle_realtime(id, timestamp, trip_update, data,true)" without the last parameter "true" since it has default value "false" ?

pbougue (commented):
Pushed some changes while reviewing @xlqian commits.
Also added integration tests when add is forbidden (hope it matches your review @kadhikari ).
Go review/merge ! :wink:

After discussion with @kadhikari I misunderstood things a bit (as you were refering to all those changes : https://github.com/CanalTP/navitia/pull/2581/files#diff-6e43747d742ba46aec0d34afddc91e23R114).
Well I did change it as it's the final/targetted version... But it's true that meanwhile it would be cool to ensure that the default value doesn't break a thing. Didn't think doubling those tests was worth it.
Any opinion is welcome!

kadhikari (approved):

PR #2583, opened 2018-10-30
Jormun: ridesharing crowfly is never equivalent to another crowfly

kinnou02:
This change the similar_journey_filter to prevent a journey with a
ridesharing crowfly to remove a journey with a walking crowfly.
Also refacto a bit the generators to factorize the streetnetwork part

JIRA: https://jira.kisio.org/browse/NAVP-1077

### DIFF
--- a/source/jormungandr/jormungandr/scenarios/journey_filter.py
+++ b/source/jormungandr/jormungandr/scenarios/journey_filter.py
@@ -366,28 +366,37 @@ def is_walk_after_parking(journey, idx_section):
     return False
 
 
-def similar_journeys_vj_generator(journey):
+def similar_pt_section_vj(section):
+    return 'pt:%s' % section.pt_display_informations.uris.vehicle_journey
+
+
+def similar_pt_section_line(section):
+    return "pt:{}".format(section.pt_display_informations.uris.line)
+
+
+def similar_journeys_generator(journey, pt_functor):
     for idx, s in enumerate(journey.sections):
         if s.type == response_pb2.PUBLIC_TRANSPORT:
-            yield 'pt:%s' % s.pt_display_informations.uris.vehicle_journey
+            yield pt_functor(s)
         elif s.type == response_pb2.STREET_NETWORK:
             # special case, we don't want to consider the walking section after/before parking a car
             # so CAR / PARK / WALK / PT is equivalent to CAR / PARK / PT
             if is_walk_after_parking(journey, idx):
                 continue
             yield 'sn:%s' % s.street_network.mode
+        elif s.type == response_pb2.CROW_FLY and s.street_network.mode == response_pb2.Ridesharing:
+            # crowfly in ridesharing must be distinct from crowfly in walking
+            yield "sn:{}".format(s.street_network.mode)
+
+
+def similar_journeys_vj_generator(journey):
+    for v in similar_journeys_generator(journey, similar_pt_section_vj):
+        yield v
 
 
 def similar_journeys_line_generator(journey):
-    for idx, s in enumerate(journey.sections):
-        if s.type == response_pb2.PUBLIC_TRANSPORT:
-            yield "pt:{}".format(s.pt_display_informations.uris.line)
-        elif s.type == response_pb2.STREET_NETWORK:
-            # special case, we don't want to consider the walking section after/before parking a car
-            # so CAR / PARK / WALK / PT is equivalent to CAR / PARK / PT
-            if is_walk_after_parking(journey, idx):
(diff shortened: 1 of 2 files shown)


source/jormungandr/jormungandr/scenarios/journey_filter.py
```diff
@@ -366,28 +366,37 @@ def is_walk_after_parking(journey, idx_section):
...
         if s.type == response_pb2.PUBLIC_TRANSPORT:
-            yield 'pt:%s' % s.pt_display_informations.uris.vehicle_journey
+            yield pt_functor(s)
         elif s.type == response_pb2.STREET_NETWORK:
             # special case, we don't want to consider the walking section after/before parking a car
             # so CAR / PARK / WALK / PT is equivalent to CAR / PARK / PT
             if is_walk_after_parking(journey, idx):
                 continue
             yield 'sn:%s' % s.street_network.mode
+        elif s.type == response_pb2.CROW_FLY and s.street_network.mode == response_pb2.Ridesharing:
```
pbougue, line 387: Why not distinguish any different crowfly mode?
I'm thinking of SNCF's coverage, we'd want to keep bike, car and walk if we have them in params, no?

Maybe also yield that it's a crowfly to distinguish it from a classic street-network.
> may happen if we have a real car_no_park and one for ridesharing?

pbougue (commented):
:+1: bonus refactor
Curious about why only ridesharing, though.
Title of the PR would change too.

nonifier (approved):

xlqian (approved):

source/jormungandr/jormungandr/scenarios/journey_filter.py
kinnou02: I didn't want to alter what is done here, I have only handled crowfly since it's very different of others modes as it will be filtered later and can remove journeys from the final response.
As for distinguish crowfly from sn I don't really have any preferences, I don't see a case where we will have both that use the same vj.

pbougue: OK, I think we should do it for all the modes.
As for the crowfly vs. SN no opinion also, it would be a guard but I don't imagine any case where it would happen.

nonifier: @pbougue: I'm not sure I get the benefits from distinguishing crowflies for all different modes ? Maybe I'm missing something here.

If we can't think of any use case and we don't have any need for this, I suggest to keep it this way.

pbougue: If we ask for first_section_mode=[bike, walk] on SNCF and hop on the same train in both cases, one is filtered if we don't distinguish it, as it's the case currently : https://api.navitia.io/v1/coverage/sncf/journeys?from=-1.53564%3B47.22167&to=stop_area%3AOCE%3ASA%3A87391003&datetime=20181107T065000&first_section_mode%5B%5D=bike&first_section_mode%5B%5D=walking&
In my opinion we want both in that case.

nonifier: huuummmmm interesting! I agree that it would make sense to differentiate the 2 types of crowfly (the same way we would do with street network).

source/jormungandr/jormungandr/scenarios/journey_filter.py
```diff
@@ -366,28 +366,37 @@ def is_walk_after_parking(journey, idx_section):
     return False
 
 
-def similar_journeys_vj_generator(journey):
+def similar_pt_section_vj(section):
+    return 'pt:%s' % section.pt_display_informations.uris.vehicle_journey
```
mbounabi, line 370: Minor: why not use the same syntax as everywhere else with 'format'?

mbounabi (approved):

source/jormungandr/jormungandr/scenarios/journey_filter.py
pbougue: It's a minor improvement itself, but .format() is a bit slower than `%s`.

source/jormungandr/jormungandr/scenarios/journey_filter.py
```diff
@@ -366,28 +366,37 @@ def is_walk_after_parking(journey, idx_section):
...
+
+
+def similar_pt_section_line(section):
+    return "pt:{}".format(section.pt_display_informations.uris.line)
```
mbounabi, line 374: Minor: why not use the syntax '%'? ;)

PR #2588, opened 2018-11-12
Kraken: prioritize main stop_area over the others stop_area

kinnou02:
With the change we reintroduce one of the rules used by navitia 1. :disappointed: 
It allows data administrators to increase the weight of some stop_area
by setting them as main stop_area.
This is a hack to make JLP happy: we need to find a better way to
implement this as `main stop_area` are used for admin to admin journeys
and geocoding.
You may want to set a coach station as the departure/arrival of a city
when doing long distance journeys, but it's a "minor" stop in the urban
network and even in the national network so you don't want to increase its visibility when doing geocoding.

The previous version of this change was always putting main stop area before, now it's more subtile, a new parameter has been added to the api: `_main_stop_area_weight_factor` that defines the factor to be applied to the weight/score of a stop_area that is a main stop area. 
The main advantage of this solution is that it won't impact any existing users since the default value is `1.0` so nothing new will be done.
It also lets us fine tune this factor to see the impacts, manual tests has been done with a value of `1.3` and provide some good result on the IdFM datasets.

- [x] Tests to come !

### DIFF
--- a/source/autocomplete/autocomplete_api.cpp
+++ b/source/autocomplete/autocomplete_api.cpp
@@ -176,8 +176,8 @@ static void update_quality(std::vector<Autocomplete<nt::idx_t>::fl_quality>& ac_
     }
 }
 
-static std::set<std::string> get_main_stop_area(const navitia::type::Data& d){
-    std::set<std::string> result;
+static std::unordered_set<std::string> get_main_stop_areas(const navitia::type::Data& d){
+    std::unordered_set<std::string> result;
     for(const auto& admin: d.geo_ref->admins){
         for(const auto& sa: admin->main_stop_areas){
             result.insert(sa->uri);
@@ -193,7 +193,8 @@ void autocomplete(navitia::PbCreator& pb_creator, const std::string &q,
                                  int nbmax,
                                  const std::vector<std::string> &admins,
                                  int search_type,
-                                 const navitia::type::Data &d) {
+                                 const navitia::type::Data &d,
+                                 float main_sa_weight_factor) {
 
     if (q.empty()) {
         pb_creator.fill_pb_error(pbnavitia::Error::bad_filter, "Autocomplete : value of q absent");
@@ -221,6 +222,15 @@ void autocomplete(navitia::PbCreator& pb_creator, const std::string &q,
                 result = d.pt_data->stop_area_autocomplete.find_partial_with_pattern(q,
                         d.geo_ref->word_weight,
                         nbmax, valid_admin_ptr(d.pt_data->stop_areas, admin_ptr), d.geo_ref->ghostwords);
+            }
+            if(main_sa_weight_factor != 1.0){
+                std::cout << "boosting main SA" << std::endl;
+                auto main_stop_areas = get_main_stop_areas(d);
+                for(auto& r: result){
+                    if(main_stop_areas.count(d.pt_data->stop_areas[r.idx]->uri)){
+                        std::get<0>(r.scores) = std::get<0>(r.scores) * main_sa_weight_factor;
+                    }
+                }
             }
             break;
         case nt::Type_e::StopPoint:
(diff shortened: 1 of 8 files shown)


source/autocomplete/autocomplete_api.cpp
```diff
@@ -176,6 +176,16 @@ static void update_quality(std::vector<Autocomplete<nt::idx_t>::fl_quality>& ac_
     }
 }
 
+static std::set<std::string> get_main_stop_area(const navitia::type::Data& d){
```
pbougue: naming: get_main_stop_area`s`

pbougue (commented):
So the "original" list of matches doesn't change, but its order does if I undestand correctly...
Can't we loose exact matches because of that ?
I'm thinking of an exact match that's not a main SA, in the middle of a long match list of main SA... Just getting the first 10 could remove that exact match, no ?

benoit-bst:
@pbougue, we need a test to respond of that. I think @kinnou02 is on the way :running_man:

kinnou02:
@pbougue fair point, I can alter the scoring to keep exact match first

source/autocomplete/autocomplete_api.cpp
```diff
@@ -176,6 +176,16 @@ static void update_quality(std::vector<Autocomplete<nt::idx_t>::fl_quality>& ac_
...
+    std::set<std::string> result;
```
nonifier: As long as you are only doing lookups into `result`, a `std::unordered_set` would be more suitable.

source/autocomplete/autocomplete_api.cpp
```diff
@@ -315,15 +325,28 @@ void autocomplete(navitia::PbCreator& pb_creator, const std::string &q,
...
+    auto compare_attributs = [&main_stop_areas](pbnavitia::PtObject a, pbnavitia::PtObject b)->bool {
         //Sort by object type
         if (a.embedded_type() != b.embedded_type()){
             const auto a_order = get_embedded_type_order(a.embedded_type());
             const auto b_order = get_embedded_type_order(b.embedded_type());
             return  a_order < b_order;
         }
+        //prioritize main stop_area over the others stop_area
+        if((a.embedded_type() == pbnavitia::STOP_AREA) && (b.embedded_type() == pbnavitia::STOP_AREA)){
+            bool a_is_main_sa = main_stop_areas.find(a.stop_area().uri()) != main_stop_areas.end();
```
nonifier: minor: to simplify readability
```suggestion
            auto a_is_main_sa = main_stop_areas.count(a.stop_area().uri());
```

benoit-bst (approved):

pbougue (approved):
Don't you want to put a param in tyr instance's conf?

kinnou02:
I would prefer not to have to do it if possible

kadhikari (approved):

source/autocomplete/autocomplete_api.cpp
TeXitoi: That's slower. I prefer the find personally.

source/autocomplete/autocomplete_api.cpp
```diff
@@ -212,6 +223,14 @@ void autocomplete(navitia::PbCreator& pb_creator, const std::string &q,
                         d.geo_ref->word_weight,
                         nbmax, valid_admin_ptr(d.pt_data->stop_areas, admin_ptr), d.geo_ref->ghostwords);
             }
+            if(main_stop_area_weight_factor != 1.0){
+                auto main_stop_areas = get_main_stop_areas(d);
+                for(auto& r: result){
+                    if(main_stop_areas.count(d.pt_data->stop_areas[r.idx]->uri)){
+                        std::get<0>(r.scores) = std::get<0>(r.scores) * main_stop_area_weight_factor;
```
TeXitoi: Minor

```suggestion
                        std::get<0>(r.scores) *= main_stop_area_weight_factor;
```

source/autocomplete/autocomplete_api.cpp
```diff
@@ -315,6 +334,7 @@ void autocomplete(navitia::PbCreator& pb_creator, const std::string &q,
     }
 
 
+
```
TeXitoi:
```suggestion
```

TeXitoi (changes requested):
Does the exact match problem fixed?

source/autocomplete/autocomplete_api.cpp
TeXitoi: That's a set, so not slower, forget my comment, count is better.

nonifier (approved):

TeXitoi (approved):

Issue #2289, opened 2018-01-10
No date_times for stop schedule on first stop of loop line


eturck:
While testing stop schedules on a more ~~stupid~~ complex line I found a problem when there is a loop.
If you want an example [here it is on navitia.io](http://api.navitia.io/v1/coverage/fr-sw/lines/line:OTL:11821949023706503/stop_points/stop_point:OTL:SP:3377699722386833/stop_schedules)
This is Tisséo's line 111 : 
![image](https://user-images.githubusercontent.com/5136553/34780582-d1134c62-f623-11e7-8209-24ef4f6ba16c.png)
It's a tiny loop. And the first and last stop_points are the same.

When I ask for stop_schedules on this stop_point, I expect to see all the departures of the loop but instead date_times is empty. It's pretty clear as to why, because [when we render the result in departure_boards](https://github.com/CanalTP/navitia/blob/dev/source/time_tables/departure_boards.cpp#L98) : 
```cpp
for(auto dt_st : id_vec.second) {
    const auto& st_calendar = navitia::StopTimeCalendar(dt_st.second, dt_st.first, calendar_id);
    // terminus or partial terminus
    if (is_last_stoptime(st_calendar.stop_time, stop_point)) {
        continue;
    }
    auto date_time = schedule->add_date_times();
    (...)
}
```
If the stop_point is equal to the last stop_point of the vehicle_journey we don't output the stop_times. Which is a problem when the line is a loop.

Removing this 'if' doesn't have much effect for 'standard' lines because we just get pick_up stop_time : 
```cpp
stop_times = routing::get_stop_times(routing::StopEvent::pick_up, (...));
```
And in ed::Data::complete pick_up_drop_of_on_borders is called : 
```cpp
void Data::pick_up_drop_of_on_borders() {
    for (auto* vj: vehicle_journeys) {
        if (vj->stop_time_list.empty()) { continue; }
        if (! vj->prev_vj) {
            vj->stop_time_list.front()->drop_off_allowed = false;
        }
        if (! vj->next_vj) {
            vj->stop_time_list.back()->pick_up_allowed = false;
        }
    }
}
```
So standard terminuses don't have any pickup stop_times because pick_up isn't allowed on the last stop_time.
But this could be a problem for : 
- blocks of vj with prev and next vjs. I guess we would display date_times at the terminus but is it a problem if pickup is allowed? I really don't know much about block_ids. I hope you can help with this :)
[...]

PR #2589, opened 2018-11-14
Kraken: display departures on stop_point used multiple times in lines

eturck:
Hello everyone,

This is to fix issue #2289. We have (well, we had, but it could come back one day) a line with the same first and last stop_point. When we did a departure_board on this stop_point, there were no date_times, and it was tagged as "terminus". It would be nice to have them :).

This PR makes some changes on this and also the way we flag terminus and partial terminus : 
- Change the check for is_last_stoptime. Before we were only checking if the stop_point of the stop_time was the same as the last stop_point of the vj. Now we check the order of the stop time instead.
- Change the way we flag terminus and partial_terminus.
  - Before
    - It was a terminus if the stop_area of the stop_point was the same as the destination of the route.
    - If not we were checking if all the stop_times were on the same stop_point as the last one of the journey_pattern they were on. If it was the case, it was a partial terminus.
    - Tagging as partial terminus meant we had stop_times to check. It was only the case when a calendar_id was passed. Otherwise, routing::get_stop_times return an empty set of pickup stop_times on terminuses and we could not check.
    - `terminus` had the priority over `no_departure_this_day` since the check is done before.
  - Now
    - To check if it's a terminus or partial terminus we need stop_times. For calendar we have them, otherwise if the pickup stop_times are empty, we request the dropoff stop_times.
    - If all stop_times are last stop_times (check on the order) we check if the stop_area of the stop_point is the destination of the route. If it is, it's a terminus, otherwise a partial_terminus.
    - Now partial terminus works for both calendar and non calendar requests.
    - `no_departure_this_day` has the priority over `terminus`. Since we need stop_times to output the `terminus` flag, if no dropoff stop_times are returned by get_stop_times we use `no_departure_this_day`.

The last commit make some changes on schema_tests.py. In `TestSwaggerSchemaDepartureBoard` there is a few changes on additional_informations (since it is where ResponseStatus is returned in the API).
I looked at all the difference, there were two cases : 
* The first one is now we output partial_terminuses. In https://github.com/CanalTP/navitia/blob/dev/source/time_tables/tests/departure_board_test_data.h#L344 we set the same destination for all the route (we overwrite some of them afterwards) so we have quite a few partial terminuses.
* Previously departure_board on StopR2 for the vj of line `R` defined here https://github.com/CanalTP/navitia/blob/dev/source/time_tables/tests/departure_board_test_data.h#L249 had a `terminus` tag. It isn't the case anymore, since we are not on the terminus of the vj. This one explains the additional error (another additional_informations at null).

Let me know if there is something to change or if you have any question.

### DIFF
--- a/source/time_tables/tests/departure_boards_test.cpp
+++ b/source/time_tables/tests/departure_boards_test.cpp
@@ -1596,3 +1596,130 @@ BOOST_AUTO_TEST_CASE(stop_schedules_order_by_line_route_stop_point) {
     test(6, "route:13rl", "JennerRL");
     test(7, "route:13rl", "RubensRL");
 }
+
+//  Check stop_schedules on loop lines.
+//  We want to display departures on a stop even if it is the terminus too.
+BOOST_AUTO_TEST_CASE(stop_schedule_on_loop) {
+    ed::builder b("20181101");
+
+    b.vj("A", "01").uri("vj:0")
+        ("stop1", "8:00"_t, "8:00"_t)
+        ("stop2", "8:05"_t, "8:05"_t)
+        ("stop1", "8:10"_t, "8:10"_t);
+
+    b.finish();
+    b.data->pt_data->sort_and_index();
+    b.data->build_raptor();
+    b.data->pt_data->build_uri();
+    auto * data_ptr = b.data.get();
+
+    navitia::PbCreator pb_creator_dep(data_ptr, bt::second_clock::universal_time(), null_time_period);
+    departure_board(pb_creator_dep, "stop_point.uri=stop1", {}, {}, d("20181101T075500"), 86400, 3,
+                    10, 0, nt::RTLevel::Base, std::numeric_limits<size_t>::max());
+    auto resp = pb_creator_dep.get_response();
+
+    BOOST_REQUIRE_EQUAL(resp.stop_schedules_size(), 1);
+    auto stop_schedule = resp.stop_schedules(0);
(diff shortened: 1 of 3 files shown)


source/time_tables/tests/departure_boards_test.cpp
```diff
@@ -1596,3 +1596,130 @@ BOOST_AUTO_TEST_CASE(stop_schedules_order_by_line_route_stop_point) {
...
+        ("stop3", "8:10"_t, "8:10"_t);
+    b.sa("real terminus");
+
+    b.finish();
+    b.data->pt_data->sort_and_index();
+    b.data->build_raptor();
+    b.data->pt_data->build_uri();
+    auto * data_ptr = b.data.get();
+    // hacky way to set another terminus to have partial terminuses
+    data_ptr->pt_data->routes[0]->destination = data_ptr->pt_data->stop_areas[1];
```
kinnou02: You should check that `data_ptr->pt_data->stop_areas[1]` is `real terminus`, if for some reason the order change later it will help to troubleshoot

kinnou02 (approved):

source/time_tables/tests/departure_boards_test.cpp
eturck: I was lazy and did not really use a proper way to set it.
I think it's better now, I still check it in the response to be sure.

kinnou02:
Jenkins, test this please.

PR #2590, opened 2018-11-14
jormungandr: integration of new parameters for Instant System

kadhikari:
Please see https://jira.kisio.org/browse/NAVP-1074 for details

### DIFF
--- a/source/jormungandr/jormungandr/scenarios/ridesharing/instant_system.py
+++ b/source/jormungandr/jormungandr/scenarios/ridesharing/instant_system.py
@@ -66,7 +66,7 @@ class InstantSystem(AbstractRidesharingService):
         rating_scale_max=None,
         timeout=2,
         crowfly_radius=200,
-        timeframe_duration=30,
+        timeframe_duration=1800,
     ):
         self.instance = instance
         self.service_url = service_url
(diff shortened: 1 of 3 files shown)


source/jormungandr/jormungandr/scenarios/ridesharing/instant_system.py
```diff
@@ -65,6 +65,8 @@ def __init__(
         rating_scale_min=None,
         rating_scale_max=None,
         timeout=2,
+        crowfly_radius=200,
+        timeframe_duration=30,
```
pbougue: It would be better to have seconds as for almost anything else (and use them all the way if you can do `PT900S`).
Maybe rename it if you have to stay to minutes?

pbougue (commented):
Don't we mock calls to InstantSystem anywhere in tests?
(Just thinking of a way to test it a tiny bit... But that won't be awesome anyway I guess).

pbougue (approved):
OK for me, I let you see if it's easy to add those mock-check tests (test that the call to InstantSystem is correct).

nonifier (approved):

nonifier:
Are we waiting for anything special here ?

PR #2591, opened 2018-11-14
Ed: Fix GTFS parse and use SuspendedCableCar

### LINKED ISSUE #2472: Inconsistencies in public transport modes
According to the docs [here](https://github.com/CanalTP/navitia/blob/e64980568abcad5872ddfcf3b562f50236554149/documentation/slate/source/includes/objects.md#physical-mode) there is a list of valid physical modes.

However, for some GTFS files the `/journey` endpoint returns modes that are not on the list, such as "Tram". It seems to come from [this part of the code](https://github.com/CanalTP/navitia/blob/6d1eb0aa209af42151ed7c30ebcbba9a38cbe5d8/source/ed/connectors/gtfs_parser.cpp#L1122-L1189) or maybe [this one](https://github.com/CanalTP/navitia/blob/64fddf67611367fec54474d457df31eb60c8d41e/source/ed/build_helper.cpp#L659-L713).

Is it a bug? Is there any complete list of physical modes Navitia can return? Thanks.

pbougue:
fix #2472 

JIRA: https://jira.kisio.org/browse/NAVP-1007

Main part of work is here: https://github.com/CanalTP/navitia/pull/2591/files#diff-e651241f605b0265f280c709d786de39
* physical_mode:**SuspendedCableCar** is added and used
* As for the GTFS reading:
  * wrong physical_mode:Tram is renamed physical_mode:**Tramway**
  * wrong physical_mode:Rail is renamed physical_mode:**Train**

TODO:
- [x] determine if we keep French names on physical modes (discussed with @prhod)
       > :fr: in doc, :gb: in code


source/ed/connectors/gtfs_parser.cpp
```diff
@@ -1123,69 +1123,68 @@ void GenericGtfsParser::fill_default_modes(Data& data){
...
+    commercial_mode->uri = "Funicular";
+    commercial_mode->name = "Funiculaire"; // French names in NTFS so far
     data.commercial_modes.push_back(commercial_mode);
     gtfs_data.commercial_mode_map["7"] = commercial_mode;
 
-    for (ed::types::CommercialMode* mt : data.commercial_modes) {
+    for (ed::types::CommercialMode* cm : data.commercial_modes) {
         // we aggregate some GTFS modes
-        if (in(mt->uri, {"Cable_car", "Gondola"})) {
+        if (in(cm->uri, {"CableCar"})) {
```
kadhikari: Why the test on Gondola -> SuspendedCableCar omitted?

pbougue: As discussed, Gondola now has its own corresponding physical mode, no need to assign it to Funicular like we did before.
And I'll add a docker-test to check we have it ;)

documentation/slate/source/includes/apis.md
```diff
@@ -1015,7 +1015,7 @@ These disruptions won't be on the `display_informations` of the sections or used
...
 | nop     | disruption_active    | boolean | For compatibility use only.<br>If true the algorithm take the disruptions into account, and thus avoid disrupted public transport.<br>Rq: `disruption_active=true` = `data_freshness=realtime` <br>Use `data_freshness` parameter instead       |  False     |
 | nop     | wheelchair           | boolean | If true the traveler is considered to be using a wheelchair, thus only accessible public transport are used<br>be warned: many data are currently too faint to provide acceptable answers with this parameter on       | False       |
-| nop     | direct_path          | enum    | Specify if direct paths should be suggested.<br>possible values: <ul><li>indifferent</li><li>none</li><li>only</li></ul>      | indifferent |
+| nop     | direct_path          | enum    | Specify if direct paths (fallback modes only are used) should be suggested.<br>possible values: <ul><li>`indifferent`</li><li>`none` to have only journeys using some PT</li><li>`only` to have only journeys without PT</li></ul>      | indifferent |
```
nonifier: Or maybe simply :
```suggestion
| nop     | direct_path          | enum    | Specify if direct paths should be suggested (using fallback modes only).<br>possible values: <ul><li>`indifferent`</li><li>`none` for journeys using PT only</li><li>`only` for journeys without PT</li></ul>      | indifferent |
```

source/ed/connectors/gtfs_parser.cpp
```diff
@@ -1122,69 +1122,91 @@ void GenericGtfsParser::fill_default_modes(Data& data){
...
-        gtfs_data.physical_mode_map[mode->uri] = mode;
-    }
+    commercial_mode = new ed::types::CommercialMode();
+    commercial_mode->uri = "CableCar";
+    commercial_mode->name = "Cable car";
+    data.commercial_modes.push_back(commercial_mode);
+    gtfs_data.commercial_mode_map["5"] = commercial_mode;
     //for physical mode, CableCar is Funicular
-    gtfs_data.physical_mode_map["CableCar"] = gtfs_data.physical_mode_map.at("Funicular");
+    gtfs_data.physical_mode_map[commercial_mode->uri] = physical_mode;
```
kadhikari, line 1209: Does it mean: 
gtfs_data.physical_mode_map['CableCar'] = physical_mode (Funicular) 
gtfs_data.physical_mode_map['Funicular'] = physical_mode (Funicular) ??

source/ed/connectors/gtfs_parser.cpp
```diff
@@ -1122,70 +1122,91 @@ void GenericGtfsParser::fill_default_modes(Data& data){
     // default commercial_mode and physical_modes
     // all modes are represented by a number in GTFS
     // see route_type in https://developers.google.com/transit/gtfs/reference?hl=fr-FR#routestxt
+    ed::types::PhysicalMode* physical_mode = new ed::types::PhysicalMode();
```
nonifier, line 1125: C++98 style spotted :scream: :scream: :scream: :scream: :scream: 
> welcome to the :point_right: :point_left: - :evergreen_tree: :deciduous_tree:

kadhikari (commented):
Since there are some modifications on commercial_mode name, is it possible to add some tests?

source/ed/connectors/gtfs_parser.cpp
pbougue: Yes... It was already the case (and also for SuspendedCableCar before).
This map is also used as a commercial2physical map in here, which was already the case, but it's quite dirty.
I didn't want to dig too hard in something about to be deprecated.

I can add test on commercial too, it's quite easy, will do that.

source/ed/connectors/gtfs_parser.cpp
```diff
@@ -1122,70 +1122,91 @@ void GenericGtfsParser::fill_default_modes(Data& data){
...
-    commercial_mode->name = "Cable car";
-    commercial_mode->uri = "Cable_car";
-    data.commercial_modes.push_back(commercial_mode);
-    gtfs_data.commercial_mode_map["5"] = commercial_mode;
-
-    commercial_mode = new ed::types::CommercialMode();
-    commercial_mode->name = "Gondola";
-    commercial_mode->uri = commercial_mode->name;
+    commercial_mode->uri = physical_mode->uri;
+    commercial_mode->name = "Gondola, Suspended cable car";
```
benoit-bst, line 1188: `Gondola` and `Suspended cable car` in the same field... Okayy funny

source/ed/connectors/gtfs_parser.cpp
benoit-bst: same question ?

benoit-bst (approved):

source/ed/connectors/gtfs_parser.cpp
pbougue: It's all the trick of converting format into another when none of them contains the other ;)
Part of how it's supposed to be done is explained there: https://github.com/CanalTP/navitia_model/blob/master/src/documentation/gtfs_read.md#mapping-of-route_type-with-modes
And the rest is just the GTFS format: http://gtfs.org/reference#routestxt

nonifier (approved):

kadhikari (approved):

PR #2592, opened 2018-11-15
Kraken & Ed: Add visible parameters to hide inputs/ouputs

benoit-bst:
Hello my kind colleagues :cat2: 

This work includes :
- Reading geopal files with visible parameter (geopal2ed.cpp with geopal_parser.h)
  Convert string field "0" or "1" in boolean
- Update Bdd with new parameters
  Commit title : _update Bdd model with visible parameter_
- Data injection in Bdd (geopal2ed.cpp with EdPersisror.h)
- Reading Bdd (ed2nav.cpp with ed_reader.h)
- Create data.nav (ed2nav.cpp with type::data.save()) **!! new Version Number !!**
- Filtering ways with visible=false
  Commit title :  _Filter way with visible=false_
- Add integration tests
  Commit  title : _add integration test_
- Fix geopal and autocomplete unit tests
  Commit title : _fix geopal and autocomplete unit tests_
- Compatibilty with osm2ed
   Commit title : _Fix osm2ed with new visible parameter_

I see you going, where are Unit tests for _geopal2ed_ binary ?
Currently, there are no tests. 
In fact, it isn't weird because _geopal2ed_ is only use for STIF specific case.
If I have to create tests for it, it is a long work...

tests QA
- run geopal2ed : Load data with visible parameters **OK**
- run osm2ed : Test and Manage Compatibility **OK**
- run eitri : Test compatibility **OK**
- run ed2nav : **OK**
- Test with playground **OK**

  Before
  before

  After
  after

JIRA 
https://jira.kisio.org/browse/NAVP-940


source/ed/connectors/geopal_parser.cpp
```diff
@@ -382,6 +383,11 @@ void GeopalParser::fill_ways_edges(){
                     }
                     wy->type ="";
                     wy->uri = wayd_uri;
+                    if (row[visible] == "0") {
```
benoit-bst: I put the priority on visible field = true, if it is broken

source/georef/georef.h
```diff
@@ -155,7 +156,7 @@ struct Way :public nt::Nameable, nt::Header{
     nt::GeographicalCoord projected_centroid(const Graph&) const;
     nt::MultiLineString make_multiline(const Graph&) const;
     template<class Archive> void serialize(Archive & ar, const unsigned int) {
-      ar & idx & name & comment & uri & way_type & admin_list & house_number_left & house_number_right & edges & geoms;
+      ar & idx & name & comment & uri & way_type & visible & admin_list & house_number_left & house_number_right & edges & geoms;
```
benoit-bst: This is why the version number is incremented

source/georef/georef.cpp
```diff
@@ -504,9 +504,17 @@ std::vector<nf::Autocomplete<nt::idx_t>::fl_quality> GeoRef::find_ways(const std
...
+        results = fl_way.find_partial_with_pattern(search_str, word_weight, nbmax, keep_element, ghostwords);
+    }
+
+    // filter with visible parameter
+    for(auto& item: results){
+       Way* way = this->ways[item.idx];
+       if (way->visible == true) {
+            to_return.push_back(item);
+       }
     }
```
kadhikari: This code is useless. You have to exclude while adding the address in autocomplete 
see: https://github.com/CanalTP/navitia/blob/bace3353ee253b67cb88e6dc38af5b88dc3c971f/source/georef/georef.cpp#L404

benoit-bst: yes, True. @kinnou02 explained me. 
I transfer the logic. 
It doesn't impact the rest of the PR

benoit-bst: this is now 

https://github.com/CanalTP/navitia/pull/2592/files#diff-992196e00edd443955062e4bad390541R420

source/sql/alembic/versions/56c1c7a19078_add_visible_parameter.py
```diff
@@ -0,0 +1,25 @@
...
+
+from alembic import op
+import sqlalchemy as sa
+import geoalchemy2 as ga
+from sqlalchemy.dialects import postgresql
+
+
+def upgrade():
+    ### Add new flag for input/ouput visibility
+    op.add_column('way', sa.Column('visible', sa.BOOLEAN(), nullable=False), schema='georef')
```
kadhikari: You need to add te default value:
op.add_column('way', sa.Column('visible', sa.BOOLEAN(), nullable=False, server_default='false'), schema='georef')

source/ed/connectors/geopal_parser.cpp
nonifier: To be backward compatible, you might want to make sure the field `visible` exists. 
In the case where the field `visible` is missing, you'll end up with `row[-1]` => :hankey: 

```suggestion
                    if (has_col(visible, row) && row[visible] == "0") {
```

source/sql/alembic/versions/56c1c7a19078_add_visible_parameter.py
benoit-bst: Actually, the autogenerate command wrote this.
I not a expert with alembic. So, I do what you want :gift: 

I'll retest the migration to be sure

source/autocomplete/tests/test.cpp
```diff
@@ -741,9 +741,12 @@ BOOST_AUTO_TEST_CASE(autocomplete_functional_test_admin_SA_and_Address_test) {
     b.data->pt_data->sort_and_index();
 
 
-    b.add_way("rue DU TREGOR", "");
-    b.add_way("rue VIS", "");
-    b.add_way("quai NEUF", "");
+    auto w = b.add_way("rue DU TREGOR", "");
+    w->visible = true;
```
nonifier: Why not overloading `add_way()` in `build_helper.cpp` to set the field `w->visible = true` by default ? That would reduce the changes.

```cpp
navitia::georef::Way* builder::add_way(const std::string& name, const std::string& way_type, bool visible = true){...}
```

source/ed/osm2ed.cpp
```diff
@@ -298,12 +298,14 @@ void OSMCache::insert_ways(){
         values.push_back(std::to_string(way.osm_id));
         values.push_back(way.name);
         values.push_back("way:"+std::to_string(way.osm_id));
+        values.push_back("");
+        values.push_back("true");
```
nonifier, line 302: Should you not take the value from the way instead ? 

```suggestion
        values.push_back(way.visible);
```

source/tests/main_autocomplete_test.cpp
```diff
@@ -88,20 +88,30 @@ int main(int argc, const char* const argv[]) {
...
     w->name = "quai NEUF";
     w->uri = w->name;
+    w->visible = true;
     w->edges.push_back(std::make_pair(2, 3));
     b.data->geo_ref->ways.push_back(w);
+    w = new Way;
+    w->idx = 3;
+    w->name = "input/output quai NEUF";
+    w->uri = w->name;
+    w->visible = false;
```
nonifier, line 109: :+1: :see_no_evil:

source/sql/alembic/versions/56c1c7a19078_add_visible_parameter.py
pbougue: The default value of visible should be `server_default=True` here most probably.

source/ed/osm2ed.cpp
```diff
@@ -287,7 +287,7 @@ void OSMCache::insert_nodes() {
  */
 void OSMCache::insert_ways(){
     auto logger = log4cplus::Logger::getInstance("log");
-    this->lotus.prepare_bulk_insert("georef.way", {"id", "name", "uri"});
+    this->lotus.prepare_bulk_insert("georef.way", {"id", "name", "uri", "type", "visible"});
```
pbougue, line 290: Oww add of `type`, OK that was not pretty before :-)

source/ed/osm2ed.cpp
pbougue: OSM doesn't have that kind of information, I guess.

source/ed/osm2ed.cpp
```diff
@@ -298,12 +298,14 @@ void OSMCache::insert_ways(){
         values.push_back(std::to_string(way.osm_id));
         values.push_back(way.name);
         values.push_back("way:"+std::to_string(way.osm_id));
+        values.push_back("");
```
pbougue, line 301: We may do something a bit smart from `highway` tag here, but it would probably be a bit too long to handle.

source/ed/connectors/geopal_parser.cpp
pbougue: Agree with @nonifier , can be nice to (hand-)test that it's backward compatible.

pbougue:
Actually geopal is not only used for IdFM, but also Destineo and some other closed coverages.
One way to test it "relatively" easily would be through ed docker-tests, like:
https://github.com/CanalTP/navitia/blob/dev/source/ed/docker_tests/ed_integration_tests.cpp#L395-L410
That's using files in those directories: https://github.com/CanalTP/navitia/tree/dev/fixtures/ed

source/ed/connectors/geopal_parser.cpp
benoit-bst: Nice catch :heart: 
you are my heroes

source/georef/georef.h
benoit-bst: Since the filtering was moved, we don't need to keep `visible` field in data.
Now, we simply don't load ways with `visible=false` into autocomplete data.

https://github.com/CanalTP/navitia/pull/2592/files#diff-992196e00edd443955062e4bad390541R420

benoit-bst: The version number no longer needs to be incremented

source/autocomplete/tests/test.cpp
benoit-bst: OK :eagle:

source/ed/osm2ed.cpp
benoit-bst: what do you mean exactly ?
From what I see **OSM** has no information about `type`, before we didn't fill this field, anyway.

source/ed/osm2ed.cpp
benoit-bst: OSM and Geopal use the same bdd struct _georef.way_ to keep `ways`. 
to work, geopal needs visible field.
so, OSM has to handle it too. but OSM hasn't information about visibility. That why
`values.push_back("true");` all the time

benoit-bst:
We lost Destineo contract... haha
That what I said @pbougue in the header , there are no automatic tests for `Geopal`. and Navitia survived so far.
I know it is not perfect but if I have to add new files + create new binary, it is not the same development time.
I think this is not my decision @kinnou02 ?

source/ed/osm2ed.cpp
pbougue: Yep, it would be an improvement on an other feature, not the point of this PR.
From OSM we could maybe find matching types but that needs spec and a fine OSM work... Not now!

source/sql/alembic/versions/56c1c7a19078_add_visible_parameter.py
kadhikari: op.add_column('way', sa.Column('visible', sa.BOOLEAN(), nullable=False, server_default='true'), schema='georef') works well

benoit-bst:
Jenkins, test this please !

source/sql/alembic/versions/56c1c7a19078_add_visible_parameter.py
benoit-bst: After krishna's test, with server_default='true' it is OK

source/georef/georef.cpp
```diff
@@ -417,9 +417,10 @@ void GeoRef::build_autocomplete_list(){
             // Same way for all address in the admin.
             // After this modification the result found with postal code in search string
             // should contain only this postal code but not others of the admin found.
-            std::string key = way->way_type + " " + way->name + " " + admin->name + " " + admin->postal_codes_to_string();
-            fl_way.add_string(key, pos, this->ghostwords, this->synonyms);
-
+            if (way->visible == true) {
```
kinnou02: minor: you should do a continue with the check on the `name` and `edges`

nonifier (approved):
Appart from the `w->visible = true;` that could be removed from the tests, nice one !

kadhikari (approved):

pbougue (approved):

source/georef/georef.cpp
benoit-bst: It already exist ?
https://github.com/CanalTP/navitia/pull/2592/files#diff-992196e00edd443955062e4bad390541R409

pbougue: The point of @kinnou02 is to put it the same way up there I guess:
```cpp
if (way->visible == false) { continue; }
```

benoit-bst:
Jenkins, test this please !

PR #2594, opened 2018-11-15
Jormungandr: implement transient zmq socket type like http keepalive

kinnou02:
This PR improve the transient zmq socket type by making them less
transient but not definitive like the `persistent` socket type.

It look a bit like http keepalive: opened sockets are reused if
possible, after some time idle the socket will be closed.
The closing of the socket is done by a dedicated *thread* that is run
every `ZMQ_SOCKET_REAPER_INTERVAL` seconds (default: 10) and it will
close the socket if it hasn't been used since `ZMQ_SOCKET_TTL_SECONDS`
seconds (default: 10).
This will help to distribute the load on every kraken instances while
reducing the connection opened that aren't used.

If jormungandr is running inside uwsgi it will use  [uwsgi timers](https://uwsgi-docs.readthedocs.io/en/latest/PythonDecorators.html#uwsgidecorators.timer) based on their [signal framework](https://uwsgi-docs.readthedocs.io/en/latest/Signals.html).
if uwsgi isn't availlable a greenlet will be spawn by the `InstanceManager`.

I also upgraded gevent as the last version is retrocompatible, faster and with less bug.


kinnou02:
Updated to use uwsgi timer, so we aren't required to enable gevent on the uwsgi level.

nonifier (approved):

pbougue (commented):
This looks OK, but I feel I might miss something from that.
Just asking... Any way to test that?

kinnou02:
Not really, I could put some integration test with transient socket.

kinnou02:
I updated serpy integration tests to use transient socket.

benoit-bst (approved):

xlqian (approved):

source/jormungandr/jormungandr/instance.py
```diff
@@ -432,28 +432,37 @@ def max_nb_crowfly_by_mode(self):
...
+        now = time.time()
+        while True:
+            try:
+                socket, t = self._sockets.popleft()
+                if now - t > ttl:
+                    logger.debug("closing one socket for %s", self.name)
+                    socket.setsockopt(zmq.LINGER, 0)
+                    socket.close()
+                else:
+                    self._sockets.appendleft((socket, t))
```
xlqian, line 448: interesting trick... i suppose this line will help to distribute more evenly the load?

kinnou02: it's not a trick, it's a feature :wink: 
In `socket()` we use the `deque` as a stack, so the most recently used socket will be on the rigth side of the `deque` while the least rencently used socket is on the left side.
Here we start by looking if we need to close the socket on the left side, if we encounter a socket that has been used too recently to be closed we can stop as all other sockets after this one has been used more recently that this one.
At this line, since the current socket has still some time left before beeing brutaly murdered, we put it back in place on the left side.

PR #2595, opened 2018-11-16
Jormungandr: Celery version bump

nonifier:
The version bump fixes the issue (crash) when Celery was loosing communication with the RabbitMQ server.

`do_not_merge` > To be merged at the beginning of next sprint, not before.


source/tyr/tyr/__init__.py
```diff
@@ -49,6 +49,14 @@ def celery_setup_logging(*args, **kwargs):
     pass
 
 
+'''
+Since Celery 4.0, JSON is the default serializer replacing pickle.
+Problem is, some of our tasks are not JSON serializable (ie. class AutocompleteUpdateData)
+So back to pickle it is !
+'''
+app.config['CELERY_TASK_SERIALIZER'] = 'pickle'
```
nonifier, line 57: Maybe the actual fix for this would be to make sure the JSON serialisation works with all tasks like `AutocompleteUpdateData` etc... 

But that would mean, being sure we have tested them all not to crash in prod.

pbougue (approved):

kinnou02 (changes requested):
Jormungandr also depend on kombu and redis, you must upgrade it's dependencies too.

kinnou02 (approved):
merge it now, or a the start of the next iteration, since we don't have a lot of test on this part it need to live a little bit on dev to be safe.
Also did you check that upgrading kombu didn't impact statistics handling?

xlqian (approved):

buildbotctp:
## Warning 

 Files: 
- `source/type/message.h`

source/jormungandr/tests/stats_tests.py
```diff
@@ -160,6 +162,39 @@ def check_stat_places_to_publish(self, stat):
...
+
+        stat_mngr.publish_request("bla", "test")
+
+        with stat_mngr.connection.Consumer([bounded_queue], callbacks=[self._on_message]):
+            time_to_live = 100
+            while self.has_message == False and time_to_live > 0:
+                stat_mngr.connection.drain_events()
+                time_to_live -= 1
+
+        bounded_queue.delete()
```
xlqian: you can creating your queue by declaring it `auto_delete` so that you don't need to delete it manually

nonifier:
Jenkins, test this please

PR #2601, opened 2018-11-22
Add effect in trip update

kadhikari:
Ticket: https://jira.kisio.org/browse/NAVP-1080

### DIFF
--- a/source/jormungandr/tests/kirin_realtime_tests.py
+++ b/source/jormungandr/tests/kirin_realtime_tests.py
@@ -801,6 +801,8 @@ class TestKirinOnNewStopTimeAtTheEnd(MockKirinDisruptionsFixture):
         disruptions_after = self.query_region('disruptions?_current_datetime=20120614T080000')
         assert nb_disruptions_before + 1 == len(disruptions_after['disruptions'])
         assert has_the_disruption(disruptions_after, 'new_stop_time')
+        last_disrupt = disruptions_after['disruptions'][-1]
+        assert last_disrupt['severity']['effect'] == 'MODIFIED_SERVICE'
 
         journey_query = journey_basic_query + "&data_freshness=realtime&_current_datetime=20120614T080000"
         response = self.query_region(journey_query)
@@ -999,12 +1001,15 @@ class TestKirinOnNewStopTimeInBetween(MockKirinDisruptionsFixture):
                 ),
             ],
             disruption_id='vjA_delayed_with_new_stop_time',
+            effect='detour',
         )
 
         # Verify disruptions
         disrupts = self.query_region('disruptions?_current_datetime=20120614T080000')
         assert len(disrupts['disruptions']) == 12
         assert has_the_disruption(disrupts, 'vjA_delayed_with_new_stop_time')
+        last_disrupt = disrupts['disruptions'][-1]
+        assert last_disrupt['severity']['effect'] == 'DETOUR'
 
         # the journey has the new stop_time in its section of public_transport
         response = self.query_region(sub_query + "&data_freshness=realtime&datetime=20120614T080200")
@@ -1125,6 +1130,7 @@ class TestKirinOnNewStopTimeAtTheBeginning(MockKirinDisruptionsFixture):
                 ),
             ],
             disruption_id='new_stop_time',
+            effect='delayed'
         )
 
         # Verify disruptions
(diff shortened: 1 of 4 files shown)


buildbotctp:
## Warning 

 Files: 
- `source/type/message.h`

buildbotctp:
## Warning 

 Files: 
- `source/type/message.h`

source/kraken/realtime.cpp
```diff
@@ -401,10 +451,11 @@ create_disruption(const std::string& id,
                 impact->aux_info.stop_times.emplace_back(std::move(st_update));
             }
 
-            trip_effect = get_trip_effect(most_important_stoptime_status);
+            trip_effect = get_calculated_trip_effect(most_important_stoptime_status);
```
pbougue, line 448: Maybe you can check if the extension is activated and do only one processing of Effect.
This is worth being grouped (has_effect declaration, this line and the use of `has_effect`).

source/kraken/realtime.cpp
```diff
@@ -316,6 +364,8 @@ create_disruption(const std::string& id,
             impact->messages.push_back(create_message(trip_update.GetExtension(kirin::trip_message)));
         }
         std::string wording;
+        // TODO: to be removed later when effect completely implemented in trip_update
```
nonifier: Maybe we should reference this somewhere in Jira

source/kraken/realtime.cpp
```diff
@@ -257,6 +256,53 @@ static nt::disruption::Effect get_trip_effect(nt::disruption::StopTimeUpdate::St
 }
 
 
+static nt::disruption::Effect get_trip_effect(transit_realtime::Alert_Effect effect) {
```
nonifier, line 257: Just by curiosity, what's the difference between a `transit_realtime::Alert_Effect` and a `nt::disruption::Effect` ? seems to be a perfect 1:1 mapping.

source/kraken/realtime.cpp
```diff
@@ -289,6 +335,8 @@ create_disruption(const std::string& id,
     if (trip_update.trip().HasExtension(kirin::contributor)) {
         disruption.contributor = trip_update.trip().GetExtension(kirin::contributor);
     }
+    bool has_effect = trip_update.HasExtension(kirin::effect);
```
nonifier: minor: It would be nice to have this line where it's being used (eg. 150 lines below :arrow_down: )

nonifier (approved):

source/kraken/realtime.cpp
kadhikari: The first is the Extension added in kirin.proto where as other is an enum class in kraken. Kraken transforms gtfs-rt effect to disruption.Effect during treatment and retransforms to disruption.severity.effect during response generation in pb_converter.

source/kraken/realtime.cpp
kadhikari: ok

source/kraken/realtime.cpp
pbougue: It looks like we can just skip that conversion, and go directly to wordings when needed, no?
So maybe remove that enum `nt::disruption::Effect` if not used anywhere else.

So just change the output type of `get_calculated_trip_effect` to match it.
And change the input type of `get_wordings`

source/kraken/realtime.cpp
```diff
@@ -413,6 +464,11 @@ create_disruption(const std::string& id,
                 wording = "trip modified";
             }
         }
+        if (has_effect){
+            trip_effect = get_trip_effect(trip_update.GetExtension(kirin::effect));
+            wording = get_wordings(trip_effect);
```
pbougue: This line should be common no matter how we get trip_effect, no?

source/jormungandr/tests/kirin_realtime_tests.py
```diff
@@ -1287,6 +1293,13 @@ def make_mock_kirin_item(vj_id, date, status='canceled', new_stop_time_list=[],
     trip.start_date = date
     trip.Extensions[kirin_pb2.contributor] = rt_topic
 
+    if effect == 'modified':
```
pbougue, line 1296: Minor (to be decided once other comments are decided): it would be better to have enum from the beginning.
But it's a helper and the enum can be long, depending on what changes from other comments, it may not be worth it.

source/kraken/realtime.cpp
pbougue: cf. other comment on mutualizing wording computation.
But it looks like we will always need that string to build impact severity.

pbougue (commented):
Functionally looks good, the code can probably be cleaned up a bit.

TeXitoi (approved):

pbougue (approved):

kadhikari:
jenkins test this please

source/tyr/requirements.txt
```diff
@@ -31,5 +31,5 @@ validate-email==1.1
 wsgiref==0.1.2
 retrying==1.3.3
 ujson==1.35
-jsonschema==2.5.1
+jsonschema==2.6.0
```
nonifier, line 34: :+1: :pray:

nonifier (approved):

kadhikari:
jenkins test this please

nonifier:
test this please

PR #2602, opened 2018-11-22
Tyr: add poi_types json configuration management to mimir autocomplete

pbougue:
At bina, a file containing POI conf will be created in the backup directory, to be passed to osm2mimir.

- [x] Some tests added on tyr's API
- [x] Hand-tested for full chain

JIRA: https://jira.kisio.org/browse/NAVP-1085

### DIFF
--- a/source/tyr/tyr/resources.py
+++ b/source/tyr/tyr/resources.py
@@ -107,7 +107,36 @@ class Job(flask_restful.Resource):
         return {'message': 'OK'}, 200
 
 
-class PoiType(flask_restful.Resource):
+def _validate_poi_types_json(poi_types_json):
+    try:
+        validate(poi_types_json, poi_type_conf_format)
+    except ValidationError as e:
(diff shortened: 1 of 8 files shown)


kinnou02 (approved):

source/tyr/tyr/resources.py
```diff
@@ -107,7 +107,36 @@ def post(self, instance_name):
         return {'message': 'OK'}, 200
 
 
-class PoiType(flask_restful.Resource):
+def _validate_poi_types_json(poi_types_json):
```
kadhikari, line 110: Since this function does more verifications than the generic function 'validate', comments to explain are always helpful
> OK, I'll add a comment on why we do those extra checks :+1:

kadhikari (approved):

kadhikari (approved):

TeXitoi (approved):

PR #2603, opened 2018-11-23
clean up tcmalloc linking

kinnou02:
With this PR only final executable are linked to tcmalloc, not the
library. This make it easier to choose another allocator later.

### DIFF
--- a/source/type/CMakeLists.txt
+++ b/source/type/CMakeLists.txt
@@ -92,7 +92,7 @@ endif("${CMAKE_CXX_COMPILER_ID}" MATCHES "Clang")
 SET_SOURCE_FILES_PROPERTIES(${PROTO_SRCS} PROPERTIES COMPILE_FLAGS ${PROTO_CMAKE_CXX_FLAGS})
 
 add_library(pb_lib ${PROTO_SRCS} pb_converter.cpp)
-target_link_libraries(pb_lib thermometer vptranslator pthread ${PROTOBUF_LIBRARY} tcmalloc)
+target_link_libraries(pb_lib thermometer vptranslator pthread ${PROTOBUF_LIBRARY})
 
 add_library(types type.cpp message.cpp datetime.cpp geographical_coord.cpp timezone_manager.cpp validity_pattern.cpp type_utils.cpp)
 target_link_libraries(types ptreferential utils pb_lib protobuf)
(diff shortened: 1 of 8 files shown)


buildbotctp:
## Warning 

 Files: 
- `source/type/message.h`

benoit-bst (approved):

benoit-bst:
Binaries like geopal2ed, ed2nav don't use `Tc Malloc` ?  
Okay, it looks fine to me.

source/type/CMakeLists.txt
```diff
@@ -92,7 +92,7 @@ endif("${CMAKE_CXX_COMPILER_ID}" MATCHES "Clang")
 SET_SOURCE_FILES_PROPERTIES(${PROTO_SRCS} PROPERTIES COMPILE_FLAGS ${PROTO_CMAKE_CXX_FLAGS})
 
 add_library(pb_lib ${PROTO_SRCS} pb_converter.cpp)
-target_link_libraries(pb_lib thermometer vptranslator pthread ${PROTOBUF_LIBRARY} tcmalloc)
+target_link_libraries(pb_lib thermometer vptranslator pthread ${PROTOBUF_LIBRARY})
```
pbougue, line 95: So `cities` is still linked with tcmalloc, because `connectors` are (I don't know about other dependencies of `pb_lib`).
It's a bit messy, one idea would be to have one global variable defining allocator and use it in every executable... But it may be a bit heavy refactor (although cleaner).

pbougue (commented):
Not feeling safe about all dependencies (even if the main in packages should be easy to secure as there are few).
It would be nice to ensure we use the same allocator in all tests also :disappointed:

kinnou02:
It's not required to have every executable use the same allocator, it's better to have test that use the same allocator, but the difference between allocator are mostly visible under load and concurrency.
In fact, we may even want to use different allocator between ed and kraken as their memory usage pattern are worlds apart.

These differences are:
  - latency
  - memory overload and fragmentation
  - ability to release memory

I will update ed fix the linking too, I did focus on kraken previously

buildbotctp:
## Warning 

 Files: 
- `source/type/message.h`

TeXitoi (approved):
@pbougue Are you OK or you want a change? If you're OK, please merge.

pbougue (approved):

PR #2606, opened 2018-11-26
Tyr: add gevent as a dependency

kinnou02:
Monkey patch tyr with gevent to handle request asynchronously.
The monkey patching is done only on the webservice to not impact celery workers.

As always with async it doesn't make thing faster but it increases concurrency, very slow request will only sightly impact fast requests done at the same time.
With threads (the current configuration) the fast requests were suffering from the latency of the slow requests.

Once this is merged we will have to merge https://github.com/CanalTP/navitia_deployment_conf/pull/509

### DIFF
--- a/source/tyr/requirements.txt
+++ b/source/tyr/requirements.txt
@@ -33,3 +33,5 @@ retrying==1.3.3
 ujson==1.35
 jsonschema==2.5.1
 python-dateutil==2.5.3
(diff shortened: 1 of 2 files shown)


nonifier:
Don't we want to patch Jormun as well ?!

kinnou02:
This change require to update deployement: way too much work.

PR #2608, opened 2018-11-27
kraken: fix tests that were failling without tcmalloc

kinnou02:
In these tests we were making a copy of the stop_time_list that was
freed before being used, so...
tcmalloc was hiding the problem since this chunk of memory was only mark
has available


buildbotctp:
## Warning 

 Files: 
- `source/type/message.h`

TeXitoi (approved):

buildbotctp:
## Warning 

 Files: 
- `source/type/message.h`

PR #2609, opened 2018-11-27
Utils: Diff against commit's ancestor

nonifier:
Make sure we diff the current branch with its forking point. 
This avoids finding differences in files that havent't been touched when the forking point isn't up to date with `origin/dev`


buildbotctp:
## Warning 

 Files: 
- `source/type/message.h`

benoit-bst (approved):

kinnou02 (approved):

PR #2610, opened 2018-11-29
Test on add effect

kadhikari:
This PR is only to test the failure of https://github.com/CanalTP/navitia/pull/2601


kadhikari:
This was a test

PR #2613, opened 2018-11-29
kraken: Remove a few use-after-free error

kinnou02:
@nonifier suggested doing some check with https://github.com/google/sanitizers/wiki/AddressSanitizer
Not it's best idea of the year...

Kraken had one in `apply_disruption()` (I'am sure everyone is shocked), I have put a band aid on it...

The second commit fix a few things in the tests.

I have run a kraken with realtime feeding, there was no error nor leak detected.
A lot of tests leak but it isn't a surprise as we allocate a lot of fixtures on the heap without even trying to free it properly. I will try to clean this shit to allow us to run test with the sanitizers on.

### DIFF
--- a/source/kraken/apply_disruption.cpp
+++ b/source/kraken/apply_disruption.cpp
@@ -59,7 +59,7 @@ static nt::VehicleJourney* create_vj_from_old_vj(nt::MetaVehicleJourney* mvj,
                               const std::string& new_vj_uri,
                               nt::RTLevel rt_level,
                               nt::ValidityPattern new_vp,
-                              const std::vector<nt::StopTime>& new_stop_times,
+                              const std::vector<nt::StopTime> new_stop_times,
                               nt::PT_Data& pt_data) {
     auto* company = vj->company;
     auto vehicle_journey_type = vj->vehicle_journey_type;
@@ -72,7 +72,7 @@ static nt::VehicleJourney* create_vj_from_old_vj(nt::MetaVehicleJourney* mvj,
             vj->route,
             std::move(new_stop_times),
             pt_data);
-    vj = nullptr;
+    vj = nullptr;//after create_discrete_vj, the vj can have been deleted
 
     new_vj->company = company;
     new_vj->vehicle_journey_type = vehicle_journey_type;
--- a/source/kraken/apply_disruption.cpp
+++ b/source/kraken/apply_disruption.cpp
@@ -59,7 +59,7 @@ static nt::VehicleJourney* create_vj_from_old_vj(nt::MetaVehicleJourney* mvj,
                               const std::string& new_vj_uri,
                               nt::RTLevel rt_level,
                               nt::ValidityPattern new_vp,
-                              const std::vector<nt::StopTime> new_stop_times,
+                              std::vector<nt::StopTime> new_stop_times,
                               nt::PT_Data& pt_data) {
     auto* company = vj->company;
     auto vehicle_journey_type = vj->vehicle_journey_type;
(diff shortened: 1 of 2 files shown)


source/kraken/apply_disruption.cpp
```diff
@@ -54,6 +54,47 @@ namespace bg = boost::gregorian;
 
 namespace {
 
+static nt::VehicleJourney* create_vj_from_old_vj(nt::MetaVehicleJourney* mvj,
+                              const nt::VehicleJourney* vj,
+                              const std::string& new_vj_uri,
+                              nt::RTLevel rt_level,
+                              nt::ValidityPattern new_vp,
+                              const std::vector<nt::StopTime>& new_stop_times,
```
TeXitoi: you want to take that by value to have a useful `stdd::move` usage.

source/kraken/apply_disruption.cpp
```diff
@@ -54,6 +54,47 @@ namespace bg = boost::gregorian;
...
+    auto odt_message = vj->odt_message;
+    auto vehicle_properties = vj->_vehicle_properties;
+
+    auto* new_vj = mvj->create_discrete_vj(new_vj_uri,
+            rt_level,
+            new_vp,
+            vj->route,
+            std::move(new_stop_times),
+            pt_data);
+    vj = nullptr;
```
TeXitoi: why?

TeXitoi (changes requested):

source/kraken/apply_disruption.cpp
kinnou02: `create_discrete_vj()` also triger the deletion of useless vjs, so when a disruption is applied on an already adapte/realtime vj, the *base* vj is deleted when we create the new one.
To be safe I set `vj = nullptr` to explicitly indicate that we can't use it anymore.

TeXitoi: A comment explaining why it is set to null would be great.

source/kraken/apply_disruption.cpp
```diff
@@ -54,6 +54,47 @@ namespace bg = boost::gregorian;
 
 namespace {
 
+static nt::VehicleJourney* create_vj_from_old_vj(nt::MetaVehicleJourney* mvj,
+                              const nt::VehicleJourney* vj,
+                              const std::string& new_vj_uri,
+                              nt::RTLevel rt_level,
+                              nt::ValidityPattern new_vp,
+                              const std::vector<nt::StopTime> new_stop_times,
```
TeXitoi: It will copy... You have to remove const.

TeXitoi (changes requested):

source/kraken/apply_disruption.cpp
```diff
@@ -442,32 +470,18 @@ struct add_impacts_visitor : public apply_impacts_visitor {
...
+            auto* new_vj = create_vj_from_old_vj(
+                    mvj,
+                    vj,
+                    new_vj_uri,
                     rt_level,
                     new_vp,
-                    vj->route,
                     std::move(new_stop_times),
                     pt_data);
+            vj = nullptr; //after the call to create_vj, vj may have been deleted :(
```
benoit-bst, line 482: But, it is already done here ?
https://github.com/CanalTP/navitia/pull/2613/commits/312b1d6da9bba5f6e197247022ff8fcee326fa73#diff-c234f9d9583465696c5f3311313759d0R75

TeXitoi (approved):

source/kraken/apply_disruption.cpp
TeXitoi: That's local variables

benoit-bst (approved):

PR #2614, opened 2018-11-29
Tyr: fix celery problem

xlqian:
http://docs.celeryq.org/en/latest/userguide/tasks.html#task-synchronous-subtasks

Following by this PR: https://github.com/CanalTP/navitia/pull/2595

### DIFF
--- a/source/tyr/tyr/command/load_data.py
+++ b/source/tyr/tyr/command/load_data.py
@@ -52,9 +52,7 @@ class LoadDataCommand(Command):
         data_dirs = data_dir.split(',')
         logging.info("loading data dir : {}".format(data_dirs))
 
-        futur_res = load_data.delay(instance.id, data_dirs)
-
         # this api is only used in artemis and artemis wants to know when this
         # task is finished, so the easiest way for the moment is to wait for
         # the answer
-        futur_res.get()
+        load_data(instance.id, data_dirs)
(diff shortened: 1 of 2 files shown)


pbougue (commented):
Well, I trust you on that one :sweat_smile: 
But how about https://github.com/CanalTP/navitia/blob/dev/source/tyr/tyr/tasks.py#L296 and https://github.com/CanalTP/navitia/blob/dev/source/tyr/tyr/tasks.py#L314 ?

kinnou02 (approved):

xlqian:
@pbougue fixed, thx

mbounabi (approved):

source/tyr/tyr/command/load_data.py
```diff
@@ -52,9 +52,7 @@ def run(self, instance_name, data_dir):
         data_dirs = data_dir.split(',')
         logging.info("loading data dir : {}".format(data_dirs))
 
-        futur_res = load_data.delay(instance.id, data_dirs)
-
         # this api is only used in artemis and artemis wants to know when this
         # task is finished, so the easiest way for the moment is to wait for
         # the answer
-        futur_res.get()
+        load_data(instance.id, data_dirs)
```
xlqian, line 58: If you call the function by `delay`, it will launch a task (aka, send a task to a worker and the worker will do the job in async), it'll cause the case  `launching synchronous subtasks`, by calling it directly, this case is avoided and the function call is blocking which is what we want

pbougue (approved):

PR #2618, opened 2018-12-05
Jormungandr: Disable uwsgi timer waiting for proper fix

kinnou02:
uwsgi has a undocumented limit on the number of timer than can be
created in a uwsgi "context", set to 64, after that creating the timer
will raise an exception that terminate all uwsgi processes.
The current implementation of transient socket create one timer per
uwsgi process.
We learned the hard way that this limit exist when our uwsgi in
production started to die at startup...

This doesn't fix the problems, it only disable uwsgi timers to let me
some time for finding a real solution.

This will be hotfixed in 2.68.2

### DIFF
--- a/source/jormungandr/jormungandr/instance_manager.py
+++ b/source/jormungandr/jormungandr/instance_manager.py
@@ -183,20 +183,19 @@ class InstanceManager(object):
                 break
 
     def init_socket_reaper(self):
-        try:
-            from uwsgidecorators import timer
+        # Disable uwsgi for now
+        # try:
+        #    from uwsgidecorators import timer
 
-            logging.getLogger(__name__).info("spawning a socket reaper with  uwsgi timer")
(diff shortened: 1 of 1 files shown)


xlqian (approved):

source/jormungandr/jormungandr/instance_manager.py
```diff
@@ -183,20 +183,19 @@ def init_kraken_instances(self):
                 break
 
     def init_socket_reaper(self):
-        try:
-            from uwsgidecorators import timer
+        # Disable uwsgi for now
```
TeXitoi: Disable uwsgi **timers** for now

TeXitoi (dismissed):

pbougue (approved):
Doing a param is not an option for now?
It's fine anyway under those circumstances :-/

mbounabi (approved):

PR #2622, opened 2018-12-07
Jomungandr: Rollback to celery and kombu 4.0.2 

kinnou02:
Kombu 4.1 and later are bugged: if rabbitmq is down they will block until the end of time...
celery/kombu#902

The last two commit are some minor fixes that don't really have anything to do with this and that should both have their own PR, but at least it will help jenkins a bit. So read by commit!

### DIFF
--- a/source/tyr/tyr/tasks.py
+++ b/source/tyr/tyr/tasks.py
@@ -205,7 +205,13 @@ def send_to_mimir(instance, filename):
 def update_data():
     for instance in models.Instance.query_existing().all():
         current_app.logger.debug("Update data of : {}".format(instance.name))
-        instance_config = load_instance_config(instance.name)
+        instance_config = None
+        try:
+            instance_config = load_instance_config(instance.name)
+        except:
+            current_app.logger.exception("impossible to load instance configuration for %s", instance.name)
+            # Do not stop the task if only one instance is missing
+            continue
         files = glob.glob(instance_config.source_directory + "/*")
         if files:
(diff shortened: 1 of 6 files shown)


source/tyr/tyr/tasks.py
```diff
@@ -205,7 +205,13 @@ def send_to_mimir(instance, filename):
 def update_data():
     for instance in models.Instance.query_existing().all():
         current_app.logger.debug("Update data of : {}".format(instance.name))
-        instance_config = load_instance_config(instance.name)
+        instance_config = None
+        try:
+            instance_config = load_instance_config(instance.name)
+        except:
+            current_app.logger.exception("impossible to load instance configuration for %s", instance.name)
+            # Do not stop the task if only one instance is missing
```
nonifier, line 213: :+1:

nonifier (approved):

source/tyr/tyr/tasks.py
```diff
@@ -205,7 +205,13 @@ def send_to_mimir(instance, filename):
...
+            continue
         files = glob.glob(instance_config.source_directory + "/*")
```
pbougue, line 215: Don't you want to put that in the try also: skip that instance's import but still save the others?
> gnaaa I'm tired: the continue does that, it's the same :zipper_mouth_face:

pbougue (approved):

TeXitoi (approved):

PR #2625, opened 2018-12-11
jormun: refactoring of zmq socket reaper

kinnou02:
This PR aim to fix the problem with the current implementaiton of the zmq socket reaper: it doesn't work in production...

There is a hard limit on the number of uwsgi timer that can be instanced in uwsgi: 64, when trying to create more uwsgi will raise a `ValueError` that will kill jormungandr.
Also uwsgi timer aren't perfect they are dispatched on workers randomly and not by using round robbin, so it's possible that a worker never get the chance to handle a signal while another will get it each time. (This comportment can also be observed with request handling)

This PR make multiple change to improve the previous implementation:

  1) Always start a gevent greenlet that should run every reaper interval
     to close idle sockets.

  2) If we are running in uwsgi without the gevent event loop enable
     create an uwsgi signal handler to close the sockets that will be
     triggered by an uwsgi timer

  3) If we can't create an uwsgi timer it's not that bad, so jormun
     continue to run in that case.

Uwsgi timers are a failsafe: By default uwsgi doesn't use the gevent event loop, so an idle worker can't run the socket reaper greenlet periodically, it's only run when waiting for an external service like kraken.

At most 64 timers will be created it there is more processes they can still handle a signal, so at most 64 processes will close their socket every reaper_interval.

### DIFF
--- a/source/jormungandr/jormungandr/instance_manager.py
+++ b/source/jormungandr/jormungandr/instance_manager.py
@@ -214,7 +214,9 @@ class InstanceManager(object):
             # workers that didn't create a timer can still run the signal handler
             # if uwsgi dispatch the signal to them
             # signal are dispatched randomly to workers (not round robbin :()
-            logging.getLogger(__name__).info("No more uwsgi timer available, only gevent will be used")
+            logging.getLogger(__name__).info(
+                "No more uwsgi timer available or not running in uwsgi, only gevent will be used"
+            )
 
     def reap_sockets(self):
         for instance in self.instances.values():
--- a/source/jormungandr/jormungandr/instance_manager.py
+++ b/source/jormungandr/jormungandr/instance_manager.py
@@ -183,26 +183,48 @@ class InstanceManager(object):
                 break
 
     def init_socket_reaper(self):
-        # Disable uwsgi timers for now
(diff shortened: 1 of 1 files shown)


source/jormungandr/jormungandr/instance_manager.py
```diff
@@ -183,26 +183,48 @@ def init_kraken_instances(self):
...
+            uwsgi.register_signal(1, 'active-workers', reaper_timer)
+            # Add a timer that trigger this signal every reaper_interval second
+            uwsgi.add_timer(1, self.reaper_interval)
+        except (ImportError, ValueError):
+            # ImportError is triggered if we aren't in uwsgi
+            # ValueError is raise if there is no more timer availlable: only 64 timers can be created
+            # workers that didn't create a timer can still run the signal handler
+            # if uwsgi dispatch the signal to them
+            # signal are dispatched randomly to workers (not round robbin :()
+            logging.getLogger(__name__).info("No more uwsgi timer available, only gevent will be used")
```
TeXitoi: The log might not be so true: if we aren't in uwsgi, the log is wrong.

TeXitoi (approved):

PR #2620, opened 2018-12-05
kraken: new StopTimeEventStatus with detour implemented

kadhikari:
This part manage a new StopTimeEventStatus to handle different status of a stop_time of impacted_objects in navitia.

- Kaken accepts new status like DELETED_FOR_DETOUR and ADDED_FOR_DETOUR 

- A new attribut 'is_detour' is added in "impacted_stops" and used in the response navitia.

For details see: https://jira.kisio.org/browse/NAVP-1084

PR #2626, opened 2018-12-11
kraken: Manages added_for_detour in advance

kadhikari:
In a trip_update, 

1. we have a stop_time replaced by another stop_time with arrival or departure time < to  the one deleted.

2. the added stop_time appears after the  deleted one in gtfs-rt.

Kraken rejected the gtfs-rt.

Correction: For a deleted stop_time, no verification is made on arrival and departure time with the previous stop_time.

Blocked by: https://github.com/CanalTP/navitia/pull/2620
Review needed from : https://github.com/CanalTP/navitia/pull/2626/commits/2f88a0635679426105c5f9cff51efa2dfba16590

### DIFF
--- a/source/jormungandr/tests/kirin_realtime_tests.py
+++ b/source/jormungandr/tests/kirin_realtime_tests.py
@@ -1387,6 +1387,7 @@ class TestKirinStopTimeOnDetourAndArrivesBeforeDeletedAtTheEnd(MockKirinDisrupti
                     arrival=tstamp("20120614T080102"),
                     departure=tstamp("20120614T080102"),
                     arrival_skipped=True,
+                    departure_skipped=True,
                     is_detour=True,
                     message='deleted for detour',
                 ),
@@ -1420,7 +1421,7 @@ class TestKirinStopTimeOnDetourAndArrivesBeforeDeletedAtTheEnd(MockKirinDisrupti
 
         assert bool(impacted_stops[1]['is_detour']) is True
         assert impacted_stops[1]['cause'] == 'deleted for detour'
-        assert impacted_stops[1]['departure_status'] == 'unchanged'
+        assert impacted_stops[1]['departure_status'] == 'deleted'
         assert impacted_stops[1]['arrival_status'] == 'deleted'
 
         assert bool(impacted_stops[2]['is_detour']) is True
@@ -1428,7 +1429,7 @@ class TestKirinStopTimeOnDetourAndArrivesBeforeDeletedAtTheEnd(MockKirinDisrupti
         assert impacted_stops[2]['departure_status'] == 'added'
         assert impacted_stops[2]['arrival_status'] == 'added'
 
-        B_C_query = "journeys?from={from_coord}&to={to_coord}&datetime={datetime}&direct_path=none".format(
+        B_C_query = "journeys?from={from_coord}&to={to_coord}&datetime={datetime}".format(
             from_coord='stop_point:stopB', to_coord='stop_point:stopC', datetime='20120614T080000'
         )
 
--- a/source/jormungandr/tests/kirin_realtime_tests.py
+++ b/source/jormungandr/tests/kirin_realtime_tests.py
@@ -1360,8 +1360,7 @@ class TestKirinStopTimeOnDetourAndArrivesBeforeDeletedAtTheEnd(MockKirinDisrupti
         """
         1. create a new_stop_time at C to replace existing one at A so that we have A deleted_for_detour
         and C added_for_detour with arrival time < to arrival time of A (deleted)
-        2. Kraken rejects this disruption and is a BUG. Adjust this test after correction in kraken
-        For details see: https://jira.kisio.org/browse/NAVP-1118
+        2. Kraken accepts this disruption
         """
         disruptions_before = self.query_region('disruptions?_current_datetime=20120614T080000')
         nb_disruptions_before = len(disruptions_before['disruptions'])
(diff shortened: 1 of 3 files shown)


source/kraken/realtime.cpp
```diff
@@ -78,6 +80,16 @@ static bool is_handleable(const transit_realtime::TripUpdate& trip_update){
     return false;
 }
 
+static bool _is_deleted(const transit_realtime::TripUpdate_StopTimeEvent& event) {
```
TeXitoi: `_` at the begining of the function name is a python idiom, not a C++ none.

source/kraken/realtime.cpp
```diff
@@ -194,18 +220,32 @@ get_relationship(const transit_realtime::TripUpdate_StopTimeEvent& event,
...
+        case kirin::StopTimeEventStatus::DELETED:
+            return nt::disruption::StopTimeUpdate::Status::DELETED;
+        case kirin::StopTimeEventStatus::DELETED_FOR_DETOUR:
+            return nt::disruption::StopTimeUpdate::Status::DELETED_FOR_DETOUR;
+        case kirin::StopTimeEventStatus::ADDED_FOR_DETOUR:
+            return nt::disruption::StopTimeUpdate::Status::ADDED_FOR_DETOUR;
+        default: break;
+        }
+    } else {
+        //TODO: to be deleted once this version is deployed in prod.
```
TeXitoi: `event.GetExtension(kirin::stop_time_event_status);` is new? if not, I think we can remove that, but I'm not sure.

TeXitoi (approved):

source/kraken/realtime.cpp
kadhikari: It's new and the else part will be deleted later

benoit-bst (approved):

TeXitoi (approved):

source/kraken/realtime.cpp
```diff
@@ -120,10 +136,15 @@ static std::ostream& operator<<(std::ostream& s, const nt::StopTime& st) {
  */
 static bool check_disruption(const nt::disruption::Disruption& disruption) {
     auto log = log4cplus::Logger::getInstance("realtime");
+    using nt::disruption::StopTimeUpdate;
     for (const auto& impact: disruption.get_impacts()) {
         boost::optional<const nt::StopTime&> last_st;
         for (const auto& stu: impact->aux_info.stop_times) {
             const auto& st = stu.stop_time;
+            bool deleted = (in(stu.arrival_status,
```
pbougue: It would be nicer with a function here also.

source/kraken/realtime.cpp
```diff
@@ -89,8 +99,11 @@ static bool check_trip_update(const transit_realtime::TripUpdate& trip_update) {
         for (const auto& st: trip_update.stop_time_update()) {
             uint32_t arrival_time = st.arrival().time();
             uint32_t departure_time = st.departure().time();
+            bool deleted = is_deleted(st.arrival()) ||
```
pbougue: We should check is_deleted(arrival) when dealing with arrival consistency, and separately is_deleted(departure) when dealing with departure consistency (so below).
And probably keep a `last_stop_event_time` to use for comparison.

source/kraken/realtime.cpp
```diff
@@ -80,6 +80,16 @@ static bool is_handleable(const transit_realtime::TripUpdate& trip_update){
     return false;
 }
 
+static bool is_deleted(const transit_realtime::TripUpdate_StopTimeEvent& event) {
+    if (event.HasExtension(kirin::stop_time_event_status) &&
```
pbougue: Minor: You can return the test directly to shorten code.
To be sure: no need to check the case without extension as it's gonna disapear soon?

source/kraken/realtime.cpp
```diff
@@ -120,10 +136,15 @@ static std::ostream& operator<<(std::ostream& s, const nt::StopTime& st) {
  */
 static bool check_disruption(const nt::disruption::Disruption& disruption) {
     auto log = log4cplus::Logger::getInstance("realtime");
+    using nt::disruption::StopTimeUpdate;
     for (const auto& impact: disruption.get_impacts()) {
         boost::optional<const nt::StopTime&> last_st;
         for (const auto& stu: impact->aux_info.stop_times) {
             const auto& st = stu.stop_time;
+            bool deleted = (in(stu.arrival_status,
+            {StopTimeUpdate::Status::DELETED, StopTimeUpdate::Status::DELETED_FOR_DETOUR}) ||
```
pbougue: Same here, I think it's better to separate stop_time events as we check consistency separately.

source/kraken/realtime.cpp
```diff
@@ -120,10 +136,15 @@ static std::ostream& operator<<(std::ostream& s, const nt::StopTime& st) {
  */
 static bool check_disruption(const nt::disruption::Disruption& disruption) {
```
pbougue, line 135: Probably not in this PR: we have 2 checks on consistency... Can't we group them?
The same goes on Kirin, and it does not seem easy to do there.

source/jormungandr/tests/kirin_realtime_tests.py
```diff
@@ -1407,7 +1406,51 @@ def test_stop_time_with_detour_and_arrival_before_deleted_at_the_end(self):
...
+        assert impacted_stops[1]['cause'] == 'deleted for detour'
+        assert impacted_stops[1]['departure_status'] == 'unchanged'
+        assert impacted_stops[1]['arrival_status'] == 'deleted'
+
+        assert bool(impacted_stops[2]['is_detour']) is True
+        assert impacted_stops[2]['cause'] == 'added for detour'
+        assert impacted_stops[2]['departure_status'] == 'added'
+        assert impacted_stops[2]['arrival_status'] == 'added'
+
+        B_C_query = "journeys?from={from_coord}&to={to_coord}&datetime={datetime}".format(
```
pbougue, line 1432: minor: you can probably add `&direct_path=none` to filter the non_pt_walk out of responses.
> not handled now, but it doesn't work

source/kraken/realtime.cpp
kadhikari: Since we don't compare same enum type in "is_deleted" i don't use a new methode  here

source/kraken/realtime.cpp
kadhikari: I discussed with Guillaume before modifying the code and we both agreed on this solution after reading your remark

source/kraken/realtime.cpp
kadhikari: Will do in the next PR no?

source/kraken/realtime.cpp
```diff
@@ -89,8 +98,9 @@ static bool check_trip_update(const transit_realtime::TripUpdate& trip_update) {
         for (const auto& st: trip_update.stop_time_update()) {
             uint32_t arrival_time = st.arrival().time();
             uint32_t departure_time = st.departure().time();
+
             if (last_st_dep != std::numeric_limits<uint32_t>::max()
-                    && last_st_dep > arrival_time) {
+                    && last_st_dep > arrival_time && !is_deleted(st.arrival())) {
```
pbougue: You need to always store the last_stop_event_time, something like (code may be prettified):
```cpp
if (! is_deleted(st.arrival())) {
    if (last_stop_event_time != std::numeric_limits<uint32_t>::max() && last_stop_event_time > arrival_time) {
        LOG...
        return false;
    last_stop_event_time = arrival_time;
}
if (! is_deleted(st.departure())) {
    if (last_stop_event_time != std::numeric_limits<uint32_t>::max() && last_stop_event_time > departure_time) {
        LOG...
        return false;
    last_stop_event_time = departure_time;
}
```
So here last_st_dep disappear (renamed for `last_stop_event_time` as the meaning is a bit wider).
Otherwise you might miss inconsistencies.

source/kraken/realtime.cpp
```diff
@@ -120,12 +133,19 @@ static std::ostream& operator<<(std::ostream& s, const nt::StopTime& st) {
  */
 static bool check_disruption(const nt::disruption::Disruption& disruption) {
     auto log = log4cplus::Logger::getInstance("realtime");
+    using nt::disruption::StopTimeUpdate;
     for (const auto& impact: disruption.get_impacts()) {
         boost::optional<const nt::StopTime&> last_st;
         for (const auto& stu: impact->aux_info.stop_times) {
             const auto& st = stu.stop_time;
+            bool dep_deleted = in(stu.departure_status,
```
pbougue: Same here, to avoid missing inconsistencies, you should do (code may be prettified):
```cpp
bool arr_deleted = in(stu.arrival_status,
                      {StopTimeUpdate::Status::DELETED, StopTimeUpdate::Status::DELETED_FOR_DETOUR});
if (! arr_deleted) {
    if (last_stop_event_time && last_stop_event_time > st.arrival_time) {
        LOG...
        return false;
    }
    last_stop_event_time = st.arrival_time
}
bool dep_deleted = in(stu.departure_status,
                      {StopTimeUpdate::Status::DELETED, StopTimeUpdate::Status::DELETED_FOR_DETOUR});
if (! dep_deleted) {
    if (last_stop_event_time && last_stop_event_time > st.departure_time) {
        LOG...
        return false;
    }
    last_stop_event_time = st.departure_time
}
```
So here last_st is replaced by an option on a datetime renamed `last_stop_event_time` but here is the point.

source/jormungandr/tests/kirin_realtime_tests.py
```diff
@@ -1407,7 +1406,51 @@ def test_stop_time_with_detour_and_arrival_before_deleted_at_the_end(self):
...
+        B_C_query = "journeys?from={from_coord}&to={to_coord}&datetime={datetime}&direct_path=none".format(
+            from_coord='stop_point:stopB', to_coord='stop_point:stopC', datetime='20120614T080000'
+        )
+
+        # Query with data_freshness=base_schedule
+        base_journey_query = B_C_query + "&data_freshness=base_schedule&_current_datetime=20120614T080000"
+
+        # There is no public transport from B to C
+        response = self.query_region(base_journey_query)
+        assert len(response['journeys']) == 1
```
pbougue, line 1441: With direct_path=none you should have `= 0` here (at least that was my point)

source/jormungandr/tests/kirin_realtime_tests.py
```diff
@@ -1407,7 +1406,51 @@ def test_stop_time_with_detour_and_arrival_before_deleted_at_the_end(self):
...
+        assert response['journeys'][0]['type'] == 'non_pt_walk'
+        assert 'data_freshness' not in response['journeys'][0]['sections'][0]  # means it's base_schedule
+
+        # Query with data_freshness=realtime
+        base_journey_query = B_C_query + "&data_freshness=realtime&_current_datetime=20120614T080000"
+
+        # There is a public transport from B to C with realtime
+        response = self.query_region(base_journey_query)
+        assert len(response['journeys']) == 2
```
pbougue, line 1450: Same, here you should have only 1 journey now.

source/jormungandr/tests/kirin_realtime_tests.py
kadhikari: i think the parameter is wrong. i will test this

source/kraken/realtime.cpp
kadhikari: done

source/kraken/realtime.cpp
kadhikari: done

pbougue (approved):

TeXitoi (approved):

PR #2616, opened 2018-12-03
Jormun: removal of destineo and default scenario 

kinnou02:
This is a great day: destineo is officialy closed and we can start to
delete all of this specific code.
:tada: 

Before merging we need to:
  - [x] wait a little bit to be safe
  - [x] remove destineo tests from artemis

PR #2629, opened 2018-12-12
Distributed: Add filter for finalized journeys

xlqian:
Some journeys are not filtered because with the fake crowflies, there was no enough detailed information for the filter, especially when it concerns a car fallback returned by Kraken(car + park + walking...) 

We re-apply some filters which need more details than other filters...

### DIFF
--- a/source/jormungandr/jormungandr/scenarios/distributed.py
+++ b/source/jormungandr/jormungandr/scenarios/distributed.py
@@ -289,6 +289,13 @@ class Scenario(new_default.Scenario):
         try:
             with FutureManager() as future_manager:
                 self._scenario.finalise_journeys(future_manager, request, responses, context, instance, is_debug)
+
+                from jormungandr.scenarios import journey_filter
+
+                # At this point, we should have every details on the journeys.
+                # We refilter again(again and again...)
+                journey_filter.filter_detailed_journeys(responses, request)
+
         except Exception as e:
             final_e = FinaliseException(e)
             return [final_e.get()]
--- a/source/jormungandr/jormungandr/scenarios/journey_filter.py
+++ b/source/jormungandr/jormungandr/scenarios/journey_filter.py
@@ -378,15 +378,10 @@ def similar_journeys_generator(journey, pt_functor):
     for idx, s in enumerate(journey.sections):
         if s.type == response_pb2.PUBLIC_TRANSPORT:
             yield pt_functor(s)
-        elif s.type == response_pb2.STREET_NETWORK:
-            # special case, we don't want to consider the walking section after/before parking a car
-            # so CAR / PARK / WALK / PT is equivalent to CAR / PARK / PT
-            if is_walk_after_parking(journey, idx):
-                continue
+        elif s.type == response_pb2.STREET_NETWORK and is_walk_after_parking(journey, idx):
+            continue
+        elif s.type in (response_pb2.STREET_NETWORK, response_pb2.CROW_FLY):
             yield 'sn:%s' % s.street_network.mode
-        elif s.type == response_pb2.CROW_FLY and s.street_network.mode == response_pb2.Ridesharing:
-            # crowfly in ridesharing must be distinct from crowfly in walking
-            yield "sn:{}".format(s.street_network.mode)
 
 
 def similar_journeys_vj_generator(journey):
--- a/source/jormungandr/jormungandr/scenarios/journey_filter.py
+++ b/source/jormungandr/jormungandr/scenarios/journey_filter.py
@@ -525,6 +525,26 @@ def apply_final_journey_filters(response_list, instance, request):
     _filter_too_much_connections(journeys, instance, request)
 
 
+def filter_detailed_journeys(responses, request):
+    journey_generator = get_qualified_journeys
+    if request.get('debug', False):
+        journey_generator = get_all_journeys
+
+    journeys = journey_generator(responses)
+
+    min_bike = request.get('_min_bike', None)
+    min_car = request.get('_min_car', None)
+    orig_modes = request.get('origin_mode', [])
+    dest_modes = request.get('destination_mode', [])
+
+    too_heavy_journey_filter = FilterTooShortHeavyJourneys(
+        min_bike=min_bike, min_car=min_car, orig_modes=orig_modes, dest_modes=dest_modes
+    )
+    f_wrapped = filter_wrapper(is_debug=request.get('debug', False), filter_obj=too_heavy_journey_filter)
+
+    [f_wrapped(j) for j in journeys]
+
+
 def _get_worst_similar(j1, j2, request):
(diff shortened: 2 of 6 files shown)


xlqian:
I'm working on the test...

source/jormungandr/jormungandr/scenarios/distributed.py
```diff
@@ -289,6 +289,13 @@ def finalise_journeys(self, request, responses, context, instance, is_debug):
         try:
             with FutureManager() as future_manager:
                 self._scenario.finalise_journeys(future_manager, request, responses, context, instance, is_debug)
+
+                from jormungandr.scenarios import journey_filter
+
+                # At this point, we should have every details on the journeys.
+                # We refilter again(again and again...)
+                journey_filter.filter_detailed_journeys(responses, request)
```
nonifier, line 297: Why can't we have this after `call_kraken` with the other filters ? (like in `def filter_journeys(responses, new_resp, instance, api_request):` for instance)

xlqian: nein, filter_journeys is called before the details are calculated (i'm not sure i understand your comment well, let's talk about it face to face)

source/jormungandr/jormungandr/scenarios/journey_filter.py
```diff
@@ -525,6 +525,26 @@ def apply_final_journey_filters(response_list, instance, request):
...
+    journeys = journey_generator(responses)
+
+    min_bike = request.get('_min_bike', None)
+    min_car = request.get('_min_car', None)
+    orig_modes = request.get('origin_mode', [])
+    dest_modes = request.get('destination_mode', [])
+
+    too_heavy_journey_filter = FilterTooShortHeavyJourneys(
+        min_bike=min_bike, min_car=min_car, orig_modes=orig_modes, dest_modes=dest_modes
+    )
```
TeXitoi, line 537: Should be removed from `filter_journeys`, no?

TeXitoi (approved):

source/jormungandr/jormungandr/scenarios/journey_filter.py
xlqian: It can also remain here. It'll optimize properly when it's Valhalla

source/jormungandr/tests/routing_tests.py
```diff
@@ -99,6 +99,12 @@ def test_shared_section(self):
         """
         pass
 
+    def test_journeys_too_short_heavy_mode_fallback_filter(self):
+        """
+        This new feature is not supported in default
```
pbougue, line 104: This may conflict with PR removing default and destineo's scenario : #2616 
No idea what's the best to deal with that.

source/jormungandr/jormungandr/scenarios/journey_filter.py
```diff
@@ -378,15 +378,10 @@ def similar_journeys_generator(journey, pt_functor):
     for idx, s in enumerate(journey.sections):
         if s.type == response_pb2.PUBLIC_TRANSPORT:
             yield pt_functor(s)
-        elif s.type == response_pb2.STREET_NETWORK:
-            # special case, we don't want to consider the walking section after/before parking a car
-            # so CAR / PARK / WALK / PT is equivalent to CAR / PARK / PT
-            if is_walk_after_parking(journey, idx):
-                continue
+        elif s.type == response_pb2.STREET_NETWORK and is_walk_after_parking(journey, idx):
```
pbougue, line 381: Seems OK, but just curious why you add it now?

source/jormungandr/jormungandr/scenarios/journey_filter.py
pbougue: I vote letting it in `filter_journeys`, early one-by-one filter is probably better in perf considering mean and worst case.
I'm just wondering if we can mutualize code more (param retrieving + journey generator)

source/jormungandr/tests/journey_common_tests.py
```diff
@@ -1080,6 +1080,39 @@ def get_length(s):
...
+        assert all('to_delete' not in j['tags'] for j in response['journeys'])
+
+        car_fallback_pt_journey = next((j for j in response['journeys'] if j['type'] == 'car'), None)
+
+        assert car_fallback_pt_journey
+        assert car_fallback_pt_journey['sections'][0]['mode'] == 'car'
+
+        car_fallback_duration = car_fallback_pt_journey['sections'][0]['duration']
+
+        query = template.format(min_car=car_fallback_duration + 1)
```
pbougue, line 1106: :+1:

source/jormungandr/tests/routing_tests_experimental.py
```diff
@@ -76,7 +76,7 @@ def test_first_and_last_section_multi_modes(self):
         # journey count = 18 / direct_path_call_count = 26 / routing_matrix_call_count = 20
         # get_directpath_count_by_mode(response, 'walking') == 5
         # get_directpath_count_by_mode(response, 'bike') == 5
-        assert len(response["journeys"]) == 10
+        assert len(response["journeys"]) == 13
```
pbougue, line 79: I don't get why there are more journeys here.

pbougue (commented):
Looks good, only doubt about the number of journeys.

source/jormungandr/jormungandr/scenarios/journey_filter.py
xlqian: I need to handle the case where the section type is CROWFLY. The modif actually fix a bug.
It's a little bit tricky.

let's say you have two journeys:

Bike - Bus - Bike
Walking - Bus - Walking

the simlar filter before fix will see:

Crowfly- Bus - Crowfly
Crowfly - Bus - Crowfly

Then one of the journeys will be deleted which is incorrect

I just take s.mode into consideration when s.type is crowfly and a little refactoring

source/jormungandr/tests/routing_tests_experimental.py
xlqian: aha i took me a lot of time to understand too.
 Notice that there is a `min_nb_jouneys=10` in the request and there is a condition that stops the search when there is no new qualified journeys(https://github.com/CanalTP/navitia/blob/dev/source/jormungandr/jormungandr/scenarios/new_default.py#L984)

what happened before the fix is, with the bad similar journey filter condition, the loop is stopped at 5th iteration because it deleted some journeys that shouldn't be deleted. The new filter condition will delete less journeys and change the loop number(7 iterations) then more journeys are found.

source/jormungandr/tests/routing_tests.py
nonifier: First come, first serve ;) And this is more critical than destineo's removal !

nonifier (approved):

PR #2632, opened 2018-12-13
Distributed: update distances also for pure crowflies

pbougue:
This also improves the way fallback **distances** and **durations** are added for street-network fallbacks.

TODO:
- [x] find a way to test that
- [ ] merge artemis reference PR after this one: https://github.com/CanalTP/artemis_references/pull/148

(you can read by commit)

### DIFF
--- a/source/jormungandr/jormungandr/scenarios/helper_classes/helper_utils.py
+++ b/source/jormungandr/jormungandr/scenarios/helper_classes/helper_utils.py
@@ -333,6 +333,15 @@ def _build_fallback(
                     pt_journey.distances.bike += fallback_dp.journeys[0].distances.bike
                     pt_journey.distances.car += fallback_dp.journeys[0].distances.car
                     pt_journey.distances.ridesharing += fallback_dp.journeys[0].distances.ridesharing
+            # if it's only a crowfly fallback, update distance if it's not a teleport
+            elif hasattr(pt_journey.distances, mode):
+                if fallback_logic.represent_start():
+                    crowfly_section = pt_journey.sections[0]
+                else:
+                    crowfly_section = pt_journey.sections[-1]
+                if crowfly_section.duration:
+                    setattr(pt_journey.distances, mode,
+                            (getattr(pt_journey.distances, mode) + crowfly_section.length))
 
     fallback_logic.set_journey_bound_datetime(pt_journey)
 
--- a/source/jormungandr/jormungandr/scenarios/helper_classes/helper_utils.py
+++ b/source/jormungandr/jormungandr/scenarios/helper_classes/helper_utils.py
@@ -147,10 +147,6 @@ def _update_fallback_sections(pt_journey, fallback_dp, fallback_period_extremity
     else:
         section_to_update = pt_journey.sections[0]
 
-    if hasattr(pt_journey.distances, mode):
-        total_fallback_length = sum(s.length for s in fallback_sections)
-        setattr(pt_journey.distances, mode, (getattr(pt_journey.distances, mode) + total_fallback_length))
-
     pt_journey.sections.remove(section_to_update)
     pt_journey.sections.extend(fallback_sections)
     pt_journey.sections.sort(SectionSorter())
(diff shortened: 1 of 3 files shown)


TeXitoi (approved):

xlqian:
Take a look at this line: https://github.com/CanalTP/navitia/blob/dev/source/jormungandr/jormungandr/scenarios/helper_classes/helper_utils.py#L73

It's the same type of problem but for `durations`.  We assume the duration of the fake crowfly is computed only for the given mode which is not exactly true when the mode is `car` and the matrix is computed by kraken (car + park + walking)... the general `durations`(`pt_journey.durations`) should also be updated as you do in this PR. 

just take this comment as a reminder, we can fix that in an another PR

xlqian (approved):

source/jormungandr/jormungandr/scenarios/helper_classes/helper_utils.py
```diff
@@ -327,7 +325,26 @@ def _build_fallback(
             if not _is_crowfly_needed(
                 pt_obj.uri, fallback_durations, accessibles_by_crowfly.crowfly, fallback_dp
             ):
-                _update_fallback_sections(pt_journey, fallback_dp, fallback_period_extremity, mode)
+                _update_fallback_sections(pt_journey, fallback_dp, fallback_period_extremity)
+
+                # update distances if it's a proper computed streetnetwork fallback
+                if fallback_dp and fallback_dp.journeys:
+                    pt_journey.distances.walking += fallback_dp.journeys[0].distances.walking
```
xlqian: i'm just curious...
can you test with this code?
```python
from operator import attrgetter, iadd

for mode in ['walking', 'bike', 'car', 'ridesharing']:
    getter = attrgetter('distances.{}'.format(mode))
    iadd(getter(pt_journey), getter(fallback_dp.journeys[0])
```

xlqian: nope... it won't work... my bad

TeXitoi (approved):

source/jormungandr/jormungandr/scenarios/helper_classes/helper_utils.py
```diff
@@ -167,13 +156,15 @@ def _extend_journey(pt_journey, fallback_dp, fallback_period_extremity):
     pt_journey.durations.walking += aligned_fallback.journeys[0].durations.walking
     pt_journey.durations.bike += aligned_fallback.journeys[0].durations.bike
     pt_journey.durations.car += aligned_fallback.journeys[0].durations.car
+    pt_journey.durations.ridesharing += aligned_fallback.journeys[0].durations.ridesharing
```
nonifier: It would be more maintainable to iterate over a list of modes. 
Maybe something along those lines:

```python
for mode in list_off_all_modes: # bike, car, walking etc...
    aligned_duration = getattr(aligned_fallback.journeys[0].durations, mode)
    duration = getattr(pt_journey.durations, mode)
    setattr(pt_journey.durations, mode, duration + aligned_duration)
```

So that if we add a new mode later (eg. scooter), we don't have to remember to add a new line here.

pbougue: Thanks!
I did something that looks better, but not sure it's all you had in mind.

We iterate on protobuf fields, so there is no list linked to that... And iterating on attributes is also hazardous (if we ever add an enum for example).

source/jormungandr/jormungandr/scenarios/helper_classes/helper_utils.py
```diff
@@ -327,7 +290,38 @@ def _build_fallback(
             if not _is_crowfly_needed(
                 pt_obj.uri, fallback_durations, accessibles_by_crowfly.crowfly, fallback_dp
             ):
-                _update_fallback_sections(pt_journey, fallback_dp, fallback_period_extremity, mode)
+                _update_fallback_sections(pt_journey, fallback_dp, fallback_period_extremity)
+
+                # update distances and durations by mode if it's a proper computed streetnetwork fallback
+                if fallback_dp and fallback_dp.journeys:
+                    all_modes = ['walking', 'bike', 'car', 'ridesharing']
```
nonifier, line 297: My point was to have a single place where we define all the different street-network's modes once for all. As of now, there is no single source of truth if you want to get those modes. 

I'm not asking to do the change now (it's a much bigger change), but I'm stressing out the point.

nonifier (approved):

PR #2633, opened 2018-12-17
Coherence of objects in a node of two sections

kadhikari:
Concerned ticket: https://jira.kisio.org/browse/NAVP-981


TeXitoi (changes requested):
We can't do that, It can break integrations by removing fields that where always present before.

TeXitoi:
That's surprising that Alex asked for that.

pbougue (commented):
Adding `do_not_merge` as we need to discuss what's needed it a bit with @kinnou02 , as I mentionned in JIRA.

kinnou02:
There is no good choice here, let's say we don't do anything

PR #2637, opened 2018-12-27
jormun: stop reporting bragi error as newrelic exception

kinnou02:
When bragi is a bit slow there is a lot of timeout on the reverse done
for /journeys, currently these timeout are reported to newrelic as
exceptions even if they do not cause the response to be on error.
This behavior is a bit misleading for operation people that see 20-30%
of error in newrelic while in fact it's just that the reverse is
failing: it's not that bad.

This PR create a new type of events in newrelic to track
`GeocodeJsonUnavailable` errors (timeouts, 503, etc...), other kind
of errors (`GeocodeJsonError`) are still reported as exception.

The newrelic event is `autocomplete_status`, is has the following
properties:
  - status: `ok`/`unavailable`/`error`
  - type: geocodejson or kraken
  - cause: exception message in case of status being `unavailable` or `error`

### DIFF
--- a/source/jormungandr/jormungandr/autocomplete/abstract_autocomplete.py
+++ b/source/jormungandr/jormungandr/autocomplete/abstract_autocomplete.py
@@ -75,7 +75,7 @@ class AbstractAutocomplete(six.with_metaclass(ABCMeta, object)):
         pass
 
     def record_status(self, status, exc=None):
-        data = {'type': type(self).__name__, 'status': status}
+         data = {'type': self.__class__.__name__, 'status': status}
         if exc is not None:
             data["cause"] = str(exc)
         record_custom_event('autocomplete_status', data)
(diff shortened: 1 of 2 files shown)


source/jormungandr/jormungandr/autocomplete/abstract_autocomplete.py
```diff
@@ -65,6 +74,12 @@ def status(self):
     def geo_status(self, instance):
         pass
 
+    def record_status(self, status, exc=None):
+        data = {'type': type(self).__name__, 'status': status}
```
xlqian: be careful with `type`
```python
class A(object):
    pass

class B:
    pass

print(type(A()).__name__)
print(type(B()).__name__)
```
```suggestion
         data = {'type': self.__class__.__name__, 'status': status}
```

nonifier: That's tricky :open_mouth: !

nonifier (approved):

source/jormungandr/jormungandr/autocomplete/abstract_autocomplete.py
kinnou02: good catch! :+1:

xlqian (approved):

PR #2639, opened 2018-12-31
"migration" to c++14

kinnou02:
we can migrate to c++14 since we do not support debian 7 anymore

### DIFF
--- a/source/cmake_modules/CompilationFlags.cmake
+++ b/source/cmake_modules/CompilationFlags.cmake
@@ -14,7 +14,7 @@ ENDIF(NOT CMAKE_BUILD_TYPE)
 # Gcc Release flags
 if(CMAKE_COMPILER_IS_GNUCXX)
     set(CMAKE_GNUCXX_WARN_FLAGS "-Wall -Wextra -Woverloaded-virtual -Wundef -pedantic")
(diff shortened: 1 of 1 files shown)


source/cmake_modules/CompilationFlags.cmake
```diff
@@ -14,7 +14,7 @@ ENDIF(NOT CMAKE_BUILD_TYPE)
 # Gcc Release flags
 if(CMAKE_COMPILER_IS_GNUCXX)
     set(CMAKE_GNUCXX_WARN_FLAGS "-Wall -Wextra -Woverloaded-virtual -Wundef -pedantic")
-    set(CMAKE_GNUCXX_COMMON_FLAGS "-std=c++0x -rdynamic -g")
+    set(CMAKE_GNUCXX_COMMON_FLAGS "-std=c++14 -rdynamic -g")
```
xlqian, line 17: milestone.... and Tisseo is ok with that?

kinnou02: For the first time they are in advance, they already "ship" on debian 9.

TeXitoi (approved):

TeXitoi (approved):

PR #2641, opened 2019-01-02
Kraken: consider reduced_service as modifying effect

xlqian:
consider REDUCED_SERVICE as modifying effect 
This should fix the test in artemis: https://github.com/CanalTP/artemis/blob/master/artemis/tests/guichet_unique_test.py#L258

### DIFF
--- a/source/jormungandr/tests/kirin_realtime_tests.py
+++ b/source/jormungandr/tests/kirin_realtime_tests.py
@@ -906,6 +906,52 @@ class TestKirinOnNewStopTimeAtTheEnd(MockKirinDisruptionsFixture):
 
 
 @dataset(MAIN_ROUTING_TEST_SETTING)
+class TestKirinReadTripEffectFromTripUpdate(MockKirinDisruptionsFixture):
+    def test_read_trip_effect_from_tripupdate(self):
+        disruptions_before = self.query_region('disruptions?_current_datetime=20120614T080000')
+        nb_disruptions_before = len(disruptions_before['disruptions'])
+        assert nb_disruptions_before == 10
+
+        vjs_before = self.query_region('vehicle_journeys')
+        assert len(vjs_before['vehicle_journeys']) == 7
+
+        self.send_mock(
+            "vjA",
+            "20120614",
+            'modified',
+            [
+                UpdatedStopTime(
+                    "stop_point:stopB",
+                    arrival=tstamp("20120614T080224"),
+                    departure=tstamp("20120614T080225"),
+                    arrival_delay=0,
+                    departure_delay=0,
+                ),
+                UpdatedStopTime(
+                    "stop_point:stopA",
+                    arrival=tstamp("20120614T080400"),
+                    departure=tstamp("20120614T080400"),
+                    message='stop_time deleted',
+                    arrival_skipped=True,
+                    departure_skipped=True,
+                ),
+            ],
+            disruption_id='reduced_service_vjA',
+            effect='reduced_service',
+        )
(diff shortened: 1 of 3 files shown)


xlqian:
I'm checking if this change may impact the disruptions from Chaos

source/kraken/realtime.cpp
```diff
@@ -291,6 +291,7 @@ static const std::string get_wordings(nt::disruption::Effect effect) {
     case Effect::MODIFIED_SERVICE:
         return  "trip modified";
     case Effect::REDUCED_SERVICE:
+        return "reduced service";
     case Effect::ADDITIONAL_SERVICE:
```
pbougue, line 296: Could be nice to improve all wordings, for next debug need :)

source/kraken/apply_disruption.cpp
```diff
@@ -501,12 +498,15 @@ static bool is_modifying_effect(nt::disruption::Effect e) {
     return in(e, {nt::disruption::Effect::NO_SERVICE,
                   nt::disruption::Effect::SIGNIFICANT_DELAYS,
                   nt::disruption::Effect::MODIFIED_SERVICE,
-                  nt::disruption::Effect::DETOUR});
+                  nt::disruption::Effect::DETOUR,
```
pbougue, line 501: Not sure: Shouldn't we define it negatively?
We might end up with the same bug next time we add/handle a value of that enum, no?

pbougue (commented):
This looks good.
Do we consider testing in artemis is enough or is it easy (and worthy) to add unit tests also?

kadhikari:
A test with effect = REDUCED_SERVICE in integration test (kirin_realtile_tests.py) should work.

source/kraken/apply_disruption.cpp
xlqian: And I have another question at a higher level: **shouldn't this be handled by the connector?** this code is common for chaos and kirin, i've no idea what to do if we get a new connector and the definition of `modifying_effect` is totally defined in a new way...

TeXitoi (approved):

kadhikari (approved):

nonifier (approved):

nonifier:
And here is the Artemis change that goes along: https://github.com/CanalTP/artemis/pull/280 :)

ps: I kicked off a [custom build on Artemis](https://ci.navitia.io/view/artemis/job/custom_artemis_reference_branch_and_artemis_branch_build/) (knowingly failing on [6 guichet unique tests](https://ci.navitia.io/view/artemis/job/Artemis/2000/#showFailuresLink)) 

ps 2: [Tests have failed](https://ci.navitia.io/view/artemis/job/custom_artemis_reference_branch_and_artemis_branch_build/182/#showFailuresLink)... Maybe due to the already failing tests on guichet-unique.

source/jormungandr/tests/kirin_realtime_tests.py
```diff
@@ -1189,7 +1235,7 @@ def test_add_modify_and_delete_one_stop_time(self):
             == 'unchanged'  # Why ?
         )
         assert disrupts['disruptions'][11]['severity']['effect'] == 'DETOUR'
-        assert disrupts['disruptions'][11]['severity']['name'] == 'trip modified'
+        assert disrupts['disruptions'][11]['severity']['name'] == 'detour'
```
nonifier, line 1238: Consistency for the win ! :+1:

nonifier (approved):

kadhikari (approved):

source/kraken/realtime.cpp
xlqian: done

PR #2644, opened 2019-01-04
"migration" to c++14 : episode 2 - clang

pbougue:
after https://github.com/CanalTP/navitia/pull/2639

updated lz4 to v1.8.3 (latest release) as compiler was not correctly handled in version used
also removed a little clang warning on ref being a hidden copy

### DIFF
--- a/source/routing/next_stop_time.cpp
+++ b/source/routing/next_stop_time.cpp
@@ -437,7 +437,7 @@ CachedNextStopTime CachedNextStopTimeManager::CacheCreator::operator()(const Cac
 
 CachedNextStopTime::DtStFromJpp::DtStFromJpp(const vDtStByJpp& map) {
     size_t s = 0;
-    for (const auto& elt: map) {
+    for (const auto elt: map) {
         s += elt.second.size();
     }
     dtsts.reserve(s);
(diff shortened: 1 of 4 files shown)


source/routing/next_stop_time.cpp
```diff
@@ -437,7 +437,7 @@ CachedNextStopTime CachedNextStopTimeManager::CacheCreator::operator()(const Cac
 
 CachedNextStopTime::DtStFromJpp::DtStFromJpp(const vDtStByJpp& map) {
     size_t s = 0;
-    for (const auto& elt: map) {
+    for (const auto elt: map) {
```
TeXitoi, line 440: Why? It yields something as `std::pair<&T, &U>` ?

kinnou02:
What was the problems? It didn't build with clang or there was some problems with gcc too?
Does this version of lz4 is rétrocompatible with the previous one? Will it load old .nav or do we need to rebuild them?

nonifier (approved):
:tada:

source/routing/next_stop_time.cpp
pbougue: It uses those : https://github.com/CanalTP/utils/blob/master/idx_map.h#L130-L131

pbougue (commented):
@kinnou02 : the problem for lz4 was the macro handling `deprecated` specifically depending on compiler, it was not handling clang c++14.
And I tested data.nav.lz4 generated without this PR : it works with a kraken from this PR.

TeXitoi (approved):

PR #2648, opened 2019-01-10
Configure codecov

kinnou02:
This is based on the default config, there is only two changes:
  - Increase of threshold to two percent, since sometime our coverage
  vary randomly...
  - Lower patch code coverage requirements to 25%, to prevent the red
  cross effect


pbougue (approved):
Looks OK and checks passed ;)

nonifier (approved):

PR #2650, opened 2019-01-14
Eitri: add cities into eitri

xlqian:
now we can create data.nav with eitri!!

This will be very useful for those who want to run **Artemis** but with their **OWN** working-in-progress Navitia

### DIFF
--- a/source/cities/alembic.ini
+++ b/source/cities/alembic.ini
@@ -1,3 +1,6 @@
+[alembic]
+script_location = alembic
+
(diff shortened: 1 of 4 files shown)


kinnou02 (approved):
looks good, did you run docker test to be sure?

xlqian:
> looks good, did you run docker test to be sure?

not yet, i'll test it

nonifier (approved):

PR #2651, opened 2019-01-14
Jormun: Fix count on /departures with realtime proxies

kinnou02:
This is a follow-up on https://github.com/CanalTP/navitia/pull/2645
The count parameter wasn't used correctly on departures when realtime
proxies were involved.
This adds a post action that will enforce the count, so we never get more
than `count` departure.

We could change the realtime proxies to let them know about the
differences between /departures and /stop_schedules to optimize calls
made to external services. It's a lot of change that I won't make it for
now.

JIRA: https://jira.kisio.org/browse/NAVITIAII-2638

### DIFF
--- a/source/jormungandr/jormungandr/schedule.py
+++ b/source/jormungandr/jormungandr/schedule.py
@@ -303,6 +303,9 @@ class MixedSchedule(object):
             return cmp(p1.stop_date_time.departure_date_time, p2.stop_date_time.departure_date_time)
 
         resp.next_departures.sort(comparator)
+        count = request['count']
+        if len(resp.next_departures) > count and rt_proxy:
+            del resp.next_departures[count:]
 
         # handle pagination :
         # If real time information exist, we have to change pagination score.
--- a/source/jormungandr/jormungandr/schedule.py
+++ b/source/jormungandr/jormungandr/schedule.py
@@ -304,7 +304,7 @@ class MixedSchedule(object):
 
         resp.next_departures.sort(comparator)
         count = request['count']
-        if len(resp.next_departures) > count and rt_proxy:
+        if len(resp.next_departures) > count:
             del resp.next_departures[count:]
 
         # handle pagination :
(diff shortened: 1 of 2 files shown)


source/jormungandr/tests/proxy_realtime_tests.py
```diff
@@ -342,6 +342,35 @@ def test_stop_schedule_limit_per_items(self):
         for stop_sched in response['stop_schedules']:
             assert len(stop_sched['date_times']) == 2
 
+    def test_departures_count(self):
+        """
+        test that count parameter is correctly taken into account when using realtime proxies
+        We search for departures on C:vj1: C:S0 -> 11:30, C:S1 -> 12:30, C:S2 -> 13:30
+        So we got departures for the route points of the line C to the stops C:S0, C:S1, C:S2
+        So there is two departures:
```
nonifier:
```suggestion
        So there are two departures:
```

source/jormungandr/jormungandr/schedule.py
```diff
@@ -303,6 +303,9 @@ def comparator(p1, p2):
             return cmp(p1.stop_date_time.departure_date_time, p2.stop_date_time.departure_date_time)
 
         resp.next_departures.sort(comparator)
+        count = request['count']
+        if len(resp.next_departures) > count and rt_proxy:
```
nonifier: Are you sure you don't want to trim the next departures event if we have no `rt_proxy` ?

nonifier (approved):

source/jormungandr/jormungandr/schedule.py
kinnou02: it's not required as kraken already return at most `count` items, but in fact I added this test because something else was failing, forgot to remove it
```suggestion
        if len(resp.next_departures) > count:
```

pbougue (approved):

TeXitoi (approved):

PR #2652, opened 2019-01-15
Fix get_base_stop_time

nonifier:
Fix base arrival/departure times that were missing after adding a new stop_time in a disruption.

3 Different implementations of `get_base_stop_time` have been found.... but [only 1 survived !](http://gph.is/1KVPGOF) 

We keep the one in [type.cpp#L178-L198](https://github.com/CanalTP/navitia/blob/14364090fa3d7cca705c48874184221ca3cfdee3/source/type/type.cpp#L178-L198) that fixes our issue. 

[#NAVP-1131](https://jira.kisio.org/browse/NAVP-1131)

### DIFF
--- a/source/kraken/tests/realtime_test.cpp
+++ b/source/kraken/tests/realtime_test.cpp
@@ -73,7 +73,7 @@ make_cancellation_message(const std::string& vj_uri, const std::string& date) {
     return trip_update;
 }
 
-pbnavitia::Response compute_iti(
+static pbnavitia::Response compute_iti(
     const ed::builder& b,
     const char* datetime, 
     const std::string& from, 
--- a/source/kraken/tests/realtime_test.cpp
+++ b/source/kraken/tests/realtime_test.cpp
@@ -1,32 +1,32 @@
-//  Copyright © 2001-2015, Canal TP and/or its affiliates. All rights reserved.
-// 
-// This file is part of Navitia,
-//     the software to build cool stuff with public transport.
-// 
-// Hope you'll enjoy and contribute to this project,
-//     powered by Canal TP (www.canaltp.fr).
-// Help us simplify mobility and open public transport:
-//     a non ending quest to the responsive locomotion way of traveling!
-// 
-// LICENCE: This program is free software; you can redistribute it and/or modify
-// it under the terms of the GNU Affero General Public License as published by
-// the Free Software Foundation, either version 3 of the License, or
-// (at your option) any later version.
-// 
-// This program is distributed in the hope that it will be useful,
-// but WITHOUT ANY WARRANTY; without even the implied warranty of
-// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
-// GNU Affero General Public License for more details.
-// 
-// You should have received a copy of the GNU Affero General Public License
-// along with this program. If not, see <http://www.gnu.org/licenses/>.
-// 
-// Stay tuned using
-// twitter @navitia
-// IRC #navitia on freenode
-// https://groups.google.com/d/forum/navitia
-// www.navitia.io
-//
+/* Copyright © 2001-2015, Canal TP and/or its affiliates. All rights reserved.
+
+This file is part of Navitia,
+    the software to build cool stuff with public transport.
+
+Hope you'll enjoy and contribute to this project,
+    powered by Canal TP (www.canaltp.fr).
+Help us simplify mobility and open public transport:
+    a non ending quest to the responsive locomotion way of traveling!
+
+LICENCE: This program is free software; you can redistribute it and/or modify
+it under the terms of the GNU Affero General Public License as published by
+the Free Software Foundation, either version 3 of the License, or
+(at your option) any later version.
+
+This program is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+GNU Affero General Public License for more details.
+
+You should have received a copy of the GNU Affero General Public License
+along with this program. If not, see <http://www.gnu.org/licenses/>.
+
+Stay tuned using
+twitter @navitia
+IRC #navitia on freenode
+https://groups.google.com/d/forum/navitia
+www.navitia.io
+*/
 
 #define BOOST_TEST_DYN_LINK
 #define BOOST_TEST_MODULE test_realtime
--- a/source/kraken/tests/realtime_test.cpp
+++ b/source/kraken/tests/realtime_test.cpp
@@ -1,32 +1,32 @@
-/* Copyright © 2001-2015, Canal TP and/or its affiliates. All rights reserved.
-
-This file is part of Navitia,
-    the software to build cool stuff with public transport.
-
-Hope you'll enjoy and contribute to this project,
-    powered by Canal TP (www.canaltp.fr).
-Help us simplify mobility and open public transport:
-    a non ending quest to the responsive locomotion way of traveling!
-
-LICENCE: This program is free software; you can redistribute it and/or modify
-it under the terms of the GNU Affero General Public License as published by
-the Free Software Foundation, either version 3 of the License, or
-(at your option) any later version.
-
-This program is distributed in the hope that it will be useful,
-but WITHOUT ANY WARRANTY; without even the implied warranty of
-MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
-GNU Affero General Public License for more details.
-
-You should have received a copy of the GNU Affero General Public License
-along with this program. If not, see <http://www.gnu.org/licenses/>.
-
-Stay tuned using
-twitter @navitia
-IRC #navitia on freenode
-https://groups.google.com/d/forum/navitia
-www.navitia.io
-*/
+//  Copyright © 2001-2015, Canal TP and/or its affiliates. All rights reserved.
+// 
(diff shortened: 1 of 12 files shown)


source/type/pb_converter.cpp
```diff
@@ -133,7 +133,7 @@ struct PbCreator::Filler::PtObjVisitor: public boost::static_visitor<> {
                                                            impacted_stop->mutable_amended_stop_time());
 
             // we need to get the base stoptime
-            const auto* base_st = impact.aux_info.get_base_stop_time(stu);
+            const auto* base_st = stu.stop_time.get_base_stop_time();
```
pbougue, line 136: So the only "possible regression" is here (Kraken used to care about index here), but it looks to me that it is used when displaying a disruption... To be confirmed if possible (poke @TeXitoi ?).

Then it wouldn't be too bad to display a lollipop disruption badly.

source/kraken/tests/realtime_test.cpp
```diff
@@ -2545,3 +2570,64 @@ BOOST_AUTO_TEST_CASE(add_new_with_earlier_arrival_and_delete_existingstop_time_i
...
+
+    b.sa("A");
+    b.sa("B");
+    b.sa("C");
+    b.sa("D");
+
+    b.vj("L1").uri("vj:1")
+        ("stop_point:A", "08:10"_t)
+        ("stop_point:C", "08:30"_t)
+        ("stop_point:D", "08:40"_t);
```
pbougue, line 2584: It would be nice to add a stop at C or B at the end, make also a delay message on all stops, and see how it behaves also on a trip from A to C and from D to C (only not crashing is OK ;-) )

pbougue (approved):
Looks good, but no-lollipop handling feels dangerous.
Although it looks like this doesn't impact the creation of VJ, so it should be safe regarding a back-to-the-future syndrome in journey-planning.

Kraken probably looses a bit on disruption's display for lollipop VJ.
And is as bad as it was for handling boarding/aligthing in lollipop case.
But improving one would help all, now :)

source/kraken/tests/realtime_test.cpp
nonifier: The original problem was that the arrivals/departures **after** an added stop time were wrong. I don"t see how adding a stop at the end would help in this regard. 

Handling lollipops is another matter that should be fixed separately as it is already (partially) broken.

source/type/pb_converter.cpp
TeXitoi: I really don't know if that's critical, but in theory, sherbrooke has lolipop and only delays, thus we break something (that might not be used).

A cheap lolipop fix can be added, something like "1. checking that we are the ith stop time using this stop point, and then searching for the ith stop time using this stop point in the base".

nonifier:
I'm having a look if we can have a quick fix for the lollipop case.

xlqian (approved):

source/type/type.cpp
```diff
@@ -176,26 +176,48 @@ bool StopTime::is_valid_day(u_int32_t day, const bool is_arrival, const RTLevel
...
-            ++nb_st_found;
+            st_base.push_back(&st_it);
         }
     }
-    //TODO handle lolipop vj, bob-commerce-commerce-bobito vj...
-    if (nb_st_found != 1) {
+
+    if (st_base.empty()) { return nullptr; }
+
+    // handle lolipop
```
pbougue: typo: lol`l`ipop

source/type/type.cpp
```diff
@@ -176,26 +176,48 @@ bool StopTime::is_valid_day(u_int32_t day, const bool is_arrival, const RTLevel
...
         }
     }
-    //TODO handle lolipop vj, bob-commerce-commerce-bobito vj...
-    if (nb_st_found != 1) {
+
+    if (st_base.empty()) { return nullptr; }
+
+    // handle lolipop 
+    size_t nb_rt_st_found = 0;
+    if (st_base.size() > 1) {
```
pbougue: Ideas:
* Count the number of stops on each VJ and compare could be nice too (if they are not the same...)
* Mention that Levenshtein is probably the best idea to improve that in a comment
* Check information from Kirin's feed (if available) of what's added/removed

source/type/type.cpp
```diff
@@ -176,26 +176,48 @@ bool StopTime::is_valid_day(u_int32_t day, const bool is_arrival, const RTLevel
...
+        for (const auto& st : vehicle_journey->stop_time_list) {
+            if (&st == this) {
+                return st_base[nb_rt_st_found];
```
pbougue: :boom: if realtime_VJ passes more times by given StopPoint than base_VJ.

source/kraken/tests/realtime_test.cpp
```diff
@@ -2631,3 +2631,103 @@ BOOST_AUTO_TEST_CASE(should_get_base_stoptime_with_realtime_added_stop_time) {
...
 }
+
+
+BOOST_AUTO_TEST_CASE(should_get_correct_base_stop_time_with_lollipop) {
+    /*
+    We create a network with a lollipop and apply delay on every stop points.
+    We expect from the "real time" VJ that its base stop times are matching the original VJ.
+
+        (08:10) A   A' (08:50)
+                ▼   ▲
```
pbougue, line 2641: :heart_eyes: unicode-art :arrow_down_small:

pbougue (dismissed):
You probably have to protect your list from being shorter (count occurrences of SP in both VJ?).

source/type/type.cpp
TeXitoi: And, if you first search for the current vj to have the rank, and then search in the base, you don't need to allocate a vector.

source/kraken/tests/realtime_test.cpp
```diff
@@ -1750,20 +1750,19 @@ BOOST_AUTO_TEST_CASE(delays_on_lollipop_with_boarding_alighting_times) {
...
+    BOOST_CHECK_EQUAL(vj->stop_time_list.at(0).departure_time, "08:11"_t);
+    BOOST_CHECK_EQUAL(vj->stop_time_list.at(0).boarding_time, "08:06"_t); // Boarding time is 5 min (300s) before departure 
+    
+    BOOST_CHECK_EQUAL(vj->stop_time_list.at(1).stop_point->uri, "stop_point:20");
+    BOOST_CHECK_EQUAL(vj->stop_time_list.at(1).departure_time, "08:21"_t);
+    BOOST_CHECK_EQUAL(vj->stop_time_list.at(1).boarding_time, "08:21"_t);
+
+    BOOST_CHECK_EQUAL(vj->stop_time_list.at(2).stop_point->uri, "stop_point:10");
+    BOOST_CHECK_EQUAL(vj->stop_time_list.at(2).arrival_time, "08:40"_t);
+    BOOST_CHECK_EQUAL(vj->stop_time_list.at(2).alighting_time, "08:45"_t); // Alighting time is 5 min (300s) after arrival
```
pbougue, line 1765: :heart:

nonifier:
I've addressed your comments guys, you can review :)

source/type/type.cpp
```diff
@@ -191,33 +191,40 @@ const StopTime* StopTime::get_base_stop_time() const {
...
-    std::vector<const nt::StopTime*> st_base;
-    for (const auto& st_it: base_vj->stop_time_list) {
-        if (stop_point == st_it.stop_point) {
-            st_base.push_back(&st_it);
+    // Partially handle lollipop 
+    //
+    // We get the rank of the stop_stime in the current VJ, and 
+    // return the one with same rank in the base VJ. 
+    // There are limitations if:
+    //   * the lollopop's node (stop_point B) is added before
```
pbougue: minor typo: loll`i`pop's

source/type/type.cpp
```diff
@@ -191,33 +191,40 @@ const StopTime* StopTime::get_base_stop_time() const {
...
         }
-
-        auto logger = log4cplus::Logger::getInstance("log");
-        LOG4CPLUS_DEBUG(logger, "Ignored stop_time finding: impossible to match exactly one base stop_time");
-        return nullptr;
     }
 
-    return st_base[0];
+    auto logger = log4cplus::Logger::getInstance("log");
+    LOG4CPLUS_DEBUG(logger, "Ignored stop_time " << stop_point->uri << ":" << departure_time << ": impossible to match exactly one base stop_time");
```
pbougue: minor: line length

pbougue (approved):
Looks good!
Only minor comments in case something else comes up they may be addressed along the way.
I'll re-trigger a build as the current was frozen... In disruption apply tests... :-/

source/type/type.cpp
xlqian: lol

xlqian (approved):

TeXitoi (approved):

source/type/type.h
```diff
@@ -757,6 +757,10 @@ struct StopTime {
 
     const StopTime* get_base_stop_time() const;
 
+    //  Check if a stop time has similar hours of departure/arrival/alighting/boarding_time etc...
+    //  We don't want to rely on properties that might change with real time object.
+    bool has_similar_timings(const StopTime& st) const;
```
pbougue: minor: `is_similar` sound better to me, as we don't rely only on timings.

pbougue (approved):
:tada: 
no easy change on safety, so let's go with that :)
Or maybe just add a comment mentioning that STU is not the "natural" owner of ST?

kadhikari (approved):

TeXitoi (approved):

PR #2654, opened 2019-01-16
Asgard direct path

AurelienLP:


kinnou02:
Jenkins, test this please.

source/jormungandr/jormungandr/street_network/asgard.py
```diff
@@ -30,24 +30,29 @@
...
+
+class Asgard(Kraken):
     def __init__(
         self, instance, service_url, asgard_socket, modes=[], id='asgard', timeout=10, api_key=None, **kwargs
     ):
         super(Asgard, self).__init__(instance, service_url, modes, id, timeout, api_key, **kwargs)
         self.asgard_socket = asgard_socket
+        self.timeout = timeout
         self._sockets = queue.Queue()
+        self.get_uri_pt_func = get_uri_pt_object_for_asgard
```
kinnou02: you could use a method that you overload in `Asgard`

TeXitoi: Is it really pythonic? @xlqian

kinnou02 (commented):
Didn't had time to read in detail :(

source/jormungandr/jormungandr/street_network/asgard.py
AurelienLP: The function used in Kraken is from utils and is not part of the class. Can I still overload it ?

xlqian: i agree with @kinnou02... overload a method is cleaner for me... :)

source/jormungandr/jormungandr/street_network/kraken.py
```diff
@@ -46,6 +46,7 @@ def __init__(self, instance, service_url, modes=[], id='kraken', timeout=10, api
         self.instance = instance
         self.modes = modes
         self.sn_system_id = id
+        self.get_uri_pt_func = get_uri_pt_object
```
xlqian: i would prefer a "private" method

source/jormungandr/jormungandr/street_network/asgard.py
```diff
@@ -30,24 +30,29 @@
...
 from zmq import green as zmq
-import six
 
 
-class Asgard(Valhalla):
+def get_uri_pt_object_for_asgard(obj):
+    return 'coord:{c.lon}:{c.lat}'.format(c=get_pt_object_coord(obj))
+
+
+class Asgard(Kraken):
```
xlqian, line 43: by inheriting `Kraken`, the behavior of `_direct_path` is changed a lot... is that wanted?

I suppose that the handling of parking section will be done later, right?

AurelienLP: "I suppose that the handling of parking section will be done later, right?"

--> Yes, it will be handled in another class

xlqian (approved):

PR #2655, opened 2019-01-16
Tyr: "correctly" handle error in async command

kinnou02:
celery is  bit funny and doesn't seems to raise exception raised by a
worker like it should. More precisely it doesn't raise once the wait as
began, but before starting blocking if the worker have encountered an
exception, wait will raise it...
So I poll on wait to get the exception...

### DIFF
--- a/source/tyr/tyr/helper.py
+++ b/source/tyr/tyr/helper.py
@@ -42,7 +42,7 @@ import sys
 import tempfile
 
 
-def wait_and_raise(async_result):
+def wait_or_raise(async_result):
     """
     wait for the end of an async task.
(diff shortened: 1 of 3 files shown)


source/tyr/tyr/helper.py
```diff
@@ -41,6 +42,22 @@
 import tempfile
 
 
+def wait_and_raise(async_result):
```
xlqian: minor: it looks more like `wait_or_raise` than `wait_and_raise`...

xlqian (approved):

TeXitoi (approved):

PR #2656, opened 2019-01-17
Handle errors in distributed scenario

AurelienLP:
This PR is the first step to handle errors in distributed mode when a service is not available.
https://jira.kisio.org/browse/NAVP-895

### DIFF
--- a/source/jormungandr/jormungandr/scenarios/helper_classes/fallback_durations.py
+++ b/source/jormungandr/jormungandr/scenarios/helper_classes/fallback_durations.py
@@ -172,7 +172,11 @@ class FallbackDurations:
         streetnetwork_service = self._instance.get_street_network(self._mode, self._request)
         sn_routing_matrix = self._get_street_network_routing_matrix(streetnetwork_service, origins, destinations)
 
-        if not sn_routing_matrix or not len(sn_routing_matrix.rows) or not len(sn_routing_matrix.rows[0].routing_response):
+        if (
+            not sn_routing_matrix
+            or not len(sn_routing_matrix.rows)
+            or not len(sn_routing_matrix.rows[0].routing_response)
+        ):
             logger.debug("no fallback durations found from %s by %s", self._requested_place_obj.uri, self._mode)
             return result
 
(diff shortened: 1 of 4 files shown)


source/jormungandr/jormungandr/scenarios/helper_classes/fallback_durations.py
```diff
@@ -103,9 +103,23 @@ def _get_duration(self, resp, place):
...
+            return self._instance.get_street_network_routing_matrix(
+                origins,
+                destinations,
+                self._mode,
+                self._max_duration_to_pt,
+                self._request,
+                **self._speed_switcher
+            )
+        except Exception as e:
+            print("Exception':{}".format(str(e)))
```
TeXitoi: We want to log, not print

source/jormungandr/jormungandr/scenarios/helper_classes/fallback_durations.py
```diff
@@ -103,9 +103,23 @@ def _get_duration(self, resp, place):
...
+            resp = response_pb2.StreetNetworkRoutingMatrix()
+            row = resp.rows.add()
+            routing = row.routing_response.add()
+            routing.duration = -1
+            routing.routing_status = response_pb2.unknown
```
TeXitoi: That's strange to return such a non-sense response, why is it needed?

TeXitoi (changes requested):

source/jormungandr/jormungandr/scenarios/helper_classes/fallback_durations.py
AurelienLP: So that it doesn't crash afterwards and is treated as an empty response.

xlqian:
Can you add the associated jira ticket in the description plz?

source/jormungandr/jormungandr/scenarios/helper_classes/fallback_durations.py
```diff
@@ -103,9 +103,23 @@ def _get_duration(self, resp, place):
...
+                self._mode,
+                self._max_duration_to_pt,
+                self._request,
+                **self._speed_switcher
+            )
+        except Exception as e:
+            logging.getLogger(__name__).error("Exception':{}".format(str(e)))
+            resp = response_pb2.StreetNetworkRoutingMatrix()
+            row = resp.rows.add()
+            routing = row.routing_response.add()
```
xlqian: It makes probably more sense that you set the error status at the level of resp... just for the sake of coherence (maybe we can use this field in the future to  raise the exception in the final response...)

TeXitoi (approved):

xlqian (approved):

PR #2657, opened 2019-01-18
Jormun/tyr: rollback to previous version of celery and redis

kinnou02:
We have a lot of timeout with the last version of redis, it's need to be
analyzed before we can switch to it.
I had to rename jormungandr.redis in jormungandr.customredis as python2.7
have some problems with relative import


pbougue (approved):
Looks OK from what we had before bump.
So you stopped any use of customredis.py by renaming it but kept the code to use it back later?

kinnou02:
Customredis isn't used directly by jormun, it needs to be configured in settings.py to be used.
We didn't migrate to it yet.

mbounabi (approved):

PR #2658, opened 2019-01-18
jormun: Prevent crash in jormun if we got a bad geom from kraken

kinnou02:
There is one very very, very long line, but it's hard to create a wrong poly that look like it's valid :(


source/jormungandr/jormungandr/tests/instance_tests.py
```diff
@@ -0,0 +1,55 @@
...
+def has_point_test():
+    geom = wkt.loads(
+        "MULTIPOLYGON(((-4.4392 48.1514,-4.4392 47.7647,-3.8881 47.7647,-3.8881 48.1514,-4.4392 48.1514)))"
+    )
+    instance = FakeInstance(geom)
+    assert instance.has_coord(-4.0392, 48.0514)
+    assert not instance.has_coord(-2.0392, 48.0514)
+
+    # same test with a invalid poly
+    geom = wkt.loads(bad_poly)
```
xlqian, line 50: :+1:

source/jormungandr/jormungandr/instance.py
```diff
@@ -525,6 +525,9 @@ def has_point(self, p):
             return self.geom and self.geom.contains(p)
         except DeadSocketException:
             return False
+        except PredicateError:
```
xlqian, line 528: just a little thought: for me, i don't know if catch the exception is a better option than throw the exception(or crash more beautifully..., i know no one like **`server internal error`**)... in this case, catch the exception means, very  probably, a misleading response...

xlqian (approved):

source/jormungandr/jormungandr/instance.py
kinnou02: But we may find another coverage that will respond.
This is a first and quick step, I will also update tyr and ed to prevent any invalid shape to be integrated.

pbougue (approved):
OK, although it's not on data input.
Curious about the type of shape that makes it crash, though.

kinnou02:
> OK, although it's not on data input.
> Curious about the type of shape that makes it crash, though.

something that look like this:
selection_183

selection_184

image

xlqian:
absolutely fantastic shape!!!

TeXitoi (approved):

PR #2660, opened 2019-01-22
Jormungandr: Module with parking

AurelienLP:
Adding a simple parking module to add additional parking time for routing_matrix and direct_path durations.

### DIFF
--- a/source/jormungandr/jormungandr/street_network/constant_parking.py
+++ b/source/jormungandr/jormungandr/street_network/constant_parking.py
@@ -27,8 +27,10 @@
 # https://groups.google.com/d/forum/navitia
 # www.navitia.io
 
+from jormungandr.street_network.abstract_parking_module import AbstractParkingModule
 
-class ConstantParking:
+
+class ConstantParking(AbstractParkingModule):
     def __init__(self, park_duration, leave_parking_duration):
         self.park_duration = park_duration
         self.leave_parking_duration = leave_parking_duration
--- a/source/jormungandr/jormungandr/street_network/constant_parking.py
+++ b/source/jormungandr/jormungandr/street_network/constant_parking.py
@@ -35,8 +35,8 @@ class ConstantParking(AbstractParkingModule):
         self.park_duration = park_duration
         self.leave_parking_duration = leave_parking_duration
 
-    def get_parking_duration(self, coords, datetime):
+    def get_parking_duration(self, coords):
         return self.park_duration
 
-    def get_leave_parking_duration(self, coords, datetime):
+    def get_leave_parking_duration(self):
         return self.leave_parking_duration
--- a/source/jormungandr/jormungandr/street_network/constant_parking.py
+++ b/source/jormungandr/jormungandr/street_network/constant_parking.py
@@ -27,7 +27,8 @@
 # https://groups.google.com/d/forum/navitia
 # www.navitia.io
 
-class ConstantParking():
+
+class ConstantParking:
     def __init__(self, park_duration, leave_parking_duration, **kwargs):
         self.park_duration = park_duration
         self.leave_parking_duration = leave_parking_duration
--- a/source/jormungandr/jormungandr/street_network/constant_parking.py
+++ b/source/jormungandr/jormungandr/street_network/constant_parking.py
@@ -1,0 +1,33 @@
+#  Copyright (c) 2001-2016, Canal TP and/or its affiliates. All rights reserved.
+#
+# This file is part of Navitia,
+#     the software to build cool stuff with public transport.
+#
+# Hope you'll enjoy and contribute to this project,
+#     powered by Canal TP (www.canaltp.fr).
+# Help us simplify mobility and open public transport:
+#     a non ending quest to the responsive locomotion way of traveling!
+#
+# LICENCE: This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU Affero General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+# GNU Affero General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with this program. If not, see <http://www.gnu.org/licenses/>.
+#
+# Stay tuned using
+# twitter @navitia
+# IRC #navitia on freenode
+# https://groups.google.com/d/forum/navitia
+# www.navitia.io
+
+class ConstantParking():
+    def __init__(self, park_duration, leave_parking_duration, **kwargs):
+        self.park_duration = park_duration
+        self.leave_parking_duration = leave_parking_duration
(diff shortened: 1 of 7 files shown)


source/jormungandr/jormungandr/street_network/constant_parking.py
```diff
@@ -0,0 +1,34 @@
...
+#
+# Stay tuned using
+# twitter @navitia
+# IRC #navitia on freenode
+# https://groups.google.com/d/forum/navitia
+# www.navitia.io
+
+
+class ConstantParking:
+    def __init__(self, park_duration, leave_parking_duration, **kwargs):
```
xlqian: `**kwargs` is not really necessary in this case

source/jormungandr/jormungandr/street_network/with_parking.py
```diff
@@ -0,0 +1,107 @@
...
+import logging
+from jormungandr.street_network.street_network import AbstractStreetNetworkService
+from jormungandr import street_network
+
+from jormungandr.street_network.asgard import Asgard
+from jormungandr import utils
+
+
+class WithParking(AbstractStreetNetworkService):
+    def __init__(self, instance, service_url, modes=[], id='with_parking', timeout=10, api_key=None, **kwargs):
```
xlqian: be careful with "objects as default values": `modes=[]` 
try this code
```python
class A(object):
    def __init__(self, a=[]):
        self.a = a
        self.a.append(1)
        print(self.a)

A()
A()
A()
```

prefer
```python
class A(object):
    def __init__(self, a=None):
        self.a = a or []
        self.a.append(1)
        print(self.a)

A()
A()
A()
```

source/jormungandr/jormungandr/street_network/with_parking.py
```diff
@@ -0,0 +1,107 @@
...
+        if response and len(response.journeys):
+            self.add_parking_time_in_direct_path(response)
+
+        return response
+
+    def add_parking_time_in_direct_path(self, response):
+        logger = logging.getLogger(__name__)
+        logger.debug("Adding parking time for direct path " + str(self.parking_module.park_duration))
+        for journey in response.journeys:
+            journey.duration += self.parking_module.park_duration
```
xlqian: hmmm, don't you need to add parking section at the end of the journey?

xlqian (commented):
if possible, can you add some integration tests, plz?

source/jormungandr/jormungandr/street_network/with_parking.py
xlqian: yeh... welcome to the evil side of python, I never understand why they make this...

nonifier: :see_no_evil:

xlqian: oops, I almost forget: `id='with_parking'` is should be changed as well

source/jormungandr/jormungandr/street_network/with_parking.py
nonifier: Shouldn't this being calculated by Asgard instead ?

AurelienLP: @xlqian Yes, I have to create a section in the response, not only adding the time. Will do now.

@nonifier The parking module and the calculator (Asgard, Kraken, Géovelo...) used are independent. The calculator doesn't know anything about the parking module.

source/jormungandr/jormungandr/street_network/with_parking.py
```diff
@@ -0,0 +1,114 @@
...
+        logger = logging.getLogger(__name__)
+        logger.debug("Creating parking section for direct path")
+        for journey in response.journeys:
+            section = journey.sections.add()
+            section.duration = self.parking_module.park_duration
+            section.type = response_pb2.PARK
+            section.id = 'section_1'
+            section.begin_date_time = journey.sections[0].end_date_time
+            section.end_date_time = section.begin_date_time + section.duration
+            journey.arrival_date_time = section.end_date_time
```
xlqian: you forgot `journey.duration += self.parking_module.park_duration`( you've done it in the last commit :) )

AurelienLP: Nice catch ! Just added it :)

source/jormungandr/jormungandr/street_network/with_parking.py
```diff
@@ -0,0 +1,128 @@
...
+
+import logging
+from jormungandr.street_network.street_network import AbstractStreetNetworkService
+from jormungandr import street_network, utils
+
+from jormungandr.street_network.asgard import Asgard
+from navitiacommon import response_pb2
+
+
+class WithParking(AbstractStreetNetworkService):
```
xlqian, line 39: it seems WithParking isn't tested yet

AurelienLP: Doing it in another PR for clarity

AurelienLP: Adding it right now

source/jormungandr/tests/direct_path_asgard_integration_tests.py
```diff
@@ -0,0 +1,162 @@
...
+    else:
+        return response_pb2.Response()
+
+
+def check_journeys(resp):
+    assert not resp.get('journeys') or sum([1 for j in resp['journeys'] if j['type'] == "best"]) == 1
+
+
+class MockAsgard(Asgard):
+    def __init__(
```
xlqian, line 109: minor:
```suggestion
    def __init__(*args, **kwargs):
        super(self).__init__(*args, **kwargs)
```

xlqian (approved):

source/jormungandr/tests/direct_path_asgard_integration_tests.py
AurelienLP: Doesn't work because of the interface. Thank you anyway ! :D

source/jormungandr/jormungandr/street_network/with_parking.py
```diff
@@ -0,0 +1,128 @@
...
+            journey.nb_sections += 1
+
+            # The origin and destination of the parking section is the destination of the journey
+            section.origin.CopyFrom(pt_object_destination)
+            section.destination.CopyFrom(pt_object_destination)
+            # And we have to complete the destination of the first section ourself
+            # Because Jormun does not do it afterwards
+            journey.sections[0].destination.CopyFrom(pt_object_destination)
+
+            section.duration = self.parking_module.park_duration
```
TeXitoi: the parking module must have a method taking the coords and the datetime that returns the park (resp leave) duration.

source/jormungandr/jormungandr/street_network/with_parking.py
```diff
@@ -0,0 +1,128 @@
...
+    def add_parking_time_in_routing_matrix(self, response, origins, destinations):
+        logger = logging.getLogger(__name__)
+        additionnal_parking_time = (
+            self.parking_module.park_duration
+            if len(origins) == 1
+            else self.parking_module.leave_parking_duration
+        )
+        logger.debug("Adding additionnal parking time for routing_matrix " + str(additionnal_parking_time))
+        for r in response:
+            r.duration += additionnal_parking_time
```
TeXitoi: As the time will depends on the coords, you'll have to call it here. Maybe it can be interesting to have a batch method on the parking module, but we can see that later.

xlqian (approved):

TeXitoi (approved):

PR #2663, opened 2019-01-28
PTref: Fix date filter in disruptions

kadhikari:
1. For a coverage like stif (IDFM), data production period is usually few weeks and we may  have many  disruptions with application periods in future out of data production period.  

2. When we use api /disruptions with/without pt_ref, we get all the disruptions (includings those in future out of data production period), but when we use parameters &since and &until, it is only possible to get disruptions having application periods intersecting production_period. We also have an error message "Error: ptref : invalid filtering period, not in production period" when we use &since and &until with values out of production period. 

3. Correction with some existing tests modified and new added.

4. Concerns: https://jira.kisio.org/browse/NAVITIAII-2648

### DIFF
--- a/source/jormungandr/tests/check_utils.py
+++ b/source/jormungandr/tests/check_utils.py
@@ -1379,7 +1379,4 @@ def get_schedule(scs, sp_uri, line_code):
 
 
 def get_disruption(disruptions, disrupt_id):
-    for d in disruptions:
-        if d['id'] == disrupt_id:
-            return d
-    return None
+    return next((d for d in disruptions if d['id'] == disrupt_id), None)
--- a/source/jormungandr/tests/check_utils.py
+++ b/source/jormungandr/tests/check_utils.py
@@ -1376,3 +1376,10 @@ def get_schedule(scs, sp_uri, line_code):
             if rp_sched['stop_point']['id'] == sp_uri and rp_sched['route']['line']['code'] == line_code
         )
     ]
+
+
+def get_disruption(disruptions, disrupt_id):
+    for d in disruptions:
+        if d['id'] == disrupt_id:
+            return d
+    return None
--- a/source/ptreferential/ptreferential_utils.cpp
+++ b/source/ptreferential/ptreferential_utils.cpp
@@ -221,27 +221,21 @@ filter_on_period(const Indexes& indexes,
     if (since && until && until < since) {
         throw ptref_error("invalid filtering period");
     }
-    auto start = bt::ptime(data.meta->production_date.begin());
-    auto end = bt::ptime(data.meta->production_date.end());
-
-    if (since) {
-        if (data.meta->production_date.is_before(since->date()) && requested_type != nt::Type_e::Impact) {
-            throw ptref_error("invalid filtering period, not in production period");
-        }
-        if ((since->date() >= data.meta->production_date.begin()) || (requested_type == nt::Type_e::Impact)) {
-            start = *since;
-        }
+    auto start = bt::ptime(bt::neg_infin);
+    auto end = bt::ptime(bt::pos_infin);
+    if (requested_type == nt::Type_e::VehicleJourney) {
+        start = bt::ptime(data.meta->production_date.begin());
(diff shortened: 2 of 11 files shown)


source/ptreferential/ptreferential_utils.cpp
```diff
@@ -221,27 +221,21 @@ filter_on_period(const Indexes& indexes,
...
-        }
-        if (until->date() <= data.meta->production_date.last()) {
-            end = *until;
-        }
+    if (until && *until < end) {
+        end = *until + bt::seconds(1);
     }
     // we want end to be in the period, so we add one seconds
-    bt::time_period period {start, end + bt::seconds(1)};
+    bt::time_period period {start, end};
```
nonifier, line 236: The comment above now is irrelevant. 
But it would be nice to comment the bloc 3 lines above, mentioning that `until` is inclusive ;)

source/jormungandr/tests/check_utils.py
```diff
@@ -1376,3 +1376,10 @@ def get_schedule(scs, sp_uri, line_code):
             if rp_sched['stop_point']['id'] == sp_uri and rp_sched['route']['line']['code'] == line_code
         )
     ]
+
+
+def get_disruption(disruptions, disrupt_id):
+    for d in disruptions:
+        if d['id'] == disrupt_id:
+            return d
+    return None
```
nonifier: In a more pythonic way, you could also use [list comprehension](https://docs.python.org/3/tutorial/datastructures.html#list-comprehensions) here :
````python
disruption = next((d for d in disruption if d['id']==disrupt_id), None)

**Review by nonifier — APPROVED** [review]

**Review by TeXitoi — APPROVED** [review]
OK for me after the correction suggested by @nonifier

File: source/ptreferential/ptreferential_utils.cpp [hunk 3]
```diff
@@ -221,27 +221,21 @@ filter_on_period(const Indexes& indexes,
     if (since && until && until < since) {
         throw ptref_error("invalid filtering period");
     }
-    auto start = bt::ptime(data.meta->production_date.begin());
-    auto end = bt::ptime(data.meta->production_date.end());
+    auto start = bt::ptime(bt::neg_infin);
+    auto end = bt::ptime(bt::pos_infin);
+    if (requested_type == nt::Type_e::VehicleJourney) {
```
pbougue: Isn't that block useless?

pbougue (dismissed):
Please also update https://github.com/CanalTP/navitia/blob/dev/documentation/rfc/disruptions_periods.md
At least mention production_period is not a limit, but maybe there's something clearer that just that to write/correct in the RFC.

source/ptreferential/ptreferential_utils.cpp
kadhikari: In case of vehicleJourney, the production period is used for the validity.

TeXitoi: The implementation for vehicle_journeys loop on the date between since and until.

TeXitoi:
@pbougue this RFC only talk about the status and if the disruption is displayed, not about since and until.

source/ptreferential/ptreferential_utils.cpp
TeXitoi: Maybe it's clearer to do that in the implementation of vehicle_journeys.

kinnou02:
I don't understand what the correction is, you explain the problem but not the fix.

also we do not load from database disruption that don't intersect the production period, so we should never have them

kadhikari:
@pbougue  i modified the doc you mentioned by reducing production period in the diagram.
Any suggestion is welcome

kinnou02 (approved):

nonifier (approved):

pbougue (approved):

nonifier (approved):

pbougue (approved):

TeXitoi (changes requested):
https://github.com/CanalTP/navitia/pull/2663/files#r251828273 must be handled, else OK for me.

TeXitoi (approved):

PR #2665, opened 2019-01-29
Jormun: do not cache coverage search by coord

kinnou02:
Since `_all_keys_by_id` also work with coord's uri we were caching the result of `_all_keys_of_coord`. 
It isn't very efficient to cache (in redis) operation on coord as they will change at almost every call.

Also check that instance gotten from cache (potentially) exist before using it


xlqian (approved):

nonifier (approved):

pbougue (approved):
Could not think of an easy/smart way to test that.

PR #2662, opened 2019-01-28
Jormungandr: rename pt_dataset to pt_dataset[]

antoine-de:
mimir multi parameters handling was very lenient, it was ok to do:

  * `pt_dataset[]=toto&pt_dataset[]=tata`
  * `pt_dataset=toto&pt_dataset=tata`
  * `pt_dataset=[toto,tata]`

with https://github.com/CanalTP/mimirsbrunn/pull/286 only `pt_dataset[]=toto&pt_dataset[]=tata` will be accepted.

In navitia `type[]` was passed as `type[]` but `pt_dataset` had no `[]`.

This PR renames all `pt_dataset` to `pt_dataset[]` for coherence and mimir compatibility

PR #2667, opened 2019-01-30
Jormun: fix bragi multiple parameters

antoine-de:
fix #2662
in #2662 `pt_dataset` has been renamed `pt_dataset[]` but all the bragi multiple parameters were still given as `param[]=[value1,value2]`

This will no longer be supported, so this PR changes all those parameters `pt_dataset[]` and `type[]` to
`param[]=value1&param[]=value2`

### DIFF
--- a/source/jormungandr/jormungandr/interfaces/v1/test/elastic_search_tests.py
+++ b/source/jormungandr/jormungandr/interfaces/v1/test/elastic_search_tests.py
@@ -504,7 +504,7 @@ def bragi_call_test():
         ]
     }
     mock_requests = MockRequests(
-        {'http://bob.com/autocomplete?q=rue+bobette&limit=10&timeout=2000': (bragi_response, 200)}
+        {'http://bob.com/autocomplete?limit=10&q=rue+bobette&timeout=2000': (bragi_response, 200)}
     )
     from jormungandr import app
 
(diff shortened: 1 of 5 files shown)


pbougue (approved):
Uglier but working :smile: 
No test on multiple datasets, but as you do it for type[]...

antoine-de:
yes there are some integration tests: https://github.com/CanalTP/navitia/blob/6ee11806b5e5c7e6703ece41b3a48447f7a86563/source/jormungandr/tests/authentication_tests.py#L543

and I also added one unit test  too: https://github.com/CanalTP/navitia/blob/943cf2c393b7ca02bd51ab4c9f6c4765d2632c9f/source/jormungandr/jormungandr/interfaces/v1/test/elastic_search_tests.py#L564

source/jormungandr/jormungandr/interfaces/v1/test/elastic_search_tests.py
```diff
@@ -540,11 +540,32 @@ def bragi_make_params_with_instance_test():
...
 
     params = bragi.make_params(request=request, instances=[instance], timeout=1)
-    rsp = {'q': 'aa', 'limit': 20, 'pt_dataset[]': ['bib'], 'timeout': 1000}
-    len(list(rsp.keys())) == len(list(params.keys()))
-    for key, value in rsp.items():
-        assert key in params
-        assert value == params[key]
+    rsp = {'q': 'aa', 'limit': 20, 'pt_dataset[]': 'bib', 'timeout': 1000}
+
+    len(rsp) == len(params)
```
xlqian: ?
```suggestion
    assert len(rsp) == len(params) 
```

besides, what you want is this: `assert rsp == dict(params)`, it works only when there is only one instance

AND If i understand your PR well, shouldn't `resp`  be a list ??

source/jormungandr/jormungandr/interfaces/v1/test/elastic_search_tests.py
```diff
@@ -504,7 +504,7 @@ def bragi_call_test():
         ]
     }
     mock_requests = MockRequests(
-        {'http://bob.com/autocomplete?q=rue+bobette&limit=10&timeout=2000': (bragi_response, 200)}
+        {'http://bob.com/autocomplete?limit=10&q=rue+bobette&timeout=2000': (bragi_response, 200)}
```
xlqian, line 507: just curious: why you switched `limit` and `q` ? does that matter?

source/jormungandr/jormungandr/interfaces/v1/test/elastic_search_tests.py
```diff
@@ -557,10 +578,10 @@ def bragi_make_params_without_instance_test():
 
     params = bragi.make_params(request=request, instances=[], timeout=0.1)
     rsp = {'q': 'aa', 'limit': 20, 'timeout': 100}
-    assert len(list(rsp.keys())) == len(list(params.keys()))
-    for key, value in rsp.items():
-        assert key in params
-        assert value == params[key]
+    len(rsp) == len(params)
```
xlqian: assert ?

xlqian: and rsp should be a list

source/jormungandr/jormungandr/interfaces/v1/test/elastic_search_tests.py
antoine-de: yes, as `MockRequest` is not a real api but does a dumb lookup in a dict the parameters must be sorted (since we compare the strings).

source/jormungandr/jormungandr/interfaces/v1/test/elastic_search_tests.py
antoine-de: oups for the `assert` :man_facepalming: 

resp is there only to test the returned values.

If I use a list for rsp I will also have sorting problems for comparing to `params` thus to dict + lookup compare.

But if you want I can also change it to a list + sort both list. what do you think ?

xlqian (approved):

nonifier (approved):

PR #2668, opened 2019-02-04
Kraken: rebuild previous raptor cache in maintenance worker after applying realtime

kinnou02:
With this PR after having apply some disruption on data we rebuild the
raptor cache that are currently in use in the current_data.
This should greatly improve performance of kraken with realtime data.

I'm not a fan of the implementation: it's totally untestable, like almost
every that is done in maintenance worker.
Also, as usual it's not clear who should have the responsibility of
building this cache:
  - maintenance_worker alone
  - the cache itself
  - data_manager
  - data & dataRaptor (currently implemented)

If someone as a strong opinion on that I listening...

I don't really see any drawback of always restoring the previous cache,
it may increase the memory usage a bit, but it has already been
allocated one time, there is no reason it won't be a second time.

Require https://github.com/CanalTP/utils/pull/75

update on the performance impact, first response time of journeys while loading realtime data as soon as possible on transilien's coverage with the v2.71.0:
image

Same things with this PR:
image
the 99th centile is improved by almost 10 time on pic.

### DIFF
--- a/source/routing/next_stop_time.h
+++ b/source/routing/next_stop_time.h
@@ -1,28 +1,28 @@
 /* Copyright © 2001-2015, Canal TP and/or its affiliates. All rights reserved.
-  
+
 This file is part of Navitia,
     the software to build cool stuff with public transport.
- 
+
 Hope you'll enjoy and contribute to this project,
     powered by Canal TP (www.canaltp.fr).
 Help us simplify mobility and open public transport:
     a non ending quest to the responsive locomotion way of traveling!
-  
+
 LICENCE: This program is free software; you can redistribute it and/or modify
 it under the terms of the GNU Affero General Public License as published by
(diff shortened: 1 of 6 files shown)


source/routing/next_stop_time.h
```diff
@@ -277,6 +277,12 @@ struct CachedNextStopTimeManager {
          const type::RTLevel rt_level,
          const type::AccessibiliteParams& accessibilite_params);
 
+    std::shared_ptr<const CachedNextStopTime> load(const CachedNextStopTimeKey& key);
```
TeXitoi: Useless, no?

kinnou02: Yes, I have removed it

TeXitoi (approved):

PR #2670, opened 2019-02-05
Kraken: Fix realtime test (on ubuntu 16.04)

benoit-bst:
Currently, _realtime tests_ failed on **Ubuntu 16.04** but not on **Debian 8** and **Ubuntu 18**.
In fact, it is the `impact list order`, because it depends on protobuf version.

Impact list is coded with  :
`google::protobuf::RepeatedPtrField< ::pbnavitia::Impact >`

The test was develop on _ubuntu 18_. This is the reason why we didn't see it on ubuntu 16.04...

:heavy_check_mark:  I fixed and tested on **Ubuntu 16.04** (native machine) and **debian 8** (docker).

### DIFF
--- a/source/disruption/tests/disruption_test.cpp
+++ b/source/disruption/tests/disruption_test.cpp
@@ -105,7 +105,7 @@ namespace {
         navitia::apply_disruption( disruption, *b.data->pt_data, *b.data->meta);
     }
 
-    std::set<std::string> get_impacts_uris(const std::set<Impact::SharedImpact> & impacts)
+    std::set<std::string> get_impacts_uris(const std::set<Impact::SharedImpact, Less> & impacts)
     {
         std::set<std::string> uris;
         boost::range::transform(impacts,
--- a/source/kraken/tests/realtime_test.cpp
+++ b/source/kraken/tests/realtime_test.cpp
@@ -75,10 +75,10 @@ make_cancellation_message(const std::string& vj_uri, const std::string& date) {
 
 static pbnavitia::Response compute_iti(
     const ed::builder& b,
-    const char* datetime, 
-    const std::string& from, 
+    const char* datetime,
+    const std::string& from,
     const std::string& to,
-    const navitia::type::RTLevel level) 
+    const navitia::type::RTLevel level)
 {
     navitia::type::EntryPoint origin(b.data->get_type_of_id(from), from);
     navitia::type::EntryPoint destination(b.data->get_type_of_id(to), to);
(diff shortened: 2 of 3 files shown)


kadhikari (approved):

TeXitoi:
Why is there a difference on the order of the impacts?

kinnou02 (approved):

benoit-bst:
@TeXitoi, I don't known why, but the fact is that it happen. 
I suppose there is a difference into `google::protobuf::RepeatedPtrField` in the protobuf lib.
Anyway, before it was not safe in the test. We have to use an `id` to access a specific element.

TeXitoi (changes requested):
I'd like that we understand why. There is maybe just somewhere we sort on a pointer value, maybe something worst.

TeXitoi:
`google::protobuf::RepeatedPtrField` is a vector, the difference is somewhere else.

benoit-bst:
**To sum up** : 

- `google::protobuf::RepeatedPtrField` is a vector, so the order is defined and unchanged
- There is a tuned `Less` comparator on Impact.
  In message.cpp:383 :
```c++
bool Impact::operator<(const Impact& other){                                                                                                                                                                                                                                   
       if(this->severity->priority != other.severity->priority){                                                                                                                                                                                                                  
           return this->severity->priority < other.severity->priority;                                                                                                                                                                                                            
       }if(this->created_at != other.created_at){                                                                                                                                                                                                                                 
           return this->created_at < other.created_at;                                                                                                                                                                                                                            
       }else{                                                                                                                                                                                                                                                                     
           return this->uri < other.uri;                                                                                                                                                                                                                                          
       }                                                                                                                                                                                                                                                                          } 
```
   Thus, We compare exactly in the same way `impacts`
- Ubuntu 16.04 order example for 3 impacts that currently failed in the test:
```c++
// print impacts in realtime_test.cpp:2397
// res = compute("20171101T073000", "stop_point:A", "stop_point:C");
// for (const auto& impact: res.impacts()) {
//     std::cout << impact.uri() << std::endl;
// }
feed-1
feed-2
feed-3
``` 
- Debian 8 order example for 3 impacts that currently passed in the test:
```c++
// print impacts in realtime_test.cpp:2397
// res = compute("20171101T073000", "stop_point:A", "stop_point:C");
// for (const auto& impact: res.impacts()) {
//     std::cout << impact.uri() << std::endl;
// }
feed-2
feed-3
feed-1

``` 

**Conclusion** :
We still have a difference between OS... For the moment i'm stuck :imp:

benoit-bst:
Aftermath invesgation, it results than in the code (pb_converter.h:176), the `std:set<...>Impact` needs a tuned `Less Comparator` (It was a lack). 
Now, for every OS, we have the good order : 
```
feed-1
feed-2
feed-3
```

End of the research :fireworks:

TeXitoi (approved):

benoit-bst:
Jenkins, test this please !

PR #2671, opened 2019-02-06
Doc: 'distance' field from 'Place' object in the swagger

nonifier:
The SDK's team has reported a missing field in the swagger... Let's make them happy :)
```
"Place" : {
         "properties" : {
+          "distance" : {
+             "type" : "integer",
+             "description" : "Distance to the object in meters"
            },
            [...]
         },
         "type" : "object",
         "required" : [
            "name",
            "id"
         ]
      },
```

poke: @Guitouille  :)

### DIFF
--- a/source/jormungandr/jormungandr/interfaces/v1/serializer/pt.py
+++ b/source/jormungandr/jormungandr/interfaces/v1/serializer/pt.py
@@ -399,10 +399,10 @@ class PlaceSerializer(PbGenericSerializer):
     stop_area = StopAreaSerializer(display_none=False)
     stop_point = StopPointSerializer(display_none=False)
     administrative_region = AdminSerializer(display_none=False)
-    embedded_type = EnumField(attr='embedded_type', required=False, pb_type=NavitiaType, display_none=True)
+    embedded_type = EnumField(attr='embedded_type', pb_type=NavitiaType, display_none=True)
     address = AddressSerializer(display_none=False)
     poi = PoiSerializer(display_none=False)
-    distance = jsonschema.IntField(required=False, display_none=True, description='Distance to the object in meters')
+    distance = jsonschema.StrField(required=False, display_none=True, description='Distance to the object in meters')
 
 class PlaceNearbySerializer(PlaceSerializer):
     pass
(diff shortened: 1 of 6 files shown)


source/jormungandr/jormungandr/interfaces/v1/serializer/pt.py
```diff
@@ -399,14 +399,13 @@ class PlaceSerializer(PbGenericSerializer):
     stop_area = StopAreaSerializer(display_none=False)
     stop_point = StopPointSerializer(display_none=False)
     administrative_region = AdminSerializer(display_none=False)
-    embedded_type = EnumField(attr='embedded_type', pb_type=NavitiaType, display_none=True)
+    embedded_type = EnumField(attr='embedded_type', required=False, pb_type=NavitiaType, display_none=True)
```
kinnou02: Seems strange to not have an `embedded_type` in places? when does it happen?

source/jormungandr/jormungandr/interfaces/v1/serializer/pt.py
```diff
@@ -399,14 +399,13 @@ class PlaceSerializer(PbGenericSerializer):
...
     address = AddressSerializer(display_none=False)
     poi = PoiSerializer(display_none=False)
-
+    distance = jsonschema.IntField(required=False, display_none=True, description='Distance to the object in meters')
```
kinnou02: In `PlaceNearby` it was a str (yes that's ugly...). Remember that we can't change the interface, so if it was an `str` before it must stay one.
In `/place_nearby` it's an str:
```
{
    "distance": "104",
    "embedded_type": "poi",
    "id": "poi:osm:node:3889022681"
    "name": "Parking vélo (Paris)",
    "poi": ⊕{8 items},
    "quality": 0,
}
```

In `/places` it's an int...
```
{
    "address": ⊕{6 items},
    "distance": 4651,
    "embedded_type": "address",
    "id": "2.37715;48.846781"
    "name": "20 Rue Hector Malot (Paris)",
    "quality": 0,
}
```

kinnou02 (changes requested):

kinnou02 (approved):

kinnou02 (approved):

source/jormungandr/tests/bragi_autocomplete_tests.py
```diff
@@ -566,7 +566,7 @@ def test_autocomplete_call(self):
             assert r[0]['embedded_type'] == 'address'
             assert r[0]['address']['name'] == 'Rue Bob'
             assert r[0]['address']['label'] == '20 Rue Bob (Bobtown)'
```
pbougue, line 568: maybe add a test on `id` as you overload it?

pbougue (approved):
It's probably (too) complex: Can you add swagger tests on both bragi and kraken cases?
Maybe around here: https://github.com/CanalTP/navitia/blob/dev/source/jormungandr/tests/schema_tests.py#L215
This would help ensure both change at the same time.

PR #2673, opened 2019-02-11
Jormun: experimentation with mypy

kinnou02:
Add some typing to python with [mypy](https://github.com/python/mypy)

Most of the change are due to flask not liking unicode object.
I added type to a few functions mostly to test a bit.

The `typing` package isn't requirement at runtime since it's only used in comments, I added it to the requirements to be safe.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,7 @@ python:
   - "2.7"
   - "3.6"
 sudo: required
+cache: pip
 
 services:
   - docker
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,7 +30,12 @@ script:
 matrix:
   include:
     - python: "3.6"
-      before_script: python3.6 -m pip install black==18.9b0;
-      script:  black --check --diff .
+      before_script:
+        - python3.6 -m pip install black==18.9b0
+        - pip install -r source/jormungandr/requirements_dev.txt
+        - bash source/scripts/build_protobuf.sh
+      script:
+        - black --check --diff .
+        - pushd source/jormungandr && MYPYPATH=.:../navitiacommon mypy --ignore-missing-imports --py2 jormungandr && popd
   exclude:
     - python: "3.6"
--- a/source/jormungandr/jormungandr/instance.py
+++ b/source/jormungandr/jormungandr/instance.py
@@ -30,6 +30,11 @@
 # www.navitia.io
 
 from __future__ import absolute_import, print_function, unicode_literals, division
+
+try:
+    from typing import Dict, Text, Any, Union
+except ImportError:
+    pass
 from contextlib import contextmanager
 from threading import Lock
 from flask_restful import abort
@@ -58,14 +63,14 @@ import time
 from collections import deque
 
 type_to_pttype = {
-    "stop_area": request_pb2.PlaceCodeRequest.StopArea,
-    "network": request_pb2.PlaceCodeRequest.Network,
-    "company": request_pb2.PlaceCodeRequest.Company,
-    "line": request_pb2.PlaceCodeRequest.Line,
-    "route": request_pb2.PlaceCodeRequest.Route,
-    "vehicle_journey": request_pb2.PlaceCodeRequest.VehicleJourney,
-    "stop_point": request_pb2.PlaceCodeRequest.StopPoint,
-    "calendar": request_pb2.PlaceCodeRequest.Calendar,
+    "stop_area": request_pb2.PlaceCodeRequest.StopArea,  # type: ignore
+    "network": request_pb2.PlaceCodeRequest.Network,  # type: ignore
+    "company": request_pb2.PlaceCodeRequest.Company,  # type: ignore
+    "line": request_pb2.PlaceCodeRequest.Line,  # type: ignore
+    "route": request_pb2.PlaceCodeRequest.Route,  # type: ignore
+    "vehicle_journey": request_pb2.PlaceCodeRequest.VehicleJourney,  # type: ignore
+    "stop_point": request_pb2.PlaceCodeRequest.StopPoint,  # type: ignore
+    "calendar": request_pb2.PlaceCodeRequest.Calendar,  # type: ignore
 }
 
 STREET_NETWORK_MODES = ('walking', 'car', 'bss', 'bike', 'ridesharing')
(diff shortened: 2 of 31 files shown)


mbounabi (commented):
Why?
If it helps the python3 migration, why not?

kinnou02:
> Why?
> If it helps the python3 migration, why not?

The python3 migration is one the motivation, more check on the code will help and a lot of the problems introduce by python3 are related to not (really) retrocompatible change around the `str` type.
Also, most of us are quite fond of typing, and we aren't confident when doing refactoring without having a compiler to cry at us.

source/jormungandr/jormungandr/instance.py
```diff
@@ -93,11 +98,14 @@ def _set_default_street_network_config(street_network_configs):
 
 
 class Instance(object):
+    name = None  # type: Text
```
pbougue, line 101: Why isn't that unicode?

source/jormungandr/jormungandr/instance.py
```diff
@@ -58,14 +63,14 @@
...
 type_to_pttype = {
-    "stop_area": request_pb2.PlaceCodeRequest.StopArea,
-    "network": request_pb2.PlaceCodeRequest.Network,
-    "company": request_pb2.PlaceCodeRequest.Company,
-    "line": request_pb2.PlaceCodeRequest.Line,
-    "route": request_pb2.PlaceCodeRequest.Route,
-    "vehicle_journey": request_pb2.PlaceCodeRequest.VehicleJourney,
-    "stop_point": request_pb2.PlaceCodeRequest.StopPoint,
-    "calendar": request_pb2.PlaceCodeRequest.Calendar,
+    "stop_area": request_pb2.PlaceCodeRequest.StopArea,  # type: ignore
```
pbougue, line 66: Here type is ignored because mypy would want a consistent type?
Should we use a parent of them all (well, probably object in that case)?

source/jormungandr/jormungandr/instance.py
```diff
@@ -217,111 +225,133 @@ def replace_experimental_scenario(s):
...
 
     @property
     def autocomplete_backend(self):
+        # type: () -> Text
         instance_db = self.get_models()
         return get_value_or_default('autocomplete_backend', instance_db, self.name)
 
     @property
     def max_walking_duration_to_pt(self):
+        # type: () -> int
```
pbougue, line 240: Just curious: Would it be possible to add it above, not in comment?
`def max_walking_duration_to_pt(self) -> int:`

source/navitiacommon/navitiacommon/models/__init__.py
```diff
@@ -107,7 +107,7 @@ def __unicode__(self):
         return self.value
 
 
-class User(db.Model):
+class User(db.Model):  # type: ignore
```
pbougue, line 110: Why all those ignore?

source/jormungandr/jormungandr/instance.py
```diff
@@ -169,8 +177,8 @@ def get_models(self):
     def __repr__(self):
         return 'instance.{}'.format(self.name)
 
-    @memory_cache.memoize(app.config['MEMORY_CACHE_CONFIGURATION'].get('TIMEOUT_PARAMS', 30))
-    @cache.memoize(app.config['CACHE_CONFIGURATION'].get('TIMEOUT_PARAMS', 300))
+    @memory_cache.memoize(app.config[str('MEMORY_CACHE_CONFIGURATION')].get(str('TIMEOUT_PARAMS'), 30))
```
pbougue, line 180: So it's tracked by mypy "natively" that flask config is using str?
If so: :tada: great !

pbougue (commented):
I'm worried about all those `ignore` mainly.

source/jormungandr/jormungandr/instance.py
kinnou02: `Text` is `unicode` in python2 and `str` in python3

source/jormungandr/jormungandr/instance.py
kinnou02: There is no stub for protobuf so mypy as some problems with it.
I need to look at https://github.com/dropbox/mypy-protobuf.
The main objective of this PR was to introduce mypy and make test pass

source/jormungandr/jormungandr/instance.py
kinnou02: This syntax is only available on python 3, so no.

source/jormungandr/jormungandr/instance.py
kinnou02: Since the last version of mypy there is [stubs](https://mypy.readthedocs.io/en/latest/stubs.html) for flask.

source/navitiacommon/navitiacommon/models/__init__.py
kinnou02: `flask_sqlalchemy` and `sqlalchemy` doesn't use mypy, so it doesn't have a lot of information on `db.Model`, and these libs are full of magic: if you look closely you will see that `db` is already an instance of a class, and `Model` is an attribute of this instance that is dynamically constructed. So this construct is a bit strange and mypy wasn't liking it.

As I said before the point of this PR is to test mypy and see if we want to use it. If so we will probably have generate some stubs for our dependencies, there is https://github.com/dropbox/sqlalchemy-stubs out there that I didn't test yet.

kinnou02:
We should talk about this Monday

.travis.yml
```diff
@@ -30,7 +31,12 @@ script:
...
     - python: "3.6"
-      before_script: python3.6 -m pip install black==18.9b0;
-      script:  black --check --diff .
+      before_script:
+        - python3.6 -m pip install black==18.9b0
+        - pip install -r source/jormungandr/requirements_dev.txt
+        - bash source/scripts/build_protobuf.sh
+      script:
+        - black --check --diff .
+        - pushd source/jormungandr && MYPYPATH=.:../navitiacommon mypy --ignore-missing-imports --py2 jormungandr && popd
```
nonifier, line 40: :+1:

nonifier (approved):
Type checking :heart: :heart: :heart: 
Much safer than.... Duck-typing

source/jormungandr/jormungandr/instance.py
pbougue: Ah, so it's ambiguous in between and a new name but it's unicode all the way.
OK  :-|

pbougue (approved):

benoit-bst (approved):

PR #2674, opened 2019-02-12
jormun: remove marshal first step

kinnou02:
This PR remove all the code that wasn't used anymore since we are always
in serpy mode. it doesn't remove the marshaller yet, so nothing should
break if they are still used somewhere.
This also remove all tests with marshal :tada:


kinnou02:
poke @eturck

mbounabi (approved):

TeXitoi (approved):

nonifier (approved):
:fireworks:

PR #2675, opened 2019-02-12
[Kraken] - UTests around clean_useless_vjs func

benoit-bst:
There was no Utests around `clean_useless_vjs` function (https://github.com/CanalTP/navitia/blob/dev/source/type/type.cpp#L527)...
This function aims to clean All VJs with _validity patterns = 0_ before adding a new one.

That's increase tests strongness and prepare for the new development (Add new trip in realtime mode).

### DIFF
--- a/source/ed/build_helper.cpp
+++ b/source/ed/build_helper.cpp
@@ -56,7 +56,8 @@ VJ::VJ(builder& b,
        const std::string& physical_mode,
        const uint32_t start_time,
        const uint32_t end_time,
-       const uint32_t headway_secs):
+       const uint32_t headway_secs,
+       const nt::RTLevel vj_type):
     b(b),
     network_name(network_name),
     line_name(line_name),
@@ -69,6 +70,7 @@ VJ::VJ(builder& b,
     start_time(start_time),
     end_time(end_time),
     headway_secs(headway_secs),
+    _vj_type(vj_type),
     _vp(b.begin, validity_pattern)
 {}
 
@@ -180,14 +182,14 @@ nt::VehicleJourney* VJ::make() {
         "vj:" + line_name + ":" + std::to_string(pt_data.vehicle_journeys.size()) :
         _uri;
     if (is_frequency) {
-        auto* fvj = mvj->create_frequency_vj(uri_str, nt::RTLevel::Base, _vp, route, sts, pt_data);
+        auto* fvj = mvj->create_frequency_vj(uri_str, _vj_type, _vp, route, sts, pt_data);
         fvj->start_time = start_time;
         const size_t nb_trips = std::ceil((end_time - start_time) / headway_secs);
         fvj->end_time = start_time + (nb_trips * headway_secs);
         fvj->headway_secs = headway_secs;
         vj = fvj;
     } else {
-        vj = mvj->create_discrete_vj(uri_str, nt::RTLevel::Base, _vp, route, sts, pt_data);
+        vj = mvj->create_discrete_vj(uri_str, _vj_type, _vp, route, sts, pt_data);
     }
     // default dataset
     if (!vj->dataset){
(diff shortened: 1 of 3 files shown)


nonifier (approved):

source/ed/build_helper.cpp
```diff
@@ -69,6 +70,7 @@ VJ::VJ(builder& b,
     start_time(start_time),
     end_time(end_time),
     headway_secs(headway_secs),
+    _vj_type(vj_type),
```
pbougue: Let's save "`_`private_attributes" for python :pray: 
(if you can remove the one below it's nice too)

source/type/tests/create_vj_test.cpp
```diff
@@ -95,3 +95,123 @@ BOOST_AUTO_TEST_CASE(create_vj_test) {
...
+    // Inside the MVJ, we add 2 Realtime VJ
+    // There is no base VJ inside (Usefull for adding a new trip in Realtime mode)
+    ed::builder b("20120614");
+    auto& pt_data = *b.data->pt_data;
+
+    const auto* rt_vj = b.vj("line_A", "000011", "block_id_A", false, "uri_C", "meta_vj_name_A", "", nt::RTLevel::RealTime)("stop1", 8000, 8000)("stop2", 8100, 8100).make();
+    BOOST_CHECK_EQUAL(rt_vj->base_validity_pattern()->days, year("000000"));
+    BOOST_CHECK_EQUAL(rt_vj->adapted_validity_pattern()->days, year("000000"));
+    BOOST_CHECK_EQUAL(rt_vj->rt_validity_pattern()->days, year("000011"));
+    rt_vj = b.vj("line_A", "001100", "block_id_A", false, "uri_C", "meta_vj_name_A", "", nt::RTLevel::RealTime)("stop1", 8000, 8000)("stop2", 8100, 8100).make();
```
pbougue: It would be a bit easier to understand with different names: rt_vj`2` and uri_`D`

source/type/tests/create_vj_test.cpp
```diff
@@ -95,3 +95,123 @@ BOOST_AUTO_TEST_CASE(create_vj_test) {
...
+    using year = navitia::type::ValidityPattern::year_bitset;
+    namespace nt = navitia::type;
+
+    // Create a route with a Meta VJ.
+    // Inside the MVJ, we add 2 Realtime VJ
+    // There is no base VJ inside (Usefull for adding a new trip in Realtime mode)
+    ed::builder b("20120614");
+    auto& pt_data = *b.data->pt_data;
+
+    const auto* rt_vj = b.vj("line_A", "000011", "block_id_A", false, "uri_C", "meta_vj_name_A", "", nt::RTLevel::RealTime)("stop1", 8000, 8000)("stop2", 8100, 8100).make();
```
pbougue: Idea: try to create a base_vj with nothing in VP on that same MetaVJ, and check if it stays along the way.
This may require to create another useless VJ on another MetaVJ before the 2 rt_vj, to triger cleanup and really test.

source/ed/build_helper.cpp
benoit-bst: There are different schools :smile:. It can be applicable for c ++ too. 
But in the scope of this class, it is not the main convention. So, I change that

source/type/tests/create_vj_test.cpp
benoit-bst: nice catch

source/type/tests/create_vj_test.cpp
benoit-bst: if you have a VJ creation with all VP = 0, the clean function will suppress it after tha fact. 
But, it is on the same meta VJ. The explication is simple, `clean_useless_vj` is a meta_vj's method. I can only clean useless vj, only on the same meta VJ.
In my opinion, there are enough tests.

source/ed/build_helper.cpp
benoit-bst: but i'll not clean all internal vars with this convention.

pbougue: That's fine, it would be nice to do it someday though, as it's not the convention in kraken.

source/type/tests/create_vj_test.cpp
pbougue: Yep, for now it won't disappear but if you use the method implemented in utils it probably will.
That's just an idea, but probably better when you use cleanup.
> edit: mentionning `erase` method from object_factory, here.

And I was curious about behavior of empty MetaVJ... Not sure a test would challenge those enough though.

pbougue (approved):

source/type/tests/create_vj_test.cpp
benoit-bst: Maybe in the future, we'll move `clean_useless_vj` in utils. Good idea.

PR #2678, opened 2019-02-14
 Jormun: deletion of marshal

kinnou02:
Most of the marshal code should be gone with this PR

First commit is from #2674


source/jormungandr/jormungandr/autocomplete/kraken.py
```diff
@@ -34,31 +34,14 @@
...
-    "places": NonNullList(NonNullNested(place)),
-    "error": PbField(error, attribute='error'),
-    "disruptions": fields.List(NonNullNested(disruption_marshaller), attribute="impacts"),
-    "feed_publishers": fields.List(NonNullNested(feed_publisher)),
-}
-
-
 class Kraken(AbstractAutocomplete):
-    @get_serializer(serpy=api.PlacesSerializer, marshall=places)
+    @get_serializer(serpy=api.PlacesSerializer)
```
mbounabi, line 44: The function get_serializer() in decorators.py still has the parameter 'marshall'. If it isn't used anymore, it can be removed.

kinnou02: I removed the `marshall` parameter, but I kept the functions

mbounabi (approved):

nonifier (approved):

PR #2679, opened 2019-02-19
Kraken: Add enable_realtime_add_trip flag

benoit-bst:
New parameter to _activate/deactivate_ adding trip.

**JIRA** : https://jira.kisio.org/browse/NAVP-1148

### DIFF
--- a/source/kraken/configuration.cpp
+++ b/source/kraken/configuration.cpp
@@ -60,8 +60,8 @@ po::options_description get_options_description(const boost::optional<std::strin
                                         "enable loading of realtime data")
         ("GENERAL.is_realtime_add_enabled", po::value<bool>()->default_value(false),
                                         "enable loading of realtime data that add stop_times")
-        ("GENERAL.enable_realtime_add_trip", po::value<bool>()->default_value(false),
-                                        "enable loading of realtime data that add trips")
+        ("GENERAL.is_realtime_add_trip_enabled", po::value<bool>()->default_value(false),
+                                        "enable loading of realtime data that add trips (only if is_realtime_add_enabled is activated)")
         ("GENERAL.kirin_timeout", po::value<int>()->default_value(60000),
                                   "timeout in ms for loading realtime data from kirin")
         ("GENERAL.kirin_retry_timeout", po::value<int>()->default_value(5*60*1000),
@@ -167,8 +167,8 @@ bool Configuration::is_realtime_add_enabled() const{
     return this->vm["GENERAL.is_realtime_add_enabled"].as<bool>();
 }
 
-bool Configuration::enable_realtime_add_trip() const{
-    return this->vm["GENERAL.enable_realtime_add_trip"].as<bool>();
+bool Configuration::is_realtime_add_trip_enabled() const{
+    return this->vm["GENERAL.is_realtime_add_trip_enabled"].as<bool>();
 }
 
 int Configuration::kirin_timeout() const{
--- a/source/kraken/configuration.cpp
+++ b/source/kraken/configuration.cpp
@@ -59,7 +59,9 @@ po::options_description get_options_description(const boost::optional<std::strin
         ("GENERAL.is_realtime_enabled", po::value<bool>()->default_value(false),
                                         "enable loading of realtime data")
         ("GENERAL.is_realtime_add_enabled", po::value<bool>()->default_value(false),
-                                        "enable loading of realtime data that add stop_times or trips")
+                                        "enable loading of realtime data that add stop_times")
+        ("GENERAL.enable_realtime_add_trip", po::value<bool>()->default_value(false),
+                                        "enable loading of realtime data that add trips")
         ("GENERAL.kirin_timeout", po::value<int>()->default_value(60000),
                                   "timeout in ms for loading realtime data from kirin")
         ("GENERAL.kirin_retry_timeout", po::value<int>()->default_value(5*60*1000),
@@ -163,6 +165,10 @@ bool Configuration::is_realtime_enabled() const{
 
 bool Configuration::is_realtime_add_enabled() const{
     return this->vm["GENERAL.is_realtime_add_enabled"].as<bool>();
+}
+
+bool Configuration::enable_realtime_add_trip() const{
+    return this->vm["GENERAL.enable_realtime_add_trip"].as<bool>();
 }
 
 int Configuration::kirin_timeout() const{
--- a/source/kraken/realtime.cpp
+++ b/source/kraken/realtime.cpp
@@ -250,17 +250,23 @@ get_status(const transit_realtime::TripUpdate_StopTimeEvent& event,
     }
 }
 
-static bool is_added_service(const transit_realtime::TripUpdate& trip_update) {
+static bool is_added_trip(const transit_realtime::TripUpdate& trip_update) {
     namespace trt = transit_realtime;
     auto log = log4cplus::Logger::getInstance("realtime");
-    using nt::disruption::StopTimeUpdate;
 
-    // adding a trip is adding service
     if (trip_update.trip().schedule_relationship() == trt::TripDescriptor_ScheduleRelationship_ADDED) {
         LOG4CPLUS_TRACE(log, "Disruption has ADDITIONAL_SERVICE effect");
         return true;
     }
-    else if (trip_update.trip().schedule_relationship() == trt::TripDescriptor_ScheduleRelationship_SCHEDULED
+    return false;
+}
+
+static bool is_added_stop_time(const transit_realtime::TripUpdate& trip_update) {
+    namespace trt = transit_realtime;
+    auto log = log4cplus::Logger::getInstance("realtime");
+    using nt::disruption::StopTimeUpdate;
+
+    if (trip_update.trip().schedule_relationship() == trt::TripDescriptor_ScheduleRelationship_SCHEDULED
                 && trip_update.stop_time_update_size()) {
         for (const auto& st: trip_update.stop_time_update()) {
             // adding a stop_time event (adding departure or/and arrival) is adding service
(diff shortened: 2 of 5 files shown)


source/kraken/configuration.cpp
```diff
@@ -59,7 +59,9 @@ po::options_description get_options_description(const boost::optional<std::strin
         ("GENERAL.is_realtime_enabled", po::value<bool>()->default_value(false),
                                         "enable loading of realtime data")
         ("GENERAL.is_realtime_add_enabled", po::value<bool>()->default_value(false),
-                                        "enable loading of realtime data that add stop_times or trips")
+                                        "enable loading of realtime data that add stop_times")
```
pbougue, line 62: To me:
is_realtime_enabled == false -> is_realtime_add_enabled = false + enable_realtime_add_trip = false
and
is_realtime_add_enabled == false -> enable_realtime_add_trip = false

So that comment was OK.

source/kraken/configuration.cpp
```diff
@@ -59,7 +59,9 @@ po::options_description get_options_description(const boost::optional<std::strin
...
+        ("GENERAL.enable_realtime_add_trip", po::value<bool>()->default_value(false),
```
pbougue: I'd rather have the same naming convention (it will appear in conf also), so is_realtime_`trip_`add_enabled (or change the other ones, but it will require changing the conf also).
Then it will change other names also in this PR (to be checked exhaustively).

source/kraken/configuration.cpp
```diff
@@ -165,6 +167,10 @@ bool Configuration::is_realtime_add_enabled() const{
     return this->vm["GENERAL.is_realtime_add_enabled"].as<bool>();
 }
 
+bool Configuration::enable_realtime_add_trip() const{
+    return this->vm["GENERAL.enable_realtime_add_trip"].as<bool>();
```
pbougue: It would probably be safer to (untested):
```c++
return (is_realtime_add_enabled() && this->vm["GENERAL.enable_realtime_add_trip"].as<bool>());
```

source/kraken/configuration.cpp
```diff
@@ -165,6 +167,10 @@ bool Configuration::is_realtime_add_enabled() const{
     return this->vm["GENERAL.is_realtime_add_enabled"].as<bool>();
```
pbougue, line 167: It would probably be safer to (untested):
```c++
return (is_realtime_enabled() && this->vm["GENERAL.is_realtime_add_enabled"].as<bool>());
```

source/kraken/realtime.cpp
```diff
@@ -250,17 +250,23 @@ get_status(const transit_realtime::TripUpdate_StopTimeEvent& event,
...
-    // adding a trip is adding service
     if (trip_update.trip().schedule_relationship() == trt::TripDescriptor_ScheduleRelationship_ADDED) {
         LOG4CPLUS_TRACE(log, "Disruption has ADDITIONAL_SERVICE effect");
         return true;
     }
-    else if (trip_update.trip().schedule_relationship() == trt::TripDescriptor_ScheduleRelationship_SCHEDULED
+    return false;
+}
+
+static bool is_added_stop_time(const transit_realtime::TripUpdate& trip_update) {
```
pbougue, line 264: We must keep the fact that adding trip is adding stop_time.
So the the final `return false` should probably be `return is_added_trip`.

source/kraken/realtime.cpp
```diff
@@ -351,17 +357,12 @@ static const type::disruption::Disruption*
 create_disruption(const std::string& id,
                   const boost::posix_time::ptime& timestamp,
                   const transit_realtime::TripUpdate& trip_update,
-                  const type::Data& data,
-                  const bool is_realtime_add_enabled) {
```
pbougue, line 355: :+1:

pbougue (changes requested):
Please, can you also:

- [ ] prepare the related config PR also, so it is activated on dev and internal, maybe custo too (prod will be "do_not_merge" PR
- [ ] add tests on that
- [ ] add tests on receiving a trip add (should be rejected) then that's same trip removal (should be rejected too, and it's probably automatically handled when kraken doesn't know about that trip).
- [ ] check that kirin's trip.status stays the same for updates on that trip previously added (if it changes add tests on that)

You can probably get inspiration from https://github.com/CanalTP/navitia/pull/2581 and the PRs referencing it.

:warning: To precise a bit, my ":-1:" on responses below mean: "I think I understand the point, but I'm not convinced by the conclusion (no more argument to oppose though)"

source/kraken/realtime.cpp
benoit-bst: yes indeed. But this function only test if we have `adding stop_time + Scheduled`. 
It corresponds to the previous case (A simple adding stop time without new trip)
So, the function return `true` only if we detect a `ADDED/ADDED_FOR_DETOUR stop time`.

source/kraken/configuration.cpp
benoit-bst: `is_realtime_enable` is the main flag. It is handled higher. It blocks all realtime functions access.
Later in the code, if we have activate this flag we can have access to `handle_realtime function`, (with a realtime request).
At the beginning of the function `handle_realtime`, we check flag in this order.
1. `enable_realtime_add_trip` and if the realtime request correspond to this type. If it corresponds and the flag is deactivate, we return directly.
2. `is_realtime_add_enabled` and if the realtime request correspond to this type. If it corresponds and the flag is deactivate, we return directly.

https://github.com/CanalTP/navitia/pull/2679/files#diff-a1c1532aa9f44bfd3b8a77e208515e23R537

Then, 
to have an adding trip :
 `is_realtime_enabled == true + enable_realtime_add_trip = true`
to handle added stop time :
 `is_realtime_enabled == true + enable_realtime_add_enabled = true`
if  `is_realtime_enabled == false`, we handle anything

Am I clear ? One main flag and one flag by feature (add trip, add stop time). I think it is a simple rule and it works.
We can talk about it with the others team boys if it necessary.

source/kraken/configuration.cpp
benoit-bst: We have decided to call the parameter `enable_realtime_add_trip`. I follow the rules in the JIRA ticket :smile_cat:

source/kraken/configuration.cpp
benoit-bst: In this part of code, we simply load params from config. We don't want to insert logic. Only load the _raw value_ of the flag.

source/kraken/configuration.cpp
benoit-bst: In this part of code, we simply load params from config. We don't want to insert logic. Only load the _raw value_ of the flag.

source/kraken/realtime.cpp
pbougue: Then we won't use it directly, I guess.
To me, especially on something as touchy as what we're trying to do: the safest is the best.

Being atomic is nice, but the main goal is to be safe (of course both is even better).

If we ever get a trip add with only updates in it (if kirin ever sends it) we must not handle the feed if `is_realtime_add_enabled=false`. And there is no guarantee on Kirin (or even COTS) from doing that anytime.

source/kraken/configuration.cpp
pbougue: Also: we will probably use that param when triggering re-indexing and all realtime ptref creations, and we want to activate them only when the 3 params are `true`.
I want to secure it that way, because it doesn't seem to make any sense to handle the case where only `enable_realtime_add_trip=true` and the other false.

source/kraken/configuration.cpp
nonifier: As long as it doesn't carry crap around, I agree with PE that consistency is usually best keeping ! 
I would be tempted to name the option `is_realtime_add_trip_enabled`.

I just wonder if `is_realtime_add_enabled` still makes sense though, and would rather be renamed `is_realtime_add_stop_enabled` ?

source/kraken/configuration.cpp
nonifier: I agree with Benoit, for the sake of consistency ! I would rather see the logic in the code, not the configuration. It would give us more flexibility.

source/kraken/configuration.cpp
benoit-bst: Okay, so we need to update all deployment conf with that.

then, new flags are :
- is_realtime_enabled
- is_realtime_add_stop_enabled
- is_realtime_add_trip_enabled

Everyone is in agreement ?

source/kraken/realtime.cpp
benoit-bst: For this point, I don't understand your thought. The code is written in such a way that, if you have _is_realtime_add_enabled=false_, we'll not handle this kind of COTS files (add stop time)

with 
```
 if (! is_realtime_add_enabled && is_added_stop_time(trip_update)) {
￼         LOG4CPLUS_DEBUG(log, "Add stop time is disabled: ignoring it");
￼         return;
}
```

source/kraken/configuration.cpp
benoit-bst: **To sum up** (After stand up)

we'll combine the logical flags new rules are
- **is_realtime_enabled** is the main flag, it can be block all realtime access
- **is_realtime_add_enabled** the stronger than **is_realtime_add_trip_enabled**
  If `is_realtime_add_enabled == false` we can't active the _adding trip option_
  so,  `is_realtime_add_trip_enabled == false` that doesn't block the _adding stop time option_

source/kraken/configuration.cpp
pbougue: To me, `is_realtime_add_enabled` can stay the same and it would save some work on conf's PR.
OK anyway.

source/kraken/realtime.cpp
benoit-bst: After the call, now we want to combine flags logics to create priority

source/kraken/configuration.cpp
benoit-bst: after debated :
- is_realtime_enabled
- is_realtime_add_enabled
- is_realtime_add_trip_enabled

source/kraken/realtime.cpp
pbougue: Good for me : combination handled earlier, it's fine either way in that part of the code, you can separate logic :+1:

pbougue (approved):

pbougue:
as discussed : tests and conf PR to come (separated for readability), we may already merge that.

benoit-bst:
Update **Conf** and **Add test** will added in others PR
Tracked within the JIRA ticket : https://jira.kisio.org/browse/NAVP-1148

nonifier (approved):

PR #2682, opened 2019-02-21
Clang-Format

nonifier:
This is a Draft PR to talk about auto formatting of our C++ code base. 
* linter `clang-format-7` 
* Line length: 120 char (based on the previous discussion we've had) (default: 80 char)
* Indentation width: 4 char. To be consistent with the actual code (default: 2) 
* cmd `clang-format-7 -sort-includes -i -style="{BasedOnStyle: chromium, ColumnLimit: 120, IndentWidth: 4}" *.cpp *.h`

### DIFF
--- a/source/.clang-format
+++ b/source/.clang-format
@@ -1,0 +1,9 @@
+---
+BasedOnStyle: chromium
+IndentWidth: 4
+---
+Language: Cpp
+ColumnLimit: 120
+AccessModifierOffset: -4
+AlignConsecutiveAssignments: true
+BreakBeforeBinaryOperators: NonAssignment


source/kraken/apply_disruption.cpp
```diff
@@ -198,45 +200,43 @@ struct add_impacts_visitor : public apply_impacts_visitor {
...
-        } else if (// we don't want to apply delay or detour without stoptime's information
-                   // if there is no stoptimes it should be modeled as a NO_SERVICE
-                   // else it is something else, like for example a SIGNIFICANT_DELAYS on a line
-                   // and in this case we do not have enough information to apply the impact
-                   ! impact->aux_info.stop_times.empty()) {
+        } else if (  // we don't want to apply delay or detour without stoptime's information
+                     // if there is no stoptimes it should be modeled as a NO_SERVICE
+                     // else it is something else, like for example a SIGNIFICANT_DELAYS on a line
+                     // and in this case we do not have enough information to apply the impact
+            !impact->aux_info.stop_times.empty()) {
```
pbougue: Not fan of this, but not fan of the comment place also (don't know what we should do).
Here condition's newline not using the same indent as code below should help.

source/kraken/apply_disruption.cpp
```diff
@@ -198,45 +200,43 @@ struct add_impacts_visitor : public apply_impacts_visitor {
...
                 stoptimes.push_back(stu.stop_time);
             }
-            auto* vj = mvj->create_discrete_vj(new_vj_uri,
-                type::RTLevel::RealTime,
-                canceled_vp,
-                r,
-                std::move(stoptimes),
-                pt_data);
+            auto* vj = mvj->create_discrete_vj(new_vj_uri, type::RTLevel::RealTime, canceled_vp, r,
+                                               std::move(stoptimes), pt_data);
```
pbougue: So newline on each param in method declaration, but not in method use.
Weird but OK: it's hard to keep track of multiple params when it's not named.

source/kraken/apply_disruption.cpp
```diff
@@ -297,39 +296,31 @@ struct add_impacts_visitor : public apply_impacts_visitor {
...
         this->log_end_action(uri);
     }
 
     void operator()(nt::StopPoint* stop_point) {
         log_start_action(stop_point->uri);
 
-        if (impact->severity->effect != nt::disruption::Effect::NO_SERVICE
-            && impact->severity->effect != nt::disruption::Effect::DETOUR) {
+        if (impact->severity->effect != nt::disruption::Effect::NO_SERVICE &&
+            impact->severity->effect != nt::disruption::Effect::DETOUR) {
```
pbougue: Not fan of condition's newline with the same indent as following code

source/kraken/configuration.cpp
```diff
@@ -29,162 +29,172 @@ www.navitia.io
...
-
-        ("CHAOS.database", po::value<std::string>(), "Chaos database connection string");
+         "path to the zmq socket used for receiving resquest")(
+            "GENERAL.instance_name",
+            name ? po::value<std::string>()->default_value(*name) : po::value<std::string>()->required(),
+            "name of the instance")
+
+            ("GENERAL.nb_threads", po::value<int>()->default_value(1), "number of workers threads")(
+                "GENERAL.is_realtime_enabled", po::value<bool>()->default_value(false),
+                "enable loading of realtime data")("GENERAL.is_realtime_add_enabled",
```
pbougue: This (kinda weird code) part is ugly now :smile: 
Don't know if a refactor to help style is possible.

source/kraken/configuration.h
```diff
@@ -29,49 +29,53 @@ www.navitia.io
...
+class Configuration {
+    boost::program_options::variables_map vm;
 
-    class Configuration{
-            boost::program_options::variables_map vm;
-        public:
-            void load(const std::string& file);
-            std::vector<std::string> load_from_command_line(const boost::program_options::options_description&,
-                                        int argc, const char* const argv[]);
+   public:
```
pbougue: Weird "3 then 4" char indent: I prefer "4 then 8", or "0 then 4".

benoit-bst:
I like the idea to have an auto formatting C++ code. I think _Clang-format_ is the best.
I am agree with @nonifier to used a light clang-format config.
Just, be careful if you use a recent version of _Clang-format_, you'll have to update the contributing doc (https://github.com/CanalTP/navitia/blob/dev/CONTRIBUTING.md) and describe the package installation.

kisses :heart:

pbougue (commented):
Looks like it's making many things better. A very few worse (see comments).

We also should pay attention to clang's version's availability on different platforms that a navitia dev would legitimately use (debian 9, don't know about ubuntu).

source/kraken/apply_disruption.cpp
benoit-bst: The place of comments is so weird initially...

source/kraken/configuration.cpp
nonifier: TBH, it was already ugly :rofl: 
You can disable the linting with : https://clang.llvm.org/docs/ClangFormatStyleOptions.html#disabling-formatting-on-a-piece-of-code.
Although, this might require some tweaks in the config if we don't want to go back to all the unit tests that are doing things in a similar way.

source/kraken/apply_disruption.cpp
nonifier: We should just document within the if bloc, after the opening brace. As normal people would do :)

source/kraken/apply_disruption.cpp
nonifier: What do you mean with `it's hard to keep track of multiple params when it's not named` ? You mean named as in Python ?

source/kraken/apply_disruption.cpp
nonifier: I agree, I prefer when they are actually on the next line, and indented like:
```cpp
    if (impact->severity->effect != nt::disruption::Effect::NO_SERVICE 
     && impact->severity->effect != nt::disruption::Effect::DETOUR) {
```
But, we can mourn it together if you want ;)

source/kraken/configuration.h
nonifier: This can be quickly fixed with  a `AccessModifierOffset: -4` in the config.

nonifier: This should be fixed !

source/kraken/configuration.cpp
```diff
@@ -45,59 +45,51 @@ po::options_description get_options_description(const boost::optional<std::strin
...
                                                 const boost::optional<bool> display_contributors) {
     po::options_description desc("Allowed options");
-    desc.add_options()("GENERAL.database", po::value<std::string>()->default_value("data.nav.lz4"),
-                       "path to the data file")
+    // clang-format off
+    desc.add_options()
+        ("GENERAL.database", po::value<std::string>()->default_value("data.nav.lz4"), "path to the data file")
 
-        // name and zmq socket can have default values (for tests)
+            //name and zmq socket can have default values (for tests)
```
pbougue: minor: here we can use format suggestion

source/kraken/configuration.cpp
```diff
@@ -45,59 +45,51 @@ po::options_description get_options_description(const boost::optional<std::strin
...
+                                  "timeout in ms before retrying to load realtime data")
+        ("GENERAL.slow_request_duration", po::value<int>()->default_value(1000),
+                                  "request running a least this number of milliseconds are logged")
+
+        ("GENERAL.display_contributors", display_contributors ?
+             po::value<bool>()->default_value(*display_contributors) : po::value<bool>()->default_value(false),
+         "display all contributors in feed publishers")
+        ("GENERAL.raptor_cache_size", po::value<int>()->default_value(10), "maximum number of stored raptor caches")
+        ("GENERAL.log_level", po::value<std::string>(), "log level of kraken")
+        ("GENERAL.log_format", po::value<std::string>()->default_value("[%D{%y-%m-%d %H:%M:%S,%q}] [%p] [%x] - %m %b:%L  %n"), "log format")
```
pbougue: minor: newlines after `,` should help with line length

source/kraken/apply_disruption.cpp
pbougue: Yep, I like align on `(` everywhere else, so...
Let's mourn together, but I will mourn on a missing double-indent, instead ;)

pbougue (approved):
Should we deal with those "side" style tasks before merging it?
(thinking of bad comment place reviewed before, and those 2 minor remarks, maybe only visible in commit).
Also, should we think about tweaking for those call chains `()()()()` ?

source/kraken/apply_disruption.cpp
pbougue: Yep, hard to admit that there are some things I prefer in python :smile:

PR #2692, opened 2019-02-27
kraken: delete a queue after a timeout for a reload from kirin

kadhikari:
https://jira.kisio.org/browse/NAVP-969

### DIFF
--- a/source/kraken/maintenance_worker.cpp
+++ b/source/kraken/maintenance_worker.cpp
@@ -112,6 +112,8 @@ void MaintenanceWorker::load_realtime(){
         data_manager.get_data()->is_realtime_loaded = true;
     }else{
         LOG4CPLUS_WARN(logger, "no realtime data receive before timeout: going without it!");
+        // Incase of timeout we can delete the existing queue as we create a new queque in each call to the function
+        channel->DeleteQueue(queue_name);
     }
 }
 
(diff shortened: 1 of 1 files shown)


source/kraken/maintenance_worker.cpp
```diff
@@ -112,6 +112,8 @@ void MaintenanceWorker::load_realtime(){
         data_manager.get_data()->is_realtime_loaded = true;
     }else{
         LOG4CPLUS_WARN(logger, "no realtime data receive before timeout: going without it!");
+        // Incase of timeout we can delete the existing queue as we create a new queque in each call to the function
+        channel->DeleteQueue(queue_name);
```
kinnou02: You should do it every time, even when we got a response.

kinnou02 (changes requested):

kinnou02 (approved):

source/kraken/maintenance_worker.cpp
```diff
@@ -113,6 +113,8 @@ void MaintenanceWorker::load_realtime(){
     }else{
         LOG4CPLUS_WARN(logger, "no realtime data receive before timeout: going without it!");
     }
+    // Finally we delete de queue as we create a new queque in each call to the function
```
pbougue: typo: "Finally (maybe remove `we`) delete `th`e queue [...] new queue (no extra `q`) `at` each call"

pbougue (approved):
This PR is fine.
As for the JIRA ticket itself, how about other queues created by Kraken? Are there some that should be deleted also?
This can be a new ticket, but it would be nice to also check that.

source/kraken/maintenance_worker.cpp
kadhikari: Done

PR #2694, opened 2019-02-28
Jormun: Add parking module connector augeas

xlqian:
Now we have car parking information in journey with asgard!!

### DIFF
--- a/source/jormungandr/jormungandr/street_network/parking/augeas.py
+++ b/source/jormungandr/jormungandr/street_network/parking/augeas.py
@@ -27,16 +27,24 @@
 # https://groups.google.com/d/forum/navitia
 # www.navitia.io
 from __future__ import absolute_import, print_function, unicode_literals, division
-import requests, pybreaker, logging
+import requests
+import pybreaker
+import logging
+from datetime import timedelta
 from jormungandr.street_network.parking.abstract_parking_module import AbstractParkingModule
+
+DEFAULT_MAX_PARKING_DURATION = int(timedelta(seconds=1200).total_seconds())
 
 
 class Augeas(AbstractParkingModule):
-    def __init__(self, service_url, max_park_duration=1200):
+    def __init__(self, service_url, max_park_duration=DEFAULT_MAX_PARKING_DURATION):
         self.service_url = service_url
         self.max_park_duration = max_park_duration
         # TODO: put into a config when the POC is validated
-        self.breaker = pybreaker.CircuitBreaker(fail_max=500, reset_timeout=60)
+        one_minute = timedelta(seconds=60).total_seconds()
+        self.breaker = pybreaker.CircuitBreaker(fail_max=10, reset_timeout=one_minute)
+        self.request_timeout = timedelta(seconds=0.1).total_seconds()
+
         self.logger = logging.getLogger(__name__)
 
     def _breaker_wrapper(self, fun, *args, **kwargs):
@@ -54,7 +62,7 @@ class Augeas(AbstractParkingModule):
             "coords": [[c.lon, c.lat] for c in coords],
         }
         url = requests.compat.urljoin(self.service_url, '/v0/park_duration')
-        r = requests.post(url=url, json=data)
+        r = requests.post(url=url, json=data, timeout=self.request_timeout)
         return r.json().get('durations')
 
     def _request(self, coord):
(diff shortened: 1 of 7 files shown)


AurelienLP (approved):

source/jormungandr/jormungandr/street_network/parking/augeas.py
```diff
@@ -0,0 +1,75 @@
...
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with this program. If not, see <http://www.gnu.org/licenses/>.
+#
+# Stay tuned using
+# twitter @navitia
+# IRC #navitia on freenode
+# https://groups.google.com/d/forum/navitia
+# www.navitia.io
+
```
kinnou02:
```suggestion
from __future__ import absolute_import, print_function, unicode_literals, division
```

source/jormungandr/jormungandr/street_network/parking/augeas.py
```diff
@@ -0,0 +1,75 @@
...
+    def _request_in_batch(self, coords):
+        data = {
+            "n": 1,
+            "walking_speed": 1.11,
+            "max_park_duration": self.max_park_duration,
+            "coords": [[c.lon, c.lat] for c in coords],
+        }
+        url = requests.compat.urljoin(self.service_url, '/v0/park_duration')
+        r = requests.post(url=url, json=data)
```
kinnou02: There is no timeout nor circuit breaker.
Timeout is mandatory.
circuitbreaker may be omit for a poc, but this should never be use without it.

source/jormungandr/jormungandr/street_network/parking/constant_parking.py
```diff
@@ -27,16 +27,22 @@
 # https://groups.google.com/d/forum/navitia
 # www.navitia.io
```
kinnou02: `from __future__ import absolute_import, print_function, unicode_literals, division`

kinnou02 (changes requested):

source/jormungandr/jormungandr/street_network/parking/augeas.py
```diff
@@ -26,15 +26,25 @@
...
+import pybreaker
+from jormungandr.street_network.parking.abstract_parking_module import AbstractParkingModule
 
 
 class Augeas(AbstractParkingModule):
     def __init__(self, service_url, max_park_duration=1200):
         self.service_url = service_url
         self.max_park_duration = max_park_duration
+        # TODO: put into a config when the POC is validated
+        self.breaker = pybreaker.CircuitBreaker(fail_max=500, reset_timeout=60)
```
kinnou02: 500 seems to be a lot

source/jormungandr/jormungandr/street_network/parking/augeas.py
kinnou02: still no timeout

source/jormungandr/jormungandr/street_network/parking/augeas.py
```diff
@@ -0,0 +1,85 @@
...
+# IRC #navitia on freenode
+# https://groups.google.com/d/forum/navitia
+# www.navitia.io
+from __future__ import absolute_import, print_function, unicode_literals, division
+import requests, pybreaker, logging
+from jormungandr.street_network.parking.abstract_parking_module import AbstractParkingModule
+
+
+class Augeas(AbstractParkingModule):
+    def __init__(self, service_url, max_park_duration=1200):
```
nonifier: It's quite confusing and obfuscated when duration types are neither explicit nor documented. 
I guess your duration here is in second because that's an assumption we've made in the codebase. But to withdraw the ambiguity, I would rather see `max_park_duration=time(0, 0, 1200)` or even better `time(0, 20)`.

source/jormungandr/jormungandr/street_network/parking/augeas.py
```diff
@@ -0,0 +1,85 @@
...
+        durations = r.json().get('durations')
+        if not durations:
+            return self.max_park_duration
+        return durations[0].get('duration', self.max_park_duration)
+
+    def get_parking_duration(self, coord):
+        return self._breaker_wrapper(self._request, coord=coord)
+
+    def get_leave_parking_duration(self, coord):
+        return self._breaker_wrapper(self._request, coord=coord)
```
nonifier, line 87: what is the difference with `get_parking_duration` ? they seem pretty similar. (same for the 2 functions above)

nonifier (approved):

source/jormungandr/jormungandr/street_network/parking/augeas.py
xlqian: you're right, they're the same in the implementation, even though they're different in real life, it should be handled differently later

source/jormungandr/jormungandr/street_network/parking/augeas.py
xlqian: oops

kinnou02 (approved):

PR #2695, opened 2019-03-01
Clang format

nonifier:
Here is a run of clang-format on the whole code-base (excluding sub-modules) using:
```sh
pre-commit run clang-format -a
```
I've tried to keep the configuration as small as possible in `source/.clang-format`. But this could be updated later. This PR still aims at reviewing the syntax before we integrate it as part of the CI and enforce it for everyone.

Clang-format has been added to Travis to ensure formatting is persistent.

Make sure to install clang-format-6 :D


pbougue:
We should probably exclude third_party and include submodule we own (or add it is submodule's repo)

kinnou02:
I'am not a big fan of `AlignConsecutiveAssignments` it tends to generate big diff because you renamed something

source/type/type.h
```diff
@@ -160,23 +156,23 @@ struct StopPoint : public Header, Nameable, hasProperties, HasMessages {
...
-    template<class Archive> void serialize(Archive & ar, const unsigned int ) {
+    template <class Archive>
+    void serialize(Archive& ar, const unsigned int) {
         // The *_list are not serialized here to avoid stack abuse
         // during serialization and deserialization.
         //
         // stop_point_connection_list is managed by StopPointConnection
-        ar & uri & label & name & stop_area & coord & fare_zone & is_zonal & idx & platform_code
-            & admin_list & _properties & impacts & dataset_list;
+        ar& uri& label& name& stop_area& coord& fare_zone& is_zonal& idx& platform_code& admin_list& _properties&
```
kinnou02, line 166: the spacing was nice here :(

source/ptreferential/ptreferential_ng.cpp
```diff
@@ -101,51 +109,40 @@ std::ostream& operator<<(std::ostream& os, const BinaryOp<Diff>& op) {
...
-            >> '('
-            >> str[phx::push_back(phx::at_c<2>(_val), phx::construct<std::string>()),
-                   phx::push_back(phx::at_c<2>(_val), _1)]
-            >> ','
-            >> str[phx::back(phx::at_c<2>(_val)) += ";",
-                   phx::back(phx::at_c<2>(_val)) += _1]
-            >> ','
-            >> str[phx::at(phx::at_c<2>(_val), 0) = _1]
-            >> ')';
+        fun   = ident >> '.' >> ident >> '(' >> -(str % ',') >> ')';
```
kinnou02: We might want to not format this, it isn't easy to understand from the start, but with everything pack it's harder

nonifier: I've surrounded the code with `// clang-format off/on`

source/type/type.h
nonifier: Unfortunately, I couldn't find any configuration to fix this :(

nonifier:
@kinnou02, I've removed `AlignConsecutiveAssignments` and disable include sorting. 
@pbougue, we should add this for every submodule we own if we want. But no point doing the formatting here as we don't have control over the source from an external repo.

benoit-bst (approved):

source/ed/connectors/synonym_parser.cpp
```diff
@@ -34,43 +34,43 @@ www.navitia.io
...
         auto row = reader.next();
-        if (reader.is_valid(key_c, row) && reader.has_col(value_c, row)){
-                key = row[key_c];
-                value = row[value_c];
-                boost::to_lower(key);
-                boost::to_lower(value);
-                this->synonym_map[key]=value;
-            }
+        if (reader.is_valid(key_c, row) && reader.has_col(value_c, row)) {
+            key = row[key_c];
```
benoit-bst, line 62: minor: indent

kinnou02 (approved):

.pre-commit-config.yaml
```diff
@@ -11,3 +11,11 @@ repos:
...
     -   id: check-merge-conflict
     #-   id: flake8
+-   repo: git://github.com/doublify/pre-commit-clang-format
+    rev: master
+    hooks:
+    -   id: clang-format
+        verbose: true
+        exclude: ^source/third_party/
+        files: \.(c|cc|cxx|cpp|h|hpp|hxx)$
+        args: [-i]
```
xlqian, line 21: new line!

xlqian (approved):

source/ed/connectors/synonym_parser.cpp
nonifier: huuu ?! 
The indentation was wrong before :confused:

PR #2700, opened 2019-03-06
Kraken: fix disruption deletion when multiple impacts impacting the same vj

eturck:
Hello.

We found a bug disruptions when multiple impacts are in the same disruption and are impacting the same meta_vj.
If we have a disruption D with two impacts I1 and I2, both impacting VJ what was happening was :
1. First application of D
  - I1 is applied, we are creating a new VJ VJ:Adapted:0.
  - I2 is applied, we are creating a new VJ VJ:Adapted:0:Adapted:1. VJ:Adapted:0 is useless and cleaned.
2. Deletion of D
  - I1 is deleted. We are deleting VJ:Adapted:0:Adapted:1. We detect that another impact was on this VJ, I2 so we keep the disruption D in a list to reapply it at the end.
  - We are applying D again. Both I1 and I2 are reapplied. We have VJ:Adapted:0:Adapted:1 again.
  - I2 is deleted. Same thing happen because I1 is still linked since we just applied it.
  - We are applying D again. We have VJ:Adapted:0:Adapted:1.

Things are pretty strange then. We have two VJ, VJ and VJ:Adapted:0:Adapted:1. The disruption was deleted so VJ has no impact left. VJ:Adapted:0:Adapted:1 still exists with the adapted validity pattern.

If we try to apply the disruption another time it fails, since we don't find any vj to impact (if it's an unserved stop between two dates, VJ:Adapted:0:Adapted:1 is still active between those two dates and don't serve this stop).

I made a test with some comments, it might be easier to understand :).

My fix is pretty simple. When we are deleting an impact, don't reapply the disruption if another impact is still in the modified_by of the meta_vj, if it's the same disruption uri.
We still need to keep the other impact in modified_by though, because we will be deleting it just after and the delete_impacts_visitor look inside modified_by to see which vj must be processed. So I just insert the impact back in modified_by.
It works, because we don't delete an impact without deleting the complete disruption. If this was to change we would need to do something smarter.

Let me know what you think and if you have better ideas on how to fix this !

### DIFF
--- a/source/kraken/apply_disruption.cpp
+++ b/source/kraken/apply_disruption.cpp
@@ -579,10 +579,18 @@ struct delete_impacts_visitor : public apply_impacts_visitor {
 
         for(const auto& wptr: modified_by_moved) {
             if (auto share_ptr = wptr.lock()){
-                // Do not reapply the same disruption. If we have more than one impact we will
-                // reapply the one we just deleted.
-                // We need to keep the link though, because other impacts from the disruption may
-                // need to be deleted after this one and they need to stay in modified_by.
+                /*
+                 * Do not reapply the same disruption. If we have more than one impact in it
+                 * we would reapply all of its impacts (even the one we are deleting).
+                 * We need to keep the link to remaining impacts of the disruption though, because
+                 * they will be deleted after this one. They must stay in modified_by for that.
+                 *
+                 * /!\ WARNING /!\
+                 * This is working because we never delete a single impact of a disruption.
+                 * delete_impact is only called by delete_disruption which delete every impacts.
+                 * If this was not the case we would need to find a way to reapply part of a
+                 * disruption's impacts, and not all of them, after each deletion of an impact.
+                 */
                 if (share_ptr->disruption->uri != impact->disruption->uri) {
                     disruptions_collection.insert(share_ptr);
                 } else {
--- a/source/kraken/apply_disruption.cpp
+++ b/source/kraken/apply_disruption.cpp
@@ -579,7 +579,15 @@ struct delete_impacts_visitor : public apply_impacts_visitor {
 
         for(const auto& wptr: modified_by_moved) {
             if (auto share_ptr = wptr.lock()){
-                disruptions_collection.insert(share_ptr);
+                // Do not reapply the same disruption. If we have more than one impact we will
+                // reapply the one we just deleted.
+                // We need to keep the link though, because other impacts from the disruption may
+                // need to be deleted after this one and they need to stay in modified_by.
+                if (share_ptr->disruption->uri != impact->disruption->uri) {
+                    disruptions_collection.insert(share_ptr);
+                } else {
+                    mvj->push_unique_impact(share_ptr);
+                }
             }
         }
         // we check if we now have useless vehicle_journeys to cleanup
--- a/source/kraken/tests/apply_disruption_test.cpp
+++ b/source/kraken/tests/apply_disruption_test.cpp
@@ -2396,7 +2396,7 @@ BOOST_AUTO_TEST_CASE(test_adapted_disruptions_on_stop_point_then_line) {
 }
 
 // We check that with a disruption with two impacts impacting the same vj the update is working.
-// It was failing before since after deleting each impact, we were reapplying the all disruption.
+// It was failing before since after deleting each impact, we were reapplying the disruption.
 BOOST_AUTO_TEST_CASE(update_disruption_with_multiple_impact_on_same_vj) {
     ed::builder b("20190301");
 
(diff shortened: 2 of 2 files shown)


source/kraken/tests/apply_disruption_test.cpp
```diff
@@ -2394,3 +2394,181 @@ BOOST_AUTO_TEST_CASE(test_adapted_disruptions_on_stop_point_then_line) {
...
+    b.sa("stop_area:5", 0, 0, false, true)("stop_point:50");
+
+    b.vj("line:A", "1111111", "", true, "vj:1")
+            ("stop_point:10", "10:00"_t)
+            ("stop_point:20", "11:00"_t)
+            ("stop_point:30", "12:00"_t)
+            ("stop_point:40", "13:00"_t)
+            ("stop_point:50", "14:00"_t);
+
+    b.generate_dummy_basis();
```
pbougue: You can use `b.make()` for the 5 lines from here (also note that `finish()` calls `build_raptor()`)

source/kraken/tests/apply_disruption_test.cpp
```diff
@@ -2394,3 +2394,181 @@ BOOST_AUTO_TEST_CASE(test_adapted_disruptions_on_stop_point_then_line) {
...
+    disruption = b.data->pt_data->disruption_holder.get_disruption("line_section_on_line:A_section_in_each_way");
+    BOOST_REQUIRE(disruption != nullptr);
+    BOOST_REQUIRE_EQUAL(disruption->get_impacts().size(), 2);
+
+    navitia::delete_disruption("line_section_on_line:A_section_in_each_way", *b.data->pt_data, *b.data->meta);
+
+    BOOST_REQUIRE_EQUAL(b.data->pt_data->vehicle_journeys.size(), 2);
+    BOOST_REQUIRE_EQUAL(b.data->pt_data->vehicle_journeys_map.at("vj:1")->get_impacts().size(), 0);
+    BOOST_REQUIRE_EQUAL(b.data->pt_data->vehicle_journeys_map.at("vj:2")->get_impacts().size(), 0);
+}
```
pbougue: missing newline

source/kraken/tests/apply_disruption_test.cpp
```diff
@@ -2394,3 +2394,181 @@ BOOST_AUTO_TEST_CASE(test_adapted_disruptions_on_stop_point_then_line) {
             ba::ends_with(rt_vj1->rt_validity_pattern()->days.to_string(), "00000000"),
             rt_vj1->rt_validity_pattern()->days);
 }
+
+// We check that with a disruption with two impacts impacting the same vj the update is working.
+// It was failing before since after deleting each impact, we were reapplying the all disruption.
```
pbougue: typo: `all` is probably too much ;)

source/kraken/apply_disruption.cpp
```diff
@@ -579,7 +579,15 @@ struct delete_impacts_visitor : public apply_impacts_visitor {
 
         for(const auto& wptr: modified_by_moved) {
             if (auto share_ptr = wptr.lock()){
-                disruptions_collection.insert(share_ptr);
+                // Do not reapply the same disruption. If we have more than one impact we will
+                // reapply the one we just deleted.
+                // We need to keep the link though, because other impacts from the disruption may
```
pbougue: "keep the link `to other impacts of the disruption` though, because"

source/kraken/apply_disruption.cpp
```diff
@@ -579,7 +579,15 @@ struct delete_impacts_visitor : public apply_impacts_visitor {
 
         for(const auto& wptr: modified_by_moved) {
             if (auto share_ptr = wptr.lock()){
-                disruptions_collection.insert(share_ptr);
+                // Do not reapply the same disruption. If we have more than one impact we will
```
pbougue: we `would` reapply

source/kraken/apply_disruption.cpp
```diff
@@ -579,7 +579,15 @@ struct delete_impacts_visitor : public apply_impacts_visitor {
...
+                // reapply the one we just deleted.
+                // We need to keep the link though, because other impacts from the disruption may
+                // need to be deleted after this one and they need to stay in modified_by.
```
pbougue: Can you add a word about the fact that it is possible because a disruption is always completely deleted (all of its impacts also)?

pbougue (commented):
Nice catch!
I'll try to think of a way to treat independently disruption and impact, but I'm not sure it's worth it.
Tried to think about trickier tests but no smart idea about that.

kinnou02 (approved):

source/kraken/tests/apply_disruption_test.cpp
eturck: Thanks, did not knew that and I just copy-pasted an old test without much thinking, my bad :).

source/kraken/apply_disruption.cpp
eturck: I've added something, I hope it's clear enough. I struggled a bit to explain this in French so in English it's not really easier :D.

pbougue (approved):

PR #2703, opened 2019-03-07
Fix empty distances returned on /places by Kraken

nonifier:
Fix nil distances returned by Kraken on `/places`.

The problem was coming from the protobuf that always allocates a distance as attribute which is never none.... 

[#NAVITIAII-2673](https://jira.kisio.org/browse/NAVITIAII-2673)

### DIFF
--- a/source/jormungandr/tests/places_tests.py
+++ b/source/jormungandr/tests/places_tests.py
@@ -240,3 +240,13 @@ class TestPlaces(AbstractTestFixture):
         assert len(places) == 1
         is_valid_places(places)
         assert len(response['disruptions']) == 0
+
+    def test_places_has_no_distance(self):
+        """So far, Kraken has no distance returned from its API (as opposed to Bragi)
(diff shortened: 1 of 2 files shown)


source/jormungandr/tests/places_tests.py
```diff
@@ -240,3 +240,13 @@ def test_disruptions_in_places(self):
         assert len(places) == 1
         is_valid_places(places)
         assert len(response['disruptions']) == 0
+
+    def test_places_has_no_distance(self):
+        """So far, Kraken has no distance returned from its API (as opposed to Bragi)
```
mbounabi, line 245: Minor: there should be a carriage return ;)

pbougue (approved):

mbounabi (approved):
:+1:

kadhikari (approved):

source/jormungandr/tests/places_tests.py
nonifier: there should be a warning from Black :smiling_imp:

PR #2705, opened 2019-03-08
Doc: small detail on fare_extention

prhod:
If the end of line is not present, the line is ignored


pbougue (approved):
:smile: :see_no_evil:

mbounabi (approved):

nonifier (approved):

PR #2706, opened 2019-03-08
Add equipment details provider configuration mechanism

nonifier:
Add preliminary mechanism to Jormun for configuring the Equipment Details Providers !

The idea is to define each provider individually, and refer to them by their keys in the instance config like:
```json
JORMUNGANDR_EQUIPMENT_DETAIL_PROVIDER_SYTRAL = {
	"class": "jormungandr.my.class", 
	"key":"sytral", 
	"args": { "url": "http://blabla.bla", "operators": ["my_operator"], "dataset": "_", "fail_max": 5, "timeout": 1 } 
}
JORMUNGANDR_EQUIPMENT_DETAIL_PROVIDER_OTHER = {
	"class": "jormungandr.my.other.class", 
	"key":"other", 
	"args": {"url": "http://other.blabla.bla", "operators": ["other_op"], "dataset": "_", "fail_max": 5, "timeout": 1 } 
}

JORMUNGANDR_INSTANCE_IDF={
	"key": "IDF", 
	"zmq_socket": "ipc:///tmp/idf_kraken", 
	"equipment_details_providers": [ "sytral", "other" ]
}

```

[#NAVP-1212](https://jira.kisio.org/browse/NAVP-1212)

TODO: Configuration from tyr and then tyr-configurator

### DIFF
--- a/source/jormungandr/jormungandr/equipments/equipment_provider_manager.py
+++ b/source/jormungandr/jormungandr/equipments/equipment_provider_manager.py
@@ -44,7 +44,9 @@ class EquipmentProviderManager(object):
         # TODO: create providers instance only if key is present in the Jormungandr instance config ?
         for configuration in equipment_providers_configuration:
             arguments = configuration.get('args', {})
-            self._equipment_providers_legacy[configuration['key']] = self._init_class(configuration['class'], arguments)
+            self._equipment_providers_legacy[configuration['key']] = self._init_class(
+                configuration['class'], arguments
+            )
 
     def _init_class(self, cls, arguments):
         """
--- a/source/jormungandr/jormungandr/equipments/equipment_provider_manager.py
+++ b/source/jormungandr/jormungandr/equipments/equipment_provider_manager.py
@@ -29,6 +29,8 @@
 
 from __future__ import absolute_import, print_function, unicode_literals, division
 
+from importlib import import_module
+
 import logging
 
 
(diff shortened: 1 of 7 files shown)


source/jormungandr/jormungandr/instance.py
```diff
@@ -111,6 +111,7 @@ def __init__(
         realtime_proxies_configuration,
         zmq_socket_type,
         autocomplete_type,
+        equipment_details_providers,  # type: List
```
kinnou02: List of what? :)

source/jormungandr/jormungandr/instance.py
```diff
@@ -156,6 +158,19 @@ def __init__(
...
+    def configure_equipment_details_providers(self, equipment_details_providers, instance_name):
+        equipment_providers = [provider['key'] for provider in app.config[str('EQUIPMENT_DETAILS_PROVIDERS')]]
+        for provider in equipment_details_providers:
+            if provider not in equipment_providers:
+                raise RuntimeError(
+                    'impossible to find equipment details provider: {} '
+                    'cannot initialize instance {}'.format(provider, instance_name)
+                )
+
+        self.equipment_details_providers = equipment_details_providers
```
kinnou02: look like we only got a list of strings here, that not realy useful...

source/jormungandr/jormungandr/instance.py
```diff
@@ -156,6 +158,19 @@ def __init__(
 
         self.zmq_socket_type = zmq_socket_type
 
+        self.configure_equipment_details_providers(equipment_details_providers, name)
+
+    def configure_equipment_details_providers(self, equipment_details_providers, instance_name):
```
xlqian: you don't need to pass `equipment_details_providers, name` as arguments, use `self.*` and I think the code will be little bit more readable with that :)

source/jormungandr/jormungandr/instance.py
```diff
@@ -156,6 +158,19 @@ def __init__(
...
+        equipment_providers = [provider['key'] for provider in app.config[str('EQUIPMENT_DETAILS_PROVIDERS')]]
```
xlqian: I'd prefer instantiate equipment_providers in jormungandr.__init__ as `bss_provider_manager`

xlqian: you will need a abstract(or not?) class for equipment_providers

source/jormungandr/jormungandr/equipments/equipment_provider_manager.py
```diff
@@ -0,0 +1,64 @@
...
+class EquipmentProviderManager(object):
+    def __init__(self, equipment_providers_configuration):
+        self.logger = logging.getLogger(__name__)
+        self.providers_config = equipment_providers_configuration
+        self.providers = {}
+
+    def get_providers(self, provider_key):
+        """
+        :param provider_key: provider set in instance configuration
+        :return: True if provider key is defined in Jormun configuration
```
mbounabi: Comment is wrong: will be fixed

source/jormungandr/jormungandr/instance.py
```diff
@@ -156,6 +158,18 @@ def __init__(
 
         self.zmq_socket_type = zmq_socket_type
 
+        self.equipment_providers = {}
+
+        self.configure_equipment_details_providers(instance_equipment_providers)
+
+    def configure_equipment_details_providers(self, instance_equipment_providers):
+        for provider_key in instance_equipment_providers:
+            provider = equipment_provider_manager.get_providers(provider_key)
```
kinnou02: minor: taking `equipment_provider_manager` as a parameter will ease testing later.
It's better to inject dependency than get them from globals.

source/jormungandr/jormungandr/instance.py
```diff
@@ -156,6 +158,18 @@ def __init__(
...
+            if provider:
+                self.equipment_providers[provider_key] = provider
```
kinnou02: This won't work once the configuration will be stored in the database.
In the instance you can just keep the keys of the wanted provider (later we will get them from the database) and get the provider from the manager each time we need them.

In the instance we will keep the list of provider id wanted, and a ref to the manager, then when needed we will get the provider from the manager, this way we always get an up to date provider without having to handle complex update system inside each instance.

source/jormungandr/jormungandr/equipments/equipment_provider_manager.py
```diff
@@ -0,0 +1,64 @@
...
+import logging
+
+
+class EquipmentProviderManager(object):
+    def __init__(self, equipment_providers_configuration):
+        self.logger = logging.getLogger(__name__)
+        self.providers_config = equipment_providers_configuration
+        self.providers = {}
+
+    def get_providers(self, provider_key):
```
kinnou02:
```suggestion
    def get_provider(self, provider_key):
```

source/jormungandr/jormungandr/instance.py
mbounabi: :heavy_check_mark:

source/jormungandr/jormungandr/instance.py
mbounabi: :heavy_check_mark:

xlqian (approved):

source/jormungandr/jormungandr/instance.py
```diff
@@ -123,6 +124,7 @@ def __init__(
         self.timezone = None  # timezone will be fetched from the kraken
         self.publication_date = -1
         self.is_initialized = False  # kraken hasn't been called yet we don't have geom nor timezone
+        self.equipment_details_providers = []
```
kinnou02: You don't need that anymore

kinnou02 (approved):

source/jormungandr/jormungandr/instance.py
mbounabi: Indeed...

PR #2707, opened 2019-03-12
Jormun: enable keepalive on http call to bragi

kinnou02:
This aims to improve performance by reducing the number of tcp
connection to open.
I have done some testing from the office network to the production:
without keepalive response time for a simple autocomplete is around
95ms. With keepalive it's around 80ms.
There is obviously some overhead in this test from our vpn and internet,
on production there will be way less hop.

Requests handle everything once a session is enabled, it will reopen the
connection if needed, etc.

In case of a problem it's possible to disable the keepalive in the
configuration by setting the parameter "disable_keepalive" to true in
the geocodejson configuration

I had to update every test that mock bragi, they were mocking `request.get` directly, obvioulsy it didn't work anymore when using session, I updated every test to use [`request_mock`](https://requests-mock.readthedocs.io/en/latest/overview.html). I also fixed a few of them as bragi wasn't since we used the instance autocomplete.

### DIFF
--- a/source/jormungandr/jormungandr/autocomplete/geocodejson.py
+++ b/source/jormungandr/jormungandr/autocomplete/geocodejson.py
@@ -186,6 +186,11 @@ class GeocodeJson(AbstractAutocomplete):
             fail_max=app.config['CIRCUIT_BREAKER_MAX_BRAGI_FAIL'],
             reset_timeout=app.config['CIRCUIT_BREAKER_BRAGI_TIMEOUT_S'],
         )
+        # create a session to allow connection pooling via keep alive
+        if kwargs.get('disable_keepalive', False):
+            self.session = requests
+        else:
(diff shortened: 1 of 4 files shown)


xlqian:
Question: shouldn't this be used everywhere in navitia? i mean not only for bragi, but also for sytral/instant system/star/all other connectors via HTTP ?

nonifier (approved):

nonifier:
Very good point from @xlqian. All the connectors would potentially benefit from a persistent connection !

kinnou02:
We may latter update others connectors, but we have to consider a few things before:
  - persistent connection have a cost: so if it's not reused it's a loss.
  - there is always a risk with persistent connection, some firewall can mess with them.
  - if the API is slow the gain is not worth it

xlqian (approved):

PR #2710, opened 2019-03-13
kraken: implements deadline on kraken request

kinnou02:
This PR let jormun define the point in time where it won't need a response
anymore: after the timeout. With the information kraken can ignore
request that have already reached timeout on the jormungandr side.

This case happen when kraken is overloaded with too many requests, these
are queued and handled gradually, but they are never invalidated, so
kraken can respond to a request a few minutes after it has been sent:
this is wasting resources and aggravate the problem.

The current implementation only check the deadline at the start of a
request but could be extended to provide cancellation lower in the call
chain.

By default these deadline are ignored by kraken, this permit server that aren't time sync to continue working

Require https://github.com/CanalTP/navitia-proto/pull/122 and https://github.com/CanalTP/utils/pull/78
Some explanations are on these.

### DIFF
--- a/source/kraken/configuration.cpp
+++ b/source/kraken/configuration.cpp
@@ -76,6 +76,7 @@ po::options_description get_options_description(const boost::optional<std::strin
         ("GENERAL.log_level", po::value<std::string>(), "log level of kraken")
         ("GENERAL.log_format", po::value<std::string>()->default_value("[%D{%y-%m-%d %H:%M:%S,%q}] [%p] [%x] - %m %b:%L  %n"), "log format")
 
+        ("GENERAL.enable_deadline", po::value<bool>()->default_value(false), "enforce deadline of request")
         ("GENERAL.metrics_binding", po::value<std::string>(), "IP:PORT to serving metrics in http")
 
         ("BROKER.host", po::value<std::string>()->default_value("localhost"), "host of rabbitmq")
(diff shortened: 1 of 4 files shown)


xlqian:
zmq doesn't handle message TTL? rabbtimq has that option...

source/kraken/configuration.cpp
```diff
@@ -76,6 +76,7 @@ po::options_description get_options_description(const boost::optional<std::strin
         ("GENERAL.log_level", po::value<std::string>(), "log level of kraken")
         ("GENERAL.log_format", po::value<std::string>()->default_value("[%D{%y-%m-%d %H:%M:%S,%q}] [%p] [%x] - %m %b:%L  %n"), "log format")
 
+        ("GENERAL.enable_deadline", po::value<bool>()->default_value(true), "enforce deadline of request")
```
xlqian: enable_deadline is a little bit generic... enable_message_deadline? or enable_zeromq_deadline? enable_zeromq_ttl?

xlqian (approved):

nonifier (approved):

source/kraken/configuration.cpp
nonifier: what about `enable_request_deadline` ?

nonifier:
```suggestion
        ("GENERAL.enable_deadline", po::value<bool>()->default_value(true), "enable deadline of request")
```

benoit-bst (approved):

kinnou02:
zmq doesn't handle TTL sadly

PR #2712, opened 2019-03-14
tyr: absent parameters added in /instances

kadhikari:
https://jira.kisio.org/browse/NAVP-1241

### DIFF
--- a/source/tyr/tyr/resources.py
+++ b/source/tyr/tyr/resources.py
@@ -542,6 +542,22 @@ class Instance(flask_restful.Resource):
             default=instance.autocomplete_backend,
         )
 
+        parser.add_argument(
+            'max_additional_connections',
+            type=int,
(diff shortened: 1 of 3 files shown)


kinnou02 (approved):

source/tyr/tyr/resources.py
```diff
@@ -542,6 +542,22 @@ def put(self, id=None, name=None):
...
+        parser.add_argument(
+            'max_additional_connections',
+            type=int,
+            help='maximum number of connections allowed in journeys',
+            location=('json', 'values'),
+            default=instance.max_additional_connections,
+        )
+
+        parser.add_argument(
+            'car_park_provider',
```
pbougue, line 554: If we can change it, "`enable_`car_park_provider" would be a clearer name

pbougue (approved):

PR #2720, opened 2019-03-22
Kraken: Skip realtime cancelled trip update on a VJ that doesn't exist + tests

benoit-bst:
Skip cancelled trip update at the `handle_realtime` function entry point, when the VJ id doesn't exist.
For the moment, it doesn't erase the Meta VJ concerned.
Tests show you the current behavior.

### DIFF
--- a/source/kraken/realtime.cpp
+++ b/source/kraken/realtime.cpp
@@ -635,13 +635,14 @@ void handle_realtime(const std::string& id,
             meta_vj = data.pt_data->meta_vjs.emplace(trip_update.trip().trip_id());
             // TODO : pick a meaningful TZ
             meta_vj->tz_handler = data.pt_data->tz_manager.get_first_timezone();
-        } else if (is_cancelled_trip(trip_update)) {
-            LOG4CPLUS_DEBUG(log, "Can't cancel a trip that doesn't exist: ignoring trip update id "
-                    << trip_update.trip().trip_id());
-            return;
         } else {
-            LOG4CPLUS_DEBUG(log, "Meta VJ doesn't exist without Added trip type: ignoring trip update id "
-                    << trip_update.trip().trip_id());
+            LOG4CPLUS_WARN(log, "Cannot perform operation on an unknown Meta VJ (other than adding trip)"
+                    << ", trip id: " << trip_update.trip().trip_id()
+                    << ", effect: " << get_wordings(get_trip_effect(trip_update.GetExtension(kirin::effect))));
+            if (trip_update.stop_time_update_size()) {
+                LOG4CPLUS_WARN(log, "Meta VJ 1st stop time departure: "
+                    << trip_update.stop_time_update(0).departure().time());
+            }
             return;
         }
     } else {
--- a/source/kraken/realtime.cpp
+++ b/source/kraken/realtime.cpp
@@ -635,6 +635,10 @@ void handle_realtime(const std::string& id,
             meta_vj = data.pt_data->meta_vjs.emplace(trip_update.trip().trip_id());
             // TODO : pick a meaningful TZ
             meta_vj->tz_handler = data.pt_data->tz_manager.get_first_timezone();
+        } else if (is_cancelled_trip(trip_update)) {
+            LOG4CPLUS_DEBUG(log, "Can't cancelled a trip that doesn't exist: ignoring trip update id "
+                    << trip_update.trip().trip_id());
+            return;
         } else {
             LOG4CPLUS_DEBUG(log, "Meta VJ doesn't exist without Added trip type: ignoring trip update id "
                     << trip_update.trip().trip_id());
--- a/source/kraken/realtime.cpp
+++ b/source/kraken/realtime.cpp
@@ -636,7 +636,7 @@ void handle_realtime(const std::string& id,
             // TODO : pick a meaningful TZ
             meta_vj->tz_handler = data.pt_data->tz_manager.get_first_timezone();
         } else if (is_cancelled_trip(trip_update)) {
-            LOG4CPLUS_DEBUG(log, "Can't cancelled a trip that doesn't exist: ignoring trip update id "
+            LOG4CPLUS_DEBUG(log, "Can't cancel a trip that doesn't exist: ignoring trip update id "
                     << trip_update.trip().trip_id());
             return;
         } else {
--- a/source/kraken/tests/realtime_test.cpp
+++ b/source/kraken/tests/realtime_test.cpp
@@ -70,6 +70,7 @@ make_cancellation_message(const std::string& vj_uri, const std::string& date) {
     trip->set_schedule_relationship(transit_realtime::TripDescriptor_ScheduleRelationship_CANCELED);
     trip->SetExtension(kirin::contributor, "cow.owner");
     trip_update.SetExtension(kirin::trip_message, "cow on the tracks");
+    trip_update.SetExtension(kirin::effect, transit_realtime::Alert_Effect::Alert_Effect_NO_SERVICE);
     return trip_update;
 }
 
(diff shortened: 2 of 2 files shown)


source/kraken/realtime.cpp
```diff
@@ -635,6 +635,10 @@ void handle_realtime(const std::string& id,
             meta_vj = data.pt_data->meta_vjs.emplace(trip_update.trip().trip_id());
             // TODO : pick a meaningful TZ
             meta_vj->tz_handler = data.pt_data->tz_manager.get_first_timezone();
+        } else if (is_cancelled_trip(trip_update)) {
+            LOG4CPLUS_DEBUG(log, "Can't cancelled a trip that doesn't exist: ignoring trip update id "
+                    << trip_update.trip().trip_id());
```
nonifier: Don't we want to return some kind of error in the protobuf ? What would be the end result from kirin in this case ?

benoit-bst: The error managing is not complete. 
In many case, we have only a log message to trace the error concerned.
The main job is to protect the code against `segfault or bad behaviours` inside Kraken engine.
We don't reject the flow in Kirin because the *COTS* format is good, only write a `Can't cancelled a trip that doesn't exist: ignoring trip update id` log Kraken message.

A good upgrade would be to fill *Prometheus probe* to track the behaviour. I have looked and is not easy to plug it at this code level. That needs a lot of changes. Maybe in a near future

source/kraken/realtime.cpp
```diff
@@ -635,6 +635,10 @@ void handle_realtime(const std::string& id,
             meta_vj = data.pt_data->meta_vjs.emplace(trip_update.trip().trip_id());
             // TODO : pick a meaningful TZ
             meta_vj->tz_handler = data.pt_data->tz_manager.get_first_timezone();
+        } else if (is_cancelled_trip(trip_update)) {
+            LOG4CPLUS_DEBUG(log, "Can't cancelled a trip that doesn't exist: ignoring trip update id "
```
kadhikari: typo : can't cancel

source/kraken/tests/realtime_test.cpp
```diff
@@ -3340,3 +3341,167 @@ BOOST_FIXTURE_TEST_CASE(trip_id_that_doesnt_exist_must_be_rejected_in_classical_
...
+
+    // Check meta vj table
+    BOOST_REQUIRE_EQUAL(pt_data.meta_vjs.size(), 3);
+
+    // Check vj table
+    BOOST_REQUIRE_EQUAL(pt_data.vehicle_journeys_map.size(), 2);
+    BOOST_REQUIRE_EQUAL(pt_data.vehicle_journeys.size(), 2);
+}
+
+BOOST_FIXTURE_TEST_CASE(cant_cancelled_trip_that_doesnt_exist, AddTripDataset) {
```
kadhikari: typo: cant_cancel_...

source/kraken/tests/realtime_test.cpp
```diff
@@ -3340,3 +3341,167 @@ BOOST_FIXTURE_TEST_CASE(trip_id_that_doesnt_exist_must_be_rejected_in_classical_
...
+
+    auto& pt_data = *b.data->pt_data;
+
+    BOOST_CHECK(!pt_data.meta_vjs.exists("vj_id_doesnt_exist"));
+    BOOST_CHECK_EQUAL(pt_data.meta_vjs.size(), 1);
+    BOOST_REQUIRE_EQUAL(pt_data.vehicle_journeys_map.size(), 1);
+    BOOST_REQUIRE_EQUAL(pt_data.vehicle_journeys.size(), 1);
+
+    // Can't cancelled trip that doesn't exist
```
kadhikari: typo: Can't cancel..

kadhikari (approved):

source/kraken/realtime.cpp
```diff
@@ -635,6 +635,10 @@ void handle_realtime(const std::string& id,
             meta_vj = data.pt_data->meta_vjs.emplace(trip_update.trip().trip_id());
             // TODO : pick a meaningful TZ
             meta_vj->tz_handler = data.pt_data->tz_manager.get_first_timezone();
+        } else if (is_cancelled_trip(trip_update)) {
```
pbougue: I don't think this should be specific case.
We want to track it all the same, whether it's deletion or modification on a nonexistent trip, no?

source/kraken/realtime.cpp
pbougue: I agree with @benoit-bst .

There would be a lot of levels involved in such a process: kirin doesn't listen anything from kraken, it only publishes updates in a rabbitMQ queue. And Kirin checks if the trip exists before sending it to kraken.
But then again, I'm pretty sure that no-one reads kirin's responses to COTS feed POST... And the tool posting COTS is just reading a queue. So no specific point in responding to the feed received.

source/kraken/tests/realtime_test.cpp
```diff
@@ -3340,3 +3341,167 @@ BOOST_FIXTURE_TEST_CASE(trip_id_that_doesnt_exist_must_be_rejected_in_classical_
...
+            RTStopTime("stop_point:E", "20190101T0830"_pts).added(),
+            RTStopTime("stop_point:F", "20190101T0900"_pts).added(),
+            RTStopTime("stop_point:G", "20190101T0930"_pts).added(),
+        },
+        transit_realtime::Alert_Effect::Alert_Effect_ADDITIONAL_SERVICE,
+        comp_uri,
+        phy_mode_uri);
+
+    navitia::handle_realtime("feed-1", timestamp, new_trip, *b.data, true, true);
+    b.data->build_raptor();
```
pbougue, line 3377: note to self: will have to change those in my PR refactoring tests after route RT add.

source/kraken/tests/realtime_test.cpp
```diff
@@ -3340,3 +3341,167 @@ BOOST_FIXTURE_TEST_CASE(trip_id_that_doesnt_exist_must_be_rejected_in_classical_
...
+
+    auto& pt_data = *b.data->pt_data;
+
+    BOOST_CHECK(!pt_data.meta_vjs.exists("vj_id_doesnt_exist"));
+    BOOST_CHECK_EQUAL(pt_data.meta_vjs.size(), 1);
+    BOOST_REQUIRE_EQUAL(pt_data.vehicle_journeys_map.size(), 1);
+    BOOST_REQUIRE_EQUAL(pt_data.vehicle_journeys.size(), 1);
+
+    // Can't cancel trip that doesn't exist
+    transit_realtime::TripUpdate remove_trip = ntest::make_delay_message("vj_id_doesnt_exist",
```
pbougue, line 3487: note to self: rename this `make_delay_message` that does more than just delays now.

source/kraken/tests/realtime_test.cpp
```diff
@@ -70,6 +70,7 @@ make_cancellation_message(const std::string& vj_uri, const std::string& date) {
     trip->set_schedule_relationship(transit_realtime::TripDescriptor_ScheduleRelationship_CANCELED);
```
pbougue, line 70: note to self: this method should probably just be sugar around `make_delay_message` once renamed.

source/kraken/realtime.cpp
benoit-bst: I think you've right. 
We must to reject all COTS if it is **not a added trip** and the **Meta VJ doesn't exist**. 
This line just allows to have a specific log for the cancelled case, but it is not critic. 
I don't care, we can suppress it if you want. 

The most important thing is to reject bad case. :100:

nonifier: I think that we all agree that this check is redundant with the following one. Although, I agree with Benoit that it would be interesting to understand where this tip_update comes from (for debugging purposes).

As Kirin should already do the check of the existing VJ, we might want to **warn** about it. And to be able to track back which VJ we are working with, we need the log the timestamp/effect.

```cpp
} else {
            LOG4CPLUS_WARN(log, "Cannot perform operation on an unknown Meta VJ (other than adding trip)"
                    << ", trip id: " << trip_update.trip().trip_id() 
                    << ", effect: " << Alert_Effect_Name(trip_update.GetExtension(kirin::effect)));
            if(trip_update.stop_time_update_size()) {
                LOG4CPLUS_WARN(log, "Meta VJ 1sr stop time departure: "
                    << trip_update.stop_time_update(0).departure().time());
            }
            return;
}
```

source/kraken/realtime.cpp
nonifier: :+1:

nonifier (approved):

pbougue (approved):

PR #2723, opened 2019-03-25
[ALL] Adding taxi mode

AurelienLP:
https://jira.kisio.org/browse/NAVP-1234
https://jira.kisio.org/browse/NAVP-1235
https://jira.kisio.org/browse/NAVP-1238


source/jormungandr/jormungandr/street_network/taxi.py
```diff
@@ -0,0 +1,190 @@
...
+        self, mode, pt_object_origin, pt_object_destination, fallback_extremity, request, direct_path_type
+    ):
+
+        copy_request = copy.deepcopy(request)
+        copy_request["car_no_park_speed"] = copy_request["taxi_speed"]
+        response = self.street_network._direct_path(
+            mode, pt_object_origin, pt_object_destination, fallback_extremity, copy_request, direct_path_type
+        )
+
+        if response and len(response.journeys):
```
xlqian: `response.journeys` may raise exception if no solution is found by the connector (`journeys` is optional in the proto), we have to use `HasField` to test if there is a solution.. then iterate over response.journeys... `len(response.journeys)` is not necessary

source/jormungandr/jormungandr/street_network/taxi.py
```diff
@@ -0,0 +1,190 @@
...
+        self, origins, destinations, street_network_mode, max_duration, request, **kwargs
+    ):
+
+        copy_request = copy.deepcopy(request)
+        copy_request["car_no_park_speed"] = copy_request["taxi_speed"]
+        response = self.street_network.get_street_network_routing_matrix(
+            origins, destinations, street_network_mode, max_duration, copy_request, **kwargs
+        )
+
+        if response and len(response.rows) and len(response.rows[0].routing_response):
```
xlqian: we have to test the presence of `rows` too

source/jormungandr/tests/routing_tests_experimental.py
```diff
@@ -347,3 +349,112 @@ def test_other_filtering(self):
...
+        assert sections[3].get('departure_date_time') == '20120614T080102'
+        assert sections[3].get('arrival_date_time') == '20120614T080222'
+        assert sections[3].get('duration') == 80
+        assert sections[3].get('type') == 'street_network'
+
+        query += "&max_duration=0"
+        response = self.query_region(query)
+        # the pt journey is eliminated
+        self.is_valid_journey_response(response, query)
+        assert len(response['journeys']) == 1
```
xlqian, line 460: just for record: since this solution is a direct_path type solution with taxi, we should tag this journey with taxi... we can do it in another PR

xlqian: test 
```python 
assert 'taxi' in response['tags']
```

source/jormungandr/jormungandr/interfaces/v1/Journeys.py
```diff
@@ -482,6 +487,13 @@ def _set_specific_params(mod):
                 new_datetime = self.convert_to_utc(original_datetime)
             args['datetime'] = date_to_timestamp(new_datetime)
 
+            scenario_name = i_manager.get_instance_scenario_name(self.region, args.get('_override_scenario'))
+
+            if scenario_name == "new_default" and (
+                "taxi" in args["origin_mode"] or "taxi" in args["destination_mode"]
+            ):
+                abort(400, message="taxi is not available with new_default scenario")
```
xlqian, line 495: it would be great if this is tested

xlqian (approved):

source/jormungandr/jormungandr/street_network/taxi.py
AurelienLP: response.journeys is repeated. I'll just remove len(response.journeys) :D

source/jormungandr/jormungandr/street_network/taxi.py
AurelienLP: Same as above

source/jormungandr/tests/routing_tests_experimental.py
AurelienLP: Yes indeed!

source/jormungandr/jormungandr/interfaces/v1/Journeys.py
AurelienLP: Adding test

TeXitoi (approved):

source/jormungandr/jormungandr/street_network/taxi.py
```diff
@@ -0,0 +1,190 @@
...
+        response = self.street_network._direct_path(
+            mode, pt_object_origin, pt_object_destination, fallback_extremity, copy_request, direct_path_type
+        )
+
+        if response:
+            for journey in response.journeys:
+                for section in journey.sections:
+                    section.street_network.mode = response_pb2.Taxi
+
+                # We don't add an addtional waiting section for direct_path
```
mbounabi: Minor: add**i**tional
Also, I don't understand the condition and the function called: If it isn't a direct path, you call a function to add a section in direct path?

AurelienLP: When you have a fallback with taxi mode, you add a waiting section before (last_section = taxi) or after (first_section = taxi) the taxi section. If it's just a direct_path you don't add anything.

mbounabi (approved):

PR #2726, opened 2019-03-28
[Tyr] New endpoint /equipments_providers + db management

mbounabi:
https://jira.kisio.org/browse/NAVP-1249
https://jira.kisio.org/browse/NAVP-1250

### DIFF
--- a/source/navitiacommon/navitiacommon/models/__init__.py
+++ b/source/navitiacommon/navitiacommon/models/__init__.py
@@ -33,14 +33,11 @@ from __future__ import absolute_import
 import uuid
 import re
 from navitiacommon.sqlalchemy import SQLAlchemy
-from geoalchemy2.types import Geography
-from flask import current_app
-from sqlalchemy.orm import load_only, backref, aliased
+from sqlalchemy.orm import backref
 from datetime import datetime
 from sqlalchemy import func, and_, UniqueConstraint, cast, true, false
 from sqlalchemy.dialects.postgresql import ARRAY, UUID, INTERVAL
 from sqlalchemy.dialects.postgresql.json import JSONB
-from navitiacommon.utils import street_source_types
 
 from navitiacommon import default_values
 import os
@@ -787,3 +784,4 @@ class AutocompleteParameter(db.Model, TimestampMixin):  # type: ignore
 
 # import at the end to prevent circular dependencies
 from navitiacommon.models.bss_provider import BssProvider
+from navitiacommon.models.equipments_providers import EquipmentsProvider
--- a/source/navitiacommon/navitiacommon/models/equipments_providers.py
+++ b/source/navitiacommon/navitiacommon/models/equipments_providers.py
@@ -36,7 +36,7 @@ from sqlalchemy.dialects.postgresql.base import ARRAY
 from navitiacommon.models import db, TimestampMixin
 
 
-class EquipmentsProvider(db.Model, TimestampMixin):
+class EquipmentsProvider(db.Model, TimestampMixin):  # type: ignore
     id = db.Column(db.Text, primary_key=True)
     instances = db.Column(ARRAY(db.Text), unique=False, nullable=False)
     klass = db.Column(db.Text, unique=False, nullable=False)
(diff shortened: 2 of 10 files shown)


source/navitiacommon/navitiacommon/models/__init__.py
```diff
@@ -33,14 +33,11 @@
 import uuid
 import re
 from navitiacommon.sqlalchemy import SQLAlchemy
-from geoalchemy2.types import Geography
-from flask import current_app
-from sqlalchemy.orm import load_only, backref, aliased
+from sqlalchemy.orm import backref
```
xlqian, line 36: it seems `backref` is not used?

xlqian (commented):
looks good to me

source/navitiacommon/navitiacommon/models/__init__.py
mbounabi: Yes it is.
https://github.com/CanalTP/navitia/blob/9eca83d6dfd6b1d3fccaf96de22aaa4a2b26730f/source/navitiacommon/navitiacommon/models/__init__.py#L267

benoit-bst (approved):

source/navitiacommon/navitiacommon/models/equipments_providers.py
```diff
@@ -0,0 +1,70 @@
...
+
+from __future__ import absolute_import
+
+from sqlalchemy.dialects.postgresql.json import JSONB
+from sqlalchemy.dialects.postgresql.base import ARRAY
+from navitiacommon.models import db, TimestampMixin
+
+
+class EquipmentsProvider(db.Model, TimestampMixin):
+    id = db.Column(db.Text, primary_key=True)
```
kadhikari, line 40: What about __tablename__ = 'equipments_provider' ?
Without tablename how can the class EquipmentsProvider be related to the table "equipments_provider" ?

source/tyr/migrations/versions/1c8334cd1997_add_equipment_provider.py
```diff
@@ -0,0 +1,32 @@
...
+from sqlalchemy.dialects import postgresql
+
+def upgrade():
+    op.create_table('equipments_provider',
+    sa.Column('id', sa.Text(), nullable=False),
+    sa.Column('instances', postgresql.ARRAY(sa.Text()), nullable=False),
+    sa.Column('klass', sa.Text(), nullable=False),
+    sa.Column('discarded', sa.Boolean(), nullable=False),
+    sa.Column('args', postgresql.JSONB(), server_default='{}', nullable=True),
+    sa.Column('created_at', sa.DateTime(), nullable=False),
```
kadhikari: For all attributs nullable=False, don't we need  "server_default='false' or 'true' "as in https://github.com/CanalTP/navitia/blob/cff21de8d98540ada4a6193f998f8023768d740d/source/sql/alembic/versions/29fc422c56cb_add_areas_in_stop_points.py#L21 ?

source/navitiacommon/navitiacommon/models/equipments_providers.py
mbounabi: From the flask-SQLAlchemy doc:

> Some parts that are required in SQLAlchemy are optional in Flask-SQLAlchemy. For instance the table name is automatically set for you unless overridden. It’s derived from the class name converted to lowercase and with “CamelCase” converted to “camel_case”. To override the table name, set the __tablename__ class attribute.

source/tyr/migrations/versions/1c8334cd1997_add_equipment_provider.py
mbounabi: Changing my message:
Except for the boolean 'discarded', there can't be default values for other parameters.

kadhikari: In that case the function "def __init__(self, id, json=None) should always send a valid json instead of None as default value. Non?

mbounabi: No. If you look at the code, if an empty json is passed, the function will look for keys that are absent. The json has to be valid if passed and it is checked in tyr before adding to db

kadhikari (approved):

PR #2732, opened 2019-04-03
Jormun: handle zone and zone_type from bragi

kinnou02:
since https://github.com/CanalTP/mimirsbrunn/pull/298 bragi return admin
as `type=zone` and added a property `zone_type` set to `city` cities.

This PR restore the compatibility for bragi by handling those.
zone_type other than `city` are still filtered, adding this notion in
navitia is tracked in NAVP-1225.

### DIFF
--- a/source/jormungandr/jormungandr/interfaces/v1/serializer/geocode_json.py
+++ b/source/jormungandr/jormungandr/interfaces/v1/serializer/geocode_json.py
@@ -263,6 +263,7 @@ class GeocodePlacesSerializer(serpy.DictSerializer):
         map_serializer = {
             'city': GeocodeAdminSerializer,
             'administrative_region': GeocodeAdminSerializer,
+            'zone': GeocodeAdminSerializer,
             'street': GeocodeAddressSerializer,
             'house': GeocodeAddressSerializer,
             'poi': GeocodePoiSerializer,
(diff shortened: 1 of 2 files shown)


source/jormungandr/jormungandr/interfaces/v1/serializer/geocode_json.py
```diff
@@ -274,6 +275,11 @@ def get_places(self, obj):
             if not type_ or type_ not in map_serializer:
                 logging.getLogger(__name__).error('Place not serialized (unknown type): {}'.format(feature))
                 continue
+            zone_type = feature.get('properties', {}).get('geocoding', {}).get('zone_type')
+            # TODO: do something smart with other zone type
+            if type_ == 'zone' and zone_type != 'city':
```
antoine-de, line 281: you're sure you don't want the other ones ?

there are only cities in navitia for the moment ?

antoine-de (approved):

source/jormungandr/jormungandr/interfaces/v1/serializer/geocode_json.py
```diff
@@ -274,6 +275,11 @@ def get_places(self, obj):
             if not type_ or type_ not in map_serializer:
                 logging.getLogger(__name__).error('Place not serialized (unknown type): {}'.format(feature))
                 continue
+            zone_type = feature.get('properties', {}).get('geocoding', {}).get('zone_type')
```
nonifier: Would be nice to read the geocoding object from the feature only once (see line 274 above).

nonifier (approved):

source/jormungandr/jormungandr/interfaces/v1/serializer/geocode_json.py
kinnou02: I don't think we should return other type without returning the zone type in the response. 
Previously in mimir we didn't import something else than cities, so we keep the same behavior

antoine-de (approved):

pbougue (approved):

PR #2737, opened 2019-04-04
Attach physical mode to vj: tests of verifications 

kadhikari:
**For a new trip to be added the physical_mode is managed as following:**

- **Kirn:** 
  For "indicateurFer": "FERRE" in Flux cots 'physical_mode:LongDistanceTrain' is used in GTFS-RT
  For "indicateurFer": "ROUTIER" in Flux cots 'physical_mode:Coach' is used in GTFS-RT

- **Kraken: Already done in kraken**
  If physical_mode exists in GTFS-RT then fetch in kraken. The disruption is rejected if the physical_mode doesn't exist.

Concerns:  https://jira.kisio.org/browse/NAVP-1257

### DIFF
--- a/source/jormungandr/tests/kirin_realtime_tests.py
+++ b/source/jormungandr/tests/kirin_realtime_tests.py
@@ -1763,7 +1763,7 @@ class TestKirinAddNewTripWithWrongPhysicalMode(MockKirinDisruptionsFixture):
         disruptions_after = self.query_region(disruption_query)
         assert nb_disruptions_before == len(disruptions_after['disruptions'])
 
-        # /journeys before (only direct walk)
+        # / Journeys: as no trip on pt added, only direct walk.
         C_B_query = (
             "journeys?from={f}&to={to}&data_freshness=realtime&"
             "datetime={dt}&_current_datetime={dt}".format(
@@ -1775,11 +1775,6 @@ class TestKirinAddNewTripWithWrongPhysicalMode(MockKirinDisruptionsFixture):
         self.is_valid_journey_response(response, C_B_query)
         assert len(response['journeys']) == 1
         assert 'non_pt_walking' in response['journeys'][0]['tags']
-
-        # /pt_objects before
-        ptobj_query = 'pt_objects?q={q}&_current_datetime={dt}'.format(q='adi', dt='20120614T080000')  # ++typo
-        response = self.query_region(ptobj_query)
-        assert 'pt_objects' not in response
 
         # Check that no vehicle_journey exists on the future realtime-trip
         vj_query = 'vehicle_journeys/{vj}?_current_datetime={dt}'.format(
--- a/source/jormungandr/tests/kirin_realtime_tests.py
+++ b/source/jormungandr/tests/kirin_realtime_tests.py
@@ -1724,8 +1724,7 @@ class TestKirinAddNewTripWithWrongPhysicalMode(MockKirinDisruptionsFixture):
     def test_add_new_trip_with_wrong_physical_mode(self):
         """
         1. send a disruption to create a new trip with physical_mode absent in kaken
-        2. test that no journey is possible using this new trip
-        3. test that no PT-Ref objects were created
+        2. check of journey, disruption and PT-Ref objects to verify that no trip is added
         """
         disruption_query = 'disruptions?_current_datetime={dt}'.format(dt='20120614T080000')
         disruptions_before = self.query_region(disruption_query)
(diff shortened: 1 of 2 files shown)


benoit-bst (approved):

source/kraken/apply_disruption.cpp
```diff
@@ -284,8 +284,7 @@ struct add_impacts_visitor : public apply_impacts_visitor {
                 if (! mvj->get_base_vj().empty()) {
                     vj->physical_mode = mvj->get_base_vj().at(0)->physical_mode;
                 } else {
-                    // for protection, use the physical_modes[0]
-                    // TODO : Create default physical mode
+                    // for protection, use the physical_modes[0] instead of creating a new one
```
pbougue, line 293: We may default to Coach (as this will probably be the most common, and shared need between transport-providers)
But this code is never supposed to be triggered, it's a protection, so...

kadhikari:
jenkins test this please

source/jormungandr/tests/kirin_realtime_tests.py
```diff
@@ -1712,6 +1714,77 @@ def test_add_new_trip(self):
...
+            disruption_id='new_trip',
+            effect='additional_service',
+            physical_mode_id='physical_mode:Toto',  # this physical mode doesn't exist in kraken
+        )
+
+        # Check there is no new disruption
+        disruptions_after = self.query_region(disruption_query)
+        assert nb_disruptions_before == len(disruptions_after['disruptions'])
+
+        # /journeys before (only direct walk)
```
pbougue: Comment is "outdated".
(same on 2 other comments below)

pbougue (commented):
As for the "empty physical_mode" case, I understand it this way:
* It's tested on other tests (trip updates) without filling the extension, so we should be OK, right?
-> if disruption (impact) doesn't have extension 'physical_mode' then it's taken from metaVJ.
-> if metaVJ doesn't have any  physical_mode then it's the first from physical_mode_list.
* How does it behave on trip add? It used to work, so we fallback on modes[0]? In that case I prefer fallback on Coach, or rejection.
-> On trip add if disruption (impact) has an extension 'physical_mode', used if exists if not trip rejected.
-> On trip add, if disruption (impact) has no extension 'physical_mode', we use the first one from the physical_mode_list

kadhikari:
jenkins test this please

source/jormungandr/tests/kirin_realtime_tests.py
```diff
@@ -1717,6 +1725,136 @@ def test_add_new_trip(self):
...
+                f='stop_point:stopC', to='stop_point:stopB', dt='20120614T080000'
+            )
+        )
+        response = self.query_region(C_B_query)
+        assert not has_the_disruption(response, 'new_trip')
+        self.is_valid_journey_response(response, C_B_query)
+        assert len(response['journeys']) == 1
+        assert 'non_pt_walking' in response['journeys'][0]['tags']
+
+        # Check that no vehicle_journey exists on the future realtime-trip
```
pbougue: As referred here: https://github.com/CanalTP/navitia/pull/2737#discussion_r272526485
It's not a "future" RT trip.

source/jormungandr/tests/kirin_realtime_tests.py
```diff
@@ -1493,8 +1494,13 @@ def test_stop_time_with_detour_and_arrival_before_deleted_at_the_end(self):
         assert response['journeys'][0]['status'] == 'DETOUR'
         assert response['journeys'][0]['sections'][0]['type'] == 'public_transport'
         assert response['journeys'][0]['sections'][0]['data_freshness'] == 'realtime'
+        assert response['journeys'][0]['sections'][0]['display_informations']['physical_mode'] == 'Tramway'
         assert has_the_disruption(response, 'stop_time_with_detour')
 
+        # Tramway is the first physical_mode in NTFS
```
pbougue: I'd add: "Tramway is the first physical_mode in NTFS`, but we might pick mode in a smarter way in the future`"

source/kraken/apply_disruption.cpp
```diff
@@ -290,8 +290,7 @@ struct add_impacts_visitor : public apply_impacts_visitor {
                 if (!mvj->get_base_vj().empty()) {
```
pbougue, line 290: As discussed with the team: this base_VJ's mode should always be used when present (and it should prevail)
We should add a test that the correct one (from baseVJ) is used when Kirin provides a physical_mode.

source/jormungandr/tests/kirin_realtime_tests.py
```diff
@@ -1785,6 +1791,70 @@ def test_add_new_trip_with_wrong_physical_mode(self):
         assert 'vehicle_journeys' not in response
 
 
+@dataset(MAIN_ROUTING_TEST_SETTING)
+class TestKirinAddNewTripWithoutPhysicalMode(MockKirinDisruptionsFixture):
+    def test_add_new_trip_without_physical_mode(self):
+        """
+        1. send a disruption to create a new trip with physical_mode absent in kaken
```
pbougue: its `without physical_mode`

source/jormungandr/tests/kirin_realtime_tests.py
```diff
@@ -1717,6 +1725,136 @@ def test_add_new_trip(self):
...
+            ],
+            disruption_id='new_trip',
+            effect='additional_service',
+        )
+
+        # Check that a new disruption is added
+        disruptions_after = self.query_region(disruption_query)
+        assert nb_disruptions_before + 1 == len(disruptions_after['disruptions'])
+
+        # / Journeys: as no trip on pt added, only direct walk.
```
pbougue: unrelated comment

kadhikari:
A new ticket is created for correction in kraken : https://jira.kisio.org/browse/NAVP-1284

pbougue (approved):

benoit-bst:
the physical mode "bug" is fix into this PR : https://github.com/CanalTP/navitia/pull/2748

PR #2738, opened 2019-04-04
Kraken: Add and use PT_Data::get_or_create_meta_vehicle_journey() 

pbougue:
Also add a PT_Data::get_main_timezone() on the way
Could not think of a better (and "light") way to handle default TZ as we lose the information in ed (initially linked to network) : https://github.com/CanalTP/navitia/blob/dev/source/ed/ed_persistor.cpp#L1102

:mag_right: Hide whitespace changes for review

TODO:
- [x] merge https://github.com/CanalTP/utils/pull/80 and update submodule once build is OK


benoit-bst (approved):
More clear :butterfly:

kadhikari (approved):
Easy to understand and useful

kinnou02 (approved):

PR #2742, opened 2019-04-08
[Kraken] Equipment reports API 

nonifier:
* Add new endpoint to Kraken `equipment_reports`
* Add the builder the ability to create stop points with codes
* forbidden uris and pagination is supported

The new equipment_reports feature lists all stop areas that have a stop point with a specific code and returns them grouped per line. 

[NAVP-1219](https://jira.kisio.org/browse/NAVP-1219)


source/equipment/tests/CMakeLists.txt
```diff
@@ -0,0 +1,11 @@
+#We use the BOOST_LIBS define is the parent
```
benoit-bst: Can suppress comment

source/equipment/equipment_api.cpp
```diff
@@ -0,0 +1,99 @@
...
+    return res;
+}
+
+void equipment_reports(PbCreator& pb_creator, const std::string& filter,
+                       int count, int depth, int start_page,
+                       const ForbiddenUris& forbidden_uris) {
+    const type::Data& data = *pb_creator.data;
+    try {
+        const auto sas_per_line = get_stop_areas_per_line(data, filter, forbidden_uris);
+        const auto paginated_sas = paginate(sas_per_line, count, start_page);
```
pbougue: We could improve that mechanism as we do a lot of useless ptref-filtering queries on SA here (same on other endpoints).
Here maybe just fill lines in vector before calling paginate, then fill usefull SAs.

Don't know the importance of this, it depends on perfs, caching, and possibility to mutualize that behavior (also complexity of doing it).

source/equipment/equipment_api.cpp
```diff
@@ -0,0 +1,99 @@
...
+    }
+
+    void fill_equipment_to_pb(const equipment::StopAreasPerLine& sas_per_line,
+                              PbCreator& pb_creator, int depth) {
+        for(const auto& line : sas_per_line) {
+            auto report = pb_creator.add_equipment_reports();
+            pb_creator.fill(line.first, report->mutable_line(), depth);
+            for(const auto& sa : line.second) {
+                auto sa_equip = report->add_stop_area_equipments();
+                pb_creator.fill(sa, sa_equip->mutable_stop_area(), depth);
```
pbougue: I think that we should fix a min depth, as we need stop_points and their codes to be able to fill it later in jormun, no?
When requesting `v1/stop_areas?depth=0` we get no SP, so that could be a problem (not sure).

source/ed/build_helper.cpp
```diff
@@ -349,42 +349,48 @@ SA::SA(builder & b, const std::string & sa_name, double x, double y,
...
-            sp->set_property(navitia::type::hasProperties::WHEELCHAIR_BOARDING);
-        if(bike_accepted)
-            sp->set_property(navitia::type::hasProperties::BIKE_ACCEPTED);
-        sp->coord.set_lon(x);
-        sp->coord.set_lat(y);
-
-        sp->stop_area = sa;
-        b.sps[sp->name] = sp;
-        sa->stop_point_list.push_back(sp);
+        SP::create_stop_point(b, sa, "stop_point:"+ sa_name, {}, x, y, wheelchair_boarding, bike_accepted);
```
pbougue: :+1:

source/equipment/tests/equipment_api_tests.cpp
```diff
@@ -0,0 +1,204 @@
...
+using boost::posix_time::time_period;
+
+struct logger_initialized {
+    logger_initialized() { navitia::init_logger(); }
+};
+BOOST_GLOBAL_FIXTURE( logger_initialized );
+
+namespace {
+
+class Test_fixture {
```
pbougue: Weird case.
Maybe also rename it to EquipmentsTestFixture (not sure about the case here)?

pbougue (commented):
Can you add test on:
* depth: to me we should have the same amount of equipments listed no matter depth (0 to 3)
  It looks OK from worker test (contrary to my simple hand-test on `/stop_areas`), but explicitly fixing it to depth=0 would be nice.
* forbidden_uris: add a tests forbidding sa:B and check we get no solution? maybe one forbidding another SA and one on line 1?

benoit-bst (changes requested):
It should be preferable to add `equipment_details->id` within the _Kraken request_ to match with the web service response.

source/equipment/equipment_api.cpp
nonifier: as we mentioned, we'll not be using the stop-points to get the codes from them. I've actually forgotten to fill the protobuf properly :confounded:

source/equipment/equipment_api.cpp
nonifier: That's done. 
Also, a much better optimisation would be to pass an array of values in the ptref filter; like 
```sh
stop_point.has_code_type("TCL_escalier", "TCL_assenceur", "TCL_catapulte")
``` 
instead of 
```sh
stop_point.has_code_type("TCL_escalier") or stop_point.has_code_type("TCL_assenceur") or ....
```

source/equipment/tests/equipment_api_tests.cpp
nonifier: :ok_hand:

source/equipment/tests/CMakeLists.txt
nonifier: done !

source/equipment/equipment_api.cpp
```diff
@@ -0,0 +1,140 @@
...
+    EquipmentReportList res;
+    for (const auto line : paginated_lines) {
+        const std::string sa_filter = build_ptref_sa_filter(line->uri, filter);
+        const type::Indexes sa_indexes = ptref::make_query(type::Type_e::StopArea, sa_filter, forbidden_uris, data);
+        const auto stop_areas = data.get_data<type::StopArea>(sa_indexes);
+
+        std::vector<equipment::StopAreaEquipment> sa_equipments;
+        for (type::StopArea* sa : stop_areas) {
+            const std::string sp_filter = build_ptref_sp_filter(sa->uri, sa_filter);
+            const type::Indexes sp_indexes = ptref::make_query(type::Type_e::StopPoint, sp_filter, data);
```
pbougue: You should probably also use forbidden_uris here (if still used depending on optim suggested).

source/equipment/equipment_api.cpp
```diff
@@ -0,0 +1,140 @@
...
+    const type::Indexes line_indices = ptref::make_query(type::Type_e::Line, filter, forbidden_uris, data);
+    const auto lines = data.get_data<type::Line>(line_indices);
+
+    total_lines = lines.size();
+    const auto paginated_lines = paginate(lines, count, start_page);
+
+    EquipmentReportList res;
+    for (const auto line : paginated_lines) {
+        const std::string sa_filter = build_ptref_sa_filter(line->uri, filter);
+        const type::Indexes sa_indexes = ptref::make_query(type::Type_e::StopArea, sa_filter, forbidden_uris, data);
```
pbougue, line 98: Another possible optim: request SP directly here, and sort/store them directly, using `sp.stop_area`.
You could then remove the corresponding test.

source/equipment/equipment_api.cpp
```diff
@@ -0,0 +1,140 @@
...
+
+const std::string build_ptref_sp_filter(const std::string& stop_area_uri, const std::string& filter) {
+    std::string sp_filter = "stop_area.uri=" + stop_area_uri;
+    if (filter.size()) {
+        sp_filter += " and (" + filter + ")";
+    }
+    return sp_filter;
+}
+
+void fill_equipment_to_pb(const equipment::EquipmentReportList& reports, PbCreator& pb_creator, int depth) {
```
benoit-bst, line 60:
```suggestion
void fill_equipment_to_pb(const equipment::EquipmentReportList& reports, PbCreator& pb_creator, const int depth) {
```

benoit-bst (approved):
Very clear ! 
I don't know, for the moment, if we can do better with runtime perf.
Let's leave that for another PR after full real perf tests, if it necessary.

Just ensure that Jenkins tests be Green

source/equipment/equipment_api.cpp
nonifier: No need to be `const` as it's passed by copy ;)

benoit-bst: but your copy can be const in this part of code ? Because it doesn't change inside the function

source/equipment/equipment_api.cpp
pbougue: As suggested by @benoit-bst it's OK to move that optim' in the JIRA ticket. (not sure about the use of forbidden_uris though).
Please just keep track of it in the ticket if you delay it (and mention the ticket here in comment).

xlqian: @pbougue Doing so, you'll have a lot of repeated stop_area

xlqian: but, yes, we'll save a query...

pbougue: To precise a bit my thinking:
Do `const type::Indexes sp_indexes = ptref::make_query(type::Type_e::StopPoint, sa_filter, forbidden_uris, data);`
Then with this fill a `map<SA, StopAreaEquipment>` (or something like that, avoiding re-creation of SA)

This would save one query per line*SA-with-code (should not be negligible).

xlqian (commented):
I agree with the optim suggested by @pbougue, the rest seems ok to me

source/equipment/equipment_api.cpp
pbougue: After discussion, given the emergency and the fact that we already tracked a :mountain: of technical debt on our projects: No JIRA ticket on that, idea is tracked here in github in case someone searches through the code.
And we wait for perf issues to optimize it.

pbougue (approved):

PR #2744, opened 2019-04-11
Remove 'instances' in 'equipments_provider'

mbounabi:


source/tyr/migrations/versions/553c4f353aeb_remove_instances_in_equipments_provider.py
```diff
@@ -0,0 +1,23 @@
...
+revision = '553c4f353aeb'
+down_revision = '204aae05372a'
+
+from alembic import op
+import sqlalchemy as sa
+from sqlalchemy.dialects import postgresql
+
+
+def upgrade():
+    op.drop_column('equipments_provider', 'instances')
```
kinnou02, line 19: You can't remove the column in this migration, it will break jormun during the update :(
You have to update the model in this PR, then do another one that update the model, this one should only be merged at the next release.

kinnou02 (changes requested):

source/tyr/migrations/versions/553c4f353aeb_remove_instances_in_equipments_provider.py
mbounabi: Even if the field was never used?

kinnou02: yes :(
to "test" it you can play the migration but keep the fields in the model, it will fail when querying the db

PR #2748, opened 2019-04-11
kraken: physical mode id in realtime, only impact additional service

benoit-bst:
After test into this PR https://github.com/CanalTP/navitia/pull/2737 that highlight a bug, this PR aims to fix that. Describe in JIRA : https://jira.kisio.org/browse/NAVP-1284

Actions :
- Fix the Bug
- Add Utest in Kraken
- Update the integration test  that highlight it


kinnou02 (approved):

pbougue (approved):
I approve but I just pushed on your branch, so... :man_judge: + :avocado: 
After my commit, review `apply_disruption.cpp` with "Hide whitespace changes" :wink:

PR #2749, opened 2019-04-12
kraken: Don't use cbegin() and cend() with RepeatedPtrField (protobuf version < 3)

benoit-bst:
The problem is visible with **Ubuntu 16.04**. The protobuf lib (_libprotoc 2.6.1_) doesn't support cbegin() and cend().

:warning: That brings the discussion to deprecated **Ubuntu 16.04** for the open source part


pbougue (approved):
Nota: issue also on debian:8

nonifier (approved):
My bad, I live in the future :D

PR #2751, opened 2019-04-15
[TYR] 'instances' in 'equipments_provider' is nullable

mbounabi:


source/navitiacommon/navitiacommon/models/equipments_providers.py
```diff
@@ -38,7 +38,7 @@
 
 class EquipmentsProvider(db.Model, TimestampMixin):  # type: ignore
     id = db.Column(db.Text, primary_key=True)
-    instances = db.Column(ARRAY(db.Text), unique=False, nullable=False)
+    instances = db.Column(ARRAY(db.Text), unique=False, nullable=True)
```
nonifier: Not sure it actually make sense to have Null instance name in the database? 
What was the use case behind this ?

nonifier (approved):

source/navitiacommon/navitiacommon/models/equipments_providers.py
mbounabi: The use is that this column is useless and we want to remove it. To avoid breaking anything during the next deployment, where 'instances' won't be set, the db has to accept that this field can be empty.

kadhikari (approved):

source/navitiacommon/navitiacommon/models/equipments_providers.py
pbougue: You have to leave the column in base, but remove it from the model (so: remove everything but the column in db).
Otherwise, it will break when we remove it from the base (old instance still trying to read it from base, event if it's doing nothing with it).

source/tyr/tyr/tasks.py
```diff
@@ -84,6 +84,7 @@ def import_data(files, instance, backup_file, async=True, reload=True, custom_ou
     :param backup_file: If True the files are moved to a backup directory, else they are not moved
     :param async: If True all jobs are run in background, else the jobs are run in sequence the function will only return when all of them are finish
     :param reload: If True kraken would be reload at the end of the treatment
+    :param custom_output_dir: subdirectory for the nav file created. If not given, the instance default one is taken
```
pbougue, line 87: the instance`'s id` is taken

pbougue (changes requested):

source/tyr/tyr/tasks.py
mbounabi: No

pbougue (approved):

PR #2755, opened 2019-04-16
[Kraken] Equipment details optimisation

nonifier:
2 optimisations performed:
- Add `collection.has_code_type(type1, type2, ...)` with multiple types as an parameter. It acts the same way as `disruption.tags(...)`.
- Retrieve stop area through stop points to avoid one level of lookup.

[#NAVP-1281](https://jira.kisio.org/browse/NAVP-1281)

### DIFF
--- a/source/equipment/tests/equipment_api_tests.cpp
+++ b/source/equipment/tests/equipment_api_tests.cpp
@@ -184,10 +184,10 @@ BOOST_FIXTURE_TEST_CASE(equipment_reports_test_api, EquipmentTestFixture) {
     const auto resp = pb_creator.get_response();
 
     map<string, set<string>> sa_per_line_uris;
-    for (const auto equip_rep : resp.equipment_reports()) {
+    for (const auto& equip_rep : resp.equipment_reports()) {
         BOOST_REQUIRE(equip_rep.has_line());
         set<string> stop_area_uris;
-        for (const auto sae : equip_rep.stop_area_equipments()) {
+        for (const auto& sae : equip_rep.stop_area_equipments()) {
             BOOST_REQUIRE(sae.has_stop_area());
             stop_area_uris.emplace(sae.stop_area().uri());
         }
--- a/source/equipment/tests/equipment_api_tests.cpp
+++ b/source/equipment/tests/equipment_api_tests.cpp
@@ -231,19 +231,18 @@ BOOST_FIXTURE_TEST_CASE(equipment_reports_should_have_stop_points_codes, Equipme
     const auto filter = "stop_point.has_code_type(CodeType1)";
     equipment::equipment_reports(pb_creator, filter, 1);
 
-    // with "count = 1", only "sa1" with "stop1" will be returned
-    // hence the expected codes are the one from stop point "stop1"
     BOOST_REQUIRE_EQUAL(pb_creator.equipment_reports_size(), 1);
-    const auto pb_codes =
-        pb_creator.get_response().equipment_reports(0).stop_area_equipments(0).stop_area().stop_points(0).codes();
 
-    vector<pair<string, string>> codes;
-    for (const auto& pb_code : pb_codes) {
-        codes.emplace_back(pb_code.type(), pb_code.value());
+    multiset<string> codes;
+    for (const auto& sae : pb_creator.get_response().equipment_reports(0).stop_area_equipments()) {
+        for (const auto& sp : sae.stop_area().stop_points()) {
+            for (const auto& c : sp.codes()) {
+                codes.emplace(c.value());
+            }
+        }
     }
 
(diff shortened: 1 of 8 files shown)


source/equipment/tests/equipment_api_tests.cpp
```diff
@@ -231,19 +231,18 @@ BOOST_FIXTURE_TEST_CASE(equipment_reports_should_have_stop_points_codes, Equipme
...
+        for (const auto& sp : sae.stop_area().stop_points()) {
+            for (const auto& c : sp.codes()) {
+                codes.emplace(c.value());
+            }
+        }
     }
 
-    vector<pair<string, string>> expected_codes = {{"CodeType1", "code1"}, {"CodeType1", "code2"}};
-
+    multiset<string> expected_codes = {"code1", "code2", "code6"};
```
pbougue, line 250: So you get only the ones from line A because of the count=1, right?
It may be worth a comment (to explain why we don't get "code8" too).

Just curious: You know why you didn't get "code6" previously?

source/ptreferential/ptreferential_utils.cpp
```diff
@@ -291,30 +291,35 @@ type::Indexes get_indexes_from_code(const type::Type_e type,
...
     auto collection = data.pt_data->collection<T>();
     for (const auto* obj : collection) {
         auto codes = data.pt_data->codes.get_codes<T>(obj);
-        if (codes.find(key) == codes.end()) {
-            continue;
+        for (const auto& key : keys) {
+            if (codes.find(key) == codes.end()) {
+                continue;
+            }
+            indexes.insert(obj->idx);
```
pbougue: You may also stop iterating on keys once you found one (`break` here, or reverse if condition for elegance)

pbougue (approved):

source/equipment/tests/equipment_api_tests.cpp
nonifier: We were previously populating only codes from the 1st stop point of the 1st stop area (aka. `stop1`). This stop point only has the 2 codes `code1` & `code2`. 
In the new version of the test, I'm gathering codes from **all** stop points of all stop areas. Hence the surfacing of `code6` which belong to 'stop3'.
I also wanted to make sure all codes were present and not duplicated (hence the the multiset).

I did add a comment anyway ;)

source/ptreferential/ptreferential_utils.cpp
nonifier: done :heavy_check_mark:

source/equipment/tests/equipment_api_tests.cpp
```diff
@@ -231,6 +231,11 @@ BOOST_FIXTURE_TEST_CASE(equipment_reports_should_have_stop_points_codes, Equipme
     const auto filter = "stop_point.has_code_type(CodeType1)";
     equipment::equipment_reports(pb_creator, filter, 1);
 
+    /*
+     * With count = 1, we'll only get stop_areas/stop_points from the 1sr line ("A")
```
pbougue: minor typo: 1s`t`

source/ptreferential/ptreferential_utils.cpp
```diff
@@ -291,30 +291,35 @@ type::Indexes get_indexes_from_code(const type::Type_e type,
...
-    get_indexes_from_code_type(const std::string& key, const Data& data) {
+    get_indexes_from_code_type(const std::vector<std::string>& keys, const Data& data) {
     Indexes indexes;
     auto collection = data.pt_data->collection<T>();
-    for (const auto* obj : collection) {
-        auto codes = data.pt_data->codes.get_codes<T>(obj);
-        if (codes.find(key) == codes.end()) {
-            continue;
+    for (const auto& key : keys) {
+        for (const auto* obj : collection) {
```
pbougue: Looks like we are losing the advantage of grouping keys, as we iterate multiple times on SP collection again, no?

In my previous comment I thought more of (sorry it wasn't clear):
```cpp
    for (const auto* obj : collection) {
        auto codes = data.pt_data->codes.get_codes<T>(obj);
        for (const auto& key : keys) {
            if (codes.find(key) != codes.end()) {
                indexes.insert(obj->idx);
                break;
            }
        }
    }
```

benoit-bst (approved):

source/ptreferential/ptreferential_utils.cpp
nonifier: I did some test with 100000 stop_areas/stop_points looking for a stop_point at the end of the list. Surprisingly enough, there is no difference in term of performances. I have a similar order of magnitude of cache miss with the 2 versions when profiling with valgrind. 

Although, I agree (as per my first implementation) that keys are likely to be much smaller than the collection size. And therefore, this would me better this way.

pbougue (approved):

kinnou02 (approved):

PR #2759, opened 2019-04-17
kraken & jormun: Attach vj to a route and line

kadhikari:
Concerned ticket: 
https://jira.kisio.org/browse/NAVP-1256
https://jira.kisio.org/browse/NAVP-1272

### DIFF
--- a/source/kraken/apply_disruption.cpp
+++ b/source/kraken/apply_disruption.cpp
@@ -193,11 +193,9 @@ static nt::Route* get_or_create_route(const nt::disruption::Impact& impact, nt::
     const auto& st_arrival = impact.aux_info.stop_times.back();
     const auto& sa_arrival = st_arrival.stop_time.stop_point->stop_area;
     std::string line_uri = "line:" + sa_depart->uri + "_" + sa_arrival->uri;
-    std::string line_name = "line:" + sa_depart->name + "_" + sa_arrival->name;
+    std::string line_name = sa_depart->name + " - " + sa_arrival->name;
     std::string route_uri = "route:" + sa_depart->uri + "_" + sa_arrival->uri;
-    std::string route_name = "route:" + sa_depart->name + "_" + sa_arrival->name;
-
-    // TODO: manage line.code when necessary
+    std::string route_name = sa_depart->name + " - " + sa_arrival->name;
     nt::Line* line = pt_data.get_or_create_line(line_uri, line_name, network, comm_mode);
     nt::Route* route = pt_data.get_or_create_route(route_uri, route_name, line, sa_arrival, "forward");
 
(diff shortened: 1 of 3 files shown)


nonifier (approved):

benoit-bst (approved):
I like the name trick : `Get the first and last stop_area to create route and line name`
More clear :+1:

source/kraken/apply_disruption.cpp
```diff
@@ -186,11 +186,20 @@ static nt::Route* get_or_create_route(const nt::disruption::Impact& impact, nt::
...
     nt::CommercialMode* comm_mode =
         pt_data.get_or_create_commercial_mode("commercial_mode:additional_service", "additional service");
+
+    // We get the first and last stop_area to create route and line
+    const auto& st_depart = impact.aux_info.stop_times.front();
+    const auto& sa_depart = st_depart.stop_time.stop_point->stop_area;
+    const auto& st_arrival = impact.aux_info.stop_times.back();
+    const auto& sa_arrival = st_arrival.stop_time.stop_point->stop_area;
+    std::string line_uri = "line:" + sa_depart->uri + "_" + sa_arrival->uri;
+    std::string line_name = "line:" + sa_depart->name + "_" + sa_arrival->name;
```
pbougue: I think we should do something like this:
```suggestion
    std::string line_name = sa_depart->name + " - " + sa_arrival->name;
```
It would look more like base-schedule's lines built the same way: https://api.navitia.io/v1/coverage/sncf/networks/network%3Asncf/lines

source/kraken/apply_disruption.cpp
```diff
@@ -186,11 +186,20 @@ static nt::Route* get_or_create_route(const nt::disruption::Impact& impact, nt::
...
+    std::string route_uri = "route:" + sa_depart->uri + "_" + sa_arrival->uri;
+    std::string route_name = "route:" + sa_depart->name + "_" + sa_arrival->name;
```
pbougue: Same:
```suggestion
    std::string route_name = sa_depart->name + " - " + sa_arrival->name;
```

source/kraken/apply_disruption.cpp
```diff
@@ -186,11 +186,20 @@ static nt::Route* get_or_create_route(const nt::disruption::Impact& impact, nt::
...
+
     // TODO: manage line.code when necessary
```
pbougue: You can remove that comment, no line has code in SNCF's coverage, so it's not a TODO anymore :wink: 
```suggestion
```

pbougue (commented):
Except the name building, all good to me.

pbougue (approved):

PR #2763, opened 2019-04-18
Remove the lowerisation of the equipment provider key

nonifier:
Yes, lowerisation actually is a word !

### DIFF
--- a/source/jormungandr/jormungandr/equipments/equipment_provider_manager.py
+++ b/source/jormungandr/jormungandr/equipments/equipment_provider_manager.py
@@ -58,7 +58,7 @@ class EquipmentProviderManager(object):
 
         # Init legacy providers from config file
         for provider in self.providers_config:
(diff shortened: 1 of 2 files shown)


mbounabi (approved):
Why? :interrobang:

kinnou02 (approved):

benoit-bst (approved):

pbougue (approved):
It's OK, but same as @mbounabi , can you add a bit of context on the bug?
(We may also lowerize when matching the keys?)
Also, any idea why it was lowerized previously?

kadhikari (approved):

mbounabi:
@pbougue The idea was to have equipments provider in lower cases to avoid having three different ones like sytral, Sytral and SYTRAL. 
However, the db is case insensitive so it can happen. So, as 4 famous persons once said, let it be!

PR #2765, opened 2019-04-19
[Jormungandr] Fix 'choose_best_instance' comparator + tests

mbounabi:
https://jira.kisio.org/browse/NAVITIAII-2721

### DIFF
--- a/source/jormungandr/jormungandr/interfaces/v1/journey_common.py
+++ b/source/jormungandr/jormungandr/interfaces/v1/journey_common.py
@@ -28,14 +28,14 @@
 # IRC #navitia on freenode
 # https://groups.google.com/d/forum/navitia
 # www.navitia.io
+
 from __future__ import absolute_import, print_function, unicode_literals, division
 from jormungandr import i_manager, fallback_modes
 from jormungandr.interfaces.v1.ResourceUri import ResourceUri
-from jormungandr.interfaces.argument import ArgumentDoc
 from datetime import datetime
 from jormungandr.resources_utils import ResourceUtc
 from jormungandr.interfaces.v1.transform_id import transform_id
-from flask_restful import reqparse, abort
+from flask_restful import abort
 import logging
 from jormungandr.exceptions import RegionNotFound
 from functools import cmp_to_key
@@ -66,6 +66,15 @@ class DatetimeRepresents(CustomSchemaType):
 
     def schema(self):
         return TypeSchema(type=str, metadata={'enum': ['arrival', 'departure'], 'default': 'departure'})
+
+
+def sort_regions(regions_list):
+    """
+    Sort the regions according to the instances_comparator (priority > is_free=False > is_free=True)
+    :param regions_list: possible regions for the journeys request
+    :return: sorted list according to the comparator
+    """
+    return sorted(regions_list, key=cmp_to_key(instances_comparator))
 
 
 def compute_regions(args):
@@ -112,13 +121,12 @@ def compute_regions(args):
 
     sorted_regions = list(possible_regions)
 
-    regions = sorted(sorted_regions, key=cmp_to_key(instances_comparator))
+    regions = sort_regions(sorted_regions)
 
     return [r.name for r in regions]
 
 
 def compute_possible_region(region, args):
-
     if region:
         region_obj = i_manager.get_region(region)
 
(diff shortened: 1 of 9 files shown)


source/jormungandr/jormungandr/interfaces/v1/test/journey_tests.py
```diff
@@ -193,3 +194,19 @@ def test_multi_coverage_overlap_chose_with_priority(self):
         assert regions[1] == self.regions['netherlands'].name
         assert regions[2] == self.regions['france'].name
         assert regions[3] == self.regions['bolivia'].name
+
+    def test_sorted_regions(self):
+        """
+        Test that the regions are sorted correctly according to the comparator criteria (priority > is_free=False > is_free=True)
+        """
+        # Remove the 2 following instances as there have same configuration as others
```
pbougue:
```suggestion
        # Remove the 2 following instances as they have same configuration as others
```

source/jormungandr/jormungandr/interfaces/v1/journey_common.py
```diff
@@ -68,6 +68,15 @@ def schema(self):
         return TypeSchema(type=str, metadata={'enum': ['arrival', 'departure'], 'default': 'departure'})
 
 
+def sort_regions(regions_list):
+    """
+    Sort the regions according to the instances_comparator (priority > is_free=False > is_free=True)
+    :param regions_list: possible regions for the journeys request
+    :return: sorted list according to the comparator
+    """
+    return sorted(regions_list, key=cmp_to_key(instances_comparator))
```
pbougue, line 77: We may also sort them alphabetically as a last criterion to have stable outputs (thinking of Artemis if we ever want to test it there).

source/jormungandr/jormungandr/tests/instance_manager_tests.py
```diff
@@ -77,3 +81,20 @@ def get_instances_by_object_id_test(manager, mocker):
...
+
+def choose_best_instance_test():
+    """
+    Test to choose the best instance according to comparator : priority > is_free=False > is_free=True
+    """
+    instances_list = [
+        FakeInstance('fr-nw', is_free=True, priority=0),
+        FakeInstance('fr-nw-c', is_free=True, priority=0),
+    ]
+    assert choose_best_instance(instances_list).name == 'fr-nw'
```
pbougue: Not sure: here you test that you keep the same order as the list? Why is it useful to test it?

source/jormungandr/tests/overlapping_routing_tests.py
```diff
@@ -34,6 +34,22 @@
 from .check_utils import *
 
 
+@dataset({"main_routing_test": {'is_free': False}, "empty_routing_test": {'is_free': True}})
+class TestOverlappingBestCoverage(AbstractTestFixture):
+    """
+    Test the answer if 2 coverages are overlapping by choosing the best one according to the instances comparator
```
pbougue: Minor-minor: add a comment on what is the other possible coverage and its output can be nice.

pbougue (approved):
It could be nice to add a precision on priority: highest value means it's chosen first.
I think of 2 places: https://github.com/CanalTP/navitia/tree/dev/source/tyr#instances and https://github.com/CanalTP/navitia/blob/bf108bc82913ce6d276bcce576a4864b0a39d6ff/source/tyr/tyr/resources.py#L445

source/jormungandr/jormungandr/tests/instance_manager_tests.py
mbounabi: Because if you look at the function 'choose_best_instance', that's what it does and it does happen in other tests. There's now a comment in the function description about it. 
We can maybe use the idea of alphabetical order here too.
-> Now tested with alphabetical order filter

source/jormungandr/tests/overlapping_routing_tests.py
mbounabi: That's minor-minor-minor-minor : looking at the PR, the instances comparator is explained several times

source/jormungandr/jormungandr/interfaces/v1/journey_common.py
mbounabi: It may be a good idea indeed. However, it may also change some tests again... TBD with the architect -> :heavy_check_mark:

mbounabi:
@pbougue, there already a comment in the tyr documentation about it, at the end of the chapter you mentionned:

> For instance, if you have many instances on the same geographical area, you may want to set the priority property to define which one should be used first when requesting navitia APIs without coverage ($navitia_uri/v1/journeys, $navitia_uri/v1/coord/, etc)

kinnou02 (approved):

benoit-bst (approved):

PR #2766, opened 2019-04-19
[Kraken] Use Astar for direct path

AurelienLP:
For some optimization purposes, we'd like to use Astar instead of Djikstra for direct path computation. This PR implements the Astar for a future benchmark.


source/georef/street_network.h
```diff
@@ -130,6 +133,9 @@ struct PathFinder {
     /// Distance array for the Dijkstra
     std::vector<navitia::time_duration> distances;
 
+    /// Distance array for the Astar
+    std::vector<navitia::time_duration> costs;
```
kinnou02, line 150: If benchmarking is good I would like to have a dedicated `PathFinder` as this is a big memory overhead

AurelienLP: Yep, and I won't have to override the init() method of PathFinder too.

source/georef/astar_shortest_paths_with_heap.h
```diff
@@ -0,0 +1,73 @@
...
+https://groups.google.com/d/forum/navitia
+www.navitia.io
+*/
+
+#include <boost/graph/astar_search.hpp>
+
+namespace boost {
+// Call breadth first search
+// Allow to pass color map so that user deals with the allocation (and white init)
+template <class Graph,
```
nonifier, line 36: Do you really need all of those templated classes ?!? :fearful: 
Looks like some of them could have default initialisation like: 
```cpp
class Compare =  std::less<navitia::time_duration>
```
Where you can late in the function use ` bfs_vis(..., Compare(), ...);`

Also, a lot of parameters are part of class `PathFinder`, why not decorating this class instead to reduce the amount of parameter ?

source/georef/astar_shortest_paths_with_heap.h
```diff
@@ -0,0 +1,73 @@
...
+Stay tuned using
+twitter @navitia
+IRC #navitia on freenode
+https://groups.google.com/d/forum/navitia
+www.navitia.io
+*/
+
+#include <boost/graph/astar_search.hpp>
+
+namespace boost {
```
nonifier, line 33: It's very confusing to be adding things into the `boost` namespace. When I saw `boost::astar_shortest_paths_no_init_with_heap` my first reaction was to open the boost documentation and to get confused by not finding anything... I would rather have something within `navitia::`.

source/georef/street_network.h
```diff
@@ -192,6 +202,30 @@ struct PathFinder {
...
+     **/
+    template <class Heuristic, class Visitor>
+    void astar(const vertex_t source, const Heuristic& heuristic, const Visitor& visitor) {
+        // Note: the predecessors have been updated in init
+        auto logger = log4cplus::Logger::getInstance("worker");
+        LOG4CPLUS_DEBUG(logger, "astar !!!");
+
+        // Fill color map in white before astar
+        std::fill(color.data.get(),
+                  color.data.get() + (color.n + color.elements_per_char - 1) / color.elements_per_char, 0);
```
nonifier, line 232: :nauseated_face:

source/georef/street_network.h
```diff
@@ -426,5 +460,70 @@ struct printer_distance_or_target_visitor : virtual public printer_distance_visi
 };
 #endif
 
+// Visitor who stops (throw a DestinationFound exception) when all targets has been visited
```
nonifier, line 478:
```suggestion
// Visitor will stop (throw a DestinationFound exception) when all targets have been visited
```

PR #2767, opened 2019-04-26
Fix spelling and change tests

mbounabi:
change "ASCENCEUR" to "ASCENSEUR"

### DIFF
--- a/source/jormungandr/jormungandr/equipments/tests/sytral_equipment_test.py
+++ b/source/jormungandr/jormungandr/equipments/tests/sytral_equipment_test.py
@@ -64,23 +64,23 @@ def equipments_get_information_test():
     """
     Test that 'equipment_details' structure is added to StopPoint proto when conditions are met
     """
-    provider = SytralProvider(url="sytral.url", timeout=3, code_types=["TCL_ASCENCEUR", "TCL_ESCALIER"])
+    provider = SytralProvider(url="sytral.url", timeout=3, code_types=["TCL_ASCENSEUR", "TCL_ESCALIER"])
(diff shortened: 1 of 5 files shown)


source/jormungandr/jormungandr/equipments/tests/sytral_equipment_test.py
```diff
@@ -64,23 +64,23 @@ def equipments_get_information_test():
...
     # equipment_details is not added
-    st = create_stop_point("TCL_ASCENCEUR", "262")
+    st = create_stop_point("TCL_ASCENSEUR", "262")
     provider.get_informations_for_journeys([st])
     assert not st.equipment_details
 
     # stop point has code with incorrect type but value present in webservice response
     # equipment_details is not added
-    st = create_stop_point("ASCENCEUR", "261")
+    st = create_stop_point("TCL_ASCENCEUR", "261")
```
pbougue, line 84: Wasn't `TCL_` removed on purpose, before?
Looks like it from the comment.

pbougue (commented):
The rest is OK.

kinnou02 (approved):

source/jormungandr/jormungandr/equipments/tests/sytral_equipment_test.py
mbounabi: The test case is to have an incorrect code, so now "TCL_ASCENCEUR" is incorrect ;)

pbougue: Aaaaaah :stuck_out_tongue_closed_eyes:
OK, nice!

pbougue (approved):

PR #2769, opened 2019-04-30
Release process hardenning

nonifier:
This aims at making sure that `pre-commit` is installed before starting the release process. 

The release is now be done via the script `release_navitia.sh` that acts as a wrapper around `script_release.py` and manages the dependencies via Pipenv.


pbougue (approved):
:+1:
Not sure about the `release` directory, but no brilliant naming comes to mind (and it's a bad practice to compile into repo, so...). Maybe we should even remove the `debug` directory from gitignore for consistency.

kinnou02 (approved):

PR #2772, opened 2019-05-03
Build: use ccache and gold if available

kinnou02:
[Gold](https://en.wikipedia.org/wiki/Gold_(linker)) is an alternative linker part of the gnu project. It's faster than
ld, [around three time faster](https://lld.llvm.org/#performance).

ccache is a compiler cache that speedup rebuild.


nonifier (approved):
What"s your average gain with Gold/CCache ?

kinnou02:
I was already using `ccache` with carefully crafted simlink, so I can't really time it.
I will time `gold` today

kinnou02:
`ld` link all executable in 210s while `gold` do it in 55s

pbougue (approved):
Tried LLD on clang, but difficult to make something mutualized https://github.com/kinnou02/navitia/pull/3.
Fine for ccache and gold :+1:

PR #2773, opened 2019-05-03
[Tyr ] Add job 'cities' and returns it in /cities/status

mbounabi:
First part of https://jira.kisio.org/browse/NAVP-1093
A job and a dataset 'cities' are now created in Tyr db.

The endpoint /cities/status now returns:
```
{
    "cities db version": "1d99ba678e3f",
    "latest_job": {
        "created_at": "2019-05-03T09:43:55.452338",
        "data_sets": [
            {
                "family_type": "cities",
                "name": "/home/mehdi/second_fork/navitia/source/tyr/france_boundaries.osm.pbf",
                "type": "cities"
            }
        ],
        "id": 8,
        "state": "done",
        "updated_at": "2019-05-03T09:44:32.344140"
    }
}
```

The 'cities' purge task will be done during the already existing 'purge' task called by the cron 'purge-old-job-every-week'.
The number of jobs and datasets to keep is determined by the already existing conf parameter 'DATASET_MAX_BACKUPS_TO_KEEP'

### DIFF
--- a/source/tyr/tyr/resources.py
+++ b/source/tyr/tyr/resources.py
@@ -60,7 +60,6 @@ from tyr.tasks import (
     remove_autocomplete_depot,
     import_autocomplete,
     cities,
-    cosmogony2cities,
     COSMOGONY_REGEXP,
 )
 from tyr.helper import get_instance_logger, save_in_tmp
@@ -1746,6 +1745,7 @@ class Cities(flask_restful.Resource):
         file_name = f.filename
         file_path = str(os.path.join(os.path.abspath(current_app.config['CITIES_OSM_FILE_PATH']), file_name))
         f.save(file_path)
+        logging.info("file received: {}".format(f))
 
         # Create Job and Dataset in db to track progress
         job = models.Job()
@@ -1753,20 +1753,19 @@ class Cities(flask_restful.Resource):
         models.db.session.add(job)
         dataset = models.DataSet()
         dataset.name = file_path
-        dataset.type = 'cities'
         dataset.family_type = 'cities'
         models.db.session.add(dataset)
         job.data_sets.append(dataset)
-        models.db.session.commit()
-
-        logging.info("file: {}".format(f))
 
         if COSMOGONY_REGEXP.match(file_name):
             # it's a cosmogony file, we import it with cosmogony2cities
-            cosmogony2cities.delay(file_path, job.id)
+            dataset.type = exe = 'cosmogony2cities'
         else:
             # we import it the 'old' way, with cities
-            cities.delay(file_path, job.id)
+            dataset.type = exe = 'cities'
+
+        models.db.session.commit()
+        cities.delay(file_path, job.id, exe)
 
         return {'message': 'OK'}, 200
 
--- a/source/tyr/tyr/resources.py
+++ b/source/tyr/tyr/resources.py
@@ -1739,7 +1739,7 @@ class Cities(flask_restful.Resource):
         args = parser.parse_args()
 
         if not args['file']:
-            logging.info("No file provided")
+            logging.error("No file provided")
             return {'message': 'No file provided'}, 400
 
         f = args['file']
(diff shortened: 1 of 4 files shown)


source/tyr/tyr/resources.py
```diff
@@ -1691,43 +1694,78 @@ def check_db():
...
         if COSMOGONY_REGEXP.match(file_name):
             # it's a cosmogony file, we import it with cosmogony2cities
-            cosmogony2cities.delay(file_path)
+            dataset.type = exe = 'cosmogony2cities'
         else:
             # we import it the 'old' way, with cities
-            cities.delay(file_path)
+            dataset.type = exe = 'cities'
+
+        models.db.session.commit()
```
xlqian, line 1767: :+1:

source/tyr/tyr/resources.py
```diff
@@ -1691,43 +1694,78 @@ def check_db():
...
+        cities.delay(file_path, job.id, exe)
```
xlqian: can you make a try catch? because `cities.delay` may throw exceptions(ex. rabbitmq is down)
```python
try:
    cities.delay(file_path, job.id, exe)
except ... :
    job.state = 'fail'
    models.db.session.commit()

```

source/tyr/tyr/tasks.py
```diff
@@ -475,38 +475,27 @@ def load_data(instance_id, data_dirs):
...
         if res != 0:
-            logging.error('cosmogony2cities failed')
+            job.state = 'failed'
+            logging.error('{} failed'.format(exe))
+        else:
+            job.state = 'done'
+
     except Exception as e:
-        logging.exception('cosmogony2cities exception : {}'.format(e.message))
+        logging.exception('{} exception : {}'.format(exe, e.message))
```
xlqian, line 525: you need job.state = 'failed' here too

source/tyr/tyr/resources.py
```diff
@@ -1691,43 +1694,82 @@ def check_db():
...
-            return {'message': 'cities db alembic version = {}'.format(msg)}, 200
+        cities_version = check_cities_db()
+        if cities_version:
+            response['cities db version'] = '{}'.format(cities_version)
         else:
             return {'message': 'cities db not reachable'}, 404
 
+        cities_job = check_cities_job()
+        if cities_job and 'instance' in cities_job:
+            # No instance associated to 'cities' job, remove the item
```
pbougue, line 1724: I don't understand this.
I'm not used to tyr though, but it may be worth explaining a bit more (or rewrite comment?) if I'm not the only one.

source/tyr/tyr/resources.py
```diff
@@ -1691,43 +1694,82 @@ def check_db():
...
-            cosmogony2cities.delay(file_path)
+            dataset.type = exe = 'cosmogony2cities'
         else:
             # we import it the 'old' way, with cities
-            cities.delay(file_path)
+            dataset.type = exe = 'cities'
+
+        models.db.session.commit()
+        try:
+            cities.delay(file_path, job.id, exe)
```
pbougue, line 1769: minor: for code "cleanness" (it's just my opinion) you could use named param `exe=dataset.type` and avoid `dataset.type = exe = 'cities'`

pbougue (commented):
This looks good and concise (not an expert of that area though).
Do we need some kind of purge of those jobs in db?

xlqian (approved):

source/tyr/tyr/resources.py
mbounabi: Let's wait to see if you're not the only one...

source/tyr/tyr/resources.py
mbounabi: same: Let's wait to see if you're not the only one... Or a more rational explanation about performance

mbounabi:
@pbougue It does need a purge somewhere, indeed. I will add it in a future commit of this PR

source/tyr/tyr/resources.py
mbounabi: The explanation on this one is rather long and not really useful for understanding the cities status.

source/tyr/tyr/tasks.py
```diff
@@ -475,38 +505,28 @@ def load_data(instance_id, data_dirs):
...
-            logging.error('cosmogony2cities failed')
+            job.state = 'failed'
+            logging.error('{} failed'.format(exe))
+        else:
+            job.state = 'done'
+
     except Exception as e:
-        logging.exception('cosmogony2cities exception : {}'.format(e.message))
+        job.state = 'failed'
+        logging.exception('{} exception : {}'.format(exe, e.message))
```
kinnou02, line 525: you need a `raise` or a `return` here

mbounabi: and a commit. Right!

nonifier (approved):

benoit-bst (approved):

kinnou02 (approved):

PR #2775, opened 2019-05-03
jormungandr: add test on trip add with trip existing in navitia

kadhikari:
Ticket: https://jira.kisio.org/browse/NAVP-1202


benoit-bst (approved):

pbougue (approved):
OK, it's clear from comment that it's not the behavior we want, but it's soon gonna be changed :+1:

pbougue (commented):
Suggestion (maybe more for the next ticket): test also on the double RT trip-add that the second one is rejected.

PR #2776, opened 2019-05-06
kraken: Realtime, cannot add new trip if trip-id corresponds to a base VJ the same day

benoit-bst:
**JIRA** : https://jira.kisio.org/browse/NAVP-1298

### DIFF
--- a/source/jormungandr/tests/kirin_realtime_tests.py
+++ b/source/jormungandr/tests/kirin_realtime_tests.py
@@ -2148,6 +2148,54 @@ class TestKirinAddNewTripPresentInNavitia(MockKirinDisruptionsFixture):
         assert nb_disruptions_before == len(disruptions_after['disruptions'])
 
 
+@dataset(MAIN_ROUTING_TEST_SETTING)
+class TestKirinAddNewTripPresentInNavitiaWithAShift(MockKirinDisruptionsFixture):
+    def test_add_new_trip_present_in_navitia_with_a_shift(self):
+
+        disruption_query = 'disruptions?_current_datetime={dt}'.format(dt='20120614T080000')
+        disruptions_before = self.query_region(disruption_query)
+        nb_disruptions_before = len(disruptions_before['disruptions'])
+
+        # The vehicle_journey vjA is present in navitia
+        pt_response = self.query_region('vehicle_journeys/vjA?_current_datetime=20120614T1337')
+
+        assert len(pt_response['vehicle_journeys']) == 1
+        assert len(pt_response['disruptions']) == 0
+
+        # New disruption, a new trip with vehicle_journey id = vjA and having 2 stop_times in realtime
+        self.send_mock(
+            "vjA",
+            "20120620",
+            'added',
+            [
+                UpdatedStopTime(
+                    "stop_point:stopC",
+                    arrival_delay=0,
+                    departure_delay=0,
+                    is_added=True,
+                    arrival=tstamp("20120620T080100"),
+                    departure=tstamp("20120620T080100"),
+                    message='on time',
+                ),
+                UpdatedStopTime(
+                    "stop_point:stopB",
+                    arrival_delay=0,
+                    departure_delay=0,
+                    is_added=True,
+                    arrival=tstamp("20120620T080102"),
+                    departure=tstamp("20120620T080102"),
+                ),
+            ],
+            disruption_id='new_trip',
+            effect='additional_service',
+        )
+
+        # Check that there should not be a new disruption
+        disruptions_after = self.query_region(disruption_query)
+        assert nb_disruptions_before == len(disruptions_after['disruptions'])
+
+
+
 def make_mock_kirin_item(
     vj_id,
     date,
--- a/source/jormungandr/tests/kirin_realtime_tests.py
+++ b/source/jormungandr/tests/kirin_realtime_tests.py
@@ -2143,11 +2143,9 @@ class TestKirinAddNewTripPresentInNavitia(MockKirinDisruptionsFixture):
             effect='additional_service',
         )
 
-        # TODO: corrections to be made in kraken to refuse the trip add and update the test below
-        # Ticket NAVP-1298
         # Check that there should not be a new disruption
         disruptions_after = self.query_region(disruption_query)
-        assert nb_disruptions_before + 1 == len(disruptions_after['disruptions'])
+        assert nb_disruptions_before == len(disruptions_after['disruptions'])
 
 
 def make_mock_kirin_item(
--- a/source/jormungandr/tests/kirin_realtime_tests.py
+++ b/source/jormungandr/tests/kirin_realtime_tests.py
@@ -2157,12 +2157,12 @@ class TestKirinAddNewTripPresentInNavitiaWithAShift(MockKirinDisruptionsFixture)
         nb_disruptions_before = len(disruptions_before['disruptions'])
 
         # The vehicle_journey vjA is present in navitia
-        pt_response = self.query_region('vehicle_journeys/vjA?_current_datetime=20120614T1337')
+        pt_response = self.query_region('vehicle_journeys/vehicle_journey:vjA?_current_datetime=20120614T1337')
 
         assert len(pt_response['vehicle_journeys']) == 1
         assert len(pt_response['disruptions']) == 0
 
-        # New disruption, a new trip with vehicle_journey id = vjA and having 2 stop_times in realtime
+        # New disruption, a new trip with meta vehicle journey id = vjA and having 2 stop_times in realtime
         self.send_mock(
             "vjA",
             "20120620",
--- a/source/jormungandr/tests/kirin_realtime_tests.py
+++ b/source/jormungandr/tests/kirin_realtime_tests.py
@@ -2190,10 +2190,10 @@ class TestKirinAddNewTripPresentInNavitiaWithAShift(MockKirinDisruptionsFixture)
             effect='additional_service',
         )
 
-        # Check that there should not be a new disruption
+        # The new trip is accepted because, it is not the same day of the base vj
+        # So a disruption is added
         disruptions_after = self.query_region(disruption_query)
-        assert nb_disruptions_before == len(disruptions_after['disruptions'])
-
+        assert nb_disruptions_before + 1 == len(disruptions_after['disruptions'])
 
 
 def make_mock_kirin_item(
--- a/source/jormungandr/tests/kirin_realtime_tests.py
+++ b/source/jormungandr/tests/kirin_realtime_tests.py
@@ -2102,8 +2102,8 @@ class TestKirinAddNewTripBlocked(MockKirinDisruptionsFixture):
 
 
 @dataset(MAIN_ROUTING_TEST_SETTING)
-class TestKirinAddNewTripPresentInNavitia(MockKirinDisruptionsFixture):
-    def test_add_new_trip_present_in_navitia(self):
+class TestKirinAddNewTripPresentInNavitiaTheSameDay(MockKirinDisruptionsFixture):
+    def test_add_new_trip_present_in_navitia_the_same_day(self):
 
         disruption_query = 'disruptions?_current_datetime={dt}'.format(dt='20120614T080000')
         disruptions_before = self.query_region(disruption_query)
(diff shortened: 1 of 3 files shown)


source/kraken/tests/realtime_test.cpp
```diff
@@ -2799,7 +2799,7 @@ BOOST_FIXTURE_TEST_CASE(add_and_update_trip_to_verify_route_line_commercial_mode
                                                    RTStopTime("stop_point:F", "20190101T0900"_pts).skipped(),
                                                    RTStopTime("stop_point:J", "20190101T0900"_pts).added(),
                                                },
-                                               transit_realtime::Alert_Effect::Alert_Effect_ADDITIONAL_SERVICE,
+                                               transit_realtime::Alert_Effect::Alert_Effect_DETOUR,
```
benoit-bst: In this test, the status should not have been `ADDITIONAL_SERVICE`

benoit-bst:
Jenkins, test this please !

source/kraken/tests/realtime_test.cpp
```diff
@@ -3000,18 +2998,18 @@ BOOST_FIXTURE_TEST_CASE(add_new_trip_and_update, AddTripDataset) {
     BOOST_REQUIRE_EQUAL(pt_data.vehicle_journeys_map.size(), 2);
     BOOST_REQUIRE_EQUAL(pt_data.vehicle_journeys.size(), 2);
 
-    // New trip added
+    // Correspond to the first added trip
```
pbougue:
```suggestion
    // Corresponds to the first added trip
```

source/kraken/tests/realtime_test.cpp
```diff
@@ -2793,14 +2793,14 @@ BOOST_FIXTURE_TEST_CASE(add_and_update_trip_to_verify_route_line_commercial_mode
...
-                                               transit_realtime::Alert_Effect::Alert_Effect_ADDITIONAL_SERVICE,
-                                               comp_uri, phy_mode_uri);
+    new_trip =
+        ntest::make_trip_update_message("vj_new_trip", "20190101",
+                                        {
+                                            RTStopTime("stop_point:A", "20190101T0800"_pts).added(),
+                                            RTStopTime("stop_point:F", "20190101T0900"_pts).skipped(),
+                                            RTStopTime("stop_point:J", "20190101T0900"_pts).added(),
+                                        },
+                                        transit_realtime::Alert_Effect::Alert_Effect_DETOUR, comp_uri, phy_mode_uri);
```
pbougue: Not sure that's how Kirin behaves (Artemis will now)... And I think this would break the output of the trip-status in /disruptions

source/kraken/realtime.cpp
```diff
@@ -646,6 +646,12 @@ void handle_realtime(const std::string& id,
         }
         return;
     }
+    if (meta_vj_exists && is_added_trip(trip_update)) {
```
pbougue: From afar, I would "only" test if there is a base VJ.

source/kraken/tests/realtime_test.cpp
```diff
@@ -2966,9 +2966,7 @@ BOOST_FIXTURE_TEST_CASE(add_new_trip_and_update, AddTripDataset) {
                       pbnavitia::StopTimeUpdateStatus::ADDED);
 
     // Recall the same new trip.
-    // The old "new trip" is replace by the new, but the meta VJ is the same (don't create).
-    // The old impact is deleted too
-    // To show that it is the new GTFS-RT, schedules has changed
+    // the new trip is rejected. There is no update
```
pbougue: You changed the spirit of this test (`add_new_trip_and_update`), no?

pbougue (commented):
This needs to roll under artemis' tests I think

source/kraken/tests/realtime_test.cpp
benoit-bst: Nope, it is a overall test for an added trip. But if you prefer, I can extract this part, to create a test :`add_new_trip_with_id_that_already_exists`
It'll be more comfortable to read

source/kraken/realtime.cpp
```diff
@@ -646,6 +646,12 @@ void handle_realtime(const std::string& id,
...
+        LOG4CPLUS_WARN(log, "Cannot add twice a new trip"
```
kinnou02:
```suggestion
        LOG4CPLUS_WARN(log, "Cannot add twice a new trip, rejecting this trip"
```
always be clear on what action as be taken in log

source/kraken/tests/realtime_test.cpp
benoit-bst: This test not concern `Additional Trip`. This is the reason why I have changed it.
If I understand correctly, we want `Update the trip recently added with destination F replaced by J`. It seems to be a `Detour`

source/kraken/realtime.cpp
benoit-bst: `Meta VJ` is closely linked to base VJ. 
If you don't have the meta VJ with the good id, you necessarily have the base VJ.
We can only check Meta VJ, the test work well.

source/kraken/realtime.cpp
benoit-bst: better : `cannot add a new trip with id that already exists`

kadhikari (approved):
Apart from the remarks already present, it seems perfect for me.

source/kraken/tests/realtime_test.cpp
pbougue: Kirin keeps the memory of an added trip, and always sends it with effect=ADDITIONAL_SERVICE: https://github.com/CanalTP/kirin/blob/master/tests/integration/cots_test.py#L1074
This is how navitia keeps the effect of the trip as ADDED: erasing previous RT, trusting Kirin.

That's why I think that:
* navitia should(can) only block RT trip-add over base-schedule VJ
* existing tests shouldn't need to be updated by this PR (except of course the one designed for it in kirin_realtime_tests.py)
* we should add a test (integration if possible) on effect after an add+delay sequence

pbougue (dismissed):
After verification, this would not work with Kirin

source/kraken/tests/realtime_test.cpp
benoit-bst: Okay I understood you remark.

We just want to reject an added trip if the base-schedule VJ is empty, because it corresponds to a real time VJ. It is the only way to distinguish a real time VJ already added.
So, I'm agree with you, `existing tests shouldn't need to be updated by this PR (except of course the one designed for it in kirin_realtime_tests.py)`

benoit-bst:
There was some confusions of my part, about the scope of the PR.
After discussion with @pbougue, it turns out that, we only want to `reject a new trip if its trip-id corresponds to a base VJ`

I updated the PR, with a Utest that highlight the good behavior.

benoit-bst:
To sum up for you @pbougue 
Now, to reject a new trip we have 
`if (meta_vj_exists && is_added_trip(trip_update) && base_vj_exists_the_same_day(data, trip_update)) `
We take into account the date to reject. 
You've right it works better.
A Kirin integration test highlight the behavior.

nonifier (approved):

mbounabi (approved):
Looks good! :eye:

source/jormungandr/tests/kirin_realtime_tests.py
```diff
@@ -2132,10 +2132,56 @@ def test_add_new_trip_present_in_navitia(self):
...
+@dataset(MAIN_ROUTING_TEST_SETTING)
+class TestKirinAddNewTripPresentInNavitiaWithAShift(MockKirinDisruptionsFixture):
+    def test_add_new_trip_present_in_navitia_with_a_shift(self):
+
+        disruption_query = 'disruptions?_current_datetime={dt}'.format(dt='20120614T080000')
+        disruptions_before = self.query_region(disruption_query)
+        nb_disruptions_before = len(disruptions_before['disruptions'])
+
+        # The vehicle_journey vjA is present in navitia
+        pt_response = self.query_region('vehicle_journeys/vjA?_current_datetime=20120614T1337')
```
pbougue: The name `vjA` is probably now `vehicle_journey:vjA` (after https://github.com/CanalTP/navitia/pull/2779).
You will need to merge/rebase on `dev` branch.

source/jormungandr/tests/kirin_realtime_tests.py
```diff
@@ -2132,10 +2132,56 @@ def test_add_new_trip_present_in_navitia(self):
...
+
+        assert len(pt_response['vehicle_journeys']) == 1
+        assert len(pt_response['disruptions']) == 0
+
+        # New disruption, a new trip with vehicle_journey id = vjA and having 2 stop_times in realtime
```
pbougue: minor: actually here it's the `MVJ` id (so it's still "vjA" after the #2779 I think).

source/kraken/realtime.cpp
```diff
@@ -46,6 +46,22 @@ namespace navitia {
 
 namespace nd = type::disruption;
 
+static bool base_vj_exists_the_same_day(const type::Data& data, const transit_realtime::TripUpdate& trip_update) {
+    const auto& mvj = *data.pt_data->get_or_create_meta_vehicle_journey(trip_update.trip().trip_id(),
```
pbougue, line 50: I think we shouldn't use `get_or_create_meta_vehicle_journey` directly here.
It's currently protected before the call to this method by testing `meta_vj_exists` but if it is re-used without thinking of it, it would create side-effect (create new VJ if not existing).

So I'd say at least port here the test on MVJ's existence (and so remove the one before calling this method):
```cpp
    if (*data.pt_data->meta_vjs.exists(trip_update.trip().trip_id()) {
        return false;
    }
```

Although a method to get an "option" would be nicer to have/use, too.

source/kraken/realtime.cpp
```diff
@@ -46,6 +46,22 @@ namespace navitia {
...
+                                                                        data.pt_data->get_main_timezone());
+    if (mvj.get_base_vj().empty()) {
```
pbougue: You can probably simplify code below using https://github.com/CanalTP/navitia/blob/d68c8536cd654c7777dff02ec9b34694ed3b50d8/source/type/type.cpp#L296

pbougue (commented):
The spirit of the code feels fine :+1: 

Kraken's use of `trip_update.trip().start_date()` looks ambiguous in case of a day shift in RT (both in past or future), but it will probably be determined later (and I hope we would think of updating your code if needed).

benoit-bst:
Jenkins, test this please !

benoit-bst:
Jenkins, test this please !

benoit-bst:
Jenkins, test this please !

source/kraken/realtime.cpp
benoit-bst: I understand your remark, but I already check if the meta vj exists (https://github.com/CanalTP/navitia/pull/2776/files#diff-a1c1532aa9f44bfd3b8a77e208515e23R658). So, I can call directly `get_or_create_meta_vehicle_journey`.
I save the `meta_vjs.exits(...)` call, because it scans the all meta_vj list.

benoit-bst:
Rebase with dev + simplify the `base_vj_exists_the_same_day()` method.

source/kraken/realtime.cpp
pbougue: This `exists(uri)` is a `map.find()` (and even if it was a vector loop), to me the cost is not an argument good enough to keep a function that has side effect on a data argument passed as `const` (even if it's currently protected now as I previously noted, yes).
So let's use a third opinion on that to move on.

source/kraken/realtime.cpp
```diff
@@ -46,6 +46,17 @@ namespace navitia {
 
 namespace nd = type::disruption;
 
+static bool base_vj_exists_the_same_day(const type::Data& data, const transit_realtime::TripUpdate& trip_update) {
+    const auto& mvj = *data.pt_data->get_or_create_meta_vehicle_journey(trip_update.trip().trip_id(),
+                                                                        data.pt_data->get_main_timezone());
+    if (mvj.get_base_vj_circulating_at_date(
```
pbougue, line 52: :+1:
minor personal opinion: I prefer `!= nullptr`

pbougue (commented):
OK, last thing is about that `get_or_create` : need an external opinion.

PR #2778, opened 2019-05-06
Kraken : Using Astar for direct path

AurelienLP:
This PR aims to :

- [ ] Creating a class PathFinder.
- [ ] Moving all the code related to the old PathFinder structure from street_network.h and street_network.cpp to path_finder.h and path_finder.cpp.
- [ ] Creating two classes DijkstraPathFinder and AstarPathFinder inherited from PathFinder which is abstract.
- [ ] Gathering the visitors for the research in visitor.h and using proper aliases.
- [ ] Fighting against huge templates.


source/georef/dijkstra_path_finder.cpp
```diff
@@ -0,0 +1,358 @@
...
+
+    return nearest_edge.first;
+}
+
+/**
+  The _DEBUG_DIJKSTRA_QUANTUM_ activate at compil time some dump used in quantum to analyze
+  the street network.
+  WARNING, it will slow A LOT the djisktra and must be used only for debug
+  */
+#ifdef _DEBUG_DIJKSTRA_QUANTUM_
```
kinnou02: does it still build? if not we may want to delete it

source/georef/astar_path_finder.cpp
```diff
@@ -0,0 +1,100 @@
...
+ * Warning, it modifies the distances and the predecessors
+ * Warning 2, the target node is NOT the destination !!!!
+ **/
+void AstarPathFinder::astar(const vertex_t source,
+                            const vertex_t target,
+                            const astar_distance_heuristic& heuristic,
+                            const astar_distance_or_target_visitor& visitor) {
+    // Note: the predecessors have been updated in init
+
+    // Fill color map in white before dijkstra
```
kinnou02:
```suggestion
    // Fill color map in white before astar
```

kinnou02 (commented):
It would be nice to have some test on the astar

source/georef/dijkstra_path_finder.cpp
AurelienLP: No idea, but I am not against removing it

kinnou02:
Jenkins, test this please.

kinnou02 (approved):

AurelienLP:
Jenkins, test this please.

AurelienLP:
Jenkins, test this please.

benoit-bst (approved):

source/georef/astar_path_finder.cpp
```diff
@@ -72,7 +73,7 @@ void AstarPathFinder::astar(const vertex_t source,
     auto weight_map = boost::get(&Edge::duration, geo_ref.graph);
     auto combiner = SpeedDistanceCombiner(speed_factor);
 
-    astar_shortest_paths_no_init_with_heap(g, &source, &target, heuristic, visitor, weight_map, combiner);
+    astar_shortest_paths_no_init_with_heap(g, vertex.cbegin(), vertex.cend(), heuristic, visitor, weight_map, combiner);
```
kinnou02: A comment explaining what happenned would help later

kinnou02 (approved):

source/georef/astar_path_finder.cpp
```diff
@@ -0,0 +1,101 @@
...
+              astar_distance_heuristic(geo_ref.graph, destinations.back(), 1. / (default_speed[mode] * speed_factor)),
+              astar_distance_or_target_visitor(radius, distances, destinations));
+    } catch (DestinationFound&) {
+    }
+}
+
+/**
+ * Launch an astar without initializing the data structure
+ * Warning, it modifies the distances and the predecessors
+ * Warning 2, the target node is NOT the destination !!!!
```
nonifier: If not the destination, what it is then ? :)

source/georef/astar_path_finder.cpp
```diff
@@ -0,0 +1,101 @@
...
+ **/
+void AstarPathFinder::astar(const vertex_t source,
+                            const vertex_t target,
+                            const astar_distance_heuristic& heuristic,
+                            const astar_distance_or_target_visitor& visitor) {
+    // Note: the predecessors have been updated in init
+    std::array<georef::vertex_t, 2> vertex{{source, target}};
+
+    // Fill color map in white before dijkstra
```
nonifier:
```suggestion
    // Fill color map in white before A*
```

source/georef/astar_path_finder.h
```diff
@@ -0,0 +1,104 @@
...
+
+struct astar_distance_heuristic : public boost::astar_heuristic<Graph, navitia::seconds> {
+    const Graph& g;
+    const type::GeographicalCoord& dest_coord;
+    const double inv_speed;
+
+    astar_distance_heuristic(const Graph& graph, const vertex_t& destination, const double inv_speed)
+        : g(graph), dest_coord(graph[destination].coord), inv_speed(inv_speed) {}
+
+    navitia::seconds operator()(vertex_t v) {
```
nonifier: you might want to pass a `const vertex_t&` here ?

source/georef/dijkstra_shortest_paths_with_heap.h
```diff
@@ -30,7 +30,7 @@ www.navitia.io
 
 #include <boost/graph/dijkstra_shortest_paths.hpp>
 
-namespace boost {
+namespace navitia {
```
nonifier, line 33: :heart_decoration:

source/georef/path_finder.cpp
```diff
@@ -0,0 +1,517 @@
...
+    }
+    if (coords_to_add.empty()) {
+        coords_to_add.push_back(projection.projected);
+    }
+
+    auto duration = crow_fly_duration(projection.distances[d]);
+    // we need to update the total length
+    p.duration += duration;
+
+    // we aither add the starting coordinate to the first path item or create a new path item if it was another way
```
nonifier:
```suggestion
    // we either add the starting coordinate to the first path item or create a new path item if it was another way
```

source/georef/path_finder.cpp
```diff
@@ -0,0 +1,517 @@
...
+}
+
+void PathFinder::add_custom_projections_to_path(Path& p,
+                                                bool append_to_begin,
+                                                const ProjectionData& projection,
+                                                ProjectionData::Direction d) const {
+    auto item_to_update = [append_to_begin](Path& p) -> PathItem& {
+        return (append_to_begin ? p.path_items.front() : p.path_items.back());
+    };
+    auto add_in_path = [append_to_begin](Path& p, const PathItem& item) {
```
nonifier, line 213: I don't know if it would make things simpler by using `std::front_inserter` or `std::back_inserter` with something like: 
```cpp
auto path_inserter = append_to_begin? std::front_inserter(p.path_tiems) : std::back_inserter(p.path_tiems);
[...]
path_inserter = item;
```

In a similar way, you could have to update your items:
```cpp
auto item_to_update = append_to_begin ? 
                     [&](){ return p.path_items.front(); } : 
                     [&](){ return p.path_items.back(); } ;
[...]
item_to_update().way_id 
```

Those are just ideas to make things nicer, feel free to disregard.
#ModernizeMyCpp

source/georef/path_finder.cpp
```diff
@@ -0,0 +1,517 @@
...
+        if (!p.path_items.empty() && item_to_update(p).way_idx == type::invalid_idx) {
+            item_to_update(p).way_idx = start_edge.way_idx;
+            item_to_update(p).duration += duration;
+        } else {
+            PathItem item;
+            item.way_idx = start_edge.way_idx;
+            item.duration = duration;
+
+            if (!p.path_items.empty()) {
+                // still complexifying stuff... TODO: simplify this
```
nonifier, line 256: Maybe start with extracting the switch case into a separate function :)
And if you're not familiar with the SOLID principles, have a look at the first one - https://en.wikipedia.org/wiki/SOLID
TL;DR -
- a class has one responsibility
- a function does one thing

nonifier (approved):

source/georef/astar_path_finder.h
xlqian: and can it be a const operator?

source/georef/dijkstra_path_finder.cpp
```diff
@@ -0,0 +1,261 @@
...
+                    distance_to_entry_point[sp_idx] = duration;
+                }
+            }
+        }
+    }
+    // case 2 : start coord is an edge (dijkstra)
+    else {
+        std::vector<routing::SpIdx> dest_sp_idx;
+        for (const auto& e : elements) {
+            dest_sp_idx.push_back(routing::SpIdx{e.first});
```
xlqian: you can use emplace_back

source/georef/path_finder.cpp
AurelienLP: Is `get_transportation_mode_item_to_update` for the function containing the switch good enough for you ?

source/georef/dijkstra_path_finder.cpp
```diff
@@ -169,7 +169,7 @@ routing::map_stop_point_duration DijkstraPathFinder::find_nearest_stop_points(
     else {
         std::vector<routing::SpIdx> dest_sp_idx;
         for (const auto& e : elements) {
-            dest_sp_idx.push_back(routing::SpIdx{e.first});
+            dest_sp_idx.emplace_back(routing::SpIdx{e.first});
```
benoit-bst, line 172: :+1:

nonifier (approved):

benoit-bst (approved):

source/georef/dijkstra_shortest_paths_with_heap.h
pbougue: I think we left that namespace because the goal was to make a PR to boost with that.

source/georef/path_finder.h
```diff
@@ -0,0 +1,201 @@
...
+    virtual ~PathFinder() = 0;
+
+    /**
+     *  Update the structure for a given starting point and transportation mode
+     *  The init HAS to be called before any other methods
+     */
+    void init(const type::GeographicalCoord& start_coord, nt::Mode_e mode, const float speed_factor);
+
+    // return the path from the starting point to the target. the target has to have been previously visited.
+    Path get_path(type::idx_t idx);
```
pbougue, line 157: This method is only used in Astar, so it should probably move there.
Also not fan of the naming as it "shadows" the other `get_path`.

source/georef/path_finder.h
```diff
@@ -0,0 +1,201 @@
...
+
+    // Add the starting point projection the the path. Add a new way if needed
+    void add_projections_to_path(Path& p, bool append_to_begin) const;
+
+    // shouldn't be used outside of class apart from tests
+    Path get_path(const ProjectionData& target,
+                  const std::pair<navitia::time_duration, ProjectionData::Direction>& nearest_edge);
+
+    // find the nearest vertex from the projection. return the distance to this vertex and the vertex
+    std::pair<navitia::time_duration, ProjectionData::Direction> find_nearest_vertex(const ProjectionData& target,
```
pbougue, line 167: This is currently the only method that needs to be public because it's used outside of tests or child classes.

Depending on the use of it, it feels like composition could be a clearer option than inheritance, as the children are not interchangeable (maybe they should be, and father class would define an interface then?)

source/georef/path_finder.h
```diff
@@ -0,0 +1,201 @@
...
+    void init(const type::GeographicalCoord& start_coord, nt::Mode_e mode, const float speed_factor);
+
+    // return the path from the starting point to the target. the target has to have been previously visited.
+    Path get_path(type::idx_t idx);
+
+    // Add the starting point projection the the path. Add a new way if needed
+    void add_projections_to_path(Path& p, bool append_to_begin) const;
+
+    // shouldn't be used outside of class apart from tests
+    Path get_path(const ProjectionData& target,
```
pbougue, line 163: This is used in tests, but we should probably move it to protected section (and friend/inherit for tests).

source/georef/path_finder.h
```diff
@@ -0,0 +1,201 @@
...
+    void add_custom_projections_to_path(Path& p,
+                                        bool append_to_begin,
+                                        const ProjectionData& projection,
+                                        ProjectionData::Direction d) const;
+
+    // Build a path with a destination and the predecessors list
+    Path build_path(vertex_t best_destination) const;
+};
+
+bool is_projected_on_same_edge(const ProjectionData& p1, const ProjectionData& p2);
```
pbougue, line 193: This could stay static to avoid unnecessary inclusion.

source/georef/astar_path_finder.cpp
```diff
@@ -0,0 +1,99 @@
...
+void AstarPathFinder::astar(const std::array<georef::vertex_t, 2>& origin_vertexes,
+                            const astar_distance_heuristic& heuristic,
+                            const astar_distance_or_target_visitor& visitor) {
+    // Note: the predecessors have been updated in init
+
+    // Fill color map in white before A*
+    std::fill(color.data.get(), color.data.get() + (color.n + color.elements_per_char - 1) / color.elements_per_char,
+              0);
+
+    // we filter the graph to only use certain mean of transport
```
pbougue, line 67: Not sure: Can't you do that at init? To avoid multiple filtering calls?

pbougue (commented):
Mostly for memory: some of these remarks are not introduced by this PR, and I did not dig much into it with all that code moving :dizzy_face: :smile: 
But in case :truck:-team wants to cleanup, here are some ideas.

The point on filtering might improve perf, though.

PR #1258, opened 2015-11-06
type: MVJ::create_{discrete,frequency}_vj with stop time list given

TeXitoi:
Still not done:
- cancel the vj that are updated by this new vj;
- handle shift (if the first stop time is greater than 24h).

PR #2779, opened 2019-05-06
Kraken: improve RT vj name and uri

pbougue:
:mag_right: review by commit advised (read commit message)

Improvements:
1. When creating a VJ from an existing one, reuse its name (used as headsign too).
    Previously it was created like the uri (assembling mvj, disruption...), it's only fallback now.
2. RT VJ uri should now always have "vehicle_journey:" upfront (like base_schedule VJ)


source/kraken/apply_disruption.cpp
```diff
@@ -66,7 +66,8 @@ static nt::VehicleJourney* create_vj_from_old_vj(nt::MetaVehicleJourney* mvj,
     auto odt_message = vj->odt_message;
     auto vehicle_properties = vj->_vehicle_properties;
 
-    auto* new_vj = mvj->create_discrete_vj(new_vj_uri, rt_level, new_vp, vj->route, std::move(new_stop_times), pt_data);
+    auto* new_vj =
+        mvj->create_discrete_vj(new_vj_uri, vj->name, rt_level, new_vp, vj->route, std::move(new_stop_times), pt_data);
```
pbougue, line 70: :dart: Main change on name is here: reuse "old" vj->name.

source/kraken/apply_disruption.cpp
```diff
@@ -237,14 +237,16 @@ struct add_impacts_visitor : public apply_impacts_visitor {
             }
 
             auto nb_rt_vj = mvj->get_rt_vj().size();
-            std::string new_vj_uri = mvj->uri + ":modified:" + std::to_string(nb_rt_vj) + ":" + impact->disruption->uri;
+            std::string new_vj_name =
```
kinnou02: I don't see any reason to change the name

kinnou02 (changes requested):

kadhikari (approved):

source/ed/build_helper.cpp
```diff
@@ -149,14 +149,14 @@ nt::VehicleJourney* VJ::make() {
     const auto uri_str =
         _uri.empty() ? "vj:" + line_name + ":" + std::to_string(pt_data.vehicle_journeys.size()) : _uri;
     if (is_frequency) {
-        auto* fvj = mvj->create_frequency_vj(uri_str, vj_type, _vp, route, sts, pt_data);
+        auto* fvj = mvj->create_frequency_vj(uri_str, uri_str, vj_type, _vp, route, sts, pt_data);
```
mbounabi: Twice the same parameter?

source/jormungandr/tests/kirin_realtime_tests.py
```diff
@@ -582,7 +582,7 @@ def test_vj_delay_day_after(self):
             '20120615T070436',
             '20120615T070520',
         ]  # pt_walk + rt 07:02:24
-        assert get_used_vj(day_after_response), [[] == ['vjA:modified:0:96231_2015-07-28_0']]
+        assert get_used_vj(day_after_response), [[] == ['vehicle_journey:vjA:modified:0:96231_2015-07-28_0']]
```
benoit-bst, line 592: Okay, if I understand, the main reason is for being ISO with the rest of the code

source/type/type.cpp
```diff
@@ -650,21 +652,23 @@ VJ* MetaVehicleJourney::impl_create_vj(const std::string& uri,
...
 
 FrequencyVehicleJourney* MetaVehicleJourney::create_frequency_vj(const std::string& uri,
+                                                                 const std::string& name,
                                                                  const RTLevel level,
                                                                  const ValidityPattern& canceled_vp,
                                                                  Route* route,
                                                                  std::vector<StopTime> sts,
                                                                  nt::PT_Data& pt_data) {
-    return impl_create_vj<FrequencyVehicleJourney>(uri, level, canceled_vp, route, std::move(sts), pt_data);
+    return impl_create_vj<FrequencyVehicleJourney>(uri, name, level, canceled_vp, route, std::move(sts), pt_data);
```
nonifier, line 662: Out of curiosity, do you know why we have to `std::move` the stop times ? Would be much nicer to have a `const std::vector<StopTime>& sts` as parameter to avoid confusion and misuse of the function.

nonifier: I guess, same applies to `MetaVehicleJourney::create_discrete_vj` further down.

nonifier (approved):
Do you reckon this will break some of the Artemis tests?

source/ed/build_helper.cpp
pbougue: Added a commit to differentiate this in tests... I know why I didn't want to do that: it's fat ;)

source/jormungandr/tests/kirin_realtime_tests.py
pbougue: Yep, RT-VJ having their uris prefixed the same way.

source/kraken/apply_disruption.cpp
pbougue: After discussion: added a commit to let it blank by default as it is displayed to traveler.
(my previous code kept the default name like before)

source/type/type.cpp
pbougue: Good question, not a big memory of that, it was done there: https://github.com/CanalTP/navitia/pull/1258
The few I remember was a perf issue/concern.

I think the move shouldn't be changed in that PR, although a rename/comment may be added? Suggestions welcome!

pbougue:
Jenkins, test this please.

pbougue:
:heavy_check_mark: No Artemis test should break (taxi is already broken): https://ci.navitia.io/job/Artemis/2145

kinnou02:
Jenkins, test this please.

kinnou02 (approved):

benoit-bst (approved):

pbougue:
Only last 2 commits are new, last one should fix debug-compiled tests.

AurelienLP:
Jenkins, test this please.

PR #2781, opened 2019-05-07
kraken: splitting of type.h Strikes Back

kinnou02:
you should read it commit per commit


benoit-bst (approved):
Thorough work for a fasty result :cat2: 
May the force be with you

nonifier (approved):

pbougue (approved):
:+1: on the commit for review :)

PR #2783, opened 2019-05-09
[Ed] add a cli param to fetch admins from cities to osm2ed

antoine-de:
add a new param to `osm2ed`: `--cities-connection-string=<cnx string to the database>`. With this params, the administrative regions are no longer fetched from osm's relation, but from the `cities` database (this db being filled up either by [cosmogony2cities](https://github.com/CanalTP/cosmogony2cities) or by the [legacy tool](https://github.com/CanalTP/navitia/blob/dev/source/cities/cities.h).

This makes it possible to have nicely read admins even in kraken.

To enable this in tyr, change the `admins_from_cities_db` instance variable to True.

#### results

Since we fetch the admins by bulk (to reduce the number of calls to the database, we end up with more admins than before.
Here are the admins imported in qgis:
Capture d’écran de 2019-05-09 12-04-32
(pink is new, green is old)

We could filter the other admins, but I don't think it's a problem to keep them.

#### bench 

This costs a bit at import time since we need to make some calls the the database
On my computer, importing a pbf on ile de france 

old | new
-----|------
2m45|3m07

Since we do the query to the db in bulk (asking for all the admins within a 20km square bbox), we only make 80 calls to the db for the idf

#### :warning: I corrected what seems to be a mistake :warning: but it might breaks stuff :roll_eyes: 
A node that was just on an admin boundary was not associated to any admin with `boost::geometry::within` is a strict inclusion. I changed `within` to `covered_by` to fix this.

I'm not sure about the implication of this change though since it's not trivial. The thing is that now, we have less ways not associated to an admins (since the use the node's admin to compute the way's admin), but more way associated to more than one admin....

on ile de france:

name| old | new
-----|-----|------
way with no admin|1276 |0
way with a name, but no admin |559|0
way with more than one admin |11131|12859

#### No more admin 9-10
cities db contains only real city (admin=8 in france), so now osm2ed will no longer import paris district. I don't think it's a problem since I don't think they are used afterward

### DIFF
--- a/source/ed/osm2ed.cpp
+++ b/source/ed/osm2ed.cpp
@@ -68,7 +68,8 @@ void ReadRelationsVisitor::relation_callback(uint64_t osm_id,
     const auto boundary = tags.find("boundary");
 
     std::string insee = "", postal_code = "", name = "";
-    if (tmp_admin_level != tags.end() && boundary != tags.end() && boundary->second == "administrative") {
+    if (!use_cities && tmp_admin_level != tags.end() && boundary != tags.end()
+        && boundary->second == "administrative") {
         // we consider only admin boundaries with level 8, 9 or 10
         std::vector<std::string> accepted_levels{"8", "9", "10"};
         const auto it_level = std::find(accepted_levels.begin(), accepted_levels.end(), tmp_admin_level->second);
@@ -100,8 +101,16 @@ void ReadRelationsVisitor::relation_callback(uint64_t osm_id,
         if (tags.find("addr:postcode") != tags.end()) {
             postal_code = tags.at("addr:postcode");
         }
-        cache.relations.insert(OSMRelation(osm_id, refs, insee, postal_code, name,
-                                           boost::lexical_cast<uint32_t>(tmp_admin_level->second)));
+
+        std::string uri;
+        if (insee.empty()) {
+            uri = "admin:osm:relation:" + std::to_string(osm_id);
+        } else {
+            uri = "admin:fr:" + insee;
+        }
+
+        cache.admins[osm_id] = std::make_unique<OSMAdminRelation>(
+            osm_id, uri, refs, insee, postal_code, name, boost::lexical_cast<uint32_t>(tmp_admin_level->second));
     } else if (tags.find("type") != tags.end() && tags.at("type") == "associatedStreet") {
         uint64_t way_id = std::numeric_limits<uint64_t>::max();
         std::vector<uint64_t> osm_ids;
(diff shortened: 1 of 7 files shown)


kinnou02 (approved):

nonifier:
jenkins, test this please

source/ed/osm2ed.cpp
```diff
@@ -204,44 +213,103 @@ void ReadNodesVisitor::node_callback(uint64_t osm_id, double lon, double lat, co
...
+        if (boost::geometry::covered_by(p, admin->polygon)) {
+            containing_admin = admin.get();
+        }
+
+        // we also store the admin for future use
+        boost::geometry::model::box<point> box;
+        boost::geometry::envelope(admin->polygon, box);
+        Rect r(box.min_corner().get<0>(), box.min_corner().get<1>(), box.max_corner().get<0>(),
+               box.max_corner().get<1>());
+        this->admin_tree.Insert(r.min, r.max, admin.get());
```
nonifier, line 307: It feels weird to create a `std::unique_ptr` but still copying the pointer twice with `this->admin_tree.Insert(r.min, r.max, admin.get())` and `containing_admin = admin.get()`.

Having a `unique_ptr` that shares its pointer sounds like a `std::shared_ptr` to me !

source/ed/osm2ed.cpp
```diff
@@ -551,7 +618,7 @@ void OSMCache::fusion_ways() {
  * Ways are supposed to be order, but they're not always.
  * Also we may have to reverse way before adding them into the polygon
  */
-void OSMRelation::build_polygon(OSMCache& cache) const {
+void OSMAdminRelation::build_polygon(OSMCache& cache) {
```
nonifier, line 621: Why have you lost the constness? The function doesn't seem to have changed :cry:

source/ed/osm2ed.h
```diff
@@ -113,33 +113,48 @@ struct OSMNode {
...
+          const std::string& name,
+          const uint32_t level);
+    Admin(u_int64_t id,
+          const std::string& uri,
+          const std::string& insee,
+          const std::string& postal_code,
+          const std::string& name,
+          const uint32_t level,
+          mpolygon_type&& polygon,
+          point&& center);
```
benoit-bst, line 130: No need a classic copy constructor too ? like
```
Admin(u_int64_t id,
￼           const std::string& uri,
￼           const std::string& insee,
           ...
           mpolygon_type& polygon,
           point& center
```

source/ed/osm2ed.h
```diff
@@ -113,33 +113,48 @@ struct OSMNode {
     mutable std::bitset<2> properties = 0;
 };
 
-struct OSMRelation {
-    const u_int64_t osm_id;
-    CanalTP::References references;
-    const std::string insee = "", name = "";
+struct Admin {
```
benoit-bst, line 116: better name :cat:

source/ed/osm2ed.h
```diff
@@ -200,14 +215,13 @@ struct OSMWay {
         return bg::distance(p, ls);
     }
 
-    std::set<const OSMRelation*> admins() const {
-        std::set<const OSMRelation*> result;
+    bool contains_admin(const Admin* admin) const {
         for (auto node : nodes) {
-            if (node->admin != nullptr) {
-                result.insert(node->admin);
+            if (node->admin == admin) {
```
nonifier: Or in a less baroque style  :stuck_out_tongue: 

```cpp
bool contains_admin(const Admin* admin) const { 
    auto has_admin = [&](const auto& n){ return n->admin == admin; };
    return boost::find_if(nodes, has_admin) != nodes.end();
}
```

Then, you realise how much boilerplate you need to for a c++ lambda  and you wish to write them the Rust way :sob: :sob: :sob:

nonifier (approved):

source/ed/osm2ed.cpp
antoine-de: hum I don't think so, `shared_ptr` is for shared ownership, but here the ownership is only for the main container

source/ed/osm2ed.cpp
antoine-de: this `constness` was just a smoke screen, the method was mutating `mutable` attributes ...

Since I removed the `mutable` attributes, this method needs needs to be not const (this method builds and sets the `polygon`)

source/ed/osm2ed.h
antoine-de: you're right, it's way better (even if too noisy 😜 ), I'll change this on monday

benoit-bst (approved):

source/ed/osm2ed.h
antoine-de: hum I'd say we'll add it when needed no ?

but if you want I can add it

benoit-bst: okay, when it will be needed

kinnou02:
Jenkins, test this please.

source/ed/osm2ed.h
pbougue, line 116: I get those warnings from clang, I think they should be addressed:
```
In file included from /home/pbougue/dev/sources/navitia/source/ed/osm2ed.cpp:31:
/home/pbougue/dev/sources/navitia/source/ed/osm2ed.h:116:8: warning: 'ed::connectors::Admin' has virtual functions but non-virtual destructor [-Wnon-virtual-dtor]
struct Admin {
       ^
/home/pbougue/dev/sources/navitia/source/ed/osm2ed.h:145:8: warning: 'ed::connectors::OSMAdminRelation' has virtual functions but non-virtual destructor [-Wnon-virtual-dtor]
struct OSMAdminRelation : public Admin {
       ^
/home/pbougue/dev/sources/navitia/source/ed/osm2ed.h:116:8: warning: 'Admin' has no out-of-line virtual method definitions; its vtable will be emitted in every translation unit [-Wweak-vtables]
struct Admin {
       ^
```

pbougue (commented):
Is there some work to do on tyr_configurator to support this, then?

xlqian:
(i've added a similar feature into eitri https://github.com/CanalTP/navitia/pull/2650) your solution is more generic

antoine-de:
I don't know for tyr_conigurator

PR #2788, opened 2019-05-15
Concurent integration tests

nonifier:
Jobs might fail on our CI when we try to run 2 concurrent jobs... This is a bummer :-1: 

This is due to 2 things:
 - Jenkins mounts `/tmp` within all dockers it pops. This is a problem as 2 kraken instances will trample each other zmq socket (`ipc://temp/kraken_{instance_name}`).
 - Similar problem exists with Rabbimq's queues. A Kraken make sure the queue is deleted before binding to it. If another Kraken is already up and using a queue, a second Kraken will delete the one already in use with what is potentially in there.

Solution:
- Making sure each test instances have its own unique id.
- Use that unique id to set a zmq_socket and a queue's name. 

We add 2 Kraken command line options:
- `--BROKER.queue=my_name` to set RabbitMQ queue's name
- `--BROKER.queue_auto_delete=true` to tell RabbitMQ to delete the queues when done with it.

### DIFF
--- a/source/jormungandr/tests/tests_mechanism.py
+++ b/source/jormungandr/tests/tests_mechanism.py
@@ -42,7 +42,7 @@ import six
 from mock import MagicMock
 
 if not 'JORMUNGANDR_CONFIG_FILE' in os.environ:
-    os.environ['JORMUNGANDR_CONFIG_FILE'] = (
+    os.environ[str('JORMUNGANDR_CONFIG_FILE')] = str(
         os.path.dirname(os.path.realpath(__file__)) + '/integration_tests_settings.py'
     )
 
@@ -59,8 +59,9 @@ from jormungandr.parking_space_availability import (
     ParkingPlaces,
 )
 from jormungandr.equipments.sytral import SytralProvider
+import uuid
 
-krakens_dir = os.environ['KRAKEN_BUILD_DIR'] + '/tests'
+krakens_dir = os.environ[str('KRAKEN_BUILD_DIR')] + '/tests'
 
 
 class FakeModel(object):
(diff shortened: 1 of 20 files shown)


source/jormungandr/tests/tests_mechanism.py
```diff
@@ -83,19 +84,27 @@ class AbstractTestFixture(unittest.TestCase):
     (the setup() and teardown() methods are called once at the initialization and destruction of the integration tests)
     """
 
+    @classmethod
+    def _get_zmq_socket_name(cls, kraken_name):
+        return "ipc:///tmp/{uid}_{name}".format(uid=cls.uid, name=kraken_name)
```
benoit-bst, line 89: okay, if I understand, this is the new **ipc** that replace `"ipc:///tmp/" + name;` for tests
:+1:

benoit-bst (approved):

source/jormungandr/tests/tests_mechanism.py
nonifier: it is indeed !

source/kraken/maintenance_worker.cpp
```diff
@@ -300,10 +300,14 @@ void MaintenanceWorker::init_rabbitmq() {
     std::string username = conf.broker_username();
     std::string password = conf.broker_password();
     std::string vhost = conf.broker_vhost();
-
     std::string hostname = get_hostname();
-    this->queue_name_task = (boost::format("kraken_%s_%s_task") % hostname % instance_name).str();
-    this->queue_name_rt = (boost::format("kraken_%s_%s_rt") % hostname % instance_name).str();
+    std::string default_queue_name = "kraken_" + hostname + "_" + instance_name;
```
kinnou02, line 304: why not put this directly into `conf.broker_queue()` ?

kinnou02 (approved):

PR #2786, opened 2019-05-13
script : Install dev env + run Navitia demo

benoit-bst:
This PR address several point:

###  Don't want to install third parts on target
Now, if you run a `make install`, the third part install some stuff, but this not usefull for a Navitia installation. I just skiped the Cmake install inside third part with `EXCLUDE_FROM_ALL` option and I tested it.

###  Fix current script, build_navitia.sh
Currently, the script doesn't work for several reasons (config file, build dir, ...). 
I tried to do my best to update it, but it is not the main purpose. 
I especially want to keep it for developers who want to install **Navitia** onto an old OS. they can read the script code what can help them.

###  Add new script that only install dev environment
Split in 2 parts the old script. the first part (ubuntu_bionic_dev_env_install.sh) is only to install dev environment. This is available only for ubuntu-18.04 for the moment. When ubuntu 19.04 will release, I will update it, if it necessary.

### Add new script that build, setup and run Navitia demo
The second part (/build_setup_and_run_navitia_demo.sh) is to run a demo. We don't need it to have a dev environement. Nonetheless, it is very usefull for **newbies** that don't known the Navitia soft compoments.

All scripts are now inside navitia/scripts for more readability (with a README). 
```
navitia/scripts/README.md
navitia/scripts/ubuntu_bionic_dev_env_install.sh
navitia/scripts/build_navitia.sh 
navitia/scripts/build_setup_and_run_navitia_demo.sh
```
A piece of the `CONTRIBUTING.md` doc has been updated.

I tested the 2 new scripts on a new clean Ubuntu 18.04 env, and on my current Ubuntu 18.04 too.
I only tested the old one on my Ubuntu 18.04 except the install dependencies part obviously.
I hope that will help new Navitia players :man_playing_handball: 

PS : This work is like a volunteering job, so be kind. It is better than nothing :cat2:

PR #2789, opened 2019-05-15
kraken: fix boost::posix_time::seconds on ubuntu disco with boost 1.67

benoit-bst:
With this fix, **Kraken** can compile on a clean **ubuntu 19.04** env (**gcc 9**, **boost 1.67**)
Boost was too lenient about the input type of **boost:posix_time:seconds()**. Now, it is fix for "long" type like in the doc.
It's job is linked with https://github.com/CanalTP/navitia/pull/2786


kinnou02 (approved):

mbounabi (approved):
Very good! Impressive! :+1:  :rainbow:

kadhikari (approved):

PR #2792, opened 2019-05-20
Kraken : Optimisation for dijkstra

AurelienLP:
When we call dijkstra for the matrix, we do some useless copies. This PR aims to avoid them and move the code of dijkstra_shortest_paths_with_heap.h into the class dijkstra_path_finder. We also remove the complexity of huge templates.

We also add some unit_tests because why not.


xlqian (approved):

source/georef/dijkstra_path_finder.cpp
```diff
@@ -210,6 +212,23 @@ DijkstraPathFinder::get_duration_with_dijkstra(const navitia::time_duration& rad
...
 
+template <class Visitor>
+void DijkstraPathFinder::dijkstra(const std::array<georef::vertex_t, 2>& origin_vertexes, const Visitor& visitor) {
+    // Note: the predecessors have been updated in init
+    // Fill color map in white before dijkstra
+    std::fill(color.data.get(), color.data.get() + (color.n + color.elements_per_char - 1) / color.elements_per_char,
+              0);
+
+    // we filter the graph to only use certain mean of transport
+    using filtered_graph = boost::filtered_graph<georef::Graph, boost::keep_all, TransportationModeFilter>;
```
pbougue, line 210: Same remark than the late review on Astar: can't you do that on init (and would it impact perf) ?

AurelienLP: Won't change anything since it is called only once, like the init.

PR #2798, opened 2019-05-21
Doc: update release doc

nonifier:
Update release documentation with correct commands to run.
[skip ci]

PR #2799, opened 2019-05-22
Fix release script when hotfixing

nonifier:
The git repository is based on the directory above why was causing `git.add()` to fail.
Enforcing the file path with on absolute path fixes the issue.

[skip ci]

### DIFF
--- a/release/script_release.py
+++ b/release/script_release.py
@@ -39,7 +39,7 @@ import subprocess
 import re
 from sys import exit, argv
 from shutil import copyfile
-from os import remove, stat
+from os import remove, stat, path
 import codecs
 import requests
 import logging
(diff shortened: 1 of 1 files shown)


pbougue (approved):

kinnou02 (approved):

release/script_release.py
```diff
@@ -271,7 +271,7 @@ def update_changelog(self):
 
         copyfile(back_filename, self.changelog_filename)
 
-        self.git.add(self.changelog_filename)
+        self.git.add(path.abspath(self.changelog_filename))
```
mbounabi, line 274: Isn't it the same fix from the PR #2798  ?

nonifier:
Duplicated as @mbounabi said

PR #2800, opened 2019-05-22
Jormungandr: fallback to kraken for /places/<uri>

antoine-de:
stop_points are not loaded into the new autocomplete system (mimir), and
are not meant to.

To not break integrations where users are using some ids from other api
results (mainly from /journeys) to get more information on the object
with `/places/<uri>`, if mimir does not find the object, we now look
into the kraken.

Note:
* this works only for `/v1/coverage/<bobette>/places/<uri>`, not for the
autocomplete (`/v1/coverage/<bobette>/places?q=bob`)
* this works only when querying one coverage
(`/v1/coverage/<bobette>/places`), not for the global autocomplete
(`/v1/places`)

### DIFF
--- a/source/jormungandr/jormungandr/instance.py
+++ b/source/jormungandr/jormungandr/instance.py
@@ -219,12 +219,8 @@ class Instance(object):
             )
         return backend
 
-    @property
-    def kraken_autocomplete(self):
-        backend = global_autocomplete.get('kraken')
-        if backend is None:
-            raise RuntimeError('impossible to find autocomplete kraken for instance {}'.format(self.name))
-        return backend
+    def stop_point_fallbacks(self):
+        return [a for a in global_autocomplete.values() if a.is_handling_stop_points()]
 
     def get_models(self):
         if self.name not in g.instances_model:
--- a/source/jormungandr/jormungandr/instance.py
+++ b/source/jormungandr/jormungandr/instance.py
@@ -219,6 +219,13 @@ class Instance(object):
             )
         return backend
 
+    @property
+    def kraken_autocomplete(self):
+        backend = global_autocomplete.get('kraken')
+        if backend is None:
+            raise RuntimeError('impossible to find autocomplete kraken for instance {}'.format(self.name))
+        return backend
+
     def get_models(self):
         if self.name not in g.instances_model:
             g.instances_model[self.name] = self._get_models()
--- a/source/jormungandr/jormungandr/scenarios/simple.py
+++ b/source/jormungandr/jormungandr/scenarios/simple.py
@@ -191,19 +191,19 @@ class Scenario(object):
                 uri=request["uri"], instances=[instance], current_datetime=request['_current_datetime']
             )
         except UnknownObject as e:
-            if not autocomplete.is_handling_stop_points():
-                # the autocomplete have not found anything and it does not handle stop_points,
-                # we'll search for the object in kraken too
-
-                kraken_res = instance.kraken_autocomplete.get_by_uri(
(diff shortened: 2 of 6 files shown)


source/jormungandr/jormungandr/scenarios/simple.py
```diff
@@ -185,9 +186,24 @@ def places(self, request, instance):
...
         autocomplete = instance.get_autocomplete(request.get('_autocomplete'))
-        return autocomplete.get_by_uri(
-            uri=request["uri"], instances=[instance], current_datetime=request['_current_datetime']
-        )
+        try:
+            return autocomplete.get_by_uri(
+                uri=request["uri"], instances=[instance], current_datetime=request['_current_datetime']
+            )
+        except UnknownObject as e:
+            if autocomplete.NAME != 'Kraken':
```
kinnou02: :nauseated_face:

kinnou02 (approved):

nonifier (approved):

source/jormungandr/jormungandr/scenarios/simple.py
```diff
@@ -185,9 +186,24 @@ def places(self, request, instance):
...
+        try:
+            return autocomplete.get_by_uri(
+                uri=request["uri"], instances=[instance], current_datetime=request['_current_datetime']
+            )
+        except UnknownObject as e:
+            if not autocomplete.is_handling_stop_points():
+                # the autocomplete have not found anything and it does not handle stop_points,
+                # we'll search for the object in kraken too
+
+                kraken_res = instance.kraken_autocomplete.get_by_uri(
```
nonifier: Does this mean that we'll never be able to fully get rid of Kraken's autocompletion ?

antoine-de: yup, at least for the /places/<uri>

(and maybe for the text search too for corner case, like sherbrooke where users explicitly search for stop_points)

source/jormungandr/jormungandr/instance.py
```diff
@@ -219,6 +219,13 @@ def autocomplete(self):
             )
         return backend
 
+    @property
+    def kraken_autocomplete(self):
+        backend = global_autocomplete.get('kraken')
```
pbougue: Can't we loop over available autocomplete instead of looking for a key?
And maybe rename method `is_possible_fallback()`, changing a bit the logic to avoid requesting mimir when kraken fails? As we don't check that the requested id starts with `stop_point:` it seems doable.

source/jormungandr/jormungandr/scenarios/simple.py
pbougue: From afar:
* adding a stop_point index to mimir may be a thing to consider (and use/activate it only on those corner cases...)
* as for the reverse autocomplete isn't it doable to have an exact search on pt-ids, and coordinates of addresses?

Don't know the priority, doability, cost of all this.

kinnou02: Maybe some day we will import stop_points and others pt_objects in bragi, but not now

source/jormungandr/jormungandr/instance.py
antoine-de: I tried another implementation based on what you asked.

I'm not sure the abstraction helps the understanding of the code, but I'll let you judge it

pbougue (approved):
The implementation looks more "logic" to me now.
Still feels weird to name it stop_point, as the implementation is generic, but it's the real diff currently.

kinnou02:
Jenkins, test this please.

PR #2801, opened 2019-05-22
Build: add pre-commit for flake8

kinnou02:
I add to disable some warning, a few of them should be enable later but
they require a lot of change. Most of them are from star import...

A few unused import are infact used, I will resolve this in another PR

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -10,7 +10,7 @@ repos:
     -   id: trailing-whitespace
     -   id: check-ast
     -   id: check-merge-conflict
-    #-   id: flake8
+    -   id: flake8
 -   repo: https://github.com/CanalTP/navitia-pre-commit
     rev: master
     hooks:
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -11,6 +11,7 @@ repos:
     -   id: check-ast
     -   id: check-merge-conflict
     -   id: flake8
+        language_version: python2.7
 -   repo: https://github.com/CanalTP/navitia-pre-commit
     rev: master
     hooks:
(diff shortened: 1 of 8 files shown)


benoit-bst (approved):
:rainbow: Can avoid the "abort" :imp:

xlqian (commented):
:+1:

xlqian (approved):

mbounabi (approved):

.pre-commit-config.yaml
```diff
@@ -10,7 +10,7 @@ repos:
     -   id: trailing-whitespace
     -   id: check-ast
     -   id: check-merge-conflict
-    #-   id: flake8
+    -   id: flake8
```
pbougue, line 13: Can you force python version (as it doesn't pass on python3.6)?
```suggestion
    -   id: flake8
        language_version: python2.7
```

source/jormungandr/requirements_dev.txt
```diff
@@ -11,3 +11,4 @@ jsonschema==2.6.0
 pytest-timeout==1.3.3
 mypy==0.670; python_version > '3.0'
 mypy-extensions==0.4.1; python_version > '3.0'
+flake8==3.5.0
```
pbougue, line 14: Aren't they all supposed to be in a different `requirements_precommit.txt` file?

kinnou02: I don't see why, you should have them when you are developing as most editor will use mypy and flake to check the code while you write it.

pbougue (approved):
:+1: It's better than before

The other problems on precommit should be handled separately:
* as discussed: maybe separate precommit requirements (and include them into each requirement_dev)
* on short term: change scope of mypy to just source/jormungandr/jormungandr as travis only checks that (before including more and fixing `precommit run --all` )
* change scope of trailing-whitespace (exclude rst and md files)
* update travis to work using the precommit for real ;)

PR #2805, opened 2019-05-24
Jormun: read stop's lines in /places

antoine-de:
linked to https://github.com/CanalTP/mimirsbrunn/pull/307

Now bragi can output the lines that pass by a given stop_area.

This will make it possible to add a nice picto in the search results with the lines passing by the stop. It will makes it easier to differentiate the results at first glance.

beware the death preview (:drum: ...):

stop_with_line (1)

The lines are sorted by the same magical sort as kraken (but some integrator might want to group/filter the lines by mode).

internally tracked as https://jira.kisio.org/browse/NAVP-845

### DIFF
--- a/source/jormungandr/jormungandr/interfaces/v1/serializer/geocode_json.py
+++ b/source/jormungandr/jormungandr/interfaces/v1/serializer/geocode_json.py
@@ -41,6 +41,7 @@ from jormungandr.interfaces.v1.serializer.base import (
     IntNestedPropertyField,
     value_by_path,
     BetaEndpointsSerializer,
+    DictGenericSerializer,
 )
 from jormungandr.utils import get_house_number
 from jormungandr.autocomplete.geocodejson import create_address_field, get_lon_lat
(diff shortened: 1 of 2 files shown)


kinnou02 (approved):

kinnou02:
We might want to add a bit of documentation stating that `line` objects from autocomplete isn't exactly the same as in other API.

nonifier (approved):

source/jormungandr/jormungandr/interfaces/v1/serializer/geocode_json.py
```diff
@@ -222,6 +223,15 @@ def get_address(self, obj):
         return AddressSerializer(obj).data
 
 
+class LineSerializer(NestedDictGenericField):
```
pbougue, line 226: OK, so you need a DictSerializer and a PbSerializer.
Can we name them differently to avoid mixup (don't know how the swagger behaves with same names)?
2 ideas :
* PbLineSerializer and DictLineSerializer (close to each other in pt.py) if we want them to match (probably with a comment to "enforce" that match)... With `display_none=True` for code and so on.
* or LineSerializer and GeocodeLineSerializer if we don't want them to match precisely (but then we should probably document it as @kinnou02 said)

pbougue (approved):
It's fine, but naming (and maybe doc) may be improved, I let you decide.

source/jormungandr/jormungandr/interfaces/v1/serializer/geocode_json.py
antoine-de: the fact that it is nested in geocode_json is not enough for you ?

the point is the same for all the types of this module (`StopAreaSerializer`, `AddressSerializer`, ...)

antoine-de: to be a bit more precise, the serializer is very tightly coupled to mimir's response, it is not a `DictLineSerializer`, it is a `MimirLineSerializer`, that's why I think it should be kept in `geocode_json` (and thus I find the naming fine, but I can change it it you prefer)

pbougue: OK, I didn't check the other names.
It's probably more some tests to add to avoid diffs between existing autocomplete services, then.
But it's for another PR if any :+1:

PR #2806, opened 2019-05-24
[Jormun] Dedup equipment details

nonifier:
Dedup equipment details when multiple stop point refer to the same equipment.

https://jira.kisio.org/browse/NAVITIAII-2749

### DIFF
--- a/source/jormungandr/jormungandr/equipments/sytral.py
+++ b/source/jormungandr/jormungandr/equipments/sytral.py
@@ -165,25 +165,28 @@ class SytralProvider(object):
         :param data: equipments data received from the webservice
         :param stop_area_equipments_list: list of stop_area_equipments from the protobuf response
         """
-        for sae in stop_area_equipments_list:
-            for st in sae.stop_area.stop_points:
-                for code in st.codes:
-                    if code.type in self.code_types:
-                        equipments_list = jmespath.search(
-                            "equipments_details[?id=='{}']".format(code.value), data
-                        )
 
-                        if equipments_list:
-                            equipment = equipments_list[0]
-                            # Fill PB
-                            equipment_details = sae.equipment_details.add()
-                            self._fill_equipment_details(
-                                equipment_form_web_service=equipment, equipment_details=equipment_details
-                            )
-                        else:
-                            equipment_details = sae.equipment_details.add()
-                            self._fill_default_equipment_details(
-                                id=code.value,
-                                embedded_type=self._embedded_type(code.type),
-                                equipment_details=equipment_details,
-                            )
+        for sae in stop_area_equipments_list:
+            """
+            Codes might be duplicated across different stop points.
+            Because we report equipments on a stop area basis, we  don't want them duplicated
+            """
+            unique_codes = {str(code): code for st in sae.stop_area.stop_points for code in st.codes}
+            for code in unique_codes.values():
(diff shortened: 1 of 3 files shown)


source/jormungandr/jormungandr/equipments/sytral.py
```diff
@@ -165,25 +165,28 @@ def _process_for_equipment_reports(self, data, stop_area_equipments_list):
...
-                for code in st.codes:
-                    if code.type in self.code_types:
-                        equipments_list = jmespath.search(
-                            "equipments_details[?id=='{}']".format(code.value), data
-                        )
+            """
+            Codes might be duplicated across different stop points.
+            Because we report equipments on a stop area basis, we  don't want them duplicated
+            """
+            unique_codes = {str(code): code for st in sae.stop_area.stop_points for code in st.codes}
```
pbougue: Shouldn't you stick to unicode here?
Also, can't you just use a set? (or maybe store stop_points when we will want to display them :-/ )

Not sure: is the type+value couple managed here? Or is it supposed to be (as test's comment makes me think that)?

source/jormungandr/jormungandr/equipments/tests/sytral_equipment_test.py
```diff
@@ -84,3 +84,31 @@ def equipments_get_information_test():
     st = create_stop_point("TCL_ASCENCEUR", "261")
     provider.get_informations_for_journeys([st])
     assert not st.equipment_details
+
+
+def equipments_get_informations_for_reports_not_duplicated():
+    """
+    When 2 stop points (from the same stop are) refer to the same equipment (type and value),
```
pbougue:
```suggestion
    When 2 stop points (from the same stop area) refer to the same equipment (type and value),
```

benoit-bst (approved):
Just make sure you don't have to fix it for `journeys API` (_process_for_journeys function)

source/jormungandr/jormungandr/equipments/sytral.py
mbounabi: A set can't be used because the item 'Code' is not hashable... However, why create a dict with key = value?

nonifier: I've updated the structure using unicodes, although this string is only used internally to identify an object. Thus, no harm done to external code that may be using unicodes char.

As @mbounabi said, Code is not `hashable`, hence the usage of a dictionary to mimic the behavior of a `set` and I'm using `key=value` because you cannot create a dict without a key. 

The usage of a dictionary was the easiest way for me to filter out identical codes without using a `set`. I'm happy to take any suggestion to do so in a more pythonic way :snake:

nonifier: @pbougue : the couple `type+value` is managed as the `str` or `unicode` function from a protobuf object serialises all members of this object. Therefore, `unicode(code)` will look like `u'type:"TCL_ESCALIER" value: "2"'`

mbounabi: I have made a few attempts to improve this part (with a list and append if not in list)... But I'm not pythonically satisfied with everything I could come up with... So, I'll validate this :sweat_smile:

mbounabi (approved):

kinnou02 (approved):

PR #2809, opened 2019-05-28
All: Update precommit

pbougue:
:mag: review by commit (to separate whitespace commit)
:dart: main commit: https://github.com/CanalTP/navitia/pull/2809/commits/2c82b8797b9e23e21ebfb1eec43911b1c1088709

- [x] separate precommit requirements
- [x] change scope of mypy to just `source/jormungandr/jormungandr` : then https://jira.kisio.org/browse/NAVP-1330 to have it everywhere
- [x] `pre-commit run trailing-whitespace --all` (allows double-space use for new line in .rst and .md files, no file to exclude)
- [x] travis now uses `pre-commit run --all` so it's the same command as locally
- [x] check that travis still spots problems using full pre-commit (cf. the last 2 commits + :red_circle: https://travis-ci.org/CanalTP/navitia/builds/538678448?utm_source=github_status&utm_medium=notification)

Work related to that comment: https://github.com/CanalTP/navitia/pull/2801#pullrequestreview-241697993


.travis.yml
```diff
@@ -45,10 +45,12 @@ matrix:
     - python: "3.6"
       before_script:
         - python3.6 -m pip install black==18.9b0
-        - pip install -r source/jormungandr/requirements_dev.txt
+        - pip install -r requirements_pre-commit.txt
+        - pip install pre-commit
```
benoit-bst, line 45: Woo and it currently works ? Thus, pre-commit is installed by default on `Travis` I guess
> previously(currently) mypy & black were launched without using pre-commit

requirements_pre-commit.txt
```diff
@@ -0,0 +1,3 @@
+mypy==0.670; python_version > '3.0'
```
benoit-bst, line 2: Nice

requirements_pre-commit.txt
```diff
@@ -0,0 +1,4 @@
+black==18.9b0; python_version > '3.6'
```
pbougue, line 1: Added black to pip requirements, as I guess it's the way every developper installs it (and the way travis does).
Let me know if it's an issue for someone's dev-setup.

.travis.yml
```diff
@@ -23,32 +23,29 @@ services:
...
-        - python3.6 -m pip install black==18.9b0
-        - pip install -r source/jormungandr/requirements_dev.txt
+        - pip install -r requirements_pre-commit.txt
+        - pip install pre-commit
+        - pre-commit install
         - bash source/scripts/build_protobuf.sh
       script:
-        - black --check --diff .
-        - pushd source/jormungandr && MYPYPATH=.:../navitiacommon mypy --ignore-missing-imports --py2 jormungandr && popd
+        - pre-commit run --all --show-diff-on-failure
```
mbounabi, line 49: Does this mean that all the hooks in the pre-commit file will be run only for python 3?

pbougue: When nothing is specified in the pre-commit-config, yes.

But actually for flake8 it uses python2.7 that's available when launching pre-commit here (if launched with python3.6, flake8 currently fails).

That's the way I launch it locally (default = 3.6), I don't know if we may end with different behaviors on hooks without version specified. But then we will just pin them to one of the version, as it's working for both flake8(2.7) and black(3.6).

.pre-commit-config.yaml
```diff
@@ -25,4 +25,5 @@ repos:
     hooks:
     -   id: mypy
         args: [--ignore-missing-imports, --py2]
+        files: source/jormungandr/jormungandr
```
nonifier, line 28: Do we have a Ticket to keep track of running Mypy on the rest of the code base ?

nonifier (approved):
Sweet ! :+1:

kinnou02 (approved):

mbounabi (approved):

.pre-commit-config.yaml
pbougue: We didn't, but now it's here: https://jira.kisio.org/browse/NAVP-1330

PR #2813, opened 2019-06-04
[Kraken]: Split sn proximity list

xlqian:
split georef proximity list into 3: one proximity list for each graph(walking, bike car)

A projection problem has been reported after merging  this PR:https://github.com/CanalTP/navitia/pull/2774

The problem is when flann try to find nearest vertices, it doesn't see if a node is on walking/bike/car graph. In case where we try to project a coord on the car graph and flann has found only some nodes on walking graph, the projection will fail. This case is particularly true when projecting a stop_area on bike/car graph, since the nodes of walking graph are highly dense.  

The advantage of doing so is be able to eliminate nodes that don't have any edges. 
The disadvantage is that it becomes more memory consuming. 

Proximity lists are built after data loading. The loading time is about 5 seconds on stif and 2 seconds on fr-idf

### DIFF
--- a/source/georef/dijkstra_path_finder.cpp
+++ b/source/georef/dijkstra_path_finder.cpp
@@ -176,9 +176,9 @@ routing::map_stop_point_duration DijkstraPathFinder::find_nearest_stop_points(
 
 struct ProjectionGetterOnFly {
     const GeoRef& geo_ref;
-    const nt::idx_t offset;
+    const type::Mode_e mode;
     const georef::ProjectionData operator()(const type::GeographicalCoord& coord) const {
-        return georef::ProjectionData{coord, geo_ref, offset, geo_ref.pl};
+        return georef::ProjectionData{coord, geo_ref, mode};
     }
 };
 
@@ -196,7 +196,7 @@ DijkstraPathFinder::get_duration_with_dijkstra(const navitia::time_duration& rad
         offset = geo_ref.offsets[mode];
     }
 
-    ProjectionGetterOnFly projection_getter{geo_ref, offset};
+    ProjectionGetterOnFly projection_getter{geo_ref, mode == type::Mode_e::Car ? nt::Mode_e::Walking : mode};
     return start_dijkstra_and_fill_duration_map<DijkstraPathFinder::coord_uri, type::GeographicalCoord,
                                                 ProjectionGetterOnFly>(radius, dest_coords, projection_getter);
 }
--- a/source/georef/dijkstra_path_finder.cpp
+++ b/source/georef/dijkstra_path_finder.cpp
@@ -188,13 +188,6 @@ DijkstraPathFinder::get_duration_with_dijkstra(const navitia::time_duration& rad
     if (dest_coords.empty()) {
         return {};
     }
-    nt::idx_t offset;
-    if (mode == type::Mode_e::Car) {
-        // on direct path with car we want to arrive on the walking graph
-        offset = geo_ref.offsets[nt::Mode_e::Walking];
-    } else {
-        offset = geo_ref.offsets[mode];
-    }
 
     ProjectionGetterOnFly projection_getter{geo_ref, mode == type::Mode_e::Car ? nt::Mode_e::Walking : mode};
     return start_dijkstra_and_fill_duration_map<DijkstraPathFinder::coord_uri, type::GeographicalCoord,
--- a/source/georef/dijkstra_path_finder.cpp
+++ b/source/georef/dijkstra_path_finder.cpp
@@ -176,7 +176,7 @@ routing::map_stop_point_duration DijkstraPathFinder::find_nearest_stop_points(
 
 struct ProjectionGetterOnFly {
     const GeoRef& geo_ref;
-    const type::Mode_e mode;
+    const type::Mode_e mode = type::Mode_e::Walking;
     const georef::ProjectionData operator()(const type::GeographicalCoord& coord) const {
         return georef::ProjectionData{coord, geo_ref, mode};
     }
--- a/source/georef/georef.cpp
+++ b/source/georef/georef.cpp
@@ -390,6 +390,9 @@ void GeoRef::build_proximity_list() {
 
     auto build_sn_pl = [this](proximitylist::ProximityList<vertex_t>& sn_pl, nt::idx_t offset) {
         for (vertex_t v = offset; v < nb_vertex_by_mode + offset; ++v) {
+            if (boost::algorithm::none_of(boost::out_edges(v, graph),
+                                          [=](const auto& e) { return is_sn_edge(*this, e); }))
+                continue;
             sn_pl.add(graph[v].coord, v);
         }
         sn_pl.build();
--- a/source/georef/georef.cpp
+++ b/source/georef/georef.cpp
@@ -38,6 +38,7 @@ www.navitia.io
 #include <boost/foreach.hpp>
 #include <boost/geometry.hpp>
 #include <boost/range/algorithm/sort.hpp>
+#include <boost/algorithm/cxx11/none_of.hpp>
 #include <boost/range/algorithm/lexicographical_compare.hpp>
 #include <boost/math/constants/constants.hpp>
 #include <array>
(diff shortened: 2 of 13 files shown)


source/georef/georef.cpp
```diff
@@ -388,17 +378,33 @@ void GeoRef::init() {
...
+    pl_bike.clear();
+    pl_car.clear();
+    poi_proximity_list.clear();
 
-    // do not build the proximitylist with the edge of other transportation mode than walking (and walking HAS to be the
-    // first graph)
-    for (vertex_t v = 0; v < nb_vertex_by_mode; ++v) {
-        pl.add(graph[v].coord, v);
-    }
+    auto log = log4cplus::Logger::getInstance("");
```
kinnou02: please name it

source/georef/dijkstra_path_finder.cpp
```diff
@@ -188,15 +188,8 @@ DijkstraPathFinder::get_duration_with_dijkstra(const navitia::time_duration& rad
...
-    nt::idx_t offset;
-    if (mode == type::Mode_e::Car) {
-        // on direct path with car we want to arrive on the walking graph
-        offset = geo_ref.offsets[nt::Mode_e::Walking];
-    } else {
-        offset = geo_ref.offsets[mode];
-    }
 
-    ProjectionGetterOnFly projection_getter{geo_ref, offset};
+    ProjectionGetterOnFly projection_getter{geo_ref, mode == type::Mode_e::Car ? nt::Mode_e::Walking : mode};
```
AurelienLP, line 192: What if it is Taxi or Ridesharing (CarNoPark) ?

source/georef/georef.cpp
```diff
@@ -38,6 +38,7 @@ www.navitia.io
 #include <boost/foreach.hpp>
 #include <boost/geometry.hpp>
 #include <boost/range/algorithm/sort.hpp>
+#include <boost/algorithm/cxx11/none_of.hpp>
```
AurelienLP, line 41: std::none_of in < algorithm > is not available ?

source/georef/georef.cpp
```diff
@@ -388,17 +378,33 @@ void GeoRef::init() {
...
 
-    poi_proximity_list.clear();
+    LOG4CPLUS_INFO(log, "Building Proximity list for walking graph");
+    build_sn_pl(pl_walking, offsets[nt::Mode_e::Walking]);
+
+    LOG4CPLUS_INFO(log, "Building Proximity list for bike graph");
+    build_sn_pl(pl_bike, offsets[nt::Mode_e::Bike]);
+
+    LOG4CPLUS_INFO(log, "Building Proximity list for car graph");
+    build_sn_pl(pl_car, offsets[nt::Mode_e::Car]);
```
AurelienLP, line 404: Again what about CarNoPark ?

source/routing/heat_map.cpp
```diff
@@ -141,26 +141,27 @@ static std::vector<std::vector<Projection>> find_projection(BoundBox box,
...
     const size_t offset_lat = floor(min_dist / (height_step * N_DEG_TO_DISTANCE)) + 1;
-    auto begin = std::lower_bound(
-        worker.pl.items.begin(), worker.pl.items.end(), box.min.lon(),
-        [](const proximitylist::ProximityList<georef::vertex_t>::Item& i, double min) { return i.coord.lon() < min; });
 
-    if (begin == worker.pl.items.end()) {
-        return dist_pixel;
+    auto box_center = type::GeographicalCoord{(box.min.lon() + box.max.lon()) / 2, (box.min.lat() + box.max.lat()) / 2};
+
+    auto radius = box.min.distance_to(box.max);
```
AurelienLP, line 147: Isn't this the diameter?

source/routing/heat_map.cpp
```diff
@@ -141,26 +141,27 @@ static std::vector<std::vector<Projection>> find_projection(BoundBox box,
...
+
+    auto objects_inside = worker.pl_walking.find_within(box_center, 1.4 * radius);
```
AurelienLP: What is 1.4 ?

source/routing/heat_map.cpp
```diff
@@ -141,26 +141,27 @@ static std::vector<std::vector<Projection>> find_projection(BoundBox box,
...
 
-    const auto coslat = cos(begin->coord.lat() * type::GeographicalCoord::N_DEG_TO_RAD);
-    for (auto it = begin; it != end; ++it) {
-        const auto& source = it->coord;
         if (!box.contains(source)) {
             continue;
         }
         const auto rank_source = find_rank(box, source, height_step, width_step);
-        BOOST_FOREACH (georef::edge_t e, boost::out_edges(it->element, worker.graph)) {
+        BOOST_FOREACH (georef::edge_t e, boost::out_edges(element, worker.graph)) {
```
AurelienLP: for(auto const& e : boost::out_edges(element, worker.graph))

source/georef/georef.cpp
xlqian: not for range

source/georef/dijkstra_path_finder.cpp
xlqian: taxi and ridesharing are already taken into consideration at this stage

source/georef/georef.cpp
xlqian: the offset of car and carnopark are the same

source/routing/heat_map.cpp
xlqian: oops

source/georef/georef.cpp
nonifier: `std::ranges::none_of` -> `since C++20`  ==>> :sob:  :sob:  :sob:  :sob:  :sob:

source/georef/dijkstra_path_finder.cpp
```diff
@@ -176,9 +176,9 @@ routing::map_stop_point_duration DijkstraPathFinder::find_nearest_stop_points(
 
 struct ProjectionGetterOnFly {
     const GeoRef& geo_ref;
-    const nt::idx_t offset;
+    const type::Mode_e mode;
```
nonifier: Any default mode by any chance ?

source/georef/georef.h
```diff
@@ -226,8 +226,14 @@ struct GeoRef {
         autocomplete::Autocomplete<unsigned int>(navitia::type::Type_e::POI);
 
     /// Indexe tous les nœuds
+
+    // TODO: Remove this before the next re-binarisation
```
nonifier: Is this comment still relevant ?

nonifier (approved):

source/georef/dijkstra_path_finder.cpp
xlqian: a test is added in artemis for taxi

source/georef/georef.h
```diff
@@ -226,8 +226,14 @@ struct GeoRef {
...
     proximitylist::ProximityList<vertex_t> pl;
```
kinnou02: do we still need `pl`?

source/georef/georef.cpp
kinnou02: we should be able to use it around 2028 :D

source/georef/georef.cpp
```diff
@@ -672,24 +676,27 @@ vertex_t GeoRef::nearest_vertex(const type::GeographicalCoord& coordinates,
...
+        case type::Mode_e::Bss:
+            return nearest_edge(coordinates, pl_walking);
+        case type::Mode_e::Bike:
+            return nearest_edge(coordinates, pl_bike);
+        case type::Mode_e::Car:
+        case type::Mode_e::CarNoPark:
+            return nearest_edge(coordinates, pl_car);
         default:
-            return false;
+            throw navitia::exception("Unknown mode when looking for nearest edges");
```
kinnou02: Please use a  `recoverable_exception`

source/georef/georef.h
xlqian: yes! the old proximity_list includes the nodes of all 3 graphs, it was binarised... now it's split into three and they're rebuilt on the fly.. So no longer need to binarise.. but if i remove them from `serialize()`, we have to binarise...

source/georef/georef.h
xlqian: no longer! if i remove it, we have to re-bina every coverage....

source/routing/heat_map.cpp
xlqian: you're right, it IS a diameter. but for the NN index, it's a search radius :) 

Apparently, the radius is too small, i have to use the diameter. 
For those points that are not inside of the box, we filter them at [this line](https://github.com/CanalTP/navitia/pull/2813/files#diff-482ca655704e9c2f731b66fdf1337634L159)

xlqian:
Jenkins, test this please!

source/georef/georef.h
xlqian: the old object is now removed ... comment removed...

PR #2816, opened 2019-06-04
Kraken: Add ability to configure max core dump file size

nonifier:
Expose a way to configure the core dump max size limit with `GENERAL.core_file_size_limit`

This is equivalent to running cmd 
```sh
$ ulimit -c <my_limit>
```

### DIFF
--- a/source/kraken/configuration.cpp
+++ b/source/kraken/configuration.cpp
@@ -80,6 +80,7 @@ po::options_description get_options_description(const boost::optional<std::strin
 
         ("GENERAL.enable_request_deadline", po::value<bool>()->default_value(true), "enable deadline of request")
         ("GENERAL.metrics_binding", po::value<std::string>(), "IP:PORT to serving metrics in http")
+        ("GENERAL.core_file_size_limit", po::value<int>()->default_value(0), "ulimit that define the maximum size of a core file")
 
         ("BROKER.host", po::value<std::string>()->default_value("localhost"), "host of rabbitmq")
         ("BROKER.port", po::value<int>()->default_value(5672), "port of rabbitmq")
@@ -179,6 +180,10 @@ bool Configuration::is_realtime_add_trip_enabled() const {
 
 int Configuration::kirin_timeout() const {
     return this->vm["GENERAL.kirin_timeout"].as<int>();
+}
+
+int Configuration::core_file_size_limit() const {
+    return this->vm["GENERAL.core_file_size_limit"].as<int>();
 }
 
 std::string Configuration::broker_host() const {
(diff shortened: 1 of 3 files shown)


kinnou02 (approved):

source/kraken/kraken_zmq.cpp
```diff
@@ -52,6 +55,27 @@ static void show_usage(const std::string& name, const boost::program_options::op
               << "\t KRAKEN_GENERAL_instance_name=<inst_name>" << std::endl;
 }
 
+void set_core_file_size_limit(int max_core_file_size, log4cplus::Logger& logger) {
+    if (max_core_file_size != 0) {
```
benoit-bst, line 59:
```suggestion
    if (max_core_file_size > 0) {
```

source/kraken/configuration.cpp
```diff
@@ -80,6 +80,7 @@ po::options_description get_options_description(const boost::optional<std::strin
 
         ("GENERAL.enable_request_deadline", po::value<bool>()->default_value(true), "enable deadline of request")
         ("GENERAL.metrics_binding", po::value<std::string>(), "IP:PORT to serving metrics in http")
+        ("GENERAL.core_file_size_limit", po::value<int>()->default_value(0), "ulimit that define the maximum size of a core file")
```
xlqian, line 83: can it be size_t by any chance?

xlqian (approved):
:+1:

source/kraken/kraken_zmq.cpp
nonifier: the reason why you might want to have a negative number is to set the max value to infinity.
```
# define RLIM_INFINITY ((__rlim_t) -1)
```

source/kraken/configuration.cpp
nonifier: nope, as you can have `-1` to set the value to `unlimited`

source/kraken/kraken_zmq.cpp
benoit-bst: ouchh oldies style

benoit-bst (approved):

kinnou02:
Jenkins, test this please.

source/kraken/configuration.cpp
pbougue: So `0` is the system's limit right ?
Can you either comment it or precise in description ?

Maybe also, precise the size is in blocks (from `ulimit -c`) or bytes (from trace below) ?

pbougue (approved):

source/kraken/configuration.cpp
nonifier: nope, `0` is absolutely valid. it tells the system that no core dump should be performed.

pbougue (approved):

PR #2821, opened 2019-06-05
kraken: fix autocomplete scores after rt update

benoit-bst:
Bug refers to https://jira.kisio.org/browse/NAVITIAII-2747

### DIFF
--- a/source/jormungandr/tests/kirin_realtime_tests.py
+++ b/source/jormungandr/tests/kirin_realtime_tests.py
@@ -207,6 +207,27 @@ class TestKirinOnVJDeletion(MockKirinDisruptionsFixture):
 
 
 @dataset(MAIN_ROUTING_TEST_SETTING)
+class TestMainStopAreaWeightFactorWithKirinUpdate(MockKirinDisruptionsFixture):
+    def test_main_stop_area_weight_factor_with_kirin_update(self):
+        response = self.query_region("places?type[]=stop_area&q=stop")
+        places = response['places']
+        assert len(places) == 3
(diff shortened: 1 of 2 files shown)


pbougue (approved):
It looks fine, and as your hand-tests didn't seem to show a time issue :+1: 

Maybe you can add (this PR or another) a test inspired by:
https://github.com/CanalTP/navitia/blob/905c2f39d5d5840360035688a8178f19d76a0630/source/jormungandr/tests/places_tests.py#L212-L225
where you send a disruption like this https://github.com/CanalTP/navitia/blob/9da4f7076aa043af311aae371b804516267c726c/source/jormungandr/tests/kirin_realtime_tests.py#L102
and check that it failed without the fix and works with it.

benoit-bst:
Add simple test to highlight the problem.
Without the fix, the test result after rt update is
```
assert places[0]['id'] == 'stopA'                                                                             
assert places[1]['id'] == 'stopB'                                                                             
assert places[2]['id'] == 'stopC'  
```
The same thing before the update but now, we do the  request with **_main_stop_area_weight_factor=5**.
We should have
```
assert places[0]['id'] == 'stopC'                                                                             
assert places[1]['id'] == 'stopA'                                                                             
assert places[2]['id'] == 'stopB'  
```
to take into account of the main stop area weight... because stopC is the only priority SA.
https://github.com/CanalTP/navitia/blob/dev/source/routing/tests/routing_api_test_data.h#L511

Now it is fix.

pbougue (approved):

kinnou02 (approved):

PR #2828, opened 2019-06-14
Doc: cost type modified to string

kadhikari:
https://jira.kisio.org/browse/NAVITIAII-2775

### DIFF
--- a/documentation/slate/source/includes/apis.md
+++ b/documentation/slate/source/includes/apis.md
@@ -1170,8 +1170,8 @@ direction       | int                    | Angle (in degree) between the previou
 
 |Field|Type|Description|
 |-----|----|-----------|
-|value|string|cost formatted as string|
-|currency|string|currency|
+|value|string|cost: float formatted as string|
+|currency|string|currency as specified in input data|
 
 #### Ticket
 
--- a/documentation/slate/source/includes/apis.md
+++ b/documentation/slate/source/includes/apis.md
@@ -1170,7 +1170,7 @@ direction       | int                    | Angle (in degree) between the previou
 
 |Field|Type|Description|
 |-----|----|-----------|
-|value|float|cost|
+|value|string|cost formatted as string|
 |currency|string|currency|
 
 #### Ticket


pbougue (approved):

documentation/slate/source/includes/apis.md
```diff
@@ -1170,7 +1170,7 @@ direction       | int                    | Angle (in degree) between the previou
 
 |Field|Type|Description|
 |-----|----|-----------|
-|value|float|cost|
+|value|string|cost formatted as string|
```
mbounabi: Isn't it necessary to mention that it is a 'float' cast into a 'string' ?

pbougue: We can. The current proposal is ambiguous enough to have integrator check the "real" response before coding :smile: (that's why I approved it).

pbougue:
We also need to decide if we do the last comment on currency in this PR : https://jira.kisio.org/browse/NAVITIAII-2775?focusedCommentId=294427&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-294427
(and if yes : who?)

documentation/slate/source/includes/apis.md
mbounabi: I'm just wondering what will happen if the integrator tries to cast the string in an 'int'

mbounabi: After discussion: 'float' needs to be specified
```suggestion
|value|string|cost: float formatted as a string|
```

documentation/slate/source/includes/apis.md
```diff
@@ -1170,7 +1170,7 @@ direction       | int                    | Angle (in degree) between the previou
...
 |currency|string|currency|
```
mbounabi:
```suggestion
|currency|string|currency as specified in input data
```

documentation/slate/source/includes/apis.md
mbounabi:
```suggestion
|value|string|cost: float formatted as string|
```

pbougue (approved):

mbounabi (approved):
:1st_place_medal:

PR #2833, opened 2019-06-19
Jormun: handle multiple SAE lines for a navitia line in sytralRT

kinnou02:
Data import for sytral is being updated: routes and lines will be
merged. A few navitia lines will map to more than one line in the sytral
SAE. This PR implements the merged of realtime schedules for all of this
lines.
fix: https://jira.kisio.org/browse/NAVP-1356


AurelienLP (approved):

pbougue (approved):
Shouldn't we do the switch for all the use of `fetch_line_id` also?

kinnou02:
> Shouldn't we do the switch for all the use of `fetch_line_id` also?

And take the risk to have duplicate or false schedules?

PR #2834, opened 2019-06-19
[Tyr] API v1

mbounabi:
https://jira.kisio.org/browse/NAVP-1339

/v1 implemented for:
- /api
- /users
- /users/<user_id>
- /users/<user_id>/keys
- /users/<user_id>/keys/<key_id>
- /users/<user_id>/authorizations
- /instances
- /instances/<instance_id>
- /equipments_providers
- /equipments_providers/<equipment_id>

Pagination for:
- /users

Tests for each methods of the endpoints

As discussed with the architect: No decorator!

### DIFF
--- a/source/navitiacommon/navitiacommon/models/equipments_providers.py
+++ b/source/navitiacommon/navitiacommon/models/equipments_providers.py
@@ -47,7 +47,7 @@ class EquipmentsProvider(db.Model, TimestampMixin):  # type: ignore
             self.from_json(json)
 
     def from_json(self, json):
-        self.klass = json['class']
+        self.klass = json['klass']
         self.args = json['args']
         self.discarded = json['discarded'] if 'discarded' in json else self.discarded
 
--- a/source/tyr/tyr/api.py
+++ b/source/tyr/tyr/api.py
@@ -54,7 +54,7 @@ api.add_resource(
     '/v<int:version>/users/<int:user_id>/keys/<int:key_id>/',
 )
 
-api.add_resource(resources.Authorization, '/v0/users/<int:user_id>/authorizations/')
+api.add_resource(resources.Authorization, '/v<int:version>/users/<int:user_id>/authorizations/')
 
 api.add_resource(resources.Index, '/')
 
--- a/source/tyr/tyr/api.py
+++ b/source/tyr/tyr/api.py
@@ -46,7 +46,7 @@ api.add_resource(
 
 api.add_resource(resources.Api, '/v<int:version>/api/')
 
-api.add_resource(resources.User, '/v0/users/', '/v0/users/<int:user_id>/')
+api.add_resource(resources.User, '/v<int:version>/users/', '/v<int:version>/users/<int:user_id>/')
 
 api.add_resource(resources.Key, '/v0/users/<int:user_id>/keys/', '/v0/users/<int:user_id>/keys/<int:key_id>/')
 
--- a/source/tyr/tyr/api.py
+++ b/source/tyr/tyr/api.py
@@ -37,7 +37,12 @@ import flask_restful
 flask_restful.representations.json.settings = {'indent': 4}
 api.app.url_map.strict_slashes = False
 
-api.add_resource(resources.Instance, '/v0/instances/', '/v0/instances/<int:id>/', '/v0/instances/<string:name>/')
+api.add_resource(
+    resources.Instance,
+    '/v<int:version>/instances/',
+    '/v<int:version>/instances/<int:id>/',
+    '/v<int:version>/instances/<string:name>/',
+)
 
 api.add_resource(resources.Api, '/v0/api/')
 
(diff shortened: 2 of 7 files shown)


source/tyr/tyr/resources.py
```diff
@@ -704,34 +711,52 @@ def get(self, user_id=None):
...
+                pagination = models.User.query.filter_by(**filter_params).paginate(
+                    args['page'], current_app.config.get('MAX_ITEMS_PER_PAGE', 5)
+                )
+                pagination_json = {
+                    'current_page': pagination.page,
+                    'items_per_page': pagination.per_page,
+                    'total_items': pagination.total,
+                }
+                if pagination.has_next:
+                    pagination_json['next'] = request.base_url + "?page={}".format(pagination.next_num)
```
kinnou02: you should use [`url_for()`](http://flask.pocoo.org/docs/1.0/quickstart/#url-building)

source/tyr/tyr/resources.py
```diff
@@ -704,34 +711,52 @@ def get(self, user_id=None):
...
+                return marshal(pagination.items, user_fields), pagination_json
```
kinnou02, line 749: this look strange, usually the second element of the tuple is the response status

source/tyr/tyr/resources.py
```diff
@@ -943,23 +971,24 @@ def delete(self, user_id):
...
-    def get(self, user_id, key_id=None):
+    def get(self, user_id, key_id=None, version=0):
         try:
-            return models.User.query.get_or_404(user_id).keys.all()
+            resp = models.User.query.get_or_404(user_id).keys.all()
         except Exception:
             logging.exception("fail")
             raise
+        if version == 1:
+            return {'keys': marshal(resp, key_fields)}
```
kinnou02: minor: You may want to call marshal only one time to avoid the need to update two lines if we want to change the marshaling later.
#DRY

source/tyr/tyr/api.py
```diff
@@ -88,7 +97,9 @@
 api.add_resource(resources.BssProvider, '/v0/bss_providers', '/v0/bss_providers/<string:id>')
 
 api.add_resource(
-    resources.EquipmentsProvider, '/v0/equipments_providers', '/v0/equipments_providers/<string:id>'
```
kinnou02, line 91: For endpoint that aren't updated I would define new paths directly to not pollute the resources with this version thing

source/navitiacommon/navitiacommon/models/equipments_providers.py
```diff
@@ -47,7 +47,7 @@ def __init__(self, id, json=None):
             self.from_json(json)
 
     def from_json(self, json):
-        self.klass = json['class']
+        self.klass = json['klass']
```
kinnou02, line 50: this will pose some compatibility issue with tyr-configurator

mbounabi: This is actually to be compatible with tyr-configurator:
https://github.com/CanalTP/tyr-configurator/pull/18

source/tyr/tyr/resources.py
mbounabi: This is an internal function, not responding to the request directly.

source/tyr/tyr/api.py
mbounabi: I would say the contrary. For instance, this endpoint isn't affected at all by the version in the resource code

source/tyr/tyr/resources.py
```diff
@@ -1890,7 +1927,7 @@ def delete(self, id=None):
 
 class EquipmentsProvider(flask_restful.Resource):
     @marshal_with(equipment_provider_list_fields)
-    def get(self, id=None):
+    def get(self, version=0, id=None):
```
pbougue, line 1940: minor: then you may use "`_`version" (meaning you don't use it)

source/tyr/tyr/resources.py
```diff
@@ -704,34 +711,52 @@ def get(self, user_id=None):
...
 
-                if filter_params:
-                    users = models.User.query.filter_by(**filter_params).all()
-                    return marshal(users, user_fields)
-                else:
-                    users = models.User.query.all()
-                    return marshal(users, user_fields)
-
-    def post(self):
+            if version == 1:
```
pbougue, line 736: This increases cyclomatic complexity to deal with all in the same file.
Maybe it's too costy, but I would like more a "dry" version of abstract-resources, inherited (and reformated) by v0-resources and v1-resources.
That way if we ever want to unplug v0 it should be easier just remove v0-resources file)

> a bit like what was done for COTS and IRE, even if Kirin is not a model on cyclomatic complexity :wink:

pbougue (commented):
Super fine tests :+1:

source/tyr/tyr/resources.py
mbounabi: To be discussed: as it 's the second time this comment comes up, I'll be fine with it

kinnou02 (approved):

pbougue (approved):

PR #2835, opened 2019-06-19
Jormun: mypy doesn't like protobuf types

kinnou02:
we should use https://github.com/dropbox/mypy-protobuf but the last
release has some bug that have been fixed since then, we need a new
release


pbougue (approved):
:disappointed:

benoit-bst (approved):

kinnou02:
Jenkins, test this please.

PR #2836, opened 2019-06-19
fusio2ed: stop_id is supported into admin_stations.txt

benoit-bst:
As @prhod said, rules are simple :
- if `stop_id` column is not present, we continue to read `station_id` (For retrocompatibility)
- if `stop_id` column is present, we read it

*station_id* refers to external code, a `NAV-I id`
*stop_id* is a pur `NAV-II id`

Later, when data team will be ready, an other PR will suppress the retrocompatibility patch
PR -> https://github.com/CanalTP/navitia/pull/2857

JIRA : https://jira.kisio.org/browse/NAVP-1285

### DIFF
--- a/fixtures/ed/ntfs/admin_stations.txt
+++ b/fixtures/ed/ntfs/admin_stations.txt
@@ -1,2 +1,2 @@
-admin_id,admin_name,station_id,station_name,stop_id,stop_name
-admin:A,"admin A",A,"Arrêt A",SA:A,"Arrêt A"
+admin_id,admin_name,station_id,station_name
+admin:A,"admin:A_name_that_is_no_read",A,"station_name_A_that_is_no_read"
--- a/fixtures/ed/ntfs/admin_stations.txt
+++ b/fixtures/ed/ntfs/admin_stations.txt
@@ -1,2 +1,2 @@
 admin_id,admin_name,station_id,station_name
-admin:A,"admin:A_name_that_is_no_read",A,"station_name_A_that_is_no_read"
+admin:A,"admin_name_A_that_is_no_read",A,"station_name_A_that_is_no_read"
--- a/fixtures/ed/ntfs/admin_stations.txt
+++ b/fixtures/ed/ntfs/admin_stations.txt
@@ -1,2 +1,2 @@
-admin_id,admin_name,station_id,station_name
-admin:A,"admin_name_A_that_is_no_read",A,"station_name_A_that_is_no_read"
+admin_id,admin_name,station_id,station_name,stop_id,stop_name
+admin:A,"admin_name_A_that_is_not_read",station_id_A_that_is_not_read,"station_name_A_that_is_not_read",SA:A,"stop_name_A_that_is_not_read"
--- a/fixtures/ed/ntfs/admin_stations.txt
+++ b/fixtures/ed/ntfs/admin_stations.txt
@@ -1,0 +1,2 @@
+admin_id,admin_name,station_id,station_name,stop_id,stop_name
+admin:A,"admin A",A,"Arrêt A",SA:A,"Arrêt A"
--- a/fixtures/ed/ntfs_dst/admin_stations.txt
+++ b/fixtures/ed/ntfs_dst/admin_stations.txt
@@ -1,0 +1,1 @@
+admin_id,admin_name,station_id,station_name,stop_id,stop_name
--- a/fixtures/ed/ntfs_dst/admin_stations.txt
+++ b/fixtures/ed/ntfs_dst/admin_stations.txt
@@ -1,1 +1,2 @@
-admin_id,admin_name,station_id,station_name,stop_id,stop_name
+admin_id,admin_name,stop_id,stop_name
+admin:A,"admin_name_A_that_is_not_read",SCF:SP:SPOCENoctilien87976902,"stop_name_A_that_is_not_read"
--- a/fixtures/ed/ntfs_dst/admin_stations.txt
+++ b/fixtures/ed/ntfs_dst/admin_stations.txt
@@ -1,1 +1,2 @@
-admin_id,admin_name,station_id,station_name,stop_id,stop_name
+admin_id,admin_name,stop_id,stop_name
+admin_id_1,"admin name 1",SCF:SP:SPOCENoctilien87976902,"Brunoy-Wittlich"
--- a/fixtures/ed/ntfs_dst/admin_stations.txt
+++ b/fixtures/ed/ntfs_dst/admin_stations.txt
@@ -1,2 +1,0 @@
-admin_id,admin_name,stop_id,stop_name
-admin_id_1,"admin name 1",SCF:SP:SPOCENoctilien87976902,"Brunoy-Wittlich"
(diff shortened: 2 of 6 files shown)


benoit-bst:
Jenkins, test this please !

source/ed/connectors/fusio_parser.h
```diff
@@ -44,6 +44,8 @@ www.navitia.io
 namespace ed {
 namespace connectors {
 
+static const int unknown_column = -1;
```
pbougue: For consistency, do you plan (not this PR I guess) on extending the use of it? As the same check is done on other places.
We try to use UPPER_CASE for static var:
```suggestion
static const int UNKNOWN_COLUMN = -1;
```

fixtures/ed/ntfs/admin_stations.txt
```diff
@@ -0,0 +1,2 @@
+admin_id,admin_name,station_id,station_name,stop_id,stop_name
+admin:A,"admin A",A,"Arrêt A",SA:A,"Arrêt A"
```
pbougue: You can put different `name`s to check you don't read them (I guess it's the one from stops.txt that's picked):
```suggestion
admin:A,"admin A",A,"Toto A",SA:A,"Tata A"
```
Or maybe just remove the `assert` from the test?

> Same remark for the `ntfs_dst` file

source/ed/connectors/fusio_parser.cpp
```diff
@@ -1545,14 +1545,22 @@ void GridCalendarTripExceptionDatesFusioHandler::handle_line(Data&, const csv_ro
...
-                        tmp_stop_area_map[external_code] = stop_area->second;
+
+    stop_area_c = csv.get_pos_col("stop_id");
+
+    // For retro compatibity
+    // TODO : to remove after the data team update, it will become useless (NAVP-1285)
+    if (stop_area_c == unknown_column) {
+        stop_id_is_present = false;
+        stop_area_c = csv.get_pos_col("station_id");
+        for (const auto& object_code_map : data.object_codes) {
```
pbougue, line 1556: I don't think we only want to associate object_codes when there is no `stop_id` column.

source/ed/connectors/fusio_parser.cpp
```diff
@@ -1566,10 +1574,23 @@ void AdminStopAreaFusioHandler::handle_line(Data& data, const csv_row& row, bool
...
+    // TODO : to remove after the data team update, it will become useless (NAVP-1285)
+    if (!stop_id_is_present) {
+        sa_it = tmp_stop_area_map.find(row[stop_area_c]);
+        if (sa_it == tmp_stop_area_map.end()) {
+            LOG4CPLUS_ERROR(logger,
+                            "AdminStopAreaFusioHandler : Impossible to find the stop_area " << row[stop_area_c]);
+            return;
+        }
+    } else {
+        sa_it = gtfs_data.stop_area_map.find(row[stop_area_c]);
```
pbougue, line 1588: Looks like the exact same code twice (if/else).
You probably don't need the `stop_id_is_present` attribute.

source/ed/connectors/fusio_parser.cpp
benoit-bst: Actually, when there is no `stop_id` column, we want to read `station_id` for retrocompatibility. 
Now, we have this code that works with `station_id` column. I just skiped this part because it works for `NAV 1 id`. 
It is a preprocessing to create a temporary stop_area list (tmp_stop_area_map)

In any case, that's what I understand

source/ed/connectors/fusio_parser.cpp
benoit-bst: if `! stop_id_is_present` we process with *tmp_stop_area_map*, not *gtfs_data.stop_area_map*

But it is temporary. After the data team update, it will remain the `else` part

source/ed/connectors/fusio_parser.h
benoit-bst: Both are good for me :smiley: 
Both exists in the current code

fixtures/ed/ntfs/admin_stations.txt
benoit-bst: Good catch :tada: 
It turns out, we read the name into `stops.txt` (true story, good feeling @pbougue). In fact, we only read the `id` here. 
So, I'll update the test to highlight this behavior.

source/ed/connectors/fusio_parser.cpp
benoit-bst: I updated the PR's header with id info

pbougue (approved):
:+1: Now that I understood the external_code ;)
Good to change the test but that's minor.
Same for upper_case, if anyone wants to vote they can weigh for one desicion, but it's minor.

fixtures/ed/ntfs_dst/admin_stations.txt
```diff
@@ -1 +0,0 @@
-admin_id,admin_name,station_id,station_name,stop_id,stop_name
```
pbougue, line 1: To me the point is to be able to read that file in the end (well with at least the one line you previously added).
So I think you should keep your test over it also, no?

fixtures/ed/ntfs/admin_stations.txt
```diff
@@ -0,0 +1,2 @@
+admin_id,admin_name,station_id,station_name
+admin:A,"admin_name_A_that_is_no_read",A,"station_name_A_that_is_no_read"
```
pbougue: :+1: and this test will eventually be removed when navitia stops reading station_id

pbougue (commented):
Not sure about one of the test-removal you did.
:point_down:

source/ed/connectors/fusio_parser.cpp
```diff
@@ -1566,10 +1574,23 @@ void AdminStopAreaFusioHandler::handle_line(Data& data, const csv_row& row, bool
...
+        if (sa_it == tmp_stop_area_map.end()) {
```
kinnou02:
```suggestion
        if (sa_it == gtfs_data.stop_area_map.end()) {
```

benoit-bst: Nice catch. :speak_no_evil: :+1:

fixtures/ed/ntfs_dst/admin_stations.txt
benoit-bst: At first, I started to implement my test with `ntfs_dst` data set. Then, I realized that it didn't have `SA` inside `stops.txt`... But we need it to retrieve stop_name.

My future changes will concern `ntfs` and `ntfs_v5` data set, because both are already SA inside stops.txt.
The first test will use stations_id and the other the stop_id.

In a near future, I will open a new PR to erase the retrocompatibility and suppress the test with `station_id`.

About the empty `admin_stations.txt` file inside ntfs_dst, I thought it was useless, but finally I think it is a good way to show, that the file can exist inside the data set, even if it is empty (Good for newbies).
So, I prefer to let it. 
Are you agree with that @pbougue ?

fixtures/ed/ntfs/admin_stations.txt
benoit-bst: Yes indeed

fixtures/ed/ntfs_dst/admin_stations.txt
pbougue: Agreed :+1:

kinnou02 (approved):

pbougue (approved):

benoit-bst:
Add extra test for nominal case (with only stop_id)

pbougue (approved):

PR #2837, opened 2019-06-21
All: Update Mypy and Pre-commit to check all projects in Python

GuillaumeLescure:
Typecheck across Python modules is not well supported. Preliminary tests show that Mypy doesn't check types if they're shared among modules. But for "simple" checks, it works well and allow us to provide some improvements.

Mypy seems to have many limitations and an harder work in the futur may be done to improve its effectiveness.

Also, this patch allow each contributor to run pre-commit with any version of python3. It improves portability for internal and external builds.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -3,7 +3,7 @@ repos:
     rev: 18.9b0
     hooks:
     - id: black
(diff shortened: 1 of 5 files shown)


GuillaumeLescure:
An other PR is created for that.

PR #2838, opened 2019-06-24
All: Use mypy for every projects

GuillaumeLescure:


GuillaumeLescure:
Expalantions about the exclude paths :
* /monitor/ is dead code
* /third_party/ we cannot modify third party code so it's not relevant to check it
* /tests/ not relevant to check tests (questionable but for now it's ok)
* /sql/ not relevant to check sql (questionable but for now it's ok)
* env.py and setup.py should be checked but mypy return a false-positif error (duplicate module)

GuillaumeLescure:
More informations about mypy :

* type checking only works when the inline comment is done, ex : # type: () -> None
(for now, it's not done in all codes yet)

* type check only works if the function has the type check comment, ex :
```
def run(): # type: () -> None
   x = "ok" # type: typeUnknown
   print "jormun", type(x)
   greet_world(x)
```
✘ $ error: Name 'typeUnknown' is not defined (edited)

```
def run():
  x = "ok" # type: typeUnknown
  print "jormun", type(x)
  greet_world(x)
```
✔ $ all good !

* some files seems to be ignored if the file "\_\_init\_\_.py" is present or missing, ex :
A type error is introduced in Jormun but the function is defined in Eitri:

| \_\_init.py\_\_ in Jormun | \_\_init.py\_\_ in Eitri | Result |
| --- | --- | --- |
| Present | Present | ✔ : no error found |
| Present | Missing | ✘ : error found |
| Missing | Present | ✔ : no error found |
| Missing | Missing | ✘ : error found |

The archive with the tests : https://github.com/CanalTP/navitia/files/3320974/test_mypy.zip

source/eitri/ed_handler.py
```diff
@@ -30,7 +30,6 @@
 import glob
 import os
 from navitiacommon import utils, launch_exec
-from navitiacommon.launch_exec import launch_exec
```
kinnou02, line 33: Are you sure that eitri still works after that?

kinnou02 (commented):
`monitor` isn't dead code, it's deployed on production and used by fabric and the monitoring platform

source/eitri/ed_handler.py
GuillaumeLescure: You're right I have this error :
'module' object is not callable

I'm investigating ...

GuillaumeLescure: Fix done.

kinnou02:
Jenkins, test this please.

GuillaumeLescure:
retest this please

kinnou02 (approved):

source/eitri/ed_handler.py
```diff
@@ -30,7 +30,6 @@
 import glob
 import os
 from navitiacommon import utils, launch_exec
```
mbounabi, line 32: from utils, only 1 function used, from launch_exec, only one function used. Not sure about performance (I'll try to do some tests or find doc about it), but I'm not a fan of 'launch_exec.launch_exec'
```suggestion
from navitiacommon.utils import type_of_data 
from navitiacommon.launch_exec import launch_exec
```

mbounabi (approved):

nonifier (approved):

PR #2839, opened 2019-06-24
[Jormun] Adding street_network_configuraton in db

AurelienLP:
https://jira.kisio.org/browse/TRR-52

We don't want to use the json in Jormun configuration to choose what back-end we want to use for each mode (Asgard for car, Kraken for walking, bike, bss, taxi, ridesharing for example) anymore.

We would like to use Tyr and Tyr-Configurator.

To do so, we have to :

1. Add a column in instances for each mode;
2. Create a new table streetnetwork_backend containing the configuration of all the different backends we'll want to use;
3. Create a relation between the values in instances and the configuration in streetnetwork_backend (a foreign key)

### DIFF
--- a/source/navitiacommon/navitiacommon/models/streetnetwork_backend.py
+++ b/source/navitiacommon/navitiacommon/models/streetnetwork_backend.py
@@ -36,6 +36,7 @@ from navitiacommon.models import db, TimestampMixin
 
 
 class StreetNetworkBackend(db.Model, TimestampMixin):  # type: ignore
+    __tablename__ = 'streetnetwork_backend'
     id = db.Column(db.Text, primary_key=True)
     klass = db.Column(db.Text, unique=False, nullable=False)
     discarded = db.Column(db.Boolean, nullable=False, default=False)
@@ -47,7 +48,7 @@ class StreetNetworkBackend(db.Model, TimestampMixin):  # type: ignore
             self.from_json(json)
 
     def from_json(self, json):
-        self.klass = json['class']
+        self.klass = json['klass']
         self.args = json['args']
         self.discarded = json['discarded'] if 'discarded' in json else self.discarded
 
--- a/source/navitiacommon/navitiacommon/models/streetnetwork_backend.py
+++ b/source/navitiacommon/navitiacommon/models/streetnetwork_backend.py
@@ -35,7 +35,7 @@ from sqlalchemy.dialects.postgresql.json import JSONB
 from navitiacommon.models import db, TimestampMixin
 
 
-class StreetNetworkConfiguration(db.Model, TimestampMixin):  # type: ignore
+class StreetNetworkBackend(db.Model, TimestampMixin):  # type: ignore
     id = db.Column(db.Text, primary_key=True)
     klass = db.Column(db.Text, unique=False, nullable=False)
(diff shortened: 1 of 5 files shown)


source/tyr/migrations/versions/57f52a929056_street_network_configuration_db.py
```diff
@@ -0,0 +1,60 @@
...
+        sa.Column('updated_at', sa.DateTime(), nullable=True),
+        sa.Column('id', sa.Text(), nullable=False),
+        sa.Column('klass', sa.Text(), nullable=False),
+        sa.Column('discarded', sa.Boolean(), nullable=False),
+        sa.Column('args', postgresql.JSONB(), server_default='{}', nullable=True),
+        sa.PrimaryKeyConstraint('id'),
+    )
+
+    op.add_column(
+        'instance', sa.Column('street_network_car', sa.Text(), nullable=False, server_default='kraken')
```
kinnou02, line 54: You have to add a foreignkey to the `street_network_configuration `

source/tyr/migrations/versions/57f52a929056_street_network_configuration_db.py
```diff
@@ -0,0 +1,60 @@
...
+down_revision = '38cf24479595'
+
+from alembic import op
+import sqlalchemy as sa
+from sqlalchemy.dialects import postgresql
+
+
+def upgrade():
+    op.create_table(
+        'street_network_configuration',
```
kinnou02: I would prefer `streetnetwork_configuration` or even better for consistency `streetnetwork_backend`

kinnou02 (changes requested):

source/tyr/migrations/versions/57f52a929056_street_network_configuration_db.py
AurelienLP: I'll go with streetnetwork_backend

source/tyr/migrations/versions/57f52a929056_street_network_configuration_db.py
AurelienLP: Don't we want `street_network_configuration` to be independent like the bss_provider table?

kinnou02: bss providers aren't referenced by any coverage, they live freely.
here we want to have that association

source/navitiacommon/navitiacommon/models/streetnetwork_backend.py
```diff
@@ -0,0 +1,67 @@
...
+    discarded = db.Column(db.Boolean, nullable=False, default=False)
+    args = db.Column(JSONB, server_default='{}')
+
+    def __init__(self, id, json=None):
+        self.id = id
+        if json:
+            self.from_json(json)
+
+    def from_json(self, json):
+        self.klass = json['class']
```
mbounabi:
```suggestion
        self.klass = json['klass']
```

This will cause some problems if a task streetnetwork_backend_diff is to be created in tyr-configurator.

kinnou02 (approved):

kadhikari (approved):

mbounabi (approved):

PR #2842, opened 2019-06-27
Add PyDoc for all Eitri project

GuillaumeLescure:


source/eitri/ed_handler.py
```diff
@@ -35,14 +35,22 @@
 import zipfile
 import logging
 
+"""
+This module contains all the functions to prepare the job, to call the binaries to create data (all the "*2ed" binaries),
```
nonifier:
```suggestion
This module contains all the functions to prepare a job, to call the binaries that ingest input data (all the "*2ed" binaries),
```

source/eitri/ed_handler.py
```diff
@@ -35,14 +35,22 @@
...
+to upload it in the database and then call "ed2nav" to extract it in a single ".nav.lz4" file.
```
nonifier:
```suggestion
into a database and then call "ed2nav" to produce a single ".nav.lz4" file.
```

source/eitri/ed_handler.py
```diff
@@ -150,7 +188,17 @@ def generate_nav(
     data_dir, docker_ed, docker_cities, output_file, ed_component_path, cities_exec_path, import_cities
 ):
     """
-    load all data either directly in data_dir if there is no sub dir, or all data in the subdir
+    Load all data from a directory to an single output file.
+    If can load all the data directly from the directory or in each sub-directories for each data kind.
```
nonifier:
```suggestion
    It can load all the data directly from the directory or in each sub-directories for each data kind.
```

source/eitri/eitri.py
```diff
@@ -34,9 +34,11 @@
 import logging
 
 """
-Nav generator
+Eitri use a set of binaries to extract data, load it in database and extract it to a single file.
```
nonifier:
```suggestion
Eitri uses a set of binaries to extract data, load it into a database and extract it to a single file.
```

source/eitri/eitri.py
```diff
@@ -52,17 +54,16 @@ def eitri(
...
     """
-    Generate a data.nav.lz4 file
+    Generate a "data.nav.lz4" file.
 
-    :param data_dir: directory with data. if several dataset (osm/gtfs/...) are available, they need to be in separate directory
-    :param output_file: output data.nav.lz4 file path
-    :param cities_exec_path: optional, where to find the command 'cities'
-    :param import_cities: optional, cities files ex:france_boundaries.osm.pbf,
-                          if given, cities will be loaded
+    :param data_dir: the directory with all data. If several dataset ("*.osm", "*.gtfs", ...) are available, they need to be in separate directories
```
nonifier:
```suggestion
    :param data_dir: the directory with all data. If several datasets ("*.osm", "*.gtfs", ...) are available, they need to be in separate directories
```

nonifier (approved):

source/eitri/docker_wrapper.py
```diff
@@ -33,28 +33,53 @@
...
+    Class to store and manipulate database parameters.
+    """
+
     def __init__(self, host, dbname, user, password):
+        """
+        Constructor of DbParams.
+
+        :param host: the host name
+        :param dbname: the database name
+        :param user: the user name to sue for the database
```
mbounabi: someone will be sued?

source/eitri/docker_wrapper.py
```diff
@@ -33,28 +33,53 @@
...
+        :param password: the password of the user
+        :return:
```
mbounabi: If it doesn't return anything, why not remove this line? Same for every other "return" alone in the file?

source/eitri/docker_wrapper.py
```diff
@@ -33,28 +33,53 @@
...
 
     def old_school_cnx_string(self):
         """
-        old C++ component does not support the classic postgres://...
+        The connection string for the database (old school version).
+        A string containg all the essentials data for a connection to a old school database (before Debian8). Old C++
+        component does not support the classic "postgres://...".
 
-         refactor when moving to debian 8 as the classic cnx string will be recognized
+        :return: the old school connection string
```
mbounabi, line 81: what is "old school" ?

source/eitri/docker_wrapper.py
GuillaumeLescure: I thought it was convention to always write the return for mypy ... but if I'm wrong, tell me, I will remove it everywhere it's not needed.

source/eitri/docker_wrapper.py
GuillaumeLescure: Before Debian8 apparently ^^
You prefere I change this with something more accurate ?

source/eitri/docker_wrapper.py
GuillaumeLescure: ^^
Done

AurelienLP (approved):

PR #2845, opened 2019-06-28
Add mypy type check in Eitri

GuillaumeLescure:


GuillaumeLescure:
Jenkins, test this please.

source/eitri/docker_wrapper.py
```diff
@@ -63,12 +90,16 @@ def old_school_cnx_string(self):
...
 
     def __init__(self):
-        log = logging.getLogger(__name__)
-        base_url = 'unix://var/run/docker.sock'
+        # type: () -> None
+        """
+        Constructor of PostgresDocker.
+        """
+        log = logging.getLogger(__name__)  # type: logging.Logger
+        base_url = 'unix://var/run/docker.sock'  # type: str
```
mbounabi: Is it useful in this case? Clearly, it's a string!

source/eitri/docker_wrapper.py
```diff
@@ -1,4 +1,3 @@
-# coding=utf-8
```
mbounabi: Are you sure about removing this header?

source/eitri/requirements.txt
```diff
@@ -7,3 +7,4 @@ GeoAlchemy2==0.2.4
 SQLAlchemy==1.3.3
 alembic==0.6.7
 requests==2.20.0
+typing
```
AurelienLP: Don't we have to specify a version here too ?

source/eitri/docker_wrapper.py
GuillaumeLescure: There is no ut8 character in this file so either we use it when we need it or we put it everywhere. Do you concur ?

source/eitri/docker_wrapper.py
GuillaumeLescure: If you want to use mypy you have to tell him what is what ;)
This comment allow him to know base_url is a string and then, when you use it in a function (that also have to comment for mypy) it will detect an error type.

source/eitri/requirements.txt
GuillaumeLescure: You're right, I will do it.

source/eitri/docker_wrapper.py
mbounabi: I know that. The way I see it, mypy helps us to have the type needed in a variable passed to a function or returned by a function. Here, the variable is set directly, it can't be anything else, so the gain offered by mypy is quite limited IMO. 
We'll call a referee :checkered_flag:

benoit-bst (approved):

source/eitri/docker_wrapper.py
nonifier: I would rather keep the utf-8 coding definition. 
There might be no utf8 char at the moment, but the entire python codebase is treated as such. What would happened if someone adds a utf8 char in the future ?

source/eitri/docker_wrapper.py
nonifier: I would agree with @mbounabi, there is very little benefits documenting an initialised variable like this one, which clearly is a `string`.

What mypy error do you actually have? would `base_url = str('unix://var/run/docker.sock')` help?

source/eitri/docker_wrapper.py
GuillaumeLescure: Ok done

source/eitri/docker_wrapper.py
GuillaumeLescure: My bad, it's not needed for strings.

xlqian (approved):

source/eitri/eitri.py
```diff
@@ -1,4 +1,5 @@
-#!python
```
nonifier, line 1: Are you sure about removing this ? 
I'm afraid that removing this will break the binary packaging/integration.

GuillaumeLescure: For what I tested the shebang doesn't work.
I have to run it with "python ./eitri" or "python2 ./eitri", I re-add an other shebang in #2852

nonifier (approved):

PR #2846, opened 2019-06-28
kraken: Fixing undefined behaviour and crash of A*

AurelienLP:
This PR MUST be merged AFTER https://github.com/CanalTP/navitia/pull/2843.

Only the commit "Fixing A* init by checking if starting edge is found" needs to be reviewed since the first one is the un-revert of the A*.

During A* init, we initialize costs vector for the starting edge. However, if this edge does not exist, its index is either undefined or -1 (max of std::size_t) creating crash or undefined behavior.

We fixed that checking if starting.edge is found. We also access cost vector with "at" and not the operator "[]" which checks the index and the bound of the vector, throwing an exception instead of crashing.

JIRA: https://jira.kisio.org/browse/NAVITIAII-2761


pbougue (approved):
Just reviewed last commit.

We may as well just keep this PR and not #2843 to avoid remerging unfixed A*

As mentioned orally :
1. It can be worth running artemis + artemis_idfm on that (let you decide)
2. Just curious of impact on perf if you know it (probably not worth rerunning bench just for that)

AurelienLP:
1. Currently building the PR to run artemis + artemis_idfm
2. It should be negligible since we are not in a loop

kinnou02 (approved):

source/georef/astar_path_finder.cpp
```diff
@@ -49,10 +49,12 @@ void AstarPathFinder::init(const type::GeographicalCoord& start_coord,
     size_t n = boost::num_vertices(geo_ref.graph);
     costs.assign(n, bt::pos_infin);
 
-    costs[starting_edge[source_e]] =
-        compute_cost_from_starting_edge_to_dist(starting_edge[source_e], dest_projected_coord);
-    costs[starting_edge[target_e]] =
-        compute_cost_from_starting_edge_to_dist(starting_edge[target_e], dest_projected_coord);
+    if (starting_edge.found) {
```
nonifier, line 52: This is a good fix, but it doesn't prevent us from doing the same mistake ever again.....
I would be in favour of raising an exception in `ProjectionData::operator[]` if `found` isn't set. 
Or you could change the way `ProjectionData::vertices` is implemented and make sure it's always in a state that will not bring you to Undefined Behavior Land.

nonifier (changes requested):

source/georef/astar_path_finder.cpp
AurelienLP: Just updated my PR to throw an exception in `ProjectionData::operator[]` when `found` is false ! :D

nonifier (approved):

pbougue:
Not sure: Is it possible to add a test that would trigger the bug before merge?
I think of 3 "situations":
1. before any fix : doesn't crash (or just using sanitizer) : return wrong result?
2. using `.at()` crashes
3. after all fix : doesn't crash, return no-result as expected?

It may be nice to have the "functional case" that triggered the bug pinned in a test.

AurelienLP:
I am trying to create an artemis test for this PR. However, I couldn't reproduce the crash without the fix. We can only create a test that returns "no solution found for this journey" when the starting edge is not found (and check that no crash happens). This is what I am trying to do now :)

PR #2849, opened 2019-07-01
Tyr: load open addresses in mimir

antoine-de:
add the capability for tyr to load [open addresses](https://openaddresses.io/) data into mimir.

### DIFF
--- a/source/navitiacommon/navitiacommon/utils.py
+++ b/source/navitiacommon/navitiacommon/utils.py
@@ -33,7 +33,7 @@ import glob
 import logging
 
 street_source_types = ['OSM']
-address_source_types = ['BANO', 'OSM']
+address_source_types = ['BANO', 'OA']
 poi_source_types = ['FUSIO', 'OSM']
 admin_source_types = ['OSM', 'COSMOGONY']
 
--- a/source/tyr/tyr/binarisation.py
+++ b/source/tyr/tyr/binarisation.py
@@ -625,6 +625,46 @@ def bano2mimir(self, autocomplete_instance, filename, job_id, dataset_uid):
 
 
 @celery.task(bind=True)
+def openaddresses2mimir(self, autocomplete_instance, filename, job_id, dataset_uid):
+    """ launch openaddresses2mimir """
+    autocomplete_instance = models.db.session.merge(autocomplete_instance)  # reatache the object
+    logger = get_autocomplete_instance_logger(autocomplete_instance, task_id=job_id)
+    job = models.Job.query.get(job_id)
+    cnx_string = current_app.config['MIMIR_URL']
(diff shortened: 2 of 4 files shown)


source/navitiacommon/navitiacommon/utils.py
```diff
@@ -33,7 +33,7 @@
 import logging
 
 street_source_types = ['OSM']
-address_source_types = ['BANO', 'OSM']
```
nonifier, line 36: `OSM` cannot be used as an address source any more?
Is this breaking coverages with OSM but no OpenAddress data?

nonifier (approved):

source/navitiacommon/navitiacommon/utils.py
antoine-de: we have never been able to load addresses from osm :stuck_out_tongue_winking_eye: 

I think the value has been set a long time ago thinking it would be done in the short term, but since it's not yet the case, I think we can delete the value :smile:

source/tyr/tyr/binarisation.py
```diff
@@ -624,6 +624,46 @@ def bano2mimir(self, autocomplete_instance, filename, job_id, dataset_uid):
...
+                working_directory,
+                '--connection-string',
+                cnx_string,
+                '--dataset',
+                autocomplete_instance.name,
+            ],
+            logger,
+        )
+        if res != 0:
+            # @TODO: exception
```
benoit-bst, line 658: Why it is a TODO ? Because you already raise a valueError.

source/tyr/tyr/tasks.py
```diff
@@ -252,6 +254,10 @@ def files_type(files):
             return 'cosmogony'
         if len(files) == 1 and files[0].endswith('.pbf'):
             return 'osm'
+        # OpenAddresses files does not have a predefined naming,
+        # so we check it last, and consider all csv as OA
+        if any(f for f in files if OPEN_ADDRESSES_REGEXP.match(f)):
```
benoit-bst, line 259: There is no possibility of confusing with a NTFS file ?

kinnou02 (approved):

source/tyr/tyr/tasks.py
antoine-de: nop they are not dropped in the same directory (the method is `type_of_autocomplete_data`)

source/tyr/tyr/binarisation.py
antoine-de: ahah I used the good ol' copy paste method for this, and all the other tasks have this TODO :roll_eyes: :see_no_evil: 

I think this TODO is to better type the raised exception and not use a generic exception

benoit-bst (approved):

PR #2850, opened 2019-07-01
Add colored logs to all Python projects

GuillaumeLescure:
* improve visibility
* standardization of the logs

### DIFF
--- a/source/jormungandr/requirements.txt
+++ b/source/jormungandr/requirements.txt
@@ -41,3 +41,4 @@ jmespath==0.9.3
 shortuuid
 click==6.7
 typing
+coloredlogs==10.0
--- a/source/navitiacommon/navitiacommon/log.py
+++ b/source/navitiacommon/navitiacommon/log.py
@@ -1,0 +1,60 @@
+# Copyright (c) 2001-2014, Canal TP and/or its affiliates. All rights reserved.
+#
+# This file is part of Navitia,
+#     the software to build cool stuff with public transport.
+#
+# Hope you'll enjoy and contribute to this project,
+#     powered by Canal TP (www.canaltp.fr).
+# Help us simplify mobility and open public transport:
+#     a non ending quest to the responsive locomotion way of traveling!
+#
+# LICENCE: This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU Affero General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+# GNU Affero General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with this program. If not, see <http://www.gnu.org/licenses/>.
+#
+# Stay tuned using
+# twitter @navitia
+# IRC #navitia on freenode
+# https://groups.google.com/d/forum/navitia
+# www.navitia.io
+
+import coloredlogs
+
+
+def enable_color(enabled):
+    # type: (bool) -> None
+    """
+    Get and check all the parameters.
+
+    :params enabled: true to enable the color in logs, otherwise false
+    """
+    if enabled:
+        coloredlogs.install(
+            level='DEBUG',
+            fmt='%(asctime)-15s %(name)s %(message)s',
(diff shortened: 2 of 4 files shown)


kinnou02 (commented):
did you check with json log? with syslog? with our custom filter to add `request_id`?

GuillaumeLescure:
Before:
before

After:
after

GuillaumeLescure:
> did you check with json log? with syslog? with our custom filter to add `request_id`?

I didn't try those but I'm note sure to understand why. For instance, those logs don't show in syslog.

But I tried to redirect thoses logs in a file ($ cmd > file) and I have the color markers that polute logs, if it's what you worried about.

GuillaumeLescure:
Ok my bad, I tested with the syslog :
`handler = logging.handlers.SysLogHandler(address = '/dev/log')`
`logger.addHandler(handler)`

And it works, no issue, logs are clear.

GuillaumeLescure:
New version disable color by default but allow any program to enable it easilly.

Default behaviour:
image

AurelienLP:
How do we enable the color ? By editing the code `log.enable_color(False)` to `log.enable_color(True)` ? Won't it be nice to do so during the creation of the cmake project ?

source/jormungandr/requirements.txt
```diff
@@ -41,3 +41,4 @@ jmespath==0.9.3
 shortuuid
 click==6.7
 typing
+coloredlogs
```
kinnou02: Always specify version of your dependencies

source/navitiacommon/navitiacommon/__init__.py
```diff
@@ -26,3 +26,7 @@
 # IRC #navitia on freenode
 # https://groups.google.com/d/forum/navitia
 # www.navitia.io
+
+from navitiacommon import log
+
+log.enable_color(False)
```
kinnou02: I don't see why importing `navitiacommon` should have a side effect on the logging.
Define your function then call it in the projects that need it.

kinnou02 (changes requested):

source/jormungandr/requirements.txt
GuillaumeLescure: Ok, done

source/navitiacommon/navitiacommon/__init__.py
GuillaumeLescure: The idea was to enable this feature for all python projects so, everytime we import "navitiacommon" (I believe it's done by every python project) we set up the log. It only enables the format, colors have to be enabled if the program needs it.

Should I remove it ?

GuillaumeLescure:
> How do we enable the color ? By editing the code `log.enable_color(False)` to `log.enable_color(True)` ? Won't it be nice to do so during the creation of the cmake project ?

Yes that's the API to use to enable or disable the logs.
For the CMake, I prefere to enable/disable it at runtime but if eveyone prefere in the CMake, I can work on it.

AurelienLP:
> > How do we enable the color ? By editing the code `log.enable_color(False)` to `log.enable_color(True)` ? Won't it be nice to do so during the creation of the cmake project ?
> 
> Yes that's the API to use to enable or disable the logs.
> For the CMake, I prefere to enable/disable it at runtime but if eveyone prefere in the CMake, I can work on it.

I don't think it is really that important, modifying python code is much much faster. Don't worry about that, especially if @kinnou02 didn't say otherwise.

source/navitiacommon/navitiacommon/__init__.py
GuillaumeLescure: Removed

AurelienLP (approved):

source/navitiacommon/navitiacommon/log.py
```diff
@@ -0,0 +1,60 @@
...
+# Stay tuned using
+# twitter @navitia
+# IRC #navitia on freenode
+# https://groups.google.com/d/forum/navitia
+# www.navitia.io
+
+import coloredlogs
+
+
+def enable_color(enabled):
```
benoit-bst, line 33: I don't understand where is call this function. Is it not used ?

GuillaumeLescure: In the #2852, source/eitri/eitri.py:145-147.

For now, only Eitri use it. Since it was remove from the init of NavitiaCommon, I will make a patch for Yormungandr and Tyr.

kinnou02 (approved):

benoit-bst (approved):

PR #2722, opened 2019-03-25
Doc: Adding pathways and levels to NTFS specs

prhod:
This PR introduce the pathways [GTFS specification](https://github.com/google/transit/pull/143/files)  to the NTFS definition.

I wondered about splitting the stops.txt file, but ended in keeping all objects in one file. 
Please, feel free to comment.

PR #2851, opened 2019-07-02
update ntfs specification

patochectp:
These changes reflect the way files are actually read.

[skip-ci]


kinnou02:
Why change this? this isn't a backward compatible.

pbougue (changes requested):
It's more consistent, but it requires Navitia to read both in the meantime.
Also @kinnou02 is right, it's a breaking change on the spec.
So if we do that, we have to issue a new version (so archive the previous one, etc.) like we did here : https://github.com/CanalTP/navitia/pull/2722/files

:warning: you orally mentionned that we use external codes to attach those... This may need to be traced/changed

PR #2854, opened 2019-07-08
[Tyr] v1 changes

mbounabi:
After review:
- Add new resource for /v1/users that inherits from resource /v0/users
- Remove useless parameter 'version' from /equipment_reports 
- Return 404 when no response for /v1/instances
- Add recommendation file for API v1

### DIFF
--- a/source/tyr/tests/integration/v1_test.py
+++ b/source/tyr/tests/integration/v1_test.py
@@ -99,7 +99,7 @@ def test_api():
 
 def test_users(create_5_users):
     check_v1_response('users')
-    check_v1_response('user', 'users/{}'.format(create_5_users))
+    check_v1_response('users', 'users/{}'.format(create_5_users))
 
 
 def test_users_methods():
@@ -183,7 +183,7 @@ def test_keys_methods(create_5_users):
 def test_authorization_methods(create_5_users, create_instance, create_api):
     user_id = create_5_users
     full_user = api_get('/v1/users/{}'.format(user_id))
-    assert len(full_user['user']['authorizations']) == 0
+    assert len(full_user['users']['authorizations']) == 0
     auth = {'instance_id': create_instance, 'api_id': create_api}
     resp_post = api_post(
         '/v1/users/{}/authorizations'.format(user_id), data=json.dumps(auth), content_type='application/json'
--- a/source/tyr/tests/integration/v1_test.py
+++ b/source/tyr/tests/integration/v1_test.py
@@ -99,7 +99,7 @@ def test_api():
 
 def test_users(create_5_users):
     check_v1_response('users')
-    check_v1_response('users', 'users/{}'.format(create_5_users))
+    check_v1_response('user', 'users/{}'.format(create_5_users))
 
 
 def test_users_methods():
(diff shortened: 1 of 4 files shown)


kinnou02 (approved):

source/tyr/tyr/resources.py
```diff
@@ -979,6 +961,42 @@ def delete(self, user_id, version=0):
...
 
+class UserV1(User):
+    def _get_args(self, user_id):
+        parser = super(UserV1, self)._get_args(user_id)
+        parser.add_argument('page', type=int, required=False, default=1)
+        return parser
+
+    def _get_all_users(self, args):
+        del args['disable_geojson']
+        # dict comprehension would be better, but it's not in python 2.6
```
pbougue: minor cleanup (not the point of this PR): All is in 2.7, so we can do it now

source/tyr/tests/integration/v1_test.py
```diff
@@ -99,7 +99,7 @@ def test_api():
 
 def test_users(create_5_users):
     check_v1_response('users')
-    check_v1_response('users', 'users/{}'.format(create_5_users))
+    check_v1_response('user', 'users/{}'.format(create_5_users))
```
pbougue: So there was an unspotted "bug" before?
Note for a choice: In Jormungandr, it's plural in that case.

pbougue (approved):
Tastes better to me :melon:

source/tyr/tests/integration/v1_test.py
mbounabi: bug or upgrade, which school is it? We agreed with @kinnou02 that the POST request which returns only 1 user should have a key "user" singular. The request GET /users/user_id also returns only 1 user. So... I changed it to "user" singular. Right or wrong? TBD

mbounabi: What we discussed about the singular is for POST and PUT requests. 
For GET, you're right, it's better to leave the key in the plural for homogeneity, even if there's only one result.

source/tyr/tyr/resources.py
```diff
@@ -987,6 +969,43 @@ def delete(self, user_id, version=0):
...
+            'items_per_page': pagination.per_page,
+            'total_items': pagination.total,
+        }
+        if pagination.has_next:
+            pagination_json['next'] = url_for(request.endpoint, page=pagination.next_num)
+        return marshal(pagination.items, user_fields), pagination_json
+
+    def get(self, user_id=None):
+        resp = super(UserV1, self).get(user_id)
+        if type(resp) == tuple:
```
AurelienLP, line 997: I am a little lost... When is it a tuple ?

mbounabi: That's what's returned above (l.992) if there are several users and it needs pagination. The response received is a tuple of the users list and the pagination object.

AurelienLP: Hoooo, I was looking in the method get from User.... My bad. Could you add a little comment `# If we need pagination` somewhere around here ? Could help blind people like me :D

AurelienLP (approved):

AurelienLP (approved):

kinnou02 (approved):

AurelienLP (approved):

PR #2855, opened 2019-07-09
Fix 'max_extra_second_pass' param from db

pbougue:
"`_`max_extra_second_pass" is stored under "max_extra_second_pass".

But it was used under the wrong name afterward (updating with db value a dict-key never used).

I don't see how to (easily) auto-test it: ideas welcome.
Hand-tested locally : works as expected with url param, from db, and none of those.

JIRA related to that: https://jira.kisio.org/browse/NAVITIAII-2690

__________

:hand: Hand-tests (after this PR):
* Find a /journeys request where `_max_extra_second_pass=999` allows finding more journey (walking at start)
* Change the max_extra_second_pass to 999 in tyr for that coverage (wait a bit so that it's effective)
* Do the same request without the param : it should work the same way

### DIFF
--- a/source/jormungandr/jormungandr/interfaces/v1/Journeys.py
+++ b/source/jormungandr/jormungandr/interfaces/v1/Journeys.py
@@ -329,7 +329,6 @@ class Journeys(JourneyCommon):
         parser_get.add_argument(
             "timeframe_duration",
             type=int,
-            dest="timeframe_duration",
             help="Minimum timeframe to search journeys.\n"
             "For example 'timeframe_duration=3600' will search for all "
             "interesting journeys departing within the next hour.\n"
--- a/source/jormungandr/jormungandr/interfaces/v1/Journeys.py
+++ b/source/jormungandr/jormungandr/interfaces/v1/Journeys.py
@@ -447,8 +447,8 @@ class Journeys(JourneyCommon):
                 args['_max_successive_physical_mode'] = mod.max_successive_physical_mode
             if args.get('_final_line_filter') is None:
                 args['_final_line_filter'] = mod.final_line_filter
-            if args.get('_max_extra_second_pass') is None:
-                args['_max_extra_second_pass'] = mod.max_extra_second_pass
+            if args.get('max_extra_second_pass') is None:
+                args['max_extra_second_pass'] = mod.max_extra_second_pass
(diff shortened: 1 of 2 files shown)


source/jormungandr/jormungandr/interfaces/v1/Journeys.py
```diff
@@ -448,8 +447,8 @@ def _set_specific_params(mod):
                 args['_max_successive_physical_mode'] = mod.max_successive_physical_mode
             if args.get('_final_line_filter') is None:
                 args['_final_line_filter'] = mod.final_line_filter
-            if args.get('_max_extra_second_pass') is None:
-                args['_max_extra_second_pass'] = mod.max_extra_second_pass
+            if args.get('max_extra_second_pass') is None:
+                args['max_extra_second_pass'] = mod.max_extra_second_pass
             if args.get('additional_time_after_first_section_taxi') is None:
```
kadhikari, line 452: What about https://github.com/CanalTP/navitia/blob/3a4d48a14deb21e7bc03cf999c35e859d29776a5/source/jormungandr/jormungandr/interfaces/v1/Journeys.py#L271 ?

Don't you also have to modify it to use a parameter 'max_extra_second_pass' ?

kadhikari:
You can at least add a test with a parameter 'max_extra_second_pass' in https://github.com/CanalTP/navitia/blob/dev/source/jormungandr/tests/stif_tests.py

> As mentioned orally: 
This would be nice, but testing only URL param (already working), not db param.
And this param is useful on very narrow cases that I don't think we have in our mini-Krakens: so we wouldn't see any effect.
A good way to test this PR would be to fill a temporary Jormun db (directly or via Tyr) then intercept a /journeys call to Kraken to check which params are filled... But this is not easy to do.

source/jormungandr/jormungandr/interfaces/v1/Journeys.py
pbougue: It's actually where the magic happens (and it was not spotted when the previous code was written).
In the line mentioned, we expose a (hidden) param used in url under the name `_max...` but we store and use it under the name `max...` (with `dest=`). So it's the only place with doc where the `_` needs to be, as we never manage args directly (almost).

kadhikari (approved):

kinnou02 (approved):

PR #2856, opened 2019-07-10
fusio2ed: Stop times are over computed with frequencies.txt

benoit-bst:
Currently, we have the following behavior

- **fusio2ed** fills `ed::data` struct with stop_times (_stop_times.txt_ file)

stop_times.txt example
stop_times

- If **frequencies.txt** exists with the same `trip_id`, **fusio2ed** shifts all the stop times to the frequencies starting value (_start_time_ col)

frequencies.txt example
frequencies

- But a bug is introduced because of an untimely function (_finalize_frequency_ function). 
In fact, after the _fusioParser_ calling, **fusio2ed** calls `data.complete()` which calls `finalize_frequency()`. 
This is an over computing that create wrong stop times values around zero.

- Solution, don't call `finalize_frequency()` and tests it with a Utest :unicorn: 

This PR shows that we need some extra _RFC doc_ about this mechanism. Maybe in a future PR.

### DIFF
--- a/source/ed/tests/fusioparser_test.cpp
+++ b/source/ed/tests/fusioparser_test.cpp
@@ -482,3 +482,38 @@ BOOST_AUTO_TEST_CASE(admin_stations_retrocompatibilty_tests) {
         BOOST_CHECK_EQUAL(data.admin_stop_areas[0]->stop_area[0]->uri, "SA:A");
     }
 }
+
+BOOST_AUTO_TEST_CASE(frequencies_tests) {
+    ed::Data data;
+    ed::connectors::FusioParser parser(ntfs_path);
+    parser.fill(data);
+    data.complete();  // complete() is called by fusio2ed
+
+    // trip_5 is composed of 4 stop_times.
+    // Check that inside fixtures/ed/ntfs/stop_times.txt
+    const auto& mvj = data.meta_vj_map["trip_5"];
+    // We split the vj in 2 parts because of the time change
+    BOOST_REQUIRE_EQUAL(mvj.theoric_vj.size(), 2);
+    const auto& vj = mvj.theoric_vj[0];
+    BOOST_REQUIRE_EQUAL(vj->stop_time_list.size(), 4);
+    BOOST_CHECK_EQUAL(vj->name, "trip_5_dst_1");
+    BOOST_CHECK_EQUAL(vj->stop_time_list[0]->departure_time, "05:00:00"_t);
+    BOOST_CHECK_EQUAL(vj->stop_time_list[0]->stop_point->uri, "SP:A");
+    BOOST_CHECK_EQUAL(vj->stop_time_list[1]->departure_time, "05:10:00"_t);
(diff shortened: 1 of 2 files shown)


source/ed/tests/fusioparser_test.cpp
```diff
@@ -482,3 +482,38 @@ BOOST_AUTO_TEST_CASE(admin_stations_retrocompatibilty_tests) {
...
+
+    // trip_5 is composed of 4 stop_times.
+    // Check that inside fixtures/ed/ntfs/stop_times.txt
+    const auto& mvj = data.meta_vj_map["trip_5"];
+    // We split the vj in 2 parts because of the time change
+    BOOST_REQUIRE_EQUAL(mvj.theoric_vj.size(), 2);
+    const auto& vj = mvj.theoric_vj[0];
+    BOOST_REQUIRE_EQUAL(vj->stop_time_list.size(), 4);
+    BOOST_CHECK_EQUAL(vj->name, "trip_5_dst_1");
+    BOOST_CHECK_EQUAL(vj->stop_time_list[0]->departure_time, "05:00:00"_t);
```
benoit-bst, line 500: Without the fix the result are
vj->stop_time_list[0]->departure_time = 5
vj->stop_time_list[1]->departure_time = 15
vj->stop_time_list[2]->departure_time = 25
vj->stop_time_list[3]->departure_time = 35

kinnou02:
I don't think this is a valid change, still investigating how trip planning with freqVJ works

kinnou02 (changes requested):
you broke [`get_next_stop_time`](https://github.com/CanalTP/navitia/blob/dev/source/routing/next_stop_time.cpp#L531)

also tests mostly do the same thing: https://github.com/CanalTP/navitia/blob/dev/source/ed/build_helper.cpp#L735-L754

benoit-bst:
Hoo, I don't see tests broken inside `next_stop_time_test`.
Maybe it is a little bit hidden ?

Currently, the process is as follows :

- The parser calls  [`FrequenciesGtfsHandler::handle_line`](https://github.com/CanalTP/navitia/blob/dev/source/ed/connectors/gtfs_parser.cpp#L1085)
we compute like https://github.com/CanalTP/navitia/blob/dev/source/ed/build_helper.cpp#L735-L754

- [`finalize_frequency`](https://github.com/CanalTP/navitia/blob/dev/source/ed/data.cpp#L741) function is called
It seems to be used only in [`ed::Data::complete()`](https://github.com/CanalTP/navitia/blob/dev/source/ed/data.cpp#L278) and in [`tests`](https://github.com/CanalTP/navitia/blob/dev/source/ed/tests/fusioparser_test.cpp#L259) (My **Rg** said that). 
I don't see why `get_next_stop_time` is impacted. Can you help me @kinnou02 ?

benoit-bst:
So, the final act of this PR is a close, because the fix can't work without a side effect...
A new PR will be open to add information inside `/vehicle_journeys` API response just to higlight that it is a **Frequency Vehicle Journey** (add `start_time end_time headway_secs` values)

PR #2857, opened 2019-07-10
fusio2ed: suppress station_id retrocompatibility inside admin_stations.txt file

benoit-bst:
Previously on PR https://github.com/CanalTP/navitia/pull/2836, we wanted to read **station_id** column or **stop_id** column inside `admin_stations.txt`. 
For retrocompatibility, **station_id** was readed if **stop_id** didn't exist.

When the _data team_ will finish its migration, we'll allowed to only read **stop_id** column.
This PR aims to suppress the retrocompatibilty for **station_id** column. 
It's now strict about field, only **stop_id** is read and it is required.

:warning:  Wait the GO of the data team to merge

### DIFF
--- a/source/ed/connectors/fusio_parser.cpp
+++ b/source/ed/connectors/fusio_parser.cpp
@@ -1543,29 +1543,9 @@ void GridCalendarTripExceptionDatesFusioHandler::handle_line(Data&, const csv_ro
 }
 }  // namespace grid_calendar
 
-void AdminStopAreaFusioHandler::init(Data& data) {
+void AdminStopAreaFusioHandler::init(Data&) {
     admin_c = csv.get_pos_col("admin_id");
-
     stop_area_c = csv.get_pos_col("stop_id");
-
-    // For retro compatibity
-    // TODO : to remove after the data team update, it will become useless (NAVP-1285)
-    if (stop_area_c == UNKNOWN_COLUMN) {
-        stop_id_is_present = false;
-        stop_area_c = csv.get_pos_col("station_id");
-        for (const auto& object_code_map : data.object_codes) {
-            for (auto& object_code : object_code_map.second) {
-                if (object_code_map.first.type == nt::Type_e::StopArea && object_code.first == "external_code") {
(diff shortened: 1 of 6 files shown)


benoit-bst:
Jenkins, test this please !

source/ed/connectors/fusio_parser.cpp
```diff
@@ -1543,29 +1543,9 @@ void GridCalendarTripExceptionDatesFusioHandler::handle_line(Data&, const csv_ro
 }
 }  // namespace grid_calendar
 
-void AdminStopAreaFusioHandler::init(Data& data) {
+void AdminStopAreaFusioHandler::init(Data&) {
```
pbougue, line 1546: You may remove the `Data&` from the signature then, no? Or is it kinda templated and you can't?

pbougue (approved):
We need to check that Artemis is OK too. If this feature is used in tests and data is not migrated there, it would break (and currently data updates from Artemis are stopped).
Actually it will break just if the file exists (and is not migrated), as mandatory field will stop bina :wink: 
I don't know if fusio has to do some changes for that.

source/ed/connectors/fusio_parser.cpp
benoit-bst: yes, it's a Template story... 
I can't suppress **Data** in the params function because we use `init()` method in a template context, even if **Data** is not used. 
Just remove the var name to avoid gcc warning.

https://github.com/CanalTP/navitia/blob/dev/source/ed/connectors/gtfs_parser.h#L430

benoit-bst:
Good Remark @pbougue. We have to pay attention of **artemis** because we change admin_stations.txt reading style with a mandatory field. It can break some tests. 
I'll run **Artemis** and report the result inside this PR

kinnou02:
This will break: a lot of platforms have old dataset

benoit-bst:
Closed, 
For the moment we don't have automatic NTFS update on dev platform. It will break our coverages

PR #2858, opened 2019-07-10
Jormun: Refacto street_network_configurations

AurelienLP:
We'll want to manage backends in db via Tyr and Tyr configurator. To do so, we want to have one class which will update the configurations every time we need it (when we call a backend).

Like `EquipmentProviderManager`, I have created a class `StreetNetworkBackendManager`.

For now, I only want to keep the retro compatibility with the actual json system.

This PR is only a refacto and shouldn't introduce any regression.

It is reviewable commit by commit !


source/jormungandr/jormungandr/instance.py
```diff
@@ -162,10 +133,9 @@ def __init__(
         self.georef = georef.Kraken(self)
         self.planner = planner.Kraken(self)
 
-        street_network_configurations = _set_default_street_network_config(street_network_configurations)
-        self.street_network_services = street_network.StreetNetwork.get_street_network_services(
-            self, street_network_configurations
-        )
+        self.streetnetwork_backend_manager = StreetNetworkBackendManager(street_network_configurations)
```
kinnou02: I would prefer having a manager shared between instances to be able to use the same instance of the backend class so circuit breaker, caches and others things can be shared.

AurelienLP:
Jenkins, test this please

AurelienLP:
jenkins, test this

AurelienLP:
Jenkins, test this please

source/jormungandr/jormungandr/instance.py
```diff
@@ -49,7 +49,7 @@
 from jormungandr.exceptions import DeadSocketException
 from navitiacommon import models
 from importlib import import_module
-from jormungandr import cache, memory_cache, app, global_autocomplete
+from jormungandr import cache, memory_cache, app, global_autocomplete, streetnetwork_backend_manager
```
kinnou02: if possible inject it as a dependency

AurelienLP: Done

kinnou02 (approved):

source/jormungandr/jormungandr/instance.py
```diff
@@ -200,6 +169,9 @@ def __init__(
         # Init equipment providers from config
         self.equipment_provider_manager.init_providers(instance_equipment_providers)
 
+        self._get_street_network = streetnetwork_backend_manager.get_street_network
```
xlqian: :smiling_imp:

source/jormungandr/jormungandr/street_network/streetnetwork_backend_manager.py
```diff
@@ -0,0 +1,118 @@
...
+            instance_configuration.append(copy.deepcopy(config))
+
+        return instance_configuration
+
+    def _create_street_network_backends(self, instance, instance_configuration):
+        for config in instance_configuration:
+            # Set default arguments
+            if 'args' not in config:
+                config['args'] = {}
+            if 'service_url' not in config['args']:
```
xlqian: config['args'].setdefault('service_url', None)
config['args'].setdefault('instance', instance)
config['args'].setdefault('modes', config.get('modes', []))

source/jormungandr/jormungandr/instance.py
AurelienLP: Does it make the code harder to read? I should maybe find an other way...

xlqian (approved):

AurelienLP:
Jenkins, test this please

PR #2859, opened 2019-07-11
Fix issue with admin out of date for osm2ed

GuillaumeLescure:
- [x] Requires #2865

### DIFF
--- a/source/ed/osm2ed.cpp
+++ b/source/ed/osm2ed.cpp
@@ -77,6 +77,20 @@ void ReadRelationsVisitor::relation_callback(uint64_t osm_id,
         if (it_level == accepted_levels.end()) {
             return;
         }
+        const auto& it_end_date = tags.find("end_date");
+        if (it_end_date != tags.end()) {
+            try {
+                const boost::gregorian::date end_date(boost::gregorian::from_string(it_end_date->second));
+                if (end_date < boost::gregorian::date(boost::gregorian::day_clock::universal_day())) {
+                    LOG4CPLUS_INFO(logger, "admin '" << tags.find("name")->second << "'with osm_id=" << osm_id
+                                                     << " is out of date: " << end_date << " (dropped)");
+                    return;
+                }
+            } catch (boost::bad_lexical_cast& exception) {
+                LOG4CPLUS_INFO(logger, "admin '" << tags.find("name")->second << "' with osm_id=" << osm_id
+                                                 << " has an invalid end date: '" << it_end_date->second << "' (kept)");
+            }
+        }
         for (const CanalTP::Reference& ref : refs) {
             switch (ref.member_type) {
                 case OSMPBF::Relation_MemberType::Relation_MemberType_WAY:
--- a/source/ed/tests/osm2ed_test.cpp
+++ b/source/ed/tests/osm2ed_test.cpp
@@ -32,6 +32,8 @@ www.navitia.io
 #define BOOST_TEST_MODULE osm2ed
 
 #include <boost/test/unit_test.hpp>
+#include "osmpbfreader/osmpbfreader.h"
+#include "utils/lotus.h"
 #include "ed/types.h"
 #include "ed/osm2ed.h"
 
@@ -42,6 +44,7 @@ struct logger_initialized {
 BOOST_GLOBAL_FIXTURE(logger_initialized);
 
 using namespace ed::connectors;
+using namespace CanalTP;
 
 BOOST_AUTO_TEST_CASE(osm2ed_with_no_param_should_start_without_throwing) {
     const char* argv[] = {"osm2ed_test"};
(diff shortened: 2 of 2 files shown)


GuillaumeLescure:
This fix, drop the out of date admins at the parsing of the protobuff file.
Current testes show that Dinan, Léhon, Hédé and Bazouges-sous-Hédé are concerned by this fix for the coverage "fr-bre". It's expected to have other similar fixes for other coverages thanks to that patch.

image

source/ed/osm2ed.cpp
```diff
@@ -76,6 +76,15 @@ void ReadRelationsVisitor::relation_callback(uint64_t osm_id,
         if (it_level == accepted_levels.end()) {
             return;
         }
+        const auto& it_end_date = tags.find("end_date");
+        if (it_end_date != tags.end()) {
+            const boost::gregorian::date end_date(boost::gregorian::from_string(it_end_date->second));
```
pbougue: Not sure about the behavior if date is ill-formated.
I guess in that case we should keep the OSM object.

pbougue (commented):
It looks OK.
I don't know if it's easy (and too expensive) to port OSM integration tests in navitia but we may get inspiration from https://github.com/CanalTP/mimirsbrunn/tree/master/tests/fixtures

source/ed/osm2ed.cpp
kinnou02: it probably throws an exception, so ...

GuillaumeLescure: If the date is ill-formated, boost::gregorian::date will set the value "not_a_date_time" for the instance and it can be checked with myDate.is_not_a_date().

But I check the behaviour with the operator '<' and '>', it's kind of weird because :
* not_a_date < valid_date == false
* not_a_date > valid_date == true

(source: https://www.boost.org/doc/libs/1_39_0/doc/html/date_time/gregorian.html)

I fix my patch to handle that case. Thanks for the heads-up :+1:

GuillaumeLescure:
Note : it would be nice if Boost were updated :
* latest - 1.70
* us - 1.55

GuillaumeLescure:
Test this please.

GuillaumeLescure:
[ ] Require #2865

source/ed/tests/osm2ed_test.cpp
```diff
@@ -67,3 +70,53 @@ BOOST_AUTO_TEST_CASE(osm_poi_id_relation_should_match_mimirs_one) {
...
+// Check if the administration is accepted if the "end_date" tag is not given
+BOOST_AUTO_TEST_CASE(osm_admin_no_end_date) {
+    OSMCache cache(std::unique_ptr<Lotus>(), boost::none);
+    ReadRelationsVisitor relations_visitor(cache, false);
+
+    Tags tags = {{"name", "NavitiaCity0"}, {"admin_level", "9"}, {"boundary", "administrative"}};
+    References ref(1, Reference());
+
+    relations_visitor.relation_callback(5, tags, ref);
+    BOOST_CHECK_EQUAL(relations_visitor.cache.admins.find(5) == relations_visitor.cache.admins.end(), false);
```
nonifier:
```suggestion
    BOOST_CHECK_NE(relations_visitor.cache.admins.find(5), relations_visitor.cache.admins.end());
```

This applies for the following assertions as well ;)

source/ed/tests/osm2ed_test.cpp
```diff
@@ -67,3 +70,53 @@ BOOST_AUTO_TEST_CASE(osm_poi_id_relation_should_match_mimirs_one) {
...
+BOOST_AUTO_TEST_CASE(osm_admin_not_out_of_date) {
+    OSMCache cache(std::unique_ptr<Lotus>(), boost::none);
+    ReadRelationsVisitor relations_visitor(cache, false);
+
+    Tags tags = {
+        {"name", "NavitiaCity1"}, {"admin_level", "9"}, {"boundary", "administrative"}, {"end_date", "2050-12-31"}};
+    References ref(1, Reference());
+
+    relations_visitor.relation_callback(5, tags, ref);
+    BOOST_CHECK_EQUAL(relations_visitor.cache.admins.find(5) == relations_visitor.cache.admins.end(), false);
```
nonifier:
```suggestion
    BOOST_CHECK_NE(relations_visitor.cache.admins.find(5), relations_visitor.cache.admins.end());
```

nonifier: et cetera.

nonifier (approved):

source/ed/tests/osm2ed_test.cpp
GuillaumeLescure: Done

source/ed/tests/osm2ed_test.cpp
GuillaumeLescure: Done

benoit-bst (approved):

GuillaumeLescure:
Test this please

PR #2869, opened 2019-07-15
rt_validity_pattern used to verify vehicle_journey validity

kadhikari:
- In api /vehicle_journeys only base_validity_pattern is used to verify if the vehicle_journey is valid.
- Since for a vehicle_journey added by a realtime (added trip) base_validity_pattern is created with all zeros where as a rt_validity_pattern is created with at least one day valid (1) , the use of filters &since and/or &until returns 0 element.

Concerns : https://jira.kisio.org/browse/NAVP-1273


benoit-bst (approved):

kinnou02 (approved):

nonifier (approved):
:ok_hand:

PR #2870, opened 2019-07-16
jormungandr: completing realtime tests on schedules

kadhikari:
Concerned ticket: https://jira.kisio.org/browse/NAVP-1368

### DIFF
--- a/source/jormungandr/tests/kirin_realtime_tests.py
+++ b/source/jormungandr/tests/kirin_realtime_tests.py
@@ -1871,13 +1871,22 @@ class TestKirinAddNewTrip(MockKirinDisruptionsFixture):
         assert departures['departures'][0]['display_informations']['name'] == 'stopC - stopB'
 
         # Check that stop_schedule on line "line:stopC_stopB" and stop_point stop_point:stopC
-        # exists with disruption
-        stop_schedules = self.query_region(ss_on_line_query)
+        # for base_schedule date_times list is empty.
+        stop_schedules = self.query_region(ss_on_line_query + '&data_freshness=base_schedule')
+        assert len(stop_schedules['stop_schedules']) == 1
+        assert stop_schedules['stop_schedules'][0]['links'][0]['type'] == 'line'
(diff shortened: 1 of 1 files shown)


kinnou02:
Jenkins, test this please.

source/jormungandr/tests/kirin_realtime_tests.py
```diff
@@ -1871,13 +1871,22 @@ def test_add_new_trip(self):
...
+
+        # Check that stop_schedule on line "line:stopC_stopB" and stop_point stop_point:stopC
+        # exists with disruption.
+        stop_schedules = self.query_region(ss_on_line_query + '&data_freshness=realtime')
         assert len(stop_schedules['stop_schedules']) == 1
         assert stop_schedules['stop_schedules'][0]['links'][0]['id'] == 'line:stopC_stopB'
         assert len(stop_schedules['disruptions']) == 1
         assert stop_schedules['disruptions'][0]['uri'] == 'new_trip'
         assert len(stop_schedules['stop_schedules'][0]['date_times']) == 1
+        assert stop_schedules['stop_schedules'][0]['date_times'][0]['date_time'] == '20120614T080100'
```
benoit-bst, line 1889: What is _date_times_ field and why is it empty in **base_schedule** and not in **realtime** mode ?

kadhikari: 1. 'date_times' represent departures at the point "stop_point:stopC"
2. Before we inject cots we have one stop_schedule but without any  'date_times'
3. After injection of cots, since many pt_objects are created:
- with &datafreshness=base_schedule we have two stop_schedules without any 'stop_times'
- with &datafreshness=realtime we have two stop_schedules with one 'stop_times' for new vj added.

benoit-bst: Thanks for the explanation :cat:

benoit-bst (approved):

mbounabi (approved):
Great! 
However, the test is getting quiiiiite long. For the next test cases, it might be worth splitting it.

kinnou02 (approved):

PR #2868, opened 2019-07-15
doc: frequency vehicle journeys RFC

benoit-bst:
Add some doc about `Frequency vehicle journeys`. 
How **Kraken** works with it
[skip-ci]

PR #2873, opened 2019-07-17
kraken: VJ frequency informations inside proto response

benoit-bst:
**Objective**: Trace informations when a VJ is in a **frequency** mode. 

Currently, we can't see any information about the frequency mode inside the Response API. The minimum is to inform the user by showing frequency params inside Vehicle Journey.

A Utest has been designed, calling the `/vehicle_journeys` API

:information_source: This PR covers only the _Kraken_ part. An other PR to come with the _Jormun_ part, to expose the result into the response API 

**Protobuf** was updated for this job : https://github.com/CanalTP/navitia-proto/pull/130

A piece of **doc** is written here https://github.com/CanalTP/navitia/pull/2868

### DIFF
--- a/source/type/pb_converter.cpp
+++ b/source/type/pb_converter.cpp
@@ -857,12 +857,8 @@ void PbCreator::Filler::fill_pb_object(const nt::VehicleJourney* vj, pbnavitia::
 
         fill(vector_bp, vehicle_journey->mutable_calendars());
 
-        if (!pair_jp.second.freq_vjs.empty()) {
-            for (const auto& frequency_vehicle_journey : pair_jp.second.freq_vjs) {
-                if (frequency_vehicle_journey->uri == vj->uri) {
-                    fill_pb_object(frequency_vehicle_journey, vehicle_journey);
-                }
-            }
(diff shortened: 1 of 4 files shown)


source/type/pb_converter.cpp
```diff
@@ -856,12 +856,25 @@ void PbCreator::Filler::fill_pb_object(const nt::VehicleJourney* vj, pbnavitia::
         const auto& vector_bp = navitia::vptranslator::translate(*vj->base_validity_pattern());
 
         fill(vector_bp, vehicle_journey->mutable_calendars());
+
+        if (!pair_jp.second.freq_vjs.empty()) {
+            for (const auto& frequency_vehicle_journey : pair_jp.second.freq_vjs) {
```
kinnou02: this seems overly complicated:  `FrequencyVehicleJourney` inherit of `VehicleJourney` so you can just `dynamic_cast` after having checked that it's a freq vj

kinnou02 (changes requested):

kinnou02 (approved):

GuillaumeLescure (approved):

benoit-bst:
Jenkins, test this please !

PR #2874, opened 2019-07-17
Jormun: Streetnetwork backend in db

AurelienLP:
This PR aims to retrieve streetnetwork_backends configurations from the db and use them.

The retro compatibility with the JSON is still there but is named "legacy" now.

To do so, I have split the behavior for the JSON and the database in the class StreetNetworkBackendManager.

This PR is readable commit by commit.

### DIFF
--- a/source/jormungandr/jormungandr/fallback_modes.py
+++ b/source/jormungandr/jormungandr/fallback_modes.py
@@ -89,12 +89,3 @@ class FallbackModes(Enum):
 all_fallback_modes = FallbackModes.modes_str()
 
 allowed_combinations = FallbackModes.get_allowed_combinations_str()
-
-street_network_modes_in_db = {
-    FallbackModes.car.name: "street_network_car",
-    FallbackModes.walking.name: "street_network_walking",
-    FallbackModes.bike.name: "street_network_bike",
-    FallbackModes.bss.name: "street_network_bss",
-    FallbackModes.ridesharing.name: "street_network_ridesharing",
-    FallbackModes.taxi.name: "street_network_taxi",
-}
--- a/source/jormungandr/jormungandr/fallback_modes.py
+++ b/source/jormungandr/jormungandr/fallback_modes.py
@@ -89,3 +89,12 @@ class FallbackModes(Enum):
 all_fallback_modes = FallbackModes.modes_str()
 
 allowed_combinations = FallbackModes.get_allowed_combinations_str()
+
+street_network_modes_in_db = {
+    FallbackModes.car.name: "street_network_car",
+    FallbackModes.walking.name: "street_network_walking",
+    FallbackModes.bike.name: "street_network_bike",
+    FallbackModes.bss.name: "street_network_bss",
+    FallbackModes.ridesharing.name: "street_network_ridesharing",
+    FallbackModes.taxi.name: "street_network_taxi",
+}
--- a/source/jormungandr/jormungandr/instance.py
+++ b/source/jormungandr/jormungandr/instance.py
@@ -55,7 +55,7 @@ from shapely.geos import ReadingError, PredicateError
 from flask import g
 import flask
 import pybreaker
-from jormungandr import georef, planner, schedule, realtime_schedule, ptref, street_network
+from jormungandr import georef, planner, schedule, realtime_schedule, ptref, street_network, fallback_modes
 from jormungandr.scenarios.ridesharing import ridesharing_service
 import six
 import time
@@ -134,7 +134,10 @@ class Instance(object):
         self.planner = planner.Kraken(self)
         self._streetnetwork_backend_manager = streetnetwork_backend_manager
 
-        self._streetnetwork_backend_manager.init_streetnetwork_backends(self, street_network_configurations)
+        if app.config[str('DISABLE_DATABASE')]:
+            self._streetnetwork_backend_manager.init_streetnetwork_backends_legacy(
+                self, street_network_configurations
+            )
 
         self.ridesharing_services = []  # type: List[ridesharing_service.AbstractRidesharingService]
         if ridesharing_configurations is not None:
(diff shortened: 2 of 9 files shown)


kinnou02 (approved):

source/jormungandr/jormungandr/fallback_modes.py
```diff
@@ -89,3 +89,12 @@ def get_allowed_combinations_str(cls):
...
 allowed_combinations = FallbackModes.get_allowed_combinations_str()
+
+street_network_modes_in_db = {
+    FallbackModes.car.name: "street_network_car",
+    FallbackModes.walking.name: "street_network_walking",
+    FallbackModes.bike.name: "street_network_bike",
+    FallbackModes.bss.name: "street_network_bss",
+    FallbackModes.ridesharing.name: "street_network_ridesharing",
+    FallbackModes.taxi.name: "street_network_taxi",
+}
```
mbounabi: These names are supposed to be the id in the db? Isn't there an enum for them to use here?
Or is it the column name?
**Update**: I checked the db upgrade PR, and it is the column name in the instance table, right? However, the comment is still valid for me. the names of the sn are repeated several times through the code. Is it safe?

source/jormungandr/jormungandr/fallback_modes.py
```diff
@@ -89,3 +89,12 @@ def get_allowed_combinations_str(cls):
 all_fallback_modes = FallbackModes.modes_str()
 
 allowed_combinations = FallbackModes.get_allowed_combinations_str()
+
+street_network_modes_in_db = {
```
nonifier: To express string enumeration from a db, you could use [sqlalchemy.types.Enum](https://docs.sqlalchemy.org/en/13/core/type_basics.html#sqlalchemy.types.Enum) or a simple [class enum.Enum](https://docs.python.org/3/library/enum.html#module-enum) could do the trick.

source/jormungandr/jormungandr/instance.py
```diff
@@ -655,34 +665,26 @@ def init(self):
...
 
     def get_street_network(self, mode, request):
-        return self._streetnetwork_backend_manager.get_street_network(self, mode, request)
+
+        if app.config[str('DISABLE_DATABASE')]:
+            return self._streetnetwork_backend_manager.get_street_network_legacy(self, mode, request)
+        else:
+            column_in_db = fallback_modes.street_network_modes_in_db.get(mode, None)
+            streetnetwork_backend_conf = getattr(self, column_in_db)
+            return self._streetnetwork_backend_manager.get_street_network_db(self, streetnetwork_backend_conf)
```
mbounabi, line 676: Here, you have to get the name of the column in the instance table (ex: sn_car), then the value that's in this column (ex: valhalla) and then get in the sn table in the db, the config of the sn. Is that it?

source/jormungandr/jormungandr/street_network/streetnetwork_backend_manager.py
```diff
@@ -86,14 +91,67 @@ def _create_street_network_backends(self, instance, instance_configuration):
...
+                sn_backend.id not in self._streetnetwork_backends_last_update
+                or sn_backend.last_update() > self._streetnetwork_backends_last_update[sn_backend.id]
+            ):
+                self._update_sn_backend(sn_backend, instance)
+
+    def get_street_network_db(self, instance, streetnetwork_backend_conf):
+        # Make sure we update the streetnetwork_backends list from the database before returning them
+        self._update_config(instance)
+
+        return self._streetnetwork_backends[streetnetwork_backend_conf]
```
nonifier: This will throw if you have a bad backend configuration setup in db. You might want to protect against looking for a wrong key: 
```python
self._streetnetwork_backends.get(streetnetwork_backend_conf)
```

nonifier (approved):

source/jormungandr/jormungandr/instance.py
AurelienLP: Yes, that's exactly that!

source/jormungandr/jormungandr/fallback_modes.py
AurelienLP: My need here is to have a safe way to match the mode use for the request with the db column in instance. For example "car" = "street_network_car"

mbounabi (approved):

kadhikari (approved):

nonifier (approved):

AurelienLP:
Jenkins, test this please

source/jormungandr/jormungandr/street_network/streetnetwork_backend_manager.py
```diff
@@ -86,14 +91,67 @@ def _create_street_network_backends(self, instance, instance_configuration):
...
+
+        return utils.create_object(config)
+
+    def _update_sn_backend(self, sn_backend, instance):
+        self.logger.info('Updating / Adding {} streetnetwork backend'.format(sn_backend.id))
+        try:
+            self._streetnetwork_backends[sn_backend.id] = self._create_backend_from_db(sn_backend, instance)
+            self._streetnetwork_backends_last_update[sn_backend.id] = sn_backend.last_update()
+        except Exception:
+            self.logger.exception('impossible to initialize streetnetwork backend')
```
xlqian: self.logger.exception('')

source/jormungandr/jormungandr/street_network/streetnetwork_backend_manager.py
```diff
@@ -86,14 +91,67 @@ def _create_street_network_backends(self, instance, instance_configuration):
...
+            return
+
+        self.logger.debug('Updating streetnetwork backends from db')
+        self._last_update = datetime.datetime.utcnow()
+
+        sn_backends = []
+        try:
+            sn_backends = self._sn_backends_getter()
+        except Exception:
+            self.logger.exception('Failure to retrieve streetnetwork backends configuration')
```
xlqian: self.logger.exception('')

xlqian (approved):

PR #2875, opened 2019-07-18
jormun: fix error in parking provider

kinnou02:
If no `api_key` was provided the provider didn't work at all :(
Also improve tests of star parking provider to detect this kind of
errors.

### DIFF
--- a/source/jormungandr/jormungandr/parking_space_availability/car/tests/star_test.py
+++ b/source/jormungandr/jormungandr/parking_space_availability/car/tests/star_test.py
@@ -30,7 +30,7 @@
 from __future__ import absolute_import, print_function, unicode_literals, division
 
 import pytest
-from mock import MagicMock
+import requests_mock
 
 from jormungandr.parking_space_availability.car.star import StarProvider
 from jormungandr.parking_space_availability.car.parking_places import ParkingPlaces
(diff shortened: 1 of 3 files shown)


source/jormungandr/jormungandr/parking_space_availability/car/tests/star_test.py
```diff
@@ -49,80 +52,77 @@ def car_park_space_availability_start_support_poi_test():
...
+    with requests_mock.Mocker() as m:
+        m.get('http://fake.url?dataset=toto', json=None)
+        assert provider.get_informations(poi) is None
+        assert m.called
 
+
+def car_park_space_get_information_no_data_test():
+    provider = StarProvider("http://fake.url", {'Keolis Rennes'}, 'toto', 42)
+    star_response = {"records": [{"fields": {"idparc": "42"}}]}
+    provider = StarProvider("http://fake.url", {'Keolis Rennes'}, 'toto', 42)
```
AurelienLP, line 93: Why do we initialize the provider twice here?

source/jormungandr/jormungandr/parking_space_availability/car/tests/star_test.py
```diff
@@ -39,6 +39,9 @@
     'properties': {'operator': 'Keolis Rennes', 'ref': '42'},
     'poi_type': {'name': 'parking relais', 'id': 'poi_type:amenity:parking'},
 }
+invalid_poi = {}
+
+empty_parking = ParkingPlaces(available=None, occupied=None, available_PRM=None, occupied_PRM=None)
```
AurelienLP: minor : `invalid_poi` and `empty_parking` seem to be used only once so they can be initialized in the function they are used in.

AurelienLP (approved):

source/jormungandr/jormungandr/parking_space_availability/car/tests/star_test.py
kinnou02: Because I'm dumb :p

nonifier (approved):

PR #2878, opened 2019-07-19
PTReferential: manage rt_vehicle_journeys with since and until params

kadhikari:
After many fails on Artemis tests (guichet_unique_test), parameter 'data_feshness' with 'base_schedule' as default value has to be added for backward compatibility (as for kirin). Api /journeys with parameters &since and &until should return only valid vjs of type base_schedule where as use of &data_fresness=realtime with other two parameter will return also valid vjs modified or added by realtime.

Merged PR with errors on Artemis tests: https://github.com/CanalTP/navitia/pull/2869

Reverted PR: https://github.com/CanalTP/navitia/pull/2872

A test artemis with this build will be realized in futur.

JIRA: https://jira.kisio.org/browse/NAVP-1273

### DIFF
--- a/source/jormungandr/jormungandr/interfaces/v1/Uri.py
+++ b/source/jormungandr/jormungandr/interfaces/v1/Uri.py
@@ -114,6 +114,16 @@ class Uri(ResourceUri, ResourceUtc):
             'of the disruptions.',
         )
         parser.add_argument(
+            "data_freshness",
+            help="Define the freshness of data to use to compute vehicle_journeys "
+            "along with parameters &since and/or &until .\n"
+            "When using `&data_freshness=base_schedule` you get original vehicle_journeys "
+            "where as when using `&data_freshness=realtime` you get vehicle_journeys "
+            "added or modified by realtime.",
+            type=OptionValue(['base_schedule', 'adapted_schedule', 'realtime']),
+            default='base_schedule',
+        )
+        parser.add_argument(
             "distance",
             type=int,
             default=200,
--- a/source/ptreferential/ptreferential_ng.cpp
+++ b/source/ptreferential/ptreferential_ng.cpp
@@ -223,8 +223,9 @@ boost::posix_time::ptime from_datetime(const std::string& s) {
 
 struct Eval : boost::static_visitor<Indexes> {
     const Type_e target;
+    const type::RTLevel rt_level;
     const type::Data& data;
-    Eval(Type_e t, const type::Data& d) : target(t), data(d) {}
+    Eval(Type_e t, const type::RTLevel rtl, const type::Data& d) : target(t), rt_level(rtl), data(d) {}
 
     Indexes operator()(const ast::All&) const { return data.get_all_index(target); }
     Indexes operator()(const ast::Empty&) const { return Indexes(); }
@@ -282,7 +283,8 @@ struct Eval : boost::static_visitor<Indexes> {
                 until = from_datetime(f.args.at(1));
             }
             const auto type = type_by_caption(f.type);
-            indexes = filter_on_period(data.get_all_index(type), type, since, until, data);
+            indexes = filter_on_period(data.get_all_index(type), type, since, until, rt_level, data);
+
         } else if (f.method == "within" && f.args.size() == 2) {
             double distance = 0.;
             try {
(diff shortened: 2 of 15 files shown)


source/jormungandr/jormungandr/interfaces/v1/Uri.py
```diff
@@ -113,6 +113,16 @@ def __init__(self, is_collection, collection, *args, **kwargs):
...
+        parser.add_argument(
+            "data_freshness",
+            help="Define the freshness of data to use to compute vehicle_journeys "
+            "along with parameters &since and/or &until .\n"
+            "When using `&data_freshness=base_schedule` you get original vehicle_journeys "
+            "where as when using `&data_freshness=realtime` you get vehicle_journeys "
+            "added or modified by realtime.",
+            type=OptionValue(['base_schedule', 'adapted_schedule', 'realtime']),
+            default='base_schedule',
+        )
```
mbounabi, line 125: Adding the parameter in this class adds it in every classes inheriting it. In some of them, 'data_freshness' doesn't make much sense. Is it worth it?
If the goal here is to fix /disruptions and /vehicle_journeys, why not add it only in these classes?

source/ptreferential/ptreferential_utils.cpp
```diff
@@ -142,14 +142,15 @@ Indexes get_impacts_by_tags(const std::vector<std::string>& tag_name, const Data
...
-static bool keep_vj(const type::VehicleJourney* vj, const bt::time_period& period) {
+static bool keep_vj(const type::VehicleJourney* vj, const bt::time_period& period, const type::RTLevel rt_level) {
     if (vj->stop_time_list.empty()) {
         return false;  // no stop time, so it cannot be valid
     }
 
     const auto& first_departure_dt = vj->earliest_time();
     for (boost::gregorian::day_iterator it(period.begin().date()); it <= period.last().date(); ++it) {
-        if (!vj->base_validity_pattern()->check(*it)) {
+        if (!(vj->base_validity_pattern()->check(*it)
```
kinnou02: You can acces the VP with the flatmap: `flat_enum_map<RTLevel, ValidityPattern*> validity_patterns = {{{nullptr, nullptr, nullptr}}};`

source/ptreferential/ptreferential_ng.cpp
```diff
@@ -434,13 +436,14 @@ Indexes make_query_ng(const Type_e requested_type,
...
                       const boost::optional<boost::posix_time::ptime>& until,
+                      const type::RTLevel rt_level,
                       const type::Data& data) {
     auto logger = log4cplus::Logger::getInstance("ptref");
     const auto request_ng = make_request(requested_type, request, forbidden_uris, odt_level, since, until, data);
     const auto expr = parse(request_ng);
     LOG4CPLUS_TRACE(logger, "ptref_ng parsed: " << expr << " [requesting: "
                                                 << navitia::type::static_data::captionByType(requested_type) << "]");
-    return Eval(requested_type, data)(expr);
+    return Eval(requested_type, rt_level, data)(expr);
```
kinnou02: Why hot add it to the expression?

source/jormungandr/jormungandr/interfaces/v1/Uri.py
kinnou02: It should be defined with `since` and  until` to be consistent (so here)

nonifier (approved):

source/ptreferential/ptreferential_utils.cpp
```diff
@@ -142,14 +142,14 @@ Indexes get_impacts_by_tags(const std::vector<std::string>& tag_name, const Data
...
-static bool keep_vj(const type::VehicleJourney* vj, const bt::time_period& period) {
+static bool keep_vj(const type::VehicleJourney* vj, const bt::time_period& period, const type::RTLevel rt_level) {
     if (vj->stop_time_list.empty()) {
         return false;  // no stop time, so it cannot be valid
     }
 
     const auto& first_departure_dt = vj->earliest_time();
     for (boost::gregorian::day_iterator it(period.begin().date()); it <= period.last().date(); ++it) {
-        if (!vj->base_validity_pattern()->check(*it)) {
+        if (!vj->validity_patterns[rt_level]->check(*it)) {
```
kinnou02: to be safe you should check that the vp isn't null

kinnou02 (approved):

nonifier (approved):

kinnou02 (approved):

PR #2880, opened 2019-07-19
ed: fix for out of range in osm2ed

kadhikari:
Following error during deployment on dev_bina, catch all exception from gregorian::date

### DIFF
--- a/source/ed/osm2ed.cpp
+++ b/source/ed/osm2ed.cpp
@@ -86,9 +86,10 @@ void ReadRelationsVisitor::relation_callback(uint64_t osm_id,
                                                      << " is out of date: " << end_date << " (dropped)");
                     return;
                 }
-            } catch (...) {
-                LOG4CPLUS_INFO(logger, "admin '" << tags.find("name")->second << "' with osm_id=" << osm_id
-                                                 << " has an invalid end date: '" << it_end_date->second << "' (kept)");
+            } catch (std::exception& e) {
+                LOG4CPLUS_WARN(logger, "admin '" << tags.find("name")->second << "' with osm_id=" << osm_id
+                                                 << " has an invalid end date (" << e.what() << "): '"
+                                                 << it_end_date->second << "' (kept)");
             }
         }
         for (const CanalTP::Reference& ref : refs) {
--- a/source/ed/osm2ed.cpp
+++ b/source/ed/osm2ed.cpp
@@ -86,7 +86,7 @@ void ReadRelationsVisitor::relation_callback(uint64_t osm_id,
                                                      << " is out of date: " << end_date << " (dropped)");
                     return;
                 }
-            } catch (boost::bad_lexical_cast& exception) {
+            } catch (...) {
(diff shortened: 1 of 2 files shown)


source/ed/tests/osm2ed_test.cpp
```diff
@@ -120,3 +120,39 @@ BOOST_AUTO_TEST_CASE(osm_admin_ill_formated_date) {
     relations_visitor.relation_callback(5, tags, ref);
     BOOST_CHECK(relations_visitor.cache.admins.find(5) != relations_visitor.cache.admins.end());
 }
+
+// Check if the administration is accepted if the "end_date" tag is well-formatted with out of range value
+BOOST_AUTO_TEST_CASE(osm_admin_out_of_range_in_end_date) {
+    OSMCache cache(std::unique_ptr<Lotus>(), boost::none);
```
AurelienLP, line 126: Isn't `std::make_unique<Lotus>()` better here ?

source/ed/osm2ed.cpp
```diff
@@ -86,7 +86,7 @@ void ReadRelationsVisitor::relation_callback(uint64_t osm_id,
                                                      << " is out of date: " << end_date << " (dropped)");
                     return;
                 }
-            } catch (boost::bad_lexical_cast& exception) {
+            } catch (...) {
```
benoit-bst: Okay, catch all exceptions

source/ed/tests/osm2ed_test.cpp
```diff
@@ -120,3 +120,39 @@ BOOST_AUTO_TEST_CASE(osm_admin_ill_formated_date) {
...
+    ReadRelationsVisitor relations_visitor(cache, false);
+
+    Tags tags = {{"name", "NavitiaCity5"}, {"admin_level", "9"}, {"boundary", "administrative"}, {"end_date", "2000-02-31"}};
+    References ref(1, Reference());
+
+    relations_visitor.relation_callback(5, tags, ref);
+    BOOST_CHECK(relations_visitor.cache.admins.find(5) != relations_visitor.cache.admins.end());
+}
+
+// Check if the administration is accepted if the "end_date" tag is well-formatted but differently
```
kadhikari: Here : rejected instead of accepted.  This will be modified later on

source/ed/osm2ed.cpp
AurelienLP: As seen orally `const std::exception& e` with a `e.what()` in log (info or warn) might be better.

source/ed/tests/osm2ed_test.cpp
GuillaumeLescure: In that case, std::unique_ptr<Lotus>() is like a NULL pointer, I don't want to have a Lotus instance to avoid connection to the database.

nonifier (approved):

source/ed/osm2ed.cpp
benoit-bst: I imagine that we want to catch only std::exceptions. Okay for me

benoit-bst (approved):

AurelienLP (approved):

PR #2885, opened 2019-07-22
jormun: upgrade to flask 1.0

kinnou02:
flask 0.14 is is vulnerable: [CVE-2019-1010083](https://nvd.nist.gov/vuln/detail/CVE-2019-1010083)


source/eitri/requirements.txt
```diff
@@ -4,7 +4,7 @@ retrying==1.2.3
 psycopg2
 clingon==0.1.4
 GeoAlchemy2==0.2.4
-SQLAlchemy==1.3.3
```
AurelienLP: Why a downgrade here ? :'(

source/sql/requirements.txt
```diff
@@ -4,5 +4,5 @@ Mako==0.9.1
 MarkupSafe==0.18
 SQLAlchemy==1.3.3
 Shapely==1.3.0
-alembic==0.6.2
-psycopg2
```
AurelienLP, line 8: Shouldn't we fix the version of psycopg2 ?

source/eitri/requirements.txt
```diff
@@ -4,7 +4,7 @@ retrying==1.2.3
...
+SQLAlchemy==1.0.11
```
mbounabi: Downgrade?

mbounabi (approved):

source/eitri/requirements.txt
kinnou02: oops, wrong line...

source/sql/requirements.txt
kinnou02: we want to use the psycopg installed by the system to not have rebuild everything

AurelienLP (approved):

PR #2891, opened 2019-07-23
[Jormun]: add try except when one pt failed

xlqian:
https://jira.kisio.org/browse/NAVITIAII-2817

Distributed return nothing at all when one of demanded pt_journey fails... 

This pr will allow the exception and return whatever it find

:warning:  Since it's quite complicated to add a test with main_routing_test... a  test will be added into Artemis directly

### DIFF
--- a/source/jormungandr/jormungandr/scenarios/helper_classes/complete_pt_journey.py
+++ b/source/jormungandr/jormungandr/scenarios/helper_classes/complete_pt_journey.py
@@ -74,29 +74,35 @@ def wait_and_build_crowflies(
     res = []
     for (dep_mode, arr_mode, future_pt_journey) in pt_journey_pool:
         logger.debug("waiting for pt journey starts with %s and ends with %s", dep_mode, arr_mode)
-        pt_journeys = wait_and_get_pt_journeys(future_pt_journey, has_valid_direct_paths)
+        try:
+            pt_journeys = wait_and_get_pt_journeys(future_pt_journey, has_valid_direct_paths)
 
-        if pt_journeys and pt_journeys.journeys:
-            origin_crowfly = {
-                "entry_point": requested_orig_obj,
-                "mode": dep_mode,
-                "places_free_access": orig_places_free_access.wait_and_get(),
-                "fallback_durations": orig_fallback_durations_pool.wait_and_get(dep_mode),
-                "fallback_type": StreetNetworkPathType.BEGINNING_FALLBACK,
-            }
+            if pt_journeys and pt_journeys.journeys:
+                origin_crowfly = {
+                    "entry_point": requested_orig_obj,
+                    "mode": dep_mode,
+                    "places_free_access": orig_places_free_access.wait_and_get(),
+                    "fallback_durations": orig_fallback_durations_pool.wait_and_get(dep_mode),
+                    "fallback_type": StreetNetworkPathType.BEGINNING_FALLBACK,
+                }
 
-            dest_crowfly = {
-                "entry_point": requested_dest_obj,
-                "mode": arr_mode,
-                "places_free_access": dest_places_free_acces.wait_and_get(),
-                "fallback_durations": dest_fallback_durations_pool.wait_and_get(arr_mode),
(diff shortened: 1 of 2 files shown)


source/jormungandr/jormungandr/scenarios/helper_classes/complete_pt_journey.py
```diff
@@ -74,29 +74,35 @@ def wait_and_build_crowflies(
...
+                    "places_free_access": dest_places_free_acces.wait_and_get(),
+                    "fallback_durations": dest_fallback_durations_pool.wait_and_get(arr_mode),
+                    "fallback_type": StreetNetworkPathType.ENDING_FALLBACK,
+                }
+
+                _build_crowflies(pt_journeys, origin_crowfly, dest_crowfly)
+
+                res.append(Pt_element(dep_mode, arr_mode, pt_journeys))
+        except PtException:
+            logger.exception(
```
kinnou02: It's normal that the PT planner doesn't find any solution for some request, it doesn't have to be logged.
Also we might want to have at least a dedicated exception in case of a no solution

source/jormungandr/jormungandr/scenarios/helper_classes/helper_exceptions.py
```diff
@@ -40,6 +40,15 @@ def get(self):
         return self._response
 
 
+class PtNoSolutionException(Exception):
```
kinnou02:
```suggestion
class PtNoSolutionException(PtException):
```

kinnou02 (changes requested):

source/jormungandr/jormungandr/scenarios/helper_classes/complete_pt_journey.py
```diff
@@ -50,7 +49,7 @@ def wait_and_get_pt_journeys(future_pt_journey, has_valid_direct_paths):
         if pt_journeys.error.id == response_pb2.Error.error_id.Value('no_solution') and has_valid_direct_paths:
             pt_journeys.ClearField(b"error")
         else:
-            raise PtException(pt_journeys)
+            raise PtNoSolutionException(pt_journeys)
```
nonifier: It seems strange that 'no solution' is treated as an exception. It's totally legit from the planner to return no journey ?

kinnou02:
Jenkins, test this please.

kinnou02 (approved):

kadhikari (approved):

AurelienLP (approved):

AurelienLP:
Jenkins, test this please.

nonifier (approved):

xlqian:
I forgot to launch the integration test after the last commit... a test was broken...

AurelienLP (approved):

kinnou02 (approved):

source/jormungandr/jormungandr/scenarios/helper_classes/complete_pt_journey.py
```diff
@@ -49,8 +48,10 @@ def wait_and_get_pt_journeys(future_pt_journey, has_valid_direct_paths):
     if pt_journeys and pt_journeys.HasField(b"error"):
         if pt_journeys.error.id == response_pb2.Error.error_id.Value('no_solution') and has_valid_direct_paths:
             pt_journeys.ClearField(b"error")
+        elif pt_journeys.error.id == response_pb2.Error.error_id.Value('date_out_of_bounds'):
+            raise InvalidDateBoundException(pt_journeys)
         else:
-            raise PtException(pt_journeys)
+            return None
```
mbounabi, line 54: Does that mean that, now, errors that aren't 'no_solution' or 'date_out_of_bounds' are ignored ?

kinnou02:
Jenkins, test this please.

source/jormungandr/jormungandr/scenarios/helper_classes/complete_pt_journey.py
xlqian: Yes! In this context, we want distributed to continue and return whatever it has found

PR #2895, opened 2019-07-24
Jormun: Expose frequency informations inside vehicle journeys

benoit-bst:
Related to https://github.com/CanalTP/navitia/pull/2873
and https://github.com/CanalTP/navitia/pull/2868

This is the last PR of the quest. 
- We expose now _frequency VJ_ informations into Jormun response
- Integration test are available :crossed_swords: 
- Updated the other test with new freq VJ (Now 4 Vjs)

**QA** part :

screenshot of a local test with **fr-auv** coverage

frequency_vj

### DIFF
--- a/source/jormungandr/tests/ptref_tests.py
+++ b/source/jormungandr/tests/ptref_tests.py
@@ -165,6 +165,15 @@ class TestPtRef(AbstractTestFixture):
         assert vjs[0]['end_time'] == '110000'
         assert vjs[0]['headway_secs'] == 600  # a 10 min step
 
+        # with a not frequency VJ
+        response = self.query_region('lines/line:A/vehicle_journeys?depth=3')
+        vjs = get_not_null(response, 'vehicle_journeys')
+
+        assert len(vjs) == 1
+        assert not 'start_time' in vjs[0]
+        assert not 'end_time' in vjs[0]
+        assert not 'headway_secs' in vjs[0]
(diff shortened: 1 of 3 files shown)


source/jormungandr/tests/ptref_tests.py
```diff
@@ -150,6 +150,21 @@ def test_vj_depth_3(self):
...
+            headway_secs
+        """
+        response = self.query_region('lines/line:freq/vehicle_journeys?depth=3')
+        vjs = get_not_null(response, 'vehicle_journeys')
+
+        assert len(vjs) == 1
+        assert vjs[0]['start_time'] == '100000'
+        assert vjs[0]['end_time'] == '110000'
+        assert vjs[0]['headway_secs'] == 600  # a 10 min step
+
```
kadhikari, line 167: What about a test of added attributs on a normal line (not frequency) ?

benoit-bst: Why not. I'll add it to be more precise

kinnou02 (approved):

source/jormungandr/tests/ptref_tests.py
benoit-bst: The test was very useful because it showed a malfunction with **headway_secs** value

kadhikari (approved):

PR #2901, opened 2019-07-29
Fix warning "extra ';'" for all BOOST_GLOBAL_FIXTURE

GuillaumeLescure:


pbougue (changes requested):
This just doesn't compile with clang if you change that.
```
navitia/source/calendar/tests/calendar_api_test.cpp:61:41: error: expected ';' after top level declarator
BOOST_GLOBAL_FIXTURE(logger_initialized)
                                        ^
                                        ;                                                                                                                                                                          
1 error generated.
```
So I think we have no choice but closing this PR. :disappointed: 

> Or it may be possible to add a line comment to have gcc ignore that warning...
> Or even deactivate that warning completely, but not fan of it.

GuillaumeLescure:
> This just doesn't compile with clang if you change that.
> 
> ```
> navitia/source/calendar/tests/calendar_api_test.cpp:61:41: error: expected ';' after top level declarator
> BOOST_GLOBAL_FIXTURE(logger_initialized)
>                                         ^
>                                         ;                                                                                                                                                                          
> 1 error generated.
> ```
> 
> So I think we have no choice but closing this PR. 
> 
> > Or it may be possible to add a line comment to have gcc ignore that warning...
> > Or even deactivate that warning completely, but not fan of it.

I didn't try with Clang but it shouldn't. Both follow the same compiler rules and a C/C++ file should be valid and invalid for both of the compilers. No differences should be noticeable. I investigate that.

benoit-bst:
I think this is the boost version. There is a break compatibility between two version. 
Lastest version don't need to `;`.
Try to compile inside debian 8 docker without the `;` for retrocompatibility

kinnou02:
> > This just doesn't compile with clang if you change that.
> > ```
> > navitia/source/calendar/tests/calendar_api_test.cpp:61:41: error: expected ';' after top level declarator
> > BOOST_GLOBAL_FIXTURE(logger_initialized)
> >                                         ^
> >                                         ;                                                                                                                                                                          
> > 1 error generated.
> > ```
> > 
> > 
> > So I think we have no choice but closing this PR.
> > > Or it may be possible to add a line comment to have gcc ignore that warning...
> > > Or even deactivate that warning completely, but not fan of it.
> 
> I didn't try with Clang but it shouldn't. Both follow the same compiler rules and a C/C++ file should be valid and invalid for both of the compilers. No differences should be noticeable. I investigate that.

this depends of the version of boost, the definition if the macro has changed at some point

GuillaumeLescure:
> I think this is the boost version. There is a break compatibility between two version.
> Lastest version don't need to `;`.
> Try to compile inside debian 8 docker without the `;` for retrocompatibility

I use Docker now (:sunglasses:) and thoses warnings came from the Debian8 image.

benoit-bst:
Okay, so now try with a modern docker like ubuntu 18.04 (with a modern boost lib). It'll probably failed because of Boost

GuillaumeLescure:
> Okay, so now try with a modern docker like ubuntu 18.04 (with a modern boost lib). It'll probably failed because of Boost

I will, but for the record, what is the policy about it ? We have to support both compiler with 2 differents versions or is there only 1 platform for development and others are bonus ?

benoit-bst:
>I will, but for the record, what is the policy about it ? We have to support both compiler with 2 differents versions or is there only 1 platform for development and others are bonus ?

Dunno bro. But I am sure that we have to support **Debian 8** environment because it is the prod env, and my OS **Ubuntu 18.04 and 19**, because I prefer :laughing:

GuillaumeLescure:
Ok so I close this PR.
The futur will need thoses ';' so it's a warning from the past :wink:
Good for me :+1:

PR #2902, opened 2019-07-30
jormun: restrict bragi autocompletion to the coverage

kinnou02:
When doing a /places on a coverage if no shape as been provided by the
user we use the shape of the coverage to prevent showing objects that
are outside of the coverage.
The shape of the coverage is simplified to speedup the search.
An evolution has to be made on bragi to not apply the shape's filter on
stop areas.
Refer to https://jira.kisio.org/browse/COR-19


kinnou02:
Jenkins, test this please.

crocme10 (approved):
Looks good

source/jormungandr/tests/authentication_tests.py
```diff
@@ -551,7 +551,7 @@ def test_places_authentication(self):
         # but when querying /v1/coverage/<something>/places only one pt_dataset is given to bragi
         with user_set(app, FakeUserAuth, 'super_user'):
             with requests_mock.Mocker() as m:
-                m.get(requests_mock.ANY, json=response)
+                m.post(requests_mock.ANY, json=response)
```
AurelienLP, line 554: Why is it a post now ?

AurelienLP (approved):

source/jormungandr/tests/authentication_tests.py
kinnou02: we use POST to send a shape in the body of the request

pbougue (approved):

PR #2894, opened 2019-07-24
Support private poi in tyr and jormungandr

crocme10:
* Adds a new binarisation step for poi2mimir
* Calls that binarisation for family_type 'poi'
* Extend instance to include private_poi_datataset
* Forward 'poi_dataset[]' to bragi

PR #2904, opened 2019-07-30
Process launch refactoring

GuillaumeLescure:
- [ ] Need #2894


kinnou02:
what is the point of this change?

GuillaumeLescure:
> what is the point of this change?

The points are :
 - naming : `launch_exec.launch_exec` become `process.run`
 - clean and simplification : remove all the dead code for the unused parameter `traces`

kinnou02:
> > what is the point of this change?
> 
> The points are :
> 
> * naming : `launch_exec.launch_exec` become `process.run`
> * clean and simplification : remove all the dead code for the unused parameter `traces`

By simplification you mean deletion of some feature like:
  - handling of log level
  - logging of stderr
  - and maybe the parsing of lines

And obviously you did run all tasks of tyr to check that everything still works as expected?

kinnou02:
if this is merged it must be after #2894

GuillaumeLescure:
> By simplification you mean deletion of some feature like:
> 
> * handling of log level

Only in the "traces" parameter that was unused and now removed
> * logging of stderr

Maybe I'm wrong but if we redirect the stderr in the stdout and then the stdout in the pipe, we should have everything. Am I missing something ?
> * and maybe the parsing of lines

Only used by the "traces" parameter that was unused and now removed
> 
> And obviously you did run all tasks of tyr to check that everything still works as expected?

Yes everything seems to be ok. I just re-test them because of the rebase, `make test` and `make docker_test` are ok :smiley:

PR #2908, opened 2019-07-31
Jormun: Add cycle lane length

AurelienLP:
We want to know the percentage of cycle path used for a bike journey.
For now, only asgard is able to retrieve these information. The aim of this PR is to compute and expose these information in the section part of the navitia response

### DIFF
--- a/source/jormungandr/jormungandr/interfaces/v1/serializer/journey.py
+++ b/source/jormungandr/jormungandr/interfaces/v1/serializer/journey.py
@@ -312,14 +312,7 @@ class SectionSerializer(PbNestedSerializer):
             return None
         return JourneySerializer(obj.ridesharing_journeys, display_none=False, many=True).data
 
-    cycle_lane_length = jsonschema.MethodField(
-        schema_type=int, display_none=False, description='Length of the section on cycle lanes'
-    )
-
-    def get_cycle_lane_length(self, obj):
-        if not hasattr(obj, 'cycle_lane_length') or not obj.cycle_lane_length:
-            return None
-        return obj.cycle_lane_length
+    cycle_lane_length = PbIntField(display_none=False)
 
 
 class JourneySerializer(PbNestedSerializer):
--- a/source/jormungandr/jormungandr/interfaces/v1/serializer/journey.py
+++ b/source/jormungandr/jormungandr/interfaces/v1/serializer/journey.py
@@ -251,9 +251,6 @@ class SectionSerializer(PbNestedSerializer):
     )
     data_freshness = EnumField(attr="realtime_level", pb_type=RTLevel, display_none=False)
     to = jsonschema.MethodField(schema_type=PlaceSerializer(), attr='destination')
-    cycle_path_percentage = jsonschema.Field(
-        schema_type=float, display_none=True, description='Percentage of cycle path used'
-    )
 
     def get_to(self, obj):
         if obj.HasField(str('type')):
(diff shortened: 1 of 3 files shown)


source/jormungandr/tests/direct_path_asgard_integration_tests.py
```diff
@@ -140,6 +160,7 @@ def test_journey_with_bike_direct_path(self):
         assert response['journeys'][0]['durations']['car'] == 500
         assert response['journeys'][0]['durations']['total'] == 500
         assert response['journeys'][0]['distances']['car'] == 50
+        assert response['journeys'][0]['sections'][0]['cycle_path_percentage'] == 0.0
```
benoit-bst: Why it is present inside a car section ? The field is not optional ?

benoit-bst (approved):

source/jormungandr/tests/direct_path_asgard_integration_tests.py
AurelienLP: Fixed that! Thank you :)

source/jormungandr/tests/direct_path_asgard_integration_tests.py
```diff
@@ -149,6 +170,7 @@ def test_journey_with_bike_direct_path(self):
         assert response['journeys'][1]['durations']['total'] == 1000
         assert response['journeys'][1]['distances']['bike'] == 100
         assert response['journeys'][1]['duration'] == 1000
+        assert response['journeys'][1]['sections'][0]['cycle_lane_length'] == 30.0
```
mbounabi: It is cast as 'int' in 'add_cycle_lane_length', isn't it?

source/jormungandr/jormungandr/interfaces/v1/serializer/journey.py
```diff
@@ -312,6 +312,15 @@ def get_ridesharing_journeys(self, obj):
             return None
         return JourneySerializer(obj.ridesharing_journeys, display_none=False, many=True).data
 
+    cycle_lane_length = jsonschema.MethodField(
+        schema_type=int, display_none=False, description='Length of the section on cycle lanes'
+    )
+
+    def get_cycle_lane_length(self, obj):
+        if not hasattr(obj, 'cycle_lane_length') or not obj.cycle_lane_length:
```
pbougue: minor: Don't you want to output `0` when you have the information and you know it's 0 ?
With this it's not possible to differentiate `0` vs. `no information available/accurate`.
Don't know if it's easily doable to differentiate though, depending on the outputs of connectors (and what protobuf allows).

source/jormungandr/tests/direct_path_asgard_integration_tests.py
pbougue: minor (for readability): So shouldn't you have :
```suggestion
        assert response['journeys'][1]['sections'][0]['cycle_lane_length'] == 30
```

pbougue (approved):
:+1: I guess geovelo will be the next one

source/jormungandr/jormungandr/interfaces/v1/serializer/journey.py
AurelienLP: That's the behavior I wanted. Did it with `PbIntField(display_none=False)`

benoit-bst (approved):

xlqian (approved):

mbounabi (approved):

PR #2910, opened 2019-08-01
PTRef: Update grammar doc and behavior of param data_freshness for since/until

pbougue:
After PR https://github.com/CanalTP/navitia/pull/2878
JIRA https://jira.kisio.org/browse/NAVP-1273

Documenting the add of `data_freshness` param to the `since`/`until` methods of PTRef **grammar**.
Also:
* Renaming RTLevels in **grammar** to the ones used in jormun, as they are exposed
* Removing from **grammar** the mandatory `data_freshness` param on disruptions (as it does not make sense to have it on anything else than VJ).

:mag: can be reviewed by commit

### DIFF
--- a/source/ptreferential/ptreferential_ng.cpp
+++ b/source/ptreferential/ptreferential_ng.cpp
@@ -254,7 +254,7 @@ struct Eval : boost::static_visitor<Indexes> {
     Indexes operator()(const ast::Empty&) const { return Indexes(); }
     Indexes operator()(const ast::Fun& f) const {
         Indexes indexes;
-        const Type_e type = type_by_caption(f.type);
+        const auto type = type_by_caption(f.type);
         if (type == Type_e::VehicleJourney && f.method == "has_headsign" && f.args.size() == 1) {
             for (auto vj : data.pt_data->headsign_handler.get_vj_from_headsign(f.args.at(0))) {
                 indexes.insert(vj->idx);
@@ -313,10 +313,9 @@ struct Eval : boost::static_visitor<Indexes> {
                 since = from_datetime(f.args.at(0));
                 until = from_datetime(f.args.at(1));
             }
-            auto rt_level = type::RTLevel::Base;  // useful only for VJ
-            if (type == Type_e::VehicleJourney) {
-                rt_level = rt_level_from_string(f.args.back());
-            }
+            // useful only for VJ
+            const auto rt_level =
+                (type == Type_e::VehicleJourney) ? rt_level_from_string(f.args.back()) : type::RTLevel::Base;
             indexes = filter_on_period(data.get_all_index(type), type, since, until, rt_level, data);
 
         } else if (f.method == "within" && f.args.size() == 2) {
@@ -375,12 +374,9 @@ struct Eval : boost::static_visitor<Indexes> {
 
 private:
     // helper to add required param to methods since(), until() and between().
-    static size_t nb_extra_args_between(const type::Type_e& type) {
+    size_t nb_extra_args_between(const type::Type_e& type) const {
         // for VehicleJourney, the data_freshness level is also required, so 1 more param is required
-        if (type == type::Type_e::VehicleJourney) {
-            return 1u;
-        }
-        return 0u;
+        return (type == type::Type_e::VehicleJourney) ? 1u : 0u;
     }
 };
 
--- a/source/ptreferential/ptreferential_ng.cpp
+++ b/source/ptreferential/ptreferential_ng.cpp
@@ -295,9 +295,15 @@ struct Eval : boost::static_visitor<Indexes> {
         } else if (type == Type_e::Impact
                    && ((f.method == "tag" && f.args.size() == 1) || (f.method == "tags" && f.args.size() >= 1))) {
             indexes = get_impacts_by_tags(f.args, data);
-        } else if ((f.method == "since" && f.args.size() == 1 + (f.type == "vehicle_journey"))
-                   || (f.method == "until" && f.args.size() == 1 + (f.type == "vehicle_journey"))
-                   || (f.method == "between" && f.args.size() == 2 + (f.type == "vehicle_journey"))) {
+        }
+        // since/until/between are available for disruption and vehicle_journey,
+        // but VJ requires the additional param `data_freshness`
+        // so below grammar provides the following methods:
+        // * disruption.between(since, until)
+        // * vehicle_journey.between(since, until, data_freshness)
+        else if ((f.method == "since" && f.args.size() == 1u + nb_extra_args_between(type))
+                 || (f.method == "until" && f.args.size() == 1u + nb_extra_args_between(type))
+                 || (f.method == "between" && f.args.size() == 2u + nb_extra_args_between(type))) {
             boost::optional<boost::posix_time::ptime> since, until;
             if (f.method == "since") {
                 since = from_datetime(f.args.at(0));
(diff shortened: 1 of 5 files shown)


source/ptreferential/ptreferential_ng.cpp
```diff
@@ -294,22 +294,23 @@ struct Eval : boost::static_visitor<Indexes> {
         } else if (f.type == "disruption"
                    && ((f.method == "tag" && f.args.size() == 1) || (f.method == "tags" && f.args.size() >= 1))) {
             indexes = get_impacts_by_tags(f.args, data);
-        } else if ((f.method == "since" && f.args.size() == 2) || (f.method == "until" && f.args.size() == 2)
-                   || (f.method == "between" && f.args.size() == 3)) {
+        } else if ((f.method == "since" && f.args.size() == 1 + (f.type == "vehicle_journey"))
```
benoit-bst: So, now `since, until and between` are only available for vehicle_journey type ?

source/ptreferential/ptreferential_ng.cpp
```diff
@@ -294,22 +294,23 @@ struct Eval : boost::static_visitor<Indexes> {
...
+                   || (f.method == "until" && f.args.size() == 1 + (f.type == "vehicle_journey"))
```
benoit-bst: you use a`+` in a logic condition ? isn't it `&&` ?

benoit-bst (approved):

source/ptreferential/ptreferential_ng.cpp
pbougue: No, they are still available for VJ and Disruption (for other PTobjects its accepted here but rejected after, inside `filter_on_periodfilter_on_period` like it was the case before).

Here the point is for VJ we check that there is one more arg (the `data_freshness`), while there is not for others.

But it may not be super clear from the code, I'm open to suggestion if you think it should be rewritten/commented.

pbougue (commented):
Previously, data_freshness param was added to VJ.since/until **and** Disruption.since/until (#2878).

This PR maintains the 3 since/until-methods for both.
This PR maintains the data_freshness param for VJ, but **removes** it for Disruption.

source/ptreferential/ptreferential_ng.cpp
pbougue: Actually it's on purpose, I implicitly (:see_no_evil:) cast the bool to an int to add one more mandatory arg.
Maybe I should cast explicitly and add a comment ?

benoit-bst: I don't know how to change the condition in a `real logic condition style`, but it is not obvious to read.
You should probably find a way.

nonifier: TBH, I don't quite understand what's happening here by reading the code :(
What exactly are you trying to express here ?

nonifier: Ok, I think I got it. It means that you expect one more argument in the case of a `vehicle_journey` isn't it ?

source/ptreferential/ptreferential_ng.cpp
benoit-bst: okay I didn't understand because of the `+`. In my reading, I interpreted that like a `&&` (like a _reflex_)

source/ptreferential/ptreferential_ng.cpp
```diff
@@ -254,20 +254,21 @@ struct Eval : boost::static_visitor<Indexes> {
     Indexes operator()(const ast::Empty&) const { return Indexes(); }
     Indexes operator()(const ast::Fun& f) const {
         Indexes indexes;
-        if (f.type == "vehicle_journey" && f.method == "has_headsign" && f.args.size() == 1) {
+        const Type_e type = type_by_caption(f.type);
```
AurelienLP: `const auto type = type_by_caption(f.type);`

source/ptreferential/ptreferential_ng.cpp
```diff
@@ -291,25 +292,31 @@ struct Eval : boost::static_visitor<Indexes> {
...
             } else if (f.method == "until") {
                 until = from_datetime(f.args.at(0));
-                rt_level = rt_level_from_string(f.args.at(1));
             } else if (f.method == "between") {
                 since = from_datetime(f.args.at(0));
                 until = from_datetime(f.args.at(1));
-                rt_level = rt_level_from_string(f.args.at(2));
             }
-            const auto type = type_by_caption(f.type);
+            auto rt_level = type::RTLevel::Base;  // useful only for VJ
```
AurelienLP: `auto const rt_level = type == Type_e::VehicleJourney ? type::RTLevel::Base : rt_level_from_string(f.args.back())`

`rt_level` can be const this way.

source/ptreferential/ptreferential_ng.cpp
```diff
@@ -365,6 +372,16 @@ struct Eval : boost::static_visitor<Indexes> {
         return res;
     }
     Indexes operator()(const ast::Expr& expr) const { return boost::apply_visitor(*this, expr.expr); }
+
+private:
+    // helper to add required param to methods since(), until() and between().
+    static size_t nb_extra_args_between(const type::Type_e& type) {
```
AurelienLP: Why static ?

source/ptreferential/ptreferential_ng.cpp
```diff
@@ -365,6 +372,16 @@ struct Eval : boost::static_visitor<Indexes> {
...
+        // for VehicleJourney, the data_freshness level is also required, so 1 more param is required
+        if (type == type::Type_e::VehicleJourney) {
```
AurelienLP: return type == type::Type_e::VehicleJourney ? 1u : 0u

source/ptreferential/ptreferential_ng.cpp
```diff
@@ -412,16 +429,19 @@ std::string make_request(const Type_e requested_type,
...
-            } else if (since) {
-                res = "(" + res + ") AND " + static_data->captionByType(requested_type) + ".since("
-                      + to_iso_string(*since) + "Z, " + rt_level_to_string(rt_level) + ")";
-            } else if (until) {
-                res = "(" + res + ") AND " + static_data->captionByType(requested_type) + ".until("
-                      + to_iso_string(*until) + "Z, " + rt_level_to_string(rt_level) + ")";
+            if (since || until) {
+                res = "(" + res + ") AND " + static_data->captionByType(requested_type);
+                if (since && until) {
+                    res = res + ".between(" + to_iso_string(*since) + "Z, " + to_iso_string(*until) + "Z";
```
AurelienLP, line 431: `res += ...`

source/ptreferential/ptreferential_ng.cpp
benoit-bst: don't need `static`. It is just a private function

benoit-bst (approved):

source/ptreferential/ptreferential_ng.cpp
pbougue: Then the `"blabla"` is a const char*, no operator+ available, needs a cast.
So it's shorter and more readable currently to me :neutral_face:

source/ptreferential/ptreferential_ng.cpp
pbougue: Yes, just add `const`

AurelienLP (approved):

kinnou02 (approved):

PR #2912, opened 2019-08-02
Tyr: Add required default value before downgrade migration

mbounabi:
https://jira.kisio.org/browse/NAVP-1302

To avoid a crash when downgrading db with value `nullable=True` to `nullable=False`, add a default value.
Crash not seen in tests because every rows are deleted before downgrade. 
Fix tested manually


pbougue (approved):

AurelienLP (approved):

AurelienLP:
Jenkins, test this please.

PR #2913, opened 2019-08-02
ptreferential: fix warning "suggest parentheses around arithmetic"

GuillaumeLescure:


benoit-bst (approved):
if tests pass...

GuillaumeLescure:
> if tests pass...

They do ... `make test` and `make docker_test` are ok !

AurelienLP (approved):

pbougue (approved):
OK, I tested locally that with the wrong parentheses some tests do break.
So it looks like it's tested enough :man_shrugging:

PR #2919, opened 2019-08-06
Tyr: do not reimport data in mimir during kraken binarisation

kinnou02:
`import_last_dataset` is use during deployment to rebuild new
*data.nav.lz4* if a binarisation is required, during this step there is
no need to reindex those datasets to mimir as navitia deployments
doesn't ship any of the mimir executable.
So to not lose time during deployment this PR disable the reindexation
of pt dataset in mimir.

### DIFF
--- a/source/tyr/tyr/command/import_last_dataset.py
+++ b/source/tyr/tyr/command/import_last_dataset.py
@@ -36,7 +36,7 @@ from tyr.helper import get_instance_logger, wait_or_raise
 
 @manager.command
 def import_last_dataset(
-    instance_name, background=False, reload_kraken=False, custom_output_dir=None, nowait=False
+    instance_name, background=False, reload_kraken=False, custom_output_dir=None, nowait=False, allow_mimir=False
 ):
     """
     reimport the last dataset of a instance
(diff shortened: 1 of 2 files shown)


source/tyr/tyr/command/import_last_dataset.py
```diff
@@ -49,6 +49,9 @@ def import_last_dataset(
 
     """
     instance = models.Instance.query_existing().filter_by(name=instance_name).first()
+    # little trick to keep a flag on the command line: -a enable mimir import,
+    # we reverse it to match import_data
+    skip_mimir = not allow_mimir
```
AurelienLP, line 54: Minor: `need_mimir` would be more understandable than `allow_mimir` for me

AurelienLP (approved):

kinnou02:
Jenkins, test this please.

source/tyr/tyr/command/import_last_dataset.py
nonifier: or maybe just use what it's in the comment: `enable_mimir_import`?
So you can have its equivalent: `disable_mimir_import = not enable_mimir_import` :)

pbougue (approved):

source/tyr/tyr/command/import_last_dataset.py
kinnou02: I didn't want to use *enable* as it doesn't enable anything, if mimir isn't setup passing this parameter won't do anything

nonifier: Ok, I see where you're going. Can't find anything better unfortunately :(

nonifier (approved):

PR #2922, opened 2019-08-07
Jormun: Update status

AurelienLP:


source/jormungandr/jormungandr/interfaces/v1/__init__.py
```diff
@@ -49,8 +49,18 @@ def add_common_status(response, instance):
         response['status']['realtime_proxies'].append(realtime_proxy.status())
 
     response['status']['street_networks'] = []
-    for sn in instance.get_all_street_networks():
-        response['status']['street_networks'].append(sn.status())
+    street_networks = instance.get_all_street_networks()
+    # street_networks is a list if it comes from the Json
+    if type(street_networks) == list:
```
xlqian: `isinstance` is better than `type`

xlqian (approved):

AurelienLP:
Jenkins, test this please.

source/jormungandr/jormungandr/interfaces/v1/__init__.py
kinnou02: it's best to check that it's some abstract type with https://docs.python.org/2.7/library/collections.html#collections-abstract-base-classes

source/jormungandr/jormungandr/interfaces/v1/serializer/status.py
```diff
@@ -119,6 +119,11 @@ class StreetNetworkSerializer(OutsideServiceCommon):
     def get_timeout(self, obj):
         return obj.get('timeout', None)
 
+    backend_class = MethodField(schema_type=str, display_none=False)
```
kinnou02: You should put it above `get_timenout()` to improve readability

kinnou02 (approved):

source/jormungandr/jormungandr/interfaces/v1/__init__.py
AurelienLP: Here, I know for certain that this is a list. Should I still check if it is an abstract class?

source/jormungandr/jormungandr/interfaces/v1/serializer/status.py
AurelienLP: `get_backend_class()` to or just the field?

source/jormungandr/jormungandr/interfaces/v1/__init__.py
kinnou02: it'r better to check for an abstract class as if you change implementation of `get_all_street_networks()` to return another kind of container, like a set or some kind of readonly list by example your code will still works.

source/jormungandr/jormungandr/interfaces/v1/serializer/status.py
kinnou02: just the field, to group them

source/jormungandr/jormungandr/interfaces/v1/__init__.py
```diff
@@ -49,8 +51,17 @@ def add_common_status(response, instance):
         response['status']['realtime_proxies'].append(realtime_proxy.status())
 
     response['status']['street_networks'] = []
-    for sn in instance.get_all_street_networks():
-        response['status']['street_networks'].append(sn.status())
+    street_networks = instance.get_all_street_networks()
```
kinnou02: The output of a method should always be the same type, whatever the configuration is

source/jormungandr/jormungandr/street_network/tests/streetnetwork_backend_manager_test.py
```diff
@@ -261,3 +262,26 @@ def wrong_streetnetwork_backend_test():
     manager._update_config("instance")
     assert not manager._streetnetwork_backends_by_instance_legacy
     assert not manager._streetnetwork_backends
+
+
+class FakeInstance:
```
kinnou02: This should inherit from Instance

kinnou02 (changes requested):

kinnou02 (approved):

xlqian (approved):

PR #2923, opened 2019-08-07
[Tests] Chaos disruption tests

nonifier:
## TL;DR
Long time awaited tests that cover Chaos disruptions database access ! 

## How does this works:
The new test is part of `docker_test` suite (eg. `make docker_test`), as it  shares the same dependencies. 

You can also run the test separately with `make chaos_tests`  

The test is a pytest that will pop a Postgres database in a Docker. The db's content comes from the sql file `source/tests/chaos/chaos_loading.sql`. Once the database is ready, a subprocess is triggered to call on `chaos_db_tests` a boost's test suite that will read the disruptions from the database and expect on the output `chaos::Disruption` protobuf objects.


source/kraken/fill_disruption_from_database.h
```diff
@@ -53,13 +53,14 @@ namespace navitia {
     FILL_REQUIRED(table_name, created_at, uint64_t) \
     FILL_NULLABLE(table_name, updated_at, uint64_t)
 
-void fill_disruption_from_database(const std::string& connection_string,
-                                   navitia::type::PT_Data& pt_data,
-                                   navitia::type::MetaData& meta,
-                                   const std::vector<std::string>& contributors);
+using Chaos_disruption_applier = std::function<
```
kinnou02: You have to choose between CamelCase and snake_case, since it's a type I would say CamelCase

source/tests/chaos/CMakeLists.txt
```diff
@@ -0,0 +1,26 @@
+#
+## Setup the BOOST tests that will exercise Chaos database
+#
+include_directories("..")
```
kinnou02: ?

source/tests/chaos/Pipfile
```diff
@@ -0,0 +1,13 @@
+[[source]]
+name = "pypi"
+url = "https://pypi.org/simple"
```
kinnou02: should we mix `Pipfile` and `requirements.txt` inside one repository?

nonifier: It's_now_Been-removed.

kinnou02 (approved):

source/tests/chaos/CMakeLists.txt
```diff
@@ -0,0 +1,24 @@
...
+set(CHAOS_TESTS_PATH "${CMAKE_CURRENT_SOURCE_DIR}")
+set(CHAOS_DB_TESTS_BUILD_DIR "${CMAKE_CURRENT_BINARY_DIR}")
+add_custom_target(
+    chaos_tests
+    DEPENDS chaos_db_tests
+)
+add_custom_command(
+    TARGET chaos_tests
+    WORKING_DIRECTORY ${CHAOS_TESTS_PATH}
+    COMMAND CHAOS_DB_TESTS_BUILD_DIR=${CHAOS_DB_TESTS_BUILD_DIR} py.test ${CHAOS_TESTS_PATH}
```
pbougue: PostgresDocker import is failing:
It misses `PYTHONPATH=${CMAKE_CURRENT_SOURCE_DIR}:${CMAKE_SOURCE_DIR}/navitiacommon` here

source/tests/chaos/chaos_db_tests.cpp
```diff
@@ -0,0 +1,150 @@
...
+
+    BOOST_REQUIRE_EQUAL(disruptions.size(), 5);
+    BOOST_CHECK_EQUAL(disruptions[0].id(), "095324ea-9370-11e9-9678-005056a40962");
+    BOOST_CHECK_EQUAL(disruptions[1].id(), "76c5f3b2-936c-11e9-9678-005056a40962");
+    BOOST_CHECK_EQUAL(disruptions[2].id(), "8a1c6cf0-8d15-11e9-b1ac-005056a40962");
+    BOOST_CHECK_EQUAL(disruptions[3].id(), "bd3ee57a-968b-11e9-951b-005056a40962");
+    BOOST_CHECK_EQUAL(disruptions[4].id(), "e14423e0-90d7-11e9-a640-005056a40962");
+
+    auto& disruption = disruptions[0];
+    BOOST_CHECK_EQUAL(disruption.reference(), "msa-20190620172619");
```
pbougue, line 91: Is it possible to explain what's tested here?
Maybe also a PR comment to link to the PR that was correcting this.

Did you try those tests without the fix?

pbougue (changes requested):
:+1: After a quick test locally, it should work fine, and it should be possible to know quickly what test is broken.
So those tests are automatable :tada: Great job! 

In addition to comments:

* After the pipenv removal, it misses a requirements_dev.txt for these tests (probably getting what's usefull from tyr/requirements_dev.txt, as it worked with it locally).
* It also misses a readme to explain how to update/create sql file.
* And maybe a readme update to explain how to launch it: better to launch directly it in travis if possible ?

source/tests/chaos/chaos_db_tests.cpp
```diff
@@ -0,0 +1,150 @@
...
+    metadata.publisher_url = "publisher_url";
+    metadata.license = "license";
+    metadata.instance_name = "instance_name";
+    metadata.dataset_created_at = "20010101T010203"_dt;
+    metadata.poi_source = "poi_source";
+    metadata.street_network_source = "kraken";
+
+    std::vector<chaos::Disruption> disruptions;
+
+    auto disruption_callback = [&disruptions](const chaos::Disruption& d, navitia::type::PT_Data&,
```
xlqian, line 70: :+1:

xlqian (approved):

source/tests/chaos/CMakeLists.txt
nonifier: well spotted ! I've had a `PYTHONPATH` with `navitiacommon` exported on my dev env...

source/tests/chaos/chaos_db_tests.cpp
nonifier: Nope I didn't try this without the fix as it was done quite some time ago and I would have diverged significantly unfortunately. 

Also, I wanted here to test the order of the disruptions that was affected by the fix. You can see that with `BOOST_CHECK_EQUAL(disruptions[0/1/2/3/4].id(), "idxxxx"`

I also wanted to make sure we have assertions on the output protobuf, and ensure objects are constructed well with no sub-objects missing. That what you have in the second part of the test when we check all the fields from `disruptions[0]`.

nonifier:
@pbougue :

> After the pipenv removal, it misses a requirements_dev.txt for these tests (probably getting what's usefull from tyr/requirements_dev.txt, as it worked with it locally).

I've added a readme explaining that we have the same dependencies as the `docker_test`, and how to run those tests separately.  

---

> It also misses a readme to explain how to update/create sql file.

There is no easy way to create/update the sql file as it comes from a dump of a real database. We couldn't think of something easier to maintain this with intelligible values unfortunately.   

---

> And maybe a readme update to explain how to launch it: better to launch directly it in travis if possible ?

I've added a `readme.md` with some explanation. I can complete it if you think I've missed something that might be of interest. 
Regarding Travis, I tried to get it under CI, but building `chaos_db_tests` would required most of the cpp dependencies (log4cpp, boost etc...) and quite frankly, I just couldn't be bothered :(

source/tests/chaos/readme.md
```diff
@@ -0,0 +1,16 @@
+# Chaos Tests
+
+This aims at testing the Navitia's dependency on [Chaos](https://github.com/CanalTP/chaos)
+
+## How to run
+Those tests are part of `docker_test` test suite, hence depend on the same requirements.
+If you haven't set any environment just yet, create a new virtual env with :
+
+```sh
+pip install -r ../../tyr/requirements.txt -r ../../eitri/requirements.txt
```
pbougue: You have to use tyr/requirements`_dev`.txt for pytest (and after a local test, eitri/requirements.txt isn't needed for chaos_tests).

So you can pick any of the 2 for the doc:
For a specific chaos_tests venv
```suggestion
pip install -r ../../tyr/requirements_dev.txt
```
Or for a whole docker_test venv
```suggestion
pip install -r ../../tyr/requirements_dev.txt -r ../../eitri/requirements.txt
```

pbougue (approved):
One big step for the tests! :footprints: 
:ok_hand: Nice work, and one can later comment or plug it to travis if needed

- [x] Please just change the tyr requirements`_dev` to use.

As for the maintenance, OK then I guess one would have to load the SQL dump, then migrate it if it requires some (using https://github.com/CanalTP/Chaos/tree/master/migrations), then handcraft it to add some rows :white_check_mark:

nonifier:
jenkins, please test this.

nonifier:
Jenkins, test this please

PR #2925, opened 2019-08-08
Revert "Jormun: journey filter sort more consistent with kraken journey sort"

pbougue:
Reverts CanalTP/navitia#2917

Previous PR gave too much importance to the walking part of transfers and changed too much journeys being kept by the filter.

JIRA ticket to investigate if something smarter should be done in jormun filters and in kraken raptor : https://jira.kisio.org/browse/NAVP-1389

### DIFF
--- a/source/jormungandr/jormungandr/scenarios/journey_filter.py
+++ b/source/jormungandr/jormungandr/scenarios/journey_filter.py
@@ -457,13 +457,11 @@ def shared_section_generator(journey):
             yield "origin:{}/dest:{}".format(s.origin.uri, s.destination.uri)
 
 
-def fallback_duration(journey, transfer_penalty):
+def fallback_duration(journey):
     duration = 0
     for section in journey.sections:
-        if section.type in (response_pb2.STREET_NETWORK, response_pb2.CROW_FLY, response_pb2.TRANSFER):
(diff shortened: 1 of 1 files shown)


source/jormungandr/jormungandr/scenarios/journey_filter.py
```diff
@@ -513,7 +511,7 @@ def _datetime_to_str(ts):
             dep=_datetime_to_str(journey.departure_date_time),
             arr=_datetime_to_str(journey.arrival_date_time),
             duration=datetime.timedelta(seconds=journey.duration),
-            fallback=datetime.timedelta(seconds=fallback_duration(journey, 0)),
+            fallback=datetime.timedelta(seconds=fallback_duration(journey)),
```
pbougue, line 514: So the debug fallback walking duration will be back to not considering transfers.
No major issue on that.

kinnou02 (approved):

pbougue (commented):
To have an idea on consequences:
filterTransferIssue
In debug=true, the 2 journeys are returned by Kraken (only transfer differs).
The one that we returned initially (and will be back after the merge of the revert) changes as soon as possible (more waiting time which is nice but also more walking).
The competitor goes one stop farther to change just to save walking (less elegant at least).

PR #2927, opened 2019-08-09
[kraken]: remove inversed variable

xlqian:
Using the inverted_nb_vertex_by_mode actually caused some exceptions... 

Since the nb of vertices is bigger than 150000 in the coverage of Spain, this trick can no longer hold true....

Too bad... I have to use the division which surprisingly doesn't have too much impact on the performance of distributed...

### DIFF
--- a/source/georef/georef.h
+++ b/source/georef/georef.h
@@ -250,8 +250,6 @@ struct GeoRef {
     /// number of vertex by transportation mode
     nt::idx_t nb_vertex_by_mode = 0;
 
-    // this number is used to avoid division operations
-    float inversed_nb_vertex_by_mode = 0;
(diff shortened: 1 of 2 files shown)


source/georef/georef.h
```diff
@@ -250,8 +250,6 @@ struct GeoRef {
     /// number of vertex by transportation mode
     nt::idx_t nb_vertex_by_mode = 0;
 
-    // this number is used to avoid division operations
-    float inversed_nb_vertex_by_mode = 0;
```
xlqian, line 254: Nah... it breaks the "big" coverages...

AurelienLP (approved):

nonifier (approved):

PR #2934, opened 2019-08-29
 Doc: update error list

pbougue:
JIRA: https://jira.kisio.org/browse/NAVITIAII-2836

* Update to match https://github.com/CanalTP/navitia/blob/dev/source/jormungandr/jormungandr/interfaces/v1/errors.py
* Also add a word to `ed` doc on db setup

### DIFF
--- a/documentation/slate/source/includes/errors.md
+++ b/documentation/slate/source/includes/errors.md
@@ -2,7 +2,7 @@
 
 Code 40x
 --------
-> When navitia is unable to give an answer, you'll receive a response with an error object
+When Navitia is unable to give an answer, you'll receive a response with an error object
 containing a unique error id, in a 4XX http code response
 
 ```shell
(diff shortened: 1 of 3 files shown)


source/ed/readme.md
```diff
@@ -25,6 +26,24 @@ example: ```--connection-string="host=localhost user=my_user password=my_pwd dbn
 
 Note: importing data will remove the data of the specified type (i.e. osm2ed and geopal2ed will remove street network data).
 
+## Setup `ed` database
+
+```bash
+sudo -i -u postgres
```
nonifier, line 32: You might want to create a user if it's missing. Otherwise the db creation will fail.

```shell
sudo -u postgres createuser my_user
sudo -u postgres createdb -O my_user my_db
sudo -u postgres psql -d my_db -c "CREATE EXTENSION postgis;"
```

nonifier (approved):

source/ed/readme.md
pbougue: Yep and it has to have creation right, I hesitated to put the following (but thought it's more postgresql doc)

```bash
CREATE USER my_user;
ALTER ROLE my_user WITH CREATEDB;
ALTER USER my_user WITH ENCRYPTED PASSWORD 'my_pwd';
```

I'm OK to add something like that if you guys think it's worthy.

> Final word after discussion is: we let it as-is, to be improved later if needed

kinnou02 (approved):

documentation/slate/source/includes/errors.md
```diff
@@ -24,23 +24,39 @@ The are two possible 40x http codes :
...
+
+
+Code 200
+--------
+Even if everything is correct (input parameters, internal processing)
+Navitia can end up finding that the correct answer is empty.
+
+| Error id                    | Description                                                                |
+|-----------------------------|----------------------------------------------------------------------------|
+| no_solution                 | From data, there is no solution available given the request                |
```
benoit-bst, line 52: Thanks :+1:

benoit-bst (approved):

PR #2935, opened 2019-08-30
Ed: protect more osm2ed from bad shapes

pbougue:
JIRA: https://jira.kisio.org/browse/NAVITIAII-2843

Issue is on https://www.openstreetmap.org/relation/8522033,
in particular this `outer` thats just a line (pointA - pointB - pointA): https://www.openstreetmap.org/way/615974857

- [x] Hand-tested with complete files for ca-bc, ca-qc and fr from geofabrik

Nota:
* Using area(), as boost::geometry::is_valid() is only available from boost 1.56.00
* Did not mutualize code for cities as it is deprecated (in favour of cosmogony)
* No test added as nothing is tested on osm2ed and it's too expensive to add some for the gain

### DIFF
--- a/source/ed/osm2ed.cpp
+++ b/source/ed/osm2ed.cpp
@@ -727,13 +727,13 @@ void OSMAdminRelation::build_polygon(OSMCache& cache) {
             }
             next_node = next_way->nodes.back();
         }
+        auto log = log4cplus::Logger::getInstance("log");
         if (tmp_polygon.outer().size() < 2 || ref == references.end()) {
             // add some logs
             // some admins are at the boundary of the osm data, so their own boundary may be incomplete
             // some other admins have a split boundary and it is impossible to compute it.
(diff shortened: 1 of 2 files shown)


pbougue:
Jenkins, test this please.

GuillaumeLescure (approved):
Is it possible to make a test for that ?

> As mentionned in main comment, it's too expensive

source/ed/osm2ed.cpp
```diff
@@ -754,6 +754,18 @@ void OSMAdminRelation::build_polygon(OSMCache& cache) {
             }
             break;
         }
+
+        if (boost::geometry::area(tmp_polygon) < std::numeric_limits<double>::epsilon()) {
```
nonifier, line 758: Duplicated code spotted ! Would be nice to have a common function that does this.

benoit-bst (approved):

nonifier (approved):

PR #2937, opened 2019-09-03
[Raptor] Lex walking duration

pbench:
Update the raptor algorithm to remember the walking duration corresponding to the earliest arrival time found.

More precisely, the raptor algorithm computes for each pair (_stop_point_, _nb_of_transfer_) the earliest arrival time at _stop_point_  using _nb_of_transfer_  change of vehicle ([example](https://github.com/CanalTP/navitia/blob/dev/documentation/rfc/doc-routing-algorithm.md#the-first-pass-on-example-1)).
Now, it also stores the walking duration of a journey that achieves this earliest arrival time.

The first commits are logs tracing the behavior of raptor.
The actual modifications of the algorithms are  :

- https://github.com/CanalTP/navitia/commit/297e405d0e8635de62990917088e73a4928e2089 adds storage to raptor Labels to remember the walking duration and updates this walking duration when the arrival time is updated for a pair (_stop_point_, _nb_of_transfer_)

- https://github.com/CanalTP/navitia/commit/6fe7efb250fd81e7dcc171fdc422a2c5c9bc5757 
The raptor algorithm works in "rounds". At round _k_ the earliest arrival time at every stop point using at most _k-1_ transfers is already computed, and the round will compute the earliest arrival time at every stop point using _k_ transfers.
In order to do so, raptor scans each _journey_pattern_ :

  - starting from the _first_stop_point_ of the _journey_pattern_, we embark on the first _vehicule_journey_ that we can catch from the earliest arrival time at _first_stop_point_ using _k-1_ transfers.
  - we "ride" the _vehicle_journey_, which gives new arrival times at each _stop_point_ along the ride. When this yields a better arrival time for a _stop_point_ we update its best known arrival time using _k_ transfers
  - it may happen that with our current ride, we arrive at a _stop_point_ later than the arrival time using _k-1_ transfers. In this case, we check if the arrival time using _k-1_ transfers allows to catch an earlier _vehicle_journey_ for this _journey_pattern_. When this happens, we "leave" the current _vehicle_journey_, embark on the earlier one, and remember the current stop_point as the "boarding" stop point.
Before the commit, the boarding stop_point was updated even if the "earlier vehicle journey" was the same as the current one. 
This commit updates the boarding stop_point if : 
    - it allows to catch a strictly earlier _vehicle_journey_
    - or it allows to catch the same _vehicle_journey_ but with a strictly smaller walking duration

- https://github.com/CanalTP/navitia/commit/64de81c2021043ab47396ea972b0bad3e4893aca 
the _fallback_duration_ of a StartingPointSecondPhase was initialized with the walking duration at the arrival stop point before https://github.com/CanalTP/navitia/commit/297e405d0e8635de62990917088e73a4928e2089 . After, _fallback_duration_ is initialized with the total walking duration.
In solution reader, the _fallback_duration_ of a StartingPointSecondPhase was used as the walking duration at the arrival stop point to build bound journeys. Now, bound journeys use the walking duration at arrival stop point computed before raptor (which is stored in a map_stop_point_duration) rather than _fallback_duration

**do_not_merge** for now as it may modify the output of navitia on several functional tests. More modifications may be needed before merge, and tests may need updates.
- tests on auvergne have been analyzed. When the output changed, it seems to yield better journeys than before. Other tests need to be analyzed.
- routing/raptor_test failed : to be analyzed

**JIRA** https://jira.kisio.org/browse/NAVITIAII-2814

### DIFF
--- a/source/routing/benchmark_full.cpp
+++ b/source/routing/benchmark_full.cpp
@@ -263,63 +263,65 @@ int main(int argc, char** argv) {
 
     std::cout << "Launching benchmark " << std::endl;
     boost::progress_display show_progress(requests.size());
-
-    // ProfilerStart("bench.prof");
     int nb_reponses = 0, nb_journeys = 0;
+    {
+        Timer total_compute_timer("Total computing time");
+        // ProfilerStart("bench.prof");
+
 #ifdef __BENCH_WITH_CALGRIND__
-    CALLGRIND_START_INSTRUMENTATION;
+        CALLGRIND_START_INSTRUMENTATION;
 #endif
-    for (auto request : requests) {
-        ++show_progress;
-        Timer t2;
+        for (auto request : requests) {
+            ++show_progress;
+            Timer t2;
 
-        if (verbose) {
-            std::cout << request.start << ", " << request.target << ", " << request.departure_posix_time << "\n";
+            if (verbose) {
+                std::cout << request.start << ", " << request.target << ", " << request.departure_posix_time << "\n";
+            }
+
+            type::EntryPoint origin = make_entry_point(request.start, data);
+            type::EntryPoint destination = make_entry_point(request.target, data);
+
+            origin.streetnetwork_params.mode = request.start_mode;
+            origin.streetnetwork_params.offset = data.geo_ref->offsets[request.start_mode];
+            origin.streetnetwork_params.max_duration = navitia::seconds(30 * 60);
+            origin.streetnetwork_params.speed_factor = 1;
+            destination.streetnetwork_params.mode = request.target_mode;
+            destination.streetnetwork_params.offset = data.geo_ref->offsets[request.target_mode];
+            destination.streetnetwork_params.max_duration = navitia::seconds(30 * 60);
+            destination.streetnetwork_params.speed_factor = 1;
+            type::AccessibiliteParams accessibilite_params;
+            navitia::PbCreator pb_creator(&data, boost::gregorian::not_a_date_time, null_time_period);
+
+            int days_since_epoch = (request.departure_posix_time.date() - boost::gregorian::date(1970, 1, 1)).days();
+            int total_second_in_day = request.departure_posix_time.time_of_day().total_seconds();
+
+            const DateTime departure_datetime = DateTimeUtils::set(days_since_epoch, total_second_in_day);
+            make_response(pb_creator, raptor, origin, destination, {departure_datetime},
+                          true,                      // clockwise ?
+                          accessibilite_params, {},  // forbidden
+                          {},                        // allowed
+                          georef_worker,
+                          type::RTLevel::Base,             // real time level
+                          2_min,                           // transfer penalty
+                          DateTimeUtils::SECONDS_PER_DAY,  // max_duration
+                          10,                              // max_transfers
+                          nb_second_pass);
+            auto resp = pb_creator.get_response();
+
+            Result result(resp.journeys().size(), t2.ms());
+            results.push_back(result);
+
+            if (resp.journeys_size() > 0) {
+                ++nb_reponses;
+                nb_journeys += resp.journeys_size();
+            }
         }
-
-        type::EntryPoint origin = make_entry_point(request.start, data);
-        type::EntryPoint destination = make_entry_point(request.target, data);
-
-        origin.streetnetwork_params.mode = request.start_mode;
-        origin.streetnetwork_params.offset = data.geo_ref->offsets[request.start_mode];
-        origin.streetnetwork_params.max_duration = navitia::seconds(30 * 60);
-        origin.streetnetwork_params.speed_factor = 1;
-        destination.streetnetwork_params.mode = request.target_mode;
-        destination.streetnetwork_params.offset = data.geo_ref->offsets[request.target_mode];
-        destination.streetnetwork_params.max_duration = navitia::seconds(30 * 60);
-        destination.streetnetwork_params.speed_factor = 1;
-        type::AccessibiliteParams accessibilite_params;
-        navitia::PbCreator pb_creator(&data, boost::gregorian::not_a_date_time, null_time_period);
-
-        int days_since_epoch = (request.departure_posix_time.date() - boost::gregorian::date(1970, 1, 1)).days();
-        int total_second_in_day = request.departure_posix_time.time_of_day().total_seconds();
-
-        const DateTime departure_datetime = DateTimeUtils::set(days_since_epoch, total_second_in_day);
-        make_response(pb_creator, raptor, origin, destination, {departure_datetime},
-                      true,                      // clockwise ?
-                      accessibilite_params, {},  // forbidden
-                      {},                        // allowed
-                      georef_worker,
-                      type::RTLevel::Base,             // real time level
-                      2_min,                           // transfer penalty
-                      DateTimeUtils::SECONDS_PER_DAY,  // max_duration
-                      10,                              // max_transfers
-                      nb_second_pass);
-        auto resp = pb_creator.get_response();
-
-        Result result(resp.journeys().size(), t2.ms());
-        results.push_back(result);
-
-        if (resp.journeys_size() > 0) {
-            ++nb_reponses;
-            nb_journeys += resp.journeys_size();
-        }
+        // ProfilerStop();
+#ifdef __BENCH_WITH_CALGRIND__
+        CALLGRIND_STOP_INSTRUMENTATION;
+#endif
     }
-    // ProfilerStop();
-#ifdef __BENCH_WITH_CALGRIND__
-    CALLGRIND_STOP_INSTRUMENTATION;
-#endif
-
     if (vm.count("output")) {
         std::fstream out_file(benchmark_output_file, std::ios::out);
         out_file << "Start point uri, Target uri, Departure posix date time, Computing time (ms), Nb of journeys found"
--- a/source/routing/benchmark_full.cpp
+++ b/source/routing/benchmark_full.cpp
@@ -38,6 +38,8 @@ www.navitia.io
 #include "utils/timer.h"
 
 #include <boost/algorithm/string/predicate.hpp>
+#include "type/pb_converter.h"
+#include "type/meta_data.h"
 #ifdef __BENCH_WITH_CALGRIND__
 #include "valgrind/callgrind.h"
 #endif
@@ -52,28 +54,22 @@ using namespace routing;
 namespace po = boost::program_options;
 namespace ba = boost::algorithm;
 
-struct PathDemand {
+struct Request {
     std::string start;
     std::string target;
 
-    unsigned int date{};
-    unsigned int hour{};
+    pt::ptime departure_posix_time;
 
     type::Mode_e start_mode = type::Mode_e::Walking;
     type::Mode_e target_mode = type::Mode_e::Walking;
 };
 
 struct Result {
-    int duration;
-    int time;
-    int arrival;
-    int nb_changes;
+    int nb_of_journeys_found;
+    int computing_time_in_ms;
 
-    explicit Result(const pbnavitia::Journey& journey)
-        : duration(journey.duration()),
-          time(-1),
-          arrival(journey.arrival_date_time()),
-          nb_changes(journey.nb_transfers()) {}
+    Result(int nb_of_journeys_found_, int computing_time_in_ms_)
+        : nb_of_journeys_found(nb_of_journeys_found_), computing_time_in_ms(computing_time_in_ms_) {}
 };
 
 static type::GeographicalCoord coord_of_entry_point(const type::EntryPoint& entry_point,
@@ -139,34 +135,36 @@ static type::EntryPoint make_entry_point(const std::string& entry_id, const type
 
 int main(int argc, char** argv) {
     navitia::init_app();
-    po::options_description desc("Options de l'outil de benchmark");
-    std::string file, output, stop_input_file, start, target, demands_output_file;
-    int iterations, date, hour, nb_second_pass;
-
-    auto logger = log4cplus::Logger::getInstance(LOG4CPLUS_TEXT("logger"));
-    logger.setLogLevel(log4cplus::WARN_LOG_LEVEL);
+    po::options_description desc("Benchmark tool options");
+    std::string data_file, benchmark_output_file, requests_input_file, requests_output_file;
+    int iterations, nb_second_pass;
 
     // clang-format off
     desc.add_options()
-            ("help", "Show this message")
+            ("help", "Show this message.")
             ("iterations,i", po::value<int>(&iterations)->default_value(100),
-                     "Number of iterations (10 requests by iteration)")
-            ("file,f", po::value<std::string>(&file)->default_value("data.nav.lz4"),
+                     "Number of (start, target) pairs to be generated. Default = 100.\n"
+                     "10 requests will be created for each (start, target) pair, with different departure date_time.")
+            ("file,f", po::value<std::string>(&data_file)->default_value("data.nav.lz4"),
                      "Path to data.nav.lz4")
-            ("start,s", po::value<std::string>(&start),
-                    "Start of a particular journey")
-            ("target,t", po::value<std::string>(&target),
-                    "Target of a particular journey")
-            ("date,d", po::value<int>(&date)->default_value(-1),
-                    "Beginning date of a particular journey")
-            ("hour,h", po::value<int>(&hour)->default_value(-1),
-                    "Beginning hour of a particular journey")
-            ("verbose,v", "Verbose debugging output")
+            ("verbose,v", "Verbose debugging output.")
             ("nb_second_pass", po::value<int>(&nb_second_pass)->default_value(0), "nb second pass")
-            ("stop_files", po::value<std::string>(&stop_input_file), "File with list of start and target")
-            ("dump_demands", po::value<std::string>(&demands_output_file), "Write a csv file with the list of demands (start, target, time) used")
-            ("output,o", po::value<std::string>(&output)->default_value("benchmark.csv"),
-                     "Output file");
+            ("requests", po::value<std::string>(&requests_input_file),
+                        "List of requests to benchmark on.\n"
+                        "Must be a comma-separated csv file where the first 3 columns are :  start point uri, target point uri, departure posix time.\n"
+                        "For example, a line could be :\n"
+                        "stop_area:SAR:SA:1443, stop_area:GT5:SA:4253, 20190717T080000\n"
+                        "The first line of the csv file is assumed to be column headers and is skipped.\n"
+                        "Can contains more than the 3 aforementionned columns.")
+            ("dump_requests", po::value<std::string>(&requests_output_file),
+                                "Write a csv file with the list of requests to be used, before starting the benchmark.\n"
+                                "Each line contains : start uri, target uri, departure posix time."
+                                )
+            ("output,o", po::value<std::string>(&benchmark_output_file),
+                     "Write a csv file with the list of requests used, the computing time (in ms) needed to answer each request, "
+                     "and the number of journey found.\n"
+                     "Each line contains : start uri, target uri, departure posix time, computing time, number of journeys found."
+                     );
     // clang-format on
 
     po::variables_map vm;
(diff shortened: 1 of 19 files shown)


source/routing/raptor.h
```diff
@@ -201,7 +201,9 @@ struct RAPTOR {
                             const nt::RTLevel rt_level,
                             const type::VehicleJourney* vj,
                             const uint16_t l_zone,
-                            DateTime workingDate);
+                            DateTime workingDate,
+                            DateTime working_walking_duration,
```
kinnou02, line 220:
```suggestion
                            DateTime working_fallback_duration,
```

source/routing/raptor_utils.h
```diff
@@ -77,6 +77,9 @@ struct Labels {
     inline void clear(const Labels& clean) {
         dt_pts = clean.dt_pts;
         dt_transfers = clean.dt_transfers;
+
+        walking_duration_pts = clean.walking_duration_pts;
```
kinnou02, line 84: same, walking should always be fallback

AurelienLP:
Jenkins, test this please

source/routing/raptor_solution_reader.h
```diff
@@ -53,7 +53,24 @@ struct Dominates {
     }
 };
 
-typedef ParetoFront<Journey, Dominates> Solutions;
+struct ParetoFrontVisitor {
+    void at_dominated(const Journey& to_insert, const Journey& in_pool) {
+        log4cplus::Logger logger = log4cplus::Logger::getInstance(LOG4CPLUS_TEXT("logger"));
```
kinnou02: You might want to instantiate you logger in the instance, also please name them to help the reading later on

source/routing/benchmark_full.cpp
```diff
@@ -241,6 +242,22 @@ int main(int argc, char** argv) {
         }
     }
 
+    // ecriture des requetes
+    if (vm.count("dump_demands")) {
+        std::fstream out_file(demands_output_file, std::ios::out);
+        out_file << "Start id, Target id, , , Day, Hour" << std::endl;
+
+        for (size_t i = 0; i < demands.size(); ++i) {
```
nonifier: If you want something a little more modern, you can

- Define an output method for `PathDemand`: 

```cpp
std::ostream& operator<<(std::ostream& o, const PathDemand& d) {
    return o << d.start << ", " << d.target << ", , ," << d.date << ", " << d.hour;
}
```
- then use the infamous ostream_iterator :D  
```cpp
boost::copy(demands, std::ostream_iterator<PathDemand>(out_file, "\n"));
```

source/routing/raptor.cpp
```diff
@@ -263,11 +285,22 @@ std::vector<StartingPointSndPhase> make_starting_points_snd_phase(const RAPTOR&
...
 
-            const unsigned walking_t = a.second.total_seconds();
-            StartingPointSndPhase starting_point = {
-                a.first, count,
-                (clockwise ? working_labels.dt_pt(a.first) + walking_t : working_labels.dt_pt(a.first) - walking_t),
-                walking_t, false};
+            const unsigned arrival_walking_t = a.second.total_seconds();
+            const DateTime walking_duration_before_arrival_stop_point = working_labels.walking_duration_pt(a.first);
+            const DateTime total_walking_duration = arrival_walking_t + walking_duration_before_arrival_stop_point;
+            StartingPointSndPhase starting_point = {a.first, count,
```
nonifier: For readability, it would make sense to split the params like:
```cpp
DateTime end_df = working_labels.dt_pt(a.first) + (clockwise ? arrival_walking_t : -arrival_walking_t);
StartingPointSndPhase starting_point = {a.first, count, end_df, total_walking_duration, false};
```

source/routing/raptor.cpp
```diff
@@ -402,9 +434,16 @@ RAPTOR::Journeys RAPTOR::compute_all_journeys(const map_stop_point_duration& dep
                                               bool clockwise,
                                               const boost::optional<navitia::time_duration>& direct_path_dur,
                                               const size_t max_extra_second_pass) {
+    log4cplus::Logger logger = log4cplus::Logger::getInstance(LOG4CPLUS_TEXT("logger"));
```
nonifier: Would be interesting to create a new logger (`"raptor"`? maybe) not to pollute the current one and be able to turn it off if it spits out too many things

source/routing/raptor.cpp
```diff
@@ -444,23 +490,44 @@ RAPTOR::Journeys RAPTOR::compute_all_journeys(const map_stop_point_duration& dep
...
+
         Journey fake_journey =
-            convert_to_bound(start, lower_bound_fb, data.dataRaptor->min_connection_time, transfer_penalty, clockwise);
+            convert_to_bound(start, data.dataRaptor->min_connection_time, transfer_penalty, clockwise);
+
+        LOG4CPLUS_TRACE(logger, " fake journey : " << fake_journey);
+
         if (solutions.contains_better_than(fake_journey)) {
+            LOG4CPLUS_TRACE(logger, "has better solution than fake journey from " << start_stop_point->uri);
+            for (auto solution : solutions) {
```
nonifier:
```suggestion
            for (const auto& solution : solutions) {
```

source/routing/raptor.cpp
```diff
@@ -673,16 +749,34 @@ void RAPTOR::raptor_loop(Visitor visitor, const nt::RTLevel rt_level, uint32_t m
...
+                ///  with the relevant departure date
+                typename Visitor::stop_time_iterator it_st;  /// item = type::StopTime
                 uint16_t l_zone = std::numeric_limits<uint16_t>::max();
+
+                LOG4CPLUS_TRACE(logger, " Scanning line  " << data.pt_data->routes[route_idx.val]->line->uri);
+
                 const auto& jpps_to_explore =
                     visitor.jpps_from_order(data.dataRaptor->jpps_from_jp, jp_idx, q_elt.second);
                 for (const auto& jpp : jpps_to_explore) {
+                    /// jpp : dataRAPTOR::JppsFromJp::Jpp
```
nonifier: If this comment does help you, you can replace `auto` with the actual type. Comment doesn't usually age well :older_man:  ^^

source/routing/raptor.cpp
```diff
@@ -794,5 +951,108 @@ int RAPTOR::best_round(SpIdx sp_idx) {
     return -1;
 }
 
+std::string RAPTOR::print_jpps_from_sp() {
```
nonifier: Are those the following `print_*` functions only for debugging purposes ? 
They don't seem to be used anywhere.

source/routing/raptor_api.cpp
```diff
@@ -1363,6 +1370,18 @@ void make_response(navitia::PbCreator& pb_creator,
         return;
     }
 
+    for (const auto stop_point_iter : *departures) {
```
nonifier: You might not want to iterate through all departures and destinations if you log information above `_TRACE` level.

It would be more interesting to log the whole collections of objects like :
```
  LOG4CPLUS_TRACE(logger, "# departure : " << *departures );
  LOG4CPLUS_TRACE(logger, "# destination : " << *destinations );
```
And you can create the appropriate `ostream& operator<<(...)` function that goes with departures and destinations.

source/routing/raptor_solution_reader.h
nonifier: the logger variable could be static to the class and used across the following 3 functions.

source/routing/raptor_solution_reader.h
```diff
@@ -53,7 +53,24 @@ struct Dominates {
     }
 };
 
-typedef ParetoFront<Journey, Dominates> Solutions;
+struct ParetoFrontVisitor {
```
nonifier: `ParetoFrontLogger` might be a bit more precise ?

nonifier (approved):
I approve but I don't understand the rational behind the changes ^^

source/routing/raptor_solution_reader.h
pbench: Done https://github.com/CanalTP/navitia/pull/2937/commits/53cd072b2a5f6596daa177532ad871f67da15f12

source/routing/raptor_solution_reader.h
pbench: Yep, done https://github.com/CanalTP/navitia/pull/2937/commits/53cd072b2a5f6596daa177532ad871f67da15f12

source/routing/raptor.cpp
pbench: Good idea ! Done https://github.com/CanalTP/navitia/pull/2937/commits/53cd072b2a5f6596daa177532ad871f67da15f12

source/routing/raptor.cpp
pbench: Good catch :) Done https://github.com/CanalTP/navitia/pull/2937/commits/53cd072b2a5f6596daa177532ad871f67da15f12

source/routing/raptor.cpp
pbench: Yes, done ! https://github.com/CanalTP/navitia/pull/2937/commits/53cd072b2a5f6596daa177532ad871f67da15f12

source/routing/raptor.cpp
pbench: Yes that's better, done https://github.com/CanalTP/navitia/pull/2937/commits/53cd072b2a5f6596daa177532ad871f67da15f12

source/routing/raptor_solution_reader.h
pbench: Ok, the logger is now a member of the struct. https://github.com/CanalTP/navitia/pull/2937/commits/53cd072b2a5f6596daa177532ad871f67da15f12
What do you mean by "name them" ?

source/routing/raptor.cpp
```diff
@@ -444,23 +490,44 @@ RAPTOR::Journeys RAPTOR::compute_all_journeys(const map_stop_point_duration& dep
     init_best_pts_snd_pass(calc_dep, departure_datetime, clockwise, best_labels_pts_for_snd_pass);
     auto best_labels_transfers_for_snd_pass = snd_pass_best_labels(clockwise, best_labels_pts);
 
+    LOG4CPLUS_TRACE(logger, "starting points 2nd phase " << std::endl
+                                                         << print_starting_points_snd_phase(starting_points));
+
     unsigned lower_bound_fb = std::numeric_limits<unsigned>::max();
```
xlqian: since it's no longer used, can you remove `lower_bound_fb` ?

source/routing/raptor_utils.h
```diff
@@ -98,6 +112,13 @@ struct Labels {
     IdxMap<type::StopPoint, DateTime> dt_pts;
     // At what time wan we reach this label with a transfer
     IdxMap<type::StopPoint, DateTime> dt_transfers;
+
+    // walking_duration_pts[stop_point] stores the walking duration elapsed during the whole
+    // journey from the starting point which reaches stop point at DateTime dt_pts[stop_point] with public transport
```
xlqian: does that contain also the fallback part?

source/routing/raptor_utils.h
```diff
@@ -98,6 +112,13 @@ struct Labels {
...
+    IdxMap<type::StopPoint, DateTime> walking_duration_pts;
+    // walking_duration_pts[stop_point] stores the walking duration elapsed during the whole
```
xlqian:
```suggestion
    // walking_duration_pts[stop_point] stores the walking transfer duration elapsed during the whole
```
no?

xlqian:
there are a lot of commented logs, can you remove or uncomment them?

source/routing/raptor.cpp
```diff
@@ -705,11 +808,44 @@ void RAPTOR::raptor_loop(Visitor visitor, const nt::RTLevel rt_level, uint32_t m
...
+                    /// the time at which we arrive at stop point jpp.sp_idx (using at most count-1 transfers)
+                    //  hence we can board any vehicle arriving after previous_dt
                     const DateTime previous_dt = prec_labels.dt_transfer(jpp.sp_idx);
-                    if (prec_labels.transfer_is_initialized(jpp.sp_idx) && valid_stop_points[jpp.sp_idx.val]
-                        && (!is_onboard || visitor.better_or_equal(previous_dt, base_dt, *it_st))) {
-                        const auto tmp_st_dt =
-                            next_st->next_stop_time(visitor.stop_event(), jpp.idx, previous_dt, visitor.clockwise());
+                    const DateTime previous_walking_duration = prec_labels.walking_duration_transfer(jpp.sp_idx);
+
+                    /// we are at stop point jpp.idx at time previous_dt
```
xlqian, line 767: can you test `previous_walking_duration <= working_walking_duration` here and `continue` if false? 

it may probably avoid the binary search in `previous_walking_duration <= working_walking_duration`?

source/routing/raptor_api.cpp
```suggestion
    for (const auto& stop_point_iter : *departures) {
```

source/routing/raptor_api.cpp
```diff
@@ -1363,6 +1370,18 @@ void make_response(navitia::PbCreator& pb_creator,
...
+        SpIdx sp_idx = stop_point_iter.first;
+        navitia::type::StopPoint* stop_point = raptor.data.pt_data->stop_points[sp_idx.val];
+        LOG4CPLUS_TRACE(logger, "departure : " << stop_point->uri << " distance : " << stop_point_iter.second);
+    }
+
+    for (const auto stop_point_iter : *destinations) {
```
```suggestion
    for (const auto& stop_point_iter : *destinations) {
```

source/routing/raptor_api.cpp
```diff
@@ -1363,6 +1370,18 @@ void make_response(navitia::PbCreator& pb_creator,
...
+        SpIdx sp_idx = stop_point_iter.first;
+        navitia::type::StopPoint* stop_point = raptor.data.pt_data->stop_points[sp_idx.val];
```
```suggestion
        auto stop_point = raptor.data.pt_data->stop_points[sp_idx.val];
```

source/routing/raptor_solution_reader.h
```diff
@@ -53,7 +53,23 @@ struct Dominates {
     }
 };
 
-typedef ParetoFront<Journey, Dominates> Solutions;
+struct ParetoFrontLogger {
+    log4cplus::Logger logger = log4cplus::Logger::getInstance(LOG4CPLUS_TEXT("raptor"));
+
+    void at_dominated(const Journey& to_insert, const Journey& in_pool) {
+        LOG4CPLUS_TRACE(logger, "Journey dominated ! " << std::endl
```
benoit-bst: Minor : Just for information it's better to use `"\n"` instead of `std::endl` that is more slowly (_endl_ flushes the output buffer)

source/routing/raptor_api.cpp
benoit-bst: FYI : Always try to use `const auto&` :cat: to protect fastness and security

source/routing/raptor.cpp
```diff
@@ -673,16 +748,33 @@ void RAPTOR::raptor_loop(Visitor visitor, const nt::RTLevel rt_level, uint32_t m
...
+                /// the vehicle journey of the current journey_pattern (jp_idx)
+                ///  with the relevant departure date
+                typename Visitor::stop_time_iterator it_st;  /// item = type::StopTime
                 uint16_t l_zone = std::numeric_limits<uint16_t>::max();
+
+                LOG4CPLUS_TRACE(logger, " Scanning line  " << data.pt_data->routes[route_idx.val]->line->uri);
+
                 const auto& jpps_to_explore =
                     visitor.jpps_from_order(data.dataRaptor->jpps_from_jp, jp_idx, q_elt.second);
-                for (const auto& jpp : jpps_to_explore) {
```
benoit-bst, line 686: why ? `auto` works

source/routing/raptor_api.cpp
benoit-bst: Just for good practice :call_me_hand:

source/routing/tests/raptor_test.cpp
```diff
@@ -38,7 +38,13 @@ www.navitia.io
 #include "utils/logger.h"
 
 struct logger_initialized {
-    logger_initialized() { navitia::init_logger(); }
+    logger_initialized() {
+        navitia::init_logger();
+        auto raptor_logger = log4cplus::Logger::getInstance(LOG4CPLUS_TEXT("raptor"));
+        raptor_logger.setLogLevel(log4cplus::FATAL_LOG_LEVEL);
```
pbougue, line 44: Why fatal?

source/routing/benchmark_full.cpp
```diff
@@ -241,6 +242,22 @@ int main(int argc, char** argv) {
         }
     }
 
+    // ecriture des requetes
```
pbougue: :fr: spotted

source/routing/dataraptor.h
```diff
@@ -80,6 +80,8 @@ struct dataRAPTOR {
     };
     JppsFromSp jpps_from_sp;
 
+    std::string print_jpps_from_sp();
```
pbougue: To be removed ?

source/routing/benchmark_full.cpp
```diff
@@ -162,6 +162,7 @@ int main(int argc, char** argv) {
             ("verbose,v", "Verbose debugging output")
             ("nb_second_pass", po::value<int>(&nb_second_pass)->default_value(0), "nb second pass")
             ("stop_files", po::value<std::string>(&stop_input_file), "File with list of start and target")
+            ("dump_demands", po::value<std::string>(&demands_output_file), "Write a csv file with the list of demands (start, target, time) used")
```
pbougue: minor on vocable used in navitia for journeys:
```suggestion
            ("dump_requests", po::value<std::string>(&demands_output_file), "Write a csv file with the list of demands (from, to, datetime) used")
```

source/routing/benchmark_full.cpp
```diff
@@ -241,6 +242,22 @@ int main(int argc, char** argv) {
         }
     }
 
+    // ecriture des requetes
+    if (vm.count("dump_demands")) {
+        std::fstream out_file(demands_output_file, std::ios::out);
+        out_file << "Start id, Target id, , , Day, Hour" << std::endl;
```
pbougue: Please never split date and time, nowhere!
Well that is unless memory or a data format holds a hostage :smiling_imp: 

(also same remark on vocabulary here)

source/routing/raptor_solution_reader.cpp
```diff
@@ -613,9 +621,14 @@ void read_solutions(const RAPTOR& raptor,
 }  // anonymous namespace
 
 std::ostream& operator<<(std::ostream& os, const Journey& j) {
-    os << "([" << navitia::str(j.departure_dt) << ", " << navitia::str(j.arrival_dt) << ", " << j.min_waiting_dur
-       << ", " << j.transfer_dur << "], [" << j.sections.size() << ", " << unsigned(j.nb_vj_extentions) << "], "
-       << j.sn_dur << ") ";
+    os << "(["
+       << "departure_dt : " << navitia::str(j.departure_dt) << ", "
+       << "arrival_dt : " << navitia::str(j.arrival_dt) << ", "
+       << "min_wainting_dur : " << j.min_waiting_dur << ", "
```
pbougue:
```suggestion
       << "min_waiting_dur : " << j.min_waiting_dur << ", "
```

source/routing/raptor_solution_reader.cpp
```diff
@@ -613,9 +621,14 @@ void read_solutions(const RAPTOR& raptor,
...
+       << "transfer_dur : " << j.transfer_dur << "], ["
+       << "nb sections : " << j.sections.size() << ", "
+       << "nb extensions : " << unsigned(j.nb_vj_extentions) << "], "
+       << "foot duration : " << j.sn_dur << ")" << std::endl;
```
pbougue:
```suggestion
       << "fallback duration : " << j.sn_dur << ")" << std::endl;
```

source/routing/raptor_utils.h
```diff
@@ -98,6 +112,13 @@ struct Labels {
     IdxMap<type::StopPoint, DateTime> dt_pts;
     // At what time wan we reach this label with a transfer
```
pbougue: minor
```suggestion
    // At what time can we reach this label with a transfer
```

source/routing/raptor_utils.h
```diff
@@ -98,6 +112,13 @@ struct Labels {
...
     IdxMap<type::StopPoint, DateTime> dt_transfers;
+
+    // walking_duration_pts[stop_point] stores the walking duration elapsed during the whole
+    // journey from the starting point which reaches stop point at DateTime dt_pts[stop_point] with public transport
+    IdxMap<type::StopPoint, DateTime> walking_duration_pts;
+    // walking_duration_pts[stop_point] stores the walking duration elapsed during the whole
+    // journey from the starting point which reaches stop point at DateTime transfers_pts[stop_point] with transfers
```
pbougue:
```suggestion
    // journey from the starting point to 'stop_point' at DateTime transfers_pts[stop_point] with a transfer
```
(same above)

source/routing/raptor_utils.h
pbougue: No, this is the same as what's done on dt_pts and dt_transfers, but for "fallback" duration (fallback + transfers if I'm correct).

See `PT1` and `TR1` in doc: https://github.com/CanalTP/navitia/blob/dev/documentation/rfc/doc-routing-algorithm.md


```suggestion
    // walking_duration_pts[stop_point] stores the fallback+transfer duration elapsed during the whole
```
(same above)

source/routing/raptor_solution_reader.h
```diff
@@ -53,7 +53,23 @@ struct Dominates {
...
+    }
+    void at_dominates(const Journey& /*to_insert*/, const Journey& in_pool) {
+        LOG4CPLUS_TRACE(logger, "Journey removed from solution pool " << std::endl << in_pool);
+    }
+    void at_inserted(const Journey& j) {
+        LOG4CPLUS_TRACE(logger, "Adding  journey to solution pool" << std::endl << j);
+    }
+};
+
+typedef ParetoFront<Journey, Dominates, ParetoFrontLogger> Solutions;
```
pbougue, line 80: Good for debug, but not sure about perf impact to add function calls in that.
(Same for all logs on highly-used code : maybe better to have a compile-flag?)

source/routing/raptor.cpp
```diff
@@ -92,8 +95,15 @@ bool RAPTOR::apply_vj_extension(const Visitor& v,
...
-
+            LOG4CPLUS_TRACE(logger, "Updating label dt count : "
+                                        << count << " sp " << data.pt_data->stop_points[sp_idx.val]->uri << " from "
+                                        << working_labels.dt_pt(sp_idx) << " to " << workingDt << " throught : "
+                                        << st.vehicle_journey->route->line->uri << " boarding_stop_point : "
+                                        << data.pt_data->stop_points[boarding_stop_point.val]->uri
+                                        << " walk : " << working_walking_duration);
             working_labels.mut_dt_pt(sp_idx) = workingDt;
+            working_labels.mut_walking_duration_pt(sp_idx) = working_walking_duration;
+            BOOST_ASSERT(working_walking_duration != DateTimeUtils::not_valid);
```
pbougue: minor: Why not a "simple" `assert` ?
You might want to properly raise a recoverable exception if you keep it, too.

> :heavy_check_mark: OK, we already have some others, that's good like that!

source/routing/raptor.cpp
```diff
@@ -62,7 +62,10 @@ bool RAPTOR::apply_vj_extension(const Visitor& v,
                                 const nt::RTLevel rt_level,
                                 const type::VehicleJourney* vj,
                                 const uint16_t l_zone,
-                                DateTime base_dt) {
+                                DateTime base_dt,
+                                DateTime working_walking_duration,
```
pbougue, line 67: Changing signature just for logs is weird.
Sorry, wrong line, it's :point_down:

source/routing/raptor.cpp
```diff
@@ -285,19 +317,18 @@ std::vector<StartingPointSndPhase> make_starting_points_snd_phase(const RAPTOR&
...
 // this journey aims at being the best we can hope from this StartingPointSndPhase
 Journey convert_to_bound(const StartingPointSndPhase& sp,
-                         uint32_t lower_bound_fb,
                          uint32_t lower_bound_conn,
                          const navitia::time_duration& transfer_penalty,
                          bool clockwise) {
     Journey journey;
     journey.sections.resize(sp.count);  // only the number of sections is part of the dominance function
-    journey.sn_dur = navitia::time_duration(0, 0, sp.fallback_dur + lower_bound_fb, 0);
+    journey.sn_dur = navitia::time_duration(0, 0, sp.fallback_dur, 0);
```
pbougue: Are you sure it's the minimal value ?
With the possible "wrong update" example we discussed, can't we end up with a too high lower bound (stopping the search too early)?

source/routing/raptor.cpp
```diff
@@ -285,19 +317,18 @@ std::vector<StartingPointSndPhase> make_starting_points_snd_phase(const RAPTOR&
...
     uint32_t nb_conn = (sp.count >= 1 ? sp.count - 1 : 0);
     if (clockwise) {
         journey.arrival_dt = sp.end_dt;
-        journey.departure_dt = sp.end_dt - journey.sn_dur.seconds() - nb_conn * lower_bound_conn;
+        journey.departure_dt = sp.end_dt - journey.sn_dur.seconds();
```
pbougue, line 326: :+1: OK, already managed in "new" sn_dur.

source/routing/raptor.cpp
```diff
@@ -673,16 +748,33 @@ void RAPTOR::raptor_loop(Visitor visitor, const nt::RTLevel rt_level, uint32_t m
          * We want to do it, to favoritize normal vj against stay_in vjs
          */
         for (auto q_elt : Q) {
+            /// We will scan the journey_pattern q_elt.first, starting from its stop numbered q_elt.second
+
             const JpIdx jp_idx = q_elt.first;
+
+            const RouteIdx route_idx = data.dataRaptor->jp_container.get(jp_idx).route_idx;
```
pbougue, line 692: Please create vars as close as possible from their use :)

pbougue (commented):
Can you update https://github.com/CanalTP/navitia/blob/dev/documentation/rfc/doc-routing-algorithm.md eventually?

I'd love to have some insight of what you changed in that loop tomorrow, coz it's hard to follow : https://github.com/CanalTP/navitia/pull/2937/files?file-filters%5B%5D=.cpp#diff-30631442826dc5f74f14e880e96c74bcR777 :wink:

source/routing/raptor.cpp
```diff
@@ -705,11 +808,44 @@ void RAPTOR::raptor_loop(Visitor visitor, const nt::RTLevel rt_level, uint32_t m
...
+                    // if we cannot board at this stop point, nothing to do
+                    if (!prec_labels.transfer_is_initialized(jpp.sp_idx) || !valid_stop_points[jpp.sp_idx.val]) {
+                        continue;
+                    }
+
+                    /// the time at which we arrive at stop point jpp.sp_idx (using at most count-1 transfers)
+                    //  hence we can board any vehicle arriving after previous_dt
                     const DateTime previous_dt = prec_labels.dt_transfer(jpp.sp_idx);
-                    if (prec_labels.transfer_is_initialized(jpp.sp_idx) && valid_stop_points[jpp.sp_idx.val]
-                        && (!is_onboard || visitor.better_or_equal(previous_dt, base_dt, *it_st))) {
```
pbougue, line 711: So `better_or_equal` is not used anymore... We should probably remove it if it stays that way.
I don't understand what this tested?

source/routing/raptor.cpp
```diff
@@ -705,11 +808,44 @@ void RAPTOR::raptor_loop(Visitor visitor, const nt::RTLevel rt_level, uint32_t m
                     // We try to get on a vehicle, if we were already on a vehicle, but we arrived
                     // before on the previous via a connection, we try to catch a vehicle leaving this
                     // journey pattern point before
+
+                    // if we cannot board at this stop point, nothing to do
+                    if (!prec_labels.transfer_is_initialized(jpp.sp_idx) || !valid_stop_points[jpp.sp_idx.val]) {
+                        continue;
```
pbougue, line 759: If `is_onboard==true`, shouldn't we do the stay_in check https://github.com/CanalTP/navitia/pull/2937/files#diff-30631442826dc5f74f14e880e96c74bcR892 like before or is it pointless ?

source/routing/raptor.cpp
```diff
@@ -705,11 +808,44 @@ void RAPTOR::raptor_loop(Visitor visitor, const nt::RTLevel rt_level, uint32_t m
...
+                    /// the next vehicle journey will arrive at
+                    ///   tmp_st_dt.second
+                    /// the corresponding StopTime is
+                    ///    tmp_st_dt.first
+                    const auto tmp_st_dt =
+                        next_st->next_stop_time(visitor.stop_event(), jpp.idx, previous_dt, visitor.clockwise());
+
+                    /// if there is no vehicle arriving after previous_dt, nothing to do
+                    if (tmp_st_dt.first == nullptr) {
+                        continue;
```
pbougue, line 778: Same, what about the stay_in check?

source/routing/benchmark_full.cpp
pbench: Corrected in 78fafeb2b78701bf233a101456b4828701b9e8e4

source/routing/benchmark_full.cpp
pbench: Corrected in 78fafeb2b78701bf233a101456b4828701b9e8e4

source/routing/raptor.h
pbench: Done in 3789955136b424900bec1d4cea2b932e3d6bb971

source/routing/raptor.cpp
pbench: Done 051ab1a4d1bf54e95d2a5d195910e797830ad317

source/routing/raptor_utils.h
pbench: For now it only stores the fallback part.
I am testing adding the transfer part, but it has effects on the resulting journeys. 
Not sure yet if it should be just fallback or fallback + transfer.
I added a more precise description comments in 051ab1a4d1bf54e95d2a5d195910e797830ad317

source/routing/raptor_api.cpp
pbench: 051ab1a4d1bf54e95d2a5d195910e797830ad317

source/routing/raptor_api.cpp
pbench: 051ab1a4d1bf54e95d2a5d195910e797830ad317

source/routing/raptor_api.cpp
pbench: Done 051ab1a4d1bf54e95d2a5d195910e797830ad317

source/routing/raptor.cpp
pbench: Mostly to help reading. There is so many different objects used that some explicit typing helps (and also help IDE tooltips).

source/routing/raptor_solution_reader.cpp
pbench: 051ab1a4d1bf54e95d2a5d195910e797830ad317

source/routing/raptor_utils.h
pbench: 051ab1a4d1bf54e95d2a5d195910e797830ad317

source/routing/raptor_solution_reader.h
pbench: Thanks ! I didn't know that

source/routing/tests/raptor_test.cpp
pbench: Just to avoid too much logs on cout during tests

xlqian (approved):

PR #2938, opened 2019-09-05
jormun: Sytral-Rt with direction_type parameter inside request

benoit-bst:
The second part of the Sytral Feature :`Take into account of direction_type with realtime.`
**JIRA** : https://jira.kisio.org/browse/NAVP-1392

The data provides a **direction_type** inside _Route struct_

response_direction_type

Into the protobuf, [direction_type parameter](https://github.com/CanalTP/navitia-proto/blob/master/type.proto#L336) is declared like optional
It can be **forward** or **backward**, if the field exists.
If it exists, the parameter is transmitted to the Go service (_Sytral-Rt_) -> PR https://github.com/CanalTP/sytralrt/pull/18

:information_source: The url code part will rework within an other PR of refactoring :revolving_hearts:

### DIFF
--- a/source/jormungandr/jormungandr/realtime_schedule/sytral.py
+++ b/source/jormungandr/jormungandr/realtime_schedule/sytral.py
@@ -129,6 +129,10 @@ class Sytral(RealtimeProxy):
         for stop_id in stop_id_list[1:]:
             url += "&stop_id={stop_id}".format(stop_id=stop_id)
 
+        direction_type = route_point.fetch_direction_type()
+        if direction_type:
+            url += "&direction_type={direction_type}".format(direction_type=direction_type)
+
         return url
 
     def _get_dt(self, datetime_str):
--- a/source/jormungandr/jormungandr/realtime_schedule/tests/sytral_test.py
+++ b/source/jormungandr/jormungandr/realtime_schedule/tests/sytral_test.py
@@ -46,6 +46,12 @@ def make_url_test():
 
     assert url == 'http://bob.com/?stop_id=stop_tutu'
 
+    url = sytral._make_url(MockRoutePoint(line_code='line_toto', stop_id='stop_tutu', direction_type='forward'))
+
+    assert validators.url(url)
+
+    assert url == 'http://bob.com/?stop_id=stop_tutu&direction_type=forward'
+
 
 def make_url_invalid_code_test():
     """
@@ -318,6 +324,11 @@ class MockRoutePoint(object):
         else:
             self._hardcoded_stop_ids = [l]
 
+        if 'direction_type' in kwargs:
+            self._hardcoded_direction_type = kwargs['direction_type']
+        else:
+            self._hardcoded_direction_type = None
+
     # def fetch_stop_id(self, object_id_tag):
     # return self._hardcoded_stop_id
 
@@ -326,6 +337,9 @@ class MockRoutePoint(object):
 
     def fetch_all_line_id(self, object_id_tag):
         return self._hardcoded_line_ids
+
+    def fetch_direction_type(self):
+        return self._hardcoded_direction_type
 
 
 def next_passage_for_route_point_test(mock_good_response):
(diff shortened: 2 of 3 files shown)


source/jormungandr/jormungandr/realtime_schedule/tests/sytral_test.py
```diff
@@ -350,6 +412,28 @@ def next_passage_for_route_point_test(mock_good_response):
...
+
+    with mock.patch('requests.get', mock_requests.get):
+        passages = sytral.next_passage_for_route_point(route_point)
+
+        assert len(passages) == 1
+
+        assert passages[0].datetime == datetime.datetime(2016, 4, 11, 12, 37, 15, tzinfo=pytz.UTC)
+        assert passages[0].is_real_time
+
+
```
kadhikari: Could you add a test with &direction_type=backward and another test with direction_type=all or both ?

kadhikari, line 436: Could you add a test with &direction_type=backward and another test with direction_type=all or both ?

benoit-bst: It is easy to add others direction_type tests, but I think that is not relevant. Why ?
Because in the test, no matter the value of the parameter. 
We write in the mock the same value as the URL. 
There is not intelligence behind.

benoit-bst:
Jenkins, test this please !

kinnou02 (approved):

source/jormungandr/jormungandr/realtime_schedule/sytral.py
```diff
@@ -129,6 +129,10 @@ def _make_url(self, route_point):
         for stop_id in stop_id_list[1:]:
             url += "&stop_id={stop_id}".format(stop_id=stop_id)
 
+        direction_type = route_point.fetch_direction_type()
+        if direction_type:
+            url += "&direction_type={direction_type}".format(direction_type=direction_type)
```
pbougue, line 134: Just to confirm: this is not our business to check how data is filled for `direction_type` in NTFS ?
We will just get a 400 from subservice and that's it if `direction_type=wtf`.

source/jormungandr/jormungandr/realtime_schedule/tests/sytral_test.py
```diff
@@ -175,6 +215,7 @@ def mock_missing_line_response():
             {
                 "direction_id": "3341",
                 "direction_name": "Piscine Chambéry",
+                "direction_type": "both",
```
pbougue, line 218: 1. We don't use this field anywhere, no?
2. Should we use it to attach it to the right route when reading response from subservice? (in case direction_id=backward in some of the response even if we requested "forward")
3. Also, if we keep it in tests, is `both` a realistic value?

source/jormungandr/jormungandr/realtime_schedule/tests/sytral_test.py
```diff
@@ -350,6 +412,28 @@ def next_passage_for_route_point_test(mock_good_response):
...
 
+def next_passage_for_route_point_with_direction_type_test(mock_direction_type_is_forward_response):
+    """
+    test the whole next_passage_for_route_point with direction_type parameter
+    mock the http call to return a good response, we should get some next_passages
+    """
+    sytral = Sytral(id='tata', service_url='http://bob.com/')
+
+    mock_requests = MockRequests(
+        {'http://bob.com/?stop_id=42&direction_type=forward': (mock_direction_type_is_forward_response, 200)}
```
pbougue, line 423: So we don't use line_code but we use direction_type... This is quite weird to ask for all "forward" routes passing at a stop, no?
> ugly,  but to be managed later (maybe) as it works as-is.

source/jormungandr/jormungandr/realtime_schedule/tests/sytral_test.py
benoit-bst: I added this field in all mock struct but it is not use, yes. 
It just to be conform with the new Sytral-Rt response version. It is mandatory now

source/jormungandr/jormungandr/realtime_schedule/sytral.py
benoit-bst: Jormun acts like a hatch. It just transmits the value contained into data. Currently, this is in the GO PR that is protect with weird values.

kadhikari (approved):

pbougue (approved):

PR #2939, opened 2019-09-05
[Jormun] Add direct path mode

AurelienLP:
Add the `direct_path_mode[]` parameter to Jormun.
Read the documentation first to understand what it does so that this PR is clearer.
This parameter is used both for new_default and distributed.

`get_kraken_calls()` now returns a tuple `(mode1, mode2, direct_path_type)` so that we know if we have to compute a direct_path or a path with public transportation for each couple of mode.

If we have `(mode1, mode2, "only")` and `(mode1, mode2, "indifferent")`, we remove `(mode1, mode2, "only")` because `(mode1, mode2, "indifferent")` will compute a direct_path anyway. We avoid a useless call to kraken this way.

We then filter the journeys if we don't want a direct_path for certain mode or if we don't want a pt journey with a certain mode (and just keep the direct_path).

The idea is not to "force the return of a direct whatever its length or duration" but "to compute the direct path of a mode which is not in first_section_mode[]" and only keep this path".


source/jormungandr/jormungandr/interfaces/v1/journey_common.py
```diff
@@ -424,6 +424,16 @@ def __init__(self, output_type_serializer):
...
         )
+        parser_get.add_argument(
+            "direct_path_mode[]",
+            type=OptionValue(fallback_modes.all_fallback_modes),
+            default=[],
+            dest="direct_path_mode",
+            action="append",
+            help="Force the modes we want a direct path."
+            "If this list is not empty, we only compute direct_path for modes in this list"
+            "And filter all the direct_paths of modes in first_section mode[]",
```
xlqian:
```suggestion
            "And filter all the direct_paths of modes in first_section_mode[]",
```

source/jormungandr/jormungandr/scenarios/journey_filter.py
```diff
@@ -332,6 +338,44 @@ def filter_func(self, journey):
...
+        self.dest_modes = [] if dest_modes is None else dest_modes
+
+    def filter_func(self, journey):
+        """
+        eliminates journeys that are not matching direct path mode parameter
+        """
+
+        is_dp = 'non_pt' in journey.tags
+        is_in_direct_path_mode_list = any(mode in self.dp_mode for mode in journey.tags)
+        is_in_fallback_mode_list = any(mode in self.orig_modes for mode in journey.tags) or any(
```
xlqian: i don't understand why do you care about this case?

source/jormungandr/jormungandr/scenarios/new_default.py
```diff
@@ -102,25 +102,40 @@ def get_kraken_calls(request):
...
-    if not res:
+    if not combinaison_mode:
         abort(
             404,
             message='the asked first_section_mode[] ({}) and last_section_mode[] '
             '({}) combination is not yet supported'.format(dep_modes, arr_modes),
         )
 
+    for first_mode, second_mode in combinaison_mode:
+        res.add((first_mode, second_mode, dp_type))
```
xlqian: how do you handle the case where the request is `first_section_mode[]=car&last_section_mode[]=car&direct_path_mode[]=car`?

documentation/slate/source/includes/apis.md
```diff
@@ -1022,6 +1022,7 @@ See how disruptions affect a journey in the [real time](#realtime) section.
...
 | nop     | wheelchair           | boolean | If true the traveler is considered to be using a wheelchair, thus only accessible public transport are used<br>be warned: many data are currently too faint to provide acceptable answers with this parameter on       | False       |
 | nop     | direct_path          | enum    | Specify if Navitia should suggest direct paths (= only fallback modes are used).<br>Possible values: <ul><li>`indifferent`</li><li>`none` for only journeys using some PT</li><li>`only` for only journeys without PT</li></ul>      | indifferent |
+| nop     | direct_path_mode[]	     | array of strings     | Force the modes we want a direct path. If this list is not empty, we only compute direct_path for modes in this list and filter all the direct_paths of modes in first_section_mode[]. It takes the following values: `walking`, `car`, `bike`, `bss`, `ridesharing`, `taxi`. It's an array, you can give multiple modes.  | []           |
```
pbougue:
```suggestion
| nop     | direct_path_mode[]	     | array of strings     | Force direct-path modes. If this list is not empty, we only compute direct_path for modes in this list no matter `max_duration_to_pt` and filter all the direct_paths of modes in first_section_mode[]. It can take the following values: `walking`, `car`, `bike`, `bss`, `ridesharing`, `taxi`. It's an array, you can give multiple modes. If this list is empty, we will compute direct_path for modes of the first_section_modes.  | first_section_modes[]           |
```

source/jormungandr/jormungandr/interfaces/v1/journey_common.py
```diff
@@ -424,6 +424,16 @@ def __init__(self, output_type_serializer):
             help="Radius length (in meters) around the coordinates of arrival "
             "in which the stop points are considered free to go (crowfly=0)",
         )
+        parser_get.add_argument(
+            "direct_path_mode[]",
+            type=OptionValue(fallback_modes.all_fallback_modes),
+            default=[],
+            dest="direct_path_mode",
+            action="append",
+            help="Force the modes we want a direct path."
```
pbougue:
```suggestion
            help="Force the direct-path modes."
```

source/jormungandr/jormungandr/interfaces/v1/journey_common.py
```diff
@@ -424,6 +424,16 @@ def __init__(self, output_type_serializer):
...
+            "If this list is not empty, we only compute direct_path for modes in this list"
```
pbougue, line 434: Depending on what you change in slate, you should report part of the changes here :wink:

pbougue (commented):
Just reviewed the "interface" and expected behavior part, looks good!

> AurelienLP wrote:
> If we have `(mode1, mode2, "only")` and `(mode1, mode2, "indifferent")`, we remove `(mode1, mode2, "only")` because `(mode1, mode2, "indifferent")` will compute a direct_path anyway. We avoid a useless call to kraken this way.

With max_duration_to_pt=30min and a direct_path length of 2h, will it work?
From what you say I expect no direct_path in result, but I think the point is also to force its processing.
So I think you can only save the `(mode1, mode2, "only")` if you had the result from indifferent, which won't always be the case.

source/jormungandr/jormungandr/scenarios/new_default.py
AurelienLP: It will returns (car, car, "indifferent") because of line 117

source/jormungandr/jormungandr/scenarios/journey_filter.py
AurelienLP: first_section=car&diret_path_mode=bike --> You have to filter direct_path of car and pt journeys with bike.

AurelienLP:
As seen with @pbougue, the direct paths which were filtered before will also be filtered with this implementation. The idea is not to "force the return of a direct whatever its length or duration" but "to compute the direct path of a mode which is not in first_section_mode[]" and only keep this path".

source/jormungandr/jormungandr/scenarios/helper_classes/pt_journey.py
```diff
@@ -214,7 +214,7 @@ def _create_parameters(request):
     def _async_request(self):
         direct_path_type = StreetNetworkPathType.DIRECT
         periode_extremity = utils.PeriodExtremity(self._request['datetime'], self._request['clockwise'])
-        for dep_mode, arr_mode in self._krakens_call:
+        for dep_mode, arr_mode, dp_type in self._krakens_call:
```
kinnou02: we don't use `dp_type` here, why using it?

documentation/slate/source/includes/apis.md
AurelienLP: As seen with you, I'll take this without "no matter `max_duration_to_pt`". Thanks again!

source/jormungandr/jormungandr/scenarios/journey_filter.py
```diff
@@ -332,6 +338,44 @@ def filter_func(self, journey):
...
+            mode in self.dest_modes for mode in journey.tags
```
AurelienLP: This line is wrong by the way. I have to remove it.

source/jormungandr/jormungandr/scenarios/new_default.py
```diff
@@ -170,7 +185,12 @@ def create_pb_request(requested_type, request, dep_mode, arr_mode):
     sn_params.origin_mode = dep_mode
     sn_params.destination_mode = arr_mode
 
-    req.journeys.max_duration = request["max_duration"]
+    # If we only want direct_paths, max_duration(time to pass in pt) is zero
+    if dp_type == "only":
+        req.journeys.max_duration = 0
```
kinnou02, line 191: doing a `DirectPathRequest` would probably be more efficient, and I don't think it will change the response

source/jormungandr/jormungandr/scenarios/helper_classes/pt_journey.py
AurelienLP: `for dep_mode, arr_mode, _ in self._krakens_call` then

source/jormungandr/jormungandr/scenarios/new_default.py
AurelienLP: Tried it but the direct paths are never returned. :(

source/jormungandr/jormungandr/scenarios/journey_filter.py
pbougue: Side-note: It would be better to work that way everywhere (this is another PR anyway).
We discussed about it with @TeXitoi: we should use first & last sections modes for direct_path modes by default.
This would make more sense and it would need to be the case on isochrones and on journeys.

source/jormungandr/jormungandr/scenarios/journey_filter.py
```diff
@@ -332,6 +338,44 @@ def filter_func(self, journey):
...
+        )
+
+        if self.dp == 'none' and is_dp:
```
pbougue: Shouldn't it be managed by another filter, called in any case and mutualized?
(same below)

And then it's the integrator's problem if he asks for direct_path_modes[]=taxi&direct_path=none (resulting in no direct_path).

source/jormungandr/jormungandr/scenarios/journey_filter.py
```diff
@@ -332,6 +338,44 @@ def filter_func(self, journey):
...
+
+        if self.dp == 'only' and not is_dp:
+            return False
+
+        # if direct_path of a mode we want the direct_path
+        if is_dp and is_in_direct_path_mode_list:
+            return True
+
+        # if fallback of a mode we want to fallback
+        if not is_dp and is_in_fallback_mode_list:
```
pbougue: Looks like it's not the responsibility of this filter, no?

source/jormungandr/jormungandr/scenarios/journey_filter.py
```diff
@@ -332,6 +338,44 @@ def filter_func(self, journey):
...
+            return True
+
+        return False
```
pbougue: So default is we remove the journey? Looks harsh.

Not sure I understand what is the responsibility of this method (maybe add a comment if I'm not the only one).
I was expecting this filter to just remove direct path that were not requested from `direct_path_mode` and `first_section_mode`, without taking any care of the rest.

source/jormungandr/jormungandr/scenarios/tests/new_default_tests.py
```diff
@@ -343,48 +343,62 @@ def add_pt_sections(journey):
 def get_kraken_calls_test():
     for md in ["bike", "walking", "bss", "car", "ridesharing", "taxi"]:
         req = {"origin_mode": [md], "destination_mode": [md]}
-        assert get_kraken_calls(req) == {(md, md)}
+        assert get_kraken_calls(req) == {(md, md, "indifferent")}
 
-    req = {"origin_mode": ["bss", "walking"], "destination_mode": ["walking"]}
-    assert get_kraken_calls(req) == {("walking", "walking")}
+    req = {"origin_mode": ["bss", "walking"], "destination_mode": ["walking"], "direct_path": "none"}
+    assert get_kraken_calls(req) == {("walking", "walking", "none")}
```
pbougue, line 349: minor: It's not your change, but I think we should have `{("bss", "walking", "none")}`, or maybe `{("bss", "walking", "none"), ("walking", "walking", "none")}`, no?

source/jormungandr/jormungandr/scenarios/tests/new_default_tests.py
```diff
@@ -343,48 +343,62 @@ def add_pt_sections(journey):
...
 
-    req = {"origin_mode": ["bike", "walking"], "destination_mode": ["walking"]}
-    assert get_kraken_calls(req) == {("walking", "walking"), ("bike", "walking")}
+    req = {"origin_mode": ["bike", "walking"], "destination_mode": ["walking"], "direct_path": "only"}
+    assert get_kraken_calls(req) == {("walking", "walking", "only"), ("bike", "walking", "only")}
```
pbougue: Not sure, it would be clearer to have ` {("walking", "walking", "only"), ("bike", "bike", "only")}` in that case.
It can bloat the code, though.

source/jormungandr/jormungandr/scenarios/tests/new_default_tests.py
```diff
@@ -343,48 +343,62 @@ def add_pt_sections(journey):
...
 
-    req = {"origin_mode": ["bike", "walking"], "destination_mode": ["bss"]}
-    assert get_kraken_calls(req) == {("bike", "bss")}
+    req = {"origin_mode": ["bike", "walking"], "destination_mode": ["bss"], "direct_path": "indifferent"}
+    assert get_kraken_calls(req) == {("bike", "bss", "indifferent")}
```
pbougue, line 355: minor: I really don't get why the walking is not there originally.

source/jormungandr/jormungandr/scenarios/tests/new_default_tests.py
```diff
@@ -343,48 +343,62 @@ def add_pt_sections(journey):
...
-        ('bike', 'taxi'),
-        ('taxi', 'walking'),
-        ('taxi', 'ridesharing'),
+        ('taxi', 'taxi', "indifferent"),
+        ('taxi', 'bss', "indifferent"),
+        ('bike', 'bss', "indifferent"),
+        ('bike', 'walking', "indifferent"),
+        ('bike', 'taxi', "indifferent"),
+        ('taxi', 'walking', "indifferent"),
+        ('taxi', 'ridesharing', "indifferent"),
```
pbougue, line 371: minor: Same, I think (bike, bss) and (bike, ridesharing) are missing originally here...

source/jormungandr/jormungandr/scenarios/tests/new_default_tests.py
```diff
@@ -343,48 +343,62 @@ def add_pt_sections(journey):
...
+    req = {"origin_mode": ["bss", "walking"], "destination_mode": ["walking"], "direct_path_mode": ["bike"]}
+    assert get_kraken_calls(req) == {("walking", "walking", "indifferent"), ("bike", "bike", "only")}
+
+    req = {
+        "origin_mode": ["bss", "walking"],
+        "destination_mode": ["walking"],
+        "direct_path_mode": ["bike"],
+        "direct_path": "none",
+    }
+    assert get_kraken_calls(req) == {("walking", "walking", "none"), ("bike", "bike", "only")}
```
pbougue: Here we could save a call and remove the (bike, bike, only) as it would be filtered by the `pt=none` filter.

source/jormungandr/jormungandr/scenarios/tests/journey_compare_tests.py
```diff
@@ -1151,3 +1151,50 @@ def test_activate_deactivate_min_taxi():
...
+    journey.tags.append("non_pt")
+    f = jf.FilterDirectPathMode(dp="indifferent", dp_mode=["car"], orig_modes=[], dest_modes=[])
+    assert f.filter_func(journey)
+
+    # is_dp and not is_in_direct_path_mode_list
+    journey = response_pb2.Journey()
+    journey.tags.append("car")
+    journey.tags.append("non_pt")
+    f = jf.FilterDirectPathMode(dp="indifferent", dp_mode=["bike"], orig_modes=["car"], dest_modes=[])
+    assert not f.filter_func(journey)
```
pbougue: To me it's not the responsibility of the filter you created.
The journey should just never be produced, not being filtered by this filter.

source/jormungandr/jormungandr/scenarios/tests/journey_compare_tests.py
```diff
@@ -1151,3 +1151,50 @@ def test_activate_deactivate_min_taxi():
...
+    # not is_dp and is_in_fallback_mode_list
+    journey = response_pb2.Journey()
+    journey.tags.append("car")
+    f = jf.FilterDirectPathMode(dp="indifferent", dp_mode=[], orig_modes=["car"], dest_modes=[])
+    assert f.filter_func(journey)
+
+    # not is_dp and not is_in_fallback_mode_list
+    journey = response_pb2.Journey()
+    journey.tags.append("car")
+    f = jf.FilterDirectPathMode(dp="indifferent", dp_mode=[], orig_modes=[], dest_modes=["car"])
```
pbougue: Same, not the responsibility of the filter to me.

source/jormungandr/jormungandr/scenarios/tests/journey_compare_tests.py
```diff
@@ -1151,3 +1151,50 @@ def test_activate_deactivate_min_taxi():
...
+    assert not f.filter_func(journey)
+
+    # not is_dp and not is_in_fallback_mode_list
+    journey = response_pb2.Journey()
+    journey.tags.append("car")
+    f = jf.FilterDirectPathMode(dp="indifferent", dp_mode=[], orig_modes=[], dest_modes=[])
```
pbougue: Same probably.

pbougue (commented):
Remarks related to (first, last) authorized pairs are probably minor as it was the case before ;)

source/jormungandr/jormungandr/scenarios/tests/new_default_tests.py
AurelienLP: According fallback_modes.py :
"we remove bss walking and  walking bss, since they are redundant for kraken (bss includes walking)"

source/jormungandr/jormungandr/scenarios/tests/journey_compare_tests.py
```diff
@@ -1151,3 +1151,44 @@ def test_activate_deactivate_min_taxi():
...
     f = jf.FilterTooShortHeavyJourneys(min_taxi=8, orig_modes=['taxi'], dest_modes=['taxi', 'walking'])
     assert not f.filter_func(journey)
+
+
+def test_filter_direct_path_mode_car():
+    # is_dp and not is_in_direct_path_mode_list
+    journey = response_pb2.Journey()
+    journey.tags.append("car")
+    journey.tags.append("non_pt")
+    f = jf.FilterDirectPathMode([])
```
pbougue: dp_mode is never gonna be empty as you test it before using it.
It's probably better to change this to the nominal case:
```suggestion
    f = jf.FilterDirectPathMode(["bike"])
```

source/jormungandr/jormungandr/scenarios/new_default.py
```diff
@@ -102,25 +102,41 @@ def get_kraken_calls(request):
...
 
     if len(dep_modes) == len(arr_modes) == 1:
-        return {(dep_modes[0], arr_modes[0])}
+        res.add((dep_modes[0], arr_modes[0], dp_type))
+        return res
 
     # this allowed_combinations is temporary, it does not handle all the use cases at all
     # We don't want to do ridesharing - ridesharing journeys
-    res = {c for c in fallback_modes.allowed_combinations if c in itertools.product(dep_modes, arr_modes)}
+    combinaison_mode = {
```
pbougue: :fr: spotted
```suggestion
    combination_mode = {
```

source/jormungandr/jormungandr/scenarios/new_default.py
```diff
@@ -102,25 +102,41 @@ def get_kraken_calls(request):
...
 
-    if not res:
+    if not combinaison_mode:
         abort(
             404,
             message='the asked first_section_mode[] ({}) and last_section_mode[] '
             '({}) combination is not yet supported'.format(dep_modes, arr_modes),
         )
 
+    for first_mode, second_mode in combinaison_mode:
```
pbougue: minor naming more accurate
```suggestion
    for dep_mode, arr_mode in combination_mode:
```

pbougue (approved):
It may be nice to launch artemis on it, prior to merge.

kinnou02 (approved):

benoit-bst (approved):
Minor: I prefer long named var like `dp_type` -> `direct_path_type`

benoit-bst:
:heart:

PR #2941, opened 2019-09-09
jormungandr: Use params for sytral-rt request

benoit-bst:
Reshape _Sytral-RT_ **GET** with _pybreaker params_.
Now, **URL** is not written with handmade concatenation, but incorporated inside a _GET params struct_.

- UTests are updated
- Tested locally with a **Sytral-Rt** instance and it works. URL transits with params :tada:

### DIFF
--- a/source/jormungandr/jormungandr/realtime_schedule/sytral.py
+++ b/source/jormungandr/jormungandr/realtime_schedule/sytral.py
@@ -81,8 +81,27 @@ class Sytral(RealtimeProxy):
         except:
             return self.rt_system_id
 
+    def _make_params(self, route_point):
+        '''
+        create params list for GET request
+        '''
+        stop_id_list = route_point.fetch_all_stop_id(self.object_id_tag)
+        if not stop_id_list:
+            logging.getLogger(__name__).debug(
+                'missing realtime id for {obj}: stop code={s}'.format(obj=route_point, s=stop_id_list),
+                extra={'rt_system_id': unicode(self.rt_system_id)},
+            )
+            self.record_internal_failure('missing id')
+            return None
+        params = [('stop_id', i) for i in stop_id_list]
+
+        direction_type = route_point.fetch_direction_type()
+        if direction_type:
+            params.append(("direction_type", direction_type))
+        return params
+
     @cache.memoize(app.config['CACHE_CONFIGURATION'].get('TIMEOUT_SYTRAL', 30))
-    def _call(self, url):
+    def _call(self, url, **kwargs):
         """
         http call to sytralRT
         """
@@ -91,7 +110,7 @@ class Sytral(RealtimeProxy):
             extra={'rt_system_id': unicode(self.rt_system_id)},
         )
         try:
-            return self.breaker.call(requests.get, url, timeout=self.timeout)
+            return self.breaker.call(requests.get, url, **kwargs)
         except pybreaker.CircuitBreakerError as e:
             logging.getLogger(__name__).error(
                 'systralRT service dead, using base ' 'schedule (error: {}'.format(e),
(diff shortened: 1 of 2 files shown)


source/jormungandr/jormungandr/realtime_schedule/tests/sytral_test.py
```diff
@@ -397,8 +402,8 @@ def next_passage_for_route_point_test(mock_good_response):
     """
     sytral = Sytral(id='tata', service_url='http://bob.com/')
 
-    mock_requests = MockRequests({'http://bob.com/?stop_id=42': (mock_good_response, 200)})
-
+    # request http://bob.com/?stop_id=42
```
benoit-bst: I updated the mocked response with the simple url, because when we use params struct, the *mock.patch* only see simple url and not the params. 
I added a comment for more clarity

source/jormungandr/jormungandr/realtime_schedule/sytral.py
```diff
@@ -91,7 +110,7 @@ def _call(self, url):
             extra={'rt_system_id': unicode(self.rt_system_id)},
         )
         try:
-            return self.breaker.call(requests.get, url, timeout=self.timeout)
```
pbougue, line 94: Shouldn't you keep the timeout (maybe in parameter build) ?

source/jormungandr/jormungandr/realtime_schedule/tests/sytral_test.py
pbougue: Maybe you can use this: https://github.com/CanalTP/navitia/blob/dev/source/jormungandr/jormungandr/tests/utils_test.py#L64-L87
Or maybe there are even better tools out there in the wild :lion:

source/jormungandr/jormungandr/realtime_schedule/tests/sytral_test.py
```diff
@@ -36,34 +36,42 @@
...
     sytral = Sytral(id='tata', service_url='http://bob.com/')
 
-    url = sytral._make_url(MockRoutePoint(line_code='line_toto', stop_id='stop_tutu'))
+    # The return is like [(stop_id, val1), (stop_id, val2), (direction_type, val), ...]
+    params = sytral._make_params(MockRoutePoint(line_code='line_toto', stop_id='stop_tutu'))
 
-    # it should be a valid url
-    assert validators.url(url)
+    assert len(params) == 1
+    assert 'stop_tutu' in params[0]
```
pbougue: Is it possible to test it like this? (to test the stop's id is in "stop_id" param)
```suggestion
    assert params == [("stop_id", "stop_tutu")]
```
or
```suggestion
    assert params[0] == ("stop_id", "stop_tutu")
```
Same below if relevant.

benoit-bst: I prefer your first proposition. 
Thanks for the trick :ok_hand:

source/jormungandr/jormungandr/realtime_schedule/sytral.py
benoit-bst: I added both with the same value. Into param struct or directly within the pybreaker param.
I don't know what is better. Any idea ?

benoit-bst: It seems to be more comfortable to separate data parameters and pure technical parameters like time out. 
It is done for others realtime connector.

source/jormungandr/jormungandr/realtime_schedule/sytral.py
```diff
@@ -81,8 +81,27 @@ def __repr__(self):
...
+        params = [('stop_id', i) for i in stop_id_list]
+
+        direction_type = route_point.fetch_direction_type()
+        if direction_type:
+            params.append(("direction_type", direction_type))
+        return params
+
     @cache.memoize(app.config['CACHE_CONFIGURATION'].get('TIMEOUT_SYTRAL', 30))
-    def _call(self, url):
+    def _call(self, url, **kwargs):
```
kinnou02: I prefer explicit argument when possible

source/jormungandr/jormungandr/realtime_schedule/tests/sytral_test.py
benoit-bst: Hoooo nice. this is what i need :ok_hand: 
Thanks

benoit-bst:
I addressed all your comments and recheck that it works always locally. 
Thanks for the help :smile_cat:

source/jormungandr/jormungandr/realtime_schedule/tests/sytral_test.py
```diff
@@ -430,7 +413,7 @@ def next_passage_for_route_point_with_direction_type_test(mock_direction_type_is
     sytral = Sytral(id='tata', service_url='http://bob.com/')
 
     mock_requests = MockRequests(
-        {'http://bob.com/?stop_id=42&direction_type=forward': (mock_direction_type_is_forward_response, 200)}
+        {'http://bob.com/?direction_type=forward&stop_id=42': (mock_direction_type_is_forward_response, 200)}
```
benoit-bst, line 416: because of the sort

pbougue (approved):

source/jormungandr/jormungandr/realtime_schedule/sytral.py
```diff
@@ -81,17 +81,36 @@ def __repr__(self):
...
         http call to sytralRT
         """
         logging.getLogger(__name__).debug(
-            'systralRT RT service , call url : {}'.format(url),
+            'systralRT RT service , call url : {}'.format(self.service_url),
             extra={'rt_system_id': unicode(self.rt_system_id)},
         )
         try:
-            return self.breaker.call(requests.get, url, timeout=self.timeout)
+            return self.breaker.call(requests.get, url=self.service_url, params=params, timeout=self.timeout)
```
xlqian, line 113: :+1:

xlqian (approved):

nonifier (approved):

kinnou02 (approved):

PR #2943, opened 2019-09-10
Jormun: disable thread in procfile

kinnou02:
Since flask 1.0 the development server is multithreaded,
mixing thread and greenlet doesn't work well, so this PR disable
threading in jormun


pbougue (approved):
I guess it's gonna be the same for Kirin when we migrate it.

benoit-bst (approved):
if you say so...

nonifier (approved):

PR #2944, opened 2019-09-10
kraken: disable case insensitive test in fare

kinnou02:
they are slow, very very slow...
During import all fare attributes are already converted to lower case, so only the sectionKey need to be converted to lowercase as is contains ntfs ids
TODO:
  - [x] check that fr-idf still works
  - [x] handle case-insensitive test at data integration if possible

### DIFF
--- a/source/fare/fare.cpp
+++ b/source/fare/fare.cpp
@@ -34,6 +34,7 @@ www.navitia.io
 #include <boost/algorithm/string.hpp>
 #include <boost/lexical_cast.hpp>
 #include <boost/optional.hpp>
+#include <boost/algorithm/string/case_conv.hpp>
 
 #include "type/datetime.h"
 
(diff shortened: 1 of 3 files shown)


nonifier:
What about lowering the case of those fields when creating the `State`, `SectionKey` and `Label` objects ?

nonifier (approved):
What about lowering the case of those fields when creating the `State`, `SectionKey` and `Label` objects ? We don't need the case sensitivity are runtime don't we ?

AurelienLP (approved):

pbougue (approved):
:+1: with initial lowering here https://github.com/CanalTP/navitia/blob/dev/source/ed/connectors/fare_utils.cpp#L63

benoit-bst (approved):

PR #2945, opened 2019-09-10
ed : grid_calendars files follow NTFS norm with backward compatibility

benoit-bst:
**Thematic** : Update ED reader to follow the [NTFS norm](https://github.com/CanalTP/ntfs-specification/blob/master/ntfs_fr.md#grid_calendarstxt-optionnel) for grid_calendars* files. Backward compatibility is mandatory for this change 

**JIRA** : https://jira.kisio.org/browse/NAVITIAII-2855

Only updated current tests and data set.
The attached test files inside JIRA ticket works well with fusio2ed binary.
Docker tests are green :green_apple:

### DIFF
--- a/source/ed/connectors/fusio_parser.cpp
+++ b/source/ed/connectors/fusio_parser.cpp
@@ -1330,7 +1330,10 @@ void PeriodFusioHandler::init(Data&) {
     if (id_c == UNKNOWN_COLUMN) {
         id_c = csv.get_pos_col("calendar_id");
     }
-    begin_c = csv.get_pos_col("begin_date");
+    begin_c = csv.get_pos_col("start_date");
+    if (begin_c == UNKNOWN_COLUMN) {
+        begin_c = csv.get_pos_col("begin_date");
+    }
     end_c = csv.get_pos_col("end_date");
 }
 
--- a/source/ed/connectors/fusio_parser.cpp
+++ b/source/ed/connectors/fusio_parser.cpp
@@ -1330,17 +1330,16 @@ void PeriodFusioHandler::init(Data&) {
     if (id_c == UNKNOWN_COLUMN) {
         id_c = csv.get_pos_col("calendar_id");
     }
-    begin_c = csv.get_pos_col("start_date");
-    if (begin_c == UNKNOWN_COLUMN) {
-        begin_c = csv.get_pos_col("begin_date");
+    start_c = csv.get_pos_col("start_date");
+    if (start_c == UNKNOWN_COLUMN) {
+        start_c = csv.get_pos_col("begin_date");
     }
     end_c = csv.get_pos_col("end_date");
 }
 
 void PeriodFusioHandler::handle_line(Data&, const csv_row& row, bool is_first_line) {
     if (!is_first_line && !has_col(id_c, row)) {
-        LOG4CPLUS_FATAL(logger, "Error while reading " << csv.filename
-                                                       << "  file has more than one period and no calendar_id column");
+        LOG4CPLUS_FATAL(logger, "Error while reading " << csv.filename << " column id : " << id_c << " not available");
         throw InvalidHeaders(csv.filename);
     }
     auto cal = gtfs_data.calendars_map.find(row[id_c]);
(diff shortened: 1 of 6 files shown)


source/ed/connectors/fusio_parser.cpp
```diff
@@ -1326,20 +1326,23 @@ static boost::gregorian::date parse_date(const std::string& str) {
 namespace grid_calendar {
 
 void PeriodFusioHandler::init(Data&) {
-    calendar_c = csv.get_pos_col("calendar_id");
+    id_c = csv.get_pos_col("grid_calendar_id");
+    if (id_c == UNKNOWN_COLUMN) {
+        id_c = csv.get_pos_col("calendar_id");
+    }
```
pbougue, line 1332: It could be nice to add a test to reject file after this if `id_c` is still `UNKNOWN_COLUMN`.
Same on all files/columns.
(this would replace the "required" columns system).

source/ed/connectors/fusio_parser.cpp
```diff
@@ -1326,20 +1326,26 @@ static boost::gregorian::date parse_date(const std::string& str) {
 namespace grid_calendar {
 
 void PeriodFusioHandler::init(Data&) {
-    calendar_c = csv.get_pos_col("calendar_id");
-    begin_c = csv.get_pos_col("begin_date");
+    id_c = csv.get_pos_col("grid_calendar_id");
+    if (id_c == UNKNOWN_COLUMN) {
+        id_c = csv.get_pos_col("calendar_id");
+    }
+    begin_c = csv.get_pos_col("start_date");
```
pbougue: very minor: rename it ?
```suggestion
    start_c = csv.get_pos_col("start_date");
```

pbougue (approved):
So you just hand-tested backward-compat' ? Too boring for the gain !

source/ed/connectors/fusio_parser.h
```diff
@@ -252,10 +252,10 @@ struct TripPropertiesFusioHandler : public GenericHandler {
 namespace grid_calendar {
 struct PeriodFusioHandler : public GenericHandler {
     PeriodFusioHandler(GtfsData& gdata, CsvReader& reader) : GenericHandler(gdata, reader) {}
-    int calendar_c, begin_c, end_c;
+    int id_c, begin_c, end_c;
     void init(Data&);
     void handle_line(Data& data, const csv_row& line, bool is_first_line);
-    const std::vector<std::string> required_headers() const { return {"calendar_id", "begin_date", "end_date"}; }
+    const std::vector<std::string> required_headers() const { return {"end_date"}; }
```
nonifier, line 258:
```suggestion
    const std::vector<std::string> required_headers() const { return {"grid_calendar_id", "start_date", "end_date"}; }
```

nonifier: Why removing the header in all the objects ? They should be required according to the specs.

nonifier (commented):
Having tests on the backward compatibility would have been nice.

source/ed/connectors/fusio_parser.h
benoit-bst: Because of the backward compatibility., we have to support new and old names. 
We can't freeze the header for those params

source/ed/connectors/fusio_parser.cpp
benoit-bst: Yes, the new name is **start_date**. Just to be conform with the new name. Why not

benoit-bst:
To be clear, It is possible to create compatibility tests, but it requires new NTFS data set and it is not so easy to create it. 
So, to test this PR, I updated the fixtures with new rules and passed old NFTS by hands

source/ed/connectors/fusio_parser.cpp
benoit-bst: if there is no **id_c** (not _grid_calendar_id_ and not _calendar_id_) the binary will break with this message :  
```
if (!is_first_line && !has_col(id_c, row)) {                                                                                                                                                                                                                           
        LOG4CPLUS_FATAL(                                                                                                                                                                                                                                                         
            logger, "Error while reading " + csv.filename + "  file has more than one grid_id and no id column");                                                                                                                                                            
    }
```

So, it breaks but not with the good message. I will update it for all parsers with 

```
"Error while reading " csv.filename << " , column id : " << id_c << " not available;
```

kinnou02 (approved):

kadhikari (approved):
:+1:

nonifier (approved):

PR #2946, opened 2019-09-11
PTRef: Increase weight of route<->SP link to avoid unwanted shortcuts

pbougue:
JIRA: https://jira.kisio.org/browse/NAVITIAII-2793

fix after #2931

### DIFF
--- a/source/type/tests/comments_test.cpp
+++ b/source/type/tests/comments_test.cpp
@@ -50,7 +50,7 @@ BOOST_GLOBAL_FIXTURE(logger_initialized);
 
 namespace navitia {
 namespace type {
-bool operator==(const std::vector<Comment>& lhs, const std::vector<std::string>& rhs) {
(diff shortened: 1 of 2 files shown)


kinnou02 (approved):

nonifier (approved):

source/type/tests/comments_test.cpp
```diff
@@ -50,7 +50,7 @@ BOOST_GLOBAL_FIXTURE(logger_initialized);
 
 namespace navitia {
 namespace type {
-bool operator==(const std::vector<Comment>& lhs, const std::vector<std::string>& rhs) {
+static bool operator==(const std::vector<Comment>& lhs, const std::vector<std::string>& rhs) {
```
benoit-bst, line 53: Why this change ?

benoit-bst (approved):

source/type/tests/comments_test.cpp
pbougue: Just remove a clang warning, as it's missing prototype.

PR #2950, opened 2019-09-13
[Jormun] Fix direct_path_mode with distributed

AurelienLP:
We don't compute pt_journey for modes in `direct_path_mode[]`

### DIFF
--- a/source/jormungandr/jormungandr/scenarios/distributed.py
+++ b/source/jormungandr/jormungandr/scenarios/distributed.py
@@ -97,6 +97,9 @@ class Distributed(object):
             mode for _, mode, direct_path_type in krakens_call if direct_path_type != "only"
         }
 
+        # These are the modes in first_section_modes[] and direct_path_modes[]
+        # We need to compute direct_paths for them either because we requested it with direct_path_modes[]
+        # Or because we need them to optimize the pt_journey computation
         requested_direct_path_modes = {
             mode for mode, _, direct_path_type in krakens_call if direct_path_type == "only"
         }
--- a/source/jormungandr/jormungandr/scenarios/distributed.py
+++ b/source/jormungandr/jormungandr/scenarios/distributed.py
@@ -90,8 +90,17 @@ class Distributed(object):
         logger = logging.getLogger(__name__)
         logger.debug('request datetime: %s', request['datetime'])
 
-        requested_dep_modes = {mode for mode, _, _ in krakens_call}
-        requested_arr_modes = {mode for _, mode, _ in krakens_call}
+        requested_dep_modes_with_pt = {
+            mode for mode, _, direct_path_type in krakens_call if direct_path_type != "only"
+        }
+        requested_arr_modes_with_pt = {
+            mode for _, mode, direct_path_type in krakens_call if direct_path_type != "only"
+        }
+
+        requested_direct_path_modes = {
+            mode for mode, _, direct_path_type in krakens_call if direct_path_type == "only"
+        }
+        requested_direct_path_modes.update(requested_dep_modes_with_pt)
 
         if context.partial_response_is_empty:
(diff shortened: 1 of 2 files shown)


source/jormungandr/jormungandr/scenarios/distributed.py
```diff
@@ -90,8 +90,17 @@ def _compute_all(self, future_manager, request, instance, krakens_call, context)
...
+            mode for mode, _, direct_path_type in krakens_call if direct_path_type != "only"
+        }
+        requested_arr_modes_with_pt = {
+            mode for _, mode, direct_path_type in krakens_call if direct_path_type != "only"
+        }
+
+        requested_direct_path_modes = {
+            mode for mode, _, direct_path_type in krakens_call if direct_path_type == "only"
+        }
+        requested_direct_path_modes.update(requested_dep_modes_with_pt)
```
xlqian, line 106: a comment on this line will be helpful for further dev

source/jormungandr/jormungandr/scenarios/helper_classes/pt_journey.py
```diff
@@ -214,35 +214,36 @@ def _create_parameters(request):
...
-                arr_mode=arr_mode,
-                periode_extremity=periode_extremity,
-                journey_params=self._journey_params,
-                bike_in_pt=bike_in_pt,
-                request=self._request,
-            )
-
-            self._value.append(PtPoolElement(dep_mode, arr_mode, pt_journey))
+        for dep_mode, arr_mode, dp_type in self._krakens_call:
+            if dp_type != "only":
```
xlqian:
```python
if dp_type == "only":
    continue

dp = self._streetnetwork_path_pool.wait_and_get(...
```
you have less indentation to do in this way :)

xlqian (approved):

source/jormungandr/jormungandr/scenarios/distributed.py
pbougue, line 106: Not sure at all : Aren't you supposed to do that only when requested_direct_path_modes is empty?

source/jormungandr/jormungandr/scenarios/helper_classes/pt_journey.py
pbougue: Won't you block some kraken calls in new_default?
Not sure this is used there, though.

pbougue (approved):
Looks OK even if I'm a bit lost in the rationale for that code (trust @xlqian 's review for that).
So the goal is to stop computing PT journeys with modes that are only intended to be direct_paths.
(a bit ambiguous from main comment).

Maybe it's worth launching artemis on this PR.

source/jormungandr/jormungandr/scenarios/distributed.py
AurelienLP: > Not sure at all : Aren't you supposed to do that only when requested_direct_path_modes is empty?

We need to compute direct_paths for modes in first_sections_modes AND modes in direct_path_modes

AurelienLP: We need to compute direct_paths for modes in first_sections_modes AND modes in direct_path_modes

source/jormungandr/jormungandr/scenarios/helper_classes/pt_journey.py
AurelienLP: Nice catch !

AurelienLP: This part is distributed only

xlqian (approved):

xlqian:
References of artemis may need to be updated.

pbougue (approved):
Thanks for clear comments :+1:

AurelienLP:
Jenkins test this please

AurelienLP:
Jenkins, test this please

PR #2951, opened 2019-09-13
[Tyr] New endpoints in API v1

mbounabi:
https://jira.kisio.org/browse/NAVP-1369

### DIFF
--- a/source/tyr/api_v1.md
+++ b/source/tyr/api_v1.md
@@ -46,7 +46,7 @@ If lots of elements expected in the response:
 ```
 
 - POST /v1/<endpoints>:  
-Used to create a new element of the endpoint. Returned status code should **201**.
+Used to create a new element of the endpoint. Returned status code should be **201**.
 ```
 {
     "endpoint": [
@@ -58,7 +58,7 @@ Used to create a new element of the endpoint. Returned status code should **201*
 ```
 
 - PUT /v1/<endpoints>:  
-Used to update a new element of the endpoint. Returned status code should **200**.  
+Used to update a new element of the endpoint. Returned status code should be **200**.  
 Note: If the POST method isn't available for an endpoint, PUT can be used to create a new element. In this case, the returned status code should be **201**
 ```
 {
--- a/source/tyr/api_v1.md
+++ b/source/tyr/api_v1.md
@@ -45,7 +45,8 @@ If lots of elements expected in the response:
 }
 ```
 
-- POST /v1/<endpoints>:
+- POST /v1/<endpoints>:  
+Used to create a new element of the endpoint. Returned status code should **201**.
 ```
 {
     "endpoint": [
@@ -56,7 +57,9 @@ If lots of elements expected in the response:
 }
 ```
 
-- PUT /v1/<endpoints>:
+- PUT /v1/<endpoints>:  
+Used to update a new element of the endpoint. Returned status code should **200**.  
+Note: If the POST method isn't available for an endpoint, PUT can be used to create a new element. In this case, the returned status code should be **200**
 ```
 {
     "endpoint": [
--- a/source/tyr/api_v1.md
+++ b/source/tyr/api_v1.md
@@ -59,7 +59,7 @@ Used to create a new element of the endpoint. Returned status code should **201*
 
 - PUT /v1/<endpoints>:  
 Used to update a new element of the endpoint. Returned status code should **200**.  
-Note: If the POST method isn't available for an endpoint, PUT can be used to create a new element. In this case, the returned status code should be **200**
+Note: If the POST method isn't available for an endpoint, PUT can be used to create a new element. In this case, the returned status code should be **201**
 ```
 {
     "endpoint": [
--- a/source/tyr/tests/integration/equipment_providers_test.py
+++ b/source/tyr/tests/integration/equipment_providers_test.py
@@ -116,13 +116,9 @@ def test_equipments_provider_put(default_equipments_config):
 
     # Update existing provider
     new_provider['args']['url'] = 'sytral3.url.update'
-    resp, status = api_put(
-        'v0/equipments_providers/sytral3',
-        data=ujson.dumps(new_provider),
-        content_type='application/json',
-        check=False,
+    resp = api_put(
+        'v0/equipments_providers/sytral3', data=ujson.dumps(new_provider), content_type='application/json'
     )
-    assert status == 200
     assert 'equipments_provider' in resp
     assert len(resp['equipments_provider']) == 1
     for key in new_provider.keys():
--- a/source/tyr/tests/integration/equipment_providers_test.py
+++ b/source/tyr/tests/integration/equipment_providers_test.py
@@ -105,10 +105,10 @@ def test_equipments_provider_put(default_equipments_config):
         check=False,
     )
     assert status == 201
-    assert 'equipments_providers' in resp
-    assert len(resp['equipments_providers']) == 1
+    assert 'equipments_provider' in resp
+    assert len(resp['equipments_provider']) == 1
     for key in new_provider.keys():
-        assert resp['equipments_providers'][0][key] == new_provider[key]
+        assert resp['equipments_provider'][0][key] == new_provider[key]
 
     resp = api_get('/v0/equipments_providers')
     assert 'equipments_providers' in resp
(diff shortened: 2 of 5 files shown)


source/tyr/tests/integration/v1_test.py
```diff
@@ -81,6 +81,18 @@ def create_api():
...
+@pytest.fixture
+def create_autocomplete_parameters():
+    with app.app_context():
+        autocomp = models.AutocompleteParameter(name='autocomp_name')
+        models.db.session.add(autocomp)
+        models.db.session.commit()
+
+        yield autocomp.name
+
+        models.db.session.delete(autocomp)
```
xlqian, line 93: i don't know if you would need a commit() after the delete?

source/tyr/tests/integration/v1_test.py
```diff
@@ -197,3 +209,151 @@ def test_authorization_methods(create_5_users, create_instance, create_api):
         '/v1/users/{}/authorizations'.format(user_id), data=json.dumps(auth), content_type='application/json'
     )
     assert len(resp_delete['user']['authorizations']) == 0
+
+
+def test_autocomplete_parameters(create_autocomplete_parameters):
+    check_v1_response('autocomplete_parameters')
+    check_v1_response(
+        'autocomplete_parameters', 'autocomplete_parameters/{}'.format(create_autocomplete_parameters)
+    )
```
xlqian, line 218: i think what you want is this:
```python
with create_autocomplete_parameters() as autocomp_name:
    check_v1_response('autocomplete_parameters', 'autocomplete_parameters/{}'.format(autocomp_name)))
```
IMHO, it's weird to use fixture like this :)

source/tyr/tyr/resources.py
```diff
@@ -1190,13 +1190,15 @@ def post(self, user_id, version=0):
...
 
 class EndPoint(flask_restful.Resource):
-    @marshal_with(end_point_fields)
-    def get(self):
-        return models.EndPoint.query.all()
+    def get(self, version=0):
+        resp = marshal(models.EndPoint.query.all(), end_point_fields)
+        if version == 1:
+            return {'end_points': resp}
+        return resp
```
xlqian, line 1197: (i know it may be overkill for this PR, it depends on if you want to organize v0 and v1 better)

It seems that we're mixing functional codes and serialization code according to the version of api. 

in order to de-correlate these logics, can we, for every endpoint,  do something like,
```python
return serializer_select(version).serialize(endpoint_impl_func(...))
```

xlqian (approved):

source/tyr/tests/integration/v1_test.py
xlqian: my bad..

source/tyr/tests/integration/v1_test.py
mbounabi: It doesn't seem necessary:
```models.db.session.delete(autocomp)        
print ("Parameters after delete : {}".format(json.dumps(api_get('v1/autocomplete_parameters'))))
```

-> `Parameters after delete : {"autocomplete_parameters":[]}`

source/tyr/tyr/resources.py
mbounabi: I created a ticket to monitor the upgrade of the API v1, which, I fully agree, isn't optimized as is:
https://jira.kisio.org/browse/NAVP-1409

kinnou02 (approved):

source/tyr/api_v1.md
```diff
@@ -45,7 +45,8 @@ If lots of elements expected in the response:
 }
 ```
 
-- POST /v1/<endpoints>:
+- POST /v1/<endpoints>:  
+Used to create a new element of the endpoint. Returned status code should **201**.
```
pbougue: minor
```suggestion
Used to create a new element of the endpoint. Returned status code should be **201**.
```

source/tyr/api_v1.md
```diff
@@ -56,7 +57,9 @@ If lots of elements expected in the response:
 }
 ```
 
-- PUT /v1/<endpoints>:
+- PUT /v1/<endpoints>:  
+Used to update a new element of the endpoint. Returned status code should **200**.
```
pbougue: minor
```suggestion
Used to update a new element of the endpoint. Returned status code should be **200**.  
```

source/tyr/api_v1.md
```diff
@@ -56,7 +57,9 @@ If lots of elements expected in the response:
 }
 ```
 
-- PUT /v1/<endpoints>:
+- PUT /v1/<endpoints>:  
+Used to update a new element of the endpoint. Returned status code should **200**.  
+Note: If the POST method isn't available for an endpoint, PUT can be used to create a new element. In this case, the returned status code should be **201**
```
pbougue, line 62: :cold_sweat:
:+1: Nice of you to document it, though!

nonifier (approved):

source/tyr/tyr/api.py
```diff
@@ -63,20 +63,31 @@
 
 api.add_resource(resources.Job, '/v0/jobs/', '/v0/jobs/<string:instance_name>/', endpoint=str('jobs'))
 
-api.add_resource(resources.EndPoint, '/v0/end_points/', '/v0/end_points/<int:id>/', endpoint=str('end_points'))
+api.add_resource(
+    resources.EndPoint,
+    '/v<int:version>/end_points/',
```
pbougue, line 68: :white_check_mark: Ohh looks like these are user's management endpoints :astonished:
:old_key: :spider_web: :card_file_box:

source/tyr/tests/integration/equipment_providers_test.py
```diff
@@ -105,20 +105,28 @@ def test_equipments_provider_put(default_equipments_config):
...
-        'v0/equipments_providers/sytral3', data=ujson.dumps(new_provider), content_type='application/json'
+    resp, status = api_put(
+        'v0/equipments_providers/sytral3',
+        data=ujson.dumps(new_provider),
+        content_type='application/json',
+        check=False,
     )
-    assert 'message' in resp
-    assert resp['message'] == 'Provider sytral3 from db is updated'
+    assert status == 200
```
pbougue: minor: you could remove that and let `check=True` above.

source/tyr/tests/integration/v1_test.py
```diff
@@ -197,3 +209,151 @@ def test_authorization_methods(create_5_users, create_instance, create_api):
...
+        'admin': 'COSMOGONY',
+        'poi': 'OSM',
+    }
+    resp_post, status_post = api_post(
+        '/v1/autocomplete_parameters',
+        data=json.dumps(autocomplete_data),
+        content_type='application/json',
+        check=False,
+    )
+    assert status_post == 201
```
pbougue, line 240: minor: you may remove that and change the check in `api_post` to test `201` and not 200 (we may be surprised in other tests though :smile: )
Same on delete for 204.

source/tyr/tests/integration/v1_test.py
```diff
@@ -197,3 +209,151 @@ def test_authorization_methods(create_5_users, create_instance, create_api):
...
+def test_end_points():
+    check_v1_response('end_points')
+
+
+def test_end_points_methods():
+    resp_get = api_get('/v1/end_points')
+    assert 'end_points' in resp_get
+    initial_num = len(resp_get['end_points'])
+
+    endpoint_data = {'name': 'tyr_v1', 'hostnames': ['host_v1']}
```
pbougue, line 330: minor advertisement possible here :smile: (also helps understand what it does?) :
```suggestion
    endpoint_data = {'name': 'navitia.io', 'hostnames': ['host_v1']}
```

pbougue (approved):
Only minor ones, let you decide on them.
Let you resolve the others :astonished: ones!

source/tyr/api_v1.md
mbounabi: Should we create a method POST for every endpoint? It can be discussed, in order to have an even better homogeneity for this v1.
However, we can note this (create a ticket?) for a future refinement and not consider it blocking for this PR.

pbougue: Yep, no remark is blocking.
I'd like more homogeneity for this, but this is just for internal use, so it's not urgent/mandatory.
OK to track it in a ticket if you agree.

source/tyr/tests/integration/v1_test.py
mbounabi: Changing it will surely break some other tests.
Also, the status checked in "api_delete" is 200. Meaning that some other v0 endpoints return 200 when deletion is OK?

source/tyr/tests/integration/v1_test.py
mbounabi: Not possible as the endpoint 'navitia.io' already exists...

source/tyr/api_v1.md
mbounabi: I created a ticket to track what is left for the full v1 experience:
https://jira.kisio.org/browse/NAVP-1425

PR #2900, opened 2019-07-26
Jormun: call realtime proxy in parallel

kinnou02:
In stop_schedules call to realtime proxies was done in serial, this PR
make theses call in parallel with gevent.
This will reduce the response time of `/stop_schedules`

PR #2953, opened 2019-09-17
Jormun: restore newrelic instrumentation on schedules and departures

kinnou02:
function run another greenlet aren't instrumented, so with #2900 we lost
metrology on external service used by realtime proxies.
Newrelic doesn't provide any way to reatach a greenlet to an already
started Transaction :(
This PR create a new Background Transaction for each greenlet to keep
track of response time of external services


kinnou02:
Jenkins, test this please.

pbougue (approved):

benoit-bst (approved):

PR #2955, opened 2019-09-18
jormun & kraken: Add direction type parameter for filtering departures API result.

benoit-bst:
**Feature** : Allow to filter `/departures` results with _direction type_ parameter (**All**, **Forward**, **Backward**)
**JIRA** : https://jira.kisio.org/browse/NAVP-1087

**QA part**:
Tested on fr-se-lyon, several request with different data type (**forward** and **backward**) and it seems to be ok for all params :ok_hand:

### DIFF
--- a/source/ed/build_helper.h
+++ b/source/ed/build_helper.h
@@ -131,7 +131,7 @@ struct VJ {
         return *this;
     }
 
-    VJ& route(const std::string& r, const std::string& d = "forward") {
+    VJ& route(const std::string& r, const std::string& d = "") {
         _route_name = r;
         _direction_type = d;
         return *this;
--- a/source/ed/build_helper.h
+++ b/source/ed/build_helper.h
@@ -131,11 +131,8 @@ struct VJ {
         return *this;
     }
 
-    VJ& route(const std::string& r) {
+    VJ& route(const std::string& r, const std::string& d = "forward") {
         _route_name = r;
-        return *this;
-    }
-    VJ& direction_type(const std::string& d) {
         _direction_type = d;
         return *this;
     }
--- a/source/ed/build_helper.h
+++ b/source/ed/build_helper.h
@@ -68,6 +68,7 @@ struct VJ {
     const std::string line_name;
     std::string _block_id;
     std::string _route_name;
+    std::string _direction_type = "";
     const bool is_frequency;
     const bool wheelchair_boarding;
     std::string _name;
@@ -132,6 +133,10 @@ struct VJ {
 
     VJ& route(const std::string& r) {
         _route_name = r;
+        return *this;
+    }
+    VJ& direction_type(const std::string& d) {
+        _direction_type = d;
         return *this;
     }
     VJ& bike_accepted(bool b) {
--- a/source/time_tables/passages.h
+++ b/source/time_tables/passages.h
@@ -30,6 +30,7 @@ www.navitia.io
 
 #pragma once
 #include "type/pb_converter.h"
+#include "type/type_utils.h"
 namespace navitia {
 namespace timetables {
 
@@ -44,6 +45,7 @@ void passages(PbCreator& pb_creator,
               const type::RTLevel rt_level,
               const pbnavitia::API api_pb,
               const uint32_t count,
-              const uint32_t start_page);
+              const uint32_t start_page,
+              const boost::optional<type::DirectionType>& direction_type = boost::none);
 }
 }  // namespace navitia
--- a/source/time_tables/passages.h
+++ b/source/time_tables/passages.h
@@ -46,6 +46,6 @@ void passages(PbCreator& pb_creator,
               const pbnavitia::API api_pb,
               const uint32_t count,
               const uint32_t start_page,
-              const boost::optional<type::DirectionType>& direction_type = boost::none);
+              const boost::optional<pbnavitia::DirectionType>& direction_type = boost::none);
 }
 }  // namespace navitia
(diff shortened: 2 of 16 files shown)


source/type/type_utils.cpp
```diff
@@ -82,4 +82,33 @@ pbnavitia::RTLevel to_pb_realtime_level(const navitia::type::RTLevel realtime_le
             throw navitia::exception("realtime level case not handled");
     }
 }
+
+namespace type {
+
+type::DirectionType get_direction_type(pbnavitia::DirectionType direction_type) {
```
nonifier: Why not returning a `boost::optional<type::DirectionType>` instead ? That would simplify the code around this.
Same applies for the function below.

source/type/type_utils.h
```diff
@@ -49,4 +50,13 @@ const type::StopTime& earliest_stop_time(const std::vector<type::StopTime>& sts)
 
 pbnavitia::RTLevel to_pb_realtime_level(const navitia::type::RTLevel realtime_level);
 
+namespace type {
+
+enum class DirectionType { All = 0, Forward, Backward };
```
nonifier: Would this enum better be in `type/type.h` ?

nonifier (approved):
Nice one !

source/type/type_utils.h
benoit-bst: Because there is no more data struct into type.h, since @kinnou02's changes :smile: 
But I hesitated to create a `direction_type.h`. What do you think ? One header for all **direction_type** stuff :crown:

source/type/type_utils.cpp
benoit-bst: Because i want to directly manipulate the enum :taco:

benoit-bst: And, I tested that the **trow** is well catched

source/type/type_utils.h
pbougue: If we keep the proto enum... Why not use it directly and remove that one?

source/ed/build_helper.h
```diff
@@ -134,6 +135,10 @@ struct VJ {
         _route_name = r;
         return *this;
     }
+    VJ& direction_type(const std::string& d) {
```
pbougue: This must be associated to route's build (above), not to VJ.
As it won't trigger a route build for "backward" if there is already one for "forward" (which is what one would hope using it).

pbougue (changes requested):
* This does not handle inbound/outbound and clockwise/anticlockwise (to be added to tests/doc/help everywhere needed).
* As route.direction_type is a string from data we must be as resilient as possible, so do a case-insensitive check.
* Please also test that a route.direction_type='toto' or ='' (which is possible in data) appears with a 'all' filter.

----

* I don't understand the rationale of not doing it as mentioned in the ticket (just re-read it): the PTRef solution doesn't require a proto change, it looks cleaner as it doesn't change interfaces, and it's usable in more places. Maybe I miss something, though.

source/ed/build_helper.h
benoit-bst: You have right, it is more clear to link the **direction_type** to the _route_ with a function like this
```
ed::builder b("20190101");
 b.vj("L1").route("route_name", "direction_type");
```

benoit-bst:
After @pbougue's remarks, i took into account:
- Into the data, the type of direction supported is : _forward, clockwise, inbound or backward, anticlockwise, outbound_. Add some tests
- Now, if the direction type, into the data, is not correct, we take all departures (no filtering). It is more safe to avoid crash
- **Lower case** is the only supported now from data field
- The builder was updated to add new param for route method like `builder.vj("1").route(route_name, direction_name);`
- Fix others tests with side effect that due to builder changes

source/type/type_utils.cpp
```diff
@@ -82,4 +82,22 @@ pbnavitia::RTLevel to_pb_realtime_level(const navitia::type::RTLevel realtime_le
...
+
+pbnavitia::DirectionType get_direction_type(std::string& direction_type) {
+    if (direction_type == "all") {
+        return pbnavitia::DirectionType::ALL;
+    } else if ((direction_type == "forward") || (direction_type == "clockwise") || (direction_type == "inbound")) {
+        return pbnavitia::DirectionType::FORWARD;
+    } else if ((direction_type == "backward") || (direction_type == "anticlockwise")
+               || (direction_type == "outbound")) {
+        return pbnavitia::DirectionType::BACKWARD;
+    } else {
```
kinnou02, line 96: maybe for an empty value, but if it ask `foo` as a direction type I expect an error

source/time_tables/passages.h
```diff
@@ -30,6 +30,7 @@ www.navitia.io
 
 #pragma once
 #include "type/pb_converter.h"
+#include "type/type_utils.h"
```
kinnou02, line 33: A forward declare isn't enough?

kinnou02 (changes requested):
I agree with @pbougue we explicitly choose to do it we a ptref filter but you didn't without providing an explanation.
Also a filter will reduce the number of JPP to explore, so the number of stop_time we got and probably the response time.
Since the number of stop_time is passed to `get_stop_time()` I expect it to only return the desired value, if you filter them after we won't have the number of stop_time that we wanted to have.

source/type/type_utils.cpp
benoit-bst: I have seen with @pbougue that for strange case like `direction_type=""` or `direction_type="poo"`, we don't want to filter and we don't want to crash the request.
But If everybody is not agree I'll change it again :tada:

benoit-bst:
It was the wrong way to find the light :goat:. 
Bye bye PR

An other PR will be open with a new way, the PT Ref way

PR #2956, opened 2019-09-18
Add some testes due to performance improvements patch

GuillaumeLescure:
WARNING : testes doesn't work, it still need work on it.
I'm unable to find why the filter is invalid.

Commit should be split :
1) to add testes
2) add logs


pbougue:
Just finished-replaced it from a non-central branch : #2957

PR #2959, opened 2019-09-19
Handle disruption with multiple stop_points in same area / repeated stop_points

eturck:
Hello,

I tried to separate my commits to be as clear as possible, it might be easier to look at the two or three tests I've added to understand the problem.

We have a few cases when disruptions on line_section were not exactly working as intended.
The first case was on a vj where the last two stop_points are in the same area (can be generalized to two consecutive stop_points in the same_area at the start / end of a section) :

Screenshot_20190919-162853

If we had a line_section disruption between a previous area X and this area Z, only the "drop off stop" was impacted by the disruption. We were still able to make a journey between a stop before X and Z.

This is because we take the shortest section, so we reset the start of the section while we see it and close the section as soon as we found a stop_time in the end area.
To solve this I just don't reset the start_idx of the section if I am in the same_area, and keep iterating over the stop_times before inserting the stops of the section until I'm outside of the end area.

While I was fixing this I found another particular case. If we have a vj `A -> B-> C-> D-> E-> B-> F` for example. If we were to disrupt the section `B -> D` the method will return a set `(B,C,D)`. Problem is, the adapted VJ will be `A/E/F`. Since we only take the shortest section it should be `A/E/B/F`, no reason to impact the second occurrence of `B`.
To solve this I changed the way we get the stop_points impacted by a section. Currently we were getting a set of stop_points to impact and iterating over every stop_time to see if its stop_point was in the set. Now we get the ranks of the stop_times impacted on the base vj.

It's not totally perfect. Since we don't have a hard link between a non-base stop_time and its base stop_time when we use [get_base_stop_time](https://github.com/Tisseo/navitia/blob/handle_repeated_stop_points_disruption/source/type/stop_time.cpp#L62), if it's not a base VJ we are impacting, we need to loop on all stop_times, and it could be wrong in a few cases.
I highlighted this limitation in a unit test, since `get_base_stop_time` is not perfect, we could be impacting some stop_times which are not supposed to be, but it's some really ~~dumb~~ specific cases (non consecutive stop_times with same stop_points and same departure / arrival times).

Let me know what you think and if there is improvements to be made!

### DIFF
--- a/source/kraken/tests/apply_disruption_test.cpp
+++ b/source/kraken/tests/apply_disruption_test.cpp
@@ -1438,7 +1438,7 @@ BOOST_AUTO_TEST_CASE(add_impact_on_line_section_cancelling_vj) {
 
 /*
  * In case we have a pretty complex line with repeated stops, since a line_section is only defined by
- * a line, two stop areas and an optional list of routes, we're impacting the longest section possible for each
+ * a line, two stop areas and an optional list of routes, we're impacting the shortest sections possible for each
  * route. Let's take a look at the line A below, with 4 routes, all departing from s1 and ending at s6.
  *
  * 1st route s1/s2/s3/s2/s3/s4/s5/s4/s5/s6 :
@@ -1563,7 +1563,8 @@ BOOST_AUTO_TEST_CASE(add_line_section_impact_on_line_with_repeated_stops) {
     base_vp = vj->base_validity_pattern()->days;
     BOOST_CHECK_MESSAGE(ba::ends_with(adapted_vp.to_string(), "111111"), adapted_vp);
     BOOST_CHECK_MESSAGE(ba::ends_with(base_vp.to_string(), "000000"), base_vp);
-    BOOST_REQUIRE_EQUAL(vj->stop_time_list.size(), 2);
+    // 6 stop_times since we don't impact the first iteration of each loop
+    BOOST_REQUIRE_EQUAL(vj->stop_time_list.size(), 6);
     BOOST_REQUIRE_EQUAL(vj->stop_time_list.front().stop_point->uri, "s1");
     BOOST_REQUIRE_EQUAL(vj->stop_time_list.back().stop_point->uri, "s6");
 
--- a/source/kraken/tests/apply_disruption_test.cpp
+++ b/source/kraken/tests/apply_disruption_test.cpp
@@ -1565,8 +1565,12 @@ BOOST_AUTO_TEST_CASE(add_line_section_impact_on_line_with_repeated_stops) {
     BOOST_CHECK_MESSAGE(ba::ends_with(base_vp.to_string(), "000000"), base_vp);
     // 6 stop_times since we don't impact the first iteration of each loop
     BOOST_REQUIRE_EQUAL(vj->stop_time_list.size(), 6);
-    BOOST_REQUIRE_EQUAL(vj->stop_time_list.front().stop_point->uri, "s1");
-    BOOST_REQUIRE_EQUAL(vj->stop_time_list.back().stop_point->uri, "s6");
+    BOOST_REQUIRE_EQUAL(vj->stop_time_list.at(0).stop_point->uri, "s1");
+    BOOST_REQUIRE_EQUAL(vj->stop_time_list.at(1).stop_point->uri, "s2");
+    BOOST_REQUIRE_EQUAL(vj->stop_time_list.at(2).stop_point->uri, "s3");
+    BOOST_REQUIRE_EQUAL(vj->stop_time_list.at(3).stop_point->uri, "s4");
+    BOOST_REQUIRE_EQUAL(vj->stop_time_list.at(4).stop_point->uri, "s5");
+    BOOST_REQUIRE_EQUAL(vj->stop_time_list.at(5).stop_point->uri, "s6");
 
     vj = b.get<nt::VehicleJourney>("vehicle_journey:vj:2:Adapted:0:line_section_sa1_sa5_routesA1-2-3");
     BOOST_CHECK_EQUAL(vj->meta_vj->get_impacts().size(), 1);
@@ -2533,13 +2537,13 @@ BOOST_AUTO_TEST_CASE(limitation_impact_repeated_stop_points_same_stop_time) {
     BOOST_CHECK_MESSAGE(ba::ends_with(adapted_vp.to_string(), "111110"), adapted_vp);
     BOOST_CHECK_MESSAGE(ba::ends_with(base_vp.to_string(), "000000"), base_vp);
     // The adapted vj should have 3 stop_times
-    // But since the previous vj is adapted we must found the base stop time associated to the adapted
+    // But since the previous vj is adapted we must find the base stop time associated to the adapted
     // one to impact it.
     // However each stop_time have the same departure_time, so we are finding the base stop_time
     // of rank 1 for the stop_time initially of rank 3 (both are on stop_point:20).
     // We are lead to believe it is impacted even if it is not in reality.
     // Nothing we can do really since there is no hard link between the adapted stop_time and base one
-    BOOST_REQUIRE_EQUAL(vj->stop_time_list.size(), 2);
+    BOOST_REQUIRE_EQUAL(vj->stop_time_list.size(), 2); // Should be 3 [10, 20, 40]
     BOOST_REQUIRE_EQUAL(vj->stop_time_list.front().stop_point->uri, "stop_point:10");
     BOOST_REQUIRE_EQUAL(vj->stop_time_list.back().stop_point->uri, "stop_point:40");
 }
--- a/source/kraken/tests/apply_disruption_test.cpp
+++ b/source/kraken/tests/apply_disruption_test.cpp
@@ -2450,3 +2450,96 @@ BOOST_AUTO_TEST_CASE(add_impact_lollipop_vj_impact_stop_only_once) {
     BOOST_REQUIRE_EQUAL(vj->stop_time_list.back().stop_point->uri, "stop_point:40");
 }
 
+// Show a limitation of the current algorithm
+// If we impact a non-base vehicle_journey we need to find the base stop_time associated
+// to each stop_time of the vj to be able to know if its origin rank is impacted or not.
+// Since get_base_stop_time is not perfect (we only check stop_point and departure_time)
+// there is some (dumb) cases where we impact some stops we should not.
+BOOST_AUTO_TEST_CASE(limitation_impact_repeated_stop_points_same_stop_time) {
+    ed::builder b("20190901");
+
+    b.sa("stop_area:1", 0, 0, false, true)("stop_point:10");
+    b.sa("stop_area:2", 0, 0, false, true)("stop_point:20");
+    b.sa("stop_area:3", 0, 0, false, true)("stop_point:30");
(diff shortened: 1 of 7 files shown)


kinnou02 (approved):

kadhikari (approved):
:+1:

source/kraken/tests/apply_disruption_test.cpp
```diff
@@ -2326,3 +2327,219 @@ BOOST_AUTO_TEST_CASE(update_disruption_with_multiple_impact_on_different_vj) {
...
+
+    // Check the adapted vj
+    vj = b.get<nt::VehicleJourney>(
+        "vehicle_journey:vj:1:Adapted:0:line_section_on_line:A_diverted:Adapted:2:line_section_on_line:A_diverted_"
+        "twice");
+    adapted_vp = vj->adapted_validity_pattern()->days;
+    base_vp = vj->base_validity_pattern()->days;
+    BOOST_CHECK_MESSAGE(ba::ends_with(adapted_vp.to_string(), "111110"), adapted_vp);
+    BOOST_CHECK_MESSAGE(ba::ends_with(base_vp.to_string(), "000000"), base_vp);
+    // The adapted vj should have 3 stop_times
```
benoit-bst, line 2539: :+1:

benoit-bst (approved):
The Utests are very clear.
Thank you for catching it :heart:

source/type/message.h
```diff
@@ -411,9 +411,9 @@ class DisruptionHolder {
 struct ImpactedVJ {
     const VehicleJourney* vj;  // vj before impact
     ValidityPattern new_vp;
-    std::set<StopPoint*> impacted_stops;
-    ImpactedVJ(const VehicleJourney* vj, ValidityPattern vp, std::set<StopPoint*> r)
-        : vj(vj), new_vp(vp), impacted_stops(std::move(r)) {}
+    std::set<uint16_t> impacted_ranks;
```
pbougue, line 414: For compile check, it would be nice to type it more (everywhere used), like what's done on `Idx` in https://github.com/CanalTP/utils/blob/f86e6eb15c0d0e84cae8cd6541239592fb5daae4/idx_map.h#L39-L56
```suggestion
    std::set<Rank<StopTime>> impacted_ranks;
```
Or maybe just typedef for clarity (especially in get_sections_ranks where different kind of indexes are mixed) ?
```suggestion
    std::set<stop_time_rank_t> impacted_ranks;
```

----

Still, we could even end-up mixing rank from base-VJ and the ones from rt-VJ... But I don't see what we can do about that :man_shrugging:

source/type/vehicle_journey.cpp
```diff
@@ -227,18 +229,27 @@ ValidityPattern VehicleJourney::get_vp_for_section(const std::set<StopPoint*>& s
     return vp_for_section;
 }
 
-std::set<StopPoint*> VehicleJourney::get_sections_stop_points(const StopArea* start_stop,
-                                                              const StopArea* end_stop) const {
+std::set<uint16_t> VehicleJourney::get_sections_ranks(const StopArea* start_stop, const StopArea* end_stop) const {
```
pbougue: As there are StopTime, StopPoint, StopArea used after, for clarity (maybe other var need renames):
```suggestion
std::set<uint16_t> VehicleJourney::get_sections_ranks(const StopArea* start_sa, const StopArea* end_sa) const {
```

source/type/vehicle_journey.cpp
```diff
@@ -267,31 +278,49 @@ std::set<StopPoint*> VehicleJourney::get_sections_stop_points(const StopArea* st
...
     const auto* vj = base_vj ? base_vj : this;
     for (const auto& st : vj->stop_time_list) {
         // The line section is using stop_areas so we make sure we have one
         if (!st.stop_point || !st.stop_point->stop_area) {
             continue;
         }
+        const auto* sa = st.stop_point->stop_area;
+
+        // Must close section before potentially starting a new one
+        if (section_ending && *last_area_idx != sa->idx) {
```
pbougue: At that point `last_area_idx` is always `end_stop->idx`so IMHO this looks easier to understand:
```suggestion
        if (section_ending && *end_stop->idx != sa->idx) {
```

source/type/vehicle_journey.cpp
```diff
@@ -267,31 +278,49 @@ std::set<StopPoint*> VehicleJourney::get_sections_stop_points(const StopArea* st
      *  - s1/s2 from the 5th to the 10th,
      *  - s1/s2/s5 from the 11th to the 15th
      * */
-    std::set<StopPoint*> res;
+    std::set<uint16_t> res;
     boost::optional<uint16_t> start_idx;
```
pbougue:
```suggestion
    boost::optional<uint16_t> section_start_rank;
```

source/type/vehicle_journey.cpp
```diff
@@ -267,31 +278,49 @@ std::set<StopPoint*> VehicleJourney::get_sections_stop_points(const StopArea* st
...
+            start_idx = boost::none;
+            section_ending = false;
+        }
 
         // we want the shortest section, thus, we set the idx each
-        // time we see it
-        if (st.stop_point->stop_area->idx == start_stop->idx) {
+        // time we see it, except if we are in the same area than the start idx
+        // Reset start_idx if the previous area was different
+        if (sa->idx == start_stop->idx && (!start_idx || !last_area_idx || *last_area_idx != sa->idx)) {
```
pbougue: I would find it easier to read with a `section_starting` flag with pretty much the same logic than section_ending.

source/kraken/tests/apply_disruption_test.cpp
```diff
@@ -1563,7 +1563,8 @@ BOOST_AUTO_TEST_CASE(add_line_section_impact_on_line_with_repeated_stops) {
     base_vp = vj->base_validity_pattern()->days;
     BOOST_CHECK_MESSAGE(ba::ends_with(adapted_vp.to_string(), "111111"), adapted_vp);
     BOOST_CHECK_MESSAGE(ba::ends_with(base_vp.to_string(), "000000"), base_vp);
-    BOOST_REQUIRE_EQUAL(vj->stop_time_list.size(), 2);
+    // 6 stop_times since we don't impact the first iteration of each loop
+    BOOST_REQUIRE_EQUAL(vj->stop_time_list.size(), 6);
     BOOST_REQUIRE_EQUAL(vj->stop_time_list.front().stop_point->uri, "s1");
     BOOST_REQUIRE_EQUAL(vj->stop_time_list.back().stop_point->uri, "s6");
```
pbougue: It would be nice to add intermediate stops too... s1 s2 s3 s4 s5 s6 if I'm correct.

source/type/tests/test.cpp
```diff
@@ -224,42 +224,42 @@ BOOST_AUTO_TEST_CASE(tz_handler_overflow_test) {
...
 }
 
-BOOST_AUTO_TEST_CASE(get_sections_stop_points) {
+BOOST_AUTO_TEST_CASE(get_sections_ranks) {
     ed::builder b("20120614");
-    // 0 1 0 2 3 0 4 2
-    const auto* vj = b.vj("A")("0", "09:00"_t)("1", "09:00"_t)("0", "09:00"_t)("2", "09:00"_t)("3", "09:00"_t)(
-                          "0", "09:00"_t)("4", "09:00"_t)("2", "09:00"_t)("5", "09:00"_t)("2", "09:00"_t)
+
+    b.sa("0", 0, 0, false, true)("01")("02")("03");
```
pbougue: For clarity:
```suggestion
    b.sa("0", 0, 0, false, true)("0a")("0b")("0c");
```

source/kraken/tests/apply_disruption_test.cpp
```diff
@@ -2326,3 +2327,219 @@ BOOST_AUTO_TEST_CASE(update_disruption_with_multiple_impact_on_different_vj) {
...
+    // Check the adapted vj
+    vj = b.get<nt::VehicleJourney>(
+        "vehicle_journey:vj:1:Adapted:0:line_section_on_line:A_diverted:Adapted:2:line_section_on_line:A_diverted_"
+        "twice");
+    adapted_vp = vj->adapted_validity_pattern()->days;
+    base_vp = vj->base_validity_pattern()->days;
+    BOOST_CHECK_MESSAGE(ba::ends_with(adapted_vp.to_string(), "111110"), adapted_vp);
+    BOOST_CHECK_MESSAGE(ba::ends_with(base_vp.to_string(), "000000"), base_vp);
+    // The adapted vj should have 3 stop_times
+    // But since the previous vj is adapted we must found the base stop time associated to the adapted
```
pbougue: minor
```suggestion
    // But since the previous vj is adapted we must find the base stop time associated to the adapted
```

source/kraken/tests/apply_disruption_test.cpp
```diff
@@ -2326,3 +2327,219 @@ BOOST_AUTO_TEST_CASE(update_disruption_with_multiple_impact_on_different_vj) {
...
+    // one to impact it.
+    // However each stop_time have the same departure_time, so we are finding the base stop_time
+    // of rank 1 for the stop_time initially of rank 3 (both are on stop_point:20).
+    // We are lead to believe it is impacted even if it is not in reality.
+    // Nothing we can do really since there is no hard link between the adapted stop_time and base one
+    BOOST_REQUIRE_EQUAL(vj->stop_time_list.size(), 2);
```
pbougue: Can you just add a comment at this line for clarity in case we ever correct it (in addition to above)?
```suggestion
    BOOST_REQUIRE_EQUAL(vj->stop_time_list.size(), 2);  // Should be 3: [10, 20, 40]
```

pbougue (commented):
This is all OK to me, only changes suggested are for clarity as the logic is quite complex.
(and may even get more complex in that area of the code with delays, advance and stop-adds).

Thank you so much for tests and commit messages :+1:

source/type/message.h
eturck: Like we said on Riot didn't have time to do it sorry :(. But feel free to do it.

source/type/vehicle_journey.cpp
eturck: Thanks for the suggestion

source/type/vehicle_journey.cpp
eturck: I simplified it, let me know if it's ok for you

source/type/vehicle_journey.cpp
eturck: 👍

source/type/vehicle_journey.cpp
eturck: I hope it's better like this

source/kraken/tests/apply_disruption_test.cpp
eturck: Good idea thanks :)

source/type/tests/test.cpp
eturck: 👍

eturck:
I will be away for the next couple of weeks. If there is any question don't hesitate to ping @vpassama :)

source/type/message.h
pbougue: Strong-typing it is an important job, worth separating PR.
I will try to open a separate one after merge of yours ;)

pbougue (approved):
:candy: super sweet case :+1:

PR #2961, opened 2019-09-24
Osm2ed: do not put transport object from OSM in autocompletion

kinnou02:
OSM contains some transport objects, like quay: https://www.openstreetmap.org/way/364024146#map=19/45.76195/4.85761
We do not want to show these object in the autocompletion as we already
have them from our transport datasets. This PR changes the way these
objects are imported to make them invisible, this way they won't be
shown in the autocompletion but will still be used for routing.

Hand testes with lyon and transilien, a few hundred way are imported as
invisible, all of them look like stops.

Refer to: https://jira.kisio.org/browse/NAVITIAII-2880

### DIFF
--- a/source/ed/connectors/osm_tags_reader.cpp
+++ b/source/ed/connectors/osm_tags_reader.cpp
@@ -73,6 +73,7 @@ std::bitset<8> parse_way_tags(const std::map<std::string, std::string>& tags) {
     int bike_direct = unknown;
     int bike_reverse = unknown;
     int foot = unknown;
+    bool visible = true;
 
     for (std::pair<std::string, std::string> pair : tags) {
         std::string key = pair.first, val = pair.second;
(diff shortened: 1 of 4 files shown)


SGrenet (approved):

benoit-bst (approved):

source/ed/connectors/osm_tags_reader.cpp
```diff
@@ -205,6 +206,17 @@ std::bitset<8> parse_way_tags(const std::map<std::string, std::string>& tags) {
                 foot = foot_allowed;
             }
         }
+
+        // We do not want to autocomplete on public transport objects
+        if (key == "public_transport") {
```
nonifier, line 211: What about excluding all objects with a key="public_transport". None of the values seem to be interesting to us - https://wiki.openstreetmap.org/wiki/Key:public_transport

nonifier (approved):

source/ed/osm2ed.h
```diff
@@ -195,6 +198,9 @@ struct OSMWay {
 
     void set_name(const std::string& name) const { this->name = name; }
 
+    void set_visible(bool visible) { properties[VISIBLE] = visible; }
```
pbougue: minor: unused method

pbougue (approved):

PR #2962, opened 2019-09-24
ed: convert fare OD to lower case too

kinnou02:
Since #2944 fare is broken on OD, everything was converted to lower
case, except the OD. This PR fix that by lowering the case of every
identifier in fare_parser

### DIFF
--- a/source/ed/connectors/fare_parser.cpp
+++ b/source/ed/connectors/fare_parser.cpp
@@ -127,6 +127,7 @@ void fare_parser::load_od() {
     // .... (with any number of ticket)
 
     int count = 0;
+    std::locale locale;
     for (row = reader.next(); !reader.eof(); row = reader.next()) {
         if (row.size() < 7) {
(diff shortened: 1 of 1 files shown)


nonifier (approved):
I hope this will not introduce conflicts of two identical strings with different casing :crossed_fingers:

benoit-bst (approved):

pbougue (approved):
@nonifier it's only on the application conditions that read is changed https://github.com/CanalTP/ntfs-specification/blob/master/ntfs_fare_extension_fr_deprecated.md#od_farescsv , not on stop_area's uris themselves.
So:
* it shouldn't introduce duplicate uris for SA.
* when lowercased SA's uris may have duplicates, therefore navitia may mix them for fare computing, but this was already possible before, as it was a case-insensitive check.

pbougue (commented):
minor: adding tests on that would be nice.

kadhikari (approved):

kinnou02:
There is no end to end test in navitia, these kind of test should be on artemis

PR #2964, opened 2019-09-25
gzip chaos db dump used as fixture

kinnou02:
It bothers me that navitia is now compose of 38% of sql :(
image

This file will never be updated manually we do not need diff on it.
:mushroom:


pbougue (approved):
Waaaah :mushroom: :mushroom: :mushroom:

nonifier (approved):

PR #2960, opened 2019-09-24
ptref: Add new method for filtering route direction type

benoit-bst:
First part for **JIRA** : https://jira.kisio.org/browse/NAVP-1087, _Add new method for filtering route_ **direction type**.
The method is specialized onfy for **route** like `route.has_direction_type(blabla)`
I started to create a generic method but it was useless for this special _route case_ (it was in the first commit for this PR).
This is **case-sensitive**. The filter value has to match perfectly with the direction type format inside the data. Into the data, it seems to be _lower style_.
RFC doc updated.
An UTest was added and it demonstrates several case.

**QA part** :
Tested several request like 
```
url/v1/coverage/fr-idf/stop_areas/id_stop_area/departures?filter=route.has_direction_type(forward)
url/v1/coverage/fr-idf/stop_areas/id_stop_area/departures?filter=route.has_direction_type(backward)

url/v1/coverage/fr-idf/stop_areas/id_stop_area/arrivals?filter=route.has_direction_type(forward)
url/v1/coverage/fr-idf/stop_areas/id_stop_area/arrivals?filter=route.has_direction_type(backward)


url/v1/coverage/fr-idf/lines/id_lines/lines?filter=route.has_direction_type(forward)
```
And it works

PR #2967, opened 2019-09-26
jormun: add direction_type parameter for departures/arrivals/stop_schedules/route_schedules Api

benoit-bst:
**Feature** : Allow to filter `/departures`, `/arrivals`, `/stop_schedules`, `/route_schedules` results with direction type parameter (all, forward, backward)
JIRA : https://jira.kisio.org/browse/NAVP-1087

The first part of the work (update ptref filter) was added here : https://github.com/CanalTP/navitia/pull/2960

- Docker tests are Green.
- I know what you will say: There is no tests for `/arrivals` Api. But don't worry about that, because _departures_ and _arrival_ use the same **Schedule** **class**.

**QA part**:
Tested on fr-se-lyon, several request with different data type (forward and backward) and it seems to be ok for all params (forward, backward, all, without param) :tada:

### DIFF
--- a/documentation/rfc/ptref_grammar.md
+++ b/documentation/rfc/ptref_grammar.md
@@ -82,7 +82,7 @@ In the following table, if the invocation starts with `collection`, any collecti
 |`poi.within(distance, coord)`|all the POIs within `distance` meters from `coord`|`distance` in a positive integer, `coord` is a navitia coord (`lon;lat`)|
 |`stop_area.within(distance, coord)`|as above for the stop areas|as above|
 |`stop_point.within(distance, coord)`|as above for the stop points|as above|
-|`route.has_direction_type(type1, type2, ...)`|all the `route` containing a given `direction_type` into list|at least one direction type must be given|
+|`route.has_direction_type(type1, type2, ...)`|all the routes whose `direction_type` is any of those provided|at least one direction type must be given|
 
 #### Interpretation
 
--- a/documentation/slate/source/includes/apis.md
+++ b/documentation/slate/source/includes/apis.md
@@ -1438,15 +1438,16 @@ access it via that kind of url: <https://api.navitia.io/v1/{a_path_to_a_resource
 
 ### Parameters
 
-Required | Name               | Type      | Description                                                                              | Default Value
----------|--------------------|-----------|------------------------------------------------------------------------------------------|--------------
-nop      | from_datetime      | [iso-date-time](#iso-date-time) | The date_time from which you want the schedules                    | the current datetime
-nop      | duration           | int       | Maximum duration in seconds between from_datetime and the retrieved datetimes.           | 86400
-nop      | depth              | int       | Json response [depth](#depth)                                                            | 1
-nop      | items_per_schedule | int       | Maximum number of columns per schedule.                                                  |
-nop      | forbidden_uris[]   | id        | If you want to avoid lines, modes, networks, etc.                                        |
-nop      | data_freshness     | enum      | Define the freshness of data to use<br><ul><li>realtime</li><li>base_schedule</li></ul>  | base_schedule
-nop      | disable_geojson    | boolean   | remove geojson fields from the response                                                  | False
+Required | Name               | Type      | Description                                                                                                                | Default Value
+---------|--------------------|-----------|----------------------------------------------------------------------------------------------------------------------------|--------------
+nop      | from_datetime      | [iso-date-time](#iso-date-time) | The date_time from which you want the schedules                                                      | the current datetime
+nop      | duration           | int       | Maximum duration in seconds between from_datetime and the retrieved datetimes.                                             | 86400
+nop      | depth              | int       | Json response [depth](#depth)                                                                                              | 1
+nop      | items_per_schedule | int       | Maximum number of columns per schedule.                                                                                    |
+nop      | forbidden_uris[]   | id        | If you want to avoid lines, modes, networks, etc.                                                                          |
+nop      | data_freshness     | enum      | Define the freshness of data to use<br><ul><li>realtime</li><li>base_schedule</li></ul>                                    | base_schedule
+nop      | disable_geojson    | boolean   | remove geojson fields from the response                                                                                    | False
+nop      | direction_type     | enum      | Allow to filter the response with the route direction type property <ul><li>all</li><li>forward</li><li>backward</li></ul> | all
 
 ### Objects
 
@@ -1571,15 +1572,16 @@ See how disruptions affect stop schedules in the [real time](#realtime) section.
 
 ### Parameters
 
-Required | Name               | Type                            | Description        | Default Value
----------|--------------------|---------------------------------|--------------------|--------------
-nop      | from_datetime      | [iso-date-time](#iso-date-time) | The date_time from which you want the schedules | the current datetime
-nop      | duration           | int                             | Maximum duration in seconds between from_datetime and the retrieved datetimes.                            | 86400
-nop      | depth              | int                             | Json response [depth](#depth) | 1
-nop      | forbidden_uris[]   | id                              | If you want to avoid lines, modes, networks, etc.    |
-nop      | items_per_schedule | int                             | Maximum number of datetimes per schedule.                                                  |
-nop      | data_freshness     | enum                            | Define the freshness of data to use to compute journeys <ul><li>realtime</li><li>base_schedule</li></ul> | realtime
-nop      | disable_geojson    | boolean                         | remove geojson fields from the response | False
+Required | Name               | Type                            | Description                                                                                                                | Default Value
+---------|--------------------|---------------------------------|----------------------------------------------------------------------------------------------------------------------------|--------------
+nop      | from_datetime      | [iso-date-time](#iso-date-time) | The date_time from which you want the schedules                                                                            | the current datetime
+nop      | duration           | int                             | Maximum duration in seconds between from_datetime and the retrieved datetimes.                                             | 86400
+nop      | depth              | int                             | Json response [depth](#depth)                                                                                              | 1
+nop      | forbidden_uris[]   | id                              | If you want to avoid lines, modes, networks, etc.                                                                          |
+nop      | items_per_schedule | int                             | Maximum number of datetimes per schedule.                                                                                  |
+nop      | data_freshness     | enum                            | Define the freshness of data to use to compute journeys <ul><li>realtime</li><li>base_schedule</li></ul>                   | realtime
+nop      | disable_geojson    | boolean                         | remove geojson fields from the response                                                                                    | False
+nop      | direction_type     | enum                            | Allow to filter the response with the route direction type property <ul><li>all</li><li>forward</li><li>backward</li></ul> | all
 
 
 ### <a name="stop-schedule"></a>Stop_schedule object
--- a/documentation/slate/source/includes/apis.md
+++ b/documentation/slate/source/includes/apis.md
@@ -1672,15 +1672,16 @@ See how disruptions affect the next departures in the [real time](#realtime) sec
 
 ### Parameters
 
-Required | Name             | Type                            | Description                                                                                              | Default Value
----------|------------------|---------------------------------|----------------------------------------------------------------------------------------------------------|--------------
-nop      | from_datetime    | [iso-date-time](#iso-date-time) | The date_time from which you want the schedules                                                          | the current datetime
-nop      | duration         | int                             | Maximum duration in seconds between from_datetime and the retrieved datetimes.                           | 86400
-nop      | count            | int                             | Maximum number of results.                                                                               | 10
-nop      | depth            | int                             | Json response [depth](#depth)                                                                            | 1
-nop      | forbidden_uris[] | id                              | If you want to avoid lines, modes, networks, etc.                                                        |
-nop      | data_freshness   | enum                            | Define the freshness of data to use to compute journeys <ul><li>realtime</li><li>base_schedule</li></ul> | realtime
-nop      | disable_geojson  | boolean                         | remove geojson fields from the response                                                                  | false
+Required | Name             | Type                            | Description                                                                                                                           | Default Value
+---------|------------------|---------------------------------|---------------------------------------------------------------------------------------------------------------------------------------|--------------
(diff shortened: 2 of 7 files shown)


source/ptreferential/tests/ptref_ng_test.cpp
```diff
@@ -76,8 +76,8 @@ BOOST_AUTO_TEST_CASE(parse_pred) {
         R"#(vehicle_journey . has_code_type ( external_code ) )#",
         R"#(vehicle_journey.has_code_type("external_code"))#");
     assert_expr(
-        R"#(route . has_code_type ( forward ) )#",
-        R"#(route.has_code_type("forward"))#");
+        R"#(route . has_direction_type ( forward ) )#",
```
benoit-bst, line 79: because of late review : https://github.com/CanalTP/navitia/pull/2960#discussion_r328495559

documentation/rfc/ptref_grammar.md
```diff
@@ -82,7 +82,7 @@ In the following table, if the invocation starts with `collection`, any collecti
 |`poi.within(distance, coord)`|all the POIs within `distance` meters from `coord`|`distance` in a positive integer, `coord` is a navitia coord (`lon;lat`)|
 |`stop_area.within(distance, coord)`|as above for the stop areas|as above|
 |`stop_point.within(distance, coord)`|as above for the stop points|as above|
-|`route.has_direction_type(type1, type2, ...)`|all the `route` containing a given `direction_type` into list|at least one direction type must be given|
+|`route.has_direction_type(type1, type2, ...)`|all the routes whose `direction_type` is any of those provided|at least one direction type must be given|
```
benoit-bst, line 85: because of late review : https://github.com/CanalTP/navitia/pull/2960#discussion_r328497174

source/time_tables/tests/departure_board_test_data.h
```diff
@@ -39,11 +39,13 @@ struct departure_board_fixture {
     departure_board_fixture() : b("20160101") {
         b.sa("SA1")("S1")("S2");
 
-        b.vj("A").name("A:vj1")("A:s", "08:00"_t)("S1", "09:00"_t)("S2", "10:00"_t)("A:e", "11:00"_t);
+        b.vj("A")
+            .route("A:0", "forward")
```
benoit-bst, line 43: I only added forward and backward into the test data set, because it seems complicated to add an other **line** with **route**(to test other direction_type like clockwise, inbound, etc...) without broken **several tests**...

nonifier (approved):

source/jormungandr/jormungandr/interfaces/v1/Schedules.py
```diff
@@ -162,6 +168,31 @@ def __init__(self, endpoint, *args, **kwargs):
...
+            elif direction_type == 'all':
+                values = 'all'
+            else:
+                abort(
+                    404,
+                    message='wrong direction type parameter selected : {}, it should be [forward, backward, all]]'.format(
+                        direction_type
+                    ),
+                )
+            return "route.has_direction_type(" + values + ")"
```
kinnou02: Please use format not string concatenation

source/jormungandr/jormungandr/interfaces/v1/Schedules.py
```diff
@@ -162,6 +168,31 @@ def __init__(self, endpoint, *args, **kwargs):
...
+
+        if filter != "":
+            return filter + " and " + create_direction_type_filter(direction_type)
```
kinnou02: you should add parenthesis to any problems with operator priorities

source/jormungandr/jormungandr/interfaces/v1/Schedules.py
```diff
@@ -203,6 +234,11 @@ def get(self, uri=None, region=None, lon=None, lat=None):
             self.region = i_manager.get_region(region, lon, lat)
         timezone.set_request_timezone(self.region)
 
+        # create direction type filter
+        if args['direction_type'] and args['direction_type'] != 'all':
+            args['filter'] = self._add_direction_type_filter(args['direction_type'], args['filter'])
+        logging.getLogger(__name__).debug("Schedule filter: {}".format(args["filter"]))
```
kinnou02:
```suggestion
        logging.getLogger(__name__).debug("Schedule filter: %s" , args["filter"])
```
It is recommended to use logging own string formatting system as it will not be run if the log isn't displayed while with `format()` it will always be executed even if we discard the log latter on.

source/jormungandr/jormungandr/interfaces/v1/Schedules.py
```diff
@@ -154,6 +155,11 @@ def __init__(self, endpoint, *args, **kwargs):
         parser_get.add_argument(
             "disable_geojson", type=BooleanType(), default=False, help="remove geojson from the response"
         )
+        parser_get.add_argument(
+            "direction_type",
+            help='Provide a route direction to filter results.',
```
kinnou02:
```suggestion
            help='Provide a route direction type to filter results.',
```

source/jormungandr/jormungandr/interfaces/v1/Schedules.py
```diff
@@ -162,6 +168,31 @@ def __init__(self, endpoint, *args, **kwargs):
...
+            else:
+                abort(
+                    404,
+                    message='wrong direction type parameter selected : {}, it should be [forward, backward, all]]'.format(
+                        direction_type
+                    ),
+                )
+            return "route.has_direction_type(" + values + ")"
+
+        if filter != "":
```
kinnou02:
```suggestion
        if filter:
```

kinnou02 (changes requested):

kinnou02 (approved):

benoit-bst:
**Update** :

The scope of the Tickket was around of **departures** and **arrivals** API. But from the moment we update the Schedule Class used inside **departures** and **arrival**. It's automatically update for **stop_schedules** and **route_schedules** too. So, this 2 APIs have the direction_type option too.

No need to had new tests for that because it is the same way of **departures** and it already coded.
I tested by hands to be certain, with **stop_schedules** and **route_schedules** and it works.
I'll just update the doc :scroll:

source/jormungandr/tests/departure_board_tests.py
```diff
@@ -931,6 +931,42 @@ def test_departure(self):
...
+        assert len(response["departures"]) == 3
+
+        # With direction type = backward
+        response = self.query_region(
+            "stop_points/S1/departures?direction_type=backward&_current_datetime=20160101T080000"
+        )
+        assert "departures" in response
+        assert len(response["departures"]) == 1
+
     def test_departure_base_sched(self):
```
kadhikari, line 970: Tests are very clear :+1:

kadhikari (approved):

documentation/slate/source/includes/apis.md
```diff
@@ -1438,15 +1438,16 @@ access it via that kind of url: <https://api.navitia.io/v1/{a_path_to_a_resource
...
+nop      | depth              | int       | Json response [depth](#depth)                                                                                              | 1
+nop      | items_per_schedule | int       | Maximum number of columns per schedule.                                                                                    |
+nop      | forbidden_uris[]   | id        | If you want to avoid lines, modes, networks, etc.                                                                          |
+nop      | data_freshness     | enum      | Define the freshness of data to use<br><ul><li>realtime</li><li>base_schedule</li></ul>                                    | base_schedule
+nop      | disable_geojson    | boolean   | remove geojson fields from the response                                                                                    | False
+nop      | direction_type     | enum      | Allow to filter the response with the route direction type property <ul><li>all</li><li>forward</li><li>backward</li></ul> | all
```
pbougue: Please add that forward filters for forward, inbound and clockwise and backward for backward, outbound and anticlockwise.
This is not straightforward, so we have to precise it.
(same on all occurences)

source/jormungandr/jormungandr/interfaces/v1/Schedules.py
```diff
@@ -154,6 +155,11 @@ def __init__(self, endpoint, *args, **kwargs):
         parser_get.add_argument(
             "disable_geojson", type=BooleanType(), default=False, help="remove geojson from the response"
         )
+        parser_get.add_argument(
+            "direction_type",
+            help='Provide a route direction type to filter results.',
```
pbougue: Please also precise here that forward and backward are extended.

source/jormungandr/jormungandr/interfaces/v1/Schedules.py
```diff
@@ -162,6 +168,31 @@ def __init__(self, endpoint, *args, **kwargs):
     def options(self, **kwargs):
         return self.api_description(**kwargs)
 
+    def _add_direction_type_filter(self, direction_type, filter):
+
+        # (forward, clokwise, inbound) are equivalent and
```
pbougue: minor
```suggestion
        # (forward, clockwise, inbound) are equivalent and
```

source/jormungandr/jormungandr/interfaces/v1/Schedules.py
```diff
@@ -162,6 +168,31 @@ def __init__(self, endpoint, *args, **kwargs):
...
+        # (backward, anticlokwise, outbound) are equivalent too.
```
pbougue: minor
```suggestion
        # (backward, anticlockwise, outbound) are equivalent too.
```

source/jormungandr/jormungandr/interfaces/v1/Schedules.py
```diff
@@ -162,6 +168,31 @@ def __init__(self, endpoint, *args, **kwargs):
...
+        def create_direction_type_filter(direction_type):
+            if direction_type == 'forward':
+                values = 'forward,clockwise,inbound'
+            elif direction_type == 'backward':
+                values = 'backward,anticlockwise,outbound'
+            elif direction_type == 'all':
+                values = 'all'
```
pbougue: In that case, you will return only the routes whose direction_type is `'all'` no?

source/jormungandr/jormungandr/interfaces/v1/Schedules.py
```diff
@@ -203,6 +234,11 @@ def get(self, uri=None, region=None, lon=None, lat=None):
             self.region = i_manager.get_region(region, lon, lat)
         timezone.set_request_timezone(self.region)
 
+        # create direction type filter
+        if args['direction_type'] and args['direction_type'] != 'all':
```
pbougue: I would find it more straightforward to check `'all'` case only inside `_add_direction_type_filter`, as for now you manage this case in 2 parts of the code (this check prevents bug from previous comment I think).

source/time_tables/tests/departure_board_test_data.h
```diff
@@ -39,11 +39,13 @@ struct departure_board_fixture {
...
+            .name("A:vj1")("A:s", "08:00"_t)("S1", "09:00"_t)("S2", "10:00"_t)("A:e", "11:00"_t);
         b.vj("A").name("A:vj2")("A:s", "09:00"_t)("S1", "10:00"_t)("S2", "11:00"_t)("A:e", "12:00"_t);
         b.vj("A").name("A:vj3")("A:s", "10:00"_t)("S1", "11:00"_t)("S2", "12:00"_t)("A:e", "13:00"_t);
 
-        b.vj("B").name("B:vj1")("B:s", "10:30"_t)("S1", "11:30"_t)("B:e", "12:30"_t);
+        b.vj("B").route("B:1", "backward").name("B:vj1")("B:s", "10:30"_t)("S1", "11:30"_t)("B:e", "12:30"_t);
```
pbougue: Can you test that backward is larger than just backward? :wink: 
```suggestion
        b.vj("B").route("B:1", "anticlockwise").name("B:vj1")("B:s", "10:30"_t)("S1", "11:30"_t)("B:e", "12:30"_t);
```

pbougue (changes requested):

source/jormungandr/jormungandr/interfaces/v1/Schedules.py
benoit-bst: For the `all` case, we don't need to create a filter. 
This is the case with https://github.com/CanalTP/navitia/pull/2967/files#diff-71deae639b1c21f792ab41c775252465R238
This case is just for being logical into the **create_direction_type_filter** function, in case of we'll use it elsewhere.

source/jormungandr/jormungandr/interfaces/v1/Schedules.py
benoit-bst: I understand your point of you and i coded like that in first step. 
I have choosen that to highlight the fact of "skip process", when it is `all` case.

source/time_tables/tests/departure_board_test_data.h
benoit-bst: **backward** is identical to **anticlockwise**. So If i test **anticlockwise**, I don't test **backward**. I prefer to let **backward** because it is more famous :star:

pbougue (commented):
Also, this does not handle the ticket part of optimizing calls to external RT providers.
So we can't close the ticket after this PR.

documentation/slate/source/includes/apis.md
benoit-bst: is it good for you ?
new_doc

source/jormungandr/jormungandr/interfaces/v1/Schedules.py
benoit-bst: An other reason is, when you don't need to add another direction_type filter option, we want to avoid to call the add_direction_type_filter function. It avoids, inside, to re-copy the filter into the output of the function. We avoid a useless copy. You see what I mean ?

But whatever, if it's more clear, i prefer to incorporate the condition inside. 
To be clear is more important says the wise :older_man:

source/jormungandr/jormungandr/interfaces/v1/Schedules.py
benoit-bst: Okay, I understood your catch. If we create, a filter like `route.has_direction_type(all)`, it is not good, you have right. But it is not possible into this code because I have protected with a `if` before.

I just have to remove the `elif direction_type == "all"` for fix that. The `all` case will go through the abort. It will be more clear. 
Thanks for the catch :ok_hand:

pbougue (approved):

PR #2968, opened 2019-09-26
[Jormun] Adding max_mode_direct_path_duration in /status and /instances

AurelienLP:


kinnou02 (approved):

pbougue (approved):
Side-question: any reason why there is no `max_direct_path_duration` to default them all?
(like for max_duration_to_pt)

Response: it's just an omission, it may be added in a close future if need arises.

nonifier (approved):

mbounabi (dismissed):
Fix Travis tests, or mergify will click :ghost:

pbougue (approved):

source/tyr/tests/integration/instance_test.py
```diff
@@ -382,6 +388,40 @@ def test_update_additional_time_for_taxi(create_instance):
...
+        'max_taxi_direct_path_duration': 1523,
+        'max_ridesharing_direct_path_duration': 4456,
+    }
+    resp = api_put('/v0/instances/fr', data=json.dumps(params), content_type='application/json')
+    assert resp['max_walking_direct_path_duration'] == 3475
+    assert resp['max_bike_direct_path_duration'] == 9512
+    assert resp['max_bss_direct_path_duration'] == 7456
+    assert resp['max_car_direct_path_duration'] == 3214
+    assert resp['max_taxi_direct_path_duration'] == 1523
+    assert resp['max_ridesharing_direct_path_duration'] == 4456
```
mbounabi: Instead of all the asserts:
```suggestion
   for key in params.keys():
       assert resp[key] == params[key]
```
Same for the GET request below

kinnou02 (approved):

mbounabi (approved):

PR #2970, opened 2019-09-30
Enable mergify

kinnou02:
Automatically merge PR done by @CanalTP/coreteam that have
  - at least 2 approved reviews
  - no comment or reuqest for change reviews
  - all requested reviewver have approved
  - all CI tests are ok
  - no label wip, dot_not_merge, etc...

PR are rebased on `dev` if needed and tests are rerun before merge.

Reviews are dismissed when a push occurs.

### DIFF
--- a/.mergify.yml
+++ b/.mergify.yml
@@ -1,0 +1,30 @@
+pull_request_rules:
+  - name: automatic merge when CI passes and 2 reviews
+    conditions:
+      - "#approved-reviews-by>=2"
+      - "#review-requested=0"
+      - "#changes-requested-reviews-by=0"
+      - "#commented-reviews-by=0"
+      - status-success=Travis CI - Pull Request
+      - status-success=default
+      - status-success=codecov/project
+      - status-success=codecov/patch
+      - base=dev
+      - label!=dot_not_merge
+      - label!=wip
+      - label!=waiting
+      - label!=merge_after_release
+      - author=@CoreTeam
+    actions:
+      merge:
+        method: merge
+        strict: true
+  - name: delete head branch after merge
+    conditions: []
+    actions:
+      delete_head_branch: {}
+  - name: remove outdated reviews
+    conditions:
+      - base=master
+    actions:
+      dismiss_reviews: {}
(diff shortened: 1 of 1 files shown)


AurelienLP (approved):

pbougue (commented):
This conf is the best I can think of (a bit of delay might be good but not doable).
I still vote to wait for a test on other repos, but I will follow the move if the majority is onboard!

pbougue (commented):
Can we do something about sub-repos needing a merge (then update parent repo) before actual merge?
Just add do-not-merge is OK, if we can think of some :robot:-thing for that it would be better.

kinnou02:
> Can we do something about sub-repos needing a merge (then update parent repo) before actual merge?
> Just add do-not-merge is OK, if we can think of some -thing for that it would be better.

This should be handled by ci, and I think @nonifier did something about this

benoit-bst (approved):

.mergify.yml
```diff
@@ -0,0 +1,30 @@
...
+      - status-success=continuous-integration/travis-ci/pr
+      - status-success=default
+      - status-success=codecov/project
+      - status-success=codecov/patch
+      - base=dev
+      - label!=dot_not_merge
+      - label!=wip
+      - label!=waiting
+      - label!=merge_after_release
+      - author=@CoreTeam
```
datanel:
```suggestion
      - author=@coreteam
```

datanel: this  should be the id, not the name (tested in https://github.com/CanalTP/transit_model)

.mergify.yml
```diff
@@ -0,0 +1,30 @@
+pull_request_rules:
+  - name: automatic merge when CI passes and 2 reviews
+    conditions:
+      - "#approved-reviews-by>=2"
+      - "#review-requested=0"
+      - "#changes-requested-reviews-by=0"
+      - "#commented-reviews-by=0"
```
datanel, line 7: if you approve and then comment, you (all the reviewers) must re-approve

nonifier (approved):

PR #2973, opened 2019-09-30
jormun: use flask_restful from upstream

kinnou02:
We do not need this ugly fork since we only updated some marshal stuff


pbougue (commented):
:tada:

pbougue (approved):

benoit-bst (approved):

PR #2977, opened 2019-10-01
[Jormungandr] use best PT journey for too_many_connections_filter

xlqian:
https://jira.kisio.org/browse/NAVP-574 
little fix on the too many connections filter

Tests passed on Artemis https://ci.navitia.io/job/custom_Artemis/7/console

### DIFF
--- a/source/jormungandr/jormungandr/scenarios/journey_filter.py
+++ b/source/jormungandr/jormungandr/scenarios/journey_filter.py
@@ -391,14 +391,14 @@ class FilterTooLongDirectPath(SingleJourneyFilter):
         return max_duration > journey.duration
 
 
-def get_best_connections(journeys, request):
+def get_best_pt_journey_connections(journeys, request):
     """
-    Returns min connection count among journeys
+    Returns the nb of connection of the best pt_journey
     Returns None if journeys empty
     """
     if not journeys:
         return None
-    best = get_ASAP_journey(journeys, request)
+    best = get_ASAP_journey((j for j in journeys if 'non_pt' not in j.tags), request)
     return get_nb_connections(best) if best else 0
 
 
@@ -712,7 +712,7 @@ def _filter_similar_journeys(journey_pairs_pool, request, similar_journey_genera
 def _filter_too_much_connections(journeys, instance, request):
     """
     eliminates journeys with a number of connections strictly superior to the
-    minimum number of connections among all journeys + _max_additional_connections
+    the number of connections of the best pt_journey + _max_additional_connections
     """
     logger = logging.getLogger(__name__)
     max_additional_connections = get_or_default(
@@ -721,10 +721,10 @@ def _filter_too_much_connections(journeys, instance, request):
     import itertools
 
     it1, it2 = itertools.tee(journeys, 2)
-    min_connections = get_best_connections(it1, request)
+    best_pt_journey_connections = get_best_pt_journey_connections(it1, request)
     is_debug = request.get('debug', False)
-    if min_connections is not None:
-        max_connections_allowed = max_additional_connections + min_connections
+    if best_pt_journey_connections is not None:
+        max_connections_allowed = max_additional_connections + best_pt_journey_connections
         for j in it2:
             if get_nb_connections(j) > max_connections_allowed:
                 logger.debug("the journey {} has a too much connections, we delete it".format(j.internal_id))
--- a/source/jormungandr/jormungandr/scenarios/journey_filter.py
+++ b/source/jormungandr/jormungandr/scenarios/journey_filter.py
@@ -399,7 +399,7 @@ def get_best_pt_journey_connections(journeys, request):
     if not journeys:
         return None
     best = get_ASAP_journey((j for j in journeys if 'non_pt' not in j.tags), request)
-    return get_nb_connections(best) if best else 0
+    return get_nb_connections(best) if best else None
 
 
 def get_nb_connections(journey):
--- a/source/jormungandr/jormungandr/scenarios/journey_filter.py
+++ b/source/jormungandr/jormungandr/scenarios/journey_filter.py
@@ -37,6 +37,7 @@ from jormungandr.scenarios.utils import compare, get_or_default
 from navitiacommon import response_pb2
 from jormungandr.utils import pb_del_if, ComposedFilter, portable_min
 from jormungandr.fallback_modes import FallbackModes
+from jormungandr.scenarios.qualifier import get_ASAP_journey
 
 
 def delete_journeys(responses, request):
@@ -390,15 +391,15 @@ class FilterTooLongDirectPath(SingleJourneyFilter):
         return max_duration > journey.duration
 
 
-def get_min_connections(journeys):
+def get_best_connections(journeys, request):
     """
     Returns min connection count among journeys
     Returns None if journeys empty
     """
     if not journeys:
         return None
-
-    return portable_min((get_nb_connections(j) for j in journeys if not to_be_deleted(j)), default=0)
+    best = get_ASAP_journey(journeys, request)
+    return get_nb_connections(best) if best else 0
 
 
 def get_nb_connections(journey):
@@ -720,7 +721,7 @@ def _filter_too_much_connections(journeys, instance, request):
     import itertools
 
     it1, it2 = itertools.tee(journeys, 2)
-    min_connections = get_min_connections(it1)
+    min_connections = get_best_connections(it1, request)
     is_debug = request.get('debug', False)
     if min_connections is not None:
         max_connections_allowed = max_additional_connections + min_connections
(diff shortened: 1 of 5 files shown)


source/jormungandr/jormungandr/scenarios/tests/new_default_tests.py
```diff
@@ -120,11 +120,11 @@
     # 10/15/2015 @ 12:08pm (UTC)
     (1444903680, 'best', (0, 8, 9)),
     # J16
-    # 10/15/2015 @ 12:08pm (UTC)
-    (1444903680, 'non_pt_bike', (0,)),
+    # 10/15/2015 @ 12:09pm (UTC)
```
xlqian: I have to modify the arrival time of these 2 journeys so that it won't become the best path

source/jormungandr/jormungandr/scenarios/qualifier.py
```diff
@@ -228,3 +228,8 @@ def nonTC_crit(j_1, j_2):
 
 def duration_crit(j_1, j_2):
     return compare_minus(j_1.duration, j_2.duration)
+
+
+def get_ASAP_journey(journeys, req):
```
pbougue, line 233: Not one more :pray: 
At least mutualize code in a function with the one used to tag 'best' (or reuse the tag if available?): https://github.com/CanalTP/navitia/blob/3cdfe8bf9156264cc04b93056b565df08fc52c29/source/jormungandr/jormungandr/scenarios/new_default.py#L749

We may probably use/adapt the most complete I know of so far: https://github.com/CanalTP/navitia/blob/3cdfe8bf9156264cc04b93056b565df08fc52c29/source/jormungandr/jormungandr/scenarios/journey_filter.py#L617-L667

source/jormungandr/jormungandr/scenarios/tests/new_default_tests.py
```diff
@@ -140,6 +140,7 @@ def build_mocked_response():
         arrival_time, jrny_type, sections_idx = jrny
         pb_j = response.journeys.add()
         pb_j.arrival_date_time = arrival_time
+        pb_j.nb_transfers = len(sections_idx) - 1
```
pbougue: Shouldn't you also remove number of street_network sections?

source/jormungandr/jormungandr/scenarios/tests/new_default_tests.py
```diff
@@ -509,3 +515,30 @@ def crowfly_in_ridesharing_test():
...
+    journey_filter.apply_final_journey_filters([mocked_pb_response], instance, mocked_request)
+    assert all('deleted_because_too_much_connections' not in j.tags for j in mocked_pb_response.journeys)
+
+    instance.max_additional_connections = 0
+    mocked_pb_response = build_mocked_response()
+    mocked_request = {'debug': True, 'datetime': 1444903200, 'clockwise': True}
+    journey_filter.apply_final_journey_filters([mocked_pb_response], instance, mocked_request)
+    assert sum('deleted_because_too_much_connections' not in j.tags for j in mocked_pb_response.journeys) == 8
+    # the best journey must be kept
+    assert mocked_pb_response.journeys[14].tags == [u'bike']
```
pbougue: clearer to test (same below):
```suggestion
    assert u'deleted_because_too_much_connections' not in mocked_pb_response.journeys[14].tags
```

source/jormungandr/jormungandr/scenarios/qualifier.py
xlqian: it's what i have done:https://github.com/CanalTP/navitia/pull/2977/files#diff-f050c03e2b24acf6b927fade7d249a87R750

pbougue (dismissed):

source/jormungandr/jormungandr/scenarios/journey_filter.py
```diff
@@ -390,15 +391,15 @@ def filter_func(self, journey):
...
     """
-    Returns min connection count among journeys
+    Returns the nb of connection of the best pt_journey
     Returns None if journeys empty
     """
     if not journeys:
         return None
-
-    return portable_min((get_nb_connections(j) for j in journeys if not to_be_deleted(j)), default=0)
+    best = get_ASAP_journey((j for j in journeys if 'non_pt' not in j.tags), request)
```
pbougue, line 401: Why not among all journeys?
If direct walking is the best, we should filter from that, no?

xlqian: it'll be problematic in distributed, because a direct path will 'always' appear in the response. with direct path by car/bike/taxi, that may shoot a lot of interesting pt journey

pbougue: OK.
:bulb: Idea if this filter becomes too permissive (later): we may consider the direct walk journey (if there is some) as a possible reference.

pbougue (approved):

source/jormungandr/jormungandr/scenarios/journey_filter.py
```diff
@@ -390,15 +391,15 @@ def filter_func(self, journey):
...
+    return get_nb_connections(best) if best else 0
```
nonifier:
```suggestion
    return get_nb_connections(best) if best else None
```
otherwise the following would not work isn't it ? `if best_pt_journey_connections is not None:`

source/jormungandr/jormungandr/scenarios/tests/new_default_tests.py
```diff
@@ -509,3 +516,30 @@ def crowfly_in_ridesharing_test():
     assert journey.durations.car == 0
     assert journey.distances.ridesharing == 43
     assert journey.distances.car == 0
+
+
+def filter_too_much_connections_test():
```
nonifier:
```suggestion
def filter_too_many_connections_test():
```
Because it's countable :)

source/jormungandr/jormungandr/scenarios/tests/new_default_tests.py
```diff
@@ -509,3 +516,30 @@ def crowfly_in_ridesharing_test():
...
+    # yes, it's a hack...
```
nonifier, line 522: no, it's a mock :rofl:

xlqian: it a mock with a hacky trick :)

source/jormungandr/jormungandr/scenarios/journey_filter.py
```diff
@@ -37,6 +37,7 @@
 from navitiacommon import response_pb2
 from jormungandr.utils import pb_del_if, ComposedFilter, portable_min
```
mbounabi, line 38:
```suggestion
from jormungandr.utils import pb_del_if, ComposedFilter
```
function now unused

source/jormungandr/jormungandr/scenarios/journey_filter.py
kinnou02: This is the kind of filter that will have to evolve when we will work to improve multimodal journeys

source/jormungandr/jormungandr/scenarios/qualifier.py
```diff
@@ -228,3 +228,8 @@ def nonTC_crit(j_1, j_2):
...
+    best_crit = arrival_crit if req["clockwise"] else departure_crit
```
mbounabi, line 234: Not sure if it has an impact on performance but why not pass only this parameter instead of the whole request object?

kinnou02 (approved):

source/jormungandr/jormungandr/scenarios/qualifier.py
```diff
@@ -228,3 +228,8 @@ def nonTC_crit(j_1, j_2):
...
+    return min_from_criteria(journeys, [best_crit, duration_crit, transfers_crit, nonTC_crit])
```
mbounabi, line 235: Just to clarify: the "best ASAP journey" is the one that arrives first, shortest, less nb of transfers, less fallback?

nonifier (approved):

source/jormungandr/jormungandr/scenarios/qualifier.py
xlqian: yes it is

source/jormungandr/jormungandr/scenarios/qualifier.py
xlqian: it won't change anything in term of performance, since the request is object and passed by reference :)

source/jormungandr/jormungandr/scenarios/journey_filter.py
xlqian: it's still used in `get_min_waiting`

mbounabi (approved):

PR #2978, opened 2019-10-02
ed: do not import stops from pathways

kinnou02:
For now we do not want to import station entrance, generic nodes and
boarding position as we can't do anything with them currently.
This PR refacto the StopHandler from the gtfs parser and fusio parser to
ignore them.
FusioParser had to evolve since the `location_type` aren't matching
between gtfs and ntfs


kinnou02:
poke @patochectp

source/ed/connectors/gtfs_parser.cpp
```diff
@@ -388,85 +388,113 @@ bool StopsGtfsHandler::parse_common_data(const csv_row& row, T* stop) {
...
+    if (has_col(timezone_c, row) && !row[timezone_c].empty()) {
+        gtfs_data.tz.stop_point_tz[sp] = row[timezone_c];
+    }
+    return sp;
+}
+
+bool StopsGtfsHandler::check_duplicate(const csv_row& row) {
     // In GTFS the file contains the stop_area and the stop_point
     // We test if it's a duplicate
     if (gtfs_data.stop_map.find(row[id_c]) != gtfs_data.stop_map.end()
```
pbougue: minor: Can you rename (it was there before but doesn't help understanding)?
```suggestion
    if (gtfs_data.stop_point_map.find(row[id_c]) != gtfs_data.stop_point_map.end()
```

source/ed/connectors/gtfs_parser.cpp
```diff
@@ -388,85 +388,113 @@ bool StopsGtfsHandler::parse_common_data(const csv_row& row, T* stop) {
...
         || gtfs_data.stop_area_map.find(row[id_c]) != gtfs_data.stop_area_map.end()) {
         LOG4CPLUS_WARN(logger, "The stop " + row[id_c] + " has been ignored");
```
pbougue, line 465: Naïve question (doesn't change the way it works anyway): So just ignoring the last read... No priority given to SP or SA?

source/ed/connectors/gtfs_parser.cpp
```diff
@@ -388,85 +388,113 @@ bool StopsGtfsHandler::parse_common_data(const csv_row& row, T* stop) {
...
+    }
+
+    // we save the tz in case the stop point is later promoted to stop area
+    if (has_col(timezone_c, row) && !row[timezone_c].empty()) {
+        gtfs_data.tz.stop_point_tz[sp] = row[timezone_c];
+    }
+    return sp;
+}
+
+bool StopsGtfsHandler::check_duplicate(const csv_row& row) {
```
pbougue: minor clearer to me:
```suggestion
bool StopsGtfsHandler::is_duplicate(const csv_row& row) {
```

pbougue (commented):
OK, in NTFS, `location_type` 3, 4, 5 are matching GTFS 2, 3, 4 :+1:

Fine for me, I let you decide but I can approve it (mergi-fear :stuck_out_tongue: )

source/ed/connectors/gtfs_parser.cpp
kinnou02: duplicate are evil!
we shouldnt' even allow them, something will obviously not works, `stop_time` reference `stop_point` but a lot of thing in the ntfs references `stop_area` like `route`'s destination.

pbougue (approved):

source/ed/connectors/fusio_parser.cpp
```diff
@@ -230,57 +232,97 @@ void StopsFusioHandler::handle_stop_point_without_area(Data& data) {
     LOG4CPLUS_INFO(logger, "Deletion of " << erase_sp.size() << " stop_point wihtout stop_area");
 }
 
-StopsGtfsHandler::stop_point_and_area StopsFusioHandler::handle_line(Data& data,
-                                                                     const csv_row& row,
-                                                                     bool is_first_line) {
-    auto return_wrapper = StopsGtfsHandler::handle_line(data, row, is_first_line);
+nm::StopArea* StopsFusioHandler::build_stop_area(Data& data, const csv_row& row) {
+    auto* sa = StopsGtfsHandler::build_stop_area(data, row);
```
xlqian, line 236: :+1:

source/ed/connectors/gtfs_parser.cpp
```diff
@@ -388,85 +388,113 @@ bool StopsGtfsHandler::parse_common_data(const csv_row& row, T* stop) {
     return true;
 }
 
-StopsGtfsHandler::stop_point_and_area StopsGtfsHandler::handle_line(Data& data, const csv_row& row, bool) {
+nm::StopArea* StopsGtfsHandler::build_stop_area(Data& data, const csv_row& row) {
+    nm::StopArea* sa = new nm::StopArea();
+    if (!parse_common_data(row, sa)) {
+        delete sa;  // don't forget to free the data
```
xlqian, line 394: can you use std::unique_ptr and return with std::unique_ptr::release()?

source/ed/connectors/gtfs_parser.cpp
```diff
@@ -388,85 +388,113 @@ bool StopsGtfsHandler::parse_common_data(const csv_row& row, T* stop) {
...
+        sa->time_zone_with_name = {data.tz_wrapper.tz_name, data.tz_wrapper.boost_timezone};
+    }
+    if (has_col(desc_c, row)) {
+        add_gtfs_comment(gtfs_data, data, sa, row[desc_c]);
+    }
+    return sa;
+}
+
+nm::StopPoint* StopsGtfsHandler::build_stop_point(Data& data, const csv_row& row) {
+    nm::StopPoint* sp = new nm::StopPoint();
```
xlqian, line 419: same, can you use std::unique_ptr?

source/ed/connectors/gtfs_parser.cpp
xlqian: ah, that's old code... take the suggestion if it seems ok to you

source/ed/connectors/fusio_parser.cpp
```diff
@@ -230,57 +232,97 @@ void StopsFusioHandler::handle_stop_point_without_area(Data& data) {
...
+
+    if (is_valid(ext_code_c, row)) {
+        data.add_object_code(sp, row[ext_code_c]);
     }
-    if (return_wrapper.first != nullptr && is_valid(geometry_id_c, row)) {
+    if (is_valid(property_id_c, row)) {
+        auto it_property = gtfs_data.hasProperties_map.find(row[property_id_c]);
+        if (it_property != gtfs_data.hasProperties_map.end()) {
+            sp->set_properties(it_property->second.properties());
+        }
```
nonifier, line 277: If you want less boiler plate you could do : 
```cpp
        utils::make_map_find(gtfs_data.hasProperties_map, row[property_id_c])
            .if_found([&](const type::hasProperties & p){ 
                sp->set_properties(p.properties());
            });
```

source/ed/connectors/gtfs_parser.cpp
kinnou02: I think it should be done in another PR, or even better, use smart ptr everywhere

source/ed/connectors/fusio_parser.cpp
```diff
@@ -230,57 +232,97 @@ void StopsFusioHandler::handle_stop_point_without_area(Data& data) {
...
+    }
+
+    if (is_valid(comment_id_c, row)) {
+        auto it_comment = data.comment_by_id.find(row[comment_id_c]);
+        if (it_comment != data.comment_by_id.end()) {
+            data.add_pt_object_comment(sp, row[comment_id_c]);
+        }
+    }
+
+    if (has_col(type_c, row) && row[type_c] == "2") {
```
nonifier, line 287: There is a whole bunch of "1", "2" etc..., living around as magic numbers. 
It would be nice to have them defined as proper constant named values.

source/ed/connectors/gtfs_parser.cpp
```diff
@@ -388,85 +388,113 @@ bool StopsGtfsHandler::parse_common_data(const csv_row& row, T* stop) {
     return true;
 }
 
-StopsGtfsHandler::stop_point_and_area StopsGtfsHandler::handle_line(Data& data, const csv_row& row, bool) {
+nm::StopArea* StopsGtfsHandler::build_stop_area(Data& data, const csv_row& row) {
```
nonifier, line 391: Returning a `unique_ptr<nm::StopArea>` here would be much better ! That would prevent you from leaking like an old  person....  
Same applies for `StopsGtfsHandler::build_stop_point()`

source/ed/connectors/fusio_parser.cpp
pbougue: As they don't mean the same thing if it's GTFS or NTFS, it may not be clearer :stuck_out_tongue_closed_eyes:

source/ed/connectors/gtfs_parser.cpp
kinnou02: returning a `unique_ptr` require adding 3 or 4 `get()` in this method to store the raw pointer where we need it, then a call to `release()` later. If the pointer was to be released we will have invalid pointer stored everywhere.
Using `shared_ptr` in `Data` and `GtfsData` would solve this problem but it's outside the scope of this PR.

xlqian (approved):

benoit-bst:
Can I have an Utest to show the change, if it not time-consuming ?
> reply from @pbougue:
> There is a little add in docker tests

If there is already a test i'm okay for sure :angel: 
But where is it ? Or if it doesn't exist because it is too complicate to implement, due to ntfs Data set, i can understand :smile:

source/ed/connectors/fusio_parser.cpp
nonifier: because GTFS and NTFS could be different, naming thing properly can only make things clearer ;)

source/ed/connectors/gtfs_parser.cpp
nonifier: np

nonifier (approved):

PR #2979, opened 2019-10-02
Kraken: POC for smarter directions in stop_schedules

pbougue:
### Short description

Stop grouping departure times by {route, stop_point} as there can be too much or too few (dumb) routes in data.

Instead, group them observing all the stations being served after considered stop_point:  
If after the same stop_point, a vehicle serves the same stations in the same order (maybe serving more on the way and after), then they are grouped.

JIRA: https://jira.kisio.org/browse/NAVP-1086

----

#### Perfs
:heavy_check_mark: No issue about that so far: on "big" requests, request duration is shorter (less protobuf to fill and serialize)

#### RER B at Gare du Nord
Before: BeforeRerB
After: rerBapres
The terminus duplicated are because starting stop_points differs (surface vs. RER)

#### RER C at Saint-Michel Notre dame
Before: rerCavant
After: rerCapres
The terminus duplicated (Dourdan) is because some vehicles are serving between Ivry and Athis, some other serving between Savigny and St-Michel-sur-Orge, but none is serving both, so they stay separated (even if geographically, they are on the same path).

#### Metro 13 at Montparnasse
Before: metro13avant
After: metro13apres
Route toward North is deduplicated as no vehicle is serving both branches (geographically separated)

#### Metro 13 at Asnières-Gennevilliers
Before: metro13terminusAvant
After: metro13terminusApres
No vehicle is **ever** departing from station direction North, so the direction does not appear.

#### SNCF at Montparnasse
Before: sncfMontparAvant
After: sncfMontparApres
It does improve a bit (mostly removes 52 terminus empty cases, win 7 in merge/expand process). Not so much improvement as stop points are different between ouigo, inoui and "classic" TGV so omnibus lines are not merging much.

### DIFF
--- a/source/jormungandr/jormungandr/interfaces/v1/serializer/pt.py
+++ b/source/jormungandr/jormungandr/interfaces/v1/serializer/pt.py
@@ -618,6 +618,7 @@ class RouteDisplayInformationSerializer(PbNestedSerializer):
     name = jsonschema.Field(schema_type=str)
     links = DisruptionLinkSerializer(attr='impact_uris', display_none=True)
     text_color = jsonschema.Field(schema_type=str)
+    description = jsonschema.Field(schema_type=str)
 
 
 class VJDisplayInformationSerializer(RouteDisplayInformationSerializer):
--- a/source/routing/raptor_utils.h
+++ b/source/routing/raptor_utils.h
@@ -40,6 +40,7 @@ namespace type {
 struct StopPoint;
 struct StopTime;
 struct Route;
+struct Line;
 struct VehicleJourney;
 struct MetaVehicleJourney;
 struct PhysicalMode;
@@ -53,6 +54,7 @@ using JppIdx = Idx<JourneyPatternPoint>;
 using JpIdx = Idx<JourneyPattern>;
 using SpIdx = Idx<type::StopPoint>;
 using RouteIdx = Idx<type::Route>;
+using LineIdx = Idx<type::Line>;
 using VjIdx = Idx<type::VehicleJourney>;
 using MvjIdx = Idx<type::MetaVehicleJourney>;
 using PhyModeIdx = Idx<type::PhysicalMode>;
--- a/source/time_tables/departure_boards.cpp
+++ b/source/time_tables/departure_boards.cpp
@@ -65,9 +65,9 @@ static bool is_terminus_for_all_stop_times(const std::vector<routing::datetime_s
 }
 
 static void render(PbCreator& pb_creator,
-                   const std::map<RoutePointIdx, pbnavitia::ResponseStatus>& response_status,
-                   const std::map<RoutePointIdx, vector_dt_st>& map_route_stop_point,
-                   const std::map<RoutePointIdx, first_and_last_stop_time>& map_route_point_first_last_st,
+                   const std::map<routing::JppIdx, pbnavitia::ResponseStatus>& response_status,
+                   const std::map<routing::JppIdx, vector_dt_st>& map_route_stop_point,
+                   const std::map<routing::JppIdx, first_and_last_stop_time>& map_route_point_first_last_st,
                    const DateTime datetime,
                    const DateTime max_datetime,
                    const boost::optional<const std::string> calendar_id,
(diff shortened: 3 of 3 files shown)


pbougue:
### TODO
- [x] Check what changed in unit-tests
  > :white_check_mark: loosing terminus routes
  > :white_check_mark: winning direction for Y-shaped route
  > :white_check_mark: loosing merged routes
  > :bulb: Some checks on order of stop schedule are changing (line>route>SP) - we should probably keep that order > noted in "remember" section.
- [x] Check RER D case at Gare de Lyon
  :heavy_check_mark: No vehicle passes by Gare de Lyon then Grand Bourg in that order
  :heavy_check_mark: Case of RER D at Juvisy does split into 2 directions to Melun (one for Grand Bourg, one for Evry-Courcouronnes)
- [x] Check behavior for SNCF coverage
  > see main comment.

### Choices
Choices already done:
1. Separate directions by stop_points (to avoid merging vehicles departing from different stops).
2. Separate directions by lines (to avoid merging vehicles from different lines).
3. For a route serving in that order A>B>C>D, if there are only vehicles doing A>B>D and others doing A>C>D, there will be separate directions ; as long as there is none doing A>B>C>D. See RER C result (and think if RER D case between Viry-Châtillon and Corbeille-Essonnes).

### Ideas
Ideas to be tested or implemented:
1. Prioritize gathering by route if possible (case of short JP that are included in almost any other JP).
2. Gather JP considering stop_areas served after (instead of stop_points), consider equivalent serving the same stop_area multiple times in a row (corner case).

### Questions
In case we want to go on with that:
1. Where should we use those notions of "omnibus-directions": departures, arrivals, route_schedules? So where should we put the code (tartare, ed, utils, just in one endpoint's manager)
2. Find a correct name for that :stuck_out_tongue_closed_eyes:
    Since a `JourneyPatternPoint` is most of the time used to represent **the final part from a JourneyPattern** (starting from StopPoint of the JPP).
3. How to present that for traveler? Use last stop_area of "omnibus-direction" as direction? How to explain the difference on the case of RER C to Dourdan (present "via" but think of SNCF's case as it can be a combination)?
4. Should we use those "omnibus-directions" everywhere direction is exposed?
5. Should we do the same kind of split/group for line/route thermometers? route-schedules? Anything else?
6. Should we expose those objects (to allow disruptions to be added to them, or other grouping/filtering in PTRef)?
7. What about the risk of clients willing to split one line per partial-terminus, one per omnibus-direction, one per direction (forward-backward)... Exposing/using this endpoint that is no more data-driven exposes to the impossibility to do that, or it would cost re-developping front to use another endpoint.
8. How about a parameter to activate it on stop_schedules ?
9. How about a parameter to activate stop_area merging (to avoid showing multiple SP-direction couples) ? :warning: As it is wrong for a front to just group SP-direction couples based on their SA and the final SA of omnibus-direction (which is the current name/identifier of omnibus-direction).
10. Should we improve the current stop_schedules for the terminus-case (don't show them, make sure they are marked as terminus)?
11. How to name the omnibus-direction for the user (as end-point is sometimes not enough) ?
12. Should we add notes on "partial" circulations' stop-times also ?

### Cleanup
Cleanup if we keep a solution isolated in a separate endpoint:
- [ ] Rename variables (minimum changes were done)
- [ ] Review in details what is done around the code changed and adapt
- [ ] Change collection used and refactor a bit direction build
- [ ] Remove unused code (get_jpp_from_route_point at least)

### To remember
- [ ] Think and adapt :electric_plug: for external realtime proxies (most are using routes)
- [ ] Keep response ordered by line, then route, then SP

pbougue:
I close this PR as it won't be merged as-is and needs more thinking on how to deploy/use this.
Also the main goal was to test and check what's doable, and it's done now, plus the branch will stay available for tests !

The discussion and the questions raised can still be done here if necessary.

pbougue:
### Grouping of JourneyPatterns

The goal of this PR is to group JourneyPatterns (JP). More precisely, the end of JP strictly after concerned stop, which is contained in the JourneyPatternPoint (JPP).

This grouping is done so that all JP that are "contained" in another JP are grouped under that "omnibus-JP" (the most complete one). That omnibus-JP becomes the "direction" in that work.

A JP that serves sames stops, in same order (but maybe not all of them) is "contained" in another JP.

#### Rough algorithm

```
"Real world" geography:   ...===X===A===B===C===D

JP1                       ...---X---A---B---C
JP2                       ...---X-------B---C
JP3                       ...---X---A---B-------D
JP4                       ...---X---A---B---C---D
```
So at stop X, considering the list JP to group: [ [A,B,C], [B,C], [A,B,D], [A,B,C,D] ]
1. Initialize result= []
2. Consider [A,B,C]: not contained in any direction of result, so create a new direction.  
    * result= [ {direction=[A,B,C], contains=[[A,B,C]]} ]
3. Consider [B,C]: contained in [A,B,C], so attach it.  
    * result= [ {d=[A,B,C], c=[[A,B,C], [B,C]]} ]
4. Consider [A,B,D]: not contained in direction (only [A,B,C] so far), and no existing direction is contained in [A,B,D], so create a new direction.  
    * result= [ {d=[A,B,C], c=[[A,B,C], [B,C]]}, {d=[A,B,D], c=[[A,B,D]]} ]
5. Consider [A,B,C,D]: not contained in any existing direction, but [A,B,C] and [A,B,D] are contains inside [A,B,C,D], so [A,B,C,D] becomes the new direction of [A,B,C], and [A,B,D] is also transfered under that direction.  
    * result= [ {d=[A,B,C,D], c=[[A,B,C], [B,C], [A,B,D], [A,B,C,D]]} ]

So we end-up with only one direction: [A,B,C,D].

##### Discussion
If no trip was serving [A,B,C,D] then we would not have grouped the 2 directions.
Because the real world geography would more likely be:

```
"Real world" geography:   ...===X===A===B===C
                                        \===D

JP1                       ...---X---A---B---C
JP2                       ...---X-------B---C
JP3                       ...---X---A---B
                                        \---D
```
And in that case consider the following trip:
```
JP5                       ...---X---A---B
```
JP5 will be attached to one of the 2 directions **randomly** (considering routes would be better).

PR #2980, opened 2019-10-03
[Eitri] Fix cities file type

mbounabi:
The cities binary expects the path to the cities file as a `str`.
Eitri retrieves this info as a `FileType` so it fails when using the argument `-i CITIES_FILE, --cities-file CITIES_FILE`
-> `TypeError: sequence item 1: expected string, FileType found`

Eitri will now retrieve the cities file path as a `str` and pass it directly to the cities binary

### DIFF
--- a/source/eitri/eitri.py
+++ b/source/eitri/eitri.py
@@ -124,10 +124,7 @@ def get_params():
         help="the output file path (default: './data.nav.lz4')",
     )
     parser.add_argument(
-        '-i',
-        '--cities-file',
-        type=argparse.FileType,
-        help='the path to the "cities" file to load  (usually a *.osm.pbf)',
+        '-i', '--cities-file', type=str, help='the path to the "cities" file to load  (usually a *.osm.pbf)'
     )
     parser.add_argument(
         '-f', '--cities-dir', type=is_directory, help='the path to the directory containing the "cities" binary'


xlqian (approved):

kinnou02 (approved):

kinnou02:
Jenkins, test this please.

source/eitri/eitri.py
```diff
@@ -124,10 +124,7 @@ def get_params():
         help="the output file path (default: './data.nav.lz4')",
     )
     parser.add_argument(
-        '-i',
-        '--cities-file',
-        type=argparse.FileType,
-        help='the path to the "cities" file to load  (usually a *.osm.pbf)',
+        '-i', '--cities-file', type=str, help='the path to the "cities" file to load  (usually a *.osm.pbf)'
```
pbougue, line 127: Maybe as above:
```suggestion
        '-i', '--cities-file', type=is_valid_file, help='the path to the "cities" file to load  (usually a *.osm.pbf)'
```

mbounabi: :negative_squared_cross_mark: It won't work as the file exists (of course) and:
https://github.com/CanalTP/navitia/blob/3cdfe8bf9156264cc04b93056b565df08fc52c29/source/eitri/eitri.py#L71

pbougue: OK, we may have the same kind of test for existing files, but it's already overkill I think :+1:

pbougue (approved):

nonifier (approved):

benoit-bst (approved):

PR #2981, opened 2019-10-03
Ed: Import entrances from gtfs and ntfs

kinnou02:
This doesn't do anything :)
It only read stops that are entrances or exits from `stops.txt` and
associate it to it's station (`StopArea` in our case).
This doesn't read others informations from gtfs pathway.


kinnou02:
next step would require a lot of refactoring, currently a stop_point is expected to be projected on only one point, once we have entrances a stop_point would be projected on multiple points as each entrance is the stop_point from the streetnetwork point of view :(

source/ed/connectors/gtfs_parser.cpp
```diff
@@ -319,6 +319,13 @@ void StopsGtfsHandler::finish(Data& data) {
         }
     }
 
+    // associate entrances
+    for (auto* sa : data.stop_areas) {
+        for (const auto* entrance : entrances_map[sa->uri]) {
```
xlqian, line 324: ( `[]` operator will create a new entry if the key does not exist.. )

nonifier:
This was just a POC to get an overview of the work needed to get the Entrance/Exit of a stop point integrated into the planner.

Some doc can be found here - https://confluence.kisio.org/display/NAV/Pathway+in+navitia

PR #2982, opened 2019-10-03
Jormun: Remove excess tickets and ticket links inside response

benoit-bst:
**JIRA**: https://jira.kisio.org/browse/NAVP-974
Remove excess **tickets** or **ticket links**, when the journey is to be deleted. 
furthermore, remove dead fare links because, when tickets are cleaned, it can stay ticket links inside `journey.fare.links`. This is the reason why `check_internal_links(response, self.tester)` not passed. This function check all **dead** **links** inside response
Now, all tests pass with this check :tada: 

- Tested on real coverage and it works, there is no extra ticket/ticket links
- Artemis is green too (https://ci.navitia.io/job/custom_Artemis/8/)


benoit-bst:
- [x] I forgot to check an other field. The famous `journey.section.links` If the journey have to be deleted, i need to check if there are dead links for ticket inside `journey.section.links` too. Work in progress

Finally, we don't need to handle `dead ticket links` inside `journey.section.links`. 
Why? Because we handle journey.section.links [after](https://github.com/CanalTP/navitia/blob/dev/source/jormungandr/jormungandr/interfaces/v1/Journeys.py#L182). We suppress excess ticket [before](https://github.com/CanalTP/navitia/pull/2982/files#diff-f050c03e2b24acf6b927fade7d249a87R796), so the link is not created :smile: 

Fake alert

source/jormungandr/jormungandr/scenarios/journey_filter.py
```diff
@@ -728,3 +728,39 @@ def _filter_too_much_connections(journeys, instance, request):
...
+            if fare_t_id not in ticket_id_list:
+                logger.debug(
+                    'remove excess fare ticket link %s into journey %s to be deleted',
+                    fare_t_id,
+                    journey.internal_id,
+                )
+                journey.fare.ticket_id.remove(fare_t_id)
+
+    for j in response.journeys:
+        if to_be_deleted(j):
```
pbougue: To me cleaning should be done as part of the `delete_journeys()` to always have a valid response, as:
* If we let it like this, the response will not be valid (on a ticket point of view) when `debug=true`.
* And we have tests on that (as they broke on the `fare` part): They do not currently test that tickets are present when they should (if one journey has tickets, the others should too, in sections and in fare).
* There would be no need to clean journey parts as the journeys themselves will be deleted.

source/jormungandr/jormungandr/scenarios/journey_filter.py
```diff
@@ -728,3 +728,39 @@ def _filter_too_much_connections(journeys, instance, request):
...
+                            'remove excess section id %s inside %s ticket after journey filtering', s_id, t.id
+                        )
+                        t.section_id.remove(s_id)
+                    else:
+                        logger.debug('remove excess ticket %s after journey filtering', t.id)
+                        tickets.remove(t)
+
+    def _remove_excess_fare_ticket_links(ticket_id_list, journey):
+        for fare_t_id in journey.fare.ticket_id:
+            if fare_t_id not in ticket_id_list:
```
pbougue: When `debug=true` sometimes we will have ticket links and sometimes not in `fare` with that, which is weird.
> (that comment will be resolved if clean is done as part of delete_journeys() )

source/jormungandr/jormungandr/scenarios/new_default.py
```diff
@@ -793,6 +793,8 @@ def merge_responses(responses, debug):
                 continue
             merged_response.impacts.extend([i])
 
+    journey_filter.remove_excess_tickets_or_ticket_links_with_journeys_to_be_deleted(merged_response)
```
kinnou02: Why not filtering at [line 772](https://github.com/CanalTP/navitia/blob/9a1d6268cab30668ff1bcccdbc20b631f2d5aedd/source/jormungandr/jormungandr/scenarios/new_default.py#L772)?
```
tickets_to_add = set(t for j in r.journeys for t in j.fare.ticket_id)
```
only adding ticket from non deleted journeys should be enough no?

benoit-bst: Good point, I can handle it to probably safe process loop call.

source/jormungandr/jormungandr/scenarios/journey_filter.py
benoit-bst: I focused my research inside **merge_journeys** function, as it is mentionned within the JIRA ticket. You have probably right about the mecanism and the delete_journeys function. I will strode inside **delete_journey** to find a way.

source/jormungandr/jormungandr/scenarios/journey_filter.py
benoit-bst: I was just because dead_link is forbidden. But this part likely change with your first remark.

kadhikari (commented):
-  Since the last use of 'to_be_deleted' is in the function cullings_journeys, you can call your functions only after this line: https://github.com/CanalTP/navitia/blob/b2918a39794041dd19146ba2d396edb169e2965a/source/jormungandr/jormungandr/scenarios/new_default.py#L1042

- As mentioned by @kinnou02, you can also filter some unwanted fare(but not all) in this line: https://github.com/CanalTP/navitia/blob/9a1d6268cab30668ff1bcccdbc20b631f2d5aedd/source/jormungandr/jormungandr/scenarios/new_default.py#L772

- Do not clean tickets for 'debug=True'

benoit-bst:
**To sum up**,

After investigation, it turns out the @kinnou02 's comment resolve the problem. Only add a condition taking into account `to_be_deleted and debug state` allows to manage the issue.
Nevertheless, I discovered a **corner case**, due to the **culling_journeys** function. Effectively, there is 2 spots where journeys can be deleted (into **merge_response** and **culling_journeys**) and tickets can rest. The **merge_responses** case will be manage with @kinnou02 's patch, but it remains **culling_journeys**. If we use an option like **max_nb_journeys**, culling_journeys removes extra journeys but forget to suppress associated tickets... So we have to suppress extra tickets for this peculiar case too.

The **todo list** become:

- [x] Add the @kinnou02's patch with taking into account of **debug state**
- [x] Suppress extra tickets after **culling_jouneys** function (**corner case**)
       Don't forget to take into account `ridesharing_journeys.fare` in the id list
- [x] Add **integration tests** that highlight all cases (the longest...)
     Sub tasks :
     - [x] Adapt builder to add new ticket with easeness
     - [x] Add new ticket inside mock kraken test
     - [x] Create a complete new integration tests bundle for all cases
            - one **classical case**
            - a **max_nb_journeys case** to test with `culling_journeys`
            - a **debug flag case**
- [x] Retest on real coverage, there is no extra ticket/ticket links

source/jormungandr/tests/journey_common_tests.py
```diff
@@ -1218,18 +1218,6 @@ def test_journeys_too_short_heavy_mode_fallback_filter(self):
 
         assert 'deleted_because_too_short_heavy_mode_fallback' in car_fallback_pt_journey['tags']
 
-    def test_journey_ticket(self):
```
benoit-bst, line 1221: This test is redondant now with [that](https://github.com/CanalTP/navitia/pull/2982/files#diff-0832703183fb3233e60448ac96ec726bR2286). The new tests are more exhaustive, too

source/jormungandr/tests/tests_mechanism.py
```diff
@@ -375,8 +375,7 @@ def is_valid_journey_response(self, response, query_str, check_journey_links=Tru
             is_valid_ticket(f, self.tester)
             all_tickets[f['id']] = f
 
-        # TODO: Implement NAVP-974 to handle tickets-sections internal links
-        # check_internal_links(response, self.tester)
+        check_internal_links(response, self.tester)
```
benoit-bst, line 378: Internal links are now check

source/jormungandr/jormungandr/scenarios/journey_filter.py
```diff
@@ -728,3 +728,23 @@ def _filter_too_much_connections(journeys, instance, request):
             if get_nb_connections(j) > max_connections_allowed:
                 logger.debug("the journey {} has a too much connections, we delete it".format(j.internal_id))
                 mark_as_dead(j, is_debug, "too_much_connections")
+
+
+def remove_excess_tickets_after_culling(response):
```
pbougue: minor:
```suggestion
def remove_excess_tickets(response):
```
As culling is not especially linked to the function itself (clean comments and log as well if you do it).

source/jormungandr/jormungandr/scenarios/journey_filter.py
```diff
@@ -728,3 +728,23 @@ def _filter_too_much_connections(journeys, instance, request):
...
+    """
+    Remove excess tickets after culling journeys process
+    """
+    logger = logging.getLogger(__name__)
+
+    fare_ticket_id_list = []
```
pbougue: minor: you can use a set here.

source/jormungandr/jormungandr/scenarios/journey_filter.py
```diff
@@ -728,3 +728,23 @@ def _filter_too_much_connections(journeys, instance, request):
...
+    for j in response.journeys:
+        fare_ticket_id_list.extend([t_id for t_id in j.fare.ticket_id])
+        # ridesharing case
+        fare_ticket_id_list.extend(
```
pbougue: minor: you may adapt the blind visitor from link-checker (adding just tickets' links) to avoid specific check cases and if we ever add tickets somewhere else in the future https://github.com/CanalTP/navitia/blob/d0c27238dea3d179379a507639f741b81ad397f9/source/jormungandr/tests/check_utils.py#L339-L345

source/ed/build_helper.cpp
```diff
@@ -602,6 +602,27 @@ void builder::connection(const std::string& name1, const std::string& name2, flo
     connexion->destination->stop_point_connection_list.push_back(connexion);
 }
 
+void builder::add_ticket(const std::string& ticket_id,
```
pbougue, line 605: :+1:

source/jormungandr/tests/journey_common_tests.py
```diff
@@ -2291,3 +2279,130 @@ def test_direct_path_mode_bike_and_direct_path_mode_none(self):
         assert len(response["journeys"]) == 1
         assert "walking" in response["journeys"][0]["tags"]
         assert "non_pt" not in response["journeys"][0]["tags"]
+
+
+@dataset({"min_nb_journeys_test": {}})
+class JourneysTickets:
+    def test_journey_tickets(self):
+        """
+        2 solutions are founded with 1 ticket for each journey
```
pbougue:
```suggestion
        2 solutions found with 1 ticket for each journey
```

source/jormungandr/tests/journey_common_tests.py
```diff
@@ -2291,3 +2279,130 @@ def test_direct_path_mode_bike_and_direct_path_mode_none(self):
...
+        assert response['tickets'][1]['source_id'] == 'B-Ticket'
+        assert response['tickets'][1]['comment'] == 'B-Ticket comment'
+
+        # Links between journeys.fare and journeys.tickets
+        assert response['tickets'][0]['id'] == response['journeys'][1]['fare']['links'][0]['id']
+        assert response['tickets'][1]['id'] == response['journeys'][0]['fare']['links'][0]['id']
+
+    def test_journey_tickets_with_max_nb_journeys(self):
+        """
+        2 solutions found, but with max_nb_journeys option, it remains 1 journey.
```
pbougue:
```suggestion
        2 solutions found, but with max_nb_journeys option, only 1 journey remains.
```

source/jormungandr/tests/journey_common_tests.py
```diff
@@ -2291,3 +2279,130 @@ def test_direct_path_mode_bike_and_direct_path_mode_none(self):
...
+        Check there is only one ticket.
+        The max_nb_journeys option force to filter one journey and the associated ticket too
```
pbougue:
```suggestion
        The max_nb_journeys option forces filtering of one journey and the associated ticket too
```

pbougue (commented):
Only minor / :gb: remarks, it's OK for me.
I let you decide if you want to change some of this, feel free to resolve comments otherwise and I'll approve.

source/jormungandr/jormungandr/scenarios/journey_filter.py
benoit-bst: I'm lazy... :sleeping:

source/jormungandr/jormungandr/scenarios/journey_filter.py
benoit-bst: much better

kadhikari (approved):

pbougue (approved):

PR #2983, opened 2019-10-04
Draft isochrone distributed

AurelienLP:
This does not compile at all, it just gives a rough idea of what I want to do for this implementation!

In distributed mode, the call to isochrone (/journeys with one from or to) is redirected to new_default.

In distributed mode, we want to call a `street_network_matrix` AND call `raptor.isochrone` with that.
The problem is that in `make_isochrone` in kraken, `add_isochrone_response` take `origin` as the `from` or `to` of the request. 

We don't send this parameter in distributed mode since we are supposed to send the result of the street_network_routing_matrix (like we do when calling pt_planner).

I have two choice :

1) Implement a new `add_isochrone_response` and do more processing in Jormun to recreate the full reponse with the result of sn_matrix and the result of raptor.isochrone
2) Send this origin (`from` or `to` of request) when calling isochrone in distributed mode and don't do anything more

We'll have the same issue for `graphical_isochrone` by the way.
Please, someone, help me.

### DIFF
--- a/source/jormungandr/jormungandr/scenarios/distributed.py
+++ b/source/jormungandr/jormungandr/scenarios/distributed.py
@@ -273,6 +273,56 @@ class Distributed(object):
             journeys=journeys_to_complete,
         )
 
+    def _isochrone(future_manager, request, instance, krakens_call, context):
+            logger = logging.getLogger(__name__)
+            logger.debug('request datetime: %s', request['datetime'])
+
+            requested_modes = { mode for mode, _, _ in krakens_call }
+
+            # context.direct_paths_by_mode = context.streetnetwork_path_pool.get_all_direct_paths()
+            context.direct_paths_by_mode = None
+
+            context.orig_proximities_by_crowfly = ProximitiesByCrowflyPool(
+                future_manager=future_manager,
+                instance=instance,
+                requested_place_obj=context.requested_orig_obj,
+                modes=requested_modes,
+                request=request,
+                direct_paths_by_mode=context.direct_paths_by_mode,
+                max_nb_crowfly_by_mode=request['max_nb_crowfly_by_mode'],
+            )
+
+            context.orig_places_free_access = PlacesFreeAccess(
+                future_manager=future_manager, instance=instance, requested_place_obj=context.requested_orig_obj
+            )
+
+            context.orig_fallback_durations_pool = FallbackDurationsPool(
+                future_manager=future_manager,
+                instance=instance,
+                requested_place_obj=context.requested_orig_obj,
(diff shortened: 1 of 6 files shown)


source/routing/raptor_api.cpp
```diff
@@ -1452,6 +1449,45 @@ void make_isochrone(navitia::PbCreator& pb_creator,
...
+                                const type::RTLevel rt_level,
+                                int max_duration = 3600,
+                                uint32_t max_transfers = std::numeric_limits<uint32_t>::max()) {
+    // Create datetime
+    auto datetimes = parse_datetimes(raptor, {datetime_timestamp}, pb_creator, clockwise);
+    if (pb_creator.has_error() || pb_creator.has_response_type(pbnavitia::DATE_OUT_OF_BOUNDS)) {
+        return;
+    }
+
+    // Get stop points for departure and destination
```
AurelienLP: This part is from `make_isochrone`

source/routing/raptor_api.cpp
```diff
@@ -1452,6 +1449,45 @@ void make_isochrone(navitia::PbCreator& pb_creator,
...
+    raptor.isochrone(departures, init_dt, bound, max_transfers, accessibilite_params, forbidden, allowed, clockwise,
+                     rt_level);
+
+    add_isochrone_response(raptor, isochrone_origin, pb_creator, raptor.data.pt_data->stop_points, clockwise, init_dt,
+                           bound, max_duration);
+    pb_creator.sort_journeys();
+    if (pb_creator.empty_journeys()) {
+        pb_creator.fill_pb_error(pbnavitia::Error::no_solution, pbnavitia::NO_SOLUTION,
+                                 "no solution found for this isochrone");
+    }
```
AurelienLP, line 1488: This part is from `make_isochrone`

source/jormungandr/jormungandr/scenarios/distributed.py
```diff
@@ -273,6 +273,56 @@ def finalise_journeys(self, future_manager, request, responses, context, instanc
...
+                future_manager=future_manager,
+                instance=instance,
+                requested_orig_obj=context.requested_orig_obj,
+                requested_dest_obj=None,
+                streetnetwork_path_pool=context.streetnetwork_path_pool,
+                krakens_call=krakens_call,
+                orig_fallback_durations_pool=context.orig_fallback_durations_pool,
+                dest_fallback_durations_pool=None,
+                request=request,
+            )
```
AurelienLP, line 322: Like `_compute_all` without direct_paths and destination (only origin)

PR #2986, opened 2019-10-04
Kraken: Strong-type Rank<> for StopTime and JourneyPatternPoint

pbougue:
After https://github.com/CanalTP/navitia/pull/2959#discussion_r331155584

For safety and readability, add strong-type for ranks.
:warning: Did not strong-type cache structure `dataRAPTOR::JppsFromSp::Jpp`, though.

:mag: Review by commit+message advised.

TODO (`do_not_merge`):
- [x] Merge #2959 then rebase
- [x] Merge https://github.com/CanalTP/utils/pull/83 then update submodule
- [x] Explore if it is easy and perf-friendly to do the same for JourneyPatternPoint
  > It looks like it does not affect perfs visibly on a local "classic" benchmark_full.

### DIFF
--- a/source/routing/next_stop_time.cpp
+++ b/source/routing/next_stop_time.cpp
@@ -62,7 +62,7 @@ void NextStopTimeData::TimesStopTimes<Getter>::init(const JourneyPattern& jp, co
     const auto jpp_order = jpp.order;
     stop_times.reserve(jp.discrete_vjs.size());
     for (const auto& vj : jp.discrete_vjs) {
-        const auto& st = vj->get_corresponding_stop_time(jpp_order);
+        const auto& st = get_corresponding_stop_time(*vj, jpp_order);
         if (!getter.is_valid(st)) {
             continue;
         }
@@ -165,7 +165,7 @@ static std::pair<const type::StopTime*, DateTime> next_valid_frequency(const Sto
 
     while (best.first == nullptr && base_dt <= bound) {
         for (const auto& freq_vj : jp.freq_vjs) {
-            const auto& st = freq_vj->get_corresponding_stop_time(jpp.order);
+            const auto& st = get_corresponding_stop_time(*freq_vj, jpp.order);
 
             if (!freq_vj->accessible(vehicle_props)) {
                 continue;
@@ -202,7 +202,7 @@ static std::pair<const type::StopTime*, DateTime> previous_valid_frequency(const
 
     while (best.first == nullptr && base_dt >= bound) {
         for (const auto& freq_vj : jp.freq_vjs) {
-            const auto& st = freq_vj->get_corresponding_stop_time(jpp.order);
+            const auto& st = get_corresponding_stop_time(*freq_vj, jpp.order);
 
             if (!freq_vj->accessible(vehicle_props)) {
                 continue;
--- a/source/routing/next_stop_time.cpp
+++ b/source/routing/next_stop_time.cpp
@@ -62,7 +62,7 @@ void NextStopTimeData::TimesStopTimes<Getter>::init(const JourneyPattern& jp, co
     const auto jpp_order = jpp.order;
     stop_times.reserve(jp.discrete_vjs.size());
     for (const auto& vj : jp.discrete_vjs) {
-        const auto& st = vj->stop_time_list[jpp_order.val];
+        const auto& st = vj->get_corresponding_stop_time(jpp_order);
         if (!getter.is_valid(st)) {
             continue;
         }
@@ -165,7 +165,7 @@ static std::pair<const type::StopTime*, DateTime> next_valid_frequency(const Sto
 
     while (best.first == nullptr && base_dt <= bound) {
         for (const auto& freq_vj : jp.freq_vjs) {
-            const auto& st = freq_vj->stop_time_list[jpp.order.val];
+            const auto& st = freq_vj->get_corresponding_stop_time(jpp.order);
 
             if (!freq_vj->accessible(vehicle_props)) {
                 continue;
@@ -202,7 +202,7 @@ static std::pair<const type::StopTime*, DateTime> previous_valid_frequency(const
 
     while (best.first == nullptr && base_dt >= bound) {
         for (const auto& freq_vj : jp.freq_vjs) {
-            const auto& st = freq_vj->stop_time_list[jpp.order.val];
+            const auto& st = freq_vj->get_corresponding_stop_time(jpp.order);
 
             if (!freq_vj->accessible(vehicle_props)) {
                 continue;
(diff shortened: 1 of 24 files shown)


pbougue:
Jenkins, test this please.

benoit-bst (dismissed):
Strong type is always better :+1: 
Sweet changes

pbougue:
FYI: I tried to add a `VJ.get_stop_time(Rank<ST>)` and it's quite complicated as VJ and ST are intricate and the readability/type gain was not so motivating as we sometimes request stop_time from a `Rank<JPP>`...

So no further work planned around this except from reviews ;-)

pbougue:
Jenkins, test this please.

source/routing/next_stop_time.cpp
```diff
@@ -59,10 +59,10 @@ bool NextStopTimeData::Arrival::is_valid(const type::StopTime& st) const {
 template <typename Getter>
 void NextStopTimeData::TimesStopTimes<Getter>::init(const JourneyPattern& jp, const JourneyPatternPoint& jpp) {
     // collect the stop times at the given jpp
-    const size_t jpp_order = jpp.order;
+    const auto jpp_order = jpp.order;
```
nonifier, line 62: cheap trick for ya :smile: 

```suggestion
    const auto jpp_order = jpp.order.val;
```

source/type/tests/test.cpp
```diff
@@ -239,27 +239,30 @@ BOOST_AUTO_TEST_CASE(get_sections_ranks) {
 
     auto sa = [&](const std::string& id) { return b.get<type::StopArea>(id); };
 
+    using rst = Rank<StopTime>;
```
nonifier: Why not typedefing this in a common place, and reusing it everywhere ? 

```cpp
using RankStopTime = Rank<StopTime>
```

nonifier (commented):
Strangely enough, `Rank` really looks like the bastard child from an `int` and an `iterator` :thinking:

source/type/fwd_type.h
```diff
@@ -29,6 +29,8 @@ www.navitia.io
 */
 #pragma once
 
+#include "utils/rank.h"
```
pbougue: Not sure if this should be included already.
Any opinion/idea welcome on that.

source/type/vehicle_journey.h
```diff
@@ -35,6 +35,7 @@ www.navitia.io
 #include "validity_pattern.h"
 #include <set>
 #include <boost/serialization/split_member.hpp>
+#include "routing/journey_pattern_container.h"
```
pbougue: Not sure VJ should know about JP and JPP (used for `VJ.get_corresponding_stop_time(JPP.order) -> ST`).
Any opinion/idea welcome on that.

An idea would be to move that method somewhere else : `JPP.get_corresponding_ST(VJ) -> ST` ?

pbougue (commented):
Poke @kinnou02 and @nonifier on those 2 notes.

source/type/vehicle_journey.cpp
```diff
@@ -293,7 +301,7 @@ std::set<uint16_t> VehicleJourney::get_sections_ranks(const StopArea* start_sa,
         // Must close section before potentially starting a new one
         // Keep going if where are still in the end stop area
         if (section_ending && end_sa->idx != sa->idx) {
-            for (uint16_t i = *section_start_rank, last = st.order() - 1; i <= last; ++i) {
+            for (auto i = *section_start_rank, last = st.order() - 1; i <= last; ++i) {
```
benoit-bst: I think you have warning signed/unsigned mismatch for the following code, no ?

source/type/vehicle_journey.cpp
```diff
@@ -324,7 +332,7 @@ std::set<uint16_t> VehicleJourney::get_sections_ranks(const StopArea* start_sa,
 
     // Must close a potential section
     if (section_ending) {
-        for (uint16_t i = *section_start_rank; i <= vj->stop_time_list.size() - 1; ++i) {
+        for (auto i = *section_start_rank; i <= RankStopTime(vj->stop_time_list.size() - 1); ++i) {
```
benoit-bst: Don't know if the readability is better now, if you have no knowledge in this part of code. You see what I mean ?
`vj->stop_time_list.size() - 1` is already a size, so why it is better to have `RankStopTime(vj->stop_time_list.size() - 1)`

source/type/vehicle_journey.h
benoit-bst: I have no opinion about except that you have to include header when only if it is necessary. 
I think you know that. The compiler can say it if that is missing :robot:

source/type/fwd_type.h
nonifier: The whole point of forward declaring is to prevent from including all the files on earth ;)
I'm strongly against it as it's breaking the benefits of this file.

Although, you can add the forward declaration of your rank types here.

source/type/vehicle_journey.h
nonifier: I would rather have a forward declaration of `RankJourneyPatternPoint` here in the include file, and this include in `vehicule_journey.cpp`.

pbougue: OK, for the forward, but the main question is "Between VJ and JPP, who should know about the other?"
Or should we even put that outside of both?

source/type/vehicle_journey.cpp
pbougue: I change it to avoid `- 1` for readability, cleanness and signed/unsigned issue (FYI no clang warning were raised, don't know about GCC).
But it's nice to keep the `Rank` type as it's inserted below, so I'll keep it for now.
Let me know if you think of a better alternative.

pbougue:
Jenkins, test this please.

source/type/vehicle_journey.h
kinnou02: Ideally `::type` should have no dependency to most of our codebase, and certainly not to `::routing`

nonifier (approved):
I had a look at that `for` loop, but couldn't figure out a better way of doing it :(

source/type/vehicle_journey.cpp
```diff
@@ -301,7 +301,7 @@ std::set<RankStopTime> VehicleJourney::get_sections_ranks(const StopArea* start_
         // Must close section before potentially starting a new one
         // Keep going if where are still in the end stop area
         if (section_ending && end_sa->idx != sa->idx) {
-            for (auto i = *section_start_rank, last = st.order() - 1; i <= last; ++i) {
+            for (auto i = *section_start_rank; i < st.order(); ++i) {
```
benoit-bst, line 300: :+1:

source/type/vehicle_journey.cpp
benoit-bst: Okay, weirdo for the `auto`. Modern compiler must manage `auto` easyly with deduction. 
Good to know

benoit-bst (approved):

pbougue:
Jenkins, test this please.

benoit-bst (approved):

PR #2991, opened 2019-10-08
Kraken: Add isochrone distributed

AurelienLP:
This PR is needed : https://github.com/CanalTP/navitia-proto/pull/137
Readable commit by commit!

Nowadays, when we call isochrone with distributed, we use the new_default scenario.

We then have to create a new interface in kraken which will receive multiple `stop_points`, result of a previous `street_network_matrix` computed by a calculator (Kraken, Asgard...).
We'll then call `raptor.isochrone` the same way we are doing for new_default.

We need `isochrone_center`, the starting point of the isochrone sent in the request to build the final response.


source/kraken/worker.cpp
```diff
@@ -642,14 +642,16 @@ JourneysArg::JourneysArg(std::vector<type::EntryPoint> origins,
...
+                         std::vector<uint64_t> datetimes,
+                         type::EntryPoint isochrone_center)
     : origins(origins),
       accessibilite_params(accessibilite_params),
       forbidden(forbidden),
       allowed(allowed),
       rt_level(rt_level),
       destinations(destinations),
-      datetimes(datetimes) {}
+      datetimes(datetimes),
```
xlqian: (hmmm,  it's not your code, we should have std::move these vectors :/ )

source/routing/raptor_api.cpp
```diff
@@ -1452,6 +1449,44 @@ void make_isochrone(navitia::PbCreator& pb_creator,
...
+                                uint32_t max_transfers) {
+    // Create datetime
+    auto const datetimes = parse_datetimes(raptor, {datetime_timestamp}, pb_creator, clockwise);
+    if (pb_creator.has_error() || pb_creator.has_response_type(pbnavitia::DATE_OUT_OF_BOUNDS)) {
+        return;
+    }
+
+    // Get stop points for departure and destination
+    auto const departures = make_map_stop_point_duration(stop_points, raptor.data.pt_data->stop_points_map);
+    auto const datetime = datetimes.front();
```
xlqian, line 1472: if `datetimes` is empty -> boom!

source/kraken/worker.h
```diff
@@ -59,13 +59,15 @@ struct JourneysArg {
     type::RTLevel rt_level;
     std::vector<type::EntryPoint> destinations;
     std::vector<uint64_t> datetimes;
+    type::EntryPoint isochrone_center;
```
xlqian: i'd make this optional isochrone_center, if i were you

AurelienLP: Yes, I hesitated between this version, a `boost::optionnal` or a `unique_ptr`

source/routing/raptor_api.cpp
AurelienLP: It is sometimes tested but I forgot it here :D

AurelienLP: `datetimes` is `request.datetimes(0)` so it would crash earlier anyway. It is the same for almost every API in kraken

source/kraken/worker.cpp
```diff
@@ -691,8 +693,12 @@ navitia::JourneysArg Worker::fill_journeys(const pbnavitia::JourneysRequest& req
 
     type::RTLevel rt_level = get_realtime_level(request.realtime_level());
 
+    const auto isochrone_center = request.has_isochrone_center()
+                                      ? create_journeys_entry_point(request.isochrone_center(), sn_params, data, true)
```
nonifier, line 697: I don't get why you need an `isochrone_center` when in new default, you can just specified an `origin` or a `destination`. 
I guess my question is: what's the difference between `origins` and `isochrone_center` when you ask for an isochrone ?

source/kraken/worker.cpp
```diff
@@ -780,6 +786,19 @@ void Worker::pt_ref(const pbnavitia::PTRefRequest& request) {
         boost::make_optional(request.has_until_datetime(), bt::from_time_t(request.until_datetime())), rt_level, *data);
 }
 
+// returns true if there is an error
+bool Worker::set_isochrone_distributed_args(const pbnavitia::JourneysRequest& request,
+                                            JourneysArg& arg,
+                                            const std::string& name) {
+    auto has_error = set_journeys_args(request, arg, name);
+    if (!has_error && !arg.isochrone_center) {
+        err_msg_isochron(this->pb_creator, "isochrone_distributed needs an isochrone_center");
```
nonifier: The check that an `isochrone_center` must exist can be done much earlier.
It's a pre-requisite to `isochrone_distributed`.

```cpp
void Worker::isochrone_distributed(const pbnavitia::JourneysRequest& request){
  if (!request.has_isochrone_center()) {
        err_msg_isochron(this->pb_creator, "isochrone_distributed needs an isochrone_center");
  [...]
```

By doing the check earlier, this function can be removed.

source/kraken/worker.cpp
```diff
@@ -990,6 +1009,20 @@ void Worker::direct_path(const pbnavitia::Request& request) {
                              dp_request.clockwise());
 }
 
+void Worker::isochrone_distributed(const pbnavitia::JourneysRequest& request) {
+    auto arg = JourneysArg();
+    bool has_error = set_isochrone_distributed_args(request, arg, "isochrone_distributed");
```
nonifier: As a good practice, it's best to return `false` on an error. 
It's way more readable to have something like : 
```cpp
 if (!set_isochrone_distributed_args(....)) {
        return; // and log an error...
    }
```

source/routing/raptor_api.cpp
```diff
@@ -1184,6 +1184,22 @@ static std::vector<bt::ptime> parse_datetimes(RAPTOR& raptor,
...
+    const std::vector<type::EntryPoint>& entryPointList,
+    const std::unordered_map<std::string, type::StopPoint*>& raptor_stop_points_map) {
+    routing::map_stop_point_duration results;
+    for (const auto& entryPoint : entryPointList) {
+        auto it = raptor_stop_points_map.find(entryPoint.uri);
+        if (it != raptor_stop_points_map.end()) {
+            results[SpIdx{*it->second}] = navitia::seconds(entryPoint.access_duration);
+        } else {
+            // for now we throw, maybe we should ignore them
+            throw navitia::recoverable_exception("stop_point " + entryPoint.uri + " not found");
```
nonifier: You have a `map_find` utility to avoid a lot of boiler plate :
```cpp
    utils::make_map_find(raptor_stop_points_map, entryPoint.uri)
        .if_found([&](const type::StopPoint* sp){
            results[SpIdx{*sp}] = navitia::seconds(entryPoint.access_duration); 
        })
        .if_not_found([](){
            // for now we throw, maybe we should ignore them
            throw navitia::recoverable_exception("stop_point " + entryPoint.uri + " not found"); 
        });
```

source/kraken/worker.cpp
AurelienLP: The point is that a didn't want to touch new_default AT ALL. In new_default, `isochrone_center` is never filled, we have either one value in `origins` (`from` of the request) or one value in `destinations` (`to` of the request).

In distributed, `origins` and `destinations` are the result of a `street_network_matrix`. One is filled, the other empty, depending on if we ask for an isochrone or a reverse_isochrone.

Now, to answer your question:
1) isochrone_center = real starting point of the isochrone (in distributed mode only !).
2) origins = real starting point of the isochrone (in new_default mode only!) OR stop_points results of a previous street_network_matrix computation (in distributed mode only!).

source/kraken/worker.cpp
AurelienLP: Totally agree with you! However, since `set_journeys_args` used in `set_isochrone_distributed_args` is doing the other way around, I preferred to stick with that and be consistent.

This is a problem though and has to be fixed.

source/routing/raptor_api.cpp
AurelienLP: I'll try that, thanks!

source/routing/raptor_api.cpp
```diff
@@ -1452,6 +1449,44 @@ void make_isochrone(navitia::PbCreator& pb_creator,
...
 }
 
+void make_isochrone_distributed(navitia::PbCreator& pb_creator,
+                                RAPTOR& raptor,
+                                const type::EntryPoint& isochrone_center,
+                                const std::vector<type::EntryPoint>& stop_points,
+                                const uint64_t datetime_timestamp,
+                                bool clockwise,
+                                const type::AccessibiliteParams& accessibilite_params,
+                                std::vector<std::string> forbidden,
```
```suggestion
                                const std::vector<std::string>& forbidden,
```
?
Try to do a maximum of const and reference if possible.

source/routing/tests/routing_api_test.cpp
```diff
@@ -1934,6 +1942,14 @@ BOOST_FIXTURE_TEST_CASE(reverse_isochrone, isochrone_fixture) {
...
     BOOST_CHECK_EQUAL(result.journeys(1).departure_date_time(), "20150615T082600"_pts);
     BOOST_CHECK_EQUAL(result.journeys(1).arrival_date_time(), "20150615T110000"_pts);
+
+    // We ask the same request in distributed
+    // And check that the response are exactly equivalent
+    const std::vector<navitia::type::EntryPoint> origins{{ep}};
+    navitia::PbCreator pb_creator_distributed(data_ptr, boost::gregorian::not_a_date_time, null_time_period);
+    nr::make_isochrone_distributed(pb_creator_distributed, raptor, ep, origins, "20150615T110000"_pts, false, {}, {},
+                                   {}, nt::RTLevel::Base, 3 * 60 * 60);
+    BOOST_CHECK_EQUAL(result.DebugString(), pb_creator_distributed.get_response().DebugString());
```
benoit-bst, line 1952: Good point !

benoit-bst (approved):

source/routing/raptor_api.cpp
AurelienLP: Yes indeed! I've missed that! Thanks 😉

benoit-bst: :+1: 
There are some input function without `const ref` input prameters. Takes a ride in the PR :pray:

AurelienLP: Done everywhere I could! :smile:

source/kraken/worker.cpp
AurelienLP: Fixed for `set_journeys_args`

source/kraken/worker.cpp
AurelienLP: Done! `set_isochrone_distributed_args` is no more :cry:

source/kraken/worker.cpp
```diff
@@ -780,37 +786,39 @@ void Worker::pt_ref(const pbnavitia::PTRefRequest& request) {
         boost::make_optional(request.has_until_datetime(), bt::from_time_t(request.until_datetime())), rt_level, *data);
 }
 
-// returns true if there is an error
+// returns false if there is an error
```
xlqian, line 789: lol

xlqian (approved):

PR #2993, opened 2019-10-08
[Kraken] fix count in places_nearby

xlqian:
https://api.navitia.io/v1/coord/-0.5729449%3B44.8303673/places_nearby?distance=10000&type%5B%5D=stop_area&filter=physical_mode.uri%3Dphysical_mode%3ATramway&count=10&

https://jira.kisio.org/browse/NAVITIAII-2896

- [x] add test (pagination, filter, total number of objects?)

### DIFF
--- a/source/jormungandr/tests/places_tests.py
+++ b/source/jormungandr/tests/places_tests.py
@@ -129,6 +129,24 @@ class TestPlaces(AbstractTestFixture):
         for place in places_nearby:
             assert place["embedded_type"] == "stop_area"
 
+    def test_places_nearby_with_coords_with_type_and_pagination(self):
+        """check the pagination"""
+
+        query = "/v1/coords/0.000001;0.000898311281954/places_nearby?type[]=stop_point&count=1"
+        response = self.query(query)
+        places_nearby = response['places_nearby']
+        assert len(places_nearby) == 1
+        is_valid_places(places_nearby)
+        assert places_nearby[0]["embedded_type"] == "stop_point"
+        assert places_nearby[0]["id"] == "stop_point:stopB"
+
+        response = self.query(query + "&start_page=1")
+        places_nearby = response['places_nearby']
+        assert len(places_nearby) == 1
+        is_valid_places(places_nearby)
+        assert places_nearby[0]["embedded_type"] == "stop_point"
+        assert places_nearby[0]["id"] == "stop_point:stopC"
+
     def test_places_nearby_with_coords_with_type_filter_and_count(self):
         """
         check places_nearby with /coords and type[]=stop_point, filter=line.uri=D, count=1
@@ -136,7 +154,7 @@ class TestPlaces(AbstractTestFixture):
         from the coord, you can find:
 
         stop_B -> 70m
-        stop_C -> 72m ( <- it's the stop_point that we are expecting)
+        stop_C -> 72m ( <- it's the stop_point that we are expecting with filter=line.uri=D)
         stop_A -> 121m
         stop_uselessA -> 121m
         """
--- a/source/jormungandr/tests/places_tests.py
+++ b/source/jormungandr/tests/places_tests.py
@@ -134,6 +134,7 @@ class TestPlaces(AbstractTestFixture):
 
         query = "/v1/coords/0.000001;0.000898311281954/places_nearby?type[]=stop_point&count=1"
         response = self.query(query)
+        assert response['pagination']['total_result'] == 4
         places_nearby = response['places_nearby']
         assert len(places_nearby) == 1
         is_valid_places(places_nearby)
@@ -141,6 +142,7 @@ class TestPlaces(AbstractTestFixture):
         assert places_nearby[0]["id"] == "stop_point:stopB"
 
         response = self.query(query + "&start_page=1")
+        assert response['pagination']['total_result'] == 4
         places_nearby = response['places_nearby']
         assert len(places_nearby) == 1
         is_valid_places(places_nearby)
--- a/source/jormungandr/tests/places_tests.py
+++ b/source/jormungandr/tests/places_tests.py
@@ -129,6 +129,32 @@ class TestPlaces(AbstractTestFixture):
         for place in places_nearby:
             assert place["embedded_type"] == "stop_area"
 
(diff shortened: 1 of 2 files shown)


source/proximity_list/proximitylist_api.cpp
```diff
@@ -149,13 +152,14 @@ void find(navitia::PbCreator& pb_creator,
             // we sort the proximity list on the indexes
```
pbougue: For a places nearby we want the 10 closest, not the 10 with lowest index, which is what we would get, no?

source/proximity_list/proximitylist_api.cpp
```diff
@@ -126,16 +127,18 @@ void find(navitia::PbCreator& pb_creator,
                 pb_creator.fill_pb_error(pbnavitia::Error::bad_filter, "ptref : " + pt_error.more);
                 return;
             }
+            // We have to find all objects within distance, then apply the filter
+            search_count = -1;
         }
         switch (type) {
             case nt::Type_e::StopArea:
-                list = pb_creator.data->pt_data->stop_area_proximity_list.find_within(coord, distance, count);
+                list = pb_creator.data->pt_data->stop_area_proximity_list.find_within(coord, distance, search_count);
```
pbougue, line 134: Can't you modify stop_area_proximity_list before? (copy then filter it?).
Or maybe pass a visitor to filter into radius_search?

source/proximity_list/proximitylist_api.cpp
```diff
@@ -149,13 +152,14 @@ void find(navitia::PbCreator& pb_creator,
...
             std::sort(list.begin(), list.end(), comp);
+            size_t added = 0;
 
             for (auto idx : indexes) {
                 auto it = std::lower_bound(list.begin(), list.end(), std::make_pair(idx, nt::GeographicalCoord()),
                                            comp);  // we create a mock pair and only compare the first elt
                 if (it != list.end() && (*it).first == idx) {
                     final_list.push_back(*it);
-                    if (final_list.size() == list.size()) {
+                    if (final_list.size() == list.size() || ++added == count) {
```
pbougue: I don't understand that change. Is it a debug leftover?

source/proximity_list/proximitylist_api.cpp
xlqian: I was thinking the same thing, if i do so, i have to rebuild the whole data structure, that means the data structure must be rebuilt per request

source/proximity_list/proximitylist_api.cpp
xlqian: list is sorted because we want to use lower_bound to accelerate the serach:https://github.com/CanalTP/navitia/pull/2993/files#diff-6aa4b0eead667bcb2df54e0b725bdb6cR158

source/proximity_list/proximitylist_api.cpp
xlqian: No.
without this, if the filter is not empty, the search_count will be -1, then it'll return all objects that it found..

source/proximity_list/proximitylist_api.cpp
pbougue: :thinking: indeed.
Maybe it's worth filtering when we have some (and keep the pre-built one most of the time), but not sure it's worthy.
You can resolve this conversation if you don't think the idea is promising :+1:

source/proximity_list/proximitylist_api.cpp
pbougue: Gnaa, now I get what it does :eyes: 

Does that work for pagination? I'm pretty sure the answer is the same in case there is no filter, also :page_with_curl: 
(In my memories pagination used to work for places_nearby, like one year ago when we looked into geovelo's perfs).
Maybe we should use `end_pagination` instead of `count` (something like this) ?
It's worth testing that too.

pbougue (commented):
FYI: Added `do_not_merge` for tests.

source/proximity_list/proximitylist_api.cpp
xlqian: to be honest, there is something i don't understand neither in the pagination, it seems to me that count is used a 'nb of object per page' not 'a total number of objects to be returned'...

source/proximity_list/proximitylist_api.cpp
xlqian: Since the performance is quite acceptable so far, i resolve this conversation

source/proximity_list/proximitylist_api.cpp
xlqian: OK, my bad, i totally misunderstood this code...

kadhikari (commented):
Is it easy to test or not?

xlqian:
@kadhikari there is small note in the description of the PR :)https://github.com/CanalTP/navitia/pull/2993#issue-325800970

pbougue (commented):
As we lost it also in the process but didn't realized it until this PR:
It may also be worth adding "/v1/coords/0.000001;0.000898311281954/places_nearby?type[]=stop_point&count=1&`start_page=1`" to check we have pagination.
And check the total number of results.

Anyway, no bug expected from this :+1:

kadhikari (approved):

source/jormungandr/tests/places_tests.py
```diff
@@ -129,6 +129,50 @@ def test_places_nearby_with_coords_without_region_and_type(self):
...
+        """check the pagination"""
+
+        query = "/v1/coords/0.000001;0.000898311281954/places_nearby?type[]=stop_point&count=1"
+        response = self.query(query)
+        places_nearby = response['places_nearby']
+        assert len(places_nearby) == 1
+        is_valid_places(places_nearby)
+        assert places_nearby[0]["embedded_type"] == "stop_point"
+        assert places_nearby[0]["id"] == "stop_point:stopB"
+
```
pbougue, line 143: minor: you may as well test `pagination.total_result` == 4

pbougue (approved):

pbougue (approved):
We have to approve after every change for mergify I guess

benoit-bst (approved):

pbougue (approved):

kadhikari (approved):

PR #2994, opened 2019-10-09
Ed: use fare_zone_id instead of zone_id in fusio_parser

kadhikari:
For details see : https://jira.kisio.org/browse/NAVP-1388


source/ed/connectors/fusio_parser.cpp
```diff
@@ -208,6 +208,11 @@ void StopsFusioHandler::init(Data& data) {
     comment_id_c = csv.get_pos_col("comment_id");
     visible_c = csv.get_pos_col("visible");
     geometry_id_c = csv.get_pos_col("geometry_id");
+    // For NTFS v0.9, we sould use fare_zone_id instead of zone_id if present
```
pbougue:
```suggestion
    // Since NTFS v0.9, using fare_zone_id (if absent, fallback to zone_id)
```

pbougue (approved):
Can you add retrocompatibility test based on ntfs_dst or ntfs_v5?

pbougue (approved):

benoit-bst (approved):

PR #2997, opened 2019-10-10
ed2nav: Writes a temp file before replacing the old data

BaptisteB-A:
Now serializes the data in a temporary file, and replaces the
previous `data.nav.lz4` file. In case of error, the temporary file is
deleted instead of the old data file as it used to be, so the old data
file doesn't get corrupted.

JIRA : https://jira.kisio.org/browse/NAVP-1301


buildbotctp:

kinnou02:
Ok to test

source/ed/ed2nav.cpp
```diff
@@ -165,6 +165,33 @@ struct FindAdminWithCities {
     }
 };
 
+bool write_data_to_file(const std::string& output_filename, navitia::type::Data& data) {
+    auto logger = log4cplus::Logger::getInstance("log");
+    std::string temp_output = output_filename + ".temp";
+    try {
+        data.save(temp_output);
+        try {
+            remove(output_filename.c_str());
```
kinnou02: If this is `remove` from `stdio` you need to check the return value, same things for `rename`.

source/ed/ed2nav.cpp
```diff
@@ -165,6 +165,33 @@ struct FindAdminWithCities {
...
+            rename(temp_output.c_str(), output_filename.c_str());
+            LOG4CPLUS_INFO(logger, "Data saved");
+        } catch (const navitia::exception& e) {
```
kinnou02: nothing in this block will throw a `navitia::exception`

source/ed/ed2nav.cpp
kinnou02: You should probably rename it first.

kinnou02 (dismissed):

source/ed/ed2nav.cpp
BaptisteB-A: I see what you mean but if I rename `temp_output` file name to `output_filename`, and then remove the `output_filename` file, wouldn't there be a problem ?

source/ed/ed2nav.cpp
```diff
@@ -165,6 +165,31 @@ struct FindAdminWithCities {
     }
 };
 
+bool write_data_to_file(const std::string& output_filename, navitia::type::Data& data) {
+    auto logger = log4cplus::Logger::getInstance("log");
+    std::string temp_output = output_filename + ".temp";
+    try {
+        data.save(temp_output);
+        if (remove(output_filename.c_str()) == 0) {
+            if (rename(temp_output.c_str(), output_filename.c_str()) == 0)
```
kinnou02: Always put brackets, it doesn't cost anything and reduce the risk of bug later on.

source/ed/ed2nav.cpp
```diff
@@ -165,6 +165,31 @@ struct FindAdminWithCities {
...
+        } else {
+            LOG4CPLUS_ERROR(logger, "Error deleting file: No such file or directory");
+        }
+    } catch (const navitia::exception& e) {
+        LOG4CPLUS_ERROR(logger, "Unable to save");
+        LOG4CPLUS_ERROR(logger, e.what());
+        if (remove(temp_output.c_str()) == 0)
+            LOG4CPLUS_INFO(logger, "Temp data removed because it was corrupted");
+        else
+            LOG4CPLUS_ERROR(logger, "Error deleting temp file: No such file or directory");
```
kinnou02: you log an error that may or may not be the one that occured, it will mislead people if the error is caused by something else like a filesystem in readonly. You should use `strerror` to get the correct error message

source/ed/ed2nav.cpp
kinnou02: What I meant was to rename the current `data.nav.lz4` to something like `data.nav.lz4.bak`, then rename `data.nav.lz4.temp`  to `data.nav.lz4`, this way if the renaming fail for whatever reason we still have a valid dataset to recover.

source/ed/ed2nav.cpp
```diff
@@ -165,6 +164,42 @@ struct FindAdminWithCities {
     }
 };
 
+bool write_data_to_file(const std::string& output_filename, navitia::type::Data& data) {
+    auto logger = log4cplus::Logger::getInstance("log");
+    std::string temp_output_filename = output_filename + ".temp";
+    std::string backup_output_filename = output_filename;
```
pbougue: Shouldn't it be 
```suggestion
    std::string backup_output_filename = output_filename + ".bak";
```
?
(then used below)

source/ed/ed2nav.cpp
```diff
@@ -165,6 +164,42 @@ struct FindAdminWithCities {
...
+    } catch (const navitia::exception& e) {
+        LOG4CPLUS_ERROR(logger, "Unable to save, going back to previous state...");
+        LOG4CPLUS_ERROR(logger, e.what());
+        if (remove(temp_output_filename.c_str()) != 0) {
+            LOG4CPLUS_ERROR(logger, "Unable to remove temp data file: " << std::strerror(errno));
+            return false;
+        }
+        LOG4CPLUS_INFO(logger, "Temp file removed");
+        if (rename(output_filename.c_str(), (backup_output_filename).c_str()) != 0) {
+            LOG4CPLUS_INFO(logger, "Unable to rename old data file: " << std::strerror(errno));
```
pbougue: Why info more than error?

source/ed/ed2nav.cpp
```diff
@@ -165,6 +164,42 @@ struct FindAdminWithCities {
     }
 };
 
+bool write_data_to_file(const std::string& output_filename, navitia::type::Data& data) {
```
pbougue: I'd like more a `void` method raising if there is an error, than a `bool` way of managing errors.

source/ed/ed2nav.cpp
```diff
@@ -165,6 +164,42 @@ struct FindAdminWithCities {
...
+    std::string backup_output_filename = output_filename;
+    if (boost::filesystem::exists(output_filename)) {
+        if (rename(output_filename.c_str(), (output_filename + ".bak").c_str()) != 0) {
+            LOG4CPLUS_INFO(logger, "Unable to rename old data file: " << std::strerror(errno));
+            return false;
+        }
+        LOG4CPLUS_INFO(logger, "Old data file successfully renamed to " << output_filename << ".bak");
+    }
+    try {
+        data.save(temp_output_filename);
```
pbougue: Why not writing it into output_filename directly (as it has been mv)?

source/ed/ed2nav.cpp
```diff
@@ -165,6 +164,42 @@ struct FindAdminWithCities {
...
+        LOG4CPLUS_INFO(logger, "File: " << output_filename << " saved");
+    } catch (const navitia::exception& e) {
+        LOG4CPLUS_ERROR(logger, "Unable to save, going back to previous state...");
+        LOG4CPLUS_ERROR(logger, e.what());
+        if (remove(temp_output_filename.c_str()) != 0) {
+            LOG4CPLUS_ERROR(logger, "Unable to remove temp data file: " << std::strerror(errno));
+            return false;
+        }
+        LOG4CPLUS_INFO(logger, "Temp file removed");
+        if (rename(output_filename.c_str(), (backup_output_filename).c_str()) != 0) {
```
pbougue: It's the opposite, no?

source/ed/ed2nav.cpp
```diff
@@ -165,6 +164,42 @@ struct FindAdminWithCities {
...
+            LOG4CPLUS_ERROR(logger, "Unable to rename temp data file: " << std::strerror(errno));
+            return false;
+        }
+        LOG4CPLUS_INFO(logger, "File: " << output_filename << " saved");
+    } catch (const navitia::exception& e) {
+        LOG4CPLUS_ERROR(logger, "Unable to save, going back to previous state...");
+        LOG4CPLUS_ERROR(logger, e.what());
+        if (remove(temp_output_filename.c_str()) != 0) {
+            LOG4CPLUS_ERROR(logger, "Unable to remove temp data file: " << std::strerror(errno));
+            return false;
```
pbougue: Here you still have to mv backup file to it's original name, it's too early to return I think.

benoit-bst:
Jenkins, test this please !

benoit-bst (approved):

source/ed/tests/ed2nav_test.cpp
```diff
@@ -50,3 +53,26 @@ BOOST_AUTO_TEST_CASE(should_throw_on_bad_connection_string) {
     const char* argv[] = {"ed2nav_test", "--connection-string=\"blahblah\""};
     BOOST_CHECK_THROW(ed::ed2nav(2, argv), std::exception);
 }
+
+BOOST_AUTO_TEST_CASE(writing_test_ed2nav_should_not_throw) {
```
pbougue, line 57: Can you add failing tests?
(like rename impossible, file not generated, etc.)

source/ed/ed2nav.cpp
BaptisteB-A: I didn't make the names very clear but I wanted to store the old output_filename in back_output_filename before adding ".bak" to the old output_filename. The new data file name with temp_output_filename is then renamed into back_output_filename, or, if you prefer, the original output_filename. Hum, sorry I'm not sure if that really is clearer..

BaptisteB-A:
Jenkins, test this please !

source/ed/ed2nav.cpp
BaptisteB-A: Fixed

source/ed/ed2nav.cpp
BaptisteB-A: Why not a `bool` ?

source/ed/ed2nav.cpp
BaptisteB-A: No, we want the old file to be renamed after back_output_filename.

source/ed/tests/ed2nav_test.cpp
BaptisteB-A: Added a test on file save

AurelienLP (approved):

source/ed/ed2nav.cpp
```diff
@@ -165,6 +164,41 @@ struct FindAdminWithCities {
     }
 };
 
+bool rename_old_file(const std::string& output_filename, std::string& backup_output_filename) {
+    auto logger = log4cplus::Logger::getInstance("log");
```
xlqian: you can give a more specific name for the logger 
`auto logger = log4cplus::Logger::getInstance("ed2nav::rename_old_file");`

source/ed/ed2nav.h
```diff
@@ -27,11 +27,39 @@ IRC #navitia on freenode
...
+namespace navitia {
+namespace type {
+struct Data;
+}  // namespace type
+}  // namespace navitia
+
 namespace ed {
 
+template <class T = navitia::type::Data>
+bool try_save_file(std::string& filename, T& data) {
```
xlqian: `const T& data` ?

source/ed/ed2nav.cpp
```diff
@@ -165,6 +164,41 @@ struct FindAdminWithCities {
...
+    if (rename(temp_output_filename.c_str(), output_filename.c_str()) != 0) {
+        LOG4CPLUS_ERROR(logger, "Unable to rename temp data file: " << temp_output_filename << std::strerror(errno));
+        return false;
+    }
+    LOG4CPLUS_INFO(logger, "File: " << output_filename << " saved");
+    return true;
+}
+
+template <class T = navitia::type::Data>
+bool write_data_to_file(const std::string& output_filename, T& data) {
```
xlqian: `const T& data` ?

source/ed/ed2nav.h
```diff
@@ -27,11 +27,39 @@ IRC #navitia on freenode
...
+        return false;
+    }
+    return true;
+}
+
+bool rename_old_file(std::string& output_filename, std::string& backup_output_filename);
+bool rename_temp_file(std::string& temp_output_filename, std::string& output_filename);
+
+template <class T = navitia::type::Data>
+bool write_data_to_file(const std::string& output_filename, T& data);
```
xlqian: `const T& data` ?

source/ed/ed2nav.cpp
pbougue: As discussed aside, no need to change here.
To trace it a bit : the conventional way of dealing with errors in C++ (and navitia) is more throwing exception.

benoit-bst (approved):

source/ed/ed2nav.cpp
```diff
@@ -165,6 +164,41 @@ struct FindAdminWithCities {
...
+
+bool rename_temp_file(const std::string& temp_output_filename, const std::string& output_filename) {
+    auto logger = log4cplus::Logger::getInstance("ed2nav::rename_temp_file");
+    if (rename(temp_output_filename.c_str(), output_filename.c_str()) != 0) {
+        LOG4CPLUS_ERROR(logger, "Unable to rename temp data file: " << temp_output_filename << std::strerror(errno));
+        return false;
+    }
+    LOG4CPLUS_INFO(logger, "File: " << output_filename << " saved");
+    return true;
+}
```
kadhikari: Minor: Two functions above can be mutualized non?

kadhikari (approved):

source/ed/ed2nav.cpp
BaptisteB-A: Right, thanks !

source/ed/tests/ed2nav_test.cpp
nonifier: Unfortunately, making a call to `rename` to fail (in a decent way), would mean to mock the free function which is very complicated without a proper mocking  framework :cry:

nonifier (approved):

AurelienLP:
Jenkins, test this please.

AurelienLP (approved):

AurelienLP:
Jenkins, test this please.

mbounabi:
Jenkins, test this please.

PR #2999, opened 2019-10-14
jormun: Upgrade status with equipment details and ridesharing informations

benoit-bst:
It was missing equipment details and ridesharing informations inside Status API.

Now we have :
status_report

### DIFF
--- a/source/jormungandr/jormungandr/interfaces/v1/serializer/status.py
+++ b/source/jormungandr/jormungandr/interfaces/v1/serializer/status.py
@@ -149,6 +149,36 @@ class RidesharingServicesSerializer(OutsideServiceCommon):
         return obj.get('network', None)
 
 
+class EquipmentProvidersSerializer(NullableDictSerializer):
+    key = MethodField(schema_type=str, display_none=False)
+    codes_types = MethodField(schema_type=str, many=True, display_none=True)
+    timeout = MethodField(schema_type=int, display_none=False)
+    fail_max = MethodField(schema_type=int, display_none=False)
+
+    def get_key(self, obj):
+        return obj.get('key', None)
+
+    def get_codes_types(self, obj):
+        return obj.get('codes_types', [])
+
+    def get_timeout(self, obj):
+        return obj.get('timeout', None)
+
+    def get_fail_max(self, obj):
+        return obj.get('fail_max', None)
+
+
+class EquipmentProvidersServicesSerializer(NullableDictSerializer):
+    equipment_providers_keys = MethodField(schema_type=str, many=True, display_none=True)
+    equipment_providers = EquipmentProvidersSerializer(many=True, display_none=False)
+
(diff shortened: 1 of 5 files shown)


benoit-bst:
Jenkins, test this please !

source/jormungandr/jormungandr/interfaces/v1/serializer/status.py
```diff
@@ -131,7 +131,52 @@ def get_backend_class(self, obj):
 
 
 class RidesharingServicesSerializer(OutsideServiceCommon):
-    pass
+    rating_scale_min = MethodField(schema_type=int, display_none=False)
+    rating_scale_max = MethodField(schema_type=int, display_none=False)
+    crowfly_radius = MethodField(schema_type=int, display_none=False)
+    network = MethodField(schema_type=str, display_none=False)
```
nonifier: minor : Why can't those just be `Field` ? 
As `Field` you don't need to write specific functions  for each of them.

nonifier (approved):

source/jormungandr/jormungandr/interfaces/v1/serializer/status.py
benoit-bst: Serpy serializer is a real fine art :alien: 
It allows to specify the value if the field is not present, for example [] for an empty list. otherwise the parameters is mandatory into the serpy serializer output. It is just an overload.
wide program.

source/jormungandr/jormungandr/interfaces/v1/serializer/status.py
```diff
@@ -131,7 +131,52 @@ def get_backend_class(self, obj):
...
+
+
+class EquipmentProvidersSerializer(NullableDictSerializer):
+    key = MethodField(schema_type=str, display_none=False)
+    codes_types = MethodField(schema_type=str, many=True, display_none=True)
+    timeout = MethodField(schema_type=int, display_none=False)
+    fail_max = MethodField(schema_type=int, display_none=False)
+
+    def get_key(self, obj):
+        return obj.get('key', None)
```
mbounabi: Are you sure it is needed to have a method for each parameter? Can't it be retrieved with:
` key =Field(attr='key', schema_type=str, display_none=False)` ?
Same comment for every others parameters. Is it because of the default values?

benoit-bst: if we use `Field`, it is not allowed to overload the parameter. Overload works with MethodField because it inheritances of `serpy.MethodFiled`. Only `serpy.MethodFiled` allows to overload function for parameters.
Thererby, if we want optional field, you have to use it. 

The question is more,` what parameters we want optional ?`
After a ride inside the code, i thinks it is possible to use some parameters in mandatory mode. I'll try.

nonifier (approved):

xlqian (approved):

PR #3000, opened 2019-10-14
jormun: fix ticket remove process with reversed method

benoit-bst:
I did few QA tests after merge on Dev and i found some bugs due to `remove ticket processing`.
Remove into a list during loop, have to be done in **reversed** mode :imp: 

I need to merge fastly because @kadhikari is going to create Ref


mbounabi (approved):
Fine QA... But after merge... :smiling_imp:

nonifier (approved):
nasty :snake:

PR #3001, opened 2019-10-16
jormun: robustify Timeo mechanism with interval errors handling and empty response

benoit-bst:
**Goals:**

- Display base schedule results list, when there is an internal timeo service error
- If timeo service return 0 errors (error code < 100) with an empty list, the empty list is return (seen with @SGrenet )
  It was obvious for me the return the base schedule in this case, but it seems to be wrong. That just says there is no next departures. And even if it is not normal, it is timeo side). I see this case 80% of the time

**JIRA** : https://jira.kisio.org/browse/NAVITIAII-2895

### DIFF
--- a/source/jormungandr/jormungandr/realtime_schedule/realtime_proxy.py
+++ b/source/jormungandr/jormungandr/realtime_schedule/realtime_proxy.py
@@ -236,8 +236,10 @@ class RealtimeProxy(six.with_metaclass(ABCMeta, object)):
     def record_external_failure(self, message):
         record_external_failure(message, 'realtime', unicode(self.rt_system_id))
 
-    def record_internal_failure(self, message, comment=''):
-        params = {'realtime_system_id': unicode(self.rt_system_id), 'message': message, 'comment': comment}
+    def record_internal_failure(self, message, comment=None):
+        params = {'realtime_system_id': unicode(self.rt_system_id), 'message': message}
+        if comment is not None:
+            params['comment'] = comment
         new_relic.record_custom_event('realtime_internal_failure', params)
 
     def record_call(self, status, **kwargs):
--- a/source/jormungandr/jormungandr/realtime_schedule/realtime_proxy.py
+++ b/source/jormungandr/jormungandr/realtime_schedule/realtime_proxy.py
@@ -236,8 +236,8 @@ class RealtimeProxy(six.with_metaclass(ABCMeta, object)):
     def record_external_failure(self, message):
         record_external_failure(message, 'realtime', unicode(self.rt_system_id))
 
-    def record_internal_failure(self, message):
-        params = {'realtime_system_id': unicode(self.rt_system_id), 'message': message}
+    def record_internal_failure(self, message, comment=''):
+        params = {'realtime_system_id': unicode(self.rt_system_id), 'message': message, 'comment': comment}
         new_relic.record_custom_event('realtime_internal_failure', params)
 
     def record_call(self, status, **kwargs):
--- a/source/jormungandr/jormungandr/realtime_schedule/timeo.py
+++ b/source/jormungandr/jormungandr/realtime_schedule/timeo.py
@@ -221,10 +221,6 @@ class Timeo(RealtimeProxy):
 
         next_st = st_responses[0]['NextStopTimesMessage']
 
-        # if realtime list is empty, only base results will send
-        if 'NextExpectedStopTime' not in next_st:
-            return None
-
         next_passages = []
         for next_expected_st in next_st.get('NextExpectedStopTime', []):
             # for the moment we handle only the NextStop and the direction
--- a/source/jormungandr/jormungandr/realtime_schedule/timeo.py
+++ b/source/jormungandr/jormungandr/realtime_schedule/timeo.py
@@ -215,13 +215,15 @@ class Timeo(RealtimeProxy):
                     resp_comment = message_response['ResponseComment']
                 else:
                     resp_comment = ''
-                logging.getLogger(__name__).error(
-                    'Timeo RT internal service error, code: {} - comment: {}'.format(resp_code, resp_comment)
-                )
                 self.record_internal_failure(
-                    'internal timeo error', 'code {} - {}'.format(resp_code, resp_comment)
+                    'Timeo RT internal service error',
+                    'ResponseCode: {} - ResponseComment: {}'.format(resp_code, resp_comment),
                 )
-                raise RealtimeProxyError('Timeo RT internal service error')
+                timeo_internal_error_message = 'Timeo RT internal service error, ResponseCode: {} - ResponseComment: {}'.format(
+                    resp_code, resp_comment
+                )
+                logging.getLogger(__name__).error(timeo_internal_error_message)
+                raise RealtimeProxyError(timeo_internal_error_message)
 
         st_responses = timeo_resp.get('StopTimesResponse')
         # by construction there should be only one StopTimesResponse
--- a/source/jormungandr/jormungandr/realtime_schedule/timeo.py
+++ b/source/jormungandr/jormungandr/realtime_schedule/timeo.py
@@ -109,6 +109,9 @@ class Timeo(RealtimeProxy):
         else:
             self.rate_limiter = FakeRateLimiter()
 
+        # We consider that all errors, greater than or equal to 100, are blocking
+        self.INTERNAL_TIMEO_ERROR_CODE_LIMIT = 100
+
     def __repr__(self):
         """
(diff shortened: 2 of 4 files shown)


benoit-bst:
Jenkins, test this please !

SGrenet (approved):
Next steps : with a proper return of Timeo's errrors, if an error occurs in Timeo, it should return the right error code, then if not, it's a Timeo issue

source/jormungandr/jormungandr/realtime_schedule/timeo.py
```diff
@@ -196,6 +199,17 @@ def _get_passages(self, timeo_resp, current_dt, line_uri=None):
             'timeo response: {}'.format(timeo_resp), extra={'rt_system_id': unicode(self.rt_system_id)}
         )
 
+        # internal timeo error handling
```
pbougue, line 206: Then you can remove the `TODO` above :point_up: line 188
minor: I would also put the code there to have all the error handling at the same place.

source/jormungandr/jormungandr/realtime_schedule/timeo.py
```diff
@@ -196,6 +199,17 @@ def _get_passages(self, timeo_resp, current_dt, line_uri=None):
...
+        message_responses = timeo_resp.get('MessageResponse')
+        for message_response in message_responses:
+            if message_response['ResponseCode'] >= self.INTERNAL_TIMEO_ERROR_CODE_LIMIT:
```
pbougue: for safety
```suggestion
            if 'ResponseCode' in message_response and message_response['ResponseCode'] >= self.INTERNAL_TIMEO_ERROR_CODE_LIMIT:
```
or use a get?

source/jormungandr/jormungandr/realtime_schedule/timeo.py
```diff
@@ -196,6 +199,17 @@ def _get_passages(self, timeo_resp, current_dt, line_uri=None):
...
+                logging.getLogger(__name__).error(
+                    'Timeo RT internal service error, code: {} - comment: {}'.format(
+                        message_response['ResponseCode'], message_response['ResponseComment']
+                    )
+                )
+                raise RealtimeProxyError('Timeo RT internal service error')
```
pbougue: You have to reuse the same (complete) log as above to have code and comment it in newrelic.

source/jormungandr/tests/proxy_realtime_timeo_integration_tests.py
```diff
@@ -89,13 +89,14 @@ def _call_timeo(self, url):
...
                 ],
+                "MessageResponse": [{"ResponseCode": 0, "ResponseComment": "success"}],
             }
         if url == 'S42':
             json = {
                 "CorrelationID": "AA",
                 "StopTimesResponse": [
                     {
-                        "StopID": "C:S0",
+                        "StopID": "S42",
```
pbougue, line 99: Did you change that only for consistency of test, or was it needed for tests to pass?

pbougue (commented):
This looks good functionally.

I have 2 more remarks (additionnal to the ones in code) that are not from this work:
1. _easy_: We may add the response code to the exception generated from real HTTP status code (newrelic).
2. _not sure_: It looks like this HTTP status code doesn't trigger the pybreaker... And it probably should.

source/jormungandr/tests/proxy_realtime_timeo_integration_tests.py
benoit-bst: for consistency

source/jormungandr/jormungandr/realtime_schedule/timeo.py
benoit-bst: I forgot the newrelic part, :speak_no_evil:

source/jormungandr/jormungandr/realtime_schedule/timeo.py
benoit-bst: okay, I take a look

benoit-bst:
> This looks good functionally.
> 
> I have 2 more remarks (additionnal to the ones in code) that are not from this work:
> 1. easy: We may add the response code to the exception generated from real HTTP status code.
> 2. not sure: It looks like this HTTP status code doesn't trigger the pybreaker... And it probably should.

1. There is no match between **http errors** and **timeo errors** (look the linked file into the JIRA ticket). So, it is difficult to convert anything. I prefer, when it is timeo error to write a log with that and send a newrelic (@kinnou02 mentioned that into the JIRA ticket)
2. As @kinnou02 said in the ticket, we don't want to break in this case.

nonifier (approved):

source/jormungandr/jormungandr/realtime_schedule/realtime_proxy.py
```diff
@@ -236,8 +236,8 @@ def status(self):
     def record_external_failure(self, message):
         record_external_failure(message, 'realtime', unicode(self.rt_system_id))
 
-    def record_internal_failure(self, message):
-        params = {'realtime_system_id': unicode(self.rt_system_id), 'message': message}
+    def record_internal_failure(self, message, comment=''):
```
pbougue: It may be a matter of taste, I'd rather like `comment=None` here, and after
```python
if comment is not None:
    params['comment'] = comment
```

source/jormungandr/jormungandr/realtime_schedule/timeo.py
```diff
@@ -181,21 +184,45 @@ def _get_next_passage_for_route_point(
...
+                resp_code = message_response['ResponseCode']
+                if 'ResponseComment' in message_response:
+                    resp_comment = message_response['ResponseComment']
+                else:
+                    resp_comment = ''
+                logging.getLogger(__name__).error(
+                    'Timeo RT internal service error, code: {} - comment: {}'.format(resp_code, resp_comment)
+                )
+                self.record_internal_failure(
+                    'internal timeo error', 'code {} - {}'.format(resp_code, resp_comment)
```
pbougue: better to build the message once for all (log, exception and newrelic) in case we update it later (I like log message more for that matters).

Also I think that exception message is traced in newrelic, so you would just need to update exception message and could remove that line : otherwise we may get duplicated messages I think.
-> You can check if we get `non 200 response` in newrelic already to know about that.

source/jormungandr/jormungandr/realtime_schedule/timeo.py
```diff
@@ -181,21 +184,45 @@ def _get_next_passage_for_route_point(
...
+                'ResponseCode' in message_response
+                and message_response['ResponseCode'] >= self.INTERNAL_TIMEO_ERROR_CODE_LIMIT
+            ):
+                resp_code = message_response['ResponseCode']
+                if 'ResponseComment' in message_response:
+                    resp_comment = message_response['ResponseComment']
+                else:
+                    resp_comment = ''
+                logging.getLogger(__name__).error(
+                    'Timeo RT internal service error, code: {} - comment: {}'.format(resp_code, resp_comment)
```
pbougue: minor to avoid confusion between HTTP code and ResponseCode:
```suggestion
                    'Timeo RT internal service error, ResponseCode: {} - ResponseComment: {}'.format(resp_code, resp_comment)
```

pbougue (commented):
Nice tests, looks OK.
The only thing left that I'm not so sure about is the newrelic part.

source/jormungandr/jormungandr/realtime_schedule/realtime_proxy.py
benoit-bst: I would to force the comment creation with an empty string to create the column into the New Relic Board. But I admit, I dunno what is the best practices :nerd_face: we'll talk about that during SU.
After your comment, the `None` become more attractive

source/jormungandr/jormungandr/realtime_schedule/timeo.py
benoit-bst: The message exeption is not traced in newrelic when we have a error code inside the response (ResponseCode field). Only the "missing id" case is traced with newrelic. So we need it.

> better to build the message once for all (log, exception and newrelic) in case we update it later (I like log message more for that matters).

Good idea, more compact for log and raiseError but not for newrelic. 
For newrelic we have 2 field, 'message' and 'comment'. this gives :
```
self.record_internal_failure('Timeo RT internal service error', 'ResponseCode: {} - ResponseComment: {}'.format(resp_code, resp_comment))
```
We have chosen to do this for distinguish the error type `Timeo intrenal error` in a column and its specificity (Code error and comment) in an other column.

pbougue (approved):

benoit-bst:
Jenkins, test this please !

PR #3002, opened 2019-10-17
[Kraken] Fix A* heuristic cost

xlqian:
https://jira.kisio.org/browse/NAVITIAII-2916

The bug is a little bit tricky to explain...

Here is why

CodeCogsEqn

When combining the current_duration and the heuristic_duration, the heuristic_duration is actually divided by factor_speed __TWICE__!! That can violate the optimality condition of A* when factor_speed is too small !!! 

The correct formula of  heuristic_duration should be:

CodeCogsEqn(1)

### DIFF
--- a/source/georef/astar_path_finder.cpp
+++ b/source/georef/astar_path_finder.cpp
@@ -66,7 +66,7 @@ void AstarPathFinder::start_distance_or_target_astar(const navitia::time_duratio
     // We start astar from source and target nodes
     try {
         astar({starting_edge[source_e], starting_edge[target_e]},
-              astar_distance_heuristic(geo_ref.graph, dest_projected, 1. / double(default_speed[mode] * speed_factor)),
+              astar_distance_heuristic(geo_ref.graph, dest_projected, 1. / double(default_speed[mode])),
               astar_distance_or_target_visitor(radius, distances, destinations));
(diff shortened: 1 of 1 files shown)


AurelienLP (approved):

xlqian:
Artemis_idfm said OK to this PR :)

benoit-bst (approved):
Whaooo, you multiply by a new factor and it doesn't change any artemis/artemis-idfm reponses ?
True story

xlqian:
@benoit-bst the reason why this factor didn't change anything in artemis is  this factor is always  set to 1 in artemis(walking_speed is always the default value). The fact that we remove this factor should lead to a better solution.

xlqian:
@benoit-bst i'll try to add some tests with non-default walking speed in artemis! thank you for the idea!

xlqian:
Ok, I find a way to test this: 
here is a journey before the fix by setting the bike_speed to **1m/s** (its default value is **4.1m/s**)
image

After the fix, with the same settings:
image

With the fix, i've got a better journey, because the excessive factor broke the optimality!!

xlqian:
test added on artemis: https://github.com/CanalTP/artemis/pull/339

pbougue (approved):

PR #3003, opened 2019-10-17
Log error code and body on geocodejson error != 200

nonifier:
We've had plenty of error with GeocodeJson with the following message:
```
 jormungandr.autocomplete.geocodejson:GeocodeJsonError: error in autocomplete request 
```
This is not ideal, as we have very little clue of what's going on.... This PR improve the error message with something like:
```
jormungandr/jormungandr/autocomplete/geocodejson.py:219: GeocodeJsonError: Autocomplete request failed with HTTP code <status_code> (<response_body>)
```

### DIFF
--- a/source/jormungandr/jormungandr/autocomplete/geocodejson.py
+++ b/source/jormungandr/jormungandr/autocomplete/geocodejson.py
@@ -213,7 +213,10 @@ class GeocodeJson(AbstractAutocomplete):
         if response.status_code == 503:
             raise GeocodeJsonUnavailable('geocodejson responded with 503')
         if response.status_code != 200:
-            raise GeocodeJsonError('error in autocomplete request')
+            error_msg = 'Autocomplete request failed with HTTP code {}'.format(response.status_code)
+            if response.text:
+                error_msg += ' ({})'.format(response.text)
+            raise GeocodeJsonError(error_msg)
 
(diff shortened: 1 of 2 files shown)


source/jormungandr/jormungandr/autocomplete/geocodejson.py
```diff
@@ -213,7 +213,10 @@ def _check_response(cls, response, uri):
         if response.status_code == 503:
             raise GeocodeJsonUnavailable('geocodejson responded with 503')
         if response.status_code != 200:
-            raise GeocodeJsonError('error in autocomplete request')
+            error_msg = 'Autocomplete request failed with HTTP code {}'.format(response.status_code)
+            if response.text:
+                error_msg += ' ({})'.format(response.text)
```
pbougue, line 218: minor: The same can be done/mutualized for the cases above

pbougue (approved):

source/jormungandr/jormungandr/autocomplete/geocodejson.py
nonifier: We already know the exact `status_code` for the other ones and I didn't wanted to pollute the error message as it's already precise enough.

benoit-bst (approved):

PR #3006, opened 2019-10-18
jormun: external_code finally works with coverage on every Uri API

benoit-bst:
**JIRA**: https://jira.kisio.org/browse/NAVP-1441

Before the fix, it was only possible to filter networks API response with **external_code**, if the coverage was not fill.
Now it works with coverage param too.
I added some Tests for networks API and external_code option
Tested in real life in local

- [ ] Waiting on Custom Artemis Build to pass to validate the PR ([custom build #111](https://ci.navitia.io/job/custom_navitia_pull_request_build/111/))

### DIFF
--- a/source/jormungandr/jormungandr/interfaces/v1/Uri.py
+++ b/source/jormungandr/jormungandr/interfaces/v1/Uri.py
@@ -163,9 +163,9 @@ class Uri(ResourceUri, ResourceUtc):
         if "odt_level" in args and args["odt_level"] != "all" and "lines" not in self.collection:
             abort(404, message="bad request: odt_level filter can only be applied to lines")
 
-        if region is None and lat is None and lon is None:
-            if "external_code" in args and args["external_code"]:
-                type_ = collections_to_resource_type[self.collection]
+        if "external_code" in args and args["external_code"]:
+            type_ = collections_to_resource_type[self.collection]
+            if region is None and lat is None and lon is None:
                 for instance in i_manager.get_regions():
                     res = i_manager.instances[instance].has_external_code(type_, args["external_code"])
                     if res:
@@ -175,7 +175,7 @@ class Uri(ResourceUri, ResourceUtc):
                 if not region:
                     abort(404, message="Unable to find an object for the uri %s" % args["external_code"])
             else:
-                abort(503, message="Not implemented yet")
+                id = i_manager.instances[region].has_external_code(type_, args["external_code"])
 
         self.region = i_manager.get_region(region, lon, lat)
 
--- a/source/jormungandr/jormungandr/interfaces/v1/Uri.py
+++ b/source/jormungandr/jormungandr/interfaces/v1/Uri.py
@@ -173,11 +173,15 @@ class Uri(ResourceUri, ResourceUtc):
                         id = res
                         break
                 if not region:
-                    abort(404, message="Unable to find an object for the uri %s" % args["external_code"])
+                    abort(
+                        404, message="Unable to find an object for the external_code %s" % args["external_code"]
+                    )
             else:
                 id = i_manager.instances[region].has_external_code(type_, args["external_code"])
                 if id == None:
-                    abort(404, message="Unable to find an object for the uri %s" % args["external_code"])
+                    abort(
+                        404, message="Unable to find an object for the external_code %s" % args["external_code"]
+                    )
 
         self.region = i_manager.get_region(region, lon, lat)
 
--- a/source/jormungandr/jormungandr/interfaces/v1/Uri.py
+++ b/source/jormungandr/jormungandr/interfaces/v1/Uri.py
@@ -176,6 +176,8 @@ class Uri(ResourceUri, ResourceUtc):
                     abort(404, message="Unable to find an object for the uri %s" % args["external_code"])
             else:
                 id = i_manager.instances[region].has_external_code(type_, args["external_code"])
+                if id == None:
+                    abort(404, message="Unable to find an object for the uri %s" % args["external_code"])
 
         self.region = i_manager.get_region(region, lon, lat)
 
--- a/source/jormungandr/jormungandr/interfaces/v1/Uri.py
+++ b/source/jormungandr/jormungandr/interfaces/v1/Uri.py
@@ -163,9 +163,9 @@ class Uri(ResourceUri, ResourceUtc):
         if "odt_level" in args and args["odt_level"] != "all" and "lines" not in self.collection:
             abort(404, message="bad request: odt_level filter can only be applied to lines")
 
-        if "external_code" in args and args["external_code"]:
+        if args.get("external_code") is not None:
             type_ = collections_to_resource_type[self.collection]
-            if region is None and lat is None and lon is None:
+            if all(obj is None for obj in (region, lat, lon)):
                 for instance in i_manager.get_regions():
                     res = i_manager.instances[instance].has_external_code(type_, args["external_code"])
                     if res:
--- a/source/jormungandr/tests/ptref_tests.py
+++ b/source/jormungandr/tests/ptref_tests.py
@@ -811,6 +811,17 @@ class TestPtRef(AbstractTestFixture):
         assert len(networks) == 1
         assert networks[0]['id'] == 'network:A'
 
+        response, code = self.query_no_assert("v1/networks?external_code=wrong_code")
+        assert code == 404
+        message = get_not_null(response, 'message')
+        assert 'Unable to find an object for the uri wrong_code' in message
+
+        # Without coverage, networks have to work with external_code
+        response, code = self.query_no_assert("v1/networks")
+        assert code == 400
+        message = get_not_null(response, 'message')
+        assert 'parameter \"external_code\" invalid: Missing required parameter' in message
+
 
 @dataset({"main_ptref_test": {}, "main_routing_test": {}})
 class TestPtRefRoutingAndPtrefCov(AbstractTestFixture):
--- a/source/jormungandr/tests/ptref_tests.py
+++ b/source/jormungandr/tests/ptref_tests.py
@@ -804,6 +804,13 @@ class TestPtRef(AbstractTestFixture):
         networks = get_not_null(response, 'networks')
         assert len(networks) == 4
 
+        # without coverage
+        response, code = self.query_no_assert("v1/networks?external_code=A")
+        assert code == 200
+        networks = get_not_null(response, 'networks')
+        assert len(networks) == 1
+        assert networks[0]['id'] == 'network:A'
+
 
 @dataset({"main_ptref_test": {}, "main_routing_test": {}})
 class TestPtRefRoutingAndPtrefCov(AbstractTestFixture):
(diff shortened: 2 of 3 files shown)


source/jormungandr/jormungandr/interfaces/v1/Uri.py
```diff
@@ -175,7 +175,7 @@ def get(self, region=None, lon=None, lat=None, uri=None, id=None):
                 if not region:
                     abort(404, message="Unable to find an object for the uri %s" % args["external_code"])
             else:
-                abort(503, message="Not implemented yet")
```
pbougue, line 178: You lost that case in the process I think ;-)
Suggested a test for that below.

source/jormungandr/jormungandr/interfaces/v1/Uri.py
```diff
@@ -163,9 +163,9 @@ def get(self, region=None, lon=None, lat=None, uri=None, id=None):
         if "odt_level" in args and args["odt_level"] != "all" and "lines" not in self.collection:
             abort(404, message="bad request: odt_level filter can only be applied to lines")
 
-        if region is None and lat is None and lon is None:
-            if "external_code" in args and args["external_code"]:
-                type_ = collections_to_resource_type[self.collection]
+        if "external_code" in args and args["external_code"]:
+            type_ = collections_to_resource_type[self.collection]
+            if region is None and lat is None and lon is None:
```
xlqian:
```suggestion
            if all((obj is None for obj in (region, lat, lon))):
```

source/jormungandr/jormungandr/interfaces/v1/Uri.py
```diff
@@ -163,9 +163,9 @@ def get(self, region=None, lon=None, lat=None, uri=None, id=None):
         if "odt_level" in args and args["odt_level"] != "all" and "lines" not in self.collection:
             abort(404, message="bad request: odt_level filter can only be applied to lines")
 
-        if region is None and lat is None and lon is None:
-            if "external_code" in args and args["external_code"]:
-                type_ = collections_to_resource_type[self.collection]
+        if "external_code" in args and args["external_code"]:
```
xlqian:
```suggestion
        if args.get(external_code) is not None:
```

source/tests/mock-kraken/main_ptref_test.cpp
```diff
@@ -139,6 +140,8 @@ struct data_set {
         b.data->pt_data->codes.add(b.lines["line:A"], "codeB", "Bise");
         b.data->pt_data->codes.add(b.lines["line:A"], "codeC", "C");
         b.data->pt_data->codes.add(b.sps.at("stop_area:stop1"), "code_uic", "bobette");
+        b.data->pt_data->codes.add(b.lines["line:A"]->network, "external_code", "A");
```
pbougue, line 144: minor : you can probably do
```suggestion
        b.data->pt_data->codes.add(b.networks["network:A"], "external_code", "A");
```

source/jormungandr/tests/ptref_tests.py
```diff
@@ -778,6 +778,39 @@ def test_pt_objects_with_disable_disruption(self):
...
+        networks = get_not_null(response, 'networks')
+        assert len(networks) == 1
+        assert networks[0]['id'] == 'network:B'
+
+        response = self.query_region("networks?external_code=wrong_code")
+        networks = get_not_null(response, 'networks')
+        assert len(networks) == 4
+
+        # without coverage
+        response, code = self.query_no_assert("v1/networks?external_code=A")
```
pbougue, line 807: Could you check that you get a 503 "Not implemented" without `external_code` param, also?

source/jormungandr/tests/ptref_tests.py
```diff
@@ -778,6 +778,39 @@ def test_pt_objects_with_disable_disruption(self):
...
+        networks = get_not_null(response, 'networks')
+        assert len(networks) == 4
+        assert networks[0]['id'] == 'network:A'
+        assert networks[1]['id'] == 'network:B'
+        assert networks[2]['id'] == 'network:CaRoule'
+        assert networks[3]['id'] == 'network:freq'
+
+    def test_networks_with_external_code(self):
+        """test networks with external_code parameter"""
+        response = self.query_region("networks?external_code=A")
```
pbougue, line 791: minor: if it's not too much work (and not already done somewhere) you can add a request on another type for that, as this param is available for any PT collection.

pbougue (commented):
Nice to finally activate it! It makes way more sense to have it with coverage (I don't know about the frequency of use though :grin: )
Nice tests too.

source/tests/mock-kraken/main_ptref_test.cpp
benoit-bst: Nope networks doesn't exist in builder

source/jormungandr/tests/ptref_tests.py
benoit-bst: Good idea :bird:

source/jormungandr/tests/ptref_tests.py
benoit-bst: `External code` is not tested well. I can find a way to put some others API test, like lines with external_code filter.

source/jormungandr/jormungandr/interfaces/v1/Uri.py
benoit-bst: Sweet

source/jormungandr/jormungandr/interfaces/v1/Uri.py
benoit-bst: all case are tested in [test_networks_with_external_code](https://github.com/CanalTP/navitia/pull/3006/files#diff-88b44ce2a7c9c473d3d31b9daf936efbR791) now

benoit-bst:
I update the main goal of this PR to others API. Effectively, The bug affects all API that use Uri.py (_networks, lines, Companies, routes, etc..._). To test in a global manner, I'll add some tests for **external_code** on others API, as mentioned above

source/jormungandr/tests/ptref_tests.py
```diff
@@ -804,6 +804,24 @@ def test_networks_with_external_code(self):
...
+        response, code = self.query_no_assert("v1/networks?external_code=wrong_code")
+        assert code == 404
+        message = get_not_null(response, 'message')
+        assert 'Unable to find an object for the uri wrong_code' in message
+
+        # Without coverage, networks have to work with external_code
+        response, code = self.query_no_assert("v1/networks")
+        assert code == 400
+        message = get_not_null(response, 'message')
+        assert 'parameter \"external_code\" invalid: Missing required parameter' in message
```
pbougue, line 822: OK, changed from the previous error code+message, but it's clearer I think.
And not breaking any integration I guess :+1:

source/jormungandr/tests/ptref_tests.py
```diff
@@ -778,6 +778,121 @@ def test_pt_objects_with_disable_disruption(self):
...
+        lines = get_not_null(response, 'lines')
+        assert len(lines) == 1
+        assert lines[0]['id'] == 'line:A'
+
+        response = self.query_region("lines?external_code=wrong_code")
+        lines = get_not_null(response, 'lines')
+        assert len(lines) == 4
+
+        # without coverage
+        response, code = self.query_no_assert("v1/lines?external_code=A")
```
pbougue, line 845: minor (same on networks): you can just query() here and remove the 200 check below.

source/jormungandr/tests/ptref_tests.py
```diff
@@ -778,6 +778,121 @@ def test_pt_objects_with_disable_disruption(self):
...
+        assert lines[3]['id'] == 'line:freq'
+
+        response = self.query_region("lines?external_code=A")
+        lines = get_not_null(response, 'lines')
+        assert len(lines) == 1
+        assert lines[0]['id'] == 'line:A'
+
+        response = self.query_region("lines?external_code=wrong_code")
+        lines = get_not_null(response, 'lines')
+        assert len(lines) == 4
```
pbougue: I don't understand that (I didn't pay attention: same for network and stop_point).
The parameter is OK and the value is not, so we should have nothing, no?
From a user point-of-view the comparison with the response without coverage doesn't make sense (and we probably should get no object here).

source/jormungandr/tests/ptref_tests.py
```diff
@@ -778,6 +778,121 @@ def test_pt_objects_with_disable_disruption(self):
...
+        response, code = self.query_no_assert("v1/networks?external_code=A")
+        assert code == 200
+        networks = get_not_null(response, 'networks')
+        assert len(networks) == 1
+        assert networks[0]['id'] == 'network:A'
+
+        response, code = self.query_no_assert("v1/networks?external_code=wrong_code")
+        assert code == 404
+        message = get_not_null(response, 'message')
+        assert 'Unable to find an object for the uri wrong_code' in message
```
mbounabi: Why the difference of behavior? 
- With a coverage, "wrong_code" returns all networks
- Without a coverage, "wrong_code" returns an error

source/jormungandr/tests/ptref_tests.py
benoit-bst: I think you've right. it should be the same behavior. Fix it fix it  :eagle:

source/jormungandr/tests/ptref_tests.py
benoit-bst: I think you've right. it should be the same behavior. Fix it fix it :eagle:

source/jormungandr/tests/ptref_tests.py
benoit-bst: **query** doesn't work without coverage, there is a check inside. Only **query_no_assert** work with it, if i want to catch the code error

benoit-bst:
@pbougue & @mbounabi, I fixed the difference for the `wrong code` case. Now, we have the same behavior for both, **404** error -> `Unable to find an object for the uri wrong_code`
Thanks for the review :heart_eyes_cat:

source/jormungandr/jormungandr/interfaces/v1/Uri.py
```diff
@@ -175,7 +175,9 @@ def get(self, region=None, lon=None, lat=None, uri=None, id=None):
                 if not region:
                     abort(404, message="Unable to find an object for the uri %s" % args["external_code"])
             else:
-                abort(503, message="Not implemented yet")
+                id = i_manager.instances[region].has_external_code(type_, args["external_code"])
+                if id == None:
+                    abort(404, message="Unable to find an object for the uri %s" % args["external_code"])
```
pbougue: Same above (sorry it becomes long, and it was there before your changes) : `uri`==`id`!=`external_code`
```suggestion
                    abort(404, message="Unable to find an object for the external_code %s" % args["external_code"])
```

benoit-bst: hoooo that was the message for no coverage case... 
fix it fix it :construction_worker_man: for both

pbougue (approved):
Sweet for API consistency, thanks for the tests and the work on that!
It should move artemis' tests on bibus I think (for the wrong external_code case) : so merge with caution after custom build :tanabata_tree:

mbounabi (approved):

mbounabi:
References to be merged with this PR:
https://github.com/CanalTP/artemis_references/pull/197

kinnou02:
This feature is about to die, it was a way to retrieve an object in navitia 2 from its identifier in navitia 1. The external code is only filed by fusio, that is slowly being deprecated.

PR #3009, opened 2019-10-22
Kraken: Refacto isochrone

AurelienLP:
Readable commit by commit!

In kraken, `make_isochrone_common` was used for graphical_isochrone and heat_maps but not for isochrone...

It changes the behavior of graphical_isochrone and heat_maps when no stop_points were found around the center. We now returns a no solution response rather than a `no_origin_nor_destination` which is used when the request does not have a `origins` or `destinations`.

This will help with isochrone_distributed.

TL;DR: 3 APIs share the same code now.


source/routing/raptor_api.cpp
```diff
@@ -1410,41 +1410,86 @@ void make_response(navitia::PbCreator& pb_creator,
...
+          bound(bound),
+          datetime(datetime) {}
+};
+
+static const boost::optional<IsochroneCommon> make_isochrone_common(
+    RAPTOR& raptor,
+    const type::EntryPoint& center,
+    const uint64_t departure_datetime,
+    const double max_duration,
+    uint32_t max_transfers,
```
```suggestion
    const uint32_t max_transfers,
```
? 
Dunno for RAPTOR& raptor and georef::StreetNetwork& worker

benoit-bst (approved):
if there a **Utest** or something like that, into the code, to prove the proper functioning ? 
I image it already exists, probably in a global way.

source/routing/raptor_api.cpp
```diff
@@ -1459,12 +1459,7 @@ static const boost::optional<IsochroneCommon> make_isochrone_common(
     if (!departures) {
         pb_creator.fill_pb_error(pbnavitia::Error::unknown_object, "The entry point: " + center.uri + " is not valid");
         return boost::optional<IsochroneCommon>{};
-    } else if (departures->empty()) {
```
pbougue: :+1: nice!

pbougue (approved):

source/routing/raptor_api.cpp
AurelienLP: Can't for RAPTOR& raptor and georef::StreetNetwork& worker :cry:

AurelienLP:
> if there a **Utest** or something like that, into the code, to prove the proper functioning ?
> I image it already exists, probably in a global way.

Yes, some Utests already exist, that is how I found the different behavior between APIs.

benoit-bst (approved):

PR #3010, opened 2019-10-22
Doc: Update to explain more data_freshness impact on collection filters

pbougue:
After questions on some requests.

Not so easy to document, as this request with `data_freshness` + `since`/`until` is the only case affected so far.
My bet is it's gonna need to be spread to other requests (filtering VJ depending on the validity_pattern if no since/until provided).
So staying a bit evasive to prepare next changes :balance_scale: 

Also:
* removes deprecated `disruption_active` param from doc
* updates the translation of `disruption_active` to match documentation > poke @kinnou02 on that


documentation/slate/source/includes/apis.md
```diff
@@ -1019,7 +1019,6 @@ See how disruptions affect a journey in the [real time](#realtime) section.
 | nop     | max_nb_transfers      | int     | Maximum number of transfers in each journey  | 10          |
 | nop     | min_nb_transfers     | int     | Minimum number of transfers in each journey  | 0           |
 | nop     | max_duration         | int     | Maximum duration of journeys in seconds (from `datetime` parameter).<br>More usefull when computing an isochrone (only `from` or `to` is provided).<br>On a classic journey (from-to), it will mostly speedup Navitia: You may have journeys a bit longer than that value (you would have to filter them).    | 86400       |
-| nop     | disruption_active    | boolean | For compatibility use only.<br>If true the algorithm take the disruptions into account, and thus avoid disrupted public transport.<br>Rq: `disruption_active=true` = `data_freshness=realtime` <br>Use `data_freshness` parameter instead       |  False     |
```
mbounabi, line 1022: I didn't even know this parameter existed... It doesn't work? Or we just don't want to document it?

documentation/slate/source/includes/realtime.md
```diff
@@ -22,6 +22,23 @@ For each one of these effects, here's how the Navitia responses will be affected
 A disruption is present in the response of the endpoints described if the request is made during its application period.
 </aside>
 
+
+## <a name="PT_object_collections_data_freshness"></a>Public transport object collections
+
+Several public transport objects have separate collections for `base_schedule` and `realtime`.
+So the data_freshness parameter may affect the number of objects returned depanding on the request.
```
mbounabi:
```suggestion
So the data_freshness parameter may affect the number of objects returned depending on the request.
```

documentation/slate/source/includes/realtime.md
```diff
@@ -22,6 +22,23 @@ For each one of these effects, here's how the Navitia responses will be affected
...
+
+For exemple when looking for a specific circulation with the collection vehicle_journey with the request like
```
mbounabi:
```suggestion
For example when looking for a specific circulation with the collection vehicle_journey using the request:
```

documentation/slate/source/includes/realtime.md
```diff
@@ -22,6 +22,23 @@ For each one of these effects, here's how the Navitia responses will be affected
...
+`http://api.navitia.io/v1/coverage/<toto>/vehicle_journeys?since=20191008T100000&until=20191008T200000&data_freshness=base_schedule`.
+A vehicle_journey circulating between since and until that is fully deleted by a disruption will
```
mbounabi: why end of line here?

documentation/slate/source/includes/apis.md
pbougue: :older_man: It was born before `data_freshness` :baby:
It works, I even changed a bit the way it does (see journey_common).
Only it's deprecated, so I removed it from doc... Want to get rid of the debt!

Maybe some people using it will wonder what it does: it sounds fair to let them the options of querying the swagger or looking at the code.

documentation/slate/source/includes/realtime.md
pbougue: To stay under 113 char while keeping it visible that the sentence does keep on :smiley: 
If you think I should change I can.

source/jormungandr/jormungandr/interfaces/v1/journey_common.py
```diff
@@ -448,7 +448,7 @@ def parse_args(self, region=None, uri=None):
 
         if args['data_freshness'] is None:
             # retrocompatibilty handling
-            args['data_freshness'] = 'adapted_schedule' if args['disruption_active'] is True else 'base_schedule'
+            args['data_freshness'] = 'realtime' if args['disruption_active'] is True else 'base_schedule'
```
kadhikari, line 451: :+1:

documentation/slate/source/includes/realtime.md
kadhikari: Minor: We can avoid two with: -- with the collection vehicle_journey with the request ...

kadhikari (approved):

mbounabi (approved):

nonifier (approved):

kinnou02 (approved):

PR #3011, opened 2019-10-22
Fix A* heuristic cost

xlqian:
Reverts CanalTP/navitia#3007

The original PR https://github.com/CanalTP/navitia/pull/3002 has broken our business tests Artemis. In order not to block the release process, we've decided to revert the original one before we could be sure of what's really going on. 

Since every failed tests have been checked and updated, we can merge this fix into dev definitively

### DIFF
--- a/source/georef/astar_path_finder.cpp
+++ b/source/georef/astar_path_finder.cpp
@@ -66,7 +66,7 @@ void AstarPathFinder::start_distance_or_target_astar(const navitia::time_duratio
     // We start astar from source and target nodes
     try {
         astar({starting_edge[source_e], starting_edge[target_e]},
-              astar_distance_heuristic(geo_ref.graph, dest_projected, 1. / double(default_speed[mode] * speed_factor)),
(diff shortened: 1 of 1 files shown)


AurelienLP (approved):

pbougue (commented):
Code OK for me :+1: 
When ready to be merged, this will need a proper title, and it would also be nice to explain why it's gonna be OK to merge the same thing this time (for external contributors).

kinnou02:
What is the reasonning behind the removal of the speed_factor?

xlqian:
@kinnou02 https://github.com/CanalTP/navitia/pull/3002

the speed_factor is added by error

pbougue (approved):
Added the `waiting` tag to prevent merge before other Artemis works.
And a bugfix must be in the changelog :wink: 
You may add the full explanation here (as it like it's the first time :smile_cat: ).

nonifier (approved):

PR #3013, opened 2019-10-23
Kraken: Add Isochrone distributed

AurelienLP:
This PR is needed : CanalTP/navitia-proto#137
Readable commit by commit!

Nowadays, when we call isochrone with distributed, we use the new_default scenario.

For distributed, we need to compute an isochrone from an `isochrone_center` and an already computed map of `<stop_point, duration_from_center_to_stop_point>`, result of a `street_network_routing_matrix`.

In this case, we don't need to compute street_network in `make_isochrone_common`, we use the `stop_points` map

### DIFF
--- a/source/routing/raptor_api.cpp
+++ b/source/routing/raptor_api.cpp
@@ -1124,6 +1124,21 @@ boost::optional<routing::map_stop_point_duration> get_stop_points(const type::En
     return result;
 }
 
+static boost::optional<routing::map_stop_point_duration> get_stop_points_if_not_already_done(
+    const navitia::type::EntryPoint& center,
+    const navitia::type::Data& data,
+    navitia::georef::StreetNetwork& worker,
+    const boost::optional<routing::map_stop_point_duration>& stop_points) {
+    boost::optional<routing::map_stop_point_duration> departures;
+    // If stop_points have already been computed, we don't need to do it here again.
+    if (stop_points) {
+        return stop_points;
+    }
+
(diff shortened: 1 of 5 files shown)


source/routing/raptor_api.cpp
```diff
@@ -1124,6 +1124,21 @@ boost::optional<routing::map_stop_point_duration> get_stop_points(const type::En
     return result;
 }
 
+static boost::optional<routing::map_stop_point_duration> get_stop_points_if_not_already_done(
+    const navitia::type::EntryPoint& center,
+    const navitia::type::Data& data,
+    navitia::georef::StreetNetwork& worker,
+    const boost::optional<routing::map_stop_point_duration>& stop_points) {
+    boost::optional<routing::map_stop_point_duration> departures;
```
pbougue: unused var, no?

pbougue (commented):
This looks OK, but I'm not sure where/how you're gonna handle the way `stop_points` are filled (from origins/destination). Same for `center`.
Is there another PR on kraken after that one?

AurelienLP:
> This looks OK, but I'm not sure where/how you're gonna handle the way `stop_points` are filled (from origins/destination). Same for `center`.
> Is there another PR on kraken after that one?

Totally !

AurelienLP:
Jenkins, test this please.

pbougue (approved):

benoit-bst (approved):

PR #3016, opened 2019-10-24
Tyr: Robustify purge and heartbeat

pbougue:
To avoid having one instance's problem stopping all following purges (and heartbeats)

Some tests are already running on purges: I did not add a specific one for this case.

JIRA: https://jira.kisio.org/browse/NAVITIAII-2932
Nota: `update_data` is already protected.


benoit-bst (approved):
great

kadhikari (approved):

PR #3019, opened 2019-10-28
/addresses collection has never been implemented.

nonifier:
:mushroom::mushroom::mushroom:
Long time ago, in 2013, was introduced `/addresses` has a place holder to prepare for an implementation. As of today, 6 years later, we are still waiting for it to be implemented......

In the meantime, `Places` and reverse geocoding were developed to fulfil that same purpose. So instead of using `/addresses` to get an address, we have to use `/coord` (`/coords`) like in :
```
/v1/coord/<lon:lon>;<lat:lat>/
/v1/coords/<lon:lon>;<lat:lat>/
/v1/coverage/
/v1/coverage/<lon:lon>;<lat:lat>/
/v1/coverage/<lon:lon>;<lat:lat>/coords
/v1/coverage/<lon:lon>;<lat:lat>/<uri:uri>/coords
/v1/coverage/<region:region>/
/v1/coverage/<region:region>/coords
/v1/coverage/<region:region>/<uri:uri>/coords
```

I suggest to drop the support of `/addresses` as an Uri (:mushroom:)

[NAVITIAII-2924](https://jira.kisio.org/browse/NAVITIAII-2924)

### DIFF
--- a/source/jormungandr/tests/end_point_tests.py
+++ b/source/jormungandr/tests/end_point_tests.py
@@ -227,3 +227,15 @@ class TestEndPoint(AbstractTestFixture):
         self.check_context(response)
         assert response.get("equipment_reports", None) is not None
         assert response['context']['timezone'] == 'UTC'
+
+    def test_region_coord_addresses(self):
+        '''
+        To get an address, you have to use the reverse API as in : /coord/<lat>;<lon>.
+        '''
+        response = self.tester.get(
+            'v1/coverage/main_routing_test/coord/0.001077974378345651;0.0005839027882705609/addresses'
+        )
+        assert response.status_code == 404, '/addresses has never been supported, and is now deprecated'
+
+        response = self.query('v1/coverage/main_routing_test/coord/0.001077974378345651;0.0005839027882705609/')
+        assert response.has_key('address')
--- a/source/jormungandr/jormungandr/interfaces/uri.py
+++ b/source/jormungandr/jormungandr/interfaces/uri.py
@@ -40,7 +40,6 @@ collections_to_resource_type = {
     "stop_areas": "stop_area",
     "lines": "line",
     "line_groups": "line_group",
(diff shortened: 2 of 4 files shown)


source/jormungandr/tests/end_point_tests.py
```diff
@@ -227,3 +227,15 @@ def test_equipment_reports_context(self):
...
+        '''
+        To get an address, you have to use the reverse API as in : /coord/<lat>;<lon>.
+        '''
+        response = self.tester.get(
+            'v1/coverage/main_routing_test/coord/0.001077974378345651;0.0005839027882705609/addresses'
+        )
+        assert response.status_code == 404, '/addresses has never been supported, and is now deprecated'
+
+        response = self.query('v1/coverage/main_routing_test/coord/0.001077974378345651;0.0005839027882705609/')
+        assert response.has_key('address')
```
kinnou02, line 241:
```suggestion
        assert 'address' in response
```
py3

kinnou02 (changes requested):
but: https://api.navitia.io/v1/coverage/fr-idf/addresses/2.5968364%3B48.5150072/? currently works, I afraid your PR will breaks it

benoit-bst:
the end point **/adresses** is not implemented so we have a ` Status: Message: Internal Server Error (500), duration of the request: 161ms`, when we call it. 
But it works without API but just **adresses** parameter... (the @kinnou02's example)
In the playground :

- :green_heart:  green param works with adresse
- :purple_heart: purple param does not work

So we have a choice to do, implement the :green_heart:  green logic for :purple_heart: purple case or erase it.
This is the same work for my JIRA ticket : https://jira.kisio.org/browse/NAVITIAII-2920

nonifier:
After @kinnou02 comment, I'll close this PR in favour of : https://github.com/CanalTP/navitia/pull/3021

PR #3020, opened 2019-10-29
Kraken: Parse center and stop points for isochrone

AurelienLP:
Readable commit by commit.

In the pbf request, we receive a `std::vector<EntryPoint>` and not a already computed `map_stop_point_duration`.

This PR aims to correct that AND to parse the `isochrone_center` and the `stop_points` from the request.

@pbougue proposition is to add a field `is_street_network_allowed_at_center` in the PBF to make this part more understandable.

### DIFF
--- a/source/kraken/worker.cpp
+++ b/source/kraken/worker.cpp
@@ -757,10 +757,17 @@ void Worker::isochrone(const pbnavitia::JourneysRequest& request) {
     if (!set_journeys_args(request, arg, "isochrone")) {
         return;
     }
-    auto const& ep = arg.origins.empty() ? arg.destinations[0] : arg.origins[0];
-    navitia::routing::make_isochrone(this->pb_creator, *planner, ep, request.datetimes(0), request.clockwise(),
+
+    // If we have a center
+    // Origins or destinations are already computed stop_points
+    const auto& center =
+        arg.isochrone_center ? *arg.isochrone_center : (arg.origins.empty() ? arg.destinations[0] : arg.origins[0]);
+    const auto& stop_points = arg.isochrone_center ? (arg.origins.empty() ? arg.destinations : arg.origins)
+                                                   : boost::optional<const std::vector<type::EntryPoint>&>{};
+
+    navitia::routing::make_isochrone(this->pb_creator, *planner, center, request.datetimes(0), request.clockwise(),
                                      arg.accessibilite_params, arg.forbidden, arg.allowed, *street_network_worker,
-                                     arg.rt_level, request.max_duration(), request.max_transfers());
+                                     arg.rt_level, request.max_duration(), request.max_transfers(), *stop_points);
 }
 
 void Worker::pt_ref(const pbnavitia::PTRefRequest& request) {
@@ -810,7 +817,12 @@ void Worker::graphical_isochrone(const pbnavitia::GraphicalIsochroneRequest& req
         return;
     }
 
-    type::EntryPoint ep = arg.origins.empty() ? arg.destinations[0] : arg.origins[0];
+    // If we have a center
+    // Origins or destinations are already computed stop_points
+    const auto& center =
+        arg.isochrone_center ? *arg.isochrone_center : (arg.origins.empty() ? arg.destinations[0] : arg.origins[0]);
+    const auto& stop_points = arg.isochrone_center ? (arg.origins.empty() ? arg.destinations : arg.origins)
+                                                   : boost::optional<const std::vector<type::EntryPoint>&>{};
     std::vector<DateTime> boundary_duration;
     for (int i = 0; i < request.boundary_duration_size(); ++i) {
         boundary_duration.push_back(request.boundary_duration(i));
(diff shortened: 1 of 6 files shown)


AurelienLP:
Jenkins, test this please.

AurelienLP:
Jenkins, test this please.

source/routing/raptor_api.cpp
```diff
@@ -1124,14 +1124,30 @@ boost::optional<routing::map_stop_point_duration> get_stop_points(const type::En
     return result;
 }
 
-static boost::optional<routing::map_stop_point_duration> get_stop_points_if_not_already_done(
+static routing::map_stop_point_duration make_map_stop_point_duration(
+    const std::vector<type::EntryPoint>& entryPointList,
```
nonifier: It would save you loads of writing by tydefing:
```cpp
using EntryPoints = std::vector<type::EntryPoint>;
```
It's also much better if we decide later on to change the container type ;)

source/kraken/worker.cpp
```diff
@@ -832,15 +844,21 @@ void Worker::heat_map(const pbnavitia::HeatMapRequest& request) {
         return;
     }
 
-    type::EntryPoint ep = arg.origins.empty() ? arg.destinations[0] : arg.origins[0];
+    // If we have a center
+    // Origins or destinations are already computed stop_points
+    const auto& center =
+        arg.isochrone_center ? *arg.isochrone_center : (arg.origins.empty() ? arg.destinations[0] : arg.origins[0]);
+    const auto& stop_points = arg.isochrone_center ? (arg.origins.empty() ? arg.destinations : arg.origins)
+                                                   : boost::optional<const std::vector<type::EntryPoint>&>{};
```
nonifier: code duplication spotted 3 times :robot: 

Also, why returning a `boost<optional>` of an empty vector ? Would `boost::none` not be better in this case ?

AurelienLP: I refactored that part in a commit but broke the CI with really ugly code, I am working on that right now.

I used a `boost<optional>` of an empty vector because auto can't resolve the type in a ternary if : `arg.origins` and `arg.destinations` are not the same type of `boost::none`.

I think of using something like `    const auto& stop_points =
        boost::make_optional(arg.isochrone_center, (arg.origins.empty() ? arg.destinations : arg.origins));`

benoit-bst (approved):

AurelienLP:
Jenkins, test this please.

source/kraken/worker.cpp
```diff
@@ -772,6 +752,31 @@ void Worker::journeys(const pbnavitia::JourneysRequest& request, pbnavitia::API
...
+static std::pair<const type::EntryPoint, const boost::optional<const type::EntryPoints&>> get_center_and_stop_points(
+    const JourneysArg& arg) {
+    // If we have a center
+    // Origins or destinations are already computed stop_points
+    const auto& center =
+        arg.isochrone_center ? *arg.isochrone_center : (arg.origins.empty() ? arg.destinations[0] : arg.origins[0]);
+    const auto& stop_points = arg.isochrone_center ? (arg.origins.empty() ? arg.destinations : arg.origins)
+                                                   : boost::optional<const type::EntryPoints&>{};
+
+    return std::make_pair(center, stop_points);
```
pbougue, line 764: I'd rather have a struct for clarity here.

source/kraken/worker.cpp
```diff
@@ -772,6 +752,31 @@ void Worker::journeys(const pbnavitia::JourneysRequest& request, pbnavitia::API
     }
 }
 
+static std::pair<const type::EntryPoint, const boost::optional<const type::EntryPoints&>> get_center_and_stop_points(
+    const JourneysArg& arg) {
+    // If we have a center
+    // Origins or destinations are already computed stop_points
```
pbougue, line 758: This is why I'd rather have a `is_street_network_allowed_at_center` parameter: here you suppose some behavior from the caller but it's implicit and undocumented (well, just here).
As the caller is quite far (python component, and boiler-platted few times in C++) I think it's better to have that param.

And it will also more clearly allow fancy uses in the future.

source/kraken/worker.cpp
```diff
@@ -772,6 +752,31 @@ void Worker::journeys(const pbnavitia::JourneysRequest& request, pbnavitia::API
...
+    const auto& center =
+        arg.isochrone_center ? *arg.isochrone_center : (arg.origins.empty() ? arg.destinations[0] : arg.origins[0]);
```
pbougue: Memory safety: Do you make sure somewhere that `args.destinations` isn't empty if you use it?
Also, better to always use `.at(0)` to fail loudly :wink:

pbougue: Also, if you keep the same behavior (not :avocado:-ing for this at all), you may fill `stop_points` first, then `center` like this:
```suggestion
        arg.isochrone_center ? *arg.isochrone_center : stop_points.at(0);
```

pbougue (commented):
I keep my opinion on the parameter, as described below.
The rest of the code is fine, doing a sweet little cleanup of that part of the code.

source/kraken/worker.cpp
AurelienLP: > Memory safety: Do you make sure somewhere that `args.destinations` isn't empty if you use it?
> Also, better to always use `.at(0)` to fail loudly wink

Yes, it is done in set_journeys_args. But I can use `at(0)` or `front()`

AurelienLP: Yes, it is done in set_journeys_args. But I can use `at(0)` or `front()`

source/kraken/worker.cpp
AurelienLP: For me a struct would be overkill. Even if the pair is not that clear at first, you know you get a center and some stop_points for the rest of the isochrone computation.

source/kraken/worker.cpp
AurelienLP: Your suggestion wouldn't work. If there is no isochrone_center, stop_points is `boost::none`.

source/kraken/worker.cpp
xlqian: i agree with that for the sake of readability

AurelienLP: I'll add that in a new PR.

source/kraken/worker.cpp
AurelienLP: I'll create a new struct `IsochronneCommonArgs` in a new PR

source/kraken/worker.cpp
AurelienLP: I've added some `front()`

xlqian (approved):

pbougue (approved):
OK, and another PR for refactors :+1:

AurelienLP:
Jenkins, test this please.

PR #3025, opened 2019-11-07
kraken: line_reports, filter excess impacts in URI filtering mode

benoit-bst:
All related information inside **JIRA** ticket : https://jira.kisio.org/browse/NAVP-1450

### DIFF
--- a/source/disruption/line_reports_api.cpp
+++ b/source/disruption/line_reports_api.cpp
@@ -31,9 +31,6 @@ www.navitia.io
 #include "line_reports_api.h"
 #include "utils/paginate.h"
 
-#include <boost/range/algorithm.hpp>
-#include <iterator>
-
 namespace bt = boost::posix_time;
 namespace nt = navitia::type;
 
@@ -104,52 +101,7 @@ struct LineReport {
         pb_creator.fill(stop_areas, report->mutable_pt_objects(), 0);
         pb_creator.fill(stop_points, report->mutable_pt_objects(), 0, with_line_sections);
     }
-
-    nt::UrisList all_uris() const {
-        nt::UrisList uris;
-        uris.insert(line->uri);
-        for (auto& network : networks) {
-            uris.insert(network->uri);
-        }
-        for (auto& route : routes) {
-            uris.insert(route->uri);
-        }
-        for (auto& stop_area : stop_areas) {
-            uris.insert(stop_area->uri);
-        }
-        for (auto& stop_point : stop_points) {
-            uris.insert(stop_point->uri);
-        }
-        return uris;
-    }
 };
-
-void filter_excess_impacts_in_uri_filtering_mode(const std::string& filter,
-                                                 navitia::PbCreator& pb_creator,
-                                                 const std::vector<LineReport>& line_reports) {
-    if (!filter.empty()) {
-        nt::UrisList line_report_uris;
-        for (auto& line_report : line_reports) {
-            auto uris = line_report.all_uris();
-            line_report_uris.insert(uris.cbegin(), uris.cend());
-        }
-
-        auto erase_allowed = [](const nt::UrisList& uris, const nt::UrisList& impacted_uris) -> bool {
-            std::vector<std::string> intersect;
-            boost::range::set_intersection(uris, impacted_uris, std::back_inserter(intersect));
-            return intersect.size() == 0;
-        };
-
-        // Clean impacts if needed
-        for (auto it = pb_creator.impacts.begin(); it != pb_creator.impacts.end();) {
-            if (erase_allowed(line_report_uris, (*it).get()->informed_entities_uris())) {
-                it = pb_creator.impacts.erase(it);
-            } else {
-                ++it;
-            }
-        }
-    }
-}
 
 void line_reports(navitia::PbCreator& pb_creator,
                   const navitia::type::Data& d,
@@ -192,8 +144,6 @@ void line_reports(navitia::PbCreator& pb_creator,
     for (const auto& line_report : paged_line_reports) {
         line_report.to_pb(pb_creator, depth);
     }
-
-    //    filter_excess_impacts_in_uri_filtering_mode(filter, pb_creator, line_reports);
 
     pb_creator.make_paginate(total_results, start_page, count, pb_creator.line_reports_size());
     if (pb_creator.line_reports_size() == 0) {
--- a/source/disruption/line_reports_api.cpp
+++ b/source/disruption/line_reports_api.cpp
@@ -31,6 +31,9 @@ www.navitia.io
 #include "line_reports_api.h"
 #include "utils/paginate.h"
 
+#include <boost/range/algorithm.hpp>
+#include <iterator>
+
 namespace bt = boost::posix_time;
 namespace nt = navitia::type;
 
@@ -132,12 +135,9 @@ void filter_excess_impacts_in_uri_filtering_mode(const std::string& filter,
         }
 
         auto erase_allowed = [](const nt::UrisList& uris, const nt::UrisList& impacted_uris) -> bool {
-            for (const auto& impacted_uri : impacted_uris) {
-                if (std::find(uris.cbegin(), uris.cend(), impacted_uri) == uris.cend()) {
-                    return true;
-                }
-            }
-            return false;
+            std::vector<std::string> intersect;
+            boost::range::set_intersection(uris, impacted_uris, std::back_inserter(intersect));
+            return intersect.size() == 0;
         };
 
         // Clean impacts if needed
--- a/source/disruption/line_reports_api.cpp
+++ b/source/disruption/line_reports_api.cpp
@@ -104,7 +104,55 @@ struct LineReport {
         pb_creator.fill(stop_areas, report->mutable_pt_objects(), 0);
         pb_creator.fill(stop_points, report->mutable_pt_objects(), 0, with_line_sections);
     }
+
+    std::set<std::string> all_uris() const {
+        std::set<std::string> uris;
+        uris.insert(line->uri);
+        for (auto& network : networks) {
+            uris.insert(network->uri);
+        }
+        for (auto& route : routes) {
+            uris.insert(route->uri);
+        }
+        for (auto& stop_area : stop_areas) {
+            uris.insert(stop_area->uri);
+        }
+        for (auto& stop_point : stop_points) {
+            uris.insert(stop_point->uri);
+        }
+        return uris;
+    }
 };
+
+void filter_excess_impacts_in_uri_filtering_mode(const std::string& filter,
+                                                 navitia::PbCreator& pb_creator,
+                                                 const std::vector<LineReport>& line_reports) {
+    if (!filter.empty()) {
+        std::set<std::string> line_report_uris;
+        for (auto& line_report : line_reports) {
+            auto uris = line_report.all_uris();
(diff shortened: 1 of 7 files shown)


benoit-bst:
Jenkins, test this please !

source/disruption/line_reports_api.h
```diff
@@ -29,8 +29,7 @@ www.navitia.io
 */
 
 #pragma once
-#include "type/type.h"
```
benoit-bst, line 32: kill the type.h :crossed_swords:

source/jormungandr/tests/disruptions_tests.py
```diff
@@ -871,3 +871,57 @@ def test_line_reports_with_since_until_value_not_valid(self):
...
+        response = self.query_region("lines/line:1/line_reports?_current_datetime=20170101T120000")
+        line_reports = get_not_null(response, 'line_reports')
+        assert len(line_reports) == 1
+        assert line_reports[0]['line']['id'] == 'line:1'
+        assert len(line_reports[0]['line']['links']) == 2
+        assert line_reports[0]['line']['links'][0]['id'] == 'line_section_on_line_1'
+        assert line_reports[0]['line']['links'][1]['id'] == 'line_section_on_line_1_other_effect'
+        # Associated disruptions
+        disruptions = get_not_null(response, 'disruptions')
+        assert len(disruptions) == 2
```
benoit-bst, line 927: Without the filtering, there is 3 disruptions...

benoit-bst:
Jenkins, test this please !

source/disruption/line_reports_api.cpp
```diff
@@ -104,8 +101,56 @@ struct LineReport {
...
+            line_report_uris.insert(uris.cbegin(), uris.cend());
+        }
+
+        auto erase_allowed = [](const nt::UrisList& uris, const nt::UrisList& impacted_uris) -> bool {
+            for (const auto& impacted_uri : impacted_uris) {
+                if (std::find(uris.cbegin(), uris.cend(), impacted_uri) == uris.cend()) {
+                    return true;
+                }
+            }
+            return false;
```
nonifier:
```cpp
auto erase_allowed = [](const nt::UrisList& uris, const nt::UrisList& impacted_uris) -> bool {
    nt::UrisList intersect;
    boost::set_intersection(uris, impacted_uris,  std::inserter(intersect));  
    return intersect.size() == 0;
}
```
This is O(N) with N is dist(uris) + dist(impacted_uris) :)

benoit-bst:
Jenkins, test this please !

source/disruption/line_reports_api.h
nonifier: kill it

source/disruption/line_reports_api.cpp
benoit-bst: delightful and stylish :apple:

source/disruption/line_reports_api.cpp
```diff
@@ -104,8 +101,56 @@ struct LineReport {
...
+        };
+
+        // Clean impacts if needed
+        for (auto it = pb_creator.impacts.begin(); it != pb_creator.impacts.end();) {
+            if (erase_allowed(line_report_uris, (*it).get()->informed_entities_uris())) {
+                it = pb_creator.impacts.erase(it);
+            } else {
+                ++it;
+            }
+        }
```
nonifier: If a protobuf repeated array is compliant with a `ForwardIterator`, you could just write:

```cpp
boost::remove_if(pb_creator.impacts, [&](const auto& impact){
    return erase_allowed(line_report_uris, impact.get()->informed_entities_uris());
});
```

benoit-bst: Good idea, but only the old way works with std::set... :cry: 
https://stackoverflow.com/questions/24263259/c-stdseterase-with-stdremove-if

benoit-bst:
Jenkins, test this please !

source/disruption/line_reports_api.cpp
benoit-bst: I had to change the result container `Urislist -> vector`, because std::set doesn't have push_back method for the std::inserter.Tthere is probably a method to do that but std::vector do the job.

nonifier (dismissed):

source/jormungandr/tests/disruptions_tests.py
```diff
@@ -871,3 +871,57 @@ def test_line_reports_with_since_until_value_not_valid(self):
...
+    def test_line_reports_with_uri_filter(self):
+        # Without filter
+        response = self.query_region("line_reports?_current_datetime=20170101T120000")
+        line_reports = get_not_null(response, 'line_reports')
+        assert len(line_reports) == 2
+        # line:1
+        assert line_reports[0]['line']['id'] == 'line:1'
+        assert len(line_reports[0]['line']['links']) == 2
+        assert line_reports[0]['line']['links'][0]['id'] == 'line_section_on_line_1'
+        assert line_reports[0]['line']['links'][1]['id'] == 'line_section_on_line_1_other_effect'
```
pbougue: We should also check for the `"pt_objects"` of the line_report without a filter:
```py
for o in line_reports[0]['pt_objects']:
    links = o[o['embedded_type']]['links']
    assert links
    for l in links:
        assert l['id'] != line_section_on_line_2
```
This is the core of the problem, as line_reports works as follow:
* from a line, we retrieve all stop_points (among other).
* then from those stop_points we retrieve all disruptions (blind to the line now). And here we get all line_sections of the stop_points, even if they are not linked to the initial line > we should have a filter here that also uses the line, so the embedding stop_point is also removed.

source/jormungandr/tests/disruptions_tests.py
```diff
@@ -871,3 +871,57 @@ def test_line_reports_with_since_until_value_not_valid(self):
...
+        assert disruptions[2]['id'] == 'line_section_on_line_2'
+
+        # With lines=line:1 filter
+        response = self.query_region("lines/line:1/line_reports?_current_datetime=20170101T120000")
+        line_reports = get_not_null(response, 'line_reports')
+        assert len(line_reports) == 1
+        assert line_reports[0]['line']['id'] == 'line:1'
+        assert len(line_reports[0]['line']['links']) == 2
+        assert line_reports[0]['line']['links'][0]['id'] == 'line_section_on_line_1'
+        assert line_reports[0]['line']['links'][1]['id'] == 'line_section_on_line_1_other_effect'
```
pbougue, line 919: Same, we should also check that there is no dead link to the disruption in pt_objects.

kadhikari (dismissed):

pbougue (changes requested):
More generally:
In principle, for any endpoint `/totos`, navitia provides a core-collection `"totos"` in the response which is "self-generated".
Side-collections are generated "from" the core-collection.

Side-collections are there to avoid duplicates and lighten the response.
So for any side-collection there are links somewhere in the core-collection `"totos"` to all objects of the side-collection.

Nota: there are some special contextual collections that are not linked like `errors`, `warnings`, `feed_publishers`, `pagination`, `links` and `context`.

> So if a side-collection has issues, the issue should be delt-with at generation time. Dealing with the final side-collection only will most probably lead to some dirty leftovers (links to that side-collection, and maybe also objects embedding links).

source/jormungandr/tests/disruptions_tests.py
pbougue: Updated the suggested test a bit, to check that we also don't have pt_objects without links.

benoit-bst: Yes, it is true :+1: 

We didn't see the dead link inside `line_reports->pt_object->embedded_type->links->associated_disruptions`.
for my part, I only check `line_reports->lines->links->associated_disruptions`.
I'll add this check inside the test, and the PR will become red because of it.

Nevertheless, it not change the fact, that I don't have, for the moment, enough legacy knowledge on this part. As I said into the ticket and during grooming, If you detect some part of code that is culprit, fell free to propose a change. I suppose we will talk all together about that next time.

source/jormungandr/tests/disruptions_tests.py
```diff
@@ -758,38 +758,99 @@ def test_disruption_with_stop_areas_and_different_parameters(self):
...
+        assert disruption['id'] == 'line_section_on_line_1'
+        assert len(disruption['impacted_objects']) == 1
+        impacted_object = disruption['impacted_objects']
+        assert impacted_object[0]['pt_object']['id'] == 'line:1'
+        # disruption : line_section_on_line_1_other_effect
+        disruption = disruptions[1]
+        assert disruption['id'] == 'line_section_on_line_1_other_effect'
+        assert len(disruption['impacted_objects']) == 2  # E -> F and F -> E
+        impacted_object = disruption['impacted_objects']
+        assert impacted_object[0]['pt_object']['id'] == 'line:1'
```
pbougue, line 844: Minor: we are more interested in `impacted_section` part here, I'd say.
(same on all impacted_objects)

source/jormungandr/tests/disruptions_tests.py
```diff
@@ -758,38 +758,99 @@ def test_disruption_with_stop_areas_and_different_parameters(self):
 
 @dataset({"line_sections_test": {}})
 class TestDisruptionsLineSections(AbstractTestFixture):
+    def _check_response(self, response):
```
pbougue: minor: not fan of the too broad method's name (risk of shadowing or just misunderstanding)

source/jormungandr/tests/disruptions_tests.py
```diff
@@ -758,38 +758,99 @@ def test_disruption_with_stop_areas_and_different_parameters(self):
...
+        assert line['links'][0]['id'] == 'line_section_on_line_1'
+        assert line['links'][1]['id'] == 'line_section_on_line_1_other_effect'
+        # line:1 associated pt_objects
+        pt_objects = line_report['pt_objects']
+        assert len(pt_objects) == 5
+        # C_1
+        assert pt_objects[0]['id'] == 'C_1'
+        assert len(pt_objects[0][pt_objects[0]['embedded_type']]['links']) == 1
+        link = pt_objects[0][pt_objects[0]['embedded_type']]['links'][0]
+        assert link['id'] == 'line_section_on_line_1'
```
pbougue: minor: shorter for 3 lines (can be done everywhere)
```python
assert [l['id'] for l in pt_objects[0][pt_objects[0]['embedded_type']]['links']] == ['line_section_on_line_1']
```
You may also use a set if you don't want to care about sorting.

pbougue (commented):
clearer and more thorough testing :+1:

source/jormungandr/tests/disruptions_tests.py
benoit-bst: I can add some check around that

source/jormungandr/tests/disruptions_tests.py
benoit-bst: Yep, we can find a function name more consistent like `_check_line_reports_response` ?

source/jormungandr/tests/disruptions_tests.py
benoit-bst: nice :+1:

benoit-bst:
**Summary**

- Upstream filtering the excess disruptions is **removed**
- Add new logic into **pb_converter** for avoiding to link wrong disruption
- **Extend tests** with all links checks

pbougue (approved):

kadhikari (approved):

PR #3031, opened 2019-11-12
Jormun: Refacto planner for future graphical isochrone implementations

AurelienLP:
In distributed, `planner.py` is responsible of creating the pbf request and send it to the street_network_backend used.

This PR focus on multiple things:

1) Implementing `_create_graphical_isochrones_request`. To do so I've refactored the code responsible of creating a `journeys_request`. In the pfb, a `GraphicalIsochroneRequests` contains a `JourneysRequest`, so I've followed the same logic and created a `GraphicalIsochronesParameters` containing a `JourneyParameters`.

2) Adding some basic Unit Tests.

3) Replacing some hard coded values in `planner.py` by declared values in `default_values.py`.

4) Moving an abort during the process of a graphical_isochrone_request for new_default which should have been done much earlier in the code.

### DIFF
--- a/source/jormungandr/jormungandr/planner.py
+++ b/source/jormungandr/jormungandr/planner.py
@@ -18,24 +18,24 @@
 # www.navitia.io
 from __future__ import absolute_import, print_function, unicode_literals, division
 from jormungandr import utils
-from navitiacommon import request_pb2, type_pb2
+from navitiacommon import request_pb2, type_pb2, default_values
 
 
 class JourneyParameters(object):
     def __init__(
         self,
-        max_duration=86400,
+        max_duration=default_values.max_duration,
         max_transfers=10,
         wheelchair=False,
         forbidden_uris=None,
         allowed_id=None,
         realtime_level='base_schedule',
-        max_extra_second_pass=None,
-        walking_transfer_penalty=120,
+        max_extra_second_pass=default_values.max_extra_second_pass,
+        walking_transfer_penalty=default_values.walking_transfer_penalty,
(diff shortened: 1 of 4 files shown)


source/jormungandr/jormungandr/planner.py
```diff
@@ -18,24 +18,24 @@
...
-from navitiacommon import request_pb2, type_pb2
+from navitiacommon import request_pb2, type_pb2, default_values
 
 
 class JourneyParameters(object):
     def __init__(
         self,
-        max_duration=86400,
+        max_duration=default_values.max_duration,
         max_transfers=10,
```
benoit-bst, line 28: there is no default parameter for max_transfers ?

source/jormungandr/jormungandr/planner.py
```diff
@@ -57,11 +57,20 @@ def __init__(
         self.isochrone_center = isochrone_center
 
 
+class GraphicalIsochronesParameters(object):
```
benoit-bst, line 60: it is only used in test ? If this is the case, maybe you should move it inside test file directly ?

AurelienLP: This code will be called later!

source/jormungandr/jormungandr/planner.py
AurelienLP: No, I didn't find one :disappointed:

benoit-bst (approved):

xlqian (approved):

PR #3032, opened 2019-11-13
Kraken: Case/accents sensitive sort on stop_areas

BaptisteB-A:
See the [Jira ticket](https://jira.kisio.org/browse/NAVP-1446).

- [x] Needs [this PR](https://github.com/CanalTP/utils/pull/85)

The stop_area list order makes more sense as following :

- special characters
- numbers
- alphabetical order : accents and uppercase letters are considered as their lowercase equivalent letters when sorting (i.e. "É" -> "e")

Added a test.
Moved `strip_accents` and `strip_accents_and_lower` functions to source/utils as well as their test.


source/type/stop_area.cpp
```diff
@@ -46,8 +46,12 @@ void StopArea::serialize(Archive& ar, const unsigned int) {
 SERIALIZABLE(StopArea)
 
 bool StopArea::operator<(const StopArea& other) const {
-    if (name != other.name) {
-        return name < other.name;
+    std::string lower_name(name);
+    lower_name = strip_accents_and_lower(lower_name);
```
nonifier: It seems that `strip_accents` works on a copy of the input string. This means you could simply write:

```
std::string lower_name = strip_accents_and_lower(name);
std::string lower_other_name = strip_accents_and_lower(other.name);
```

source/type/tests/test.cpp
```diff
@@ -141,6 +144,47 @@ BOOST_AUTO_TEST_CASE(projection) {
...
+    return sa;
+}
+
+bool operator==(const navitia::type::StopArea& sa, const std::string& sa_name) {
+    return sa.name == sa_name;
+}
+
+bool operator!=(const navitia::type::StopArea& sa, const std::string& sa_name) {
+    return !(sa.name == sa_name);
+}
```
nonifier: Do you actually need those two comparators anymore ?

source/type/tests/test.cpp
```diff
@@ -141,6 +144,47 @@ BOOST_AUTO_TEST_CASE(projection) {
...
+
+BOOST_AUTO_TEST_CASE(human_sort_on_stop_areas_should_not_throw) {
+    std::vector<std::shared_ptr<StopArea>> stop_areas {
```
nonifier: To wrap up with the `unique_ptr` story :D

Turns out, when initialising a `std::vector` of with an `initializer_list`, a **copy** of the initializer list is performed at the vector's constructor level : [libstdc++/stl_vector.h](https://github.com/gcc-mirror/gcc/blob/gcc-9_2_0-release/libstdc++-v3/include/bits/stl_vector.h#L611-L628).... And copying a `std::unique_ptr` is forbidden :no_good_man: :no_entry_sign: :no_good_man: 

The way to do it with `unique_ptr` would have been to use a `std::array` like in - https://godbolt.org/z/usFACF

nonifier (approved):

source/type/tests/test.cpp
BaptisteB-A: Alright! Should I do it that way with `unique_ptr` using `std::array` or leave it as it is right now ?

source/type/stop_area.cpp
```diff
@@ -46,8 +46,12 @@ void StopArea::serialize(Archive& ar, const unsigned int) {
 SERIALIZABLE(StopArea)
 
 bool StopArea::operator<(const StopArea& other) const {
-    if (name != other.name) {
-        return name < other.name;
+    std::string lower_name(name);
```
benoit-bst: why this is not already in lower mode ? 
Maybe during the construction we may put name var in lower directly ?

benoit-bst (approved):
nice shot

source/type/tests/test.cpp
nonifier: You can leave it as it it. It's just for our education ;)

BaptisteB-A: Thanks for the heads up :)

benoit-bst:
Jenkins, test this please !

pbougue (commented):
This looks good.

Thoughts:
* How about perfs (as recently we had troubles using `to_lower` from boost)?
* This should be done on any pt_objects to have sorting as consistent as possible throughout Navitia
* A little bit of integration tests on a real call would be nice if it's not too heavy to do.

BaptisteB-A:
About the sort on the rest of pt_objects, I plan to do it in another PR.

nonifier:
About the perf, we need to measure it before eventually fixing it.  
For the test, we already have unit tests that cover the actual changes, adding a integration tests is redundant.

PR #3034, opened 2019-11-13
Tyr: Rename 'async' as it's reserved from python 3.7

pbougue:
It's messing with ast in precommit
And messing with future Tyr ;)

No use of the param detected in fabric.


mbounabi (approved):

benoit-bst (approved):
Okay, so the parameter already exits

PR #3039, opened 2019-11-19
[Kraken] set nb_nearest_vertices to -1

xlqian:
This little is due to these tickets: https://jira.kisio.org/browse/NAVITIAII-2950 https://jira.kisio.org/browse/NAVITIAII-2956

The projection was wrong because this number is big enough. In order not to impact the performance, we use 50.

### DIFF
--- a/source/georef/georef.cpp
+++ b/source/georef/georef.cpp
@@ -702,8 +702,7 @@ edge_t GeoRef::nearest_edge(const type::GeographicalCoord& coordinates,
     // The number indicates the number of nearest vertices that should be returned by find_with
     // This number is determined by balancing the performance and the practical results (Artemis)
     // The bigger the number is, the better the projection will be and slower it will run.
-    // With 30, we have broken less than 1% tests on Artemis_idfm.
-    constexpr int nb_nearest_vertices = 30;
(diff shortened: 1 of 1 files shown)


kinnou02 (commented):
So next time if we have a longer street we set it to `100`?

xlqian:
@kinnou02 I set this magic number to -1 so that flann will find all nodes inside 500 meters

for walking and bike mode, the change didn't impact a lot the performance during my benchmark. But it did impact like 10% for car mode

AurelienLP (approved):

xlqian:
Jenkins, test this please!

xlqian:
Artemis OK!
https://ci.navitia.io/job/custom_Artemis/13/

kinnou02 (approved):
should be good

pbougue (commented):
:warning: Looks like your artemis' test was without IdFM, so I think it's worth launching for IdFM too.

This is a fine change, and the trade-off on perfs vs. quality is OK :+1:

pbougue (approved):

xlqian:
Jenkins, test this please!

xlqian:
Jenkins, test this please!

PR #3042, opened 2019-11-21
[Jormun] GeoJson geometry to be configured to a car park provider

nonifier:
A GeoJson shape can now be configured for a car park provider. In our case,
a polygon (even with holes) can be used.

A POI will have its coordinates check against that geometric shape to
define if it belongs to the provider or not.

The shape can be of any geometric object supported by GeoJson like rectangle, polygon, multipolygon etc... as long as a Point (the poi's coordinate) could be defined as 'inside' or 'outside' of the say shape.

http://geojson.io 
https://fr.wikipedia.org/wiki/GeoJSON
https://jira.kisio.org/browse/NAVP-1445

### DIFF
--- a/source/jormungandr/jormungandr/parking_space_availability/abstract_parking_places_provider.py
+++ b/source/jormungandr/jormungandr/parking_space_availability/abstract_parking_places_provider.py
@@ -61,22 +61,34 @@ class AbstractParkingPlacesProvider(six.with_metaclass(ABCMeta, object)):
                 self.log.error("Unable to parse geometry object : ", boundary_geometry)
 
     def has_boundary_shape(self):  # type () : bool
-        return self.boundary_shape != None
+        return hasattr(self, 'boundary_shape') and self.boundary_shape != None
 
-    def is_coord_within_boundary_shape(self, lon, lat):
-        # type (float, float) -> bool
-
+    def is_poi_coords_within_shape(self, poi):
         if self.has_boundary_shape() == False:
-            return False
+            '''
+            We assume that a POI is within a provider with no shape to be backward compatible.
+            So that we don't have to configure a shape for every single provider.
+            '''
+            return True
 
         try:
+            coord = poi['coord']
+            lon = float(coord['lon'])
+            lat = float(coord['lat'])
             return self.boundary_shape.contains(shapely.geometry.Point([lon, lat]))
+        except KeyError as e:
+            self.log.error(
+                "Coords illformed, 'poi' needs a coord dict with 'lon' and 'lat' attributes': ", str(e)
+            )
         except ValueError as e:
             self.log.error("Cannot convert POI's coord to float : ", str(e))
         except Exception as e:
-            self.log.error("Couldn't find if coordinate is within the shape: ", str(e))
+            self.log.error("Cannot find if coordinate is within shape: ", str(e))
 
         return False
+
+    def handle_poi(self, poi):
+        return self.support_poi(poi) and self.is_poi_coords_within_shape(poi)
 
     @abstractmethod
     def support_poi(self, poi):
--- a/source/jormungandr/jormungandr/parking_space_availability/abstract_parking_places_provider.py
+++ b/source/jormungandr/jormungandr/parking_space_availability/abstract_parking_places_provider.py
@@ -29,12 +29,54 @@
 from __future__ import absolute_import, print_function, unicode_literals, division
 from abc import abstractmethod, ABCMeta
 import six
+import logging
+import shapely
 
 
 class AbstractParkingPlacesProvider(six.with_metaclass(ABCMeta, object)):
     """
     abstract class managing calls to external service providing real-time parking places info
     """
+
+    def __init__(self, **kwargs):
+        self.boundary_shape = None
+        self.log = logging.getLogger(__name__)
+        self._init_boundary_shape(kwargs.get('geometry'))
+
+    '''
+    Initialise the object's shape using a GeoJson object.
+
+    boundary_geometry : a dictionary representing a GeoJSON object. see https://geojson.org/
+    '''
+
+    def _init_boundary_shape(self, boundary_geometry):
+        if boundary_geometry:
+            try:
+                boundary_shape = shapely.geometry.shape(boundary_geometry)
+                if not boundary_shape.is_valid:
+                    raise Exception("Geometry shape is invalid")
+                self.boundary_shape = boundary_shape
+            except Exception as e:
+                self.log.error('Error while loading boundary shape :', str(e))
+                self.log.error("Unable to parse geometry object : ", boundary_geometry)
+
+    def has_boundary_shape(self):  # type () : bool
+        return self.boundary_shape != None
+
+    def is_coord_within_boundary_shape(self, lon, lat):
+        # type (float, float) -> bool
+
+        if self.has_boundary_shape() == False:
+            return False
+
+        try:
+            return self.boundary_shape.contains(shapely.geometry.Point([lon, lat]))
+        except ValueError as e:
+            self.log.error("Cannot convert POI's coord to float : ", str(e))
+        except Exception as e:
+            self.log.error("Couldn't find if coordinate is within the shape: ", str(e))
+
+        return False
 
     @abstractmethod
     def support_poi(self, poi):
--- a/source/jormungandr/jormungandr/parking_space_availability/car/common_car_park_provider.py
+++ b/source/jormungandr/jormungandr/parking_space_availability/car/common_car_park_provider.py
@@ -43,8 +43,7 @@ from abc import abstractmethod
 
 class CommonCarParkProvider(AbstractParkingPlacesProvider):
     def __init__(self, url, operators, dataset, timeout, feed_publisher, **kwargs):
-
-        AbstractParkingPlacesProvider.__init__(self, **kwargs)
+        super(CommonCarParkProvider, self).__init__(**kwargs)
         self.ws_service_template = url + '?dataset={}'
         self.operators = [o.lower() for o in operators]
         self.timeout = timeout
--- a/source/jormungandr/jormungandr/parking_space_availability/car/common_car_park_provider.py
+++ b/source/jormungandr/jormungandr/parking_space_availability/car/common_car_park_provider.py
@@ -33,7 +33,6 @@ from __future__ import absolute_import, print_function, unicode_literals, divisi
 import logging
 import pybreaker
 import requests as requests
-import shapely
 
 from jormungandr import cache, app, new_relic
 from jormungandr.parking_space_availability import AbstractParkingPlacesProvider
@@ -45,6 +44,7 @@ from abc import abstractmethod
 class CommonCarParkProvider(AbstractParkingPlacesProvider):
     def __init__(self, url, operators, dataset, timeout, feed_publisher, **kwargs):
 
+        AbstractParkingPlacesProvider.__init__(self, **kwargs)
         self.ws_service_template = url + '?dataset={}'
         self.operators = [o.lower() for o in operators]
         self.timeout = timeout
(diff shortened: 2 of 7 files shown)


source/jormungandr/jormungandr/parking_space_availability/car/common_car_park_provider.py
```diff
@@ -73,9 +83,32 @@ def get_informations(self, poi):
...
-    def support_poi(self, poi):
+    def _is_poi_coords_within_shape(self, poi):
+        if self.shape is None:
+            return True
+
+        try:
+            coord = poi['coord']
+            coords = [float(coord['lon']), float(coord['lat'])]
+            return self.shape.contains(shapely.geometry.Point(coords))
+        except KeyError as e:
```
benoit-bst: Maybe adding an ` except ValueError` if members are not float type

source/jormungandr/jormungandr/parking_space_availability/car/tests/common_car_park_provider_test.py
```diff
@@ -0,0 +1,124 @@
...
+    assert carParkProvider.support_poi({}) == False
+    assert carParkProvider.support_poi({'properties': {}}) == False
+    assert carParkProvider.support_poi({'properties': {'operator': 'other_operator'}}) == False
+
+
+@pytest.fixture
+def carParkProviderWithGeometry(defaultCarParkParameters):
+    params = defaultCarParkParameters
+    params['geometry'] = {
+        'type': 'Polygon',
```
benoit-bst, line 79: I'm curious, an other type like `rectangle` can work well ?

source/jormungandr/jormungandr/parking_space_availability/car/common_car_park_provider.py
```diff
@@ -60,6 +61,15 @@ def __init__(self, url, operators, dataset, timeout, feed_publisher, **kwargs):
 
         self.api_key = kwargs.get('api_key')
 
+        self.shape = None
+        arg_geometry = kwargs.get('geometry')
```
benoit-bst: Maybe a solid name like `car_park_matching_geometry` ?

benoit-bst (commented):
The logic is good for me.
May be adding an extra integration test can be great. Up to you

source/jormungandr/jormungandr/parking_space_availability/car/common_car_park_provider.py
```diff
@@ -60,6 +61,15 @@ def __init__(self, url, operators, dataset, timeout, feed_publisher, **kwargs):
...
+        if arg_geometry:
+            try:
+                self.shape = shapely.geometry.shape(arg_geometry)
+            except Exception as e:
+                self.log.error("Enable to parse geometry object from : {}".format(arg_geometry))
```
mbounabi: Does this print something readable?

source/jormungandr/jormungandr/parking_space_availability/car/common_car_park_provider.py
```diff
@@ -73,9 +83,32 @@ def get_informations(self, poi):
...
+        except Exception as e:
+            self.log.error("Cannot find if coords is within shape: {}".format(str(e)))
+
+        return False
+
+    def _has_supported_poi_operator(self, poi):
         properties = poi.get('properties', {})
-        return properties.get('operator', '').lower() in self.operators
+        operator = properties.get('operator', '').lower()
+        return operator in self.operators
```
mbounabi, line 79: Why separate it in 2 lines?

mbounabi (approved):

source/jormungandr/jormungandr/parking_space_availability/car/common_car_park_provider.py
pbougue:
```suggestion
                self.log.error("Unable to parse geometry object from : {}".format(arg_geometry))
```

source/jormungandr/jormungandr/parking_space_availability/car/common_car_park_provider.py
```diff
@@ -73,9 +83,32 @@ def get_informations(self, poi):
         if data:
             return self.process_data(data, poi)
 
-    def support_poi(self, poi):
+    def _is_poi_coords_within_shape(self, poi):
```
pbougue: Not sure: this logic may be shared with BSS also, so moved one class up.

pbougue (commented):
OK for the logic and the interface.

source/jormungandr/jormungandr/parking_space_availability/car/common_car_park_provider.py
```diff
@@ -60,6 +61,15 @@ def __init__(self, url, operators, dataset, timeout, feed_publisher, **kwargs):
 
         self.api_key = kwargs.get('api_key')
 
+        self.shape = None
+        arg_geometry = kwargs.get('geometry')
+        if arg_geometry:
+            try:
+                self.shape = shapely.geometry.shape(arg_geometry)
```
kinnou02: You have to check the validity of the shape: https://shapely.readthedocs.io/en/latest/manual.html#object.is_valid

see https://github.com/CanalTP/navitia/pull/2658

source/jormungandr/jormungandr/parking_space_availability/car/common_car_park_provider.py
```diff
@@ -73,9 +83,32 @@ def get_informations(self, poi):
...
+            coords = [float(coord['lon']), float(coord['lat'])]
+            return self.shape.contains(shapely.geometry.Point(coords))
+        except KeyError as e:
+            self.log.error(
+                "Coords illformed, 'poi' needs a coords dict with 'lon' and 'lat' attributes' : {}".format(
+                    str(e)
+                )
+            )
+        except Exception as e:
+            self.log.error("Cannot find if coords is within shape: {}".format(str(e)))
```
kinnou02: logging should use logging formatting not `format`

kinnou02 (changes requested):

source/jormungandr/jormungandr/parking_space_availability/car/common_car_park_provider.py
nonifier: This will be caught further down within the generic `except Exception as e:`

source/jormungandr/jormungandr/parking_space_availability/car/tests/common_car_park_provider_test.py
nonifier: There isn't such type as 'rectangle' unfortunately. The GeoJson way of declaring a rectangle is to use a 4 points polygon :)

source/jormungandr/jormungandr/parking_space_availability/car/common_car_park_provider.py
nonifier: Well, the formating might not be amazing, especially if your shape is big. But at least you have an idea that you error comes from the shape you're passing in.
I'm open for suggestion for a more friendly user error message :)

source/jormungandr/jormungandr/parking_space_availability/car/common_car_park_provider.py
nonifier: I did that to ease readability for the same as for the above line. Otherwise, I would have ended-up with something like:

```python
        return poi.get('properties', {}).get('operator', '').lower() in self.operators
```

source/jormungandr/jormungandr/parking_space_availability/car/common_car_park_provider.py
nonifier: Moved one class up !

source/jormungandr/jormungandr/parking_space_availability/car/common_car_park_provider.py
nonifier: :+1:

source/jormungandr/jormungandr/parking_space_availability/car/common_car_park_provider.py
```diff
@@ -87,13 +75,19 @@ def get_informations(self, poi):
...
 
     def _is_poi_coords_within_shape(self, poi):
-        if self.boundary_shape is None:
+
+        if self.has_boundary_shape() == False:
+            '''
+            We assume that a POI is within a provider with no shape to be backward compatible.
+            So that we don't have to configure a shape for every single provier.
+            '''
             return True
```
pbougue: minor : you can move that no-filter logic up one class too.
I think the default for any class would be no filtering if no shape is provided.
(let you resolve this conversation once you chose)

source/jormungandr/jormungandr/parking_space_availability/car/common_car_park_provider.py
```diff
@@ -44,6 +44,7 @@
 class CommonCarParkProvider(AbstractParkingPlacesProvider):
     def __init__(self, url, operators, dataset, timeout, feed_publisher, **kwargs):
 
+        AbstractParkingPlacesProvider.__init__(self, **kwargs)
```
pbougue: For python2 (:disappointed:) it's more:
```suggestion
        super(CommonCarParkProvider, self).__init__(self, **kwargs)
```

source/jormungandr/jormungandr/parking_space_availability/abstract_parking_places_provider.py
```diff
@@ -29,13 +29,55 @@
...
+    def is_coord_within_boundary_shape(self, lon, lat):
+        # type (float, float) -> bool
+
+        if self.has_boundary_shape() == False:
+            return False
+
+        try:
+            return self.boundary_shape.contains(shapely.geometry.Point([lon, lat]))
+        except ValueError as e:
+            self.log.error("Cannot convert POI's coord to float : ", str(e))
```
pbougue, line 84: minor: technically here it's not mandatory that it is POI's coords.
```suggestion
            self.log.error("Cannot convert coord to float : ", str(e))
```

source/jormungandr/jormungandr/parking_space_availability/car/common_car_park_provider.py
pbougue: minor: you may also move this method one class up, up to you.
(let you resolve this conversation once you chose)

source/jormungandr/tests/tests_mechanism.py
```diff
@@ -507,6 +507,7 @@ def deco(cls):
 
 class MockBssProvider(AbstractParkingPlacesProvider):
     def __init__(self, pois_supported, name='mock bss provider'):
+        AbstractParkingPlacesProvider.__init__(self)
```
pbougue: Don't know if this is mandatory, but this is cleaner.
However, this should be done on **all** class inheriting from AbstractParkingPlacesProvider :wink: 

Also same : use `super`

source/jormungandr/jormungandr/parking_space_availability/car/tests/common_car_park_provider_test.py
```diff
@@ -0,0 +1,159 @@
...
+@pytest.fixture
+def carParkProvider(defaultCarParkParameters):
+    return CommonCarParkProviderTest(**defaultCarParkParameters)
+
+
+def test_create_default_common_car_park_provider_with_no_shape(carParkProvider):
+    assert carParkProvider.has_boundary_shape() == False
+
+
+def test_default_car_park_provider_should_handle_poi(carParkProvider):
```
pbougue: Not sure if the naming or the test should be changed :
```suggestion
def test_default_car_park_provider_should_support_poi(carParkProvider):
```

source/jormungandr/jormungandr/parking_space_availability/abstract_parking_places_provider.py
```diff
@@ -29,13 +29,67 @@
...
+            try:
+                boundary_shape = shapely.geometry.shape(boundary_geometry)
+                if not boundary_shape.is_valid:
+                    raise Exception("Geometry shape is invalid")
+                self.boundary_shape = boundary_shape
+            except Exception as e:
+                self.log.error('Error while loading boundary shape :', str(e))
+                self.log.error("Unable to parse geometry object : ", boundary_geometry)
+
+    def has_boundary_shape(self):  # type () : bool
```
pbougue, line 63: Looks like this method is not used outside of tests.
You may also use it (and move the `getattr` part into it) ?

pbougue (approved):
Let you decide about those 2 little remarks (and test orally mentioned), then OK to merge.

PR #3044, opened 2019-11-22
Jormun: Check that the origin and destination of journeys are not the same

AurelienLP:
https://jira.kisio.org/browse/NAVITIAII-2944

`if not (args['destination'] or args['origin'])` is check earlier because it can.

We check `if api == 'journeys' and args['origin'] == args['destination']` and send a message error.

The tests `test_no_origin_or_destination_point_in_request` and `test_same_origin_and_destination_point_in_request` are added.

The test `test_journeys_do_not_loose_precision` is change to use two different coords for `from` and `to`.


source/jormungandr/tests/journey_common_tests.py
```diff
@@ -346,24 +346,6 @@ def test_journeys_date_valid_not_zeropadded(self):
         assert 'message' in response
         assert "Unable to parse datetime, year" in response['message']
 
-    def test_journeys_do_not_loose_precision(self):
```
benoit-bst, line 349: why is it deleted ?
This test is used to check different precisions

source/jormungandr/tests/journey_common_tests.py
```diff
@@ -346,24 +346,6 @@ def test_journeys_date_valid_not_zeropadded(self):
...
-        """do we have a good precision given back in the id"""
```
mbounabi, line 350: Why was this test added when it was? Because it was clearly added for a purpose?

benoit-bst (approved):

benoit-bst:
Jenkins, test this please !

benoit-bst:
Jenkins, test this please !

AurelienLP:
Jenkins, test this please.

AurelienLP:
Jenkins, test this please.

pbougue (commented):
We still have the issue if there is 1m of distance between from and to.
IMHO this is managing in code a specific case that should properly be managed in a more generic way. If we merge that, the next ticket we will probably get on that subject will be about the 1m journey, and when we will manage it, this code/behavior will have to be maintained for retrocompat (worst case) or cleaned.

So I think we should use this case as a way to improve many more and not "hide" it in between (unless there is an urgent need on that).

AurelienLP:
Jenkins, test this please.

mbounabi (approved):

AurelienLP:
Jenkins, test this please.

xlqian (approved):

PR #3046, opened 2019-11-22
[Release] Pull release branch before checking-out it.

nonifier:
On your first release, if you haven't pulled the release branch already, the
script would just fail on you.

### DIFF
--- a/release/script_release.py
+++ b/release/script_release.py
@@ -71,6 +71,7 @@ class ReleaseManager:
         self.git.rebase(remote_name + "/dev", "dev")
         self.dev_data_version = self.get_data_version()
         try:
+            self.git.pull("release")
             self.git.checkout("release")
         except Exception as e:
             print("Cannot checkout 'release':{}, creating from distant branch".format(str(e)))
--- a/release/script_release.py
+++ b/release/script_release.py
@@ -70,13 +70,12 @@ class ReleaseManager:
         # TODO quit on error
         self.git.rebase(remote_name + "/dev", "dev")
         self.dev_data_version = self.get_data_version()
+        remote_release = remote_name + "/release"
         try:
-            self.git.pull("release")
-            self.git.checkout("release")
+            self.git.checkout("-B release ", remote_release)
         except Exception as e:
             print("Cannot checkout 'release':{}, creating from distant branch".format(str(e)))
-            self.git.checkout("-b", "release", remote_name + "/release")
-        self.git.rebase(remote_name + "/release", "release")
+            self.git.checkout("-b", "release", remote_release)
 
         print("checking that release was merged into dev...")
         unmerged = self.git.branch("--no-merged", "dev", '--no-color')


release/script_release.py
```diff
@@ -71,6 +71,7 @@ def __init__(self, release_type, remote_name="canalTP"):
         self.git.rebase(remote_name + "/dev", "dev")
         self.dev_data_version = self.get_data_version()
         try:
+            self.git.pull("release")
```
pbougue: 1. In case someone have done extra unpushed work on `release` branch locally (and forgot about it) : this should be `--ff-only` as one would push weird things if a merge commit is needed between local `release` and `CanalTP:release`.

2. I guess the rebase below is not needed anymore. If it still is, the same `--ff-only` may be done there.

3. We should precise the distant release branch we want to pull from: in case referent distant repo is personal one and not `remote_name` (CanalTP).

nonifier: I've removed `git.pull("release")` in favour of `git checkout -B <local branch> <remote branch>`

This has 2 benefits:

1. It creates a local branch if it doesn't exist
2. it reset the local branch with the remote one

https://git-scm.com/docs/git-checkout#Documentation/git-checkout.txt-emgitcheckoutem-b-Bltnewbranchgtltstartpointgt

release/script_release.py
```diff
@@ -70,12 +70,13 @@ def __init__(self, release_type, remote_name="canalTP"):
...
+        remote_release = remote_name + "/release"
         try:
-            self.git.checkout("release")
+            self.git.checkout("-B release " + remote_release)
         except Exception as e:
             print("Cannot checkout 'release':{}, creating from distant branch".format(str(e)))
-            self.git.checkout("-b", "release", remote_name + "/release")
-        self.git.rebase(remote_name + "/release", "release")
+            self.git.checkout("-b", "release", remote_release)
+        self.git.rebase(remote_release, "release")
```
pbougue: minor: I think you don't need it anymore, then.

pbougue (approved):
:+1: the only possible problem I imagine is a local loss of work done only on `release` branch, seems OK.

benoit-bst (approved):

release/script_release.py
```diff
@@ -70,12 +70,13 @@ def __init__(self, release_type, remote_name="canalTP"):
         # TODO quit on error
         self.git.rebase(remote_name + "/dev", "dev")
         self.dev_data_version = self.get_data_version()
+        remote_release = remote_name + "/release"
         try:
-            self.git.checkout("release")
+            self.git.checkout("-B release " + remote_release)
```
mbounabi: Minor, for consistency with line 79... Or the other way with '+' 
```suggestion
            self.git.checkout("-B release ", remote_release)
```

mbounabi (approved):

nonifier:
Jenkins, test this please

PR #3049, opened 2019-11-27
Fix instabilities in fares

pbench:
Fare calculation sometimes return two tickets for a journey whereas one would be enough.
This was due to a bad comparison operator in fare computation.
Fixed in https://github.com/CanalTP/navitia/pull/3049/commits/ab4e1d44887a4cb8ee332428d2ffe899260be8df
Relates to https://jira.kisio.org/browse/NDTMA-57

Sometimes the fare calculator cannot find an appropriate ticket. In this case, an "unknown_ticket" is
returned for some sections of a journey.
These "unkown_ticket"s sometimes disappear in the response obtained by jormungandr.
It seems to be due to the pointer in PbCreator.unknown_ticket https://github.com/CanalTP/navitia/blob/ed565e0e87caddfd3af94461cdd5b838f978a4fc/source/type/pb_converter.h#L203
The first time an unknown_ticket is found, this PbCreator.unknown_ticket is assigned to a pointer to a Ticket object in PbCreator::fill_fare_section https://github.com/CanalTP/navitia/blob/ed565e0e87caddfd3af94461cdd5b838f978a4fc/source/type/pb_converter.cpp#L1722

This ticket object is destroyed to the next call to journeys  https://github.com/CanalTP/navitia/blob/ed565e0e87caddfd3af94461cdd5b838f978a4fc/source/kraken/worker.cpp#L378
After that, PbCreator.unknown_ticket points to deallocated memory.
However, in the next call to fill_fare_section, this dealllocated memory is read and put in the protobuf response. I guess that protobuf reject this object as invalid, and thus does not appear in the response.
Fixed in https://github.com/CanalTP/navitia/pull/3049/commits/c6640d4918c3b2d8ae0ef8101e2cbf1bebf4949f
Fix https://jira.kisio.org/browse/NAVP-1467

### DIFF
--- a/source/fare/fare.cpp
+++ b/source/fare/fare.cpp
@@ -134,7 +134,7 @@ results Fare::compute_fare(const routing::Path& path) const {
     results res;
     int nb_nodes = boost::num_vertices(g);
 
-    LOG4CPLUS_TRACE(logger, "Computing fare for journey : \n" << path);
+    LOG4CPLUS_DEBUG(logger, "Computing fare for journey : \n" << path);
 
     if (nb_nodes < 2) {
         LOG4CPLUS_TRACE(logger, "no fare data loaded, cannot compute fare");
--- a/source/fare/fare.cpp
+++ b/source/fare/fare.cpp
@@ -39,6 +39,9 @@ www.navitia.io
 #include "type/datetime.h"
 #include "routing/routing.h"
 
+#define FARE_LOG_ENABLED false
+
+
 namespace greg = boost::gregorian;
 
 namespace navitia {
@@ -70,10 +73,14 @@ static Label next_label(Label label, Ticket ticket, const SectionKey& section) {
         // empty ticket, it is juste a change
         // we have to update the number of changes and the duration with the same ticket
         if (ticket.caption == "" && ticket.value == 0) {
-            LOG4CPLUS_TRACE(logger, "Empty ticket, free change");
+            #if(FARE_LOG_ENABLED)
+                LOG4CPLUS_TRACE(logger, "Empty ticket, free change");
+            #endif
             label.nb_changes++;
         } else {
-            LOG4CPLUS_TRACE(logger, "We buy a new ticket");
+            #if(FARE_LOG_ENABLED)
+                LOG4CPLUS_TRACE(logger, "We buy a new ticket");
+            #endif
             // we bought a new ticket
             // we save the global cost, and we reset the number of changes and duration
             if (ticket.value.undefined)
@@ -134,7 +141,9 @@ results Fare::compute_fare(const routing::Path& path) const {
     results res;
     int nb_nodes = boost::num_vertices(g);
 
-    LOG4CPLUS_DEBUG(logger, "Computing fare for journey : \n" << path);
+    #if(FARE_LOG_ENABLED)
+        LOG4CPLUS_DEBUG(logger, "Computing fare for journey : \n" << path);
+    #endif
 
     if (nb_nodes < 2) {
         LOG4CPLUS_TRACE(logger, "no fare data loaded, cannot compute fare");
@@ -150,8 +159,9 @@ results Fare::compute_fare(const routing::Path& path) const {
             section_idx++;
             continue;
         }
-
-        LOG4CPLUS_TRACE(logger, "In section " << section_idx << " : \n" << item);
+        #if(FARE_LOG_ENABLED)
+            LOG4CPLUS_TRACE(logger, "In section " << section_idx << " : \n" << item);
+        #endif
         SectionKey section_key(item, section_idx++);
 
         std::vector<std::vector<Label>> new_labels(nb_nodes);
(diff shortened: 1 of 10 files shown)


source/fare/fare.h
```diff
@@ -259,10 +263,16 @@ struct Label {
             return nb_undefined_sub_cost < l.nb_undefined_sub_cost;
         if (cost.value != l.cost.value)
             return cost.value < l.cost.value;
+        if (tickets.size() != l.tickets.size()) {
```
pbougue, line 266: :+1: This explains the 1-cent fix for fares, I guess.

source/type/pb_converter.cpp
```diff
@@ -1712,18 +1712,14 @@ void PbCreator::fill_fare_section(pbnavitia::Journey* pb_journey, const fare::re
 
         pbnavitia::Ticket* pb_ticket = nullptr;
         if (ticket.is_default_ticket()) {
-            if (!unknown_ticket) {
```
pbougue, line 1715: So no more try to only build one unknown_ticket : looks OK, probably not so bad.
Any idea why that was like that? Just to shorten a bit the response?
The other possible choice of securing the lonely unknown-ticket but "keeping" the feature is too costly?

pbougue (commented):
Looks nice.

Didn't look much at the form, just some questions about the context and the "why" (looks like you fixed 2 issues).

Please, can you change the title of the PR to have something meaningful for external people (JIRA is not public) ?

Edit: I don't think this PR is blocked by any artemis PR, as there is no ticket check currently, no?

kadhikari:
I compared two artemis custom build using this PR as navitia, artemis and artemis_reference as mentioned above and did not find and differences(there is no more random unknown_ticket generation):  
- https://ci.navitia.io/view/custom_artemis/job/custom_artemis_reference_branch_and_artemis_branch_build/291/
- https://ci.navitia.io/view/custom_artemis/job/custom_artemis_reference_branch_and_artemis_branch_build/294/

I also compared one of the above build with a nightly build  and seems correct to me:
- nightly build without tickets in short_response: https://ci.navitia.io/job/Artemis_nightly/1288/

:+1: Pascal

source/fare/fare.h
pbench: yep ! After this commit, it should work well with 0-cent ticket

source/type/pb_converter.cpp
pbench: I have no idea what was the functionnality behind this.
Since the bug has been present for 4 years https://github.com/CanalTP/navitia/commit/380eefab27a5b6e2c7f77a38629fc10c58c1632c I guess the functionnality is not relevant anymore

pbench:
> Please, can you change the title of the PR to have something meaningful for external people (JIRA is not public) ?

Ok edited, and added more meaningful explaination in the PR

> Edit: I don't think this PR is blocked by any artemis PR, as there is no ticket check currently, no?

You're right !

pbench:
Jenkins, test this please.

source/fare/fare.cpp
```diff
@@ -138,27 +153,44 @@ results Fare::compute_fare(const routing::Path& path) const {
     labels[0].push_back(Label());
     size_t section_idx(0);
 
-    for (const auto& item : path.items) {
+    for (const navitia::routing::PathItem& item : path.items) {
```
pbougue, line 144: minor: `auto` was fine, no?

pbougue (approved):

source/fare/fare.cpp
pbench: yes it was, being explicit just help for code navigation

pbougue:
Jenkins, test this please.

pbougue:
Jenkins, test this please.

pbench:
Jenkins, test this please.

source/fare/fare.cpp
```diff
@@ -67,8 +72,14 @@ static Label next_label(Label label, Ticket ticket, const SectionKey& section) {
         // empty ticket, it is juste a change
         // we have to update the number of changes and the duration with the same ticket
         if (ticket.caption == "" && ticket.value == 0) {
+#if (FARE_LOG_ENABLED)
+            LOG4CPLUS_TRACE(logger, "Empty ticket, free change");
+#endif
```
nonifier: you can leave the `TRACE` without the #if/#endif for 2 reasons:
1 - we rarely use the trace level for outputting log; and even if you want, you can still activate/deactivate the `fare` logger on demand at the logger's configuration
2 - whatever is inside this #if will never be compiled by the CI. This code will end-up rotting until somebody decide to re-activate it in the future. This is a really bad practice inherited from C's style practices ;)

source/routing/routing.h
```diff
@@ -112,6 +112,52 @@ struct PathItem {
     }
 };
 
+inline std::ostream& operator<<(std::ostream& ss, const PathItem& t) {
```
nonifier: Why do you need to inline this ?! There is no template or apparent reason of moving it to a header.
It was better when this was in the .cpp file :(

source/fare/fare.cpp
pbench: Ok, the macro has been removed in 24eb4dcc95284c5210af415bf8c2338ba80477b5
I introduced it to avoid checking the condition if(log_level > trace) when the log is disabled, but it makes the code harder to read anyway

source/routing/routing.h
pbench: Good catch ! Fixed in 07005c15df4b9e52f9680128932e65490fbdbad4
I think I had some cyclic dependency linking problem at some point, that's why it ended up in the header. Seems to work now.

pbench: Hum, doesn't compile after a make clean, so the dependecy problem is stil here

pbench: Ok, fixed in 0e9da87ca6a4d79b03309c273c87e7785c3acb69

nonifier (approved):
Much cleaner ! Nice one

PR #3050, opened 2019-11-27
[Distributed] Use transient_socket for asgard

xlqian:
Why this PR?

Because jormungandr creates sockets via the AWS's autobalancer, when a new instance is popped by autoscaling, despite the autobalancer, sockets created previously will still stick to the old instance. We have to close the old socket(when TTL runs out) and reopen one so that traffic will be lead to new instances.

This PR is inspired by this code: https://github.com/CanalTP/navitia/pull/2594/

But a little bit different, in the mentioned PR, we close these sockets are not used anymore but not those sockets that are created a long time ago but recently used... 

In this PR, we close all sockets whose TTL runs out, even though it's used recently...


AurelienLP:
This is ok for me except the status part : the new attribute is not return since you need to update the serializer and the UTs

AurelienLP (commented):
This is ok for me except the status part : the new attribute is not return since you need to update the serializer and the UTs

kinnou02 (commented):
I think there is a better design for your needs.

Invert the sort of the priority queue to always get the most recent one, check its lifetime in `sockets()` to not use an old socket.
For your need the reaper doesn't require the usage of uwsgi's timers.
An alternative is to not use a reaper and to spawn a cleanning greenlet only when an old socket is encountered.

xlqian:
@kinnou02 I agree with you. The reason why I used `queue.PriorityQueue` is that this queue is a synchronized object. OK, I'll see how to do your suggestion.

xlqian:
"An alternative is to not use a reaper and to spawn a cleaning greenlet only when an old socket is encountered." that sounds a good idea, but I cannot spawn a greentlet inside a greenlet (not that the socket is required already inside of a greenlet)

xlqian:
@kinnou02 there is one thing that bothers me is that if we check TTL in 'socket()', it means the check will only happen we a request is received? and invert the priority queue seems a good idea, but it will complicate a lot the implementation in _reaper

AurelienLP (approved):

source/jormungandr/jormungandr/transient_socket.py
```diff
@@ -0,0 +1,155 @@
...
+    @staticmethod
+    def _reap(o, sockets):
+        try:
+            while True:
+                with _semaphore:
+                    now = time.time()
+                    # sockets's last element is the oldest one
+                    t, socket = sockets[-1]
+                    if now - t > o.ttl:
+                        TransientSocket.close_socket(socket, o.name)
```
nonifier: Do you need to keep the lock on the semaphore for closing the socket ?

source/jormungandr/jormungandr/transient_socket.py
```diff
@@ -0,0 +1,154 @@
...
+    @contextmanager
+    def socket(self):
+        # We don't want to waste time to close sockets in this function since the performance is critical
+        # The cleaning job is done in another greenlet.
+        try:
+            with _semaphore:
+                # self._sockets's first element is the most recent one
+                t, socket = self._sockets[self][
+                    0
+                ]  # If self._sockets is empty, a IndexError exception will be raised
```
xlqian, line 90: pffff... black did this ugly thing...

nonifier (approved):

PR #3053, opened 2019-11-28
Fare: Using clang-tidy on fare

AurelienLP:
Same thing as here https://github.com/CanalTP/navitia/pull/3052 but on fare.


AurelienLP:
Jenkins, test this please.

kadhikari (approved):

PR #3066, opened 2019-12-02
Type: Using clang-tidy on type

AurelienLP:
Same thing as here https://github.com/CanalTP/navitia/pull/3052 but on type.

### DIFF
--- a/source/type/message.cpp
+++ b/source/type/message.cpp
@@ -29,91 +29,89 @@ www.navitia.io
 */
 
 #include "type/message.h"
+
 #include "type/pt_data.h"
+#include "type/serialization.h"
 #include "type/type.h"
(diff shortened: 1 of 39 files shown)


source/type/message.cpp
```diff
@@ -523,7 +517,7 @@ const Disruption* DisruptionHolder::get_disruption(const std::string& uri) const
     return it->second.get();
 }
 
-void DisruptionHolder::add_weak_impact(boost::weak_ptr<Impact> weak_impact) {
+void DisruptionHolder::add_weak_impact(const boost::weak_ptr<Impact>& weak_impact) {
```
pbench, line 520: Not really sure if a const ref to a weak_ptr and a copy of a weak_ptr does the same thing ?

PR #3069, opened 2019-12-04
CI: Using a fixed number of tries instead of time for the init of mock kraken

AurelienLP:
For the Jormungandr_integration tests, we use multiple mocks of Kraken (main_routing_test, basic_schedule_test...).

In test_mechanism, we launch them and initialize them asking them their metadatas and updating their properties. We use `retry` in `test_mechanism.py` to do so, every 10 ms up until 5 seconds.

Since `_send_and_receive` already has a timeout, let's just use a fixed number of tries instead of handling an interval and an other timeout. We would have less trouble this way.

### DIFF
--- a/source/jormungandr/tests/tests_mechanism.py
+++ b/source/jormungandr/tests/tests_mechanism.py
@@ -223,7 +223,9 @@ class AbstractTestFixture(unittest.TestCase):
             instance = i_manager.instances[name]
             try:
                 retrying.Retrying(
-                    stop_max_attempt_number=5, retry_on_result=lambda x: not instance.is_initialized
+                    stop_max_attempt_number=5,
+                    wait_fixed=10,
+                    retry_on_result=lambda x: not instance.is_initialized,
                 ).call(instance.init)
             except RetryError:
                 logging.exception('impossible to start kraken {}'.format(name))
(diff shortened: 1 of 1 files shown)


xlqian:
so it still remains unsolved that why that happens only now?? interesting it didn't happen before...

AurelienLP:
> so it still remains unsolved that why that happens only now?? interesting it didn't happen before...

No, we have no idea...

AurelienLP:
Jenkins, test this please.

source/jormungandr/tests/tests_mechanism.py
```diff
@@ -223,7 +223,7 @@ def setup_class(cls):
             instance = i_manager.instances[name]
             try:
                 retrying.Retrying(
-                    stop_max_delay=5000, wait_fixed=10, retry_on_result=lambda x: not instance.is_initialized
+                    stop_max_attempt_number=5, retry_on_result=lambda x: not instance.is_initialized
```
mbounabi: Shouldn't there be a "wait_fixed" parameter also? As it is here, the attempts will be consecutive with no wait between each one.

AurelienLP:
Jenkins, test this please.

AurelienLP:
Jenkins, test this please.

xlqian (approved):

mbounabi (approved):

PR #3070, opened 2019-12-05
Kraken: Clang tidy on kraken

AurelienLP:
I've used clang-tidy 6.0 on kraken with the following command :
`clang-tidy -checks='*, -fuchsia-overloaded-operator, -fuchsia-default-arguments,-google*, -cppcoreguidelines-pro-bounds-array-to-pointer-decay, -hicpp-no-array-decay, -readability-implicit-bool-conversion, -misc-macro-parentheses' ../source/kraken/worker.cpp  -header-filter="*.pb.h" -fix -fix-errors`

It fixes a lot of basics things. There is still some warnings that may take a lot more effort to fix ourselves.

`-checks='*, -fuchsia-overloaded-operator, -fuchsia-default-arguments,-google*, -cppcoreguidelines-pro-bounds-array-to-pointer-decay, -hicpp-no-array-decay, -readability-implicit-bool-conversion, -misc-macro-parentheses'` runs all available checks except :

- `fuchsia-overloaded-operator` : just yell at us for overloading operators...
- all of google checks because they have some weird conventions sometime like passing a non const pointer instead of a nonconst reference in a function. Their reasoning is that it makes it clear that it is an in-out parameter. Meh...
- `cppcoreguidelines-pro-bounds-array-to-pointer-decay` : yell at us for using all of the `LOG4CPLUS` macros. May be a good thing to fix it someday
- `hicpp-no-array-decay` : same as above.
- `readability-implicit-bool-conversion` : replace all test of pointers from `if (pointer)` to `if (pointer != nullptr)` and all returns of function from `if function()` to `if function() == 0u` which is very dull...
- `misc-macro-parentheses` which destroy most of the macros...

I've also reordered the included files following this order:

- .h file of the acutal .cpp file
- .h from our project
- .h from external sources
- system includes

### DIFF
--- a/source/kraken/worker.cpp
+++ b/source/kraken/worker.cpp
@@ -30,22 +30,23 @@ www.navitia.io
 
 #include "worker.h"
 
-#include "routing/raptor_api.h"
 #include "autocomplete/autocomplete_api.h"
-
+#include "calendar/calendar_api.h"
+#include "disruption/line_reports_api.h"
+#include "disruption/traffic_reports_api.h"
+#include "equipment/equipment_api.h"
 #include "proximity_list/proximitylist_api.h"
 #include "ptreferential/ptreferential.h"
 #include "ptreferential/ptreferential_api.h"
-#include "time_tables/route_schedules.h"
-#include "time_tables/passages.h"
-#include "time_tables/departure_boards.h"
-#include "disruption/traffic_reports_api.h"
-#include "disruption/line_reports_api.h"
-#include "calendar/calendar_api.h"
 #include "routing/raptor.h"
+#include "routing/raptor_api.h"
+#include "time_tables/departure_boards.h"
+#include "time_tables/passages.h"
+#include "time_tables/route_schedules.h"
 #include "type/meta_data.h"
-#include "equipment/equipment_api.h"
+
 #include <numeric>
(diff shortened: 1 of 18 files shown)


nonifier (approved):
Hell Yeah ! :+1:

source/kraken/worker.cpp
```diff
@@ -1118,7 +1138,7 @@ void Worker::nearest_stop_points(const pbnavitia::NearestStopPointsRequest& requ
     entry_point.streetnetwork_params.max_duration = navitia::seconds(request.max_duration());
     street_network_worker->init(entry_point, {});
     // kraken don't handle reverse isochrone
-    auto result = routing::get_stop_points(entry_point, *data, *street_network_worker, false);
+    auto result = routing::get_stop_points(entry_point, *data, *street_network_worker, 0u);
```
pbougue, line 1141: It will work the same way it used to, but this change looks too big to be the initial intent of the developer.

AurelienLP: This is a mistake from the developer since the last (optional) argument of get_stop_points are an unsigned int (0u by default) and a boolean (false by default).

kinnou02 (approved):

PR #3076, opened 2019-12-05
Ed: Clang tidy on ed

AurelienLP:
Same as https://github.com/CanalTP/navitia/pull/3070 but on Ed.

### DIFF
--- a/source/ed/build_helper.cpp
+++ b/source/ed/build_helper.cpp
@@ -570,7 +570,7 @@ SA builder::sa(const std::string& name,
                const bool create_sp,
                const bool wheelchair_boarding,
                const bool bike_accepted) {
-    return SA(*this, name, x, y, create_sp, wheelchair_boarding, bike_accepted);
+    return {*this, name, x, y, create_sp, wheelchair_boarding, bike_accepted};
 }
 
 builder::builder(const std::string& date,
--- a/source/ed/build_helper.cpp
+++ b/source/ed/build_helper.cpp
@@ -29,8 +29,11 @@ www.navitia.io
 */
 
 #include "build_helper.h"
+
 #include "ed/connectors/gtfs_parser.h"
+
 #include <boost/range/algorithm/find_if.hpp>
+#include <utility>
 
 namespace pt = boost::posix_time;
 namespace dis = nt::disruption;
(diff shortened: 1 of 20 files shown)


source/ed/build_helper.cpp
```diff
@@ -250,14 +254,15 @@ VJ& VJ::operator()(const std::string& sp_name,
...
+        if (!b.data->pt_data->networks.empty()) {
             sp->network = b.data->pt_data->networks.front();
+        }
 
         b.sps[sp_name] = sp;
         b.data->pt_data->stop_points.push_back(sp);
         auto sa_it = b.sas.find(sp_name);
         if (sa_it == b.sas.end()) {
-            navitia::type::StopArea* sa = new navitia::type::StopArea();
+            auto* sa = new navitia::type::StopArea();
```
nonifier, line 265: Can I have no raw pointers in the code base for :christmas_tree: :santa: :christmas_tree: ?

source/ed/ed_persistor.cpp
```diff
@@ -1045,21 +1055,21 @@ void EdPersistor::insert_vehicle_journeys(const std::vector<types::VehicleJourne
     }
     this->lotus.finish_bulk_insert();
     for (types::VehicleJourney* vj : vehicle_journeys) {
-        std::string values = "";
+        std::string values;
         if (vj->prev_vj) {
-            values = "previous_vehicle_journey_id = " + boost::lexical_cast<std::string>(vj->prev_vj->idx);
+            values = "previous_vehicle_journey_id = " + std::to_string(vj->prev_vj->idx);
```
nonifier, line 1060: Are you really sure about that ? is it actually the same thing ?

nonifier (approved):

source/ed/build_helper.cpp
AurelienLP: I have this dream too but it won't be as easy as that... Plus I am not sure if the effort is worth the benefit...

source/ed/ed_persistor.cpp
AurelienLP: I can check that just in case!

AurelienLP: `vj->prev_vj->idx` is a uint_32_t so it works! :smile:

xlqian (approved):

PR #3087, opened 2019-12-13
[jorm] reduce simplification of coverage shapes used by bragi

kinnou02:
`0.1` exclude a lot of the coverage :(

image
red is the original shape, green is simplified with a tolerance of `0.1` and purple with `0.01`

### DIFF
--- a/source/jormungandr/jormungandr/instance.py
+++ b/source/jormungandr/jormungandr/instance.py
@@ -658,7 +658,7 @@ class Instance(object):
             self.geojson = None
             return
         # simplify the geom to prevent slow query on bragi
-        geom = self.geom.simplify(tolerance=0.1)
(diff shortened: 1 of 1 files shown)


mbounabi (approved):

pbougue (approved):
It's better anyway.

But shouldn't it strictly include the original one?
Maybe a convex envelope is too much, though.

kinnou02:
> It's better anyway.
> 
> But shouldn't it strictly include the original one?
> Maybe a convex envelope is too much, though.

I tried a convex envelope but it's too big :(

SGrenet (approved):
With 0.01, excluded areas are only edge cases on boundaries
=> few chances to have a town centers in these exclusions

PR #3088, opened 2019-12-13
osm2ed: import car speed in ed database

kinnou02:
This PR import car speed on every edges of the routing graph.
The tag `maxspeed` is used if present else a speed is chosen from the
highway type.
A factor is applied on max speed to get a somewhat realistic speed for
the edge.

Default value are valid for France and should be configurable if we want
to use it internationally. Rural zone vs urban zone isn't handled so
secondary and tertiary highway have a max speed of 50km/h everywhere if
not specified by a tag, primary are at 80km/h so sometime max speed is
higher than the legal value :/

TODO:
  - [x] add tests
  - [x] add flag to disable import
  - [x] format & rebase

### DIFF
--- a/source/ed/connectors/speed_parser.cpp
+++ b/source/ed/connectors/speed_parser.cpp
@@ -40,11 +40,11 @@ www.navitia.io
 namespace ed {
 namespace connectors {
 
-constexpr float kmh_to_ms(float val){
+constexpr float kmh_to_ms(float val) {
     return val / 3.6;
 }
 
-template<typename T>
+template <typename T>
 static void update_if_unknown(boost::optional<T>& target, const T& source) {
     if (!target) {
         target = source;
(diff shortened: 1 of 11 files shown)


xlqian:
are we competing with Valhalla? LOL

kinnou02:
> are we competing with Valhalla? LOL

This is for @eturck !

source/ed/connectors/speed_parser.cpp
```diff
@@ -0,0 +1,172 @@
...
+        update_if_unknown(speed, it->second);
+    } else {
+        std::cerr << highway << " not handled" << std::endl;
+    }
+
+    auto it_factor = speed_factor_by_type.find(highway);
+    if (it_factor != speed_factor_by_type.end() && speed) {
+        speed = speed.get() * it_factor->second;
+    } else {
+        std::cerr << highway << " not handled" << std::endl;
```
xlqian: what will happen if you find the max_speed but not the speed factor? can be thowed an exception?

kinnou02: We don't apply any factor, so we will go at `max_speed`, we may want to use a default factor, like `0.5`

source/ed/connectors/speed_parser.cpp
```diff
@@ -0,0 +1,172 @@
...
+    return parser;
+}
+
+boost::optional<float> SpeedsParser::get_speed(const std::string& highway,
+                                               const boost::optional<std::string>& max_speed) const {
+    boost::optional<float> speed;
+    boost::optional<float> factor;
+
+    if (max_speeds_by_type.empty() && speed_factor_by_type.empty()) {
+        return boost::none;
```
eturck, line 139: Shouldn't we return max_speed if a value is passed to get_speed even if they are both empty?
Maybe this check needs to go below `if (max_speed)`?

source/ed/connectors/speed_parser.cpp
eturck: A default factor of `0.5` would probably be nice :).

xlqian (approved):

source/ed/connectors/speed_parser.cpp
kinnou02: This is a way to indicate that the speed parser hasn't been initialized so speed aren't imported. I can add a boolean to make things clear

eturck: Ok for me then.
If we can configure the speed parser in the future there is a stupid case where you would pass empty maps for both of them and could still want to import maxspeed datas with no default value and only with the default speed_factor. But it's stupid :).

eturck (approved):

AurelienLP (approved):

xlqian (approved):

PR #3094, opened 2019-12-17
TimeTables: Clang tidy on time_tables

AurelienLP:
Same as https://github.com/CanalTP/navitia/pull/3070 but on Time Tables.

### DIFF
--- a/source/time_tables/departure_boards.h
+++ b/source/time_tables/departure_boards.h
@@ -43,10 +43,10 @@ typedef std::pair<boost::optional<routing::datetime_stop_time>, boost::optional<
     first_and_last_stop_time;
 
 void departure_board(PbCreator& pb_creator,
-                     const std::string& filter,
-                     const boost::optional<const std::string> calendar_id,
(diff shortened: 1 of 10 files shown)


nonifier (approved):

source/time_tables/departure_boards.h
```diff
@@ -43,10 +43,10 @@ typedef std::pair<boost::optional<routing::datetime_stop_time>, boost::optional<
     first_and_last_stop_time;
 
 void departure_board(PbCreator& pb_creator,
-                     const std::string& filter,
-                     const boost::optional<const std::string> calendar_id,
+                     const std::string& request,
```
xlqian, line 46: what? it changes variable's name too?

xlqian (approved):

source/time_tables/departure_boards.h
AurelienLP: What the name between the declaration and definition don't match yes!

PR #3103, opened 2019-12-30
[Tyr] Add dataset.state param

mbounabi:
https://jira.kisio.org/browse/NAVP-1482

Below, an example of the new `/jobs` response, with `state` in each dataset:
```
{
    "jobs": [
        {
            "created_at": "2020-01-02T13:09:57.505830",
            "updated_at": null,
            "state": "running",
            "data_sets": [
                {
                    "family_type": "poi",
                    **"state": "done",**
                    "type": "poi",
                    "name": "/srv/ed/backup/bibus/20200102-130957501933/fusio-poi.zip"
                },
                {
                    "family_type": "pt",
                    **"state": "done",**
                    "type": "fusio",
                    "name": "/srv/ed/backup/bibus/20200102-130957502764/fusio.zip"
                },
                {
                    "family_type": "streetnetwork",
                    **"state": "running",**
                    "type": "geopal",
                    "name": "/srv/ed/backup/bibus/20200102-130957503244/fusio-geopal.zip"
                }
            ],
            "id": 2
        }
```

### DIFF
--- a/source/tyr/tyr/binarisation.py
+++ b/source/tyr/tyr/binarisation.py
@@ -28,8 +28,9 @@
 # www.navitia.io
 
 """
-Functions to launch the binaratisations
+Functions to launch the binarizations
 """
+
 from __future__ import absolute_import, print_function, division
 import logging
 import os
@@ -173,12 +174,21 @@ def collect_metric(task_type, job, dataset_uid):
         logger.exception('unable to persist Metrics data: ')
 
 
+def _retrieve_dataset_and_set_state(dataset_type, job_id):
+    dataset = models.DataSet.find_by_type_and_job_id(dataset_type, job_id)
+    logging.getLogger(__name__).debug("Retrieved dataset: {}".format(dataset.id))
+    dataset.state = "running"
+    models.db.session.commit()
+    return dataset
+
+
 @celery.task(bind=True)
 @Lock(timeout=30 * 60)
 def fusio2ed(self, instance_config, filename, job_id, dataset_uid):
     """ Unzip fusio file and launch fusio2ed """
 
     job = models.Job.query.get(job_id)
+    dataset = _retrieve_dataset_and_set_state("fusio", job.id)
     instance = job.instance
 
     logger = get_instance_logger(instance, task_id=job_id)
@@ -201,15 +211,18 @@ def fusio2ed(self, instance_config, filename, job_id, dataset_uid):
         params.append("--log_comment")
         params.append(instance_config.name)
         res = None
-        with collect_metric('fusio2ed', job, dataset_uid):
+        with collect_metric("fusio2ed", job, dataset_uid):
             res = launch_exec("fusio2ed", params, logger)
         if res != 0:
-            raise ValueError('fusio2ed failed')
+            raise ValueError("fusio2ed failed")
+        dataset.state = "done"
     except:
-        logger.exception('')
-        job.state = 'failed'
-        models.db.session.commit()
+        logger.exception("")
+        job.state = "failed"
+        dataset.state = "failed"
         raise
+    finally:
+        models.db.session.commit()
 
 
 @celery.task(bind=True)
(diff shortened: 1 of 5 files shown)


source/tyr/tyr/binarisation.py
```diff
@@ -201,15 +211,18 @@ def fusio2ed(self, instance_config, filename, job_id, dataset_uid):
...
             res = launch_exec("fusio2ed", params, logger)
         if res != 0:
-            raise ValueError('fusio2ed failed')
+            raise ValueError("fusio2ed failed")
+        dataset.state = "done"
     except:
-        logger.exception('')
-        job.state = 'failed'
-        models.db.session.commit()
+        logger.exception("")
```
nonifier, line 220: Is it useful to log an empty exception here ?

source/tyr/tyr/binarisation.py
```diff
@@ -240,23 +254,26 @@ def gtfs2ed(self, instance_config, gtfs_filename, job_id, dataset_uid):
...
-            raise ValueError('gtfs2ed failed')
+            raise ValueError("gtfs2ed failed")
+        dataset.state = "done"
     except:
-        logger.exception('')
-        job.state = 'failed'
-        models.db.session.commit()
+        logger.exception("")
+        job.state = "failed"
+        dataset.state = "failed"
```
nonifier, line 265: Handling the dataset state at the job level seems redundant across all jobs.
I think it would be better to have it one level above ? 

Like in task.py:
```python
def import_data(...)
    [...]
    binarisation = [ed2nav.si(instance_config, job.id, custom_output_dir)]
    actions.append(chain(*binarisation))
    actions.append(finish_job.si(job.id))
    
    dataset = _retrieve_dataset_and_set_state("gtfs", job.id)
    try:
        if asynchronous:
            return chain(*actions).delay()
        else:
            return chain(*actions).apply()
    except:
        dataset.state = "failed" 
        models.db.session.commit()
```

source/tyr/tyr/binarisation.py
```diff
@@ -282,19 +299,22 @@ def osm2ed(self, instance_config, osm_filename, job_id, dataset_uid):
...
             args.extend(["--cities-connection-string", cities_db])
-        with collect_metric('osm2ed', job, dataset_uid):
-            res = launch_exec('osm2ed', args, logger)
+        with collect_metric("osm2ed", job, dataset_uid):
+            res = launch_exec("osm2ed", args, logger)
         if res != 0:
             # @TODO: exception
-            raise ValueError('osm2ed failed')
+            raise ValueError("osm2ed failed")
+        dataset.state = "done"
```
nonifier, line 310: Should this rather be the `ed2nav` job ?

source/tyr/tyr/binarisation.py
mbounabi: At this point, you can't know which task (associated to a dataset) has failed. So I don't think it's possible to do it this way.
Updating dataset state here is based on what is done for the job state.

source/tyr/tyr/binarisation.py
mbounabi: This will log the call stack with the exception, only with an empty custom message before it. So, yes, it is useful. Maybe there's a better way to do it but it is legacy and done like this everywhere

source/tyr/tyr/binarisation.py
mbounabi: why?

nonifier: I thought your dataset would only be "done" when the `.nav` is generated at the end of all tasks. 
I'm maybe confused, but in this case, what's the difference between the `job.state` and `dataset.state` here ?

mbounabi: When giving several types of data for binarization to Tyr, a task will be run for each type of data. Only one job is created for one binarization, with several datasets corresponding to each type of data (see description).
The `dataset.state` represents the state of the task `*2ed` associated to the dataset.
The `job.state` represents the state of all tasks `*2ed` + `ed2nav`

Now, if one task fails, we can know which one did.

nonifier (approved):

xlqian (approved):

benoit-bst (approved):

PR #3105, opened 2020-01-02
[Kraken] add log for zmq error

xlqian:
log when a bad ZMQ message is recieved

### DIFF
--- a/source/kraken/kraken_zmq.cpp
+++ b/source/kraken/kraken_zmq.cpp
@@ -149,6 +149,8 @@ int main(int argn, char** argv) {
     do {
         try {
             lb.run();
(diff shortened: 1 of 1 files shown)


source/kraken/kraken_zmq.cpp
```diff
@@ -149,6 +149,8 @@ int main(int argn, char** argv) {
     do {
         try {
             lb.run();
+        } catch (const navitia::recoverable_exception& e) {
+            LOG4CPLUS_ERROR(logger, e.what());
         } catch (const zmq::error_t&) {
```
AurelienLP, line 154: Don't you want to log something here just in case?

AurelienLP (approved):

nonifier (approved):

PR #3106, opened 2020-01-08
[Fare] update fare engine to handle fare v2 to fare v1 conversion

pbench:
Two updates to the fare engine :
- removed the lower casing of ids in fare transition (e.g. network=BlaBla is no longer equivalent to network=bLablA)
- accept "line=myLine" conditions for a fare transition

Needed to accept the conversion of fare v2 to fare v1 in https://github.com/CanalTP/transit_model/pull/509

https://jira.kisio.org/browse/NAVP-1465

### DIFF
--- a/source/ed/connectors/fare_parser.cpp
+++ b/source/ed/connectors/fare_parser.cpp
@@ -31,6 +31,7 @@ www.navitia.io
 #include "fare_parser.h"
 #include "utils/csv.h"
 #include "utils/base64_encode.h"
+#include "utils/functions.h"
 #include "ed/data.h"
 #include "fare_utils.h"
 
@@ -114,13 +115,8 @@ void fare_parser::load_transitions() {
 
 bool fare_parser::is_valid(const navitia::fare::State& state) {
     if (!state.mode.empty()) {
-        bool found = false;
-        for (const auto& mode : data.physical_modes) {
-            if (mode->uri == state.mode) {
-                found = true;
-                break;
-            }
-        }
+        bool found =
+            navitia::contains_if(data.physical_modes, [&](const auto& mode) { return mode->uri == state.mode; });
         if (!found) {
             LOG4CPLUS_WARN(logger, "A transition is valid only for the mode "
                                        << state.mode << " but this mode does not appears in the data.");
@@ -129,13 +125,8 @@ bool fare_parser::is_valid(const navitia::fare::State& state) {
     }
 
     if (!state.stop_area.empty()) {
-        bool found = false;
-        for (const auto& stop_area : data.stop_areas) {
-            if (stop_area->uri == state.stop_area) {
-                found = true;
-                break;
-            }
-        }
+        bool found = navitia::contains_if(data.stop_areas,
+                                          [&](const auto& stop_area) { return stop_area->uri == state.stop_area; });
         if (!found) {
             LOG4CPLUS_WARN(logger, "A transition is valid only for the stop_area "
                                        << state.stop_area << " but this stop_area does not appears in the data.");
@@ -144,13 +135,7 @@ bool fare_parser::is_valid(const navitia::fare::State& state) {
     }
 
     if (!state.line.empty()) {
-        bool found = false;
-        for (const auto& line : data.lines) {
-            if (line->uri == state.line) {
-                found = true;
-                break;
-            }
-        }
+        bool found = navitia::contains_if(data.lines, [&](const auto& line) { return line->uri == state.line; });
         if (!found) {
             LOG4CPLUS_WARN(logger, "A transition is valid only for the line "
                                        << state.line << " but this line does not appears in the data.");
@@ -159,13 +144,8 @@ bool fare_parser::is_valid(const navitia::fare::State& state) {
     }
 
     if (!state.network.empty()) {
-        bool found = false;
-        for (const auto& network : data.networks) {
-            if (network->uri == state.network) {
-                found = true;
-                break;
-            }
-        }
+        bool found =
+            navitia::contains_if(data.networks, [&](const auto& network) { return network->uri == state.network; });
         if (!found) {
             LOG4CPLUS_WARN(logger, "A transition is valid only for the network "
                                        << state.network << " but this network does not appears in the data.");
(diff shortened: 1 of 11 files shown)


source/ed/tests/fare_parser_test.cpp
```diff
@@ -52,13 +52,13 @@ BOOST_AUTO_TEST_CASE(parse_state_test) {
...
     // on n'est pas case sensitive
-    BOOST_CHECK(parse_state("mode=Metro").mode == navitia::encode_uri("metro"));
+    BOOST_CHECK(parse_state("mode=metro").mode == navitia::encode_uri("metro"));
 
     BOOST_CHECK(parse_state("zone=1").zone == "1");
 
     // on ignore les espaces
-    BOOST_CHECK(parse_state(" mode = Metro  ").mode == navitia::encode_uri("metro"));
-    BOOST_CHECK(parse_state("line=L1").line == navitia::encode_uri("l1"));
+    BOOST_CHECK(parse_state(" mode = metro  ").mode == navitia::encode_uri("metro"));
```
pbougue, line 59: Don't you want to change the right operand instead ?

source/fare/fare.cpp
```diff
@@ -383,16 +385,26 @@ bool Transition::valid(const SectionKey& section, const Label& label) const {
...
+            LOG4CPLUS_TRACE(logger, "dest_zone " << cond.value << " vs " << section.dest_zone);
+
             return false;
-        }
-        if (cond.key == "stoparea" && cond.value != section.dest_stop_area) {
+        } else if (cond.key == "stoparea" && cond.value != section.dest_stop_area) {
+            LOG4CPLUS_TRACE(logger, "dest_stop_are " << cond.value << " vs " << section.dest_stop_area);
+
             return false;
         }
```
pbougue, line 444: Shouldn't you also do that line check on the end_condition ?

source/fare/fare.cpp
```diff
@@ -383,16 +385,26 @@ bool Transition::valid(const SectionKey& section, const Label& label) const {
                 return false;
             }
         } else if (cond.key == "ticket" && !label.tickets.empty()) {
+            LOG4CPLUS_TRACE(logger, "ticket " << cond.value << " vs " << label.tickets.back().key);
+
             if (!compare(label.tickets.back().key, cond.value, cond.comparaison)) {
                 return false;
             }
+        } else if (cond.key == "line") {
```
pbougue, line 428: I think this should be documented in NTFS specs (fares v1): https://github.com/CanalTP/ntfs-specification/blob/master/ntfs_fare_extension_fr_deprecated.md

pbougue (commented):
I guess you are planning some of those:
1. Can you detail a bit (comment in PR, JIRA or in code if it fits): what is the use case for line add (functional case, v2 format, v1 format?)
2. It would be nice to have tests on those line transition.
3. Is there a warning/error checking that all conditions are valid ones (networks, lines, etc. actually exist)? This check would be very useful on idfm datasets to check nothing is broken by no-lowercasing (nothing more than before).
4. For safety, maybe add a feature-trigger on that no-lowercasing in case it causes problems?

source/ed/tests/fare_parser_test.cpp
```diff
@@ -52,13 +52,13 @@ BOOST_AUTO_TEST_CASE(parse_state_test) {
     BOOST_CHECK(parse_state("") == state);
 
     // on n'est pas case sensitive
```
benoit-bst: The comment is always valid ?

benoit-bst (approved):

source/ed/tests/fare_parser_test.cpp
pbench: what does it change ?

source/fare/fare.cpp
pbench: That would give two ways to perform the same check. 
I updated the spec to precise that a "line=..." condition can only appear in a start condition

source/fare/fare.cpp
pbench: done https://github.com/CanalTP/ntfs-specification/pull/12

source/ed/tests/fare_parser_test.cpp
pbougue: Check that no lowercasing is done (but that may be minor)

source/ed/tests/fare_parser_test.cpp
pbench: removed

source/ed/tests/fare_parser_test.cpp
pbench: you mean something like that ? :
`BOOST_CHECK(parse_state(" mode = Metro  ").mode != navitia::encode_uri("metro"));`

source/ed/connectors/fare_parser.cpp
```diff
@@ -101,6 +112,155 @@ void fare_parser::load_transitions() {
     }
 }
 
+bool fare_parser::is_valid(const navitia::fare::State& state) {
+    if (!state.mode.empty()) {
+        bool found = false;
+        for (const auto& mode : data.physical_modes) {
```
pbougue: You may probably shorten those checks using `contains()` https://github.com/CanalTP/utils/blob/master/functions.h#L217
And speedup using `data.physical_modes_map`.

pbench: physical_modes_map is not present in ed::Data (the "data" available here)

source/ed/connectors/fare_parser.cpp
```diff
@@ -101,6 +113,119 @@ void fare_parser::load_transitions() {
...
+bool fare_parser::is_valid(const navitia::fare::State& state) {
+    if (!state.mode.empty()) {
+        bool found =
+            navitia::contains_if(data.physical_modes, [&](const auto& mode) { return mode->uri == state.mode; });
+        if (!found) {
+            LOG4CPLUS_WARN(logger, "A transition is valid only for the mode "
+                                       << state.mode << " but this mode does not appears in the data.");
+            return false;
+        }
+    }
```
AurelienLP, line 125: Can't you refactor this part and use it for all the code below?

AurelienLP (approved):

benoit-bst (approved):

PR #3114, opened 2020-01-23
[Doc/Jormun] Errors

mbounabi:
https://jira.kisio.org/browse/NAVITIAII-3012

- [Doc] Update errors section 

- [Jormun] Use pythonic conditions

### DIFF
--- a/source/jormungandr/jormungandr/exceptions.py
+++ b/source/jormungandr/jormungandr/exceptions.py
@@ -56,11 +56,11 @@ class RegionNotFound(HTTPException):
             elif object_id[:6] == "coord:":
                 lon, lat = object_id[6:].split(":")
                 object_id = None
-        if region == lon == lat is None == object_id:
+        if not any([region, lon, lat, object_id]):
             self.data = format_error("unknown_object", "No region nor " "coordinates given")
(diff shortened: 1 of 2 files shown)


source/jormungandr/jormungandr/exceptions.py
```diff
@@ -56,11 +56,11 @@ def __init__(self, region=None, lon=None, lat=None, object_id=None, custom_msg=N
             elif object_id[:6] == "coord:":
                 lon, lat = object_id[6:].split(":")
                 object_id = None
-        if region == lon == lat is None == object_id:
+        if not any([region, lon, lat, object_id]):
```
nonifier, line 59: `not any([...])` should be expressed as `none_of([...])` really, it is much easier to read and understand.
 
It's weird that Python hasn't such a built-in function :( Maybe this exists in another lib ?

nonifier (approved):

source/jormungandr/jormungandr/exceptions.py
mbounabi: I guess that there's no point in having two opposite functions when there's `not`

benoit-bst (approved):

PR #3115, opened 2020-01-27
[Ed2Nav] Remove backup nav file

nonifier:
If we have successfully written and renamed the new data.nav file, we don't need the old one any more.

If you still want to retrieve the previous file, there is a daily backup done by the DevOps team that can be recovered from on demand.


source/ed/tests/ed2nav_test.cpp
```diff
@@ -72,11 +72,9 @@ BOOST_AUTO_TEST_CASE(writing_test_ed2nav_should_not_throw) {
     // Checks if the already existing binary is renamed and the new one is created
     BOOST_CHECK(ed::write_data_to_file(test, data));
     BOOST_CHECK(boost::filesystem::exists("ed2nav_test.nav.lz4"));
-    BOOST_CHECK(boost::filesystem::exists("ed2nav_test.nav.lz4.bak"));
```
AurelienLP, line 75: Why not BOOST_CHECK(!boost::filesystem::exists("ed2nav_test.nav.lz4.bak")); instead ?

source/ed/ed2nav.cpp
```diff
@@ -261,6 +261,18 @@ bool rename_file(const std::string& source_name, const std::string& dest_name) {
     return true;
 }
 
+bool remove_file(const std::string& filename) {
+    if(!boost::filesystem::exists(filename))
```
AurelienLP, line 265: if (..) {
 // blabla
}

AurelienLP (approved):

source/ed/ed2nav.cpp
nonifier: Do you really need this ?
```cpp
// if the filename doesn't exist, it's already deleted....
```

AurelienLP: I mean, you need to put the braces after the if, even if you have one instruction after :stuck_out_tongue:

nonifier: oooh, I see!
 
It's weird, I thought `clang-format` would do that type of formatting automatically :thinking:
I think we should rather harden the clang-formatting config to enforce this behaviour.

AurelienLP: I thought it would too, or maybe it is clang-tidy... :thinking:

benoit-bst (approved):

mbounabi (approved):

pbench (approved):

source/ed/ed2nav.cpp
nonifier: yup, looks like it's a clang-tidy thing - https://clang.llvm.org/extra/clang-tidy/checks/readability-braces-around-statements.html

PR #3116, opened 2020-01-27
[doc] Document the Service Translation algorithm

nonifier:
Add some sparse documentation around the Service Translation for a vehicle journey's validity pattern.

More documentation can be found here:
 - https://confluence.kisio.org/pages/viewpage.action?pageId=3768839
 - https://github.com/TeXitoi/pinot2015presenter/blob/master/pres/pinot2015presenter-pres.pdf

### DIFF
--- a/documentation/slate/source/includes/stuff.md
+++ b/documentation/slate/source/includes/stuff.md
@@ -106,6 +106,7 @@ This is due to a re-interpretation of the service pattern, with one specific goa
 
 This means that Navitia will re-organise the input data, to produce a smaller and more comprehensive response that respects the semantics of the input data.
 
+For more information about the algorithm (in French): [https://github.com/TeXitoi/pinot2015presenter/blob/master/pres/pinot2015presenter-pres.pdf](https://github.com/TeXitoi/pinot2015presenter/blob/master/pres/pinot2015presenter-pres.pdf)
 
 
 <a name="ridesharing-stuff"></a>Ridesharing
--- a/documentation/slate/source/includes/stuff.md
+++ b/documentation/slate/source/includes/stuff.md
@@ -101,10 +101,10 @@ You expected to have 3 `add` exceptions but you only have 1 `remove`. Also, the 
 This is due to a re-interpretation of the service pattern, with one specific goal :
 
 <aside class="notice">
- Navitia interprets the input service with the **least** number of exceptions
+ Navitia analyzes the input service and computes an equivalent pair of (week_pattern, exceptions) with the **smallest** number of exceptions
 </aside>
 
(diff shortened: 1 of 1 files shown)


documentation/slate/source/includes/stuff.md
```diff
@@ -53,6 +53,61 @@ The different journey types are:
...
+
+
+For instance, if you have a trip that:
+
+-   occurs every Saturday of January 2020 (in your `calendar.txt`)
+-   has 3 exceptions that `add` the service the first 3 Sundays (`calendar_dates.txt`)
+
+
+You might expect to have that same exact representation for your `/vehicle_journeys`, but instead you find something different.
+
```
pbench, line 98: Maybe move here the code block with "Navitia's interpreted service response"

documentation/slate/source/includes/stuff.md
```diff
@@ -53,6 +53,61 @@ The different journey types are:
...
+You expected to have 3 `add` exceptions but you only have 1 `remove`. Also, the week pattern is set to `true` on Sunday even though your `calendar.txt` said Saturday only.
+
+This is due to a re-interpretation of the service pattern, with one specific goal :
+
+<aside class="notice">
+ Navitia interprets the input service with the **least** number of exceptions
```
pbench:
```suggestion
 Navitia analyze the input service and compute an equivalent pair of (week_pattern, exceptions) but with the **smallest** number of exceptions
```

documentation/slate/source/includes/stuff.md
```diff
@@ -53,6 +53,61 @@ The different journey types are:
...
+</aside>
+
+This means that Navitia will try to re-organise the input data, to produce a smaller and more comprehensive response that respects the integrity of the data.
```
pbench:
```suggestion
This means that Navitia will re-organise the input data, to produce a smaller and more comprehensive response that respects the semantics of the input data.
```

pbench (approved):
Nice doc ! 
Just some suggestions for minor improvments
And maybe add a link to https://github.com/TeXitoi/pinot2015presenter/blob/master/pres/pinot2015presenter-pres.pdf ?

benoit-bst (approved):

documentation/slate/source/includes/stuff.md
nonifier: That's what I did at first but the rendering was weird as the code gets displayed on the right hand side of the web page. 
I agree it makes more sense when you read the code, but when looking at the rendered HTML page, it's much better up there :/

pbench: Ok !

PR #3117, opened 2020-01-28
fusio2ed: add stop_time_precison parsing for ntfs stop_times files

benoit-bst:
**Aim** : Change the **Stop Time precison** field inside the **NTFS norm**
There is a field inside the **stop_times.txt** file which provides information on the Stop Time reliability. 
Since the NTFS was updated, the name has changed and the value too (https://github.com/CanalTP/ntfs-specification/blob/master/ntfs_fr.md#stop_timestxt-requis)

**JIRA** : https://jira.kisio.org/browse/NAVITIAII-2943

We want **backward compatibility** for more suppleness, so the last field was :

- **date_time_estimated**
0 -> false
1 -> true

now the replacing field is :

- **stop_time_precision**
0 -> false
1 -> false
2 -> true

We support both

Nota: I let the date_time_estimated Var name everywhere because I want to be consistent with the others parts of the code (inside data struct and bdd it is date_time_estimated and I don’t want to do a complex migration). 

- [x] Utests
- [x] Artemis is Green


kadhikari (approved):
Well tested :+1:

mbounabi (approved):

PR #3041, opened 2019-11-20
kraken: move MetaVehicleJourney and get_type_e from type.h

benoit-bst:
**Quest** : Wipe out _type.h_ :crossed_swords: 

**Second round**
This PR aims to clean a bit more _type.h_ by moving **MetaVehicleJourney** and **get_type_e** method from type.h. For this, a  meta_vehicle_journey.h file is created and get_type_e is moved inside  type_interfaces.h.
It remains to move _EntryPoint_, _AccessibiliteParams_, _StreetNetworkParams_ struct and _get_admin_name_ method from _type.h_. 
I'll do it in an other PR.

PR #3119, opened 2020-02-03
kraken: type/type.h is not use now

benoit-bst:
**Quest :** Wipe out type.h :crossed_swords: 

We have many include with `#include "type/type.h"` (about 30 times), and the fact is, this header is too large. We don't need all includes inside it.
I cleaned the code to call the usefull header in place. To perform this task i needed to :

- move **entry_point** from _type.h_
- move **accessibility_params** from _type.h_
- Create a header `base_Pt_objects.h` to incude only base objects (_line, route, stop_area, stop_point_)
- replace types.h by usefull headers

Functionally, it is the same code.
This is not very fun to review because it is only headers moving :neutral_face: 

**Nota:** this work follows https://github.com/CanalTP/navitia/pull/3041


nonifier (approved):

pbench (approved):
Looks good !

PR #3120, opened 2020-02-03
Mask highway=trunk, trunk_link, motorway, motorway_link at autocomplete

BaptisteB-A:
[Jira ticket](https://jira.kisio.org/browse/NAVITIAII-2949)
Added tests

### DIFF
--- a/source/ed/connectors/osm_tags_reader.cpp
+++ b/source/ed/connectors/osm_tags_reader.cpp
@@ -153,7 +153,7 @@ std::bitset<8> parse_way_tags(const std::map<std::string, std::string>& tags) {
                 bike_direct = bike_lane;
         } else if (key == "bicycle") {
             if (val == "yes" || val == "permissive" || val == "destination" || val == "designated" || val == "private"
-                || val == "true" || val == "allowed" || val == "official" || val == "destination")
+                || val == "true" || val == "allowed" || val == "official")
                 bike_direct = bike_allowed;
             else if (val == "no" || val == "dismount" || val == "VTT")
                 bike_direct = bike_forbiden;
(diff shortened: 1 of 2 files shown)


source/ed/connectors/osm_tags_reader.cpp
```diff
@@ -153,7 +153,7 @@ std::bitset<8> parse_way_tags(const std::map<std::string, std::string>& tags) {
                 bike_direct = bike_lane;
         } else if (key == "bicycle") {
             if (val == "yes" || val == "permissive" || val == "destination" || val == "designated" || val == "private"
-                || val == "true" || val == "allowed" || val == "official" || val == "destination")
+                || val == "true" || val == "allowed" || val == "official")
```
nonifier, line 156: Are you sure you want to remove `destination` ? 
From https://wiki.openstreetmap.org/wiki/Bicycle#Bicycle_Restrictions, I understand that the access by bike is only allowed if the street is you destination. 

I guess this is now a problem is you want to cycle to your place, and you leave in a street that have that `destination` restriction.  :thinking:

source/ed/tests/osm_tags_reader_test.cpp
```diff
@@ -249,3 +249,28 @@ BOOST_AUTO_TEST_CASE(colon_tagging) {
     BOOST_CHECK_EQUAL(colon_params.get_applicable_poi_rule(bss_tags)->poi_type_id, "amenity:bicycle_rental");
     BOOST_CHECK_EQUAL(colon_params.get_applicable_poi_rule(effia_tags)->poi_type_id, "amenity:parking");
 }
+
+BOOST_AUTO_TEST_CASE(highway_trunk_not_visible) {
+    std::bitset<8> res = ed::connectors::parse_way_tags({{"highway", "trunk"}});
```
nonifier, line 254: Nice one !

source/ed/connectors/osm_tags_reader.cpp
BaptisteB-A: `val == "destination"` was there twice in the code, I simply removed one :)

mbounabi (approved):

nonifier (approved):

PR #3122, opened 2020-02-05
update checkout version to v2

nonifier:


.github/workflows/workflow.yml
```diff
@@ -12,31 +12,33 @@ on:
 
 jobs:
   build:
+    runs-on: ubuntu-latest
+
     container:
       image: navitia/debian8_dev
 
-    # tests
```
benoit-bst, line 18: I was wondering what it was for. A WM and a container ?
It seems to be a mistake, yes

.github/workflows/workflow.yml
```diff
@@ -12,31 +12,33 @@ on:
...
-
     services:
       rabbitmq:
         image: rabbitmq:3-alpine
         ports:
           - 5672:5672
 
     steps:
-    - uses: actions/checkout@v1
+    - uses: actions/checkout@v2
```
benoit-bst: is it pertinent to do the same here ?
-> [1](https://github.com/CanalTP/navitia/pull/3122/files#diff-f09ce9d24f2aeb928cee946c13324a06R70)
-> [2](https://github.com/CanalTP/navitia/pull/3122/files#diff-f09ce9d24f2aeb928cee946c13324a06R90)

PR #3124, opened 2020-02-06
[Jormungandr] Remove keep_alive http session in geocodejson

xlqian:
https://jira.kisio.org/browse/ND-638

the keep_alive http session seems to emit some nonsense `FIN/ACK` to Bragi. These `FIN/ACK` are not attached to any session and are treated by Firewall as Non-Sync TCP which impact the performance...

We've tried deactivation of keep_alive on a jormungandr machine, it seems that it worked, non more random `FIN/ACK` is emitted by jormungandr


benoit-bst (approved):

xlqian:
so far, during 24 hours, no more random fin/ack is observed on the test machine... so far so good...

PR #3136, opened 2020-02-17
Ed: Parse postal_code in osm

AurelienLP:
https://jira.kisio.org/browse/NAVITIAII-3033

In osm, for postal codes we have two codes : `addr:postcode` and `postal_code`.

We only parsed `addr:postcode` but now, if `addr:postcode` does not exist and `postal_code` exists, we use it.


nonifier (approved):

benoit-bst (approved):

mbounabi (approved):

benoit-bst:
docker tests are :green_apple:

PR #3137, opened 2020-02-19
CI: temporarily fix github CI

benoit-bst:
To not block PRs and having time to investigate:

- Run only debian8_dev docker
- Decrease disk space after tests by removing useless compiled tests
- Erase docker tests

### DIFF
--- a/.github/workflows/workflow.yml
+++ b/.github/workflows/workflow.yml
@@ -28,7 +28,7 @@ jobs:
 
     strategy:
         matrix:
-            docker_image: [navitia/debian8_dev, navitia/debian9_dev]
+            docker_image: [navitia/debian8_dev]
 
     container:
         image: ${{matrix.docker_image}}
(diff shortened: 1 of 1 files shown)


mbounabi (approved):

.github/workflows/workflow.yml
```diff
@@ -56,15 +56,15 @@ jobs:
         export JORMUNGANDR_BROKER_URL='amqp://guest:guest@rabbitmq:5672//'
         export KRAKEN_RABBITMQ_HOST='rabbitmq'
         ctest --output-on-failure
+    - name: remove useless tests
+      run: |
+        rm -rf tests/mock_kraken kraken/tests ed/tests fare/tests routing/tests calendar/tests
+        rm -rf georef/tests proximity_list/tests time_tables/tests equipment/tests
+        rm -rf ptreferential/tests
     - name: install tyr dependencies
```
mbounabi, line 64: There's no need for tyr dependencies if docker_test isn't run

benoit-bst: Yes but It is a partial change. 
I hope, in short times, to reactivate docker test

nonifier (approved):

PR #3141, opened 2020-02-21
[Tyr] Split brokers used by Tyr for Celery and Kraken

nonifier:
Celery and Kraken are using Rabbimq, but for complete different purpose.
The service used to be hosted on the exact same machine, which is not the case any more.

 Hence the need to address 2 different  brokers:
    * One for Tyr to interact with Celery: CELERY_BROKER_URL
    * One for Tyr to interact with the Krakens: KRAKEN_BROKER_URL

TODO :
 * [x] Needs https://github.com/CanalTP/fabric_navitia/pull/301

### DIFF
--- a/source/tyr/tyr/binarisation.py
+++ b/source/tyr/tyr/binarisation.py
@@ -546,7 +546,7 @@ def reload_data(self, instance_config, job_id):
         task.action = navitiacommon.task_pb2.RELOAD
 
         rabbit_mq_handler = RabbitMqHandler(
-            current_app.config['CELERY_BROKER_URL'], instance_config.exchange, "topic"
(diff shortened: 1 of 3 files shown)


mbounabi (commented):
Shouldn't there be a default value and a way to set it by environment variable in the settings file?

nonifier:
@mbounabi good catch, it's been added in [356756544fd74e2847637bdfcfa664f807ed5e28](https://github.com/CanalTP/navitia/pull/3141/commits/356756544fd74e2847637bdfcfa664f807ed5e28)

benoit-bst (approved):

mbounabi (approved):

kadhikari (approved):

PR #3158, opened 2020-03-03
[release] Release version type

nonifier:
Document versioning. 
Improve release script with questions to be more likely to generate the correct release type version.


nonifier:
I'll reopen the same PR, but from Canaltp repo so that the new workflow can be tested.
https://github.com/CanalTP/navitia/pull/3159

PR #3159, opened 2020-03-03
[Release] Release version type

nonifier:
* Document versioning.
* Improve release script with questions to be more likely to generate the correct release type version.
* Update release doc


.github/workflows/release_script.yml
```diff
@@ -0,0 +1,24 @@
...
+
+on:
+    push:
+        branches:
+            - dev
+            - release
+    pull_request:
+
+jobs:
+    release_script_tests:
```
benoit-bst: Can you put this cute little one inside https://github.com/CanalTP/navitia/blob/dev/.github/workflows/workflow.yml ?

benoit-bst: and i don't know if it is possible, but can we trigger this test with the **make test** command, as the others  ?

benoit-bst (approved):

.github/workflows/release_script.yml
nonifier: Unfortunately, the release script is one folder above the rest of the code... It would be a real pain to integrate it to cmake :cry:

xlqian (approved):

nonifier:
@Mergifyio rebase

PR #3160, opened 2020-03-05
Add end point terminus schedules

kadhikari:
- New end point /terminus_schedules added
- Some unit tests added in kraken
- Some more integration tests to come
- Ticket : https://jira.kisio.org/browse/NAVP-1515

Note: please see https://jira.kisio.org/browse/NAVP-1515 for more information.
Based on POC #2979

Some benchmark figures (comparisons between stop_schedules and terminus_schedules with IDFM data):
1. gare du nord -> 3 times faster
2. gare du nor / rer B -> 5 times faster
3. St Michel Notre Dame / rer -> 5 times faster

### DIFF
--- a/source/kraken/worker.cpp
+++ b/source/kraken/worker.cpp
@@ -249,6 +249,7 @@ static bool get_geojson_state(const pbnavitia::Request& request) {
         case pbnavitia::NEXT_DEPARTURES:
         case pbnavitia::NEXT_ARRIVALS:
         case pbnavitia::DEPARTURE_BOARDS:
+        case pbnavitia::terminus_schedules:
             result = request.next_stop_times().disable_geojson();
             break;
         case pbnavitia::PTREFERENTIAL:
@@ -471,6 +472,14 @@ void Worker::next_stop_times(const pbnavitia::NextStopTimeRequest& request, pbna
                     forbidden_uri, from_datetime, request.duration(), request.items_per_schedule(), request.depth(),
                     request.count(), request.start_page(), rt_level);
                 break;
+             case pbnavitia::terminus_schedules:
+                timetables::terminus_schedules(
+                    this->pb_creator, request.departure_filter(),
+                    request.has_calendar() ? boost::optional<const std::string>(request.calendar())
+                                           : boost::optional<const std::string>(),
+                            forbidden_uri, from_datetime, request.duration(), request.depth(), request.count(),
+                            request.start_page(), request.items_per_schedule());
+                        break;
             default:
                 LOG4CPLUS_WARN(logger, "Unknown timetable query");
                 pb_creator.fill_pb_error(pbnavitia::Error::unknown_api, "Unknown time table api");
@@ -1056,6 +1065,7 @@ void Worker::dispatch(const pbnavitia::Request& request, const nt::Data& data) {
         case pbnavitia::NEXT_DEPARTURES:
         case pbnavitia::NEXT_ARRIVALS:
         case pbnavitia::DEPARTURE_BOARDS:
+        case pbnavitia::terminus_schedules:
             next_stop_times(request.next_stop_times(), request.requested_api());
             break;
         case pbnavitia::NMPLANNER:
--- a/source/kraken/worker.cpp
+++ b/source/kraken/worker.cpp
@@ -472,14 +472,14 @@ void Worker::next_stop_times(const pbnavitia::NextStopTimeRequest& request, pbna
                     forbidden_uri, from_datetime, request.duration(), request.items_per_schedule(), request.depth(),
                     request.count(), request.start_page(), rt_level);
                 break;
-             case pbnavitia::terminus_schedules:
-                timetables::terminus_schedules(
-                    this->pb_creator, request.departure_filter(),
-                    request.has_calendar() ? boost::optional<const std::string>(request.calendar())
-                                           : boost::optional<const std::string>(),
-                            forbidden_uri, from_datetime, request.duration(), request.depth(), request.count(),
-                            request.start_page(), request.items_per_schedule());
-                        break;
+            case pbnavitia::terminus_schedules:
+                timetables::terminus_schedules(this->pb_creator, request.departure_filter(),
+                                               request.has_calendar()
+                                                   ? boost::optional<const std::string>(request.calendar())
+                                                   : boost::optional<const std::string>(),
+                                               forbidden_uri, from_datetime, request.duration(), request.depth(),
+                                               request.count(), request.start_page(), request.items_per_schedule());
+                break;
             default:
                 LOG4CPLUS_WARN(logger, "Unknown timetable query");
                 pb_creator.fill_pb_error(pbnavitia::Error::unknown_api, "Unknown time table api");
--- a/source/time_tables/departure_boards.cpp
+++ b/source/time_tables/departure_boards.cpp
@@ -140,6 +140,90 @@ static void render(PbCreator& pb_creator,
     }
 }
 
+static void render(PbCreator& pb_creator,
+                   const std::map<routing::JppIdx, pbnavitia::ResponseStatus>& response_status,
+                   const std::map<routing::JppIdx, vector_dt_st>& map_route_stop_point,
+                   const std::map<routing::JppIdx, first_and_last_stop_time>& map_route_point_first_last_st,
+                   const DateTime datetime,
+                   const DateTime max_datetime,
+                   const boost::optional<const std::string> calendar_id,
+                   const uint32_t depth) {
+    pb_creator.action_period =
+        pt::time_period(to_posix_time(datetime, *pb_creator.data), to_posix_time(max_datetime, *pb_creator.data));
+
+    for (const auto& id_vec : map_route_stop_point) {
+        auto schedule = pb_creator.add_terminus_schedules();
+        // Each schedule has a stop_point and a route
+        const auto& jpp_dir = pb_creator.data->dataRaptor->jp_container.get(id_vec.first);
+        const auto& jp_dir = pb_creator.data->dataRaptor->jp_container.get(jpp_dir.jp_idx);
+        const type::Route* route = pb_creator.data->pt_data->routes[jp_dir.route_idx.val];
+        const type::StopPoint* stop_point = pb_creator.data->pt_data->stop_points[jpp_dir.sp_idx.val];
+        pb_creator.fill(stop_point, schedule->mutable_stop_point(), depth);
+
+        auto m_route = schedule->mutable_route();
+
+        pb_creator.fill(route, m_route, depth);
+        if (route->line != nullptr) {
+            auto m_line = m_route->mutable_line();
+            pb_creator.fill(route->line, m_line, 0);
+        }
+        auto pt_display_information = schedule->mutable_pt_display_informations();
+
+        pb_creator.fill(route, pt_display_information, 0);
+
+        // add informations for debug. TODO : clean that
+        const auto& jpc = pb_creator.data->dataRaptor->jp_container;
+        pt_display_information->set_direction(
+            pb_creator.data->pt_data->stop_points[jpc.get(jp_dir.jpps.back()).sp_idx.val]->stop_area->name);
+        pt_display_information->set_description(std::string("JP = ") + std::to_string(jpp_dir.jp_idx.val)
+                                                + " JPP = " + std::to_string(id_vec.first.val));
+
+        // Now we fill the date_times
+        for (auto dt_st : id_vec.second) {
+            const auto& st_calendar = navitia::StopTimeCalendar(dt_st.second, dt_st.first, calendar_id);
+            // terminus or partial terminus
+            if (is_last_stoptime(st_calendar.stop_time)) {
+                continue;
+            }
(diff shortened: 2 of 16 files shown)


source/kraken/worker.cpp
```diff
@@ -249,6 +249,7 @@ static bool get_geojson_state(const pbnavitia::Request& request) {
         case pbnavitia::NEXT_DEPARTURES:
         case pbnavitia::NEXT_ARRIVALS:
         case pbnavitia::DEPARTURE_BOARDS:
+        case pbnavitia::terminus_schedules:
```
benoit-bst, line 252: why not in **Upper Case** like the others ?

source/time_tables/departure_boards.h
```diff
@@ -54,6 +54,17 @@ void departure_board(PbCreator& pb_creator,
...
 
+void terminus_schedules(PbCreator& pb_creator,
+                        const std::string& request,
+                        const boost::optional<const std::string>& calendar_id,
+                        const std::vector<std::string>& forbidden_uris,
+                        const pt::ptime date,
+                        const uint32_t duration,
+                        const uint32_t depth,
+                        const int count,
+                        const int start_page,
```
benoit-bst: Must be in **unit32**. It seems to be not negative

source/time_tables/departure_boards.cpp
```diff
@@ -140,6 +140,90 @@ static void render(PbCreator& pb_creator,
     }
 }
 
+static void render(PbCreator& pb_creator,
+                   const std::map<routing::JppIdx, pbnavitia::ResponseStatus>& response_status,
+                   const std::map<routing::JppIdx, vector_dt_st>& map_route_stop_point,
+                   const std::map<routing::JppIdx, first_and_last_stop_time>& map_route_point_first_last_st,
+                   const DateTime datetime,
+                   const DateTime max_datetime,
```
benoit-bst:
```suggestion
                   const DateTime& datetime,
                   const DateTime& max_datetime,
```

source/time_tables/departure_boards.cpp
```diff
@@ -339,6 +454,172 @@ void departure_board(PbCreator& pb_creator,
...
+    //  <stop_point_route, status>
+    std::map<routing::JppIdx, pbnavitia::ResponseStatus> response_status;
+
+    std::map<routing::JppIdx, vector_dt_st> map_route_stop_point;
+    std::map<routing::JppIdx, first_and_last_stop_time> map_route_point_first_last_st;
+
+    using LinePointIdx = std::pair<routing::LineIdx, routing::SpIdx>;
+    // group JourneyPatternEndsByDirection by StopPoint and Line so that directions at different
+    // stop points or different lines stay separated
+    std::map<LinePointIdx, std::vector<JourneyPatternEndsByDirection>> jp_ends_by_directions_from_lp;
```
benoit-bst, line 492: why not an hash struct (_flat_map_) ?
It is nice to use **std::map** (red-black tree) when the ordering is beneficial for algorithm.
I notice that `jp_ends_by_directions_from_lp` is only used with a **find** method (**O(logN)** for _std::map_).
So, with a hash the find is **O(1)**

the only question when we use a std::map is : do I need to keep order ?  for which algorithm ?

to recap
Becarefull with std::map and especially when there is a high number of elements to process. It can be not optimal for simple access

source/time_tables/tests/departure_boards_test.cpp
```diff
@@ -91,6 +91,24 @@ BOOST_AUTO_TEST_CASE(departureboard_test1) {
         BOOST_REQUIRE_EQUAL(resp.stop_schedules(0).date_times_size(), 1);
         BOOST_REQUIRE_EQUAL(resp.stop_schedules(1).date_times_size(), 1);
     }
+
+    // comparing terminus_schedule with above stop_schedules
```
benoit-bst: where stop_schedules is called to compare ?

source/time_tables/tests/departure_boards_test.cpp
```diff
@@ -91,6 +91,24 @@ BOOST_AUTO_TEST_CASE(departureboard_test1) {
...
+        terminus_schedules(pb_creator, "stop_point.uri=stop1", {}, {}, d("20150615T094500"), 43200, 0, 10, 0,
+                           std::numeric_limits<size_t>::max());
+
+        resp = pb_creator.get_response();
+        BOOST_REQUIRE_EQUAL(resp.terminus_schedules_size(), 2);
+        BOOST_REQUIRE_EQUAL(resp.terminus_schedules(0).date_times_size(), 1);
+        BOOST_REQUIRE_EQUAL(resp.terminus_schedules(1).date_times_size(), 1);
+        BOOST_REQUIRE_EQUAL(resp.terminus_schedules(0).stop_point().uri(), "stop1");
+        BOOST_REQUIRE_EQUAL(resp.terminus_schedules(0).pt_display_informations().direction(), "stop2");
+        BOOST_REQUIRE_EQUAL(resp.terminus_schedules(1).pt_display_informations().direction(), "stop3");
```
benoit-bst, line 110:
```suggestion
        BOOST_REQUIRE_EQUAL(resp.terminus_schedules(1).stop_point().uri(), "stop1");
        BOOST_REQUIRE_EQUAL(resp.terminus_schedules(1).pt_display_informations().direction(), "stop3");
```

source/time_tables/tests/departure_boards_test.cpp
```diff
@@ -146,6 +200,19 @@ BOOST_AUTO_TEST_CASE(departureboard_test1) {
         BOOST_REQUIRE_EQUAL(resp.stop_schedules(1).route().name(), "B");
         BOOST_CHECK_EQUAL(resp.stop_schedules(1).date_times_size(), 1);
     }
+
+    // No terminus_schedules on terminus for toute "A"
```
benoit-bst:
```suggestion
    // No terminus_schedules on terminus for route "A"
```
stop2 is not the _departure_ and the _terminus_ at the same time ?
I explain my point of view, when we are already on a terminus (like stop2) the response can contains itself, nope ?

source/time_tables/tests/departure_boards_test.cpp
```diff
@@ -1557,6 +1697,22 @@ BOOST_AUTO_TEST_CASE(stop_schedules_order_by_line_route_stop_point) {
     test(5, "route:13lr", "RubensLR");
     test(6, "route:13rl", "JennerRL");
     test(7, "route:13rl", "RubensRL");
+
+    // We compare terminus_schedume with stop_schedule:
+    navitia::PbCreator pb_creator_1(data_ptr, bt::second_clock::universal_time(), null_time_period);
+    terminus_schedules(pb_creator_1, "network.uri=base_network", {}, {}, d("20170103T070000"), 86400, 3, 20, 0,
```
benoit-bst, line 1704: `base_network` is for wich line-route ? all ?

source/time_tables/tests/departure_boards_test.cpp
```diff
@@ -1636,6 +1792,33 @@ BOOST_AUTO_TEST_CASE(stop_schedule_on_terminus) {
     BOOST_CHECK_EQUAL(stop_schedule.response_status(), pbnavitia::ResponseStatus::no_departure_this_day);
 }
 
+//  Check that there is no terminus_schedules on terminuses.
```
benoit-bst, line 1796: is the same here ? -> https://github.com/CanalTP/navitia/pull/3160/files#diff-71e4eb4cf5a3ff125342d54c13ff27d9R228

source/time_tables/tests/departure_boards_test.cpp
```diff
@@ -1678,3 +1861,216 @@ BOOST_AUTO_TEST_CASE(stop_schedule_on_partial_terminus) {
...
+    b.finish();
+    b.data->pt_data->sort_and_index();
+    b.data->build_raptor();
+    b.data->pt_data->build_uri();
+    auto* data_ptr = b.data.get();
+    // Set a different terminus on the route to have partial terminuses
+    data_ptr->pt_data->routes[0]->destination = b.sas.find("real terminus")->second;
+
+    navitia::PbCreator pb_creator_dep(data_ptr, bt::second_clock::universal_time(), null_time_period);
+    terminus_schedules(pb_creator_dep, "stop_point.uri=stop3", {}, {}, d("20181101T075500"), 86400, 3, 10, 0,
```
benoit-bst, line 1882: Where can I see it is **partial terminus** ? I only see that you call terminus_schedules on stop3

source/time_tables/tests/departure_boards_test.cpp
```diff
@@ -1678,3 +1861,216 @@ BOOST_AUTO_TEST_CASE(stop_schedule_on_partial_terminus) {
...
+     * Bobette:  C -> B -> A / D -> B -> A
+     *                           C
+     *                      -
+     * A -------------- B
+     *                      -
+     *                           D
+     */
+    ed::builder b("20160802");
+    b.vj("bob").route("bob")("A", "10:00"_t)("B", "11:00"_t)("C", "12:00"_t);
+    b.vj("bob").route("bob")("A", "10:15"_t)("B", "11:15"_t)("D", "12:15"_t);
```
benoit-bst, line 1910: Is it realistic to have a line with 2 routes with the same name ?

source/kraken/worker.cpp
kadhikari: After some api types at the begining, we started to use in lower case. I just followed ..

benoit-bst:
I have a main question. I don't master the algorithmic part of this PR and I am questioning about performance. 
Did you bench the response time compare to **stop_schedules** ? It is acceptable ?

source/time_tables/tests/departure_boards_test.cpp
kadhikari: It's a filter. base_network being the only network will give all the the lines in terminus_schedules.

source/time_tables/tests/departure_boards_test.cpp
kadhikari: This test is compared with the test : stop_schedule_on_partial_terminus
In stop_schedules, we may transform a terminus to a partial_terminus.
In terminus_schedules as the terminus as well as partial terminus is eliminated from the beginning, we don't have neither terminus nor partial_terminus.

source/time_tables/tests/departure_boards_test.cpp
kadhikari: Even in the production we have same name for two roots: http://api.navitia.io/v1/coverage/stif/stop_areas/stop_area%3A0%3ASA%3A59259/lines/line%3A0%3A100110013%3A13/routes?count=100&

kadhikari:
> I have a main question. I don't master the algorithmic part of this PR and I am questioning about performance.
> Did you bench the response time compare to **stop_schedules** ? It is acceptable ?

In this phase, as this end point is only for private use and will not be documented soon, i don't think that it is necessary non?   
@pbougue  has also mentioned in his poke : https://github.com/CanalTP/navitia/pull/2979

source/time_tables/tests/departure_boards_test.cpp
kadhikari: This compares with stop_schedule_on_terminus above.

benoit-bst:
> In this phase, as this end point is only for private use and will not be documented soon, i don't think that it is necessary non?

I just want to know if it is realistic. If it takes 10 secs, we can not accept.
It is a large mesh benchmark.

source/time_tables/tests/departure_boards_test.cpp
kadhikari: As stop2 has arrival as well as departure time, it might be the both.
In stop_schedules, we have an object stop_schedule without date_time for a terminus.
In terminus_schedule, we don't have object terminus_schedule for a terminus.

source/time_tables/departure_boards.cpp
kadhikari: I took the code of the POC

benoit-bst: This remark is especially available, if there is an issue with performance.

benoit-bst (approved):
> Some benchmark figures (comparisons between stop_schedules and terminus_schedules with IDFM data):
> gare du nord -> 3 times faster
> gare du nor / rer B -> 5 times faster
> St Michel Notre Dame / rer -> 5 times faster

I just saw your complement  on the performance. It looks pretty nice

pbench (approved):
:+1:

benoit-bst (approved):
Useful added tests. Nice

source/time_tables/departure_boards.cpp
```diff
@@ -140,6 +140,90 @@ static void render(PbCreator& pb_creator,
     }
 }
 
+static void render(PbCreator& pb_creator,
```
nonifier, line 143: This function is way too loooooooooooooooooooooooooooooong. 
We should definitely break this down into smaller functions. This would ease readability, maintainability and increase modularity for better unit testing.

nonifier: I'm happy to merge this like that if we really need to, but I think we really want to avoid that kind of coding practices ! 

We should really put in practice clean code:
https://alvinalexander.com/programming/clean-code-quotes-robert-c-martin

PR #3161, opened 2020-03-05
[CI] Split build workflow into 2

nonifier:
This is a test to make sure that different workflows are run on different VM. This is to mitigate the space limit that we are hitting on the main build.


nonifier:
Didn't worfk, closing it.

PR #3162, opened 2020-03-05
CI: Several actions to increase the CI quality

benoit-bst:
Several actions to increase the CI quality

- Pass **checkout in V2 mode**. The actions/checkout@v2 is less buggy. We can re-build jobs easlily without problems (https://github.com/actions/checkout/issues/23). Re-triggering a workflow finally works.
However, checkout V2 use a really new version of Git (Setup Git 2.18 or higher). So, this is not available for ubuntu 18 or debian 8/9...
- **Re-sync submodules** (SimpleAmqClient) and check only if submodules HEAD is contained inside origin/HEAD with **check_submodules** script
- Resurrect **Docker tests**
- Resurrect **debian 9 container** for build/tests

### DIFF
--- a/.github/workflows/workflow.yml
+++ b/.github/workflows/workflow.yml
@@ -28,7 +28,7 @@ jobs:
 
     strategy:
         matrix:
-            docker_image: [navitia/debian8_dev]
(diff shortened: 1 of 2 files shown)


nonifier (approved):
DROP DA MIC !

PR #3164, opened 2020-03-10
[Jormungandr] Set parameter for region

mbounabi:
https://jira.kisio.org/browse/NAVITIAII-3013

When no coverage is set in a request, the parameters from the chosen coverage needs to be applied.

Also:
Some changes in comments

### DIFF
--- a/source/jormungandr/jormungandr/interfaces/v1/Journeys.py
+++ b/source/jormungandr/jormungandr/interfaces/v1/Journeys.py
@@ -475,6 +475,8 @@ class Journeys(JourneyCommon):
 
         if region:
             _set_specific_params(i_manager.instances[region])
+        elif possible_regions:
+            _set_specific_params(i_manager.instances[possible_regions[0]])
         else:
             _set_specific_params(default_values)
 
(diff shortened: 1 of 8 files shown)


kadhikari (approved):
It seems that we didn't have a test on default parameter without coverage. There is not any test failure after this modification.

mbounabi:
> It seems that we didn't have a test on default parameter without coverage. There is not any test failure after this modification.

@kadhikari Indeed. However, there's no need to apply default values in any case as, in the end, there will always be a coverage to be queried on.

source/jormungandr/jormungandr/interfaces/v1/Journeys.py
```diff
@@ -499,7 +493,7 @@ def _set_specific_params(mod):
         responses = {}
         for r in possible_regions:
             self.region = r
-
+            _set_specific_params(i_manager.instances[r])
```
nonifier, line 496: :+1:

nonifier (approved):

source/jormungandr/jormungandr/interfaces/v1/journey_common.py
```diff
@@ -127,16 +127,15 @@ def compute_regions(args):
...
+    :return:    If region is set in the query, return it
+                If not, return a list of possible regions for the query
+    """
     if region:
-        region_obj = i_manager.get_region(region)
-
-    if not region:
-        # TODO how to handle lon/lat ? don't we have to override args['origin'] ?
-        possible_regions = compute_regions(args)
+        return [i_manager.get_region(region)]
```
xlqian, line 135: :+1:

xlqian (approved):

PR #3165, opened 2020-03-10
[Doc] Actual definition of 'max_duration'

mbounabi:
[skip ci]


documentation/slate/source/includes/apis.md
```diff
@@ -1023,7 +1023,7 @@ See how disruptions affect a journey in the [real time](#realtime) section.
 | nop     | count                | int     | Fixed number of different journeys<br>More in multiple_journeys  |             |
 | nop     | max_nb_transfers      | int     | Maximum number of transfers in each journey  | 10          |
 | nop     | min_nb_transfers     | int     | Minimum number of transfers in each journey  | 0           |
-| nop     | max_duration         | int     | Maximum duration of journeys in seconds (from `datetime` parameter).<br>More usefull when computing an isochrone (only `from` or `to` is provided).<br>On a classic journey (from-to), it will mostly speedup Navitia: You may have journeys a bit longer than that value (you would have to filter them).    | 86400       |
+| nop     | max_duration         | int     | Maximum duration to the beginning of the journey public transports section in seconds (from `datetime` parameter).<br>If `datetime` represents the arrival of the journeys requested, then the journey public transports section must start after `datetime` - `max_duration` <br>More useful when computing an isochrone (only `from` or `to` is provided)    | 86400       |
```
pbench:
```suggestion
| nop     | max_duration         | int     | In seconds. If `datetime` represents the beginning of the journeys requested, then the last public transport section of all journeys will end before `datetime + max_duration` .<br>If `datetime` represents the arrival of the journeys requested, then the first public transport section of all journeys will start after `datetime` - `max_duration` <br>More useful when computing an isochrone (only `from` or `to` is provided)    | 86400       |
```

mbounabi: Thanks @pbench 
I just made small changes: 
- `departure` instead of `beginning` to match the value of the parameter.
- Units at the end to be consistent with other parameters

pbench (approved):
:+1:

kadhikari (approved):

PR #3167, opened 2020-03-11
[GeoRef] Add projection cache to `get_duration_with_dijkstra`

nonifier:
Street Network Matrix can be computed 2 different ways with Kraken using `DijkstraPathFinder`:
 1. via `find_nearest_stop_points()` with a proximity that holds stop points ids (New Default scenario).
 2.  via `get_duration_with_dijsktra()` with a list of coordinates (Distributed scenario).

In the second case, we want to avoid projecting the coordinate for every single object. We could use this exact same caching mechanism at the one done for 1. (in `ProjectionGetterByCache`). 

This PR adds a cache that maps each stop points coordinate to its projected data on the street network. 

Read by commit to avoid all the header moves...

### DIFF
--- a/source/georef/georef.h
+++ b/source/georef/georef.h
@@ -205,6 +205,9 @@ struct GeoRef {
     typedef flat_enum_map<nt::Mode_e, ProjectionData> ProjectionByMode;
     std::vector<ProjectionByMode> projected_stop_points = {};
 
+    typedef std::unordered_map<nt::GeographicalCoord, ProjectionByMode> ProjectedCoords;
+    ProjectedCoords projected_coords;
+
     /// Graphe pour effectuer le calcul d'itinéraire
     Graph graph;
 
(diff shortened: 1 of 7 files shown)


source/georef/georef.h
```diff
@@ -238,6 +205,9 @@ struct GeoRef {
     typedef flat_enum_map<nt::Mode_e, ProjectionData> ProjectionByMode;
     std::vector<ProjectionByMode> projected_stop_points = {};
 
+    typedef std::unordered_map<nt::GeographicalCoord, ProjectionByMode> ProjectedCoords;
+    ProjectedCoords projected_coords;
```
xlqian, line 209: this is not serialized?

xlqian (commented):
I'm also interested in how much the data.nav gets larger?

xlqian (approved):

source/georef/georef.h
nonifier: Good point, but it;s not serialized. The projection pre-compute is performed by `build_proximity_list()` each time we load or re-load data.

benoit-bst (approved):

PR #3168, opened 2020-03-11
[Action] Publish images with condition

mbounabi:
Publish images:
- with tag `dev` when merging on dev
- with tag `relase`/`latest` when merging on release


benoit-bst (approved):
If condition is Sexy

PR #3172, opened 2020-03-16
fix debian compatibility to build sources

benoit-bst:
Currently, we can't compile with the new last debian version (**debian 10 - buster**). The fact is, the **pqxx** libraries broke backward compatibility by changing the Major number.
So, we have a difference in the API with the new version (version 6). The new Pqxx library has changed with the debian 10 buster...

The last version: **libpqxx-dev 4.xx**
The new version: **libpqxx-dev 6.xx**

My first idea was to use _Apt Debian backport_, but it seemed there was not the Pqxx old version into the backport (https://packages.debian.org/search?keywords=libpqxx-dev).
I did not want to re-compile the librairie after a curl acces to the sources. Therefore, I decided to use the master weapon :gun: the **CMake OS detection**.
It consists to detect OS for applicating the good struct with a **#if** into the code

I tested that for:
- debian 8
- debian 9
- debian 10
- ubuntu 18.04

Now, we can compile with the last debian version.

- [x] debian 10 added to CI

**Note:** The detection script is generic, it can detect a lot of OS (mint, fedora, opensuse, debian, ubuntu). I only tested OS that we support

### DIFF
--- a/.github/workflows/workflow.yml
+++ b/.github/workflows/workflow.yml
@@ -29,7 +29,7 @@ jobs:
 
     strategy:
         matrix:
-            docker_image: [navitia/debian8_dev, navitia/debian9_dev]
+            docker_image: [navitia/debian8_dev, navitia/debian9_dev, navitia/debian10_dev]
 
     container:
         image: ${{matrix.docker_image}}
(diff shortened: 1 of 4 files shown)


xlqian (approved):

pbench (approved):
I don't really speak Cmake, but it looks good

thegreathir:
Same problem is present about ubuntu 19.10
You  should set PQXX_COMPATIBILITY off in this ubuntu version

PR #3175, opened 2020-03-17
Minor perf enhancement in jormun

xlqian:
minor perf enhancement, the gain is about 3~5%.

If the direct_path is excluded by `max_{mode}_direct_path_duration`, the gain may be around 10%

### DIFF
--- a/source/jormungandr/jormungandr/street_network/kraken.py
+++ b/source/jormungandr/jormungandr/street_network/kraken.py
@@ -42,6 +42,7 @@ from jormungandr.street_network.street_network import (
 from jormungandr import utils
 import six
 from functools import cmp_to_key
+from jormungandr.street_network.utils import crowfly_distance_between
 
 
 class Kraken(AbstractStreetNetworkService):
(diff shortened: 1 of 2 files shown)


benoit-bst (approved):

source/jormungandr/jormungandr/street_network/kraken.py
```diff
@@ -99,9 +116,19 @@ def _direct_path(
...
-                direct_path_request['max_{mode}_direct_path_duration'.format(mode=mode)] / 2
+                request['max_{mode}_direct_path_duration'.format(mode=mode)] / 2
+            )
+            # if the crowfly distance between origin and destination is too large, there is no need to call kraken
+            crowfly_distance = crowfly_distance_between(
+                utils.get_pt_object_coord(pt_object_origin), utils.get_pt_object_coord(pt_object_destination)
             )
 
+            if (
+                crowfly_distance / float(direct_path_request['{mode}_speed'.format(mode=kraken_mode)])
```
kadhikari, line 127: I suppose the devider value is never 0.

kadhikari (approved):

mbounabi (approved):

source/jormungandr/jormungandr/street_network/kraken.py
xlqian: i'll add a try catch!!

xlqian: @kadhikari  after a quick check, there is no need,  this param is always > 0 !  https://github.com/CanalTP/navitia/blob/dev/source/jormungandr/jormungandr/interfaces/v1/journey_common.py#L287

PR #3178, opened 2020-03-24
[Kraken] optim proximitylist_api

xlqian:
https://jira.kisio.org/browse/TRR-103

a new interface is added to `find_within()`  in `proximity_list.cpp` which allows returning the crow fly distance.

two things are optimized in` proximitylist_api.cpp`:

1 no more 'sort', because the 'list' returned by flann are already sorted regarding the distance
2 no more computation of distance, the distances are now returned by `find_within()` 

The gain is about 10% ~ 40% for the call of proximity list, since the elapsed time (1ms ~ 10ms) in this function is quite small, the global gain for '/journeys' is small too...

### DIFF
--- a/source/proximity_list/proximity_list.cpp
+++ b/source/proximity_list/proximity_list.cpp
@@ -148,8 +148,8 @@ static int radius_search(const std::shared_ptr<index_t>& NN_index,
 
 template <class T>
 auto ProximityList<T>::find_within_impl(const GeographicalCoord& coord,
-                                        double radius,
-                                        int size,
+                                        const double radius,
+                                        const int size,
                                         IndexCoord /*unused*/) const
     -> std::vector<typename ReturnTypeTrait<T, IndexCoord>::ValueType> {
     // Containers are auto-sized by NN_index, Flann will return all objects inside of the given radius
@@ -166,10 +166,10 @@ auto ProximityList<T>::find_within_impl(const GeographicalCoord& coord,
     return res;
 }
 
-template <class T>
+template <typename T>
 auto ProximityList<T>::find_within_impl(const GeographicalCoord& coord,
-                                        double radius,
-                                        int size,
+                                        const double radius,
+                                        const int size,
                                         IndexCoordDistance /*unused*/) const
     -> std::vector<typename ReturnTypeTrait<T, IndexCoordDistance>::ValueType> {
     // Containers are auto-sized by NN_index, Flann will return all objects inside of the given radius
@@ -188,8 +188,8 @@ auto ProximityList<T>::find_within_impl(const GeographicalCoord& coord,
 
 template <class T>
 auto ProximityList<T>::find_within_impl(const GeographicalCoord& coord,
-                                        double radius,
-                                        int size,
+                                        const double radius,
+                                        const int size,
                                         IndexOnly /*unused*/) const
     -> std::vector<typename ReturnTypeTrait<T, IndexOnly>::ValueType> {
     // Using small sized std::array will avoid heap allocation and limit the research
@@ -203,7 +203,6 @@ auto ProximityList<T>::find_within_impl(const GeographicalCoord& coord,
     std::vector<typename ReturnTypeTrait<T, IndexOnly>::ValueType> res;
     auto op = [](const Item& item, float /*unused*/) { return item.element; };
     make_result<T>(coord, items, indices_data, distances_data, nb_found, res, op);
-    ;
 
     return res;
 }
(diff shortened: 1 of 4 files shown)


source/proximity_list/proximity_list.cpp
```diff
@@ -93,12 +93,20 @@ void ProximityList<T>::build() {
...
 template <typename T, typename Item>
-static auto extract(const Item& item, IndexCoord /*unused*/) -> typename ReturnTypeTrait<T, IndexCoord>::ValueType {
+static auto extract(const Item& item, float /*unused*/, IndexCoord /*unused*/) ->
+    typename ReturnTypeTrait<T, IndexCoord>::ValueType {
     return std::make_pair(item.element, item.coord);
 }
 
 template <typename T, typename Item>
-static auto extract(const Item& item, IndexOnly /*unused*/) -> typename ReturnTypeTrait<T, IndexOnly>::ValueType {
+static auto extract(const Item& item, float distance, IndexCoordDistance /*unused*/) ->
```
nonifier: By templating the 3rd parameter's type, you could only keep only 1 `extract` implementation:
```cpp
 template <typename T, typename IndexType, typename Item,>
 static auto extract(const Item& item, float distance) -> 
    typename ReturnTypeTrait<T, IndexType>::ValueType {
    return std::make_tuple(item.element, item.coord, distance);
}
```
So that later, you only need 
```cpp
 res.push_back(extract<T, Tag>(items[res_ind], distances.at(i)));
```

That should also simplify types to only work on `tuple` (rather than mixing pair, tuple and other)

source/proximity_list/proximity_list.cpp
```diff
@@ -185,7 +199,13 @@ auto ProximityList<T>::find_within_impl(const GeographicalCoord& coord,
     std::array<index_t::DistanceType, max_size> distances_data{};
     flann::Matrix<index_t::DistanceType> distances(&distances_data[0], 1, size == -1 ? max_size : size);
     int nb_found = radius_search(NN_index, coord, radius, size, indices, distances);
-    return make_result<T>(coord, items, indices_data, distances_data, nb_found, IndexOnly{});
+
+    std::vector<typename ReturnTypeTrait<T, IndexOnly>::ValueType> res;
+    auto op = [](const Item& item, float /*unused*/) { return item.element; };
+    make_result<T>(coord, items, indices_data, distances_data, nb_found, res, op);
+    ;
```
nonifier: :ghost:

source/proximity_list/proximity_list.cpp
```diff
@@ -125,10 +115,9 @@ static auto make_result(const type::GeographicalCoord& coord,
         LOG4CPLUS_TRACE(
             log4cplus::Logger::getInstance("log"),
             "Distance(squared) from the coord: " << coord.lon() << " " << coord.lat() << " is " << distances.at(i));
-
-        res.push_back(extract<T>(items[res_ind], Tag{}));
+        const auto& item = items[res_ind];
+        out.push_back(op(item, distances.at(i)));
```
nonifier, line 119: :+1:

source/proximity_list/proximity_list.cpp
```diff
@@ -169,7 +158,32 @@ auto ProximityList<T>::find_within_impl(const GeographicalCoord& coord,
...
+
+    return res;
+}
+
+template <class T>
+auto ProximityList<T>::find_within_impl(const GeographicalCoord& coord,
+                                        double radius,
+                                        int size,
+                                        IndexCoordDistance /*unused*/) const
+    -> std::vector<typename ReturnTypeTrait<T, IndexCoordDistance>::ValueType> {
```
nonifier: This function is the same as`find_within_impl` above. 
The same way as previously, if you use `unused` parameter as template parameter, you should be able to remove the version of `find_within_impl` above.

nonifier (approved):

source/proximity_list/proximity_list.cpp
xlqian: oops

source/proximity_list/proximity_list.cpp
xlqian: no, they're not the same... it's actually at dispatch tag for function overload... they're used as "typename" not as "object"

xlqian: i've tried to refactorize these two functions, since you mentioned they're very much alike, it just complicated the codes even more.... I've no magic solution this time :/

nonifier (approved):
Refacto for the win !

benoit-bst (approved):
Is there some tests to cover the new changes ?

PR #3179, opened 2020-03-25
[Jormun] Handle params for sereval possible regions

mbounabi:
https://jira.kisio.org/browse/NAVITIAII-3013
Use base request args for every regions queried

### DIFF
--- a/source/jormungandr/jormungandr/__init__.py
+++ b/source/jormungandr/jormungandr/__init__.py
@@ -104,7 +104,7 @@ i_manager = InstanceManager(
     instance_filename_pattern=app.config.get(str('INSTANCES_FILENAME_PATTERN'), '*.json'),
     start_ping=app.config.get(str('START_MONITORING_THREAD'), True),
 )
-i_manager.initialisation()
+i_manager.initialization()
 
 
 from jormungandr.stat_manager import StatManager
--- a/source/jormungandr/tests/journey_common_tests.py
+++ b/source/jormungandr/tests/journey_common_tests.py
@@ -1612,11 +1612,17 @@ class WithoutPt:
         assert response['debug']['regions_called'][1] == "main_routing_test"
 
 
-@dataset({"main_routing_test": {'max_nb_journeys': 1}})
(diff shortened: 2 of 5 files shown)


source/jormungandr/tests/journey_common_tests.py
```diff
@@ -1612,11 +1612,17 @@ def test_one_region_without_pt_new_default(self):
         assert response['debug']['regions_called'][1] == "main_routing_test"
 
 
-@dataset({"main_routing_test": {'max_nb_journeys': 1}})
+@dataset(
+    {
+        "main_routing_without_pt_test": {"priority": 42, "min_nb_journeys": 10},
+        "main_routing_test": {"max_nb_journeys": 1},
+    }
+)
```
kadhikari, line 1620: If i understand, you added a new dataset to show that even with the lower priority value (0), "main_routing_test" is chosen as it has a solution with PT. As the name indicates main_routing_without_pt_test doesn't have any solution with PT ?

mbounabi: You understood correctly. The dataset with the higher priority will be chosen first and its parameters set.
However, as there's no PT, the request will be done on the second dataset and its parameters will also be set: the result will only 1 journey as the parameter "max_nb_journeys = 1" is set for this coverage only

kadhikari (approved):

source/jormungandr/jormungandr/__init__.py
```diff
@@ -104,7 +104,7 @@
     instance_filename_pattern=app.config.get(str('INSTANCES_FILENAME_PATTERN'), '*.json'),
     start_ping=app.config.get(str('START_MONITORING_THREAD'), True),
 )
-i_manager.initialisation()
+i_manager.initialization()
```
nonifier, line 107: :gb: 0 - :us: 1

nonifier (approved):

kadhikari:
@mbounabi I think you will have to rebase with dev.

PR #3185, opened 2020-03-31
[ALL] Add tidy target in cmake

AurelienLP:
" - Knock Knock.
 - Who's there ?
- Clang
- Clang who ?
- Clang-tidy !"

Hello guys, hope the confinement is going well for all of you !
Since I've been missing you all I figured I would annoy you again with clang tidy !
I've finally found a way to use clang tidy effectively via Cmake héhé! :smile: 

I've used the same flag as discussed here https://github.com/CanalTP/navitia/pull/3070

1) I've created a .clang-tidy file with the command described in the file. Its usage is the same as the .clang-format file.
2) I've created 2 new targets in the cmake `make tidy` and `make tidy_fix`. This commands will check all `.cpp` files of the project excluding the ones in a `tests` directory, `third_party` or `utils`. `make tidy_fix` will automatically modify the files in place if it can.
3) I've applied all the automatic fix.
4) I've applied some manual ones.
5) I've excluded some checks in ` WarningsAsErrors` because there were so much errors (as discussed in https://github.com/CanalTP/navitia/pull/3070 again).
6) I've added some "NOLINT" for some really really bad code...

This idea is to use this in the CI. When an error is detected, the step in the CI should be red.

However, since i don't know how much time it will take to run it and your problematic in the CI, I won't test it yet (and we have to discuss it, I don't want a revert afterwards #REVERT-92).

I'd like to, on a second time, add a target that will run clang-tidy only on the diff with the base so that it'll be faster in the CI.

La bise ! #One-core #Love #vivelecodepropre
:squid:


AurelienLP:
It compiles now ! Have fun with it ! :)

nonifier:
@AurelienLP - Thanks man, this is really appreciated. 

I've been fiddling with you PR and cland-tidy and I have the following question:

1. Which configuration have you been using ? when I run the tool using the `.chang-tidy` file at the root folder I got a lot more errors/warnings that break the codebase when applying the fixes :sob: 
2. It takes sooooo long to run. What did you have in mind to integrate this as part of the CI ? I don't think there is a way to add clang-tidy as part of pre-commit.

AurelienLP:
Hi @nonifier, sorry to check that so late!

1) It should run by itself the same way clang-format does... I should take some times to rebase this PR and try this again.

2) I thought it would be a first step to introduce this in the CI. We should maybe check tools like clang-tidy-diff.py. I've tried that in the past but didn't dig too far unfortunately...

nonifier:
I'll close this PR in favour of https://jira.kisio.org/browse/NAVP-1563.

PR #3189, opened 2020-04-06
[Tyr] Add DELETE to /jobs

mbounabi:
CRUD-Style

### DIFF
--- a/source/tyr/tests/integration/jobs_test.py
+++ b/source/tyr/tests/integration/jobs_test.py
@@ -100,10 +100,10 @@ def test_jobs_deletion(create_instances):
 
     # ---1--- DELETE WITH FILTER BY INSTANCE AND BY STATE
     # delete "running" job for 'Instance_1'
-    api_delete("/v0/jobs/{}?state=running".format("Instance_1"))
+    api_delete("/v0/jobs/{}?confirm=yes&state=running".format("Instance_1"))
 
     # two jobs remaining for 'Instance_1'
-    resp = api_get("/v0/jobs/Instance_1")
+    resp = api_get("/v0/jobs/Instance_1?confirm=yes")
     assert len(resp["jobs"]) == 2
 
     # other instances still have 3 jobs
@@ -112,7 +112,7 @@ def test_jobs_deletion(create_instances):
         assert len(resp["jobs"]) == 3
 
     # --- 2 --- DELETE WITH FILTER BY STATE
-    api_delete("/v0/jobs?state=running")
+    api_delete("/v0/jobs?confirm=yes&state=running")
 
     # two jobs remaining for each instance
     for instance in instances:
@@ -120,7 +120,7 @@ def test_jobs_deletion(create_instances):
         assert len(resp["jobs"]) == 2
 
     # --- 3 --- DELETE WITH FILTER BY INSTANCE
-    api_delete("/v0/jobs/Instance_1")
+    api_delete("/v0/jobs/Instance_1?confirm=yes")
 
     # no job remaining for 'Instance_1'
     resp = api_get("/v0/jobs/Instance_1")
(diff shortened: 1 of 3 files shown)


source/tyr/tyr/resources.py
```diff
@@ -138,6 +138,31 @@ def post(self, instance_name):
...
+        if instance_name:
+            query = query.join(models.Instance).filter(models.Instance.name == instance_name)
+
+        jobs_to_delete = query.all()
+        if not jobs_to_delete:
+            return json.loads('{}'), 204
+
+        for job in jobs_to_delete:
+            try:
+                db.session.delete(job)
```
xlqian: the job is deleted, do we lose the traceability, too? can we add 'delete' to job's state instead of removing it totally from db?

mbounabi: Indeed, traceability is lost. We can add a new value "deleted" for job state:
https://github.com/CanalTP/navitia/blob/60d49021c1f28a16eacb7f11df939ea049c954e8/source/navitiacommon/navitiacommon/models/__init__.py#L749

source/tyr/tyr/resources.py
```diff
@@ -138,6 +138,31 @@ def post(self, instance_name):
...
+    def delete(self, instance_name=None):
+        parser = reqparse.RequestParser()
+        parser.add_argument('state', type=str, required=False, case_sensitive=True, help='filter by job state')
+        args = parser.parse_args()
+
+        query = models.Job.query
+        if "state" in args and args["state"]:
+            query = query.filter_by(state=args["state"])
+        if instance_name:
+            query = query.join(models.Instance).filter(models.Instance.name == instance_name)
```
nonifier: Can't you just delete a job based on it's `id` ? Like if there is only one specific job that you want to be deleted ?

source/tyr/tests/integration/jobs_test.py
```diff
@@ -0,0 +1,143 @@
...
+    resp = api_get("/v0/jobs/Instance_1")
+    assert len(resp["jobs"]) == 0
+
+    # other instances still have 2 jobs
+    for instance in ["Instance_2", "Instance_3"]:
+        resp = api_get("/v0/jobs/{}".format(instance))
+        assert len(resp["jobs"]) == 2
+
+    # --- 4 --- DELETE WITHOUT FILTER - ARMAGEDDON
+    api_delete("/v0/jobs")
```
nonifier: :scream: woooow this is extreme :scream: 
But I get there could be use cases for that. 
with great power

mbounabi: Interesting idea found on a forum:
DELETE all jobs needs a query parameter to be performed:
`DELETE /v0/jobs?confirm=yes`

source/tyr/tyr/resources.py
mbounabi: **To Be Discussed:**
Changing the job state to "deleted" is only relevant for "running" job. "done" and "failed" jobs are finished jobs and won't be taken into account anymore.
In this case, adding an enum might not be relevant, the job "running" can be set to "failed".
Also, changing only the state might need additional operations to actually stop the job from being performed.

Also, to respect CRUD-Style, I think that DELETE should be available for this endpoint.

source/tyr/tyr/resources.py
mbounabi: Done. 
BTW, GET /jobs/<job_id> isn't currently possible. Maybe it's worth adding it (in an other PR)

xlqian (approved):

benoit-bst (approved):

source/tyr/tyr/resources.py
```diff
@@ -138,6 +138,39 @@ def post(self, instance_name):
...
+                try:
+                    db.session.delete(job)
+                except Exception:
+                    logging.exception("Failed to delete Job {} - Operation aborted".format(job.id))
+                    raise
+
+            db.session.commit()
+            return json.loads('{}'), 200
+        else:
+            return {'message': 'confirm the deletion'}, 400
```
nonifier:
```suggestion
            return {'message': 'confirm the deletion with parameter "confirm=yes"'}, 400
```
Being explicit is better not to go back to the source code if you ever want to use the endpoint.

nonifier (approved):

PR #3190, opened 2020-04-07
[Fusio2ed] take into account of stop_code within stops.txt

benoit-bst:
Now, the **stop_code** parameter is read within the **stop.txt** file. 
Each **stop_code** are inserted inside **object_code** with `gtfs_stop_code` key.
To keep the backward compatibility, **object_system/object_code** are always read within **object_codes.txt**
All details in JIRA : https://jira.kisio.org/browse/NAVP-1520


source/ed/connectors/gtfs_parser.cpp
```diff
@@ -287,8 +287,10 @@ void StopsGtfsHandler::init(Data& data) {
     wheelchair_c = csv.get_pos_col("wheelchair_boarding");
     platform_c = csv.get_pos_col("platform_code");
     timezone_c = csv.get_pos_col("stop_timezone");
-    if (code_c == -1) {
+    if (code_c == UNKNOWN_COLUMN) {
```
woshilapin, line 290: Is there a reason why you do not do `csv.get_pos_col("stop_code")`?

source/ed/connectors/gtfs_parser.cpp
```diff
@@ -400,6 +402,7 @@ nm::StopArea* StopsGtfsHandler::build_stop_area(Data& data, const csv_row& row)
     }
     gtfs_data.stop_area_map[sa->uri] = sa;
     data.stop_areas.push_back(sa);
+    add_gtfs_stop_code(data, sa, row[code_c]);
```
woshilapin, line 405: I do not believe adding this `gtfs_stop_code` when it's missing is correct. You do not know if the input NTFS has been generated from GTFS or from another data format.

So far, the object code `gtfs_stop_code` has not been removed on `transit_model` side. We can definitely let it until we're sure it doesn't impact Navitia consumers. The only important thing would be to communicate on Navitia side that it will eventually be replaced by `stop_code` (so consumers can change their code).

nonifier: You might want to add the stop code only if the column exists (otherwise you'll access `row[-1]` :boom: ) 
```suggestion
if(stop_code_is_present) {
    add_gtfs_stop_code(data, sa, row[code_c]);
}
```

source/ed/connectors/gtfs_parser.cpp
```diff
@@ -438,6 +441,8 @@ nm::StopPoint* StopsGtfsHandler::build_stop_point(Data& data, const csv_row& row
     }
     gtfs_data.stop_point_map[sp->uri] = sp;
     data.stop_points.push_back(sp);
+    add_gtfs_stop_code(data, sp, row[code_c]);
```
nonifier: Same here regarding the condition with `if(stop_code_is_present)`

source/ed/connectors/gtfs_parser.h
```diff
@@ -214,20 +216,23 @@ struct DefaultContributorHandler : public GenericHandler {
...
 struct StopsGtfsHandler : public GenericHandler {
-    StopsGtfsHandler(GtfsData& gdata, CsvReader& reader) : GenericHandler(gdata, reader) {}
+    StopsGtfsHandler(GtfsData& gdata, CsvReader& reader) : GenericHandler(gdata, reader), stop_code_is_present(false) {}
     using stop_point_and_area = std::pair<ed::types::StopPoint*, ed::types::StopArea*>;
 
     int id_c, code_c, name_c, desc_c, lat_c, lon_c, zone_c, type_c, parent_c, wheelchair_c, platform_c, timezone_c,
         ext_code_c;
 
     int ignored = 0;
+    bool stop_code_is_present;
```
nonifier, line 226: Minor - if you use ` bool stop_code_is_present = false;` that would save you from initialising it in the constructor.

source/ed/tests/fusioparser_test.cpp
```diff
@@ -543,3 +543,14 @@ BOOST_AUTO_TEST_CASE(stop_time_precision_retrocompatibilty_tests) {
...
+BOOST_AUTO_TEST_CASE(gtfs_stop_code_tests) {
+    ed::Data data;
+    ed::connectors::FusioParser parser(ntfs_path);
+    parser.fill(data);
+    BOOST_REQUIRE_EQUAL(data.if_object_code_exist("stop_code_for_sa", "gtfs_stop_code"), true);
+    BOOST_REQUIRE_EQUAL(data.if_object_code_exist("stop_code_for_sp", "gtfs_stop_code"), true);
+    BOOST_REQUIRE_EQUAL(data.if_object_code_exist("J", "gtfs_stop_code"), true);
+    BOOST_REQUIRE_EQUAL(data.if_object_code_exist("E", "external_code"), true);
+    BOOST_REQUIRE_EQUAL(data.if_object_code_exist("A", "external_code"), true);
+}
```
kadhikari, line 557: Minor: a test with a 'false' may also be added

kadhikari (commented):
It's fine for me with corrections on @nonifier 's comments.

nonifier (approved):

kadhikari (approved):

PR #3201, opened 2020-04-21
[Kraken] Add stop_points_nearby option within places_nearby

benoit-bst:
**Subject** : In order to optimize performance for a car distributed request, we need to update places_nearby (This is the first part P+R quest)

Currently, we need to call a places_nearby to find **all P+R Pois** next to the origin. After that, a bunch of request have to find, for each P+R, the nearby StopPoints. 
So, at the end we do 1 places_nearby call + n places_nearby call for each P+R found.
This work optimize it in 1 call that send directly all informations (A list of P+R with the StopPoint list attached)
The **backward compatibility is keeped** because it is coded like an option.

**Nota**: For the moment this option is not activated. It will be integrated in a future PR

Todo:

- [x] navitia-proto : ~change local branch to master after~ https://github.com/CanalTP/navitia-proto/pull/140 ~is merged. This is the reason why **CI check_submodules** is~ :red_circle: 

**Jira**: https://jira.kisio.org/browse/TRR-101

### DIFF
--- a/source/kraken/worker.cpp
+++ b/source/kraken/worker.cpp
@@ -502,8 +502,13 @@ void Worker::proximity_list(const pbnavitia::PlacesNearbyRequest& request) {
         this->pb_creator.fill_pb_error(pbnavitia::Error::bad_format, e.what());
         return;
     }
+    bool active_stop_points_nearby_options = false;
+    if (request.has_find_stop_points_nearby() && request.find_stop_points_nearby() == true) {
+        active_stop_points_nearby_options = true;
+    }
     proximitylist::find(this->pb_creator, coord, request.distance(), vector_of_pb_types(request), request.filter(),
-                        request.depth(), request.count(), request.start_page(), *data);
+                        request.depth(), request.count(), request.start_page(), *data,
+                        active_stop_points_nearby_options);
 }
 
 static type::StreetNetworkParams streetnetwork_params_of_entry_point(const pbnavitia::StreetNetworkParams& request,
--- a/source/kraken/worker.cpp
+++ b/source/kraken/worker.cpp
@@ -502,10 +502,7 @@ void Worker::proximity_list(const pbnavitia::PlacesNearbyRequest& request) {
         this->pb_creator.fill_pb_error(pbnavitia::Error::bad_format, e.what());
         return;
     }
-    bool active_stop_points_nearby_options = false;
-    if (request.has_find_stop_points_nearby() && request.find_stop_points_nearby() == true) {
-        active_stop_points_nearby_options = true;
-    }
+    bool active_stop_points_nearby_options = request.has_find_stop_points_nearby() && request.find_stop_points_nearby();
     proximitylist::find(this->pb_creator, coord, request.distance(), vector_of_pb_types(request), request.filter(),
                         request.depth(), request.count(), request.start_page(), *data,
                         active_stop_points_nearby_options);
--- a/source/proximity_list/proximitylist_api.cpp
+++ b/source/proximity_list/proximitylist_api.cpp
@@ -80,12 +80,10 @@ static void make_pb(navitia::PbCreator& pb_creator,
                 break;
         }
         // add stop points nearby into response
-        if (!stop_points_nearby_idx_distance.empty()) {
-            for (const auto& sp_idx_distance : stop_points_nearby_idx_distance) {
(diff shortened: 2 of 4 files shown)


source/proximity_list/proximitylist_api.cpp
```diff
@@ -103,14 +112,14 @@ void find(navitia::PbCreator& pb_creator,
...
                 // we'll regenerate the good number in make_pb
-                result.emplace_back(nb_w.second->idx, coord, 0, type);
+                result.emplace_back(nb_w.second->idx, std::move(coord), 0, type, std::move(std::vector<nt::idx_t>()));
                 ++total_result;
             } catch (const proximitylist::NotFound&) {
             }
             continue;
         }
 
-        vector_idx_coord_distance list;
```
benoit-bst, line 113: `list` is a bad name for a var
:see_no_evil:

source/kraken/worker.cpp
```diff
@@ -502,8 +502,13 @@ void Worker::proximity_list(const pbnavitia::PlacesNearbyRequest& request) {
         this->pb_creator.fill_pb_error(pbnavitia::Error::bad_format, e.what());
         return;
     }
+    bool active_stop_points_nearby_options = false;
+    if (request.has_find_stop_points_nearby() && request.find_stop_points_nearby() == true) {
+        active_stop_points_nearby_options = true;
+    }
```
xlqian:
```suggestion
    bool active_stop_points_nearby_options = request.has_find_stop_points_nearby() && request.find_stop_points_nearby();
```

source/proximity_list/proximitylist_api.cpp
```diff
@@ -77,12 +78,19 @@ static void make_pb(navitia::PbCreator& pb_creator,
             default:
                 break;
         }
+        // add stop points nearby into response
+        if (!stop_points_nearby_idx.empty()) {
```
xlqian: the check is not necessary :)

source/proximity_list/proximitylist_api.cpp
```diff
@@ -77,12 +78,19 @@ static void make_pb(navitia::PbCreator& pb_creator,
...
+            for (const auto& sp_idx : stop_points_nearby_idx) {
+                pbnavitia::StopPoint* sp = place->add_stop_points_nearby();
+                pb_creator.fill(data.pt_data->stop_points[sp_idx], sp, depth);
```
xlqian: we also need the distance between the stop_point and the P+R to be able to calculate the crowfly duration later

source/proximity_list/tests/test.cpp
```diff
@@ -368,3 +368,88 @@ BOOST_AUTO_TEST_CASE(test_poi_filter) {
...
+    navitia::type::Data data;
+
+    // origin
+    navitia::type::GeographicalCoord c;
+    c.set_lon(-1.554514);
+    c.set_lat(47.218515);
+
+    // POIs
+    {
+        auto poitype = new navitia::georef::POIType();
```
xlqian: I know this was how other tests were written, I still prefer not seeing `new` in c++14 :)

xlqian:
```suggestion
        auto poitype = std::make_unique<navitia::georef::POIType>();
```

source/proximity_list/tests/test.cpp
```diff
@@ -368,3 +368,88 @@ BOOST_AUTO_TEST_CASE(test_poi_filter) {
...
+        poitype->uri = "poi_type_1";
+        poitype->idx = 0;
+        data.geo_ref->poitypes.push_back(poitype);
```
xlqian:
```suggestion
        data.geo_ref->poitypes.push_back(poitype.get());
```

source/proximity_list/proximitylist_api.cpp
benoit-bst: good point. I will transform `repeated StopPoint stop_points_nearby` in `repeated PtObject stop_points_nearby`
So we will have the **distance** too

source/kraken/worker.cpp
benoit-bst: haha :+1:

xlqian (approved):

kadhikari (approved):
Well tested. 
Functionally good for me.

PR #3210, opened 2020-04-28
[Kraken] Limit stop points nearby for place_nearby by radius option

benoit-bst:
It seems to be important in term of performance to select an other size of radius for **stop_points_nearby**. 
In the **Jormungandr** part, this radius will set to a distance of **5 min duration by walking**. It will allow to limit the searching field.

In extra, fix the distance **from Poi to stop_point**. I forgot the square root :speak_no_evil:

### DIFF
--- a/source/proximity_list/proximitylist_api.cpp
+++ b/source/proximity_list/proximitylist_api.cpp
@@ -83,7 +83,7 @@ static void make_pb(navitia::PbCreator& pb_creator,
         for (const auto& sp_idx_distance : stop_points_nearby_idx_distance) {
             pbnavitia::PtObject* pt_obj = place->add_stop_points_nearby();
             pb_creator.fill(data.pt_data->stop_points[std::get<0>(sp_idx_distance)], pt_obj, depth);
-            pt_obj->set_distance(std::get<1>(sp_idx_distance));
+            pt_obj->set_distance(sqrt(std::get<1>(sp_idx_distance)));
         }
     }
 }
--- a/source/proximity_list/proximitylist_api.cpp
+++ b/source/proximity_list/proximitylist_api.cpp
@@ -102,7 +102,7 @@ void find(navitia::PbCreator& pb_creator,
           const uint32_t count,
           const uint32_t start_page,
           const type::Data& data,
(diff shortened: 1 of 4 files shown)


source/proximity_list/proximitylist_api.cpp
```diff
@@ -83,7 +83,7 @@ static void make_pb(navitia::PbCreator& pb_creator,
         for (const auto& sp_idx_distance : stop_points_nearby_idx_distance) {
             pbnavitia::PtObject* pt_obj = place->add_stop_points_nearby();
             pb_creator.fill(data.pt_data->stop_points[std::get<0>(sp_idx_distance)], pt_obj, depth);
-            pt_obj->set_distance(std::get<1>(sp_idx_distance));
+            pt_obj->set_distance(sqrt(std::get<1>(sp_idx_distance)));
```
benoit-bst, line 86: the forgotten sqrt

source/proximity_list/tests/test.cpp
```diff
@@ -428,21 +428,21 @@ BOOST_AUTO_TEST_CASE(test_find_stop_points_nearby_options) {
...
+    find(pb_creator, c, 15, {navitia::type::Type_e::POI}, "poi_type.uri=poi_type_2", 1, 10, 0, data, 15);
     auto result = pb_creator.get_response();
 
     // Only poi_2 is available (poi_type_2)
     BOOST_REQUIRE_EQUAL(result.places_nearby().size(), 1);
     BOOST_CHECK(result.places_nearby(0).uri() == "poi_2");
     BOOST_REQUIRE_EQUAL(result.places_nearby(0).stop_points_nearby().size(), 2);
     BOOST_REQUIRE_EQUAL(result.places_nearby(0).stop_points_nearby(0).uri(), "sp_1");
-    BOOST_REQUIRE_EQUAL(result.places_nearby(0).stop_points_nearby(0).distance(), 2);
+    BOOST_REQUIRE_EQUAL(result.places_nearby(0).stop_points_nearby(0).distance(), 1);
```
benoit-bst, line 439: change test results with the sqrt

xlqian (approved):

PR #3218, opened 2020-05-05
[CI] Add clang-tidy as part of our CI's worflow

nonifier:
BEFORE builds are being performed, we run clang-tidy on the c++ source
code and apply all changes onto the current branch.

This way, we keep the code tidy and formated before it's being built and
tested.

== Clang-Tidy as introduced changes that break my code....
> If this happens, run clang-tidy localy using:
```
make tidy_fix
```
> Fix the code and push it all to you local branch !

== Hang on a minute ! Why aren't we using pre-commit ?
> Clang tidy uses LLVM frontend to parse and understand the code which
 makes clang-tidy really slow... But it's what makes clang-tidy so powerful
 aas well.
> Not to bloat pre-commit with a unbearable waiting time, the process is
done in the CI side.

== What are we checking against ?
> So far, we only check `clang-analyzwe` to validate the workflow.
> If we are fine with it, we could alway add more interesting checks
like : `mordernize-*, cppcoreguidelines-*, readability-*` etc...
> I wanted this to be an iterative process not to introduce to many
changes in one time.

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -1,0 +1,6 @@
+Checks:          '-*, clang-analyzer-*'
+HeaderFilterRegex: '*'
+AnalyzeTemporaryDtors: false
+FormatStyle:     file
+User:            Navitia
+CheckOptions:
--- a/.github/workflows/workflow.yml
+++ b/.github/workflows/workflow.yml
@@ -11,7 +11,15 @@ on:
 
 
 jobs:
-  check_submodules:
+  info:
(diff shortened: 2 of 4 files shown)


AurelienLP:
Nice! 

I've tried to do the other way around, adding all exceptions and removing them in the future but it's seems easier this way! 

Since you've used the cmake part of my PR, don't hesitate to close it if it stays open for too long!

#one-core

PR #3220, opened 2020-05-07
[Tyr] update virtualenv requirements

pbench:
Tyr virtualenv requirements now includes eitri requirements (otherwise a package was missing when running docker_test)

### DIFF
--- a/source/tyr/requirements_dev.txt
+++ b/source/tyr/requirements_dev.txt
@@ -1,5 +1,6 @@
 -r requirements.txt
 -r ../../requirements_pre-commit.txt
+-r ../eitri/requirements.txt
 
 docker==3.4.1
 pytest==4.6.5


kadhikari (approved):

source/tyr/requirements_dev.txt
```diff
@@ -1,5 +1,6 @@
 -r requirements.txt
 -r ../../requirements_pre-commit.txt
+-r ../eitri/requirements.txt
```
nonifier, line 3: I'm not keen en tiding up tyr with eitri's requirements are they are 2 distinct projects and have no real reason to rely en each other. 

The `docker_tests` are part of the CI and are passing  (see https://github.com/CanalTP/navitia/blob/dev/.github/workflows/workflow.yml#L80-L85). 

Where did you experience the problem ? I don't see why you would need this :thinking:

pbench: I tried running docker_test in the `tyr` environment built from tyr/requirements_dev.
The package GeoAlchemy was missing, and the tests were failing.
I can add only the missing package to the tyr requirements if you prefer.
Or maybe this was the wrong environment to run the tests in ?

pbench: Maybe the CI works because before running `docker_tests` we 
`pip install -r source/sql/requirements.txt`

nonifier: I agree that a dependency should be missing somewhere. But I don't think this is the right solution. 

As you mentioned, the `docker_test` are not meant to be run within the `tyr` virtualenv. 
Maybe a `requirements.txt` that includes tyr and sql is missing somewhere to help run the docker tests.

pbougue: The makefile ties them too closely (or/and names them badly).
You can use `make tyr_tests` for the tyr part, and `make run_tests` for the eitri part :-/

xlqian: I agree with @pbougue ,  docker_tests should be run with `make` command, which is simpler

pbench:
I close for now, more discussion is needed on how to proceed

PR #3224, opened 2020-05-20
[Jorm] get_realtime_proxy on non-realtime object causing 'no handler found'

nonifier:
We were casting a `None` object into a string (u"None") causing it to fail the
check condition.

This is raising a "no handle found" message to new-relic whereas no
realtime proxy is actually configured, hence generating a false-positive error.

fun fact: This is raising 10K `realtime_internal_failure` a minute on new relic. 

Kudos to @SGrenet for spotting it.


mbounabi (approved):

SGrenet (approved):

kadhikari (approved):
:+1:

benoit-bst (approved):

PR #3227, opened 2020-05-28
Purge instance after each bina

kadhikari:
At the end of import_data, we should clean unnecessary backup data files. 
Note: 

- Not easy to test the function import_data
- Added task 'purge_instance' is already well tested and only deletes delete able files. 
- Tested in local with tyr environment.
Ticket: https://jira.kisio.org/browse/NAVP-1567

### DIFF
--- a/source/tyr/tests/integration/purge_test.py
+++ b/source/tyr/tests/integration/purge_test.py
@@ -298,6 +298,8 @@ def test_purge_instance_jobs():
     """
     Delete old jobs created before the time limit
     Do not delete jobs with state = 'running'
+    Deletes only directories of delete able jobs but not in the table 'job'
+    Is used by the job 'purge_datasets' planned daily to purge old jobs
     """
     app.config['JOB_MAX_PERIOD_TO_KEEP'] = 1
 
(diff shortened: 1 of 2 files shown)


source/tyr/tyr/tasks.py
```diff
@@ -169,6 +169,9 @@ def import_data(
             current_app.logger.info("skipping mimir import")
 
         actions.append(finish_job.si(job.id))
+
+        # We should delete old backup directories related to this instance
+        actions.append(purge_instance.si(instance.id, current_app.config['DATASET_MAX_BACKUPS_TO_KEEP']))
```
nonifier, line 174: this is going to greatly improve our resilience :+1:

nonifier: Do you think the CRON job is still needed ? Should we remove it at the same time ?

nonifier (approved):

source/tyr/tyr/tasks.py
kadhikari: We can leave for some time and remove it after a verification.

source/tyr/tests/integration/purge_test.py
```diff
@@ -298,6 +298,8 @@ def test_purge_instance_jobs():
     """
     Delete old jobs created before the time limit
     Do not delete jobs with state = 'running'
+    Deletes only directories of delete able jobs but not in the table 'job'
```
mbounabi, line 301:
```suggestion
    Deletes only directories of deletable jobs but not in the table 'job'
```

mbounabi (approved):

PR #3229, opened 2020-06-04
[Jormun] crowfly replaces street network section when the computation has failed

benoit-bst:
### Topic

Nowaday, in **Distributed** scenario, if the street network computation return an empty response, an exception occurs, `no origin nor destination`
To be consistent with the **New Default** scenario, we have to create a crowfly section that replaces street network.

It will fix the **JIRA** https://jira.kisio.org/browse/NAVP-1571

### DIFF
--- a/source/jormungandr/tests/direct_path_asgard_integration_tests.py
+++ b/source/jormungandr/tests/direct_path_asgard_integration_tests.py
@@ -216,6 +216,8 @@ class TestAsgardDirectPath(AbstractTestFixture):
         assert response['journeys'][2]['duration'] == 2000
         assert not response['journeys'][2]['sections'][0].get('cycle_lane_length')
 
+        assert not response.get('feed_publishers')
+
 
 @dataset(
     {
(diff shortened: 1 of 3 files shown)


source/jormungandr/tests/direct_path_asgard_integration_tests.py
```diff
@@ -114,13 +127,17 @@ def route_response(mode):
     return response
 
 
-def response_valid(request):
+def valid_response(request):
```
nonifier, line 130: :gb: :+1:

source/jormungandr/tests/direct_path_asgard_integration_tests.py
```diff
@@ -182,4 +216,28 @@ def test_journey_with_bike_direct_path(self):
         assert response['journeys'][2]['duration'] == 2000
         assert not response['journeys'][2]['sections'][0].get('cycle_lane_length')
 
-        assert not response.get('feed_publishers')
```
nonifier, line 185: Should this assert really be removed ?

source/jormungandr/tests/direct_path_asgard_integration_tests.py
```diff
@@ -182,4 +216,28 @@ def test_journey_with_bike_direct_path(self):
...
+               In this case, we replace the lost street network section by a crowfly, like in New Default
+        """
+        response, status = self.query_region(journey_basic_query, check=False)
+        assert status == 200
+        check_journeys(response)
+
+        assert len(response['journeys']) == 1
+
+        assert len(response['journeys'][0]['sections']) == 3
+        assert response['journeys'][0]['sections'][0]['type'] == 'crow_fly'  # replace the street network
```
nonifier: to be even more expressive and leave a hint if the test fails, you can decorate your assertion with an error message :
```suggestion
        assert response['journeys'][0]['sections'][0]['type'] == 'crow_fly', "A crow fly should replace the street network"
```

But that's only if you want to be fancy :)

nonifier (approved):

xlqian (approved):

PR #3231, opened 2020-06-16
[Jormun] Set distributed scenario on call to kraken.

nonifier:
This will ensure reporting is correctly handled when an instance has its
scenario configured from the database. Otherwise, this would only be set
when overriding the parameter using `_override_scenario`


benoit-bst (approved):
meowwww

xlqian (approved):

PR #3232, opened 2020-06-17
[Jormun] Add Here street network connector

benoit-bst:
# Integrate Here street network within Navitia

Now, **Here** works with navitia. It allows car sections with **realtime traffic** information.
The integration follows the **Here** developer documentation : https://developer.here.com/documentation/routing/dev_guide/topics/routing-mode-parameter-combinations.html

A bit of documentation was done : https://github.com/benoit-bst/navitia/blob/ef0302e32e5475bf1dd42042f2182ff8af105bb8/source/jormungandr/jormungandr/street_network/street_networks.md
The **configuration** is exposed, and the **debug method** too

For the moment, we can only activate the feature on the entire coverage with the **jormun configuration file** or the **DB**.

## Tests

An important difference with Kraken street network has to be highlighted. 
**Here** is capable of compute 100 matrix points (The max value). It is a huge difference with Kraken, because it allows thousands of matrix points !! (so thousands of potential available SP in _car_no_park_ mode). These possibilities to **Car+TC journeys** are better. 
You can test the difference with a request + `first_section_mode=car_no_park`. Sometimes **Kraken** offers an solution and **here** not...

Even if kraken has more **Car+TC solutions**, the results have a poor quality (street network are too rough). 
On the other hand, with **Here SN**, the car section is very appreciable (Realtime traffic is a gift :gift:)

## performance

I roughly tested with few requests and it shows that:

### for a simple direct path
```
Request : 
http://localhost:5000/v1/coverage/transilien/journeys?from=2.20794%3B48.89901&to=2.39166%3B48.84661&_override_scenario=distributed&first_section_mode%5B%5D=car_no_park&_min_car=0&_min_car_no_park=0&direct_path=only&
```
Kraken : 250 ms
Here : 600 ms
**Ratio** : +60% for Here

### for a Car+TC
```
Request : 
http://localhost:5000/v1/coverage/transilien/journeys?from=2.14408%3B48.89812&to=2.33443%3B48.84189&_override_scenario=distributed&first_section_mode%5B%5D=car_no_park&_min_car=0&_min_car_no_park=0
```
Kraken : 450 ms
Here : 1200 ms
**Ratio** : +62% for Here

Considering the results, we need to find bottlenecks to improve the speed if it is possible.

### DIFF
--- a/source/jormungandr/jormungandr/interfaces/v1/Journeys.py
+++ b/source/jormungandr/jormungandr/interfaces/v1/Journeys.py
@@ -383,6 +383,24 @@ class Journeys(JourneyCommon):
             help="how long it takes to park the car, " "used especially in distributed scenario",
         )
         parser_get.add_argument(
+            "_here_realtime_traffic",
+            type=six.text_type,
+            hidden=True,
+            help="Here, Active or not the realtime traffic information (enabled/disabled)",
+        )
+        parser_get.add_argument(
+            "_here_matrix_type",
+            type=six.text_type,
+            hidden=True,
+            help="Here, street network matrix type (simple_matrix/multi_direct_path)",
+        )
+        parser_get.add_argument(
+            "_here_max_matrix_points",
+            type=six.text_type,
+            hidden=True,
+            help="Here, Max number of matrix points for the street network computation (limited to 100)",
+        )
+        parser_get.add_argument(
             "equipment_details",
             default=True,
             type=BooleanType(),
--- a/source/jormungandr/jormungandr/interfaces/v1/Journeys.py
+++ b/source/jormungandr/jormungandr/interfaces/v1/Journeys.py
@@ -384,9 +384,10 @@ class Journeys(JourneyCommon):
         )
         parser_get.add_argument(
             "_here_realtime_traffic",
-            type=six.text_type,
+            type=BooleanType(),
+            default=True,
             hidden=True,
-            help="Here, Active or not the realtime traffic information (enabled/disabled)",
+            help="Here, Active or not the realtime traffic information (True/False)",
         )
         parser_get.add_argument(
             "_here_matrix_type",
@@ -396,7 +397,8 @@ class Journeys(JourneyCommon):
         )
         parser_get.add_argument(
             "_here_max_matrix_points",
-            type=six.text_type,
+            type=int,
+            default=default_values.here_max_matrix_points,
             hidden=True,
             help="Here, Max number of matrix points for the street network computation (limited to 100)",
         )
--- a/source/jormungandr/jormungandr/street_network/here.py
+++ b/source/jormungandr/jormungandr/street_network/here.py
@@ -1,4 +1,4 @@
-#  Copyright (c) 2001-2016, Canal TP and/or its affiliates. All rights reserved.
+#  Copyright (c) 2001-2020, Canal TP and/or its affiliates. All rights reserved.
 #
 # This file is part of Navitia,
 #     the software to build cool stuff with public transport.
@@ -36,11 +36,26 @@ import pybreaker
 import requests as requests
 from jormungandr import app
 from jormungandr.exceptions import TechnicalError
-from jormungandr.utils import get_pt_object_coord
+from jormungandr.utils import get_pt_object_coord, PeriodExtremity
 from jormungandr.street_network.street_network import AbstractStreetNetworkService, StreetNetworkPathKey
 from jormungandr.ptref import FeedPublisher
 from jormungandr.fallback_modes import FallbackModes as fm
 from six import text_type
+from enum import Enum
+
+MAX_MATRIX_POINTS = 100  # The limit is fixed by Here API
+
+# Possible values to active/deactivate realtime traffic
+class RealTimeTraffic(Enum):
+    enabled = "enabled"
+    disabled = "disabled"
+
+
+# Possible values to choose a matrix method
+class MatrixType(Enum):
+    simple_matrix = "simple_matrix"
+    multi_direct_path = "multi_direct_path"
+
 
 DEFAULT_HERE_FEED_PUBLISHER = {
     'id': 'here',
@@ -83,8 +98,10 @@ class Here(AbstractStreetNetworkService):
         modes=[],
         id='here',
         timeout=10,
-        api_id=None,
-        api_code=None,
+        apiKey=None,
+        matrix_type=MatrixType.simple_matrix.value,
+        max_matrix_points=MAX_MATRIX_POINTS,
+        realtime_traffic=RealTimeTraffic.enabled.value,
         feed_publisher=DEFAULT_HERE_FEED_PUBLISHER,
         **kwargs
     ):
@@ -97,17 +114,50 @@ class Here(AbstractStreetNetworkService):
         self.matrix_service_url = 'https://matrix.{base_url}/calculatematrix.json'.format(
             base_url=service_base_url
         )
-        self.api_id = api_id
-        self.api_code = api_code
+        self.log = logging.LoggerAdapter(logging.getLogger(__name__), extra={'streetnetwork_id': text_type(id)})
+        self.apiKey = apiKey
         self.modes = modes
         self.timeout = timeout
-        self.max_points = 100  # max number of point asked in the routing matrix
+        self.matrix_type = matrix_type
+        if (
+            self.matrix_type != MatrixType.simple_matrix.value
+            and self.matrix_type != MatrixType.multi_direct_path.value
+        ):
+            self.log.error(
+                'Here confifguration matrix_type={} is badly formatted (option simple_matrix/multi_direct_path) - force to simple_matrix'.format(
+                    self.matrix_type
+                )
+            )
+            self.matrix_type = MatrixType.simple_matrix.value
+        self.max_matrix_points = (
+            max_matrix_points
+        )  # max number of allowed point for the routing matrix computation
+        if int(max_matrix_points) > MAX_MATRIX_POINTS:
+            self.log.error(
+                'Here confifguration max_matrix_points={} is > 100. Force to 100'.format(max_matrix_points)
+            )
+            self.max_matrix_points = MAX_MATRIX_POINTS
+        self.realtime_traffic = realtime_traffic
+        if (
+            self.realtime_traffic != RealTimeTraffic.enabled.value
+            and self.realtime_traffic != RealTimeTraffic.disabled.value
+        ):
+            self.log.error(
+                'Here confifguration realtime_traffic={} is badly formatted (option enabled/disabled) - force to disabled'.format(
+                    self.realtime_traffic
+                )
+            )
+            self.realtime_traffic = RealTimeTraffic.disabled.value
         self.breaker = pybreaker.CircuitBreaker(
             fail_max=app.config['CIRCUIT_BREAKER_MAX_HERE_FAIL'],
             reset_timeout=app.config['CIRCUIT_BREAKER_HERE_TIMEOUT_S'],
         )
 
-        self.log = logging.LoggerAdapter(logging.getLogger(__name__), extra={'streetnetwork_id': text_type(id)})
+        self.log.debug(
+            'Here, load confifguration max_matrix_points={} - matrix_type={} - realtime_traffic={}'.format(
+                self.max_matrix_points, self.matrix_type, self.realtime_traffic
+            )
+        )
         self._feed_publisher = FeedPublisher(**feed_publisher) if feed_publisher else None
 
     def status(self):
@@ -116,7 +166,9 @@ class Here(AbstractStreetNetworkService):
             'class': self.__class__.__name__,
             'modes': self.modes,
             'timeout': self.timeout,
-            'max_points': self.max_points,
+            'matrix_type': self.matrix_type,
+            'max_matrix_points': self.max_matrix_points,
+            'realtime_traffic': self.realtime_traffic,
             'circuit_breaker': {
                 'current_state': self.breaker.current_state,
                 'fail_counter': self.breaker.fail_counter,
@@ -131,17 +183,17 @@ class Here(AbstractStreetNetworkService):
             self.record_call('ok')
             return r
         except pybreaker.CircuitBreakerError as e:
-            self.log.error('Valhalla routing service dead (error: {})'.format(e))
+            self.log.error('Here routing service dead (error: {})'.format(e))
             self.record_call('failure', reason='circuit breaker open')
             raise TechnicalError('HERE service not available')
         except requests.Timeout as t:
-            self.log.error('Valhalla routing service dead (error: {})'.format(t))
+            self.log.error('Here routing service dead (error: {})'.format(t))
             self.record_call('failure', reason='timeout')
             raise TechnicalError('impossible to access HERE service, timeout reached')
         except Exception as e:
-            self.log.exception('Valhalla routing error')
+            self.log.exception('Here routing error')
             self.record_call('failure', reason=str(e))
-            raise TechnicalError('impossible to access HERE service')
+            raise TechnicalError('impossible to access Here service')
 
     @staticmethod
     def _read_response(response, origin, destination, mode, fallback_extremity, request):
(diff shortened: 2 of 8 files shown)


source/jormungandr/jormungandr/interfaces/v1/Journeys.py
```diff
@@ -382,6 +382,24 @@ def __init__(self):
             hidden=True,
             help="how long it takes to park the car, " "used especially in distributed scenario",
         )
+        parser_get.add_argument(
+            "_here_realtime_traffic",
+            type=six.text_type,
```
kadhikari: The code could be much simpler if we use type=BooleanType() with default=True or False non?

source/jormungandr/jormungandr/interfaces/v1/Journeys.py
```diff
@@ -382,6 +382,24 @@ def __init__(self):
...
+        )
+        parser_get.add_argument(
+            "_here_matrix_type",
+            type=six.text_type,
+            hidden=True,
+            help="Here, street network matrix type (simple_matrix/multi_direct_path)",
+        )
+        parser_get.add_argument(
+            "_here_max_matrix_points",
+            type=six.text_type,
```
kadhikari: Same as above with type=int and controlling de upper limit value in https://github.com/CanalTP/navitia/blob/dev/source/navitiacommon/navitiacommon/default_values.py

source/jormungandr/jormungandr/street_network/here.py
```diff
@@ -36,11 +36,26 @@
...
 from jormungandr.exceptions import TechnicalError
-from jormungandr.utils import get_pt_object_coord
+from jormungandr.utils import get_pt_object_coord, PeriodExtremity
 from jormungandr.street_network.street_network import AbstractStreetNetworkService, StreetNetworkPathKey
 from jormungandr.ptref import FeedPublisher
 from jormungandr.fallback_modes import FallbackModes as fm
 from six import text_type
+from enum import Enum
+
+MAX_MATRIX_POINTS = 100  # The limit is fixed by Here API
```
nonifier: If this is only relevant for Here, it should a member attribute to the Here class.

source/jormungandr/jormungandr/street_network/here.py
```diff
@@ -97,17 +114,50 @@ def __init__(
...
+        self.apiKey = apiKey
         self.modes = modes
         self.timeout = timeout
-        self.max_points = 100  # max number of point asked in the routing matrix
+        self.matrix_type = matrix_type
+        if (
+            self.matrix_type != MatrixType.simple_matrix.value
+            and self.matrix_type != MatrixType.multi_direct_path.value
+        ):
+            self.log.error(
```
nonifier: You could just simply write

```suggestion
        if matrix_type in MatrixType:
            self.matrix_type = matrix_type
        else:
            self.log.error(
```

source/jormungandr/jormungandr/interfaces/v1/Journeys.py
```diff
@@ -382,6 +382,24 @@ def __init__(self):
             hidden=True,
             help="how long it takes to park the car, " "used especially in distributed scenario",
         )
+        parser_get.add_argument(
+            "_here_realtime_traffic",
```
nonifier, line 386: If think it would be better not to mentioned `here` in the input parameters as they could be used for other connectors in the near future. 
```suggestion
            "_realtime_traffic",
```
Same applies for the above new param.

source/jormungandr/jormungandr/street_network/here.py
```diff
@@ -36,11 +36,26 @@
...
+
+MAX_MATRIX_POINTS = 100  # The limit is fixed by Here API
+
+# Possible values to active/deactivate realtime traffic
+class RealTimeTraffic(Enum):
+    enabled = "enabled"
+    disabled = "disabled"
+
+
+# Possible values to choose a matrix method
```
pbench, line 63: Maybe document here (or somewhere else) what these two options means (not obvious at first)

source/jormungandr/jormungandr/street_network/here.py
```diff
@@ -299,15 +400,36 @@ def _get_matrix(cls, json_response, origins, destinations):
...
+        routes = json_response.get('response', {}).get('route', [])
+        if not routes:
+            return row
+        route = routes[0]
+        travel_time = route.get('summary', {}).get('travelTime')
+        if travel_time:
+            routing.duration = travel_time
+            routing.routing_status = response_pb2.reached
+        else:
+            # since we limit the number of points given to HERE, we say that all the others cannot be
```
pbench: Does this comment applies here ? There is not limit on the number of points for a direct path query right ?

source/jormungandr/jormungandr/street_network/here.py
```diff
@@ -318,32 +440,73 @@ def get_matrix_params(self, origins, destinations, mode, max_duration, request):
...
+        for origin in origins[: int(max_matrix_points)]:
+            for destination in destinations[: int(max_matrix_points)]:
+                params = self.get_direct_path_params(
+                    origin,
+                    destination,
+                    mode,
+                    PeriodExtremity(datetime=request['datetime'], represents_start=request['clockwise']),
+                    realtime_traffic,
+                )
+                r = self._call_here(self.routing_service_url, params=params)
```
pbench, line 477: If I'm not mistaken, each call is made sequentially.
Is it possible to make them concurrently to speed things up ?

pbench (approved):
Small comments/ideas, looks good otherwise :+1:

source/jormungandr/jormungandr/street_network/here.py
```diff
@@ -222,22 +274,67 @@ def _read_response(response, origin, destination, mode, fallback_extremity, requ
 
         return resp
 
-    def get_direct_path_params(self, origin, destination, mode, fallback_extremity):
+    def get_realtime_traffic_parameter(self, request):
```
nonifier, line 282: This logic is exactly the same as the one in the Here class' constructor. Instead of repeating yourself, you can re-use this function up there by refactoring it a little :

```suggestion
    def get_realtime_traffic_parameter(self, realtime_traffic):
        [...]
```

source/jormungandr/jormungandr/street_network/here.py
```diff
@@ -222,22 +274,67 @@ def _read_response(response, origin, destination, mode, fallback_extremity, requ
...
+        _realtime_traffic = request.get('_here_realtime_traffic', None)
+        if _realtime_traffic == None:
+            return self.realtime_traffic
+        if (
+            _realtime_traffic != None
```
nonifier: this is not needed as you have already returned `if _realtime_traffic == None` on the line above

source/jormungandr/jormungandr/street_network/here.py
```diff
@@ -318,32 +440,73 @@ def get_matrix_params(self, origins, destinations, mode, max_duration, request):
...
+                    mode,
+                    PeriodExtremity(datetime=request['datetime'], represents_start=request['clockwise']),
+                    realtime_traffic,
+                )
+                r = self._call_here(self.routing_service_url, params=params)
+                if r.status_code != 200:
+                    self.log.error(
+                        'impossible to query HERE: {} with {} response: {}'.format(r.url, r.status_code, r.text)
+                    )
+                    raise TechnicalError('invalid HERE call, impossible to query: {}'.format(r.url))
```
nonifier: `Requests` comes with helper functions for that - https://requests.readthedocs.io/en/master/user/quickstart/#response-status-codes

source/jormungandr/jormungandr/street_network/street_networks.md
```diff
@@ -0,0 +1,48 @@
...
+
+## Here
+
+Provide a external API that delivers street network solution with realtime traffic informations.<br>
+Navitia uses it to perform car sections.<br>
+Here documentation : https://developer.here.com/
+
+### How to connect Here with Navitia.
+
+Insert into **Jormun** configuration:
```
nonifier, line 11: :+1:

source/jormungandr/jormungandr/street_network/here.py
benoit-bst: Exactly, this comment is an evil copy/paste ! 
Nice catch

source/jormungandr/jormungandr/street_network/here.py
benoit-bst: yes, this is a step by step calling method. This is not efficient. 
But since we have choose to call the matrix Here API... this method is not used.
I added a TODO for the future to propose the parallelize method.

source/jormungandr/jormungandr/street_network/here.py
benoit-bst: @kadhikari proposed to add this little cute inside the default_value.py file. I added it

source/jormungandr/jormungandr/interfaces/v1/Journeys.py
benoit-bst: I did it on purpose to prefix with Here, in case of multi street network core, like Asgard for the bike and Here for the car. This is just to avoid collision between identical parameters.
I see the future :mage_man: 
Moreover, I created this parameter only for debug or demonstration purpose. Yes I think about @SGrenet :cat:

source/jormungandr/jormungandr/street_network/here.py
benoit-bst: Finally, I changed the method with a **true/false** value from @kadhikari remark

source/jormungandr/jormungandr/street_network/here.py
benoit-bst: hooo sweet

source/jormungandr/jormungandr/street_network/here.py
benoit-bst: whoo nice

source/jormungandr/jormungandr/interfaces/v1/Journeys.py
xlqian: I don't how this should be handled.... For the moment, we can only limite users' access to certain `coverage` by their token, but not the feature... I mean we cannot close Here for user A and open Here for user B on the same coverage... there is probably something to do **on the authentication level**...  For me, and this parameter should be defaulted to `true`, we can trun off this feature if user's token doesn't allow the usage...

source/jormungandr/jormungandr/interfaces/v1/Journeys.py
```diff
@@ -382,6 +382,24 @@ def __init__(self):
...
+            help="Here, Active or not the realtime traffic information (enabled/disabled)",
+        )
+        parser_get.add_argument(
+            "_here_matrix_type",
+            type=six.text_type,
+            hidden=True,
+            help="Here, street network matrix type (simple_matrix/multi_direct_path)",
+        )
+        parser_get.add_argument(
+            "_here_max_matrix_points",
```
xlqian, line 399: this parameter is actually defined by https://github.com/CanalTP/navitia/blob/dev/source/jormungandr/jormungandr/interfaces/v1/Journeys.py#L365, I think you don't have to set here...

source/jormungandr/jormungandr/street_network/here.py
```diff
@@ -299,15 +400,36 @@ def _get_matrix(cls, json_response, origins, destinations):
 
         return sn_routing_matrix
 
-    def get_matrix_params(self, origins, destinations, mode, max_duration, request):
+    @classmethod
+    def _create_matrix_response_with_direct_path(cls, json_response, origin, destination, row):
```
xlqian: this function is not a classmethod but a staticmethod

source/jormungandr/jormungandr/street_network/here.py
```diff
@@ -318,32 +440,73 @@ def get_matrix_params(self, origins, destinations, mode, max_duration, request):
         datetime = request['datetime']
         params['departure'] = _str_to_dt(datetime)
 
-        # Note: for the moment we limit to self.max_points the number of n:m asked
-        for i, o in enumerate(origins[: self.max_points]):
+        for i, o in enumerate(origins[: int(max_matrix_points)]):
```
xlqian: just a little python tip, the operator `[]` in python is a simple copy, it will make allocations.. you can use `islice`... for the sake for performance...  
```python
import itertools
for i, o in enumerate(itertools.islice(origins, max_matrix_points)):
```
in this case, since the origins and destinations are of small size, it won't change anything...

source/jormungandr/jormungandr/street_network/here.py
```diff
@@ -299,15 +400,36 @@ def _get_matrix(cls, json_response, origins, destinations):
...
-    def get_matrix_params(self, origins, destinations, mode, max_duration, request):
+    @classmethod
+    def _create_matrix_response_with_direct_path(cls, json_response, origin, destination, row):
+        routing = row.routing_response.add()
+        routes = json_response.get('response', {}).get('route', [])
+        if not routes:
+            return row
+        route = routes[0]
+        travel_time = route.get('summary', {}).get('travelTime')
+        if travel_time:
```
xlqian: can `travel_time` be 0?
`if travel_time is not None:`

source/jormungandr/jormungandr/street_network/here.py
xlqian: `origin, destination` are not used

source/jormungandr/jormungandr/street_network/here.py
```diff
@@ -97,17 +114,50 @@ def __init__(
...
+            self.log.error(
+                'Here confifguration matrix_type={} is badly formatted (option simple_matrix/multi_direct_path) - force to simple_matrix'.format(
+                    self.matrix_type
+                )
+            )
+            self.matrix_type = MatrixType.simple_matrix.value
+        self.max_matrix_points = (
+            max_matrix_points
+        )  # max number of allowed point for the routing matrix computation
+        if int(max_matrix_points) > MAX_MATRIX_POINTS:
```
xlqian: why this need to be casted to int?

source/jormungandr/jormungandr/street_network/here.py
```diff
@@ -36,11 +36,26 @@
...
 from jormungandr.ptref import FeedPublisher
 from jormungandr.fallback_modes import FallbackModes as fm
 from six import text_type
+from enum import Enum
+
+MAX_MATRIX_POINTS = 100  # The limit is fixed by Here API
+
+# Possible values to active/deactivate realtime traffic
+class RealTimeTraffic(Enum):
+    enabled = "enabled"
```
xlqian, line 49: :heart:

source/jormungandr/jormungandr/street_network/here.py
benoit-bst: it was not useful.

source/jormungandr/jormungandr/interfaces/v1/Journeys.py
benoit-bst: I created this parameter, with the prefix _here, for debug purpose or demonstration.
It is specific to Here because it is limited between 0 and 100

source/jormungandr/jormungandr/street_network/here.py
benoit-bst: oops this is an evil copy/paste

source/jormungandr/jormungandr/street_network/here.py
benoit-bst: **origins** or **destinatons** are smaller than 100, but I like the tip.
thanks :1st_place_medal:

source/jormungandr/jormungandr/street_network/here.py
benoit-bst: :+1:

benoit-bst:
@xlqian 

> I don't how this should be handled.... For the moment, we can only limite users' access to certain coverage by their token, but not the feature... I mean we cannot close Here for user A and open Here for user B on the same coverage... there is probably something to do on the authentication level... For me, and this parameter should be defaulted to true, we can trun off this feature if user's token doesn't allow the usage...

I understand your point of view. 
For the moment we don't have the opportunity to select a **token** for this feature. 
we can only activate realtime traffic for an entire coverage...
However, I created the `_here_realtime_traffic` param only for debug or demonstration purpose. This is not a way to by pass the token way :smile:

source/jormungandr/jormungandr/interfaces/v1/Journeys.py
xlqian: if it's only for debug purpose, I'm Ok with that. What I meant is if the user set `_max_nb_crowfly_by_car` to 50, the connector **should** take 50 into consideration

nonifier (approved):
So much nicer  :+1: 
I'm not a massive fan of the controversial `_here_*` parameter, but as long as it stays hidden I can live with it. I'm just a little worried to see a constellation of specific parameters popping up in the futur. Let's be careful.

xlqian (approved):

PR #3238, opened 2020-06-30
[Jormungandr] remove administrative_region from places_nearby param &type[]

kadhikari:
Concerned ticket: https://jira.kisio.org/browse/NAVITIAII-2929

### DIFF
--- a/documentation/slate/source/includes/apis.md
+++ b/documentation/slate/source/includes/apis.md
@@ -762,7 +762,7 @@ coordinates, returning a [places](#place) collection.
   Required | name        | Type            | Description                       | Default value
   ---------|-------------|-----------------|-----------------------------------|---------------------------------------------------------------
   nop      | distance    | int             | Distance range in meters          | 500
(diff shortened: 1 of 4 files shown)


pbench (commented):
Should the check https://github.com/CanalTP/navitia/blob/56abcc686756c6ba6a93622dbc24514528059273/source/jormungandr/jormungandr/interfaces/v1/Places.py#L335 also be updated to use `places_nearby_type` ?

kadhikari:
> Should the check
> 
> https://github.com/CanalTP/navitia/blob/56abcc686756c6ba6a93622dbc24514528059273/source/jormungandr/jormungandr/interfaces/v1/Places.py#L335
> also be updated to use `places_nearby_type` ?

Non, The code above verifies the object type(stop_areas) in /v1/coverage/toto/stop_areas/sa_id/places_nearby?....
It does not verify value of the filter &type[].

pbench (approved):

nonifier (approved):
Nice one ! :+1:

PR #3239, opened 2020-07-02
[Jormun] take SN maneuvers into account with Here

benoit-bst:
**Objective** : Adding **Here** Maneuvers and keep backward compatibilty.
We use the same **Path** struct with new optionnal fields to extend it
**JIRA** : https://jira.kisio.org/browse/NAVP-1612

The output **API** is like that:

here_maneuvers

A full **instruction** in html format is available with several parameters : length, duration, direction, ...

For the moment Maneuvers are only in **english**. A next PR will allow to configure language with this list of possibilities :
https://developer.here.com/documentation/routing/dev_guide/topics/resource-param-type-languages.html#languages

TODO : 
- [x] Merge https://github.com/CanalTP/navitia-proto/pull/143 and change submodule to master

At the end, It will be presented like this in the playground :

playground_with_guidance

The Playground PR link https://github.com/CanalTP/navitia-playground/pull/326

### DIFF
--- a/source/jormungandr/jormungandr/interfaces/v1/serializer/journey.py
+++ b/source/jormungandr/jormungandr/interfaces/v1/serializer/journey.py
@@ -38,7 +38,12 @@ from jormungandr.interfaces.v1.serializer.pt import (
     StringListField,
 )
 from jormungandr.interfaces.v1.serializer.time import DateTimeField
-from jormungandr.interfaces.v1.serializer.fields import LinkSchema, RoundedField, SectionGeoJsonField
+from jormungandr.interfaces.v1.serializer.fields import (
+    LinkSchema,
+    RoundedField,
+    SectionGeoJsonField,
+    CoordSerializer,
+)
 from jormungandr.interfaces.v1.serializer.base import (
     AmountSerializer,
     PbNestedSerializer,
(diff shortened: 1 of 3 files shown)


source/jormungandr/jormungandr/street_network/here.py
```diff
@@ -251,13 +251,27 @@ def _read_response(response, origin, destination, mode, fallback_extremity, requ
             fm.car_no_park.name: response_pb2.Car,
         }
         section.street_network.mode = map_mode[mode]
+
+        # handle maneuvers to fill Path field
+        def _convert_direction(direction):
```
xlqian, line 256: I'm not sure we should show this information which is potentially false...

xlqian: what's is instruction when a roundabout is involved?

xlqian (approved):

source/jormungandr/jormungandr/street_network/here.py
benoit-bst: This is Here isntruction exemple for roundabout
```
"Take the <span class="exit" ="">1st exit</span> from Place François Mitterrand roundabout onto <span class="next-street">Boulevard Périphérique</span> toward <span class="sign"><span lang="fr">Nancy</span>/<span lang="fr">Aéroport Orly</span>/<span lang="fr">Lyon</span>/<span lang="fr">Porte de Vincennes</span>/<span lang="fr">Périphérique Intérieur</span></span>. <span class="distance-description">Go for <span class="length">2.1 km</span>.</span>"
```
To handle the `direction` field, the rules is simplified. **Here API** offers a `direction` field the can be :

- right
- left
- forward
- bearRight
- etc...

We only convert right to 90 and left to -90 and the others equals to 0 to be backward compatible.

source/jormungandr/jormungandr/interfaces/v1/serializer/journey.py
```diff
@@ -167,10 +172,31 @@ class JourneyDebugSerializer(PbNestedSerializer):
 
 
 class PathSerializer(PbNestedSerializer):
+    id = jsonschema.MethodField(schema_type=int, display_none=False)
     length = RoundedField(display_none=True)
     name = jsonschema.Field(schema_type=str, display_none=True)
     duration = RoundedField(display_none=True)
     direction = jsonschema.Field(schema_type=int, display_none=True)
+    instruction = jsonschema.MethodField(schema_type=int, display_none=False)
```
kadhikari: schema_type=str non?

kadhikari (changes requested):

source/jormungandr/jormungandr/street_network/here.py
```diff
@@ -251,13 +251,27 @@ def _read_response(response, origin, destination, mode, fallback_extremity, requ
...
+            sentence = direction.lower()
+            if "right" in sentence:
+                return 90
+            elif "left" in sentence:
+                return -90
+            else:
+                return 0
+
```
kadhikari, line 264: I also found an example of "direction": 113 in a street_network.

benoit-bst: yes for the classical Path kraken compute a angle.
This is for backward compatibility that I convert, for HERE, the `right/left` string to an angle. And I choose `+90/-90` to reflect right/left

source/jormungandr/jormungandr/interfaces/v1/serializer/journey.py
benoit-bst: hoooooo thanks

kadhikari (approved):

PR #3245, opened 2020-07-13
[Jormun] Add more accurate tags in direct path

xlqian:
https://jira.kisio.org/browse/NAVITIAII-2925

### DIFF
--- a/source/jormungandr/jormungandr/scenarios/utils.py
+++ b/source/jormungandr/jormungandr/scenarios/utils.py
@@ -65,6 +65,7 @@ mode_weight = {
     FallbackModes.taxi: 6,
     FallbackModes.ridesharing: 5,
     FallbackModes.car: 4,
+    FallbackModes.car_no_park: 4,
     FallbackModes.bike: 3,
     FallbackModes.bss: 2,
     FallbackModes.walking: 1,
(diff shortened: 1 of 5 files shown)


source/jormungandr/jormungandr/scenarios/utils.py
```diff
@@ -59,6 +59,7 @@
     FallbackModes.taxi: 6,
     FallbackModes.ridesharing: 5,
     FallbackModes.car: 4,
+    FallbackModes.car_no_park: 4,
```
kadhikari, line 68: No difference in weight between car and car_no_park ? What about 1 to 7 instead of 1 to 6 ??

source/jormungandr/jormungandr/street_network/kraken.py
```diff
@@ -139,7 +139,8 @@ def _direct_path(
         response = instance.send_and_receive(req, request_id=request_id)
         if should_invert_journey:
             return self._reverse_journeys(response)
-
+        if response.journeys:
+            response.journeys[0].sections[0].street_network.mode = FallbackModes[mode].value
```
kadhikari, line 143: I don't understand this line ?

source/jormungandr/jormungandr/scenarios/utils.py
xlqian: yes, this is an internal param to measure how "heavy" it is to compute with the given mode, so 'car' is as heavy as 'car_no_park' to compute

source/jormungandr/jormungandr/street_network/kraken.py
xlqian: this line gives a 'tag' to the returned direct_path. Notice that this line is located in 'kraken.py', a connector for distributed

pbench (approved):

nonifier (approved):

PR #3246, opened 2020-07-16
[DOC] add new badges into the main readme

benoit-bst:
- Build **badges** are dead due to the change of CI flow

badges

Now, _github actions_ is the system for building packages

- Add new badges like riot chat link, license and version release

### DIFF
--- a/readme.rst
+++ b/readme.rst
@@ -7,17 +7,20 @@
 =========
 ``(pronounce [navi-sia])``
 
-.. class:: no-web no-pdf
+.. image:: https://img.shields.io/github/v/tag/CanalTp/navitia?style=flat-square
+    :target: https://github.com/CanalTP/navitia/releases
+    :alt: version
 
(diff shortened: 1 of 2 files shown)


readme.rst
```diff
@@ -11,10 +11,13 @@
 
     Release branch: |last_release_build|
 
-.. |last_release_build| image:: https://ci.navitia.io/job/navitia_release/badge/icon
-    :target: https://ci.navitia.io/job/navitia_release/
+.. |last_release_build| image:: https://github.com/CanalTP/navitia/workflows/Build%20Navitia%20Packages%20For%20Release/badge.svg
```
pbougue: You can probably use shields to "flat-square" it too, like https://github.com/CanalTP/transit_model

benoit-bst: yes it is possible. I didn't think about that "flat-square". Thanks

kadhikari (approved):

PR #3247, opened 2020-07-17
[Jormungadr] fix traveler_type for several regions

xlqian:
https://jira.kisio.org/browse/NAVITIAII-3118

the travler_type was not updated when different regions are called...


kadhikari (approved):

kadhikari:
I tested it locally and admit that it is hard to test.

benoit-bst (approved):

pbench (approved):
:+1:

PR #3248, opened 2020-07-20
[JORMUN] Fix scenario and add debug infos

kadhikari:
When we have multi-coverage sharing part of geographical as well as public transport data, there is an error while handling scenario related to the coverage used for generating itinerary result.   

For details: https://jira.kisio.org/browse/NAVITIAII-3116

### DIFF
--- a/source/jormungandr/jormungandr/instance.py
+++ b/source/jormungandr/jormungandr/instance.py
@@ -212,11 +212,11 @@ class Instance(object):
         return models.Instance.get_by_name(self.name)
 
     def scenario(self, override_scenario=None):
-        if hasattr(g, 'scenario') and g.scenario:
-            """
-            once a scenario has been chosen for a request, we cannot change it
-            """
-            return g.scenario
+        """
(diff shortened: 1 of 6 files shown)


SGrenet (approved):

benoit-bst (approved):

pbench (approved):
Looks :+1:  to me ! 
Just wondering why we store the "scenario" in the "g" object (which seems to global to a request).
If I understand correctly, without a coverage specified, a request to journeys calls several regions, so several instances. The call to the first instance will set `g.scenario` for the duration of the request. Then, the calls to the subsequent instances will keep using the `g.scenario` set for the first instance.
Why not just return a fresh scenario in each call to `instance.scenario()`?

benoit-bst:
@pbench, « g » is the global context of the Flask API. It is useful to use it for storing some stuff that you want to keep all the Flask lifetime

kadhikari:
@pbench it's to avoid calling database for each sub query towards kraken. Once saved in g, can be used during all the sub queries for a query global to navitia.

PR #3249, opened 2020-07-23
[fallback with car] Filter parking for origin fallback with allowed_id[] and forbidden_uris[]

kadhikari:
To use a car or bss for origin fallback, we should be able to eliminate/force use of particular parkings. To do that followings are the modifications:
1. forbidden_uris[] added in the request PlacesNearbyRequest.
2. Since allowed_id[] is only used in journeys, poi ids present in this parameter are transformed as filter text (see : https://jira.kisio.org/browse/NAVP-1632)

For details refer to : https://jira.kisio.org/browse/NAVP-1628


benoit-bst (approved):

pbench (approved):
Well done !

source/jormungandr/jormungandr/georef.py
```diff
@@ -98,13 +100,33 @@ def get_crow_fly(
...
+
+        allowed_id_filter = ''
+        if allowed_id is not None:
+            allowed_id_count = len(allowed_id)
+            if allowed_id_count > 0:
+                # Retrieve first element(uri) of the set
+                allowed_id_items = 'poi.id={}'.format(allowed_id.pop())
+                # For the rest of elements use 'or' in between two elements(uris)
+                for uri in allowed_id:
+                    allowed_id_items = allowed_id_items + ' or poi.id={}'.format(uri)
```
xlqian:
```suggestion
                pois_ids = ('poi.id={}'.format(allowed_id) for allowed_id in  allowed_id)
                allowed_id_items = '  or  '.join(pois_ids)
```

source/jormungandr/jormungandr/georef.py
```diff
@@ -98,13 +100,33 @@ def get_crow_fly(
...
+                # Format the filter for all allowed_ids uris
+                if allowed_id_count == 1:
+                    allowed_id_filter = ' and {}'.format(allowed_id_items)
+                elif allowed_id_count > 1:
+                    allowed_id_filter = ' and ({})'.format(allowed_id_items)
```
xlqian, line 113: are you sure that when allowed_id_count == 1, this syntax doesn't work?

source/jormungandr/jormungandr/utils.py
```diff
@@ -640,3 +640,13 @@ def portable_min(*args, **kwargs):
...
     return round(3.6 * speed)
+
+
+def get_poi_params(codes):
+    poi_params = set()
+    if codes is not None:
+        for code in codes:
+            split_code = code.split(':')
+            if len(split_code) > 0 and split_code[0] == 'poi':
+                poi_params.add(code)
```
xlqian: what you want here is `startwith`
```suggestion
    for code in (codes or []):
        if split_code.startwith('poi:'):
           poi_params.add(code)
```

source/jormungandr/jormungandr/georef.py
```diff
@@ -83,6 +83,8 @@ def get_crow_fly(
         filter=None,
         stop_points_nearby_duration=300,
         request_id=None,
+        forbidden_uris=None,
+        allowed_id=None,
```
nonifier: Having for default value an empty array would make your code much simpler
```suggestion
        forbidden_uris=[],
        allowed_id=[],
```

source/jormungandr/jormungandr/georef.py
```diff
@@ -98,13 +100,33 @@ def get_crow_fly(
...
-            req.places_nearby.filter = filter
+            req.places_nearby.filter = filter + allowed_id_filter
         if streetnetwork_mode == "car":
             req.places_nearby.stop_points_nearby_radius = (
                 kwargs.get("walking", 1.11) * stop_points_nearby_duration
             )
             req.places_nearby.depth = 1
+        if forbidden_uris is not None:
+            for uri in forbidden_uris:
+                req.places_nearby.forbidden_uris.append(uri)
```
nonifier, line 125: Why can't we do the same for `allowed_ids` ? Is there a missing interface in PTref ? we should definitely add it to PtRef if it's not there. Otherwise, we will duplicate the same business logic everywhere we need allowed ids.

kadhikari: allowed_ids are only used by JourneyRequest (api journeys) but not in other endpoints. https://github.com/CanalTP/navitia-proto/blob/85ae96e65a1b54764cb1f0c614353909eed51a48/request.proto#L87

In other requests (endpoints) we use these ids in filter

source/jormungandr/jormungandr/georef.py
```diff
@@ -98,13 +100,29 @@ def get_crow_fly(
         req.places_nearby.start_page = 0
         req.disable_feedpublisher = True
         req.places_nearby.types.append(object_type)
+
+        allowed_id_filter = ''
+        if allowed_id is not None:
+            allowed_id_count = len(allowed_id)
+            if allowed_id_count > 0:
+                poi_ids = ('poi.id={}'.format(uri) for uri in allowed_id)
+                allowed_id_items = '  or  '.join(poi_ids)
```
xlqian, line 109: :heart:

xlqian (approved):

nonifier (approved):
It would have been nice to have `allowed_ids` as part of the PTref API (like `forbidden_uri) so that all endpoints could benefit from it.

PR #3250, opened 2020-07-24
[jormun] Here, add exclusion areas option

benoit-bst:
### Topic
**Here** is capable to work with a list of **exclusion areas**. In fact this is exclusion box, created with 2 coords (the **avoidAreas** parameter: https://developer.here.com/documentation/routing/dev_guide/topics/resource-calculate-route.html).
In navitia, a parameter is added `_here_exclusion_area[]`. The format is as follows:
`coord_1!coord_2` -> example `_here_exclusion_area[]=2.36753;48.89632!2.32119;48.90422`

### Example

I wrote a request without exclusion areas :
`http://localhost:5000/v1/coverage/default/journeysfrom=2.26867%3B48.88134&to=2.40553%3B48.84866&first_section_mode%5B%5D=car_no_park&_override_scenario=distributed&direct_path=only`
the response gives me a trip with the _"périphérique intérieur"_ (at this point, more fast than _"périphérique extérieur"_)

p_int

I wrote the same request with an exclusion area that cross the "périphérique" (in the north). it forced, the journey to the south way

p_ext

This option is usefull to prevent journeys in a city with forbidden areas (market place the sunday, yellow jacket, national day, and so on) :smile_cat:

### DIFF
--- a/source/jormungandr/jormungandr/interfaces/v1/Journeys.py
+++ b/source/jormungandr/jormungandr/interfaces/v1/Journeys.py
@@ -459,16 +459,16 @@ class Journeys(JourneyCommon):
             help="Here, Max number of matrix points for the street network computation (limited to 100)",
         )
         parser_get.add_argument(
-            '_sn_exclusion_area[]',
+            '_here_exclusion_area[]',
             type=six.text_type,
             case_sensitive=False,
             hidden=True,
             action='append',
-            dest='_sn_exclusion_areas',
+            dest='_here_exclusion_area[]',
             help='Give 2 coords for an exclusion box. The format is like that:\n'
             'Coord_1!Coord_2 with Coord=lat;lon\n'
-            ' - exemple : _sn_exclusion_area[]=2.40553;48.84866!2.41453;48.85677\n'
-            ' - This is a list, you can add to the maximun 20 _sn_exclusion_area[]\n',
+            ' - exemple : _here_exclusion_area[]=2.40553;48.84866!2.41453;48.85677\n'
+            ' - This is a list, you can add to the maximun 20 _here_exclusion_area[]\n',
         )
         parser_get.add_argument(
             "equipment_details",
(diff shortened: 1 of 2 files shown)


source/jormungandr/jormungandr/interfaces/v1/Journeys.py
```diff
@@ -458,6 +458,18 @@ def __init__(self):
             hidden=True,
             help="Here, Max number of matrix points for the street network computation (limited to 100)",
         )
+        parser_get.add_argument(
+            '_sn_exclusion_area[]',
+            type=six.text_type,
+            case_sensitive=False,
+            hidden=True,
+            action='append',
+            dest='_sn_exclusion_areas',
```
pbench: If I'm not mistaken, this parameters applies only to the car fallbacks obtained with Here. But the name  `sn_exclusion_areas` suggests it applies to all street network computations.
Maybe rename it to something more specific ? Like "here_sn_exclusion_areas" ?

source/jormungandr/jormungandr/interfaces/v1/Journeys.py
```diff
@@ -458,6 +458,18 @@ def __init__(self):
...
+            help='Give 2 coords for an exclusion box. The format is like that:\n'
+            'Coord_1!Coord_2 with Coord=lat;lon\n'
+            ' - exemple : _sn_exclusion_area[]=2.40553;48.84866!2.41453;48.85677\n'
+            ' - This is a list, you can add to the maximun 20 _sn_exclusion_area[]\n',
```
pbench: The max of 20 exclusion areas is not checked in the code. What happens if more than 20 are given ? 
Maybe keep only the first 20 to give to Here, and log a message in jormun ?

pbench (approved):
Some minor suggestions (feel free to dismiss it).
Looks good otherwise !

source/jormungandr/jormungandr/interfaces/v1/Journeys.py
benoit-bst: Good point, I can protect and inform

source/jormungandr/jormungandr/interfaces/v1/Journeys.py
benoit-bst: i didn't think about that. I accept your change, it will be more understandable

PR #3251, opened 2020-07-28
Alternative socket for pt queries

pbench:
Allows to configure a coverage so that "pt_planner" request are send to a specific socket, instead of the main socket of the coverage. 
This allows, in the distributed context, to use another engine than kraken for the "pt_planner" request, but still use kraken for all other kind of request

A coverage can be configured as follows :
```bash
JORMUNGANDR_INSTANCE_AUVERGNE= '{"key": "fr-auv", "zmq_socket": "ipc:///tmp/fr_auv_kraken", "pt_zmq_socket" : "ipc:///tmp/fr_auv_pt"}'
```

### DIFF
--- a/source/jormungandr/jormungandr/instance.py
+++ b/source/jormungandr/jormungandr/instance.py
@@ -617,18 +617,18 @@ class Instance(object):
                         break  # remaining socket are still in "keep alive" state
                 except IndexError:
                     break
+
         _reap_sockets(self._sockets)
         _reap_sockets(self._pt_sockets)
-
 
     @contextmanager
     def socket(self, context, pt_socket=False):
 
         socket = None
-        if pt_socket :
+        if pt_socket:
             sockets = self._pt_sockets
             socket_path = self.pt_socket_path
-        else :
+        else:
             sockets = self._sockets
             socket_path = self.socket_path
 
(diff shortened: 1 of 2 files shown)


source/jormungandr/jormungandr/instance.py
```diff
@@ -549,20 +556,45 @@ def reap_socket(self, ttl):
...
-            socket, _ = self._sockets.pop()
-        except IndexError:  # there is no socket available: lets create one
-            socket = context.socket(zmq.REQ)
-            socket.connect(self.socket_path)
-        try:
-            yield socket
-        finally:
-            if not socket.closed:
-                self._sockets.append((socket, time.time()))
+        if not pt_socket:
```
benoit-bst: Maybe you can create a function to avoid the duplicated code into the « if » « else ».

benoit-bst (approved):
Can be a good way to plug your new cute journeys core 🐥

nonifier:
I don't really get the point of your extra socket ? Looks like you are doing exactly the same thing depending on which socket is being set. 
I think that it would be better if a 'distributed service' is holding the socket, pretty much the same way a connector holds an url. Or maybe having an socket array at the instance level, so we could dispatch a request to any socket based on a routing key.

nonifier (approved):

PR #3252, opened 2020-07-29
[jormun] feed publishers are added even without direct path

benoit-bst:
**JIRA**: https://jira.kisio.org/browse/NAVP-1627
There is a bug with SN feed publishers in some cases :
- without Direct path (_direct_path=none_)
- max_duration = 0 (equivalent to only direct path mode)

The PR fixes that :tulip:

### DIFF
--- a/source/jormungandr/jormungandr/street_network/street_network.py
+++ b/source/jormungandr/jormungandr/street_network/street_network.py
@@ -88,7 +88,6 @@ class AbstractStreetNetworkService(ABC):  # type: ignore
             request_id,
         )
 
-        self._add_feed_publisher(resp)
         return resp
 
     @abc.abstractmethod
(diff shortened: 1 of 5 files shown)


source/jormungandr/jormungandr/street_network/tests/geovelo_test.py
```diff
@@ -325,12 +325,6 @@ def direct_path_geovelo_zero_test():
         assert geovelo_resp.journeys[0].sections[0].street_network.path_items[0].length == 0
         assert geovelo_resp.journeys[0].sections[0].street_network.path_items[0].duration == 0
 
-        assert len(geovelo_resp.feed_publishers) == 1
```
benoit-bst, line 328: Remove, because this is managed in a upper level with Distributed now

source/jormungandr/jormungandr/street_network/street_network.py
```diff
@@ -88,7 +88,6 @@ def direct_path_with_fp(
             request_id,
         )
 
-        self._add_feed_publisher(resp)
```
nonifier, line 91: How does it work for `new_default` now this is gone ?

xlqian (approved):

pbench (approved):

source/jormungandr/jormungandr/street_network/street_network.py
benoit-bst: it has been tested for new_default and distributed and it works. It is managed in a upper level for the 2 modes

PR #3258, opened 2020-08-10
[CI] Clang tidy - Automatic PR 

nonifier:
Here come the Robots ! :robot: :robot: :robot: 

## When
Every time we will **merge to** the **release** branch (every 2 weeks), clang-tidy will be run against the code base and  changes will be applied if needed.

## What
All **warnings** from clang-tidy are ignored for the moment. We only are interested in the automatic changes here.
So far, we only are performing the following checks:
 
* [modernize-avoid-bind.html](https://clang.llvm.org/extra/clang-tidy/checks/modernize-avoid-bind.html)
* [modernize-concat-nested-namespaces](https://clang.llvm.org/extra/clang-tidy/checks/modernize-concat-nested-namespaces.html)
* [modernize-deprecated-headers](https://clang.llvm.org/extra/clang-tidy/checks/modernize-deprecated-headers.html)
* [modernize-deprecated-ios-base-aliases](https://clang.llvm.org/extra/clang-tidy/checks/modernize-deprecated-ios-base-aliases.html)
* [modernize-loop-convert](https://clang.llvm.org/extra/clang-tidy/checks/modernize-loop-convert.html)

A pull request is then automatically created like https://github.com/CanalTP/navitia/pull/3257. 

More checks will be added once we agree on a workflow that fits everybody's need. 
https://releases.llvm.org/10.0.0/tools/clang/tools/extra/docs/clang-tidy/checks/list.html

## Why
Because modern code is beautiful (and because @AurelienLP )


source/cmake_modules/EnableClangTidy.cmake
```diff
@@ -0,0 +1,64 @@
+
+message(STATUS "cland-tools")
```
pbench:
```suggestion
message(STATUS "clang-tools")
```

pbench (approved):
Nice job ! :+1:

xlqian (approved):

PR #3269, opened 2020-08-18
[Jormungandr] fix bss walking overlapping problem

xlqian:
https://jira.kisio.org/browse/NAVITIAII-3133

- [x] Still need to add tests

### DIFF
--- a/source/jormungandr/jormungandr/scenarios/journey_filter.py
+++ b/source/jormungandr/jormungandr/scenarios/journey_filter.py
@@ -448,14 +448,14 @@ def similar_pt_section_line(section):
     return "pt:{}".format(section.pt_display_informations.uris.line)
 
 
-def similar_journeys_generator(journey, pt_functor, sn_functor=None):
+def similar_journeys_generator(journey, pt_functor, sn_functor=lambda mode: mode):
     for idx, s in enumerate(journey.sections):
         if s.type == response_pb2.PUBLIC_TRANSPORT:
             yield pt_functor(s)
         elif s.type == response_pb2.STREET_NETWORK and is_walk_after_parking(journey, idx):
             continue
         elif s.type in (response_pb2.STREET_NETWORK, response_pb2.CROW_FLY):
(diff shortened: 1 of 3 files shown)


source/jormungandr/jormungandr/scenarios/journey_filter.py
```diff
@@ -669,7 +686,9 @@ def is_fallback(section):
 
 
 def filter_similar_vj_journeys(journey_pairs_pool, request):
-    _filter_similar_journeys(journey_pairs_pool, request, similar_journeys_vj_generator)
+    _filter_similar_journeys(
+        journey_pairs_pool, request, similar_journeys_vj_generator, similar_bss_walking_vj_generator
```
nonifier, line 692:
```suggestion
        journey_pairs_pool, request, (similar_journeys_vj_generator, similar_bss_walking_vj_generator)
```

Would it make sense to group the generators altogether (in a generator :smile:)  ?
That would avoid changing the `*similar_journey_generators` later down.

source/jormungandr/jormungandr/scenarios/journey_filter.py
```diff
@@ -448,14 +448,31 @@ def similar_pt_section_line(section):
...
+def detailed_pt_section_vj(section):
+    return 'pt:{vj} dep:{dep} arr:{arr}'.format(
+        vj=section.pt_display_informations.uris.vehicle_journey,
+        dep=section.begin_date_time,
+        arr=section.end_date_time,
+    )
+
+
+def bss_walking_sn_functor(mode):
+    return None if mode in (FallbackModes.walking.value, FallbackModes.bss.value) else mode
```
nonifier: Would it not be better to return the `walking` mode instead ? 
Just wondering if a crow fly would not have it's mode set, it would be equal to a `walking` and a `bss`.

source/jormungandr/jormungandr/scenarios/journey_filter.py
xlqian: I did this so that there would be minimum changes... :)

nonifier (approved):

PR #3274, opened 2020-09-04
[Jromun] use a better method to generate the real time info for bss

xlqian:
Dummy BSS Provider:

Use a hash algo to generate radom real time info

### DIFF
--- a/source/jormungandr/jormungandr/parking_space_availability/bss/dummy.py
+++ b/source/jormungandr/jormungandr/parking_space_availability/bss/dummy.py
@@ -76,6 +76,11 @@ class DummyProvider(CommonBssProvider):
         ref = poi.get('properties', {}).get('ref')
         if not ref:
             return
-        # calculate dummy available stands number
-        s = int(bytearray(str(ref))[0])
-        return Stands(int(s / 2), s, StandsStatus.closed if s % 2 else StandsStatus.open)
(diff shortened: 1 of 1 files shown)


SGrenet (approved):

source/jormungandr/jormungandr/parking_space_availability/bss/dummy.py
```diff
@@ -76,6 +76,11 @@ def _get_informations(self, poi):
...
-        # calculate dummy available stands number
-        s = int(bytearray(str(ref))[0])
-        return Stands(int(s / 2), s, StandsStatus.closed if s % 2 else StandsStatus.open)
+        # generate dummy available stands number using a hash algo
+        import hashlib
+
+        m = hashlib.md5()
+        for k, v in poi.get('properties', {}).items():
+            m.update(v.encode('utf-8', errors='ignore'))
+        s = int(bytearray(str(m.hexdigest()))[0])
```
pbench, line 85: Looks good, but why not something simpler like :
`s = random.randint(0,50)`
?

xlqian: because I want the realtime info is random but also 'fixed', so that team FINA can test it.

pbench (approved):

PR #3275, opened 2020-09-07
[Jormun+Tyr] Asynchronous ridesharing

benoit-bst:
### Objective

Make **Ridesharing** calls asynchronous. 
For that, we create an option that active the new mode. This mode is simple :

- We call a journeys with ridesharing option (_first_section_mode[] or last_section_mode[] =ridesharing_). 
- The response is a car response without ridesharing, but a link is added into the response to call the _new ridesharing API_ (Work in progress).

**JIRA**: https://jira.kisio.org/browse/NAVP-1646

### Activation

- By var env : **JORMUNGANDR_ASYNCHRONOUS_RIDESHARING**
- By a parameter, within the settings file : **ASYNCHRONOUS_RIDESHARING=True**
- By the Db field : **asynchronous_ridesharing** into instance table
- By a API parameter : **_asynchronous_ridesharing=True**

### Notes

- For the moment, the _**link is fake**_ because the ridesharing API doesn't exist
- Tyr is ready

### DIFF
--- a/source/jormungandr/jormungandr/instance.py
+++ b/source/jormungandr/jormungandr/instance.py
@@ -134,6 +134,7 @@ class Instance(object):
         self.georef = georef.Kraken(self)
         self.planner = planner.Kraken(self)
         self._streetnetwork_backend_manager = streetnetwork_backend_manager
+        self.asynchronous_ridesharing = False
 
         if app.config[str('DISABLE_DATABASE')]:
             self._streetnetwork_backend_manager.init_streetnetwork_backends_legacy(
@@ -171,6 +172,12 @@ class Instance(object):
                 app.config[str('EQUIPMENT_DETAILS_PROVIDERS')], self.get_providers_from_db
             )
 
+        # New ridesharing type
+        if app.config[str('DISABLE_DATABASE')]:
+            self.asynchronous_ridesharing = app.config[str('ASYNCHRONOUS_RIDESHARING')]
+        else:
+            self.asynchronous_ridesharing = self.get_asynchronous_ridesharing_from_db
+
         # Init equipment providers from config
         self.equipment_provider_manager.init_providers(instance_equipment_providers)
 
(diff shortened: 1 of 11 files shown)


source/jormungandr/jormungandr/instance.py
```diff
@@ -134,6 +134,7 @@ def __init__(
         self.georef = georef.Kraken(self)
         self.planner = planner.Kraken(self)
         self._streetnetwork_backend_manager = streetnetwork_backend_manager
+        self.asynchronous_ridesharing = False
```
nonifier, line 137: You are a brave man ! 
I would have been just fine with `async_`, but you are not the kind of guy that takes shortcuts :laughing:

nonifier (approved):

source/jormungandr/jormungandr/scenarios/new_default.py
```diff
@@ -1097,19 +1097,8 @@ def fill_journeys(self, request_type, api_request, instance):
...
-            try:
-                decorate_journeys(pb_resp, instance, api_request)
-            except Exception:
-                logger.exception('Error while retrieving ridesharing ads')
-        else:
-            for j in pb_resp.journeys:
-                if 'ridesharing' in j.tags:
-                    journey_filter.mark_as_dead(j, api_request.get('debug'), 'no_matching_ridesharing_found')
+        # Handle ridesharing service
+        self.handle_ridesharing_services(logger, instance, ridesharing_req, pb_resp)
```
xlqian, line 1101: :+1:

source/jormungandr/jormungandr/scenarios/new_default.py
```diff
@@ -1194,6 +1183,30 @@ def isochrone(self, request, instance):
 
         return resp
 
+    def handle_ridesharing_services(self, logger, instance, request, pb_response):
+        def active_asynchronous_ridesharing(request, instance):
+            if request.get("_asynchronous_ridesharing", False):
```
xlqian: hmmm, what if the user put `_asynchronous_ridesharing=false` and instance.asynchronous_ridesharing is set `true`?

I think what you want actually is:

```python
if request.get("_asynchronous_ridesharing", None) is None:
    # when the param is not set in the request, use the instance config
    return instance.asynchronous_ridesharing
else:
    return request.get("_asynchronous_ridesharing")
```

benoit-bst: It seems to be more clear. Why not. 
I will do the change. 
Thanks

xlqian (approved):

PR #3276, opened 2020-09-08
[Distributed] fix_bss_journeys_without_bss_sections

xlqian:
https://jira.kisio.org/browse/NAVITIAII-3133

### DIFF
--- a/source/jormungandr/jormungandr/street_network/kraken.py
+++ b/source/jormungandr/jormungandr/street_network/kraken.py
@@ -138,9 +138,9 @@ class Kraken(AbstractStreetNetworkService):
 
         response = instance.send_and_receive(req, request_id=request_id)
         if should_invert_journey:
-            return self._reverse_journeys(response)
+            response = self._reverse_journeys(response)
 
-        def has_bss_section(journey):
+        def has_bss_rent_before_put_back_section(journey):
             # Here is a little trick with python's generator
             # the next 3 lines check not only the existences of BSS_RENT and BSS_PUT_BACK, but also check the fact that
             # BSS_RENT must be located Before BSS_PUT_BACK
@@ -150,7 +150,13 @@ class Kraken(AbstractStreetNetworkService):
             return bss_rent and bss_put_back
 
         if response.journeys:
-            if mode == FallbackModes.bss.name and (not has_bss_section(response.journeys[0])):
+            # Note that: the journey returned by Kraken is a direct path. A direct path of walking/bike/car/car_no_park/
+            # contains only one section. But for bss, there may be one or three sections.
+            # For bss,we only need attribute the mode to the first section. The most significant mode will be chosen
+            # later
+            if mode == FallbackModes.bss.name and (
+                not has_bss_rent_before_put_back_section(response.journeys[0])
+            ):
                 response.journeys[0].sections[0].street_network.mode = FallbackModes.walking.value
             else:
                 response.journeys[0].sections[0].street_network.mode = FallbackModes[mode].value
--- a/source/jormungandr/jormungandr/street_network/kraken.py
+++ b/source/jormungandr/jormungandr/street_network/kraken.py
@@ -139,8 +139,21 @@ class Kraken(AbstractStreetNetworkService):
         response = instance.send_and_receive(req, request_id=request_id)
         if should_invert_journey:
             return self._reverse_journeys(response)
+
+        def has_bss_section(journey):
+            # Here is a little trick with python's generator
+            # the next 3 lines check not only the existences of BSS_RENT and BSS_PUT_BACK, but also check the fact that
+            # BSS_RENT must be located Before BSS_PUT_BACK
+            sections = (s for s in journey.sections)
+            bss_rent = next((True for s in sections if s.type == response_pb2.BSS_RENT), False)
+            bss_put_back = next((True for s in sections if s.type == response_pb2.BSS_PUT_BACK), False)
+            return bss_rent and bss_put_back
+
         if response.journeys:
-            response.journeys[0].sections[0].street_network.mode = FallbackModes[mode].value
+            if mode == FallbackModes.bss.name and (not has_bss_section(response.journeys[0])):
+                response.journeys[0].sections[0].street_network.mode = FallbackModes.walking.value
+            else:
+                response.journeys[0].sections[0].street_network.mode = FallbackModes[mode].value
         return response
 
     def _hanlde_car_no_park_modes(self, mode):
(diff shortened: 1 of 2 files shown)


source/jormungandr/jormungandr/street_network/kraken.py
```diff
@@ -139,8 +139,21 @@ def _direct_path(
         response = instance.send_and_receive(req, request_id=request_id)
         if should_invert_journey:
             return self._reverse_journeys(response)
```
pbench: Shouldn't this early return be after the code you added ?

source/jormungandr/jormungandr/street_network/kraken.py
```diff
@@ -139,8 +139,21 @@ def _direct_path(
...
+
+        def has_bss_section(journey):
+            # Here is a little trick with python's generator
+            # the next 3 lines check not only the existences of BSS_RENT and BSS_PUT_BACK, but also check the fact that
+            # BSS_RENT must be located Before BSS_PUT_BACK
```
pbench, line 146: Where is hidden the magic that check the BSS_RENT is before BSS_PUT_BACK ?

source/jormungandr/jormungandr/street_network/kraken.py
```diff
@@ -139,8 +139,21 @@ def _direct_path(
...
+            sections = (s for s in journey.sections)
+            bss_rent = next((True for s in sections if s.type == response_pb2.BSS_RENT), False)
+            bss_put_back = next((True for s in sections if s.type == response_pb2.BSS_PUT_BACK), False)
+            return bss_rent and bss_put_back
+
         if response.journeys:
-            response.journeys[0].sections[0].street_network.mode = FallbackModes[mode].value
+            if mode == FallbackModes.bss.name and (not has_bss_section(response.journeys[0])):
+                response.journeys[0].sections[0].street_network.mode = FallbackModes.walking.value
```
pbench, line 160: Maybe add a comment to explain why you only do this for `journeys[0]` and its `sections[0]`.
I took me a little while to understand that there should be only one journey in the response, and if its a direct walk path it should have only one section

source/jormungandr/jormungandr/street_network/kraken.py
```diff
@@ -139,8 +139,21 @@ def _direct_path(
...
+            else:
+                response.journeys[0].sections[0].street_network.mode = FallbackModes[mode].value
```
pbench, line 162: Could you have more than one section here ?

source/jormungandr/jormungandr/street_network/kraken.py
nonifier: The magic comes from the `next` function, that will iterate the sections until it finds a bss_rent.
Then iterate over what's left of the generator to find a bass_put_back.

source/jormungandr/jormungandr/street_network/kraken.py
```diff
@@ -139,8 +139,21 @@ def _direct_path(
         response = instance.send_and_receive(req, request_id=request_id)
         if should_invert_journey:
             return self._reverse_journeys(response)
+
+        def has_bss_section(journey):
```
nonifier: Should this rather be like: 
```suggestion
        def has_bss_rent_before_put_back_section(journey):
```

source/jormungandr/jormungandr/street_network/kraken.py
pbench: But here the generator is not kept between bss_rent and bss_put_back, right ?

source/jormungandr/jormungandr/street_network/kraken.py
xlqian: OK

source/jormungandr/jormungandr/street_network/kraken.py
```diff
@@ -139,8 +139,21 @@ def _direct_path(
...
+            # the next 3 lines check not only the existences of BSS_RENT and BSS_PUT_BACK, but also check the fact that
+            # BSS_RENT must be located Before BSS_PUT_BACK
+            sections = (s for s in journey.sections)
+            bss_rent = next((True for s in sections if s.type == response_pb2.BSS_RENT), False)
+            bss_put_back = next((True for s in sections if s.type == response_pb2.BSS_PUT_BACK), False)
+            return bss_rent and bss_put_back
+
         if response.journeys:
-            response.journeys[0].sections[0].street_network.mode = FallbackModes[mode].value
+            if mode == FallbackModes.bss.name and (not has_bss_section(response.journeys[0])):
```
nonifier: I'm not sure I understand why we do have `bss` and no `walking` section in the first place.

From what I understand, with distributed we do 2 calls to the street network service: 
* one for walking 
* one for bss
So we should end up with 2 identical journeys with the same fallback sections ? But yet, we only have the (fake) bss one. Is the walking journey filtered out earlier, or am I missing something ?

source/jormungandr/jormungandr/street_network/kraken.py
nonifier: But it is between the 2 calls to `next` :) 
The generator `sections` can be seen as an iterator if you prefer.

nonifier (approved):
This seems a bit hacky to me and I afraid this comes back at us later. 
I'll approve it anyway according the the urgency of the fix !

source/jormungandr/jormungandr/street_network/kraken.py
xlqian: I just had a quick discussion with @nonifier, I assume that the logic of journey tags is not concentrated at all, it's no wonder that you get lost... In fact, we may have several sections, but  https://github.com/CanalTP/navitia/blob/dev/source/jormungandr/jormungandr/scenarios/new_default.py#L697 will take care of it, it will take the "heaviest" mode (the most significant) to tag the journey

source/jormungandr/jormungandr/street_network/kraken.py
xlqian: The bug that this PR fix is not an issue of duplicated journeys due to bss-walking. It's really due to the bad journey tagging. :)

source/jormungandr/jormungandr/street_network/kraken.py
pbench: Ok, I get it ! :+1:

xlqian: @pbench 
```python
l = [1,2,3,4,5,6]
magic_generator = (x for x in l)
should_be_true  = next( (True for x in magic_generator if x == 3 ), False)
should_be_false = next( (True for x in magic_generator if x == 2 ), False)

```

xlqian:
@nonifier The fix is indeed a little bit tricky, because of distributed, in new_default, this logic is handled directly by kraken and here we have to handle a part of it in jormungandr which seems very strange to those (including me) who don't have a whole picture of these logics...

pbench (approved):

PR #3280, opened 2020-09-17
[Jormun] asynchronous ridesharing mode

benoit-bst:
## Final Objective

**Confluence**: https://confluence.kisio.org/display/NAV/Covoiturage+%3A+le+RideSharing+dans+Navitia
**JIRA**: https://jira.kisio.org/browse/NAVP-1660

### Current mode
We have added the new _ridesharing_journeys_ link

old_ridesharing

### New mode

**first call** is the same response of the _current mode_ without ridesharing journeys (you have to click on the link)
first_new_ridesharing

when you click on **ridesharing_jouneys link**, it calls the same request with a additional parameter **"partner_services=ridesharing"**.
It returns only the ridesharing journeys.
second_new_ridesharing

finally, the request time consumption is now during the second call (via the ridesharing_journeys link)

## new mode activation (for the migration)
By the Db field : asynchronous_ridesharing into instance table (navitia_configurator field :heart:)
By a API parameter : _asynchronous_ridesharing=True (for tests purpose)

## Notes

- whether old or new mode, a link is added (ridesharing_journeys) to call only ridesharing response
- Tyr is ready
- There is a debate around the parameter name : **"partner_services[]=ridesharing"**. It is open, feel free to propose a different name (@stifoon @wassimbenaissa and more)

### DIFF
--- a/source/jormungandr/jormungandr/default_settings.py
+++ b/source/jormungandr/jormungandr/default_settings.py
@@ -24,9 +24,6 @@ SQLALCHEMY_DATABASE_URI = os.getenv(
 
 DISABLE_DATABASE = boolean(os.getenv('JORMUNGANDR_DISABLE_DATABASE', False))
 
-# Active the asynchronous ridesharing mode
-ASYNCHRONOUS_RIDESHARING = boolean(os.getenv('JORMUNGANDR_ASYNCHRONOUS_RIDESHARING', False))
-
 # disable authentication
 PUBLIC = boolean(os.getenv('JORMUNGANDR_IS_PUBLIC', True))
 
--- a/source/jormungandr/jormungandr/default_settings.py
+++ b/source/jormungandr/jormungandr/default_settings.py
@@ -24,6 +24,9 @@ SQLALCHEMY_DATABASE_URI = os.getenv(
 
 DISABLE_DATABASE = boolean(os.getenv('JORMUNGANDR_DISABLE_DATABASE', False))
 
+# Active the asynchronous ridesharing mode
+ASYNCHRONOUS_RIDESHARING = boolean(os.getenv('JORMUNGANDR_ASYNCHRONOUS_RIDESHARING', False))
+
 # disable authentication
 PUBLIC = boolean(os.getenv('JORMUNGANDR_IS_PUBLIC', True))
 
--- a/source/jormungandr/jormungandr/instance.py
+++ b/source/jormungandr/jormungandr/instance.py
@@ -134,7 +134,6 @@ class Instance(object):
         self.georef = georef.Kraken(self)
         self.planner = planner.Kraken(self)
         self._streetnetwork_backend_manager = streetnetwork_backend_manager
-        self.asynchronous_ridesharing = False
 
         if app.config[str('DISABLE_DATABASE')]:
             self._streetnetwork_backend_manager.init_streetnetwork_backends_legacy(
@@ -172,12 +171,6 @@ class Instance(object):
                 app.config[str('EQUIPMENT_DETAILS_PROVIDERS')], self.get_providers_from_db
             )
 
-        # New ridesharing type
-        if app.config[str('DISABLE_DATABASE')]:
-            self.asynchronous_ridesharing = app.config[str('ASYNCHRONOUS_RIDESHARING')]
-        else:
-            self.asynchronous_ridesharing = self.get_asynchronous_ridesharing_from_db  # type: ignore
-
         # Init equipment providers from config
         self.equipment_provider_manager.init_providers(instance_equipment_providers)
 
@@ -186,11 +179,6 @@ class Instance(object):
         :return: a callable query of equipment providers associated to the current instance in db
         """
         return self._get_models().equipment_details_providers
-
-    def get_asynchronous_ridesharing_from_db(self):
-        # type: () -> bool
-        instance_db = self.get_models()
-        return get_value_or_default('asynchronous_ridesharing', instance_db, self.name)
 
     @property
     def autocomplete(self):
@@ -501,6 +489,12 @@ class Instance(object):
         # type: () -> int
         instance_db = self.get_models()
         return get_value_or_default('max_extra_second_pass', instance_db, self.name)
+
+    @property
+    def asynchronous_ridesharing(self):
+        # type: () -> bool
+        instance_db = self.get_models()
+        return get_value_or_default('asynchronous_ridesharing', instance_db, self.name)
 
     @property
     def max_nb_crowfly_by_mode(self):
--- a/source/jormungandr/jormungandr/instance.py
+++ b/source/jormungandr/jormungandr/instance.py
@@ -134,6 +134,7 @@ class Instance(object):
         self.georef = georef.Kraken(self)
         self.planner = planner.Kraken(self)
         self._streetnetwork_backend_manager = streetnetwork_backend_manager
+        self.asynchronous_ridesharing = False
 
         if app.config[str('DISABLE_DATABASE')]:
             self._streetnetwork_backend_manager.init_streetnetwork_backends_legacy(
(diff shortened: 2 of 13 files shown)


source/jormungandr/jormungandr/interfaces/v1/journey_common.py
```diff
@@ -455,6 +455,19 @@ def __init__(self, output_type_serializer):
             hidden=True,
             help="define the duration to reach stop points by crow fly",
         )
+        parser_get.add_argument(
+            "only_ridesharing",
```
nonifier: minor: 
 :gb: `ridesharing_only` would be more grammatically correct english-wise

source/jormungandr/jormungandr/instance.py
```diff
@@ -171,6 +172,12 @@ def __init__(
                 app.config[str('EQUIPMENT_DETAILS_PROVIDERS')], self.get_providers_from_db
             )
 
+        # New ridesharing type
+        if app.config[str('DISABLE_DATABASE')]:
+            self.asynchronous_ridesharing = app.config[str('ASYNCHRONOUS_RIDESHARING')]
```
xlqian: `self.asynchronous_ridesharing` is a bool here, but 2 lines further, `self.asynchronous_ridesharing` becomes a functor... something is wrong here, no?

source/jormungandr/jormungandr/scenarios/new_default.py
```diff
@@ -1194,6 +1183,39 @@ def isochrone(self, request, instance):
 
         return resp
 
+    def handle_ridesharing_services(self, logger, instance, request, pb_response):
+
+        if instance.ridesharing_services and (
+            'ridesharing' in request['origin_mode'] or 'ridesharing' in request['destination_mode']
+        ):
+
+            def active_asynchronous_ridesharing(request, instance):
```
```suggestion
            def is_asynchronous_ridesharing(request, instance):
```

source/jormungandr/jormungandr/scenarios/new_default.py
```diff
@@ -1194,6 +1183,39 @@ def isochrone(self, request, instance):
...
+                if request.get("_asynchronous_ridesharing", None) is None:
+                    # when the param is not set in the request, use the instance config
+                    return instance.asynchronous_ridesharing
+                else:
+                    return request.get("_asynchronous_ridesharing")
+
+            if not active_asynchronous_ridesharing(request, instance) or request.get('only_ridesharing', False):
```
```suggestion
            if not is_asynchronous_ridesharing(request, instance) or request.get('only_ridesharing', False):
```

source/jormungandr/jormungandr/interfaces/v1/journey_common.py
```diff
@@ -455,6 +455,19 @@ def __init__(self, output_type_serializer):
...
+            type=BooleanType(),
+            hidden=True,
+            default=False,
+            help="active the asynchronous mode for the ridesharing services",
+        )
+        parser_get.add_argument(
```
xlqian, line 465: haven't we finally decided to use `asynchronous_ridesharing`? I mean it won't be hidden

source/jormungandr/jormungandr/interfaces/v1/journey_common.py
```diff
@@ -455,6 +455,19 @@ def __init__(self, output_type_serializer):
             hidden=True,
             help="define the duration to reach stop points by crow fly",
         )
+        parser_get.add_argument(
+            "only_ridesharing",
+            type=BooleanType(),
+            hidden=True,
+            default=False,
+            help="active the asynchronous mode for the ridesharing services",
```
xlqian: it seems that the help message is not up to date

source/jormungandr/jormungandr/interfaces/v1/journey_common.py
xlqian: the hidden argument should be started with an underscore.

source/jormungandr/jormungandr/scenarios/new_default.py
```diff
@@ -1194,6 +1183,39 @@ def isochrone(self, request, instance):
...
+                logger.debug('trying to add ridesharing journeys')
+                try:
+                    decorate_journeys_with_ridesharing_offers(pb_response, instance, request)
+                except Exception:
+                    logger.exception('Error while retrieving ridesharing ads')
+                if request.get('only_ridesharing', False):
+                    for j in pb_response.journeys:
```
xlqian, line 1210: if `_ridesharing_only` is on, we should remove all other kraken_calls which don't contain ridesharing... so that we can slightly reduce the load of the platform.

https://github.com/CanalTP/navitia/blob/dev/source/jormungandr/jormungandr/scenarios/new_default.py#L126

xlqian (commented):
bla bla :)

source/jormungandr/jormungandr/scenarios/new_default.py
benoit-bst: I let the all multimodal possiblities happen (without filtering or disabled kraken call).
Because i don't want to biaise the choice algorithm.

source/jormungandr/jormungandr/instance.py
benoit-bst: I need your help for this part.
I not sure about this part of code.
This fonction, https://github.com/CanalTP/navitia/pull/3280/files#diff-24e77452880e6f3b205e94a7ba89e22fR190, return a value. Maybe i need to cast tha value in boolean ?

source/jormungandr/jormungandr/instance.py
```diff
@@ -180,6 +187,11 @@ def get_providers_from_db(self):
         """
         return self._get_models().equipment_details_providers
 
+    def get_asynchronous_ridesharing_from_db(self):
```
xlqian: try
```python
@property
def asynchronous_ridesharing_from_db(self):
```

source/jormungandr/jormungandr/scenarios/new_default.py
xlqian: you're right, we can do it later if we observe an obvious load problem

source/jormungandr/jormungandr/default_settings.py
```diff
@@ -24,6 +24,9 @@
 
 DISABLE_DATABASE = boolean(os.getenv('JORMUNGANDR_DISABLE_DATABASE', False))
 
+# Active the asynchronous ridesharing mode
+ASYNCHRONOUS_RIDESHARING = boolean(os.getenv('JORMUNGANDR_ASYNCHRONOUS_RIDESHARING', False))
```
xlqian, line 28: this config shouldn't be in global config

xlqian: this default value is already here: https://github.com/CanalTP/navitia/pull/3280/files#diff-3eb516880819f9f6747981aefb7a07f6R175, you don't need this in the global setting...

benoit-bst:
Thank @xlqian for the DB tips !

benoit-bst:
`partner_services=ridesharing` option list is added !

source/jormungandr/jormungandr/partner_services.py
```diff
@@ -0,0 +1,46 @@
...
+# channel `#navitia` on riot https://riot.im/app/#/room/#navitia:matrix.org
+# https://groups.google.com/d/forum/navitia
+# www.navitia.io
+
+from enum import Enum
+
+
+# The different partner services
+class PartnerServices(Enum):
+    ridesharing = "ridesharing"
```
benoit-bst, line 35: for the moment only **ridesharing** is added into _partner services_.
Soon VTC, and more...

source/jormungandr/jormungandr/interfaces/v1/journey_common.py
kadhikari: As mentioned in https://confluence.kisio.org/display/NAV/Covoiturage+%3A+le+RideSharing+dans+Navitia the parameter name should be asynchronous_ridesharing non?

kadhikari (approved):
Except the parameter name _asynchronous_ridesharing instead of asynchronous_ridesharing it seems fine to me :)

benoit-bst:
@kadhikari __asynchronous_ridesharing_ is an hidden parameter. It is for **debug purpose** (ask by @SGrenet), allow to change on a fly the mode old/new. But it will done by db configuration with the cute _navitia-configurator_

nonifier (approved):

xlqian (approved):

PR #3281, opened 2020-09-21
[Jormun] fix co2 emission unit for Here

benoit-bst:
**Here** provide a Co2 emission in **kEC** unit (https://developer.here.com/documentation/routing/dev_guide/topics/resource-type-route-summary.html)
But **navitia** uses **gEC**. A simple conversion fix the problem


SGrenet (approved):

nonifier (approved):
:tada:

PR #3285, opened 2020-09-25
[kraken + ed] remove clang warnings

pbougue:
After compiling in clang (been a while :smile:) some warnings appeared.

* easy warnings
* allow deprecated `proj` API (as my lib is newer): I still have `conv_coord_test` test breaking on that (56m distance between what's tested and the conversion using the newer version of PROJ). This PR does not fix the test (I don't know if we should accept any of the 2 values, only one, a range or something else).
* :warning: last commit is more debatable I think (I miss context on those lines and their adds): please review carefully

:mag: please review by commit (read comment first).

### DIFF
--- a/source/ed/connectors/conv_coord.h
+++ b/source/ed/connectors/conv_coord.h
@@ -29,6 +29,7 @@ www.navitia.io
 */
 
 #pragma once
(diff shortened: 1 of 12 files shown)


benoit-bst (approved):
Thanks for the cleaning :heart:

nonifier (approved):

PR #3288, opened 2020-09-30
[Jormun] update the way to choose between two similar journeys

pbench:
With the previous version of the filter, the order in which the journeys were compared changed the output.
For example, given the list of journeys [A,B,C] it would select A, and given the list [B, C, A] it would select C.
Now it should be independent of the order.

### DIFF
--- a/source/jormungandr/jormungandr/scenarios/journey_filter.py
+++ b/source/jormungandr/jormungandr/scenarios/journey_filter.py
@@ -514,6 +514,22 @@ def fallback_duration(journey):
     return duration
 
 
+def end_fallback_duration(journey):
+    for section in reversed(journey.sections):
+        if section.type in (response_pb2.STREET_NETWORK, response_pb2.CROW_FLY):
+            return section.duration
+
+    return 0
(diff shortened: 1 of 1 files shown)


kadhikari (approved):
:+1:

source/jormungandr/jormungandr/scenarios/journey_filter.py
```diff
@@ -514,6 +514,22 @@ def fallback_duration(journey):
     return duration
 
 
+def end_fallback_duration(journey):
+    for section in reversed(journey.sections):
+        if section.type in (response_pb2.STREET_NETWORK, response_pb2.CROW_FLY):
+            return section.duration
+
```
xlqian, line 521:
```python
duration = next( section.duration for section in reversed(journey.sections) if section.type in (response_pb2.STREET_NETWORK, response_pb2.CROW_FLY), None )
return duration or 0
```

xlqian (approved):

benoit-bst (approved):

PR #3291, opened 2020-10-01
[Raptor] ignore the max_extra_second_pass parameter and set it to zero

pbench:
To be merged with https://github.com/CanalTP/artemis_references/pull/226

### DIFF
--- a/source/kraken/worker.cpp
+++ b/source/kraken/worker.cpp
@@ -751,14 +751,14 @@ void Worker::journeys(const pbnavitia::JourneysRequest& request, pbnavitia::API 
                                            "no origin point nor destination point given");
             return;
         }
-
+        uint32_t max_extra_second_pass = 0;  // request.max_extra_second_pass()
         switch (api) {
             case pbnavitia::pt_planner:
                 routing::make_pt_response(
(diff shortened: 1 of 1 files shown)


xlqian (approved):

source/kraken/worker.cpp
```diff
@@ -751,14 +751,14 @@ void Worker::journeys(const pbnavitia::JourneysRequest& request, pbnavitia::API
                                            "no origin point nor destination point given");
             return;
         }
-
+        uint32_t max_extra_second_pass = 0;  // request.max_extra_second_pass()
```
kadhikari, line 754: We don't use any more the parameter  max_extra_second_pass ?

pbench: It is to deploy the raptor PR on all plaforms (dev, internal, etc.) without having to update the navitia-deployment-conf.
With the PR, the parameter should be zero, without the PR we want to keep using the parameters in navitia-deployment-conf.

benoit-bst (approved):

kadhikari (approved):

nonifier (approved):
Can we track somewhere to revert this once deployed ? 
We are so good at forgetting things like this :smile:

pbench:
> Can we track somewhere to revert this once deployed ?
> We are so good at forgetting things like this smile

I'll make a ticket :)

PR #3293, opened 2020-10-06
[Jormun] Ridesharing async mode with greenlet

benoit-bst:
## Desire
Parallelize ridesharing services call with greenlet technics.
It'll speed up the response time that can be very slow with sequential mode

```
with ridesharing providers :
- Rs provider 1
- Rs provider 2

response
|
|------journeys 1 --|----first section 1
|                   |----last section 1
|
|------journeys 2 --|----first section 2
|                   |----last section 2
|
|------journeys 2 --|----first section 3
|                   |----last section 3
```
For a response like this, we iterate sequentially for each sections and for each section, we call sequentially ridesharing providers.

calls are like this :

```
# sequentially
first section 1 - Rs provider 1
first section 1 - Rs provider 2

last section 1 - Rs provider 1
last section 1 - Rs provider 2

first section 2 - Rs provider 1
first section 2 - Rs provider 2

last section 2 - Rs provider 1
last section 2 - Rs provider 2

first section 3 - Rs provider 1
first section 3 - Rs provider 2

last section 3 - Rs provider 1
last section 3 - Rs provider 2
```

With the PR now we have two levels of parallelization :

```
        Pool main                          second Pool
|coroutine for first section 1 | coroutine for Rs provider 1
|                              | coroutine for Rs provider 2
|
|                                          second pool
|coroutine for last section 1  | coroutine for Rs provider 1
|                              | coroutine for Rs provider 2
|
|                                          second pool
|coroutine for first section 2 | coroutine for Rs provider 1
|                              | coroutine for Rs provider 2
|
|                                          second pool
|coroutine for last section 2  | coroutine for Rs provider 1
|                              | coroutine for Rs provider 2
|
|                                          second pool
|coroutine for first section 3 | coroutine for Rs provider 1
|                              | coroutine for Rs provider 2
|
|                                          second pool
|coroutine for last section 3  | coroutine for Rs provider 1
|                              | coroutine for Rs provider 2
```

## Results
For a simple call with _Instant System_ services (8 calls - 4 first section mode - 4 last section mode) the response time is about 13300 ms. With the async mode it is about 4890ms.

## Settings

For the moment, it is set up by a classic parameter that are not in the Tyr DB.
parameter `GREENLET_POOL_FOR_RIDESHARING_SERVICES=True/False`

- [x] TODO: Add config parameter into Jormun bdd, to change the value by the fly. In case of problems


source/jormungandr/jormungandr/default_settings.py
```diff
@@ -24,6 +24,11 @@
 
 DISABLE_DATABASE = boolean(os.getenv('JORMUNGANDR_DISABLE_DATABASE', False))
 
+# Active the asynchronous ridesharing mode
+ASYNCHRONOUS_RIDESHARING = boolean(os.getenv('JORMUNGANDR_ASYNCHRONOUS_RIDESHARING', False))
```
benoit-bst, line 28: it was an oblivion of the PR https://github.com/CanalTP/navitia/pull/3280

pbench:
From the explaination on your PR, I understand that we can mix 2 ridesharing providers on a single journey. 
Is it something we want to offer ?

benoit-bst:
> From the explaination on your PR, I understand that we can mix 2 ridesharing providers on a single journey.
Is it something we want to offer ?

@pbench yes it is possible. 
Example with **Paris**. For one coverage, you can have a bunch of **ridesharing providers** like _Blablacar_, _Klaxit_, and so on...

source/jormungandr/jormungandr/instance.py
```diff
@@ -765,10 +766,30 @@ def get_autocomplete(self, requested_autocomplete):
...
         fps = set()
-        for service in self.ridesharing_services:
-            rsjs, fp = service.request_journeys_with_feed_publisher(
-                from_coord, to_coord, period_extremity, limit
-            )
-            res.extend(rsjs)
-            fps.add(fp)
+        greenlet_pool_actived = app.config.get('GREENLET_POOL_FOR_RIDESHARING_SERVICES', False)
+
+        with FutureManager() as future_manager:
```
xlqian: :+1: 

It's a great idea to reuse the `FutureMananger`, but be careful, here, you're using the same greenlet configuration for `distributed`'s streetnetwork... and in terms of dependency, it'd better pull `FutureManager()` out of `jormungandr.scenarios.helper_classes` because `FutureManager()` was meant to be called only by `distributed`...

source/jormungandr/jormungandr/instance.py
```diff
@@ -765,10 +766,30 @@ def get_autocomplete(self, requested_autocomplete):
...
+            futures = []
+            for service in self.ridesharing_services:
```
xlqian:
```python
calls = []
res = []
fps = []

for service in self.ridesharing_services:
    def _call(s=service):
        return service.request_journeys_with_feed_publisher(from_coord, to_coord, period_extremity, limit)
    calls.append(_call)

call_res = []
if greenlet_pool_activated:
        futures = [future_manager.create_future(call) for call in calls]
    # it's just a generator, a lazy call 
    call_res = (f.wait_and_get() for f in futures)
else:
    call_res = (c() for c in calls)

for rsjs, fp in call_res:
    res.extend(rsjs)
    fps.add(fp)      
        
return res, fps

File: source/jormungandr/jormungandr/default_settings.py [hunk 4]
```diff
@@ -24,6 +24,11 @@
...
+# Active ridesharing service call with async greenlet
+GREENLET_POOL_FOR_RIDESHARING_SERVICES = boolean(os.getenv('JORMUNGANDR_GREENLET_POOL_FOR_RIDESHARING', False))
```
nonifier, line 30: Can we also have this in the configuration DB ?

source/jormungandr/jormungandr/instance.py
nonifier: oh yeah :tada:

nonifier (approved):

source/jormungandr/jormungandr/default_settings.py
benoit-bst: yes we need to have it, to change the value by the fly. In case of problems

source/jormungandr/jormungandr/instance.py
benoit-bst: I dunno why. If i create a new instance of FutureManager, this is a new greenlet pool ?

source/jormungandr/jormungandr/instance.py
benoit-bst: I think it didn't work.
I need to create future for all `service.request_journeys_with_feed_publisher(...)` call.
So when you do
```
for service in self.ridesharing_services:
    def _call(s=service):
        return service.request_journeys_with_feed_publisher(from_coord, to_coord, period_extremity, limit)
    calls.append(_call)
```
You call the function, and sequentially no ? Unless it return a functor ?

source/jormungandr/jormungandr/instance.py
xlqian: yes, there will be a new greenlet pool. i'm just talking about the configuration :) 
I mean the configuration: https://github.com/CanalTP/navitia/blob/dev/source/jormungandr/jormungandr/default_settings.py#L198  is shared by streetnetwork and async ridesharing.

xlqian: Note also that under the current architecture, the FutureMananger is meant to be used exclusively by streetnetwork. you can probablely pull this class out

source/jormungandr/jormungandr/instance.py
xlqian: `calls` contains a bunch of functors

xlqian: functors are triggered by:
```python
futures = [future_manager.create_future(call) for call in calls]
```

Disclaimer: I didn't test the code :)

source/jormungandr/jormungandr/default_settings.py
xlqian: :+1:

xlqian: I think you also need

`GREENLET_POOL_SIZE_FOR_RIDESHARING_SERVICES`

source/jormungandr/jormungandr/instance.py
benoit-bst: okay :+1: 
This will be a problem to mix greenlet pool size for streetnetwork and for ridesharing. I wll code a solution

xlqian (approved):

PR #3297, opened 2020-10-09
[Jormungadr] Overrided mode

xlqian:
https://jira.kisio.org/browse/NAVITIAII-3174

This PR will solve the case where both `bss` and `walking` are given as fallback mode. Since  `bss` includes `walking` potentially if the `walking` speed are set differently in both modes, the fallback may be incoherent . In this PR, when computing the Matrix, we will fill the matrix by picking the fastest fallback duration/mode.

### DIFF
--- a/source/jormungandr/jormungandr/scenarios/helper_classes/fallback_durations.py
+++ b/source/jormungandr/jormungandr/scenarios/helper_classes/fallback_durations.py
@@ -346,13 +346,8 @@ class FallbackDurationsPool(dict):
 
         for uri, duration_item in six.iteritems(main_fb):
             duration = duration_item.duration
-            # if the duration of the main_mode is smaller than all other mode, we keep the uri in the final res
-            # else the uri is removed because the fallback duration is worse than one of the overriding mode
-            # Note that:
-            #            all([]) == True
-            # So when
-            # 1. overriding_fbs is empty, we fill in the result with main_fb's content
-            # 2. uri can't be found in any of overriding_fbs, we fill in the result with main_fb's content
+            # if the duration of the main_mode is strictly smaller than all other mode, we keep the uri in the final res
+            # else the fallback duration is replaced by the smallest duration among all modes.
             mode_durations = [(main_mode, duration)] + [
                 (mode, fb.get(uri).duration) for mode, fb in overriding_fbs if fb.get(uri)
             ]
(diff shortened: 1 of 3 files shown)


source/jormungandr/jormungandr/scenarios/helper_classes/fallback_durations.py
```diff
@@ -349,6 +353,14 @@ def get_best_fallback_durations(self, main_mode):
...
             # 2. uri can't be found in any of overriding_fbs, we fill in the result with main_fb's content
-            if all(duration <= fb.get(uri).duration for fb in overriding_fbs if fb.get(uri)):
-                res[uri] = duration
+            mode_durations = [(main_mode, duration)] + [
+                (mode, fb.get(uri).duration) for mode, fb in overriding_fbs if fb.get(uri)
+            ]
+            best_mode, best_duration = min(mode_durations, key=lambda o: o[1])
+            if best_mode != main_mode:
+                self._overrided_uri_map[main_mode][uri] = best_mode
+            res[uri] = best_duration
```
pbench, line 357: Be careful for the case where `main_mode` and another `mode` have exactly the same duration, which one will be kept ? 
I would do something like : "I keep the first mode in the list overriding_modes that have a duration strictly lower than main_mode_duration (if any)"

source/jormungandr/jormungandr/scenarios/helper_classes/fallback_durations.py
```diff
@@ -349,6 +353,14 @@ def get_best_fallback_durations(self, main_mode):
             # So when
```
pbench: Since `all()` is not used anymore below, this comment should be removed

kadhikari:
I think some comments on the purpose of the modification and the tests will help the reviewers!

source/jormungandr/jormungandr/scenarios/helper_classes/fallback_durations.py
xlqian: why would you keep the "First" strictly smaller one? why not the "Smallest" one?

xlqian: (Just for record)
https://docs.python.org/3/library/functions.html#min : If multiple items are minimal, the function returns the first one encountered.

so there won't be any ambiguities when several modes have the same samllest value.

SGrenet (approved):

kadhikari (approved):

source/jormungandr/jormungandr/scenarios/helper_classes/fallback_durations.py
```diff
@@ -31,7 +31,7 @@
 
 import jormungandr.street_network.utils
 from navitiacommon import response_pb2
-from collections import namedtuple
+from collections import namedtuple, defaultdict
```
nonifier, line 34: I learned something new today with `defaultdict` :+1:

nonifier (approved):

PR #3299, opened 2020-10-12
[Debian 10] build_packages for debian 10

xlqian:
build_packages for debian 10


.github/workflows/build_navitia_packages_for_release_debian_10.yml
```diff
@@ -0,0 +1,48 @@
...
+on:
+  pull_request:
+
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+
+    container:
+      image: navitia/debian10_dev
```
nonifier: Why not using a matrix build on the previous script ? So we can do `[navitia/debian8_dev, navitia/debian10_dev]`
https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions#jobsjob_idstrategymatrix

.github/workflows/build_navitia_packages_for_dev.yml
```diff
@@ -1,17 +1,20 @@
 name: Build Navitia Packages For Dev
 
 on:
-  push:
-    branches:
-      - dev
+  pull_request:
```
xlqian: just for test

nonifier:
Do you plan on keeping the original file ? 
https://github.com/CanalTP/navitia/blob/dev/.github/workflows/build_navitia_packages_for_dev.yml

benoit-bst (approved):

xlqian:
@nonifier we should remove the old one, once all CI jobs are migrated to the new one

nonifier (approved):

PR #3303, opened 2020-10-19
[raptor] count stay_in as a transfer in the pareto front

pbench:
A stay-in in a journey didn't count as a transfer in the comparison operator used in the Pareto front.
As a results, we obtained journeys that used a stay-in just to gain a few seconds of walking or arrive a few seconds earlier.

This PR count a stay-in as a transfer.

Artemis run https://ci.navitia.io/view/custom_artemis/job/custom_Artemis/47/

To be merged with https://github.com/CanalTP/artemis_references/pull/229

- [x] waiting for some QA before merge

### DIFF
--- a/source/routing/journey.cpp
+++ b/source/routing/journey.cpp
@@ -53,14 +53,16 @@ bool Journey::is_pt() const {
 bool Journey::better_on_dt(const Journey& that,
                            bool request_clockwise,
                            const navitia::time_duration transfer_penalty) const {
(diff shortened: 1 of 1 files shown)


nonifier (approved):
Nice one ! And with a proper parameter for both transfer and stay-in penalty, we'll be in a very nice spot :+1:

xlqian (approved):

SGrenet (approved):

PR #3312, opened 2020-10-29
[Jormungandr] Add lines in StopArea and StopPoint when depth = 3 in the query

kadhikari:
Some modifications:
- Display commercial_mode and physical_modes in Line even without parameter depth(depth = 0) as basic attributes.
- with the parameter depth = 3 display lines in
-- StopArea (Line with basic attributes).
-- StopPoint (Line with basic attributes).

For details see : https://jira.kisio.org/browse/NAVP-1630


benoit-bst (approved):

nonifier (approved):
Looking good :+1:

PR #3315, opened 2020-11-03
[Jormungandr] Add a deeplink in a on_demand_transport section

kadhikari:
For details: 
https://jira.kisio.org/browse/NAVP-1625
and 
https://jira.kisio.org/browse/NAVP-1626

### DIFF
--- a/source/jormungandr/jormungandr/interfaces/v1/Journeys.py
+++ b/source/jormungandr/jormungandr/interfaces/v1/Journeys.py
@@ -46,7 +46,7 @@ from jormungandr.interfaces.v1.make_links import (
 from jormungandr.interfaces.v1.errors import ManageError
 from collections import defaultdict
 from navitiacommon import response_pb2
-from jormungandr.utils import date_to_timestamp, dt_to_str
+from jormungandr.utils import date_to_timestamp, dt_to_str, has_invalid_reponse_code, journeys_absent
 from jormungandr.interfaces.v1.serializer import api
 from jormungandr.interfaces.v1.decorators import get_serializer
 from navitiacommon import default_values
@@ -108,7 +108,7 @@ class add_journey_href(object):
         @wraps(f)
         def wrapper(*args, **kwargs):
             objects = f(*args, **kwargs)
-            if objects[1] != 200 or 'journeys' not in objects[0]:
+            if has_invalid_reponse_code(objects) or journeys_absent(objects):
                 return objects
 
             for journey in objects[0]['journeys']:
@@ -186,9 +186,7 @@ class add_fare_links(object):
         @wraps(f)
         def wrapper(*args, **kwargs):
             objects = f(*args, **kwargs)
-            if objects[1] != 200:
-                return objects
-            if "journeys" not in objects[0]:
+            if has_invalid_reponse_code(objects) or journeys_absent(objects):
                 return objects
 
             if 'tickets' not in objects[0]:
@@ -231,9 +229,7 @@ class add_tad_links(object):
         @wraps(f)
         def wrapper(*args, **kwargs):
             objects = f(*args, **kwargs)
-            if objects[1] != 200:
-                return objects
-            if "journeys" not in objects[0]:
+            if has_invalid_reponse_code(objects) or journeys_absent(objects):
                 return objects
 
             for j in objects[0]['journeys']:
@@ -303,15 +299,15 @@ class rig_journey(object):
         @wraps(f)
         def wrapper(*args, **kwargs):
             objects = f(*args, **kwargs)
-            response, status, _ = objects
-            if status != 200:
+            if has_invalid_reponse_code(objects):
                 return objects
 
             if not hasattr(g, 'origin_detail') or not hasattr(g, 'destination_detail'):
                 return objects
 
+            response = objects[0]
             for j in response.get('journeys', []):
-                if not 'sections' in j:
+                if 'sections' not in j:
                     continue
                 logging.debug(
                     'for journey changing origin: {old_o} to {new_o}'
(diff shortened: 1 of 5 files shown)


source/jormungandr/jormungandr/interfaces/v1/Journeys.py
```diff
@@ -217,6 +224,68 @@ def wrapper(*args, **kwargs):
...
+                if "sections" not in j:
+                    continue
+                for s in j['sections']:
+                    # For a section with type = on_demand_transport
+                    if s.get('type') == 'on_demand_transport':
+                        # get network uri from the link
+                        networks = [link['id'] for link in s.get('links', []) if link['type'] == "network"]
+                        if len(networks) == 0:
+                            continue
+                        network_id = networks[0]
```
azime:
```suggestion
                        network_id = next((link['id'] for link in s.get('links', []) if link['type'] == "network"), None)
                        if not network_id:
                            continue
```

source/jormungandr/jormungandr/interfaces/v1/Journeys.py
```diff
@@ -217,6 +224,68 @@ def wrapper(*args, **kwargs):
...
+                        region = kwargs.get('region')
+                        if region is None:
+                            continue
+
+                        # Get the Network details and verify if it contains codes with type = "app_code"
+                        instance = i_manager.instances.get(region)
+                        network_details = instance.ptref.get_objs(
+                            type_pb2.NETWORK, 'network.uri={}'.format(network_id)
+                        )
+                        network_dict = protobuf_to_dict(next(network_details))
```
azime, line 264: I don't know if necessary to do "protobuf_to_dict"

source/jormungandr/jormungandr/interfaces/v1/Journeys.py
```diff
@@ -217,6 +224,68 @@ def wrapper(*args, **kwargs):
...
+                        app_value = [
+                            code['value']
+                            for code in network_dict.get('codes', [])
+                            if code.get('type') == "app_code"
+                        ]
+                        if len(app_value) == 0:
+                            continue
```
azime:
```suggestion
                        app_value = next((
                            code['value']
                            for code in network_dict.get('codes', [])
                            if code.get('type') == "app_code"
                        ), None)
                        if not app_value:
                            continue
```

source/jormungandr/jormungandr/interfaces/v1/make_links.py
```diff
@@ -81,6 +81,13 @@ def create_internal_link(rel, _type, id, templated=False, description=None):
     return d
 
 
+def make_external_service_link(url, rel, _type, templated=False, **kwargs):
+    call_params = ""
+    for key, value in kwargs.items():
+        call_params += "{}={}&".format(key, value)
```
azime:
```suggestion
    call_params = "&".join(["{}={}".format(key, value) for key, value in kwargs.iteritems()])
```

source/jormungandr/jormungandr/interfaces/v1/Journeys.py
```diff
@@ -217,6 +224,68 @@ def wrapper(*args, **kwargs):
...
 
 
+# Add TAD deep links in each section of type on_demand_transport if network contains app_code in codes
+class add_tad_links(object):
+    def __call__(self, f):
+        @wraps(f)
+        def wrapper(*args, **kwargs):
+            objects = f(*args, **kwargs)
+            if objects[1] != 200:
+                return objects
```
nonifier: Can we have this in a separate function please (`has_invalid_response_code`) that can be shared across all journey decorators ? 
It would be easier to read/understand, and if the response ever changes (if the code later moves to objects[2] for instance, or if we want to decorate 201/203 etc..) it would be much easier to maintain.

(this applies for the 2 lines below as well)

kadhikari: :+1:

source/jormungandr/jormungandr/interfaces/v1/Journeys.py
kadhikari: :+1:

source/jormungandr/jormungandr/interfaces/v1/Journeys.py
kadhikari: :+1:

source/jormungandr/jormungandr/interfaces/v1/Journeys.py
kadhikari: We need to transform as they don't have same structures

source/jormungandr/jormungandr/interfaces/v1/make_links.py
kadhikari: :+1:

nonifier (approved):
Much nicer !

azime (approved):

source/jormungandr/jormungandr/utils.py
```diff
@@ -55,6 +55,7 @@
 PY3 = sys.version_info[0] == 3
 
 DATETIME_FORMAT = "%Y%m%dT%H%M%S"
+UTC_DATETIME_FORMAT = "%Y-%m-%dT%H:%M:%S%z"
```
nonifier, line 58:
```suggestion
UTC_DATETIME_FORMAT = "%Y%m%dT%H%M%SZ"
```
I agree that local date times have to be expressed in UTC. 
But, the **format** needs to be consistent across the API. As per our documentation : 
```
Navitia exposes every date times as local times of the coverage via an ISO 8601 “YYYYMMDDThhmmss” string
```
Hence, `2012-06-15T11:00:00+0000` (the extended format) can be expressed using the base format by appending a `Z` at the end as :  `20120615T110000Z`.

https://en.wikipedia.org/wiki/ISO_8601#Coordinated_Universal_Time_(UTC)

nonifier: AFter discussion with @kadhikari , we will leave it that way, and come back to it with a proper way of generating deeplinks based on the connector.

nonifier (approved):

benoit-bst (approved):
Huge work !

PR #3320, opened 2020-11-10
[Places_nearby] filter with 'or' having atleast one matching criteria should work 

kadhikari:
Do not catch the exception on filter where there are more than one object types in places_nearby.
Ticket : https://jira.kisio.org/browse/NAVITIAII-3227

### DIFF
--- a/source/proximity_list/proximitylist_api.cpp
+++ b/source/proximity_list/proximitylist_api.cpp
@@ -127,6 +127,10 @@ void find(navitia::PbCreator& pb_creator,
         if (!filter.empty()) {
             try {
                 indexes = ptref::make_query(type, filter, forbidden_uris, *pb_creator.data);
+            } catch (const ptref::parsing_error& parse_error) {
+                pb_creator.fill_pb_error(pbnavitia::Error::unable_to_parse,
+                                         "Problem while parsing the query:" + parse_error.more);
+                return;
             } catch (const std::exception&) {
             }
         }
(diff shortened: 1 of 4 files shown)


SGrenet (approved):

SGrenet:
Missing test from impact to dataset can be avoided, it's not a real use case

source/proximity_list/proximitylist_api.cpp
```diff
@@ -131,9 +131,7 @@ void find(navitia::PbCreator& pb_creator,
                 pb_creator.fill_pb_error(pbnavitia::Error::unable_to_parse,
                                          "Problem while parsing the query:" + parse_error.more);
                 return;
-            } catch (const ptref::ptref_error& pt_error) {
-                pb_creator.fill_pb_error(pbnavitia::Error::bad_filter, "ptref : " + pt_error.more);
-                return;
+            } catch (const std::exception&) {
```
benoit-bst, line 134: I don't know why this is empty ?

benoit-bst: maybe you should unwind global errors with a generic message ?

kadhikari: Here we catch error but continue to search for the rest object types in the query (request). As exception is raised, we are obliged to catch and continue.

nonifier: I got confused at first because I thought we were using `pt_ref` to figure out the proximity list. But `pt_ref` is only here to find out which objects are filtered.

nonifier (approved):

PR #3321, opened 2020-11-16
[Autocomplete] Add filter in pt_objects request

kadhikari:
Ticket: https://jira.kisio.org/browse/NAVP-1702

### DIFF
--- a/source/autocomplete/autocomplete_api.cpp
+++ b/source/autocomplete/autocomplete_api.cpp
@@ -366,7 +366,8 @@ void autocomplete(navitia::PbCreator& pb_creator,
                   const std::vector<std::string>& admins,
                   int search_type,
                   const navitia::type::Data& d,
-                  float main_stop_area_weight_factor) {
+                  float main_stop_area_weight_factor,
+                  const std::string& ptref_filter) {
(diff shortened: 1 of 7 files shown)


nonifier (approved):

source/autocomplete/autocomplete_api.cpp
```diff
@@ -389,8 +390,28 @@ void autocomplete(navitia::PbCreator& pb_creator,
...
+                    try {
+                        indexes = ptref::make_query(type, ptref_filter, d);
+                    } catch (const ptref::parsing_error& parse_error) {
+                        pb_creator.fill_pb_error(pbnavitia::Error::unable_to_parse,
+                                                 "Problem while parsing the query:" + parse_error.more);
+                        return;
+                    } catch (const std::exception&) {
+                    }
+                }
+            }
```
nonifier, line 407: If we want to keep the code clean, it's nice to follow the SOLID principles. Especially the S - https://en.wikipedia.org/wiki/Single-responsibility_principle. 
A function should do one thing and a Class have one responsibility. Hence moving this into a separate function.

benoit-bst (approved):

PR #3322, opened 2020-11-16
[Jormungandr] missing ridesharing db parameters fetching to parallelize calls

benoit-bst:
It was an oblivion of https://github.com/CanalTP/navitia/pull/3293

the db is ready https://github.com/CanalTP/navitia/pull/3293/files#diff-f5db4f5cc492be0758738f0d601cd8883ee9c62888186df6264804b530570e43 but we don't use it...

### DIFF
--- a/source/jormungandr/jormungandr/scenarios/ridesharing/ridesharing_service_manager.py
+++ b/source/jormungandr/jormungandr/scenarios/ridesharing/ridesharing_service_manager.py
@@ -58,8 +58,6 @@ class RidesharingServiceManager(object):
         self._last_update = datetime.datetime(1970, 1, 1)
         self._update_interval = update_interval
         self.instance = instance
-        self.greenlet_pool_size = app.config.get('RIDESHARING_GREENLET_POOL_SIZE', 8)
-        self.greenlet_pool_actived = app.config.get('GREENLET_POOL_FOR_RIDESHARING_SERVICES', False)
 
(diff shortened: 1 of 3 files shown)


source/jormungandr/jormungandr/scenarios/ridesharing/ridesharing_service_manager.py
```diff
@@ -160,13 +158,13 @@ def get_all_ridesharing_services(self):
         return self._ridesharing_services_legacy + list(self._ridesharing_services.values())
 
     def decorate_journeys_with_ridesharing_offers(self, response, request, instance):
-        # TODO: disable same journey schedule link for ridesharing journey?
-        if self.greenlet_pool_actived:
+        greenlet_pool_actived = instance.greenlet_pool_for_ridesharing_services
```
kadhikari, line 161: 'greenlet_pool_actived' can be replaced by instance.greenlet_pool_for_ridesharing_services Non?

kadhikari (approved):

source/jormungandr/jormungandr/scenarios/ridesharing/ridesharing_service_manager.py
benoit-bst: this is for calling the db one time

nonifier (approved):
You can blame the reviewers for having missing this :smile:

PR #3324, opened 2020-11-18
jormungandr: Limiting param count value to 200 except for journeys

kadhikari:
Ticket: https://jira.kisio.org/browse/NAVITIAII-3240


SGrenet (changes requested):
max_count has to be at 200 for /places only
other endpoints don't have to change now

(except journeys in my opinion for which max_count 1000 is too much, but is an other subject because isochron can be impacted)

SGrenet (approved):

xlqian (approved):

azime (approved):

nonifier (approved):

PR #3332, opened 2020-11-30
[Tyr] detect and unzip compressed osm files sent to tyr

pbench:
For some reason, Artemis NG send zipped osm files to tyr, but tyr was only able to handle raw .pbf files, hence the auvergne dataset was not able to be set up, and artemis failed.
Here we detect if a .zip was sent to tyr, and in this case unzip it before calling osm2ed.


nonifier (approved):
Good catch !

kadhikari (approved):
:+1:

xlqian (approved):

PR #3338, opened 2020-12-09
[CI] Build packages in workflow

pbench:
This PR aims at building navitia packages on the CI workflow.

At first I tried to build everything (binaries and tests) only once, but I hit the space limitations on github runners (17Go).
So I went back to doing two builds : 
 - one with the tests (and STRIP_SYMBOLS) to be able to run all tests
 - a second one without the tests (but without STRIP_SYMBOLS) to build the packages

This basically amounts to the merging the previous worflow.yml and build_navitia_packages_for_dev_multi_distribution.yml.

We could also keep the two seperate worflow (worflow.yml and build_navitia_packages_for_dev_multi_distribution.yml) but run  build_navitia_packages_for_dev_multi_distribution.yml on pull requests.

This is a first step towards the goal to run Artemis NG on the new CI and on pull requests.
Next steps : 
- build navitia dockers from packages dowloaded from github
- launch Artemis NG from these dockers


xlqian:
I agree with your idea in principle. The only thing that bothers me is that the check time will be doubled... 

> We` could also keep the two seperate worflow (worflow.yml and build_navitia_packages_for_dev_multi_distribution.yml) but run build_navitia_packages_for_dev_multi_distribution.yml on pull requests.

IMHO How about using another job for building the packages?

pbench:
> I agree with your idea in principle. The only thing that bothers me is that the check time will be doubled...

That also bothers me !

> IMHO How about using another job for building the packages?

This is the other option I see.

If we agree that having two jobs is better, then I think the easiest way to do it is :

- keep worflow.yml as before this PR
- just change build_navitia_packages_for_dev_multi_distribution.yml so that it also runs on pull_request

pbench:
Looking at the runs to have some numbers :

- before this PR [the CI job took ~ 1h](https://github.com/CanalTP/navitia/actions/runs/408029313) 
- after this PR [the CI job took ~ 1h20mn](https://github.com/CanalTP/navitia/actions/runs/410230029)

pbench:
I'll close this one, an open another where building packages for pull request is done in another worflow

PR #3339, opened 2020-12-09
Impact: Fix application period end value

kadhikari:
* In the module pb_converter, period.last() is used while setting protobuf application_period.end() value in the response.
* This value is smaller (-1) than the real application.end() value used while crated in chaos.
Concerned ticket: https://jira.kisio.org/browse/NAVITIAII-3265

Before:
```javascript
"application_periods": ⊖[
       ⊖{
            "begin": "20201201T154000",
            "end": "20210228T223959"
        }
    ]
```

After:
```javascript
"application_periods": ⊖[
       ⊖{
            "begin": "20201201T154000",
            "end": "20210228T224000"
        }
    ]
```

### DIFF
--- a/source/jormungandr/tests/chaos_disruptions_tests.py
+++ b/source/jormungandr/tests/chaos_disruptions_tests.py
@@ -662,7 +662,7 @@ class TestChaosDisruptionsBlockingOverlapping(ChaosDisruptionsFixture):
         pt_sections = get_pt_sections(response)
         assert len(pt_sections) == 1
         assert pt_sections[0]['departure_date_time'] == '20120614T080100'
-        assert pt_sections[0]['arrival_date_time'] == u'20120614T080102'
+        assert pt_sections[0]['arrival_date_time'] == '20120614T080102'
         line_id = next(o['id'] for o in pt_sections[0]['links'] if o['type'] == 'line')
         assert line_id == 'A'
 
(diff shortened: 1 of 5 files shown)


benoit-bst (approved):

source/jormungandr/tests/chaos_disruptions_tests.py
```diff
@@ -640,17 +640,130 @@ def test_disruption_on_journey(self):
...
+        # Check that the journey with PT exists and line 'A' is used in PT
+        response = self.query_region(journey_basic_query + "&data_freshness=realtime")
+        assert contains_pt_journey(response) is True
+        pt_sections = get_pt_sections(response)
+        assert len(pt_sections) == 1
+        assert pt_sections[0]['departure_date_time'] == '20120614T080100'
+        assert pt_sections[0]['arrival_date_time'] == u'20120614T080102'
+        line_id = next(o['id'] for o in pt_sections[0]['links'] if o['type'] == 'line')
+        assert line_id == 'A'
+
```
SGrenet, line 720: My observations : 
In step 1 : ok, disruption don't impact line A, but is this ignored disruption displayed on this journey ?

source/jormungandr/tests/chaos_disruptions_tests.py
```diff
@@ -640,17 +640,130 @@ def test_disruption_on_journey(self):
...
+        # Check that the journey with PT exists but line 'M' is used in PT as 'A' is impacted
+        response = self.query_region(journey_basic_query + "&data_freshness=realtime")
+        assert contains_pt_journey(response) is True
+        pt_sections = get_pt_sections(response)
+        assert len(pt_sections) == 1
+        assert pt_sections[0]['departure_date_time'] == '20120614T080101'
+        assert pt_sections[0]['arrival_date_time'] == u'20120614T080103'
+        line_id = next(o['id'] for o in pt_sections[0]['links'] if o['type'] == 'line')
+        assert line_id == 'M'
+
```
SGrenet, line 842: theses 3 steps test 3 limits : 
- disruption end = journey start
- disruption start = journey end
- disrution start 1s before journey end

one limit isn't tested : 
- disruption end 1s after journey start

SGrenet (changes requested):

source/jormungandr/tests/chaos_disruptions_tests.py
kadhikari: I will add another test : STEP 4 along with verification of disruptions in journey for each  STEP.

source/jormungandr/tests/chaos_disruptions_tests.py
kadhikari: Will be added

pbougue (commented):
Nice to add more tests on the way of this work.

This seems reasonable as it only changes display, no internal mechanism.
As you checked that no other output exists using disruption's application period `.last()` no inconsistencies are expected (I don't have much experience, so don't remember anything opposed to that).
Integrators might be impacted only if they previously "badly" compensated the 1 second they wanted to print instead of rounding, but probably not visible.
Functional inconsistency of not displaying for day 2 a disruption finishing at 00:00:00 on day 2 is accepted as I understand it (and it's OK to me).

SGrenet (approved):

source/jormungandr/tests/chaos_disruptions_tests.py
```diff
@@ -640,17 +640,207 @@ def test_disruption_on_journey(self):
         We delete the network disruption and test if there's a journey using
         the line B.
         """
-        response = self.query_region(journey_basic_query)
 
+        def contains_pt_journey(resp):
+            return any(
+                [
+                    len([s for s in j["sections"] if s["type"] == "public_transport"]) == 1
```
pbench, line 647:
```suggestion
                    len([s for s in j["sections"] if s["type"] == "public_transport"]) >= 1
```
otherwise this would return true if only if there is a journey which with exactly one section of public_transport.
Given the name of the function, I expect it to return true if resp contains a journey with 2 sections of public_transport

pbench (approved):
minor suggestion, otherwise looks :ok_hand:  !

PR #3342, opened 2020-12-23
[Jormun] Fix - Timeo should raise on HTTP error

nonifier:
When a call to Timeo fail with a HTTP status code of 400 and above, we were silently ignoring the call.

This PR changes the behaviour to make sure an exception is raised when an error occurred so we can monitor in New Relic failing calls.


SGrenet (approved):

SGrenet:
give link with https://jira.kisio.org/browse/NAVITIAII-3287 in description please

azime (approved):

PR #3350, opened 2021-01-11
[Jenkins] Trigger deploy on Artemis machine after a push

benoit-bst:
Related to : https://github.com/CanalTP/fabric_navitia/pull/320

### DIFF
--- a/.github/workflows/build_navitia_packages_for_dev_multi_distribution.yml
+++ b/.github/workflows/build_navitia_packages_for_dev_multi_distribution.yml
@@ -52,5 +52,5 @@ jobs:
     - name: run artemis NG on https://jenkins-core.canaltp.fr/job/artemis_ng
       run: http -v -f POST https://${{secrets.JENKINS_NG_TOKEN}}@jenkins-core.canaltp.fr/job/artemis_ng/buildWithParameters
     - name: run deploy on artemis machine https://jenkins-core.canaltp.fr/job/deploy-navitia
-      if: ${{ github.event_name == 'pull_request' }}
+      if: ${{ github.event_name == 'push' }}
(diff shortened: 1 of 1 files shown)


.github/workflows/build_navitia_packages_for_dev_multi_distribution.yml
```diff
@@ -51,3 +51,6 @@ jobs:
       run: apt update && apt install -y httpie
     - name: run artemis NG on https://jenkins-core.canaltp.fr/job/artemis_ng
       run: http -v -f POST https://${{secrets.JENKINS_NG_TOKEN}}@jenkins-core.canaltp.fr/job/artemis_ng/buildWithParameters
+    - name: run deploy on artemis machine https://jenkins-core.canaltp.fr/job/deploy-navitia
+      if: ${{ github.event_name == 'push' }}
```
pbench, line 55: are the `}}` at the right position ?
```suggestion
      if: ${{ github.event_name }} == 'push' 
```

benoit-bst: i found this syntax here:  https://docs.github.com/pt/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions#jobsjob_idstepsif
```
steps:
 - name: My first step
   if: ${{ github.event_name == 'pull_request' && github.event.action == 'unassigned' }}
   run: echo This event is a pull request that had an assignee removed.
```

pbench: allright, that's one of the mysteries of syntax !

pbench (approved):

PR #3353, opened 2021-01-12
Hide pwd in tyr

xlqian:
https://jira.kisio.org/browse/NAVITIAII-3285


azime (commented):
Curious, why not use a filter in the logger ?

nonifier (approved):
:see_no_evil:

azime (approved):

PR #3354, opened 2021-01-12
correct trigger of artemis_ng on pull request

pbench:


benoit-bst (approved):
I don't know if it works. 
I propose you to follow the syntax bellow to trig deploy-navitia. As you want !!

pbench:
> I don't know if it works.
> I propose you to follow the syntax bellow to trig deploy-navitia. As you want !!

Your syntax seems to also work. I'll keep yours as it is more readable

PR #3358, opened 2021-01-21
[ridesharing] Karaos - fix main shape without useless coords

benoit-bst:
### issue

**Karos** sends 2 fields`driverArrival` and `driverDeparture`, that we don't want to use. 
Because it doesn't represent the start and the arrival of the offer. Just the start and the arrival of the **driver**. 
We directly use shape to catch arrival and departure

UTest coords was false. Now it is fix

**Real examples** :

**1. Ride**

before
ridesharing_karos_1
after
ridesharing_karos_avec_correctif_1

**2. Ride**

before
ridesharing_karos_2
after
ridesharing_karos_avec_correctif_2


azime:
:ok_hand:

azime (approved):

kadhikari (approved):
:1st_place_medal: for Ridesharing

PR #3362, opened 2021-01-28
[Tyr] Set `distributed` as default scenario for Navitia's instances

nonifier:
Update our model so that default scenario are set to `distributed` for new coverages


xlqian (approved):
:partying_face:

benoit-bst (approved):

PR #3370, opened 2021-02-24
[Raptor] Labels refacto

nonifier:
⚠️ Please read per commit for a better understanding ⚠️ 

Depends on : 
 - [x] https://github.com/CanalTP/utils/pull/96
 - [x] https://github.com/CanalTP/utils/pull/97

This PR is purely about refactoring how Labels are used and implemented within Raptor. 
There is NO functional change ( no itinerary has been harmed during this refacto ⛔ 🔪 ).

Motivation : 
 * The re-use a the Label concept (and code) across Raptor implementation
 * Preparing the ground for a route-point based Label (instead of Stop-point based)
 * Better data localisation has label internal data are now inlined.

Performance tests have been done using Artemis IDFM, and this PR gives the same performance results as `dev` branch (it is actually slightly better by a 1% or 2% margin but it's not significant).


source/routing/raptor.cpp
```diff
@@ -339,8 +335,8 @@ Journey convert_to_bound(const StartingPointSndPhase& sp, bool clockwise) {
 }  // namespace
 
 // copy and do the off by one for strict comparison for the second pass.
-static IdxMap<type::StopPoint, DateTime> snd_pass_best_labels(const bool clockwise,
-                                                              IdxMap<type::StopPoint, DateTime> best_labels) {
+static IdxMap<type::StopPoint, DateTime>& snd_pass_best_labels(const bool clockwise,
```
pbench: Before this function was copying its input, modifying the copy, and returning the copy.
Now it modify its input and returns it.
Is it ok ?

source/routing/raptor.cpp
```diff
@@ -438,14 +440,17 @@ RAPTOR::Journeys RAPTOR::compute_all_journeys(const map_stop_point_duration& dep
     auto starting_points = make_starting_points_snd_phase(*this, calc_dest, accessibilite_params, clockwise);
     swap(labels, first_pass_labels);
 
+    const auto inrows_labels = best_labels.inrow_labels();
+
     auto best_labels_for_snd_pass = Labels(
```
pbench: Here you create  `inrows_labels`, a full copy  of the data present in `best_labels`, and then you create another copy of this data in `best_labels_for_snd_pass`.
Before, the temporary copy with `inrows_labels` was not needed. 
Could you avoid creating it ?

source/routing/raptor.cpp
nonifier: It's now modifying the copy provided by ` const auto inrows_labels = best_labels.inrow_labels();` (cf you comment below :) )

source/routing/raptor.cpp
nonifier: Good catch, Label is now taking `const ref` Map as contructor parameter to avoid the copy 👍🏻

source/routing/labels.cpp
```diff
@@ -0,0 +1,50 @@
+
```
xlqian, line 30:
```suggestion
/* Copyright © 2001-2021, Kisio Digital and/or its affiliates. All rights reserved.
This file is part of Navitia,
    the software to build cool stuff with public transport.
Hope you'll enjoy and contribute to this project,
    powered by Canal TP (www.canaltp.fr).
Help us simplify mobility and open public transport:
    a non ending quest to the responsive locomotion way of traveling!
LICENCE: This program is free software; you can redistribute it and/or modify
it under the terms of the GNU Affero General Public License as published by
the Free Software Foundation, either version 3 of the License, or
(at your option) any later version.
This program is distributed in the hope that it will be useful,
but WITHOUT ANY WARRANTY; without even the implied warranty of
MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
GNU Affero General Public License for more details.
You should have received a copy of the GNU Affero General Public License
along with this program. If not, see <http://www.gnu.org/licenses/>.
Stay tuned using
twitter @navitia
channel `#navitia` on riot https://riot.im/app/#/room/#navitia:matrix.org
https://groups.google.com/d/forum/navitia
www.navitia.io
*/
```

source/routing/labels.h
```diff
@@ -0,0 +1,84 @@
+
```
xlqian, line 30: /* Copyright © 2001-2014, Kisio Digital and/or its affiliates. All rights reserved.
This file is part of Navitia,
    the software to build cool stuff with public transport.
Hope you'll enjoy and contribute to this project,
    powered by Canal TP (www.canaltp.fr).
Help us simplify mobility and open public transport:
    a non ending quest to the responsive locomotion way of traveling!
LICENCE: This program is free software; you can redistribute it and/or modify
it under the terms of the GNU Affero General Public License as published by
the Free Software Foundation, either version 3 of the License, or
(at your option) any later version.
This program is distributed in the hope that it will be useful,
but WITHOUT ANY WARRANTY; without even the implied warranty of
MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
GNU Affero General Public License for more details.
You should have received a copy of the GNU Affero General Public License
along with this program. If not, see <http://www.gnu.org/licenses/>.
Stay tuned using
twitter @navitia
channel `#navitia` on riot https://riot.im/app/#/room/#navitia:matrix.org
https://groups.google.com/d/forum/navitia
www.navitia.io
*/

source/routing/labels.cpp
```diff
@@ -0,0 +1,50 @@
...
+Labels::Labels(const Map& dt_pts, const Map& dt_transfers, const Map& walkings, const Map& walking_transfers) {
+    const auto& zip = boost::combine(dt_pts, dt_transfers.values(), walkings.values(), walking_transfers.values());
+    labels.resize(zip.size());
+    for (const auto& z : zip) {
+        const auto sp_idx = z.get<0>().first;
+        labels[sp_idx] = Label{z.get<0>().second, z.get<1>(), z.get<2>(), z.get<3>()};
+    }
+}
+
+std::array<Labels::Map, 4> Labels::inrow_labels() {
```
xlqian, line 49: it seems `row_labels` is built based on `labels`, can't you just make `row_labels` an member of `Labels` class then initialize `row_labels` in the constructor and return a reference here?

source/routing/labels.cpp
```diff
@@ -0,0 +1,50 @@
...
+
+    return row_labels;
+}
+
+void Labels::fill_values(DateTime pts, DateTime transfert, DateTime walking, DateTime walking_transfert) {
+    Label default_label{pts, transfert, walking, walking_transfert};
+    boost::fill(labels.values(), default_label);
+}
+
+void Labels::init(const std::vector<type::StopPoint*>& stops, DateTime val) {
```
xlqian, line 73: can't `init` be a part of the constructor?

source/routing/labels.cpp
nonifier: Label are now stored as an array of structure with :
```cpp
    using LabelsMap = IdxMap<type::StopPoint, Label>;
```
This way I will be able to only change the key value (to Route_ point), this will help the future work without changing Raptor's implementation : 
```cpp
    using LabelsMap = IdxMap<type::ROUTEPoint, Label>;
```

The problem is that Raptor is switching its transfers and non transfers labels (cf, https://github.com/CanalTP/navitia/blob/dev/source/routing/raptor.cpp#L444-L448). So being able to represent labels in row, help doing this. Having row_labels as a member class would mean keeping and maintaining 2 versions of the labels which is not ideal.

source/routing/labels.cpp
nonifier: I wish. but Raptor doesn't use RAII pattern in its implementation. 
Raptor re-use previously created label to initialize/clear previous one.
Apparently this is to "fast init labels with a memcpy"
https://github.com/CanalTP/navitia/blob/dev/source/routing/dataraptor.h#L105-L107.

I'm not sure there is actually much benefits from doing this copy. But I didn't want to change everything at once and there is no need to change it at the moment.

source/routing/labels.cpp
nonifier: good catch, it's now been added !

source/routing/labels.h
nonifier: it's been added 👍🏻

xlqian (approved):

azime (approved):

PR #3371, opened 2021-02-25
Kraken: Fix headsign for transilien coverage

azime:
Fix  headsign for transilien

Jira: [NAVITIAII-3320](https://jira.kisio.org/browse/NAVITIAII-3320)

before modification (Dev env):
image

After modification (Local env): 
image


SGrenet (approved):

kadhikari (commented):
I don't understand why there is a difference of direction: POISSY (Poissy), -> POISSY ??

benoit-bst (approved):

SGrenet:
> 
> 
> I don't understand why there is a difference of direction: POISSY (Poissy), -> POISSY ??

Probably : 1st screen on dev and 2nd on local whithout cities (to speed up test and validation)

kadhikari (approved):

PR #3379, opened 2021-03-04
[docker compose] waiting artifacts to run GA job

benoit-bst:
We want to chain jobs like:
```
build packages -> publish docker compse
```
Because docker compose needs debian packages...
@pbench found a way with https://docs.github.com/en/actions/reference/events-that-trigger-workflows#workflow_run
Dunno if it works. **Merge** it and **revert** If it does not work :duck:


azime (approved):

xlqian (approved):
:100:

PR #3384, opened 2021-03-11
Delete user email in release changelog

kadhikari:
For details: https://jira.kisio.org/browse/NAVITIAII-3341

### DIFF
--- a/release/script_release.py
+++ b/release/script_release.py
@@ -269,12 +269,11 @@ class ReleaseManager:
             write_lines.append(u'  * \n')
 
         author_name = self.git.config('user.name')
-        author_mail = self.git.config('user.email')
         write_lines.extend(
             [
                 u'\n',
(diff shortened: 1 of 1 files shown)


benoit-bst (commented):
I am not sure about that.
I think It will not work, because debian package need an Author. Maybe a name like Navitia and a generic email will be better.

kadhikari:
I don't see any use of email apart writing name + email + ... in a line. I don't understand "debian package need an Author"

"Maybe a name like Navitia and a generic email will be better" -> good Idea

xlqian:
and our organization email will still stand in previous commits...

benoit-bst:
In any case, the **github** **history**  exposes our mail... This is only for the future

kadhikari:
Email address will not be deleted. The user is suggested to use his/her company's email address.

PR #3388, opened 2021-03-17
[ed::builder] Refactorize tests to accomodate future work

nonifier:
Ok, I know this is scary.... BUT

don't be afraid

# What is is ?
The main purpose of this change is to always build object's relationship before we build raptor's internal data.
The relationship that we are interested in is the `route <-> stop point` one, populated from [build_route_and_stop_point_relations](https://github.com/CanalTP/navitia/blob/6cd52a684db252052455de76ec8b5e22656a073f/source/type/data.cpp#L416-L427).  

# Why do we need this ?
Because, route points will be populated according to the `route <-> stop point` relationship within `dataraptor` once at startup, so that  each journey pattern point can be attached to its matching route-point. 
Hence the need to always build [relationship before dataraptor](https://github.com/CanalTP/navitia/compare/dev...nonifier:raptor-route-points?diff=unified&expand=1#diff-f965e99255fc0119c323b338b3eadd33032497f5d44354cabb7061f98d4c6a9dR805-R806).

This is [what happened](https://thecodinglove.com/when-they-ask-me-to-remove-a-function-from-a-legacy-project) when you dare updating `ed::builder`....

## Ed::builder Constructor
Also, this offers a new design when using `ed::builder` to create your data : 

Instead of 
```cpp
        ed::builder b("20170101");
        // Define my Data
        b.sa("stop_area:stop1", 4.2973, 46.945686);
        b.vj("A")("stop_area:stop1", 8 * 3600 + 10 * 60, 8 * 3600 + 11 * 60)

        // I'm done, let's call a handful lot of obscure functions
        b.finish();  // ok seems fair
        b.data->build_uri(); // hummmm not so sure about this one....
        b.data->pt_data->sort_and_index(); //  ¯\_(ツ)_/¯
        b.data->build_raptor(); // sounds fun !
        b.make();  // Isn't this already done ??
```
  
You can simply write :

```cpp
        ed::builder b("20170101", [](ed::builder& b{
            // Define my Data
            b.sa("stop_area:stop1", 4.2973, 46.945686);
            b.vj("A")("stop_area:stop1", 8 * 3600 + 10 * 60, 8 * 3600 + 11 * 60)        
        });

        // You can use your data ! we got you covered ! 
```


source/ed/build_helper.cpp
```diff
@@ -625,6 +628,11 @@ builder::builder(const std::string& date,
     }
 
     tz_handler = data->pt_data->tz_manager.get_or_create(timezone_name, data->meta->production_date.begin(), timezone);
+
+    if (no_dummy == false)
+        generate_dummy_basis();
```
pbench, line 631:
```suggestion
    if (no_dummy == false) {
        generate_dummy_basis();
    }
```

source/type/pt_data.cpp
```diff
@@ -46,6 +46,7 @@ www.navitia.io
 #include "type/physical_mode.h"
 #include "utils/functions.h"
 
+#include <boost/none.hpp>
```
benoit-bst: it is used ?

source/ed/build_helper.cpp
```diff
@@ -31,6 +31,7 @@ www.navitia.io
 #include "build_helper.h"
 
 #include "ed/connectors/gtfs_parser.h"
+#include "utils/map_find.h"
```
benoit-bst: the famous :wink:

benoit-bst (approved):

source/ed/build_helper.cpp
```diff
@@ -820,16 +800,16 @@ void builder::finish() {
 }
 
 void builder::make() {
-    generate_dummy_basis();
```
pbench, line 823: you don't need to generate_dummy_basis() anymore ?

pbench (approved):
A few naive questions.
Very nice refacto :clap:

source/ed/build_helper.cpp
benoit-bst: same same @pbench but risky

nonifier: It's fine by me either way, but like with every (endless and pationate) questions about code formatting, I would rather have the rule defined with our formatting tool.  

One may add that rule to our `clang-tidy` if you really want it  :) 
https://clang.llvm.org/extra/clang-tidy/checks/readability-braces-around-statements.html

source/ed/build_helper.cpp
nonifier: We still do indeed :)  
As a lot of tests need the dummy objects prior to the `make()` call,  I moved the call to `generate_dummy_basis()` [just before the callback](https://github.com/CanalTP/navitia/pull/3388/files#diff-f965e99255fc0119c323b338b3eadd33032497f5d44354cabb7061f98d4c6a9dR632-R633) in ed::builder's constructor.

source/type/pt_data.cpp
nonifier: good catch ! now removed 🙏🏻

source/ed/build_helper.cpp
nonifier: the unfamous `map_find` :D 
It's now been removed because of its uselessness. (god that's a lot of `s` in this word....)

source/ed/build_helper.cpp
pbench: allright :+1:

source/ed/build_helper.cpp
pbench: I do prefer with curly braces but not enough for an endless and passionate debate ;)
I leave the choice to you @nonifier !

A rule in clang-tidy is a good idea !

source/type/pt_data.cpp
```diff
@@ -93,6 +93,22 @@ type::Network* PT_Data::get_or_create_network(const std::string& uri, const std:
     return network;
 }
 
+type::Company* PT_Data::get_or_create_company(const std::string& uri, const std::string& name) {
+    auto company_it = companies_map.find(uri);
+
+    if (company_it != companies_map.end())
+        return company_it->second;
+
+    auto* company = new navitia::type::Company();
```
xlqian, line 102: still raw pointer? (maybe  off the subject)

xlqian (approved):
GG I love those roasting comments

source/type/pt_data.cpp
nonifier: Yes I didn't dare touching this subject (even though I wish I had :D) 
I just followed the actual design :(

PR #3401, opened 2021-03-31
[Raptor] Compute Route Points add associate then to StopPoint/JPP

nonifier:
This is the first real change that will introduce the unfamous `ROUTE POINT` : 
* A very simple data-structure which does the connection between a `Route` and (wait for it....) a `Stop Point` ! : 
```cpp
struct RoutePoint {
    idx_t idx;
    StopPoint* stop_point;
    Route* route;
};
```

The list of `RoutePoint` belongs to `PT_Data` like most the PT objects : 
```
class PT_Data : { 
    ...
    std::vector<RoutePoint> route_points;
}
```

And unlike the rest of the `PT_Data`, the objects are "referenced" not with a raw pointer (I know we all got tired of the raw pointer forest) but with proper c++ reference semantic through [std::reference_wrapper](https://en.cppreference.com/w/cpp/utility/functional/reference_wrapper) 🎉 
```cpp
struct StopPoint :
    [...] 
    boost::container::flat_map<Route*, std::reference_wrapper<RoutePoint>> route_point_list;
```

No call to `new` or `delete` has been harm during the development....

### DIFF
--- a/source/type/pt_data.cpp
+++ b/source/type/pt_data.cpp
@@ -44,6 +44,7 @@ www.navitia.io
 #include "type/multi_polygon_map.h"
 #include "type/commercial_mode.h"
 #include "type/physical_mode.h"
+#include "type/route_point.h"
 #include "utils/functions.h"
 
 #include <boost/range/algorithm/find_if.hpp>
--- a/source/type/route_point.h
+++ b/source/type/route_point.h
@@ -1,4 +1,4 @@
-/* Copyright © 2001-2019, Canal TP and/or its affiliates. All rights reserved.
+/* Copyright © 2001-2021, Canal TP and/or its affiliates. All rights reserved.
 
 This file is part of Navitia,
     the software to build cool stuff with public transport.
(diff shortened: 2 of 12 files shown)


source/type/route_point.h
```diff
@@ -0,0 +1,55 @@
...
+#include <boost/range/any_range.hpp>
+
+#include <utility>
+
+namespace navitia {
+namespace type {
+
+struct RoutePoint {
+    idx_t idx;
+    StopPoint* stop_point;
```
benoit-bst, line 45: :see_no_evil:

source/type/stop_point.h
```diff
@@ -56,6 +57,7 @@ struct StopPoint : public Header, Nameable, hasProperties, HasMessages {
     std::vector<StopPointConnection*> stop_point_connection_list;
     std::set<Dataset*> dataset_list;
     boost::container::flat_set<Route*> route_list;
```
benoit-bst, line 59: does it become useless?
Because of https://github.com/CanalTP/navitia/pull/3401/files#diff-e0441f873b68f4bee05b0865750b7d996e6e4f706b2d9a297dbacc64dfb9a388R60

source/type/pt_data.cpp
```diff
@@ -44,6 +44,7 @@ www.navitia.io
 #include "type/multi_polygon_map.h"
 #include "type/commercial_mode.h"
 #include "type/physical_mode.h"
+#include "type/route_point.h"
```
benoit-bst, line 47: it is alreeady include here ? https://github.com/CanalTP/navitia/pull/3401/files#diff-adeca346ad47a2fe4760401a823171620dcf4bb21b22a2729e0de603845cb230R36

PR #3405, opened 2021-04-06
[Jormungandr] max_waiting added in instance object and /journeys API

azime:
Jira: [NAVITIAII-3363](https://jira.kisio.org/browse/NAVITIAII-3363)

> **Before correction**: Dafault max_waiting: 4h
image

> **After correction**: Override max_waiting: 15minutes
image

### DIFF
--- a/source/jormungandr/jormungandr/interfaces/v1/Journeys.py
+++ b/source/jormungandr/jormungandr/interfaces/v1/Journeys.py
@@ -580,7 +580,9 @@ class Journeys(JourneyCommon):
         if handle_poi_infos(args["add_poi_infos"], args["bss_stands"]):
             self.get_decorators.insert(1, ManageParkingPlaces(self, 'journeys'))
 
-        parser_get.add_argument("max_waiting", type=PositiveInteger(), help='Maximum waiting, must be > 0')
+        parser_get.add_argument(
+            "max_waiting_duration", type=PositiveInteger(), help='Maximum waiting, must be > 0'
+        )
 
     @add_tad_links()
     @add_debug_info()
--- a/source/jormungandr/jormungandr/interfaces/v1/Journeys.py
+++ b/source/jormungandr/jormungandr/interfaces/v1/Journeys.py
@@ -581,7 +581,10 @@ class Journeys(JourneyCommon):
             self.get_decorators.insert(1, ManageParkingPlaces(self, 'journeys'))
 
         parser_get.add_argument(
-            "max_waiting_duration", type=PositiveInteger(), help='Maximum waiting, must be > 0'
+            "max_waiting_duration",
+            type=PositiveInteger(),
+            help='A journey containing a waiting section with a duration greater or equal to  max_waiting_duration '
+            'will be discarded. Units : seconds. Must be > 0. Default value : 4h',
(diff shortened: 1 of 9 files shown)


source/jormungandr/jormungandr/interfaces/v1/Journeys.py
```diff
@@ -580,6 +580,10 @@ def __init__(self):
         if handle_poi_infos(args["add_poi_infos"], args["bss_stands"]):
             self.get_decorators.insert(1, ManageParkingPlaces(self, 'journeys'))
 
+        parser_get.add_argument(
+            "max_waiting_duration", type=PositiveInteger(), help='Maximum waiting, must be > 0'
```
kadhikari: Do we really want to expose this parameter or only for debug purpose? If only for debug,  use '_max_waiting_duration' instead.

azime: No, this is not a debug parameter !

pbench:
```suggestion
            "max_waiting_duration", type=PositiveInteger(), 
            help='A journey containing a waiting section with a duration greater or equal to  max_waiting_duration will be discarded. Units : seconds. Must be > 0. Default value : 4h'
```
Just to be a bit more precise

source/navitiacommon/navitiacommon/models/__init__.py
```diff
@@ -583,6 +583,8 @@ class Instance(db.Model):  # type: ignore
     external_services = db.relationship(
         "ExternalService", secondary=associate_instance_external_service, backref="instances", lazy='joined'
     )
+    # max_waiting default value 4h: 4*60*60 = 14400 minutes
+    max_waiting = db.Column(db.Integer, nullable=False, server_default='14400')
```
pbench: can you use navitiacommon/default_values.py/max_waiting_duration instead of hard-coding  the default value at two different place ?

azime: :+1:

source/jormungandr/jormungandr/interfaces/v1/Journeys.py
azime: :+1:

source/navitiacommon/navitiacommon/models/__init__.py
```diff
@@ -583,6 +583,8 @@ class Instance(db.Model):  # type: ignore
     external_services = db.relationship(
         "ExternalService", secondary=associate_instance_external_service, backref="instances", lazy='joined'
     )
+    # max_waiting_duration default value 4h: 4*60*60 = 14400 minutes
+    max_waiting_duration = db.Column(db.Integer, nullable=False, server_default='14400')
```
kadhikari:
```suggestion
    max_waiting_duration = db.Column(db.Integer, nullable=False, server_default=default_values.max_waiting_duration)
```

azime: :+1:

kadhikari (approved):

pbench (approved):

PR #3406, opened 2021-04-08
[Kraken] fix sort application_patterns of impact

azime:
Jira: [NAVITIAII-3369](https://jira.kisio.org/browse/NAVITIAII-3369)

**Before correction:**
image

**Afetr correction**
image

### DIFF
--- a/source/kraken/tests/apply_disruption_test.cpp
+++ b/source/kraken/tests/apply_disruption_test.cpp
@@ -2543,3 +2543,549 @@ BOOST_AUTO_TEST_CASE(limitation_impact_repeated_stop_points_same_stop_time) {
     BOOST_REQUIRE_EQUAL(vj->stop_time_list.front().stop_point->uri, "stop_point:10");
     BOOST_REQUIRE_EQUAL(vj->stop_time_list.back().stop_point->uri, "stop_point:40");
 }
+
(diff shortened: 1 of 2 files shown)


pbench:
Ok so according to https://en.cppreference.com/w/cpp/language/operators paragraph "Comparison operators", an overload of "<"
is expected to be a "[strict weak ordering](https://en.wikipedia.org/wiki/Weak_ordering)", which means that it should be irreflexive, which means that " a < a " should be false.

Here, your operator< on TimeSlot is not irreflexive. 
Maybe just use what is suggested in cppreference 
`
std::tie(this.begin, this.end) < std::tie(other.begin, other.end);
`

On ApplicationPattern it also seems to me that the operator< is not irreflexive.
If I understand correctly : 
- you first compare the application_pattern.begin
- if they are equal, you compare the application_pattern.end
- if they are equal, you compare the first time slots (why not use the TimeSlot operator< here ?)
If the first time slots are equal, what happens ? Maybe you should keep comparing the subsequent time slots until they are not equal (or you run out of time slot on one of the pattern)

pbench:
And that's a very thorough test coverage ! :+1:

kadhikari (approved):
Thanks for the complete test. :+1:

xlqian:
@pbench you're right, we merge the PR, there will be a follow-up PR to correct.

PR #3407, opened 2021-04-09
[Raptor] JPP based Raptor's Labels

nonifier:
⚠️ This is a work in progress ⚠️  

## What is this about ? 
We are still working on implementing a version of Raptor, where Labels would be computed not just per stop points, but also by taking into account the multiple routes that might pass by that stop point. 

NAVP-1686

## The Route Points Implementation
The first idea was to leverage the use of a `Route-Point` object that would define the association of a Route at a specific Stop point.
[After reviewing this implementation ](https://github.com/CanalTP/navitia/pull/3401), the idea of leveraging the `Journey Pattern Point` instead of the Route Point was raised and seemed to better fit the Raptor's design.

A `JPP` is to a `Journey Pattern` what a `Route Point` is to a `route`. 
This means that we are going to find the best label, for every journey pattern at all stop points. 

Because a `Route` only hold a few `Journey Patterns`, the use of `Jpp` instead of `Route Point` should not dramatically blow up in memory.  

## The JPP implementation
This PR is only a start of the implementation.  It is not finished an still needs some work, It doesn't even compile 👎🏼 . I've only done `raptor.cpp` and `dataRaptor.cpp` to put you on the track.

If you want to continue the work, let the compiler guide you :D 

As a reference, you can have a look at the work done for the `Route Point` [implementation](https://github.com/CanalTP/navitia/compare/dev...nonifier:raptor-route-points-2). It should give you a hint on the different places that need to be updated (but with less changes).

Keep on the good work and as always : 
Good luck

### DIFF
--- a/source/routing/dataraptor.cpp
+++ b/source/routing/dataraptor.cpp
@@ -92,8 +92,8 @@ void dataRAPTOR::JppsFromJp::load(const JourneyPatternContainer& jp_container) {
 
 void dataRAPTOR::load(const type::PT_Data& pt_data, size_t cache_size) {
     jp_container.load(pt_data);
-    labels_const.init_inf(pt_data.stop_points);
-    labels_const_reverse.init_min(pt_data.stop_points);
+    labels_const.init_inf(jp_container.get_jpps_values());
+    labels_const_reverse.init_min(jp_container.get_jpps_values());
 
     connections.load(pt_data);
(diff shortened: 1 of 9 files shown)


pbench:
Close and reopen in #3413

pbench:
Close and reopen in #3413

PR #3408, opened 2021-04-13
add addess node into stop_points for any depth

AzHicham:
Address node are always added into stop_points
Test incoming ...

Related to JIRA ticket : : https://jira.kisio.org/browse/NAVITIAII-3377

### DIFF
--- a/source/type/pb_converter.cpp
+++ b/source/type/pb_converter.cpp
@@ -618,9 +618,7 @@ void PbCreator::Filler::fill_pb_object(const nt::StopPoint* sp, pbnavitia::StopP
         fill(sp->admin_list, stop_point->mutable_administrative_regions());
     }
 
-    if (depth > 2) {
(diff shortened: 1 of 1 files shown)


xlqian (commented):
any tests? :)

benoit-bst:
Maybe a test inside https://github.com/CanalTP/navitia/blob/dev/source/jormungandr/tests/journey_common_tests.py ?
This is _high level integration tests_.

You have to familiarize yourself with _fake kraken_ coverage like -> https://github.com/CanalTP/navitia/blob/dev/source/jormungandr/tests/journey_common_tests.py#L58
For this example, the source of the _coverage_ is -> https://github.com/CanalTP/navitia/blob/dev/source/tests/mock-kraken/main_routing_test.cpp

It allows us to simulate **jormun + kraken** effects. A very powerfull non-regression test :cat:

PR #3410, opened 2021-04-13
fix unit test with boost version > 1.66

AzHicham:
Exception thrown by boost gregorian calendar was changed in boost 1.67


xlqian (approved):

pbench (approved):
I never knew we were testing boost error messages !

benoit-bst:
👆The push force is automatic (it is not me 😀).
Mergify apply a rule to rebase and it takes the first github name of the team list... benoit-bst

PR #3411, opened 2021-04-14
Send slack message only for failures on push event and not for pull request

pbench:


.github/workflows/build_navitia_packages_for_dev_multi_distribution.yml
```diff
@@ -38,7 +38,7 @@ jobs:
     - name: remove useless temporary files
       run: rm -rf ../navitia-*
     - name: slack notification (the job has failed)
-      if: failure()
+      if: ${{ failure() && github.event_name == 'push' }}
```
benoit-bst: I use this kind of syntax : https://github.com/CanalTP/navitia/blob/dev/.github/workflows/build_navitia_packages_for_release.yml#L57

pbench: Ok, that's more readable :+1:

benoit-bst (approved):

azime (approved):

PR #3418, opened 2021-04-21
[Jormungandr] Not return occupancy if undefined

azime:
Jira: [NAVITIAII-3387](https://jira.kisio.org/browse/NAVITIAII-3387)


source/jormungandr/jormungandr/schedule.py
```diff
@@ -379,7 +379,9 @@ def _manage_occupancies(self, schedules):
                 # TODO: this call may be parallelized with gevent
                 # Not done now as it's only a POC and our WS Forseti/vehicle_occupancies will only have fewer
                 # elements (less than 5000 for line 40 and 45). The Cache with 1 hour duration works well
-                date_time.occupancy = vo_service.get_response(args)
+                occupancy = vo_service.get_response(args)
+                if occupancy:
```
xlqian, line 383: Is it possible that occupancy turns out to be 0?
```suggestion
                if occupancy is not None:
```

xlqian (approved):

pbench (approved):

PR #3425, opened 2021-04-27
[Kraken] Modification in sort of application_patterns

azime:
Jira: [NAVITIAII-3378](https://jira.kisio.org/browse/NAVITIAII-3378)


source/type/message.cpp
```diff
@@ -344,28 +344,42 @@ bool Impact::is_line_section_of(const Line& line) const {
...
+        const auto other_ts = *next(other.time_slots.begin(), n);
+        if (!(this_ts == other_ts)) {
+            return this_ts < other_ts;
         }
-        return this->time_slots.begin()->begin < other.time_slots.begin()->begin;
     }
-    if (this->application_period.begin() == other.application_period.begin()) {
-        return this->application_period.end() <= other.application_period.end();
+    if (this->time_slots.size() != other.time_slots.size()) {
+        if (this->time_slots.empty() || other.time_slots.empty()) {
```
xlqian: can you just do 
```return this->time_slots.size() < other.time_slots.size() ``` ?

File: source/type/message.cpp [hunk 2]
```diff
@@ -344,28 +344,42 @@ bool Impact::is_line_section_of(const Line& line) const {
...
+        const auto this_ts = *next(this->time_slots.begin(), n);
+        const auto other_ts = *next(other.time_slots.begin(), n);
+        if (!(this_ts == other_ts)) {
+            return this_ts < other_ts;
         }
-        return this->time_slots.begin()->begin < other.time_slots.begin()->begin;
     }
-    if (this->application_period.begin() == other.application_period.begin()) {
-        return this->application_period.end() <= other.application_period.end();
+    if (this->time_slots.size() != other.time_slots.size()) {
```
xlqian, line 375: shouldn't this test be done before the [loop](https://github.com/CanalTP/navitia/pull/3425/files#diff-347d523f6725669454dd92c1db510fcbd5f8326b7dd2830c779b1629ca3fd18dR368)  ????

source/type/message.cpp
```diff
@@ -344,28 +344,42 @@ bool Impact::is_line_section_of(const Line& line) const {
...
+        if (this->time_slots.empty() || other.time_slots.empty()) {
+            return this->time_slots.empty() ? false : true;
+        }
+        return this->time_slots.size() > other.time_slots.size() ? false : true;
```
pbench:
```suggestion
    if (this->time_slots.size() != other.time_slots.size()) {
        return this->time_slots.size() < other.time_slots.size();
```
should do the same, no ?

source/type/message.cpp
azime: [Response](https://github.com/CanalTP/navitia/pull/3425/files#diff-62d4cf1b2a67eeff3a23408773de54bdc30ff5c4d3a9c137e325a0ea7c23c1fdR3242)!

source/type/message.cpp
azime: OK

source/type/message.cpp
azime: OK

xlqian (approved):

pbench (approved):

PR #3431, opened 2021-05-03
store address in StopPoint, and set these address during Data::complete()

pbench:
and use the stored addresses while filling protobuf responses

### DIFF
--- a/source/ed/CMakeLists.txt
+++ b/source/ed/CMakeLists.txt
@@ -5,7 +5,7 @@ SET(SOURCE_LIB
 )
 
 add_library(ed ${SOURCE_LIB})
-target_link_libraries(ed types fare)
+target_link_libraries(ed types fare data)
 
 FIND_LIBRARY(OSMPBF osmpbf)
 
--- a/source/type/pb_converter.cpp
+++ b/source/type/pb_converter.cpp
@@ -618,7 +618,8 @@ void PbCreator::Filler::fill_pb_object(const nt::StopPoint* sp, pbnavitia::StopP
         fill(sp->admin_list, stop_point->mutable_administrative_regions());
     }
 
-    fill(&sp->coord, stop_point);
+    *(stop_point->mutable_address()) = sp->address;
+    // fill(&sp->coord, stop_point);
 
     if (depth > 0) {
         fill(sp->stop_area, stop_point);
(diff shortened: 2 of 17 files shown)


source/type/pt_data.cpp
```diff
@@ -491,6 +492,14 @@ const StopPointConnection* PT_Data::get_stop_point_connection(const StopPoint& f
     return *search;
 }
 
+void PT_Data::fill_stop_point_address(PbCreator* pb_creator) {
+    for (StopPoint* stop_point : stop_points) {
+        const nt::GeographicalCoord* coord = &(stop_point->coord);
+        pbnavitia::Address* address = &(stop_point->address);
+        pb_creator->fill(coord, address, 3);
```
pbench, line 499: is depth=3 right ?

source/ed/CMakeLists.txt
```diff
@@ -5,7 +5,7 @@ SET(SOURCE_LIB
 )
 
 add_library(ed ${SOURCE_LIB})
-target_link_libraries(ed types fare)
+target_link_libraries(ed types fare data)
```
xlqian, line 8: why `ed` is also involved?

source/type/pb_converter.cpp
```diff
@@ -618,9 +618,8 @@ void PbCreator::Filler::fill_pb_object(const nt::StopPoint* sp, pbnavitia::StopP
         fill(sp->admin_list, stop_point->mutable_administrative_regions());
     }
 
-    if (depth > 2) {
-        fill(&sp->coord, stop_point);
-    }
+    *(stop_point->mutable_address()) = sp->address;
```
xlqian, line 621: fill(stop_point->mutable_address(), stop_point)

source/type/stop_point.h
```diff
@@ -56,11 +57,12 @@ struct StopPoint : public Header, Nameable, hasProperties, HasMessages {
     std::vector<StopPointConnection*> stop_point_connection_list;
     std::set<Dataset*> dataset_list;
     boost::container::flat_set<Route*> route_list;
+    pbnavitia::Address address;
```
xlqian, line 60: as @azime pointed out, in principle, we should create a new struct for address and fill the `pbnavitia:Address` in pb_converter... just like `type::Admin`

source/ed/CMakeLists.txt
pbench: what happens is that I store the protobuf Address in the StopPoint object (maybe not the best thing to do), consequently `types` depends on the protobuf files, and this impact dependencies everywhere :crying_cat_face:

xlqian (commented):
:)

PR #3432, opened 2021-05-04
[Kraken] Decide which type of next stoptime to use in runtime

xlqian:
Reuse the uncached next_stop_time when the request's `datetime` is not contained in the time window( [Today,  Today + N]).
 
(no cache)         (cache)         ( no cache)
----------[Fixed Time Window ]------------

The performance drop-off in production MAY be due to the frequent invalidation of next_stop_time. In fact, this actually unveiled the shortcomings of the LRU cache, which is, when if the cache hit is frequently missed, instead of improving the performance, we lost much more time in rebuilding the cache in the place of reusing pre-computed data...

The solution is to fix a date window, let's say [N, N+5]. Any request whose datetime is out of the window, we'll use the uncached next_stop_time. Eventually, we will observe a performance dropoff, but only for a small portion of requests.  The advantage of the solution is that cache invalidation is largely reduced.

### DIFF
--- a/source/routing/next_stop_time.cpp
+++ b/source/routing/next_stop_time.cpp
@@ -466,13 +466,13 @@ boost::iterator_range<CachedNextStopTime::vDtSt::const_iterator> CachedNextStopT
 }
 
 std::pair<const type::StopTime*, DateTime> CachedNextStopTime::next_stop_time(const StopEvent stop_event,
-        const JppIdx jpp_idx,
-        const DateTime dt,
-        const bool clockwise,
-        const type::RTLevel ,
-        const type::VehicleProperties& ,
-        const bool check_freq ,
-        const boost::optional<DateTime>&) const {
+                                                                              const JppIdx jpp_idx,
+                                                                              const DateTime dt,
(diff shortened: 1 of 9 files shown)


source/routing/next_stop_time.cpp
```diff
@@ -636,7 +640,7 @@ DateTime get_previous_stop_time(const StopEvent stop_event,
     // overnight case and hour > midnight
     if (hour > upper_bound) {
         // case with hour in [upper_bound, lower_boud]
-        if (freq_vj.is_valid(date, rt_level)) {
+        if (freq_vj.is_valid(date - 1, rt_level)) {
```
pbench, line 643: why do you need that with this PR ?

pbench (approved):

source/routing/next_stop_time.cpp
xlqian: This code which has been deactivated was not thoroughly tested... when I brought this code back, a [test](https://github.com/CanalTP/navitia/blob/dev/source/routing/tests/frequency_raptor_test.cpp#L517) has failed...

PR #3435, opened 2021-05-06
[Kraken] replace size_t by idx_t

azime:
Correction for the [remark](https://github.com/CanalTP/navitia/pull/3434#discussion_r627133222)

### DIFF
--- a/source/type/data.cpp
+++ b/source/type/data.cpp
@@ -490,7 +490,7 @@ void Data::aggregate_odt() {
 }
 
 void Data::fill_stop_point_address() {
-    size_t without_address = 0;
+    idx_t without_address = 0;
     for (auto sp : pt_data->stop_points) {
         if (!sp->coord.is_initialized()) {
             ++without_address;


kadhikari (approved):

benoit-bst (approved):

xlqian (approved):

AzHicham:
std::size_t is already unsigned I think. (could be u32 or u64 depending on the platform)
https://en.cppreference.com/w/cpp/types/size_t

xlqian:
@AzHicham `navitia::type::idx_t` is just an alias uint32_t. IMHO, using explicit unsigned types can avoid different behaviors on different platforms. (I've already experienced this due to the implicitness of `size_t` depending on the platform. :smile: It took me two days to figure out what was happening... )

AzHicham:
It's clear ! 
It had the same issue once when I used boost.serialization and I was streaming data between two computer (one was 32b and the other was 64b) ... I let you guess what happened

PR #3413, opened 2021-04-14
[Raptor] JPP based Raptor's Labels

pbench:
Reopening  https://github.com/CanalTP/navitia/pull/3407 in my fork.

Ticket [NAVP-1686](https://jira.kisio.org/browse/NAVP-1686)

PR #3442, opened 2021-05-18
Route points again, with 4mn walking penalty

pbench:
Follow up on https://github.com/CanalTP/navitia/pull/3413
We now penalize the walking duration with the waiting duration : i.e when comparing two journeys, we compare the quantity : 
` 2 * walking_duration + waiting_duration`
So that waiting for 1 more minute means we should reduce the walking duration by at least 30s

Todo : 

- [ ] validate the changes in artemis tests

- [ ] update artemis references

- [ ] update doc in confluence with the new formula


pbench:
Closing, performance impact too high

PR #3443, opened 2021-05-18
[Jormungandr] Manage "is_realtime" in the Timeo response

azime:
Jira: [NAVP-1827](https://jira.kisio.org/browse/NAVP-1827)

### DIFF
--- a/source/jormungandr/jormungandr/realtime_schedule/timeo.py
+++ b/source/jormungandr/jormungandr/realtime_schedule/timeo.py
@@ -231,10 +231,12 @@ class Timeo(RealtimeProxy):
             )
             raise RealtimeProxyError('invalid response')
 
-        next_st = st_responses[0]['NextStopTimesMessage']
-
+        next_st = st_responses[0]['NextStopTimesMessage'].get("NextExpectedStopTime", [])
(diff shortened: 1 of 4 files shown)


source/jormungandr/jormungandr/realtime_schedule/timeo.py
```diff
@@ -231,10 +231,12 @@ def _get_passages(self, response, current_dt, line_uri=None):
             )
             raise RealtimeProxyError('invalid response')
 
-        next_st = st_responses[0]['NextStopTimesMessage']
-
+        next_st = st_responses[0]['NextStopTimesMessage'].get("NextExpectedStopTime", [])
+        new_next_st = [n_st for n_st in next_st if n_st.get("is_realtime", True)]
+        if not len(new_next_st) and len(next_st):
+            return None
```
pbench, line 237: You slightly change what is returned by the function when `next_st` is empty, before the function would return `[]` but now it will return `None`. 
Is that Ok ?

pbench (approved):

source/jormungandr/jormungandr/realtime_schedule/timeo.py
azime: Yes !

kadhikari (approved):

PR #3456, opened 2021-05-31
[Jormungandr] Manage realtime proxies from database

azime:
Jira: [NAVP-1831](https://jira.kisio.org/browse/NAVP-1831)

### DIFF
--- a/source/jormungandr/jormungandr/instance.py
+++ b/source/jormungandr/jormungandr/instance.py
@@ -161,8 +161,9 @@ class Instance(object):
         self.ptref = ptref.PtRef(self)
 
         self.schedule = schedule.MixedSchedule(self)
+        providers_getter = None if disable_database else self.get_realtime_proxies_from_db
         self.realtime_proxy_manager = realtime_schedule.RealtimeProxyManager(
-            realtime_proxies_configuration, self
+            realtime_proxies_configuration, self, providers_getter
         )
 
         self._autocomplete_type = autocomplete_type
@@ -213,7 +214,15 @@ class Instance(object):
         """
         :return: a callable query of external services associated to the current instance in db
         """
-        return self._get_models().external_services
+        result = self._get_models().external_services
+        return [res for res in result if res.navitia_service != 'realtime_proxies']
+
+    def get_realtime_proxies_from_db(self):
+        """
+        :return: a callable query of external services associated to the current instance in db
+        """
+        result = self._get_models().external_services
+        return [res for res in result if res.navitia_service == 'realtime_proxies']
 
     @property
     def autocomplete(self):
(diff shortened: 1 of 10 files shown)


source/jormungandr/jormungandr/instance.py
```diff
@@ -213,7 +214,15 @@ def get_external_service_providers_from_db(self):
         """
         :return: a callable query of external services associated to the current instance in db
         """
-        return self._get_models().external_services
+        result = self._get_models().external_services
+        return [res for res in result if res.navitia_service != 'realtime_proxies']
```
kadhikari: It seems fine now but when we add a new service, we will have to modify here and below. What I propose:
return [res for res in result if res.navitia_service in ['free_floatings', 'vehicle_occupancies']]

kadhikari (commented):
It seems fine to me. Could you test in local the followings:

- Only config in file
- One config in file and another in db but different id
- Same config in the file and also in db with one parameter value modified and verify that the new modified value in db is used.
- Config only in db

source/jormungandr/jormungandr/instance.py
azime: OK

kadhikari (approved):

source/jormungandr/jormungandr/realtime_schedule/realtime_proxy_manager.py
```diff
@@ -31,57 +31,136 @@
...
+        try:
+            object_id_tag = rt_proxy.args.get('object_id_tag', rt_proxy.id)
+            if 'object_id_tag' in rt_proxy.args:
+                del rt_proxy.args["object_id_tag"]
+            self._realtime_proxies[rt_proxy.id] = self._init_class(
+                rt_proxy.id, rt_proxy.klass, object_id_tag, rt_proxy.args
+            )
+            self._realtime_proxies_last_update[rt_proxy.id] = rt_proxy.last_update()
+        except Exception:
+            self.logger.exception('impossible to initialize realtime proxy')
```
pbench:
```suggestion
            self.logger.exception('impossible to update realtime proxy {}'.format(rt_proxy.id))
```

source/jormungandr/jormungandr/realtime_schedule/realtime_proxy_manager.py
```diff
@@ -31,57 +31,136 @@
...
+
+        # If the realtime proxy added in db is also defined in legacy, delete it.
```
pbench:
```suggestion
        # If a legacy realtime proxy exists in db, remove it from self._realtime_proxies_legacy
```

source/jormungandr/jormungandr/realtime_schedule/realtime_proxy_manager.py
azime: OK

source/jormungandr/jormungandr/realtime_schedule/realtime_proxy_manager.py
azime: ok

pbench (approved):

PR #3472, opened 2021-06-14
Filter disruption on current status 

AzHicham:
The status of a disruption could be past/active/future. 
Thank to this PR, an user could retrieve only a limited set of disruption based on the status.

### DIFF
--- a/source/type/message.cpp
+++ b/source/type/message.cpp
@@ -429,7 +429,7 @@ bool Impact::is_valid(const boost::posix_time::ptime& publication_date,
         return false;
     }
 
-    // if we have a active_period, we check if the impact applies on this period
+    // if we have a active_period we check if the impact applies on this period
     if (active_period.is_null()) {
         return true;
     }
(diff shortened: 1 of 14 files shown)


source/type/message.cpp
```diff
@@ -442,6 +442,23 @@ bool Impact::is_valid(const boost::posix_time::ptime& publication_date,
...
 }
 
+ActiveStatus Impact::get_active_status(const boost::posix_time::ptime& publication_date) const {
+    bool is_future = false;
+    for (const auto& period : application_periods) {
+        if (period.contains(publication_date)) {
+            return ActiveStatus::active;
+        }
+        if (!period.is_null() && period.begin() >= publication_date) {
+            is_future = true;
```
azime, line 452:
```suggestion
           return ActiveStatus::future;
```
No?

AzHicham: We loop over all application_periods to check if there is one currently active, I don't think application_periods are ordered ?

benoit-bst (approved):

source/type/message.cpp
azime: OK

azime (approved):

PR #3478, opened 2021-06-21
remove duplicated journeys because of bss

xlqian:
https://jira.kisio.org/browse/NAVITIAII-3447

In this PR, I replaced bss fallback by a walking fallback if the bss fallback is actually a pure walking fallback. At the end of distributed_scenario, I added another filter to remove the duplicated journeys

### DIFF
--- a/source/jormungandr/jormungandr/scenarios/helper_classes/helper_utils.py
+++ b/source/jormungandr/jormungandr/scenarios/helper_classes/helper_utils.py
@@ -388,15 +388,11 @@ def _build_crowfly(pt_journey, entry_point, mode, places_free_access, fallback_d
 
 
 def _is_pure_walking(dp):
-    try:
-        if len(dp.journeys[0].sections) != 1:
-            return False
+    if dp.journyes and len(dp.journeys[0].sections) == 1:
         section = dp.journeys[0].sections[0]
         if section.type == response_pb2.STREET_NETWORK and section.street_network.mode == response_pb2.Walking:
             return True
-        return False
-    except IndexError:
-        return False
+    return False
 
 
 def _build_fallback(
--- a/source/jormungandr/jormungandr/scenarios/helper_classes/helper_utils.py
+++ b/source/jormungandr/jormungandr/scenarios/helper_classes/helper_utils.py
@@ -40,6 +40,7 @@ from jormungandr.utils import (
 from jormungandr.street_network.utils import crowfly_distance_between
 from jormungandr.fallback_modes import FallbackModes, all_fallback_modes
 from .helper_exceptions import *
+from navitiacommon import response_pb2
 import copy
 import logging
 import six
@@ -386,6 +387,18 @@ def _build_crowfly(pt_journey, entry_point, mode, places_free_access, fallback_d
     return _create_crowfly(pt_journey, crowfly_origin, crowfly_destination, begin, end, mode)
 
 
+def _is_pure_walking(dp):
+    try:
+        if len(dp.journeys[0].sections) != 1:
+            return False
+        section = dp.journeys[0].sections[0]
+        if section.type == response_pb2.STREET_NETWORK and section.street_network.mode == response_pb2.Walking:
+            return True
+        return False
+    except IndexError:
+        return False
+
+
 def _build_fallback(
     requested_obj,
     pt_journey,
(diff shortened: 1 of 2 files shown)


source/jormungandr/jormungandr/scenarios/helper_classes/helper_utils.py
```diff
@@ -385,6 +386,18 @@ def _build_crowfly(pt_journey, entry_point, mode, places_free_access, fallback_d
...
+def _is_pure_walking(dp):
+    try:
+        if len(dp.journeys[0].sections) != 1:
+            return False
+        section = dp.journeys[0].sections[0]
+        if section.type == response_pb2.STREET_NETWORK and section.street_network.mode == response_pb2.Walking:
+            return True
+        return False
+    except IndexError:
+        return False
```
pbench:
```suggestion
   if len(dp.journeys > 0):
      if len(dp.journeys[0].sections) == 1:
         section =  dp.journeys[0].sections[0]
         if section.type == response_pb2.STREET_NETWORK and section.street_network.mode == response_pb2.Walking :
                        return True
   return False                        
```

source/jormungandr/jormungandr/scenarios/journey_filter.py
```diff
@@ -647,6 +647,20 @@ def apply_final_journey_filters(response_list, instance, request):
     _filter_too_much_connections(journeys, instance, request)
 
 
+def replace_bss_tag(journeys):
+    """
+    replace the gss tag by walking, if there's no bss section in the journey
```
pbench:
```suggestion
    replace the `bss` tag by `walking`, if there's no bss section in the journey
```

source/jormungandr/jormungandr/scenarios/journey_filter.py
```diff
@@ -647,6 +647,20 @@ def apply_final_journey_filters(response_list, instance, request):
...
+    """
+    for j in journeys:
+        if not j.tags or "bss" not in j.tags:
+            continue
+        has_bss = any((s.type == response_pb2.BSS_RENT for s in j.sections))
+        if has_bss:
+            continue
+        j.tags.remove("bss")
+        j.tags.append("walking")
```
pbench, line 648: Can it happen that there is no walking at all (e.g. the starting point is a stop point) and thus the "walking" tag should not be appended ?

xlqian: the goal is to replace the bss tag by walking ONLY when `bss` tag is existent in `journeys.tags`. In fact, with the presence of `bss` tag, we're sure that there must be a fallback return by the `bss` connector but we're not sure if it's pure walking or bss...

pbench: Ok, I guess the case I am thinking about is the following : you give a stop_point as the starting point and ask for a `bss` fallback.
In this case, I expect the `bss` fallback to be empty (maybe I am wrong here), but it should be tagged as `bss`. 
So here, you remove the `bss` tag and add the `walking` tag. But the fallback has no walking at all.

kadhikari (commented):
Is it difficult to find and add a test for this correction ?

source/jormungandr/jormungandr/scenarios/journey_filter.py
xlqian: this case is protected by[ this line](https://github.com/CanalTP/navitia/pull/3478/files/14f6a6e217d8ab86ee93ddf80bbb623030047a99#diff-51842a7f082fe7a1d3dfc36f56a6370e922036194f379d51f0c92c497dc55a69R655)

source/jormungandr/jormungandr/scenarios/helper_classes/helper_utils.py
```diff
@@ -385,6 +386,14 @@ def _build_crowfly(pt_journey, entry_point, mode, places_free_access, fallback_d
     return _create_crowfly(pt_journey, crowfly_origin, crowfly_destination, begin, end, mode)
 
 
+def _is_pure_walking(dp):
+    if dp.journyes and len(dp.journeys[0].sections) == 1:
```
pbench:
```suggestion
    if dp.journeys and len(dp.journeys[0].sections) == 1:
```

pbench (approved):
Looks good to me :+1:

xlqian:
@kadhikari hard to place a tests :( I tried...

kadhikari (approved):

PR #3480, opened 2021-06-23
Fix bug bypass_disruption link

AzHicham:
This PR fix the bypass_disruption link : 
Previously the bypass_disruption link was the last link to be filled, therefore added parameters by prev/next link were present in bypass_disruption link. 
Now : The first link to be filled is bypass_disruption link, but we use a copy of dict parameters, so data_freshness=realtime will not be present in prev/next link


kadhikari (approved):

pbench (commented):
What was the bug and how does this change correct it ?

AzHicham:
> What was the bug and how does this change correct it ?

Please see edited comment of this PR :)

benoit-bst (approved):

pbench (approved):

PR #3481, opened 2021-06-23
Remove terminus_schedules without stop_times

azime:
Jira: [NAVP-1659](https://jira.kisio.org/browse/NAVP-1659)

Example:
**The parameters used**
image

**Before correction:**
image

**After correction:**
image

Bench:
**Command used**: 
`ab -n 1000 http://navitia/v1/coverage/fr-se-lyon/stop_areas/stop_area%3Atcl%3ASA%3A11273/lines/line%3Atcl%3AC8/terminus_schedules?data_freshness=base_schedule&`

**Before correction:**
image

**After correction:**
image

### DIFF
--- a/source/time_tables/departure_boards.cpp
+++ b/source/time_tables/departure_boards.cpp
@@ -473,10 +473,13 @@ void terminus_schedules(PbCreator& pb_creator,
         // check whether that calendar exists, to raise an early error
         if (pb_creator.data->pt_data->calendars_map.find(*calendar_id)
             == pb_creator.data->pt_data->calendars_map.end()) {
-            pb_creator.fill_pb_error(pbnavitia::Error::bad_filter, "stop_schedules : calendar does not exist");
+            pb_creator.fill_pb_error(pbnavitia::Error::bad_filter, "terminus_schedules : calendar does not exist");
             return;
         }
     }
+    // <route, <route_point>>
+    std::map<const type::Route*, vector_jpp_idx> map_route_route_point;
+
     //  <stop_point_route, status>
     std::map<routing::JppIdx, pbnavitia::ResponseStatus> response_status;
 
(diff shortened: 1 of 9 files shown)


source/time_tables/departure_boards.cpp
```diff
@@ -621,14 +624,51 @@ void terminus_schedules(PbCreator& pb_creator,
...
 
+void clean_terminus_schedules(const std::map<const type::Route*, vector_jpp_idx>& map_route_route_point,
+                              std::map<routing::JppIdx, pbnavitia::ResponseStatus>& response_status,
+                              std::map<routing::JppIdx, vector_dt_st>& map_route_stop_point) {
+    for (const auto& rt_rp : map_route_route_point) {
+        bool with_stop_times = false;
+        if (rt_rp.second.size() > 1) {
+            for (const auto& route_point : rt_rp.second) {
+                const auto& it = map_route_stop_point.find(route_point);
+                if (!it->second.empty()) {
```
pbench, line 650: Will crash if `it == map_route_stop_point.end()`. Are you sure it cannot happen ?

source/time_tables/departure_boards.cpp
```diff
@@ -621,14 +624,51 @@ void terminus_schedules(PbCreator& pb_creator,
...
+            for (const auto& route_point : rt_rp.second) {
+                const auto& it = map_route_stop_point.find(route_point);
+                if (it->second.empty()) {
+                    if (with_stop_times) {
+                        map_route_stop_point.erase(it);
+                        const auto& st_it = response_status.find(route_point);
+                        response_status.erase(st_it);
+                    } else {
+                        with_stop_times = true;
+                    }
```
pbench: The loop above set `with_stop_times` to `true` if the route has at least one `route_point/jpp` with one `stop_time`.
If such a route is found, then, in this loop you erase the **all** `route_point/jpp` of the route that have no stop_time. 

If the route has no route_point with a stop_time, then in this loop you erase **all** `route_point/jpp` of the route, except the first one. 
Why do you keep the first one ?

Why not just remove all route_points without  stop_times ?

pbench (approved):

benoit-bst (approved):
You accelerated da game :crossed_swords:

PR #3486, opened 2021-06-29
[Jormungandr] read params from database

azime:


kadhikari:
If i understand, for those connectors, we use max_fail and reset_timeout value from database if exist. Actions to fetch these values from database and create arguments (**args) must have been done already non?

azime:
Yes, these params (circuit_breaker_max_fail, circuit_breaker_reset_timeout) are in navitia-configurator

benoit-bst (approved):

pbench (approved):

PR #3490, opened 2021-07-13
[Kraken] move deadline into dispatch to avoid crash

xlqian:
move deadline into dispatch to avoid a crash

Kraken crashed because of the timeout(deadline is reached) when a metadata request is received. One of the pointers in `Worker` which is invalidated by the real-time handling remains invalid, which caused a crash in the handling of metrics. 

We moved deadline into the function `dispatch` so that, whether deadline is hit or not, the re-initialisation is always done

### DIFF
--- a/source/kraken/data_manager.h
+++ b/source/kraken/data_manager.h
@@ -41,6 +41,7 @@ www.navitia.io
 #include <boost/make_shared.hpp>
 #include <boost/optional.hpp>
 
+#include <mutex>
 #include <memory>
 #include <iostream>
 #include <atomic>
@@ -76,6 +77,8 @@ private:
         }
     }
 
+    std::unique_ptr<std::mutex> mutex = std::make_unique<std::mutex>();
+
 public:
     DataManager() : current_data(create_data(0)) { data_identifier = 0; }
 
@@ -87,7 +90,10 @@ public:
         data->is_connected_to_rabbitmq = current_data->is_connected_to_rabbitmq.load();
         current_data = std::move(data);
     }
-    boost::shared_ptr<const Data> get_data() const { return current_data; }
+    boost::shared_ptr<const Data> get_data() const {
+        std::lock_guard<std::mutex> lock(*mutex);
+        return current_data;
+    }
     boost::shared_ptr<Data> get_data_clone() {
         ++data_identifier;
         auto data = create_data(data_identifier.load());
--- a/source/kraken/data_manager.h
+++ b/source/kraken/data_manager.h
@@ -88,6 +88,8 @@ public:
             throw navitia::exception("Giving a null Data to DataManager::set_data");
         }
         data->is_connected_to_rabbitmq = current_data->is_connected_to_rabbitmq.load();
+
+        std::lock_guard<std::mutex> lock(*mutex);
         current_data = std::move(data);
     }
     boost::shared_ptr<const Data> get_data() const {
--- a/source/kraken/data_manager.h
+++ b/source/kraken/data_manager.h
@@ -40,6 +40,7 @@ www.navitia.io
 
 #include <boost/make_shared.hpp>
 #include <boost/optional.hpp>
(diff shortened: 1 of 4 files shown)


source/kraken/data_manager.h
```diff
@@ -87,7 +90,10 @@ class DataManager {
         data->is_connected_to_rabbitmq = current_data->is_connected_to_rabbitmq.load();
         current_data = std::move(data);
     }
-    boost::shared_ptr<const Data> get_data() const { return current_data; }
+    boost::shared_ptr<const Data> get_data() const {
+        std::lock_guard<std::mutex> lock(*mutex);
```
xlqian: this line is NOT the cause of the crash. I added this because we have a doubt over the race condition this line may introduce...

source/kraken/data_manager.h
```diff
@@ -76,6 +77,8 @@ class DataManager {
         }
     }
 
+    std::unique_ptr<std::mutex> mutex = std::make_unique<std::mutex>();
```
AzHicham: why use a unique_ptr<mutex> instead of mutex directly ?

source/kraken/kraken_zmq.h
```diff
@@ -128,8 +128,7 @@ inline void doWork(zmq::context_t& context,
         LOG4CPLUS_DEBUG(logger, "deadline set to " << deadline.get());
         const auto data = data_manager.get_data();
         try {
-            deadline.check();
```
xlqian, line 131: voilà, the mischievous line...

source/kraken/data_manager.h
pbench: ~~Don't you effectively disable concurrent request handling with this ?~~ 
~~If data is behind a mutex, two threads cannot read data at the same time, right ?~~
Edit : Ok, I got it now :ok_hand:

benoit-bst: I think you have right.
It can be a case, the access between get_data() and set_data(). 
We have 2 threads for example: The first do `set_data()`, and during the std::move, the second call get_data()... 
It can be unlikely but it is real !

AzHicham: Here we could use boost::shared_mutex (std::shared_mutex since c++17) with : 
```
std::lock_guard<boost::shared_mutex> lock(write); in set_data
std::shared_lock<boost::shared_mutex> lock(write);    in get_data 
```
For multiple reader/single writer configuration

pbench:
> Kraken crashed because of the timeout(deadline is reached) when a metadata request is received. One of the pointers in `Worker` which is invalidated by the real-time handling remains invalid, which caused a crash in the handling of metrics.

Is this another option to fix the bug ? : don't send metrics when the deadline is reached

source/kraken/data_manager.h
pbench: ~~another silly question (I am not that familiar with lock/mutex) : isn't you mutex released at the end of the `get_data` function (i.e. immediately) ?~~
Ok, it protects the pointer to data and not the data itself , thanks @AzHicham

AzHicham: The mutex will be released after making a copy of boost::shared_ptr<const Data>, then the copy will be returned

benoit-bst:
> When you talk about metrics, you talk about what ? The Prometheus handler ?

I have the response

source/kraken/data_manager.h
xlqian: because `get_data` is const, I cannot lock a non-mutable mutex in `get_data`...

xlqian: otherwise, I must declare mutex as a mutable member which is disturbing for me

benoit-bst: ho noooo, the mutable is a weird by pass :flying_saucer: .
Otherwise, You can erase the const ?

benoit-bst (approved):

pbench (approved):

PR #3498, opened 2021-07-21
Safeguard code while accessing elements in the vector

kadhikari:
Since all the cases are already tested by @benoit-bst i haven't added any test in this PR.

Jira: https://jira.kisio.org/browse/NAVITIAII-3522


azime:
can you add a test?

kadhikari:
> can you add a test?

It will be very difficult to find a test.

benoit-bst (approved):

azime (approved):

PR #3530, opened 2021-08-24
[Tyr] Rename dataset file name if paranthesis is present

kadhikari:
If the filename contains characters as (, ) .. rename file replacing characters from this character by a datetime formatted string.
Example: 'eu_euskadi_ntfs_final(1).zip' becomes 'eu_euskadi_ntfs_final_20210819-220730832040.zip'

Jira: https://jira.kisio.org/browse/NAVP-1777

### DIFF
--- a/source/tyr/tyr/binarisation.py
+++ b/source/tyr/tyr/binarisation.py
@@ -71,17 +71,42 @@ def unzip_if_needed(filename):
     return working_directory
 
 
-def move_to_backupdirectory(filename, working_directory):
+def manage_file_with_unwanted_char(file_basename, sub_string):
+    """
+    Returns a treated base file name if unwanted characters like '(', ')' are present
+    :param file_basename: filename without full path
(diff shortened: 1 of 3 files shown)


source/tyr/tyr/binarisation.py
```diff
@@ -71,17 +71,42 @@ def unzip_if_needed(filename):
     return working_directory
 
 
-def move_to_backupdirectory(filename, working_directory):
+def manage_file_with_unwanted_char(file_basename, sub_string):
```
xlqian, line 74: what happens if file's name is `myfile(1).osm(2).pbf`?

xlqian (approved):

source/tyr/tyr/binarisation.py
kadhikari: It will rename as myfile_xxxx.pbf where xxx = formatted_datatime

benoit-bst (approved):

PR #3534, opened 2021-08-31
[TYR] handle multiple files and job ids

xlqian:
https://jira.kisio.org/browse/NAVITIAII-3550

### DIFF
--- a/source/tyr/tests/integration/datapost_test.py
+++ b/source/tyr/tests/integration/datapost_test.py
@@ -82,14 +82,14 @@ def test_post_pbf(create_instance_fr):
             f[0].close()
     assert resp.status_code == 200
 
-    assert os.path.isfile('/tmp/_JOB_ID_DELIMITER_1_JOB_ID_DELIMITER_empty_pbf.osm.pbf')
-    assert os.path.isfile('/tmp/_JOB_ID_DELIMITER_1_JOB_ID_DELIMITER_fusio.zip')
+    assert os.path.isfile('/tmp/<JOB_ID_DELIMITER>1<JOB_ID_DELIMITER>empty_pbf.osm.pbf')
+    assert os.path.isfile('/tmp/<JOB_ID_DELIMITER>1<JOB_ID_DELIMITER>fusio.zip')
 
-    os.remove('/tmp/_JOB_ID_DELIMITER_1_JOB_ID_DELIMITER_empty_pbf.osm.pbf')
-    os.remove('/tmp/_JOB_ID_DELIMITER_1_JOB_ID_DELIMITER_fusio.zip')
+    os.remove('/tmp/<JOB_ID_DELIMITER>1<JOB_ID_DELIMITER>empty_pbf.osm.pbf')
+    os.remove('/tmp/<JOB_ID_DELIMITER>1<JOB_ID_DELIMITER>fusio.zip')
 
-    assert not os.path.isfile('/tmp/_JOB_ID_DELIMITER_1_JOB_ID_DELIMITER_empty_pbf.osm.pbf')
-    assert not os.path.isfile('/tmp/_JOB_ID_DELIMITER_1_JOB_ID_DELIMITER_fusio.zip')
+    assert not os.path.isfile('/tmp/<JOB_ID_DELIMITER>1<JOB_ID_DELIMITER>empty_pbf.osm.pbf')
+    assert not os.path.isfile('/tmp/<JOB_ID_DELIMITER>1<JOB_ID_DELIMITER>fusio.zip')
 
 
 def test_post_no_file(create_instance_fr):
--- a/source/tyr/tests/integration/datapost_test.py
+++ b/source/tyr/tests/integration/datapost_test.py
@@ -63,22 +63,33 @@ def get_jobs_from_db(id=None):
 
 
 def test_post_pbf(create_instance_fr):
-    assert not os.path.isfile('/tmp/empty_pbf.osm.pbf')
+    assert not os.path.isfile('/tmp/_JOB_ID_DELIMITER_1_JOB_ID_DELIMITER_empty_pbf.osm.pbf')
+    assert not os.path.isfile('/tmp/_JOB_ID_DELIMITER_1_JOB_ID_DELIMITER_fusio.zip')
 
-    filename = 'empty_pbf.osm.pbf'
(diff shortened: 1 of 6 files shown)


source/tyr/tyr/binarisation.py
```diff
@@ -144,10 +144,14 @@ def __init__(self, timeout):
     def __call__(self, func):
         @wraps(func)
         def wrapper(*args, **kwargs):
+            job_ids = get_named_arg('job_ids', func, args, kwargs) or []
```
kadhikari, line 147: Where do we add argument_name 'job_ids' in kwargs ?

source/tyr/tyr/binarisation.py
```diff
@@ -144,10 +144,14 @@ def __init__(self, timeout):
...
             job_id = get_named_arg('job_id', func, args, kwargs)
+            if job_id:
+                job_ids.append(job_id)
+
             logging.debug('args: %s -- kwargs: %s', args, kwargs)
-            job = models.Job.query.get(job_id)
-            logger = get_instance_logger(job.instance, task_id=job_id)
+            job = models.Job.query.get(job_ids[0])
```
kadhikari, line 153: We always get the first job but about other jobs ?

source/tyr/tyr/binarisation.py
```diff
@@ -215,7 +220,7 @@ def fusio2ed(self, instance_config, filename, job_id, dataset_uid):
     dataset = _retrieve_dataset_and_set_state("fusio", job.id)
     instance = job.instance
 
-    logger = get_instance_logger(instance, task_id=job_id)
+    logger = get_instance_logger(instance, jobs_ids=[job_id])
```
kadhikari, line 223: Only one job id for jobs_ids ?

source/tyr/tyr/binarisation.py
```diff
@@ -578,19 +590,25 @@ def reload_data(self, instance_config, job_id):
         rabbit_mq_handler.publish(task.SerializeToString(), instance.name + '.task.reload')
     except:
         logger.exception('')
-        job.state = 'failed'
+        for job in jobs:
+            job.state = 'failed'
```
kadhikari, line 594: If one job fail, it will penalize all the other jobs ?

source/tyr/tyr/tasks.py
```diff
@@ -115,8 +124,26 @@ def import_data(
...
 
+    new_job = None
     for _file in files:
-        filename = None
+        # we check if there's a job_id in file's name, there's no job id is found, we have to create a new_job
+        job_id = retrive_job_id(_file)
+        if not job_id:
+            new_job = models.Job()
+            new_job.instance = instance
+            new_job.state = 'running'
```
kadhikari, line 152: Here we loose all the new_job created from second to second last  file without job id.

kadhikari (changes requested):

source/tyr/tyr/binarisation.py
xlqian: it's a decorator, 'jod_ids' come from the wrapped function..

source/tyr/tyr/binarisation.py
xlqian: yes, in fact, the reason why we extract a job from the db is to get the instance, and all jobs SHOULD possess the same instance,

source/tyr/tyr/binarisation.py
xlqian: yes

source/tyr/tyr/binarisation.py
xlqian: yep

kadhikari (approved):

source/tyr/tests/integration/datapost_test.py
```diff
@@ -63,22 +63,33 @@ def get_jobs_from_db(id=None):
 
 
 def test_post_pbf(create_instance_fr):
-    assert not os.path.isfile('/tmp/empty_pbf.osm.pbf')
+    assert not os.path.isfile('/tmp/_JOB_ID_DELIMITER_1_JOB_ID_DELIMITER_empty_pbf.osm.pbf')
+    assert not os.path.isfile('/tmp/_JOB_ID_DELIMITER_1_JOB_ID_DELIMITER_fusio.zip')
```
pbench, line 67: Maybe 
```suggestion
    assert not os.path.isfile('/tmp/<JOB_ID_DELIMITER>1<JOB_ID_DELIMITER>fusio.zip')
```
as below ?

source/tyr/tyr/__init__.py
```diff
@@ -88,4 +88,6 @@ def celery_setup_logging(*args, **kwargs):
 
 tyr_rabbit_mq_handler = RabbitMqHandler(app.config['CELERY_BROKER_URL'], "tyr_event_exchange")
 
+JOB_ID_DELIMITER = "<JOB_ID_DELIMITER>"
```
pbench, line 91: This will be written as part of filenames right ? 
It seems strange to me to use "<>" in filenames. Why not  "_JOB_ID_DELIMITER_" ?

source/tyr/tyr/binarisation.py
```diff
@@ -560,12 +565,19 @@ def shape2ed(self, instance_config, filename, job_id, dataset_uid):
...
 @celery.task(bind=True)
-def reload_data(self, instance_config, job_id):
+def reload_data(self, instance_config, job_ids):
     """ reload data on all kraken of this instance"""
-    job = models.Job.query.get(job_id)
-    instance = job.instance
-    logging.info("Unqueuing job {}, reload data of instance {}".format(job.id, instance.name))
-    logger = get_instance_logger(instance, task_id=job_id)
+    jobs = [models.Job.query.get(job_id) for job_id in job_ids]
+    assert jobs
```
pbench, line 571: Maybe log an error message and raise an exception instead of crashing ?

source/tyr/tyr/binarisation.py
```diff
@@ -578,19 +590,27 @@ def reload_data(self, instance_config, job_id):
...
-    instance = job.instance
+    jobs = []
+    for job_id in job_ids:
+        job = models.Job.query.get(job_id)
+        job.state = 'running ed2nav'
+        jobs.append(job)
+    models.db.session.commit()
 
-    logger = get_instance_logger(instance, task_id=job_id)
+    assert jobs
```
pbench, line 610: Maybe log an error message and raise an exception instead of crashing ?

pbench (approved):

PR #3537, opened 2021-09-03
ed: add adresses from ntfs to ed

benoit-bst:
### Add a new table that contains addresses from ntfs

The **NTFS** is allowed to contains addresses for **SP** and **SA** (https://github.com/CanalTP/ntfs-specification/blob/master/ntfs_fr.md#addressestxt-optionnel). The work consists to create a new table

table_adresses

and add a **address_id** field inside stop_point table to link it

### Next step
The next step will be to parse these fields and use these addresses in place of georef addresses for SP (furure PR)

### related ticket
https://jira.kisio.org/browse/NAVP-1895


source/sql/alembic/versions/2233bc5aef67_add_addresses_from_ntfs.py
```diff
@@ -0,0 +1,31 @@
...
+        sa.Column('house_number', sa.TEXT(), nullable=True),
+        sa.Column('street_name', sa.TEXT(), nullable=True),
+        schema='navitia',
+    )
+    op.add_column('stop_point', sa.Column('address_id', sa.TEXT(), nullable=True), schema='navitia')
+
+
+def downgrade():
+    op.drop_table('address', schema='navitia')
+    op.drop_column('stop_point', 'address_id', schema='navitia')
```
kadhikari, line 31: Can you invert above two lines as below ?
op.drop_column('stop_point', 'address_id', schema='navitia')
op.drop_table('address', schema='navitia')

kadhikari (changes requested):

source/sql/alembic/versions/2233bc5aef67_add_addresses_from_ntfs.py
benoit-bst: This is not the same table.
The first is **address** and the second **stop_point**. So the order is not mandatory :cat:

kadhikari: As address_id in the table stop_point is optional (Not NULL false), order of deletion is not important.

kadhikari (approved):

pbench (approved):

PR #3540, opened 2021-09-07
[Kraken] Expire rabbitmq message sent to Kirin (task.load_realtime)

AzHicham:
See ticket NAVP-1892
See also rabbitmq doc about TTL for more informations https://www.rabbitmq.com/ttl.html

### DIFF
--- a/source/kraken/maintenance_worker.cpp
+++ b/source/kraken/maintenance_worker.cpp
@@ -103,6 +103,9 @@ void MaintenanceWorker::load_realtime() {
     std::string stream;
     task.SerializeToString(&stream);
     auto message = AmqpClient::BasicMessage::Create(stream);
+    // Add expiration timeout (in ms) to the message (TTL)
+    message->Expiration(std::to_string(conf.kirin_timeout()));
+
     // we ask for realtime data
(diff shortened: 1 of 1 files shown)


pbougue (approved):
I trust you on how to use amqp library :see_no_evil: 
Otherwise, it's exactly what I was thinking of :+1: 

As for the QA, here are my ideas doable on dev (don't know who will do it, but I could):
* nominal (did not mess normal behavior): restart kraken (reload data) and check that the number of navitia's disruptions is the same or slightly more after reload than before.
* too slow kirin (the purpose of this work): shut kirin-reloader, restart kraken, check kirin_reloader queue for ~10 min (at least 5) and see that message stays at 1 (TTL erases previous request at the moment the new one is emitted).
Then restart kirin_reloader and check that message is consumed, and Navitia disruptions come back to the same amount as before restart.
:warning: Depending on how the messages are discarded, one might have to wait for the consumer to be triggered to have them all consumed at the start of consumer (one served, all other discarded).

pbench (approved):

benoit-bst (approved):

PR #3541, opened 2021-09-09
[kraken] require data version 13

benoit-bst:
Forgot the increment data version for https://github.com/CanalTP/navitia/pull/3537
Because bina is required


kadhikari:
This should only be merged along with the future PR on integration of NTFS address in kraken (ed2nav) non ?

PR #3544, opened 2021-09-10
[Jormun] Handle geo_status for bragi properly

AzHicham:
See ticket NAVITIAII-3598

Instead of Error 500 we now get the following message

Screenshot from 2021-09-10 13-02-40

### DIFF
--- a/source/jormungandr/tests/bragi_autocomplete_tests.py
+++ b/source/jormungandr/tests/bragi_autocomplete_tests.py
@@ -1667,6 +1667,25 @@ class AbstractAutocompleteAndRouting:
             assert address['id'] == '3.282103;49.847586'
             assert len(address['administrative_regions']) == 1
 
+    def test_geo_status(self):
+        """
+        geo_status is still not implemented for bragi
+        we test the return code and error message
+        """
+        url = 'https://host_of_bragi'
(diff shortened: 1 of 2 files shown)


source/jormungandr/tests/bragi_autocomplete_tests.py
```diff
@@ -1667,6 +1667,25 @@ def test_global_coords_uri_new_bragi_coord_id_fmt(self):
...
 
+    def test_geo_status(self):
+        """
+        geo_status is still not implemented for bragi
+        we test the return code and error message
+        """
+        url = 'https://host_of_bragi'
+        params = {'pt_dataset[]': 'main_routing_test', 'lon': 3.282103, 'lat': 49.84758, 'timeout': 200}
+
+        url += "/reverse?{}".format(urlencode(params, doseq=True))
```
kadhikari: I don't see any use of this line same as above 'url = 'https://host_of_bragi' '

kadhikari (changes requested):

source/jormungandr/tests/bragi_autocomplete_tests.py
AzHicham: fixed, thanks @kadhikari

kadhikari (approved):

azime (approved):

PR #3559, opened 2021-09-20
[Jormungandr] Upgrade "kombu" and "amqp"

azime:


source/tyr/requirements.txt
```diff
@@ -11,15 +11,14 @@ SQLAlchemy==1.3.3
...
 alembic==1.0.11
-amqp==1.4.9
+amqp==2.1.1
 aniso8601==0.82
-anyjson==0.3.3
 argparse==1.2.1
-billiard==3.3.0.13
-celery==3.1.7
+billiard==3.5.0.5
+celery==4.0.0
```
pbougue: I didn't check all other libs, but it looks weird to not go at least on the last "hotfix" version https://pypi.org/project/celery/#history
So in this case 4.0.2 would probably be more secure (and maybe 4.1.0 to match kirin, or even 4.1.1 to apply the same reasoning).

xlqian:
Following various dependencies problems, we have to update packages' versions one after another. In order to maintain at most the stability, as suggested, we just aligned these versions with Kirin  https://github.com/CanalTP/kirin/blob/master/requirements.txt#L50

.github/workflows/workflow.yml
```diff
@@ -47,6 +47,10 @@ jobs:
         ports:
           - 5672:5672
 
+      redis:
+        image: redis:6-alpine
+        ports:
+          - 6379:6379
```
pbench, line 53: Now I am curious as why redis was not needed before :thinking:

pbench (approved):

.github/workflows/workflow.yml
xlqian: exactly

xlqian: that's what i'm looking for since yesterday

pbougue: Indeed, that's weird.
For the record, we added a Redis container inside unit tests in Kirin: https://github.com/CanalTP/kirin/blob/72ae5d7a65c79bcca4ae9fbdf850b59d5ed353eb/tests/docker_wrapper.py#L243
It seems a little better to avoid forcing redis availability on platform that's running test... But it probably depends on what requires redis.

xlqian: It's actually due to Celery 4.1.0. After comparing the behaviour "before/after", celery does now check the status of Redis when a task is sent(https://github.com/CanalTP/navitia/blob/dev/source/tyr/tyr/resources.py#L2095), which was not the case.

pbougue: :facepalm: What a mess... Are you sure celery is correctly setup?
We may have the same silent issue on kirin (as tests provide redis).

But it means you are actually testing celery tasks? (too complicated to be worth it on Kirin)

xlqian: Redis_result_backend is used as an call-back object when a task is sent (the redis-connection is sent to the worker, too).
This was probably lazily done before (and tests' celery tasks aren't returning result).

PR #3569, opened 2021-09-30
[Kraken] Add new param (is_active): select only active vehicle journey at current_datetime

azime:
This param used by vehicle_positions API
Jiar: [NAVP-1862](https://jira.kisio.org/browse/NAVP-1862)


source/ptreferential/ptreferential_ng.cpp
```diff
@@ -263,6 +263,13 @@ struct Eval : boost::static_visitor<Indexes> {
             }
         } else if (type == Type_e::VehicleJourney && f.method == "has_disruption" && f.args.empty()) {
             indexes = get_indexes_by_impacts(type::Type_e::VehicleJourney, data);
+        } else if (type == Type_e::VehicleJourney && f.method == "has_active") {
+            // useful only for VJ for vehicle_positions api
+            const auto rt_level =
+                (type == Type_e::VehicleJourney) ? rt_level_from_string(f.args.back()) : type::RTLevel::Base;
```
pbench: I fear using `f.args.back()` will cause problem in the future if we add more arguments to the function.
Maybe use the index of the argument instead of `back()` ?

source/type/vehicle_journey.cpp
```diff
@@ -162,6 +162,11 @@ uint32_t VehicleJourney::earliest_time() const {
     return std::min(st.arrival_time, st.boarding_time);
 }
 
+uint32_t VehicleJourney::last_time() const {
+    const auto& st = navitia::last_stop_time(stop_time_list);
```
pbench: why not just return stop_time_list.back() ? The stop_time_list may not be sorted by increasing times ?

source/ptreferential/ptreferential_ng.cpp
```diff
@@ -263,6 +263,13 @@ struct Eval : boost::static_visitor<Indexes> {
             }
         } else if (type == Type_e::VehicleJourney && f.method == "has_disruption" && f.args.empty()) {
             indexes = get_indexes_by_impacts(type::Type_e::VehicleJourney, data);
+        } else if (type == Type_e::VehicleJourney && f.method == "has_active") {
```
pbougue: :gb: Shouldn't it be "is_active" ?

source/type/vehicle_journey.h
```diff
@@ -107,6 +107,7 @@ struct VehicleJourney : public Header, Nameable, hasVehicleProperties {
 
     // Return the smallest time within its stop_times
     uint32_t earliest_time() const;
+    uint32_t last_time() const;
```
pbougue: To match earliest ('last' matches 'first'):
```suggestion
    uint32_t latest_time() const;
```

pbougue (commented):
This should be added to `?filter=<f>` documentation too: https://github.com/CanalTP/navitia/blob/dev/documentation/rfc/ptref_grammar.md

source/ptreferential/ptreferential_utils.cpp
```diff
@@ -152,6 +152,22 @@ Indexes get_impacts_by_tags(const std::vector<std::string>& tag_name, const Data
...
+        return false;
+    }
+    if ((vj->validity_patterns[rt_level] != nullptr)
+        && !vj->validity_patterns[rt_level]->check(current_datetime.date())) {
+        return false;
+    }
+    const auto& first_departure_dt = vj->earliest_time();
+    const auto& last_arrival_dt = vj->last_time();
+    bt::time_period period = {bt::ptime(current_datetime.date(), bt::seconds(first_departure_dt)),
+                              bt::ptime(current_datetime.date(), bt::seconds(last_arrival_dt))};
```
pbench: I think you may miss some active vj here because of over midnight stop_times.
If first_departure_dt and/or last_arrival_dt are after midnight, you also need to do the same check with the date just before current_datetime.date()

source/ptreferential/ptreferential_ng.cpp
```diff
@@ -436,6 +444,13 @@ std::string make_request(const Type_e requested_type,
             }
             break;
         case Type_e::VehicleJourney:
+            if (current_datetime != boost::none) {
+                res = "(" + res + ") AND " + navitia::type::static_data::get()->captionByType(requested_type);
+                res = res + ".has_active(" + to_iso_string(*current_datetime) + "Z";
+                res = res + ", " + rt_level_to_string(rt_level);
+                res = res + ")";
+                break;
+            }
```
pbench, line 452:
```suggestion
            }    
            break;
```
Unless you *do* want to execute the case `Type_e::Impact` when `current_datetime == boost::none`

azime: No, Only vehicle_journey

source/ptreferential/ptreferential_ng.cpp
azime: OK

source/ptreferential/ptreferential_ng.cpp
azime: ok

source/ptreferential/ptreferential_ng.cpp
kadhikari: I vote for is_active also :smile:

source/ptreferential/ptreferential_ng.cpp
kadhikari: I think we better not use break as this case will only applied for vehicle_journey and with current_datetime, non?

source/position/position_api.cpp
```diff
@@ -39,19 +39,13 @@ void vehicle_positions(PbCreator& pb_creator,
...
-    const auto start = get_optional_value_or(since, pb_creator.now);
-    const auto end = get_optional_value_or(until, pb_creator.now + boost::posix_time::minutes(2 * 10));
-    pb_creator.action_period = boost::posix_time::time_period(start, end);
-
     type::Indexes idx_vjs;
     try {
-        idx_vjs = ptref::make_query(type::Type_e::VehicleJourney, filter, forbidden_uris, type::OdtLevel_e::all, start,
-                                    end, data);
+        idx_vjs = ptref::make_query(type::Type_e::VehicleJourney, filter, forbidden_uris, type::OdtLevel_e::all,
+                                    boost::none, boost::none, type::RTLevel::RealTime, data, pb_creator.now);
```
kadhikari, line 48: Why RealTime ?

source/ptreferential/ptreferential_utils.cpp
kadhikari: As we suppose a vehicle_journey is active for less than 24 hours, we can test as below:
if last_arrival_dt < first_departure_dt then, vehicle_journey end date_time = bt::ptime(current_datetime.date() + 1 day, bt::seconds(last_arrival_dt))

source/type/vehicle_journey.cpp
kadhikari: Since 'earliest_stop_time' exists, i think stop_time_list probably not ordered.

source/type/vehicle_journey.h
kadhikari: I think here last doesn't mean latest but destination arrival_time -> date_time of end station.

source/type/type_utils.cpp
```diff
@@ -69,6 +69,14 @@ const nt::StopTime& earliest_stop_time(const std::vector<nt::StopTime>& sts) {
...
 }
 
+const nt::StopTime& last_stop_time(const std::vector<nt::StopTime>& sts) {
+    assert(!sts.empty());
+    const auto& max_st = std::max_element(sts.begin(), sts.end(), [](const nt::StopTime& st1, const nt::StopTime& st2) {
+        return std::max(st1.boarding_time, st1.arrival_time) < std::max(st2.boarding_time, st2.arrival_time);
+    });
+    return *max_st;
+}
+
```
kadhikari: I have a doubt, since the value st1.boarding_time or st1.arrival_time is 'seconds since midnight' what if the last StopTime is next day ?

source/type/vehicle_journey.h
pbougue: Then it's the same above  for "earliest" and it should be "first"? Either is fine, but it looks like it should match.

kadhikari (approved):
Great tests :+1:

benoit-bst (approved):

PR #3573, opened 2021-10-03
[CI] Adding debian 11 support

AzHicham:
This PR make the following change:
- ensure navitia could be build on debian11 platform, without breaking compatibility with Debian 8 to 10

However, on Debian 11 we build, test and package navitia only with python2.7, for python3.9 an other PR will follow soon.


benoit-bst (approved):

xlqian (approved):
:+1:

PR #3582, opened 2021-10-12
[CI] Fix tag publish docker images navitia

AzHicham:


pbench (approved):

benoit-bst (commented):
This change allows to have the **latest** tag and the **version** tag too ?

pbench:
> This change allows to have the **latest** tag and the **version** tag too ?

The version tag is always set https://github.com/CanalTP/navitia-docker-compose/blob/011fbe8eddc2bd6feaa1a1664ef186820322ef27/builder_from_package/build.sh#L170-L175

benoit-bst (approved):

AzHicham:
> This change allows to have the **latest** tag and the **version** tag too ?

Yes, you can see [here](https://hub.docker.com/r/navitia/jormungandr/tags) previous build docker, image with sha fc90e9e390f1 has two tags : navitia_builder & v13.0.0-70-g32011b4ce

PR #3590, opened 2021-10-19
Manage skipped_stop as new stop_time type from data file to response json

kadhikari:
A new staus of a vehicle on a stop 'skipped_stop' is added in stop_times. It will allow to inform passengers in a station that the vehicle will not stop.. Below are some details on modifications:
* An attribute skipped_stop (bool) is added  in the table navitia.stop_time.
* An attribute skipped_stop (bool) is added  in the navitia-proto.type.StopTime
* New attribute skipped_stop added in vehicle_journey.stop_time in the response json navitia.
* New attribute skipped_stop added in journey.section.stop_date_time.additional_informations

For details: https://jira.kisio.org/browse/NAVP-1942


pbench (approved):

source/sql/alembic/versions/2c69b03d20fc_add_skipped_stop_in_stop_time.py
```diff
@@ -0,0 +1,27 @@
...
+import sqlalchemy as sa
+import geoalchemy2 as ga
+
+
+def upgrade():
+    op.add_column(
+        'stop_time',
+        sa.Column('skipped_stop', sa.BOOLEAN(), nullable=False, server_default="false"),
+        schema='navitia',
+    )
```
azime, line 23: We should do an update for all stoptimes in db, no ?

kadhikari: The attribute is initialized with false for each existing record in stop_time

xlqian (approved):

source/routing/raptor_api.cpp
```diff
@@ -565,8 +565,10 @@ static bt::ptime handle_pt_sections(pbnavitia::Journey* pb_journey,
                     continue;
                 }
 
-                // skipping stop_time having a 'deleted' or 'deleted for detour' stop_point
-                if (!item.stop_times[i]->pick_up_allowed() && !item.stop_times[i]->drop_off_allowed()) {
+                // skipping stop_time having a 'deleted' or 'deleted for detour' stop_point except skipped_stop
+                // Make sure that a realtime deleted stop_time has skipped_stop = false
```
pbench:
```suggestion
                // do not add a stop_time with pick_up_allowed = false and drop_off_allowed = false
                // unless skipped_stop = true 
                // Make sure that a realtime deleted stop_time has skipped_stop = false
```

pbench (approved):

azime (approved):

source/routing/raptor_api.cpp
kadhikari: Done

benoit-bst (approved):
You rules.

PR #3599, opened 2021-11-04
Skip inserting last_dataset files into ed database

kadhikari:
Add an option to avoid all the actions of type xxx_2ed (inserting source data files into ed database) while using the command **import_last_dataset** in tyr

Jira: https://jira.kisio.org/browse/NAVITIAII-3647

* Tested in local with @benoit-bst on tyr having **two** coverages and 5 source data files(five xxx_2ed to be done).
* command line tested: 
** PYTHONPATH=/../navitiacommon TYR_CONFIG_FILE=dev_settings.py python manage_tyr.py import_last_dataset idfm --skip_2ed
** PYTHONPATH=/../navitiacommon TYR_CONFIG_FILE=dev_settings.py python manage_tyr.py import_last_dataset idfm
* Tested also adding a new data source in /source
* When command line used with --skip_2ed, there is no dataset attached to the job. This fact doesn't have any unwanted effect on **last_dataset**.

### DIFF
--- a/source/tyr/tyr/command/import_last_dataset.py
+++ b/source/tyr/tyr/command/import_last_dataset.py
@@ -36,7 +36,13 @@ from tyr.helper import get_instance_logger, wait_or_raise
 
 @manager.command
 def import_last_dataset(
-    instance_name, background=False, reload_kraken=False, custom_output_dir=None, nowait=False, allow_mimir=False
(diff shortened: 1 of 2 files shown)


kadhikari:
Replaced by https://github.com/CanalTP/navitia/pull/3601

PR #3600, opened 2021-11-04
[Tyr] Indexing data for Elasticsearch2 en Elasticsearch7 while doing Import_data depending on configuration.

azime:
Jira: [NAVP-1952](https://jira.kisio.org/browse/NAVP-1952)


source/tyr/tyr/tasks.py
```diff
@@ -208,30 +208,35 @@ def send_to_mimir(instance, filename, family_type):
     job.instance = instance
     job.state = 'running'
 
-    dataset = models.DataSet()
-    dataset.family_type = 'mimir'
-    dataset.type = 'fusio'
+    if is_activate_autocomplete_version(7):
```
kadhikari, line 211: We can add a comment above this line 'We create a dataset for each autocomplete version if active'

source/tyr/tests/integration/mimir_test.py
```diff
@@ -71,7 +62,21 @@ def test_mimir_family_type_poi(create_instance, enable_mimir):
         assert len(actions) == 2  # poi2mimir + finish
 
 
-def test_mimir_ntfs_false(create_instance, enable_mimir):
+def test_mimir2_and_mimir_family_type_poi(create_instance, enable_mimir2_and_mimir):
+    with app.app_context():
+        instance = models.Instance.query.get(create_instance)
+        actions = []
+        actions.extend(tasks.send_to_mimir(instance, 'test.poi', 'poi'))
+        assert actions[0].task == 'tyr.binarisation.poi2mimir'
```
kadhikari, line 70: assert actions[1].task ??

kadhikari (commented):
great job :+1: 
Seems fine to me. Since we can't test all with existing unit and integration test, I propose to test with a tyr local for each functionality.

source/tyr/tests/integration/datapost_test.py
```diff
@@ -62,7 +62,7 @@ def get_jobs_from_db(id=None):
         return models.Job.get(id=id)
 
 
-def test_post_pbf(create_instance_fr):
+def test_post_pbf(create_instance_fr, enable_mimir2):
```
benoit-bst, line 65: i don't know if it is easy, but maybe `enable_mimir_es7` is better ?

benoit-bst (approved):

kadhikari (approved):

kadhikari (approved):

PR #3606, opened 2021-11-09
[Jormun] Filter duplicate journey

AzHicham:
Please see ticket NAVITIAII-3724


xlqian (approved):

kadhikari (commented):
Will it be possible to add a test ??

AzHicham:
> Will it be possible to add a test ??

Unfortunately for this case, we don't have enough bss data to reproduce the bug in ticket

kadhikari (approved):

PR #3608, opened 2021-11-15
[loki] Add _use_pt_socket parameter

pbench:
When adding `&_override_scenario=distributed&_use_pt_socket=True` to a request, jormungandr will use the `pt_zmq_socket` to solve `pt_planner` distributed subrequest.
This will have no effect if pt_zmq_socket is not set in the jormungandr configuration file of the instance.

This will be useful to deploy and test the loki backend.

### DIFF
--- a/source/jormungandr/jormungandr/instance.py
+++ b/source/jormungandr/jormungandr/instance.py
@@ -748,8 +748,15 @@ class Instance(object):
         logger = logging.getLogger(__name__)
         deadline = datetime.utcnow() + timedelta(milliseconds=timeout)
         request.deadline = deadline.strftime('%Y%m%dT%H%M%S,%f')
-        pt_socket = request.requested_api == type_pb2.pt_planner
-        with self.socket(self.context, pt_socket) as socket:
+        # we use the pt_socket if :
+        #  - we are asking for the pt_planner api
+        #  - _use_pt_socket is set to True in the request's parameters
+        use_pt_socket = (
+            request.requested_api == type_pb2.pt_planner
+            and 'use_pt_socket' in kwargs
+            and kwargs['use_pt_socket']
+        )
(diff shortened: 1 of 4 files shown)


AzHicham (approved):

source/jormungandr/jormungandr/interfaces/v1/Journeys.py
```diff
@@ -389,6 +389,12 @@ def __init__(self):
             hidden=True,
             help="debug param to specify a custom scenario",
         )
+        parser_get.add_argument(
+            "_use_pt_socket",
+            type=six.text_type,
```
xlqian: shouldn't it be boolean?

source/jormungandr/jormungandr/instance.py
```diff
@@ -748,8 +748,15 @@ def _send_and_receive(
         logger = logging.getLogger(__name__)
         deadline = datetime.utcnow() + timedelta(milliseconds=timeout)
         request.deadline = deadline.strftime('%Y%m%dT%H%M%S,%f')
-        pt_socket = request.requested_api == type_pb2.pt_planner
-        with self.socket(self.context, pt_socket) as socket:
+        # we use the pt_socket if :
```
xlqian: i think it would be a little bit clearer you pass this parameter where the pt_journey is requested... so don't have to check here

source/jormungandr/jormungandr/interfaces/v1/Journeys.py
pbench: yep ! Thanks

source/jormungandr/jormungandr/instance.py
pbench: Allright, done

xlqian (approved):

kadhikari (approved):

PR #3611, opened 2021-11-18
[Jormun] Fix link by_pass_disruption

AzHicham:


source/jormungandr/jormungandr/scenarios/simple.py
```diff
@@ -364,8 +364,8 @@ def _add_ridesharing_link(self, resp, params, instance):
         add_link(resp, rel='ridesharing_journeys', **req)
 
     def _add_bypass_disruptions_link(self, resp, params):
-        # find first impact wit a NO_SERVICE severity
-        if 'data_freshness' in params and params['data_freshness'] == 'realtime':
+        # find first impact with a NO_SERVICE severity
+        if 'data_freshness' in params and 'realtime' in params['data_freshness']:
```
azime:
```suggestion
        if params.get("data_freshness", None) == 'realtime':
```
No ?

AzHicham: Unfortunatly params['data_freshness'] returns ['realtime'] and not 'realtime' 
Does your suggestion still works in this case ?

pbench (approved):

source/jormungandr/jormungandr/scenarios/simple.py
xlqian:
```suggestion
        if 'realtime' in params.get('data_freshness', []):
```

xlqian (approved):

PR #3614, opened 2021-11-22
[Ed]: add inputs, outputs and pathways inside ed db

benoit-bst:
## Topic

This is the first part of the **Input/Output** journey.
Several steps need to be done
- Read and **insert** Inputs/Outputs and Pathways inside ED DB
- **Create** the data.nav with ED that includes IO
- Use this information and delivers it with the **proximity list** EndPoint
- ... The future Steps are an adaptation of the response with IO gateway

A full grooming explamation is available here -> https://confluence.kisio.org/pages/viewpage.action?pageId=84321143

## The scope
We focus on _Read and insert **Inputs/Outputs** and **Pathways** inside ED DB_ 
The JIRA ticket -> https://jira.kisio.org/browse/NAVP-1965
To be easily readable, I propose a DB schema that follows the same NTFS structure for **inputs/outputs** and **pathways**

**Input/output** is a stop inside **stops.txt** with a _location_type=3_ -> 
https://github.com/CanalTP/ntfs-specification/blob/master/ntfs_fr.md#stopstxt-requis

The corresponding table is:
input_output

**Pathway.txt** is a new file for the fusio parser. The NTFS says -> https://github.com/CanalTP/ntfs-specification/blob/master/ntfs_fr.md#pathwaystxt-optionnel

The corresponding table is:
pathway_table

## Tests

A batch of tests are coded and I hand tested the fusio2ed with a real data (IDFM NTFS Files)


source/type/input_output.h
```diff
@@ -0,0 +1,59 @@
...
+
+#include <boost/container/flat_set.hpp>
+
+#include <vector>
+#include <set>
+
+namespace navitia {
+namespace type {
+
+struct InputOutput : public Header, Nameable, hasProperties, HasMessages {
```
benoit-bst: for the moment, this class is not used. It will be in a next PR to read ED DB and create the data.nav

source/ed/connectors/gtfs_parser.cpp
```diff
@@ -501,12 +517,91 @@ StopsGtfsHandler::stop_point_and_area StopsGtfsHandler::handle_line(Data& data,
...
+    traversal_time_c = csv.get_pos_col("traversal_time");
+    stair_count_c = csv.get_pos_col("stair_count");
+    max_slope_c = csv.get_pos_col("max_slope");
+    min_width_c = csv.get_pos_col("min_width");
+    signposted_as_c = csv.get_pos_col("signposted_as");
+    reversed_signposted_as_c = csv.get_pos_col("reversed_signposted_as");
+}
+
+int PathWayGtfsHandler::fill_pathway_field(const csv_row& row, const int key, const std::string type) {
+    int pw_value = UNKNOWN_COLUMN;
```
pbench: The constant UNKNOWN_COLUMN is used in the rest of the code to handle a column name which is absent in the csv, but expected to be there by the code.
Here you use the same constant to signal that some value in the csv is ill-formed (a float was expected).
Maybe use another constant ?

source/ed/connectors/gtfs_parser.cpp
```diff
@@ -501,12 +517,91 @@ StopsGtfsHandler::stop_point_and_area StopsGtfsHandler::handle_line(Data& data,
...
+    if (has_col(key, row) && row[key] != "") {
+        try {
+            return (int)boost::lexical_cast<float>(row[key]);
```
pbench, line 552: you try to parse the value into a float, and this float is then casted into an int.
Could you just try to parse an int ?

source/ed/connectors/gtfs_parser.cpp
```diff
@@ -501,12 +517,91 @@ StopsGtfsHandler::stop_point_and_area StopsGtfsHandler::handle_line(Data& data,
...
+    length_c = csv.get_pos_col("length");
+    traversal_time_c = csv.get_pos_col("traversal_time");
+    stair_count_c = csv.get_pos_col("stair_count");
+    max_slope_c = csv.get_pos_col("max_slope");
+    min_width_c = csv.get_pos_col("min_width");
+    signposted_as_c = csv.get_pos_col("signposted_as");
+    reversed_signposted_as_c = csv.get_pos_col("reversed_signposted_as");
+}
+
+int PathWayGtfsHandler::fill_pathway_field(const csv_row& row, const int key, const std::string type) {
```
pbench:
```suggestion
int PathWayGtfsHandler::fill_pathway_field(const csv_row& row, const int key, const std::string column_name) {
```
here 'type' is the name of the column, right ?

source/type/input_output.cpp
```diff
@@ -0,0 +1,63 @@
...
+
+namespace navitia {
+namespace type {
+template <class Archive>
+void InputOutput::serialize(Archive& ar, const unsigned int /*unused*/) {
+    ar& uri& name& stop_area& coord& idx;
+}
+SERIALIZABLE(InputOutput)
+
+bool InputOutput::operator<(const InputOutput& other) const {
```
pbench: Out of curiosity : why do you need a comparator ? Do you sort these objects at some point ?

pbench (approved):

source/ed/connectors/gtfs_parser.cpp
benoit-bst: probably a `DEFAULT_PATHWAYS_VALUE` is better yes

source/ed/connectors/gtfs_parser.cpp
benoit-bst: This is a little funny trick to have one function to convert a **string** to **int/float**

source/type/input_output.cpp
benoit-bst: All objet (Line, StopArea, ...) into the repository are with this operator. For a sort purpose

kadhikari (approved):

PR #3616, opened 2021-11-30
ed: read access_point and pathway table then aggregate it

benoit-bst:
This is the second part of the Epic **Entrance/Exit**. 
Following the first job https://github.com/CanalTP/navitia/pull/3614, we have to read the ED table to create the new AccessPoint vector linked into each StopPoint.
That is an AccessPoint -> https://github.com/CanalTP/navitia/pull/3616/files#diff-640a43f77adb4783b283fa233d472b13b5228d3313ce7e83bce717d1ea0e9740R39

At the end **AccessPoint** are exposed inside **StopPoint** like this -> https://github.com/CanalTP/navitia/pull/3616/files#diff-e0441f873b68f4bee05b0865750b7d996e6e4f706b2d9a297dbacc64dfb9a388R56

### DIFF
--- a/source/ed/ed_reader.cpp
+++ b/source/ed/ed_reader.cpp
@@ -666,8 +666,8 @@ void EdReader::fill_access_point_field(navitia::type::AccessPoint* access_point,
     if (!const_it["signposted_as"].is_null()) {
         const_it["signposted_as"].to(access_point->signposted_as);
     }
-    if (!const_it["reverse_signposted_as"].is_null()) {
-        const_it["reverse_signposted_as"].to(access_point->reversed_signposted_as);
+    if (!const_it["reversed_signposted_as"].is_null()) {
+        const_it["reversed_signposted_as"].to(access_point->reversed_signposted_as);
     }
     // link with SP
     auto sp_key = uri_to_idx_stop_point.find(sp_id);
@@ -727,7 +727,7 @@ void EdReader::fill_access_points(nt::Data& data, pqxx::work& work) {
         "max_slope, "
         "min_width, "
         "signposted_as, "
-        "reverse_signposted_as "
+        "reversed_signposted_as "
         "FROM navitia.pathway";
 
     result = work.exec(request);
--- a/source/ed/ed_reader.cpp
+++ b/source/ed/ed_reader.cpp
@@ -751,6 +751,9 @@ void EdReader::fill_access_points(nt::Data& data, pqxx::work& work) {
             fill_access_point_field(to_access_p->second, const_it, false, from_stop_id);
             continue;
         }
+
+        // An other case exists: StopPoint <=> StopPoint connection.
+        // In the future, it will be handled like AccessPoint <=> SP
     }
 }
 
--- a/source/ed/ed_reader.cpp
+++ b/source/ed/ed_reader.cpp
@@ -670,7 +670,7 @@ void EdReader::fill_access_point_field(navitia::type::AccessPoint* access_point,
         const_it["reversed_signposted_as"].to(access_point->reversed_signposted_as);
     }
     // link with SP
-    auto sp_key = uri_to_idx_stop_point.find(sp_id);
+    auto sp_key = uri_to_idx_stop_point.find("stop_point:" + sp_id);
     if (sp_key != uri_to_idx_stop_point.end()) {
         auto sp = stop_point_map.find(sp_key->second);
         if (sp != stop_point_map.end()) {
@@ -699,7 +699,7 @@ void EdReader::fill_access_points(nt::Data& data, pqxx::work& work) {
         ap->coord.set_lon(const_it["lon"].as<double>());
         ap->coord.set_lat(const_it["lat"].as<double>());
         // parent station is Stop Area
-        auto sa_key = uri_to_idx_stop_area.find(const_it["parent_station"].as<std::string>());
+        auto sa_key = uri_to_idx_stop_area.find("stop_area:" + const_it["parent_station"].as<std::string>());
         if (sa_key != uri_to_idx_stop_area.end()) {
             auto sa = stop_area_map.find(sa_key->second);
             if (sa != stop_area_map.end()) {
--- a/source/ed/ed_reader.cpp
+++ b/source/ed/ed_reader.cpp
@@ -684,10 +684,10 @@ void EdReader::fill_access_point_field(navitia::type::AccessPoint* access_point,
 void EdReader::fill_access_points(nt::Data& data, pqxx::work& work) {
     // access_point
     std::string request =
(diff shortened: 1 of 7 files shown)


source/type/stop_point.h
```diff
@@ -53,6 +53,7 @@ struct StopPoint : public Header, Nameable, hasProperties, HasMessages {
 
     StopArea* stop_area;
     std::vector<navitia::georef::Admin*> admin_list;
+    std::set<navitia::type::AccessPoint*> access_points;
```
benoit-bst, line 56: An AccessPoint has to be unique for the StopPoint point of view.
This is the reason of the set

source/ed/ed_reader.cpp
```diff
@@ -619,6 +622,135 @@ void EdReader::fill_stop_points(nt::Data& data, pqxx::work& work) {
...
+            continue;
+        }
+        // Access Point URI match for to_stop_id
+        // so, from_stop_id have to be a StopPoint
+        auto to_access_p = access_point_map.find(to_stop_id);
+        if (to_access_p != access_point_map.end()) {
+            fill_access_point_field(to_access_p->second, const_it, false, from_stop_id);
+            continue;
+        }
     }
```
pbench: Maybe log an error here, in case neither from_stop_id and to_stop_id are Access Point URI

source/type/stop_point.cpp
```diff
@@ -49,7 +50,7 @@ void StopPoint::serialize(Archive& ar, const unsigned int /*unused*/) {
     //
     // stop_point_connection_list is managed by StopPointConnection
     ar& uri& label& name& stop_area& coord& fare_zone& is_zonal& address_id& idx& platform_code& admin_list&
-        _properties& impacts& dataset_list& address;
+        access_points& _properties& impacts& dataset_list& address;
```
pbench, line 52: The comment above says that serializing lists may create problems.
Is it complicated to do something like what is done for stop_point_connection_list ? 
It seems to be filled here : https://github.com/CanalTP/navitia/blob/1049b0a053b4d54b3ce2b9bd64ced9390bc42398/source/type/connection.cpp#L49-L51

source/type/access_point.h
```diff
@@ -33,24 +33,33 @@ www.navitia.io
...
+    // parameters
+    std::string stop_code = "";
+    bool is_entrance = false;
+    bool is_exit = false;
+    uint pathway_mode = -1;
+    uint length = -1;
+    uint traversal_time = -1;
+    uint stair_count = -1;
+    uint max_slope = -1;
+    uint min_width = -1;
```
pbench: Setting an **unsigned** int to -1 always feels strange to me.
Maybe use 0 as a default value ?

benoit-bst: hahaha my bad, at the start it was 0, but i needed to handle the 0, so i decided to put -1 for the default value and it is not an **uint** obviously. thanks !
It compiles...
Don't know why it is work here https://github.com/CanalTP/navitia/pull/3616/files#diff-65c56ad7590d85028f4c4d16f93f969c03d1640a35a333f2be0ba097ea7a2e2fR461

source/type/stop_point.cpp
benoit-bst: it seems to be true for **stop_connection_list** but not for the others list like `admin_list, data_set_list, route_list`.
The comment is not true ...change it

source/ed/ed_reader.cpp
benoit-bst: this case can exist with a SP->SP connection but we don't want to handle for the moment.
I'll add a comment to explain the unhandled case

source/type/stop_point.cpp
pbench: :+1:

source/type/access_point.h
pbench: It works because -1 is a 'valid' value for uint, it is equal to  UINT::Max :grimacing:

pbench (approved):

source/ed/ed_reader.cpp
```diff
@@ -619,6 +622,138 @@ void EdReader::fill_stop_points(nt::Data& data, pqxx::work& work) {
...
+            access_point->is_exit = true;
+        } else {
+            if (from_access_point) {
+                access_point->is_entrance = true;
+                access_point->is_exit = false;
+            } else {
+                access_point->is_entrance = false;
+                access_point->is_exit = true;
+            }
+        }
```
xlqian, line 649: maybe we can reduced nested `if` a little bit?
```suggestion
       [access_point](){
           if (is_bidirectional) {
               access_point->is_entrance = true;
               access_point->is_exit = true;
               return;
           }          
           if (from_access_point) {
               access_point->is_entrance = true;
               access_point->is_exit = false;
               return;
           }
           access_point->is_entrance = false;
           access_point->is_exit = true;
       }();
        
```

source/ed/ed_reader.cpp
```diff
@@ -619,6 +622,138 @@ void EdReader::fill_stop_points(nt::Data& data, pqxx::work& work) {
...
         this->stop_point_map[const_it["id"].as<idx_t>()] = sp;
+        this->uri_to_idx_stop_point[const_it["uri"].as<std::string>()] = const_it["id"].as<idx_t>();
+    }
+}
+
+void EdReader::fill_access_point_field(navitia::type::AccessPoint* access_point,
+                                       const pqxx::result::iterator const_it,
+                                       const bool from_access_point,
+                                       const std::string& sp_id) {
+    if (!const_it["pathway_mode"].is_null()) {
```
xlqian, line 633: very minor suggestion:

```suggestion
    auto is_null_else = [](auto obj, auto f){
    if (obj.is_null()) return;
    f(obj);    
    }; 
    is_null_else(const_it["pathway_mode"], [access_point](auto obj){ access_point->pathway_mode = obj; } );
    is_null_else(const_it["length"], [access_point](auto obj){ access_point->length = obj; }  );
   .......
```

source/ed/ed_reader.cpp
```diff
@@ -619,6 +622,138 @@ void EdReader::fill_stop_points(nt::Data& data, pqxx::work& work) {
...
+        auto sp = stop_point_map.find(sp_key->second);
+        if (sp != stop_point_map.end()) {
+            sp->second->access_points.insert(access_point);
+        }
+    } else {
+        LOG4CPLUS_ERROR(log, "pathway.to_stop_id not match with a stop point uri " << sp_id);
+    }
+}
+
+void EdReader::fill_access_points(nt::Data& data, pqxx::work& work) {
```
xlqian, line 684: data is an unused argument, is it mandatory?

xlqian (approved):
:+1:

PR #3623, opened 2021-12-14
filter twisted journeys

xlqian:


source/routing/raptor_api.h
```diff
@@ -188,6 +188,15 @@ bool is_way_later(const Journey& j1, const Journey& j2, const NightBusFilter::Pa
...
 void filter_late_journeys(RAPTOR::Journeys& journeys, const NightBusFilter::Params& params);
 
+/**
+ * @brief filter "twisted" journeys, a journeys is considered "twisted" if its last/first stop_point's
+ * stop_area has been served in the previous/following sections
+ *
+ * @param journeys A container of Journeys
+ * @param clockwise depart after or arrive before
+ */
+void filter_twisted_journeys(RAPTOR::Journeys& journeys, const bool clockwise);
```
pbench, line 198:
```suggestion
 * @brief remove "backtracking" journeys.
 * A "depart after' journey is considered "backtracking" if its last stop_point's stop_area has been visited in the previous sections.
 * An "arrive before' journey is considered "backtracking" if its first stop_point's stop_area has been visited in the subsequent sections
 *
 * @param journeys A container of Journeys
 * @param clockwise depart after or arrive before
 */
void filter_twisted_journeys(RAPTOR::Journeys& journeys, const bool clockwise);
```

pbench (approved):
I suggest to replace "twisted" by "backtracking". Otherwise everything looks good !

source/routing/raptor_api.cpp
```diff
@@ -1365,6 +1367,65 @@ void filter_late_journeys(RAPTOR::Journeys& journeys, const NightBusFilter::Para
...
+            continue;
+        }
+
+        bool found = false;
+
+        if (clockwise) {
+            auto section_it = journey.sections.rbegin();
+            const auto& last_stop_area_uri = section_it->get_out_st->stop_point->stop_area->uri;
+            for (++section_it; section_it != journey.sections.rend(); ++section_it) {
+                assert(section_it->get_in_st < section_it->get_out_st);
```
pbench, line 1394: maybe log an error and exit this function instead of asserting, so that the program don't crash if that happens

PR #3630, opened 2022-01-04
[jormun]: Add brahi7 in autocomplete

kadhikari:
* bragi7 added in instance.autocomplete_backend
* we can use _aucomplete=bragi7 in jormun

### DIFF
--- a/source/jormungandr/jormungandr/interfaces/v1/Places.py
+++ b/source/jormungandr/jormungandr/interfaces/v1/Places.py
@@ -138,6 +138,7 @@ class Places(ResourceUri):
             "_autocomplete",
             type=six.text_type,
             hidden=True,
+            choices=['kraken', 'bragi', 'bragi7'],
             help="name of the autocomplete service, used under the hood",
         )
         self.parsers['get'].add_argument(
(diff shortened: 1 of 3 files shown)


AzHicham (approved):

source/jormungandr/jormungandr/interfaces/v1/Places.py
```diff
@@ -198,7 +199,10 @@ def get(self, region=None, lon=None, lat=None):
             response = i_manager.dispatch(args, "places", instance_name=self.region)
         else:
             available_instances = get_all_available_instances(user)
-            autocomplete = global_autocomplete.get('bragi')
+            # If parameter '_autocomplete' is absent use 'bragi' as default value
+            if args["_autocomplete"] is None:
```
xlqian, line 203: maybe you can set a default value in that argument?

pbench (approved):

benoit-bst (approved):

source/jormungandr/jormungandr/interfaces/v1/Places.py
kadhikari: * Here, parameter exist but without any value : args["_autocomplete"] is None
* If we don't use the private parameter '_autocomplete' it is present without any value.
* In this case we cannot init a default value.

xlqian (approved):

PR #3634, opened 2022-01-06
Access Point: Expose access_point inside stop_point output api

benoit-bst:
### Sum up

- Create a new ed builder function to add Access Point in test.
- Create a new kraken mock for future needs
- Fill the protobuf with a new field : the famous AccessPoint
- Expose within the api the Access Point list into Stop Point
- Tested with a real ntfs file with multiple AccessPoint.

A typical response is represented like that :

accessPoint

### DIFF
--- a/source/jormungandr/jormungandr/interfaces/v1/serializer/pt.py
+++ b/source/jormungandr/jormungandr/interfaces/v1/serializer/pt.py
@@ -34,6 +34,7 @@ from jormungandr.interfaces.v1.serializer.base import (
     EnumListField,
     LiteralField,
     SortedGenericSerializer,
+    get_proto_attr_or_default,
 )
 from jormungandr.interfaces.v1.serializer.jsonschema.fields import Field, DateType
 from jormungandr.interfaces.v1.serializer.time import (
@@ -456,34 +457,34 @@ class AccessPointSerializer(PbGenericSerializer):
     parent_station = jsonschema.MethodField(schema_type=lambda: StopAreaSerializer(), display_none=False)
 
     def get_is_entrance(self, obj):
-        return obj.is_entrance if obj.HasField(str('is_entrance')) else None
+        return get_proto_attr_or_default(obj, 'is_entrance')
 
     def get_is_exit(self, obj):
-        return obj.is_exit if obj.HasField(str('is_exit')) else None
+        return get_proto_attr_or_default(obj, 'is_exit')
 
     def get_pathway_mode(self, obj):
-        return obj.pathway_mode if obj.HasField(str('pathway_mode')) else None
+        return get_proto_attr_or_default(obj, 'pathway_mode')
 
     def get_length(self, obj):
-        return obj.length if obj.HasField(str('length')) else None
+        return get_proto_attr_or_default(obj, 'length')
 
     def get_traversal_time(self, obj):
-        return obj.traversal_time if obj.HasField(str('traversal_time')) else None
+        return get_proto_attr_or_default(obj, 'traversal_time')
 
     def get_stair_count(self, obj):
-        return obj.stair_count if obj.HasField(str('stair_count')) else None
+        return get_proto_attr_or_default(obj, 'stair_count')
 
     def get_max_slope(self, obj):
-        return obj.max_slope if obj.HasField(str('max_slope')) else None
+        return get_proto_attr_or_default(obj, 'max_slope')
 
     def get_min_width(self, obj):
-        return obj.min_width if obj.HasField(str('min_width')) else None
+        return get_proto_attr_or_default(obj, 'min_width')
 
     def get_signposted_as(self, obj):
-        return obj.signposted_as if obj.HasField(str('signposted_as')) else None
+        return get_proto_attr_or_default(obj, 'signposted_as')
 
     def get_reversed_signposted_as(self, obj):
-        return obj.reversed_signposted_as if obj.HasField(str('reversed_signposted_as')) else None
+        return get_proto_attr_or_default(obj, 'reversed_signposted_as')
 
     def get_parent_station(self, obj):
         if obj.HasField(str('parent_station')):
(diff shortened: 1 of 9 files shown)


source/jormungandr/tests/access_points_tests.py
```diff
@@ -0,0 +1,98 @@
...
+# channel `#navitia` on riot https://riot.im/app/#/room/#navitia:matrix.org
+# https://groups.google.com/d/forum/navitia
+# www.navitia.io
+from __future__ import absolute_import, print_function, unicode_literals, division
+from .tests_mechanism import AbstractTestFixture, dataset
+from .check_utils import get_not_null, is_valid_places
+
+
+@dataset({"access_points_test": {}})
+class TestAccessPoints(AbstractTestFixture):
```
benoit-bst, line 42: tests are related to https://github.com/CanalTP/navitia/pull/3634/files#diff-bd5b2b6c9576adab49e34700ec199c54db96e13085979bfdde7b2fbd2ffc166dR41

source/type/pb_converter.cpp
```diff
@@ -593,6 +594,49 @@ void PbCreator::Filler::fill_pb_object(const ng::Admin* adm, pbnavitia::Administ
     }
 }
 
+void PbCreator::Filler::fill_access_points(const std::set<nt::AccessPoint*> access_points,
```
```suggestion
void PbCreator::Filler::fill_access_points(const std::set<nt::AccessPoint*> & access_points,
```

pbench (approved):

source/jormungandr/jormungandr/interfaces/v1/serializer/pt.py
```diff
@@ -434,6 +437,60 @@ def get_stop_area(self, obj):
...
+    pathway_mode = jsonschema.MethodField(schema_type=int, display_none=False)
+    length = jsonschema.MethodField(schema_type=int, display_none=False)
+    traversal_time = jsonschema.MethodField(schema_type=int, display_none=False)
+    stair_count = jsonschema.MethodField(schema_type=int, display_none=False)
+    max_slope = jsonschema.MethodField(schema_type=int, display_none=False)
+    min_width = jsonschema.MethodField(schema_type=int, display_none=False)
+    signposted_as = jsonschema.MethodField(schema_type=str, display_none=False)
+    reversed_signposted_as = jsonschema.MethodField(schema_type=str, display_none=False)
+    parent_station = jsonschema.MethodField(schema_type=lambda: StopAreaSerializer(), display_none=False)
+
```
kadhikari, line 458: Instead of using **jsonschema.MethodField**, i think you can use **jsonschema.Field** for simple attributes. Forexample : direction_type = jsonschema.Field(schema_type=str, display_none=True)

source/jormungandr/tests/access_points_tests.py
```diff
@@ -0,0 +1,101 @@
...
+            # spC
+            if sp['name'] == 'spC':
+                assert len(get_not_null(sp, 'access_points')) == 2
+                for ap in sp['access_points']:
+                    if ap['name'] == 'AP3':
+                        assert ap['is_entrance'] == True
+                        assert ap['is_exit'] == False
+                        assert ap['length'] == 12
+                        assert ap['traversal_time'] == 36
+                    if ap['name'] == 'AP1':
```
kadhikari: It shouldn't be AP2 no?

AzHicham (approved):

source/jormungandr/jormungandr/interfaces/v1/serializer/pt.py
xlqian, line 458:
```python

def get_proto_attr_or_default(obj, attr_name, default=None):
    return getattr(obj, attr_name) if obj.HasField(str(attr_name)) else default
```

source/jormungandr/jormungandr/interfaces/v1/serializer/pt.py
```diff
@@ -434,6 +437,60 @@ def get_stop_area(self, obj):
...
+    def get_is_entrance(self, obj):
+        return obj.is_entrance if obj.HasField(str('is_entrance')) else None
```
xlqian: return get_proto_attr_or_default(obj, 'is_entrance')

source/jormungandr/jormungandr/interfaces/v1/serializer/pt.py
```diff
@@ -434,6 +437,60 @@ def get_stop_area(self, obj):
...
+
+    def get_is_exit(self, obj):
+        return obj.is_exit if obj.HasField(str('is_exit')) else None
```
xlqian: return get_proto_attr_or_default(obj, 'is_exit')

source/jormungandr/jormungandr/interfaces/v1/serializer/pt.py
```diff
@@ -434,6 +437,60 @@ def get_stop_area(self, obj):
...
+
+    def get_pathway_mode(self, obj):
+        return obj.pathway_mode if obj.HasField(str('pathway_mode')) else None
```
xlqian: return get_proto_attr_or_default(obj, 'pathway_mode')

source/jormungandr/jormungandr/interfaces/v1/serializer/pt.py
```diff
@@ -434,6 +437,60 @@ def get_stop_area(self, obj):
...
+
+    def get_length(self, obj):
+        return obj.length if obj.HasField(str('length')) else None
```
xlqian: return get_proto_attr_or_default(obj, 'length')

source/type/pb_converter.cpp
xlqian: just a small remark, since your access_point are sorted by their memory address(`std::set<nt::AccessPoint*>`), the outcome maybe random, that could be problematic for your tests , tests may fail randomly

source/jormungandr/jormungandr/interfaces/v1/serializer/pt.py
xlqian: etc....

source/jormungandr/jormungandr/interfaces/v1/serializer/pt.py
benoit-bst: It don't work if we want to not display a field if it is empty...

source/jormungandr/tests/access_points_tests.py
benoit-bst: Yes exactly, it is a mistake. :1st_place_medal:

source/jormungandr/jormungandr/interfaces/v1/serializer/pt.py
benoit-bst: nice simplification Python master :heart: !

kadhikari (approved):

xlqian (approved):

PR #3636, opened 2022-01-13
[Jormun] Integrate access point into journeys

xlqian:
https://navitia.atlassian.net/wiki/spaces/NAV/pages/7100538811/Les+Entr+es+Sorties

In this PR, we're taking AccessPoint into consideration when computing journeys. 

* When `distributed` is selected, stop_points will be requested surrounding the from/to coordinates. As we are about to collect access_points from stop_points,  `depth` must be set to 3.
* When computing fallback durations(street network routing matrix), stop points are no longer the destinations/origins but access points. But the final fallback duration map is still a map of `stop_point.uri` vs `DurationElement`. Note that the `DurationElement` contains all information of by which access point the underlying stop point is attainable, the distance and the traversal time.
* When computing fallbacks, requests are now (from, access_point_entrance) and (access_point_exit, to). 
* The final response is enriched with more information about access_point.

### DIFF
--- a/source/jormungandr/jormungandr/interfaces/v1/Journeys.py
+++ b/source/jormungandr/jormungandr/interfaces/v1/Journeys.py
@@ -732,13 +732,6 @@ class Journeys(JourneyCommon):
             hidden=True,
             help="used to adjust the search range in Asgard when computing matrix",
         )
-        parser_get.add_argument(
-            "_access_points",
-            type=BooleanType(),
-            default=False,
-            hidden=True,
-            help="use/disuse the entrance/exit in journeys computation",
-        )
 
     @add_tad_links()
     @add_debug_info()
--- a/source/jormungandr/jormungandr/interfaces/v1/Journeys.py
+++ b/source/jormungandr/jormungandr/interfaces/v1/Journeys.py
@@ -732,6 +732,13 @@ class Journeys(JourneyCommon):
             hidden=True,
             help="used to adjust the search range in Asgard when computing matrix",
         )
+        parser_get.add_argument(
+            "_access_points",
+            type=BooleanType(),
+            default=False,
+            hidden=True,
+            help="used to adjust the search range in Asgard when computing matrix",
+        )
 
     @add_tad_links()
     @add_debug_info()
--- a/source/jormungandr/jormungandr/interfaces/v1/Journeys.py
+++ b/source/jormungandr/jormungandr/interfaces/v1/Journeys.py
@@ -737,7 +737,7 @@ class Journeys(JourneyCommon):
             type=BooleanType(),
             default=False,
             hidden=True,
-            help="used to adjust the search range in Asgard when computing matrix",
+            help="use/disuse the entrance/exit in journeys computation",
         )
 
     @add_tad_links()
--- a/source/jormungandr/jormungandr/interfaces/v1/serializer/pt.py
+++ b/source/jormungandr/jormungandr/interfaces/v1/serializer/pt.py
@@ -405,16 +405,16 @@ class EquipmentDetailsSerializer(PbNestedSerializer):
 
 class AccessPointSerializer(PbGenericSerializer):
     coord = CoordSerializer(required=False)
-    is_entrance = jsonschema.Field(schema_type=bool, display_none=False)
-    is_exit = jsonschema.Field(schema_type=bool, display_none=False)
-    pathway_mode = jsonschema.Field(schema_type=int, display_none=False)
-    length = jsonschema.Field(schema_type=int, display_none=False)
-    traversal_time = jsonschema.Field(schema_type=int, display_none=False)
-    stair_count = jsonschema.Field(schema_type=int, display_none=False)
-    max_slope = jsonschema.Field(schema_type=int, display_none=False)
-    min_width = jsonschema.Field(schema_type=int, display_none=False)
-    signposted_as = jsonschema.Field(schema_type=str, display_none=False)
-    reversed_signposted_as = jsonschema.Field(schema_type=str, display_none=False)
+    is_entrance = jsonschema.MethodField(schema_type=bool, display_none=False)
+    is_exit = jsonschema.MethodField(schema_type=bool, display_none=False)
+    pathway_mode = jsonschema.MethodField(schema_type=int, display_none=False)
+    length = jsonschema.MethodField(schema_type=int, display_none=False)
+    traversal_time = jsonschema.MethodField(schema_type=int, display_none=False)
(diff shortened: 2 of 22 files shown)


source/jormungandr/jormungandr/interfaces/v1/Journeys.py
```diff
@@ -732,6 +732,13 @@ def __init__(self):
             hidden=True,
             help="used to adjust the search range in Asgard when computing matrix",
         )
+        parser_get.add_argument(
```
benoit-bst: okay this is the flag to active the feature

source/jormungandr/jormungandr/interfaces/v1/serializer/pt.py
```diff
@@ -438,60 +457,6 @@ def get_stop_area(self, obj):
...
-    length = jsonschema.MethodField(schema_type=int, display_none=False)
-    traversal_time = jsonschema.MethodField(schema_type=int, display_none=False)
-    stair_count = jsonschema.MethodField(schema_type=int, display_none=False)
-    max_slope = jsonschema.MethodField(schema_type=int, display_none=False)
-    min_width = jsonschema.MethodField(schema_type=int, display_none=False)
-    signposted_as = jsonschema.MethodField(schema_type=str, display_none=False)
-    reversed_signposted_as = jsonschema.MethodField(schema_type=str, display_none=False)
-    parent_station = jsonschema.MethodField(schema_type=lambda: StopAreaSerializer(), display_none=False)
-
-    def get_is_entrance(self, obj):
```
benoit-bst: if you erase `get` function inside the AccessPointSerializer, you will have troubles with that... I tested it and if the `get` no exists we print empty string and -1 value... So this is mandatory.

source/jormungandr/jormungandr/interfaces/v1/serializer/pt.py
```diff
@@ -531,6 +496,8 @@ class PlaceSerializer(PbGenericSerializer):
     embedded_type = EnumField(attr='embedded_type', pb_type=NavitiaType, display_none=True)
     address = AddressSerializer(display_none=False)
     poi = PoiSerializer(display_none=False)
+    access_point = AccessPointSerializer(display_none=False)
```
benoit-bst, line 529: we have **AccessPoint** inside **StopArea** ?

source/jormungandr/jormungandr/utils.py
```diff
@@ -396,6 +396,7 @@ def get_pt_object_coord(pt_object):
         type_pb2.ADDRESS: "address",
         type_pb2.ADMINISTRATIVE_REGION: "administrative_region",
         type_pb2.POI: "poi",
+        type_pb2.ACCESS_POINT: "access_point",
```
benoit-bst, line 401: so, for you AccessPoint is a PtObject ? I thought it was not

source/jormungandr/jormungandr/interfaces/v1/serializer/pt.py
xlqian: OK, you're right, I didn't test that case

source/jormungandr/jormungandr/interfaces/v1/serializer/pt.py
xlqian: It is in a  `PlaceSerializer`. I serialized `access_point` as if it is a `PtObject`, that's what we want.

source/jormungandr/jormungandr/utils.py
xlqian: Since every object in `proximities_by_crowfly` (https://github.com/CanalTP/navitia/pull/3636/files#diff-77c5b35626862ae09c1bfed12bdeb812b49b47284ad3e313402ec8fc3a982bfcR170) is a `PtObject`( stop_point, car_park and access_point), converting `AccessPoint` to `PtObject` will make it easier for me to handle them instead of doing `if` every time. We also need to do this when we start to implement `AccessPoint` into `pt_ref`

source/jormungandr/jormungandr/scenarios/helper_classes/fallback_durations.py
```diff
@@ -157,12 +160,55 @@ def _do_request(self):
...
+
+                if p.uri in all_free_access:
+                    continue
+                # what we are looking to compute, is not the stop_point, but the entrance and exit of a stop_point
+                # if any of them are existent
+                if not p.stop_point.access_points:
+                    places_isochrone.append(p)
+
+                if self._direct_path_type == StreetNetworkPathType.BEGINNING_FALLBACK:
+                    for ap in (ap for ap in p.stop_point.access_points if ap.is_entrance):
```
pbench: So if `ap.is_entrance == False` for all `ap` in `p.stop_point.access_points`, then we will not compute any fallback duration to `p`.
Is it what we want to do ?

source/jormungandr/jormungandr/scenarios/helper_classes/fallback_durations.py
```diff
@@ -157,12 +160,55 @@ def _do_request(self):
...
+                                    name=ap.name,
+                                    uri=ap.uri,
+                                    embedded_type=type_pb2.ACCESS_POINT,
+                                    access_point=ap,
+                                )
+                            )
+                        access_points_map[ap.uri].append((p.stop_point.uri, ap.length, ap.traversal_time))
+
+                if self._direct_path_type == StreetNetworkPathType.ENDING_FALLBACK:
+                    for ap in (ap for ap in p.stop_point.access_points if ap.is_exit):
```
pbench: Same as https://github.com/CanalTP/navitia/pull/3636/files#r786087976

source/jormungandr/jormungandr/scenarios/helper_classes/fallback_durations.py
```diff
@@ -157,12 +160,55 @@ def _do_request(self):
...
+                            )
+                        access_points_map[ap.uri].append((p.stop_point.uri, ap.length, ap.traversal_time))
+
+        result = defaultdict(lambda: DurationElement(float('inf'), None, None, 0, None))
         # Since we have already places that have free access, we add them into the result
-        [result.update({uri: DurationElement(0, response_pb2.reached, None, 0)}) for uri in all_free_access]
+        [
+            result.update({uri: DurationElement(0, response_pb2.reached, None, 0, None)})
+            for uri in all_free_access
+        ]
```
pbench, line 217: This creates a list but do nothing with it ??

xlqian: it's a pythonic way to do for_each :)

PR #3646, opened 2022-01-24
Access points entry point

benoit-bst:
I am a little bit sorry because the PR is just huge... I mixed the **Kraken** and the **Jormungandr** part.
Some details to understand more easly
- Into sources, we have a complete folder with all the _access_points_ kraken part (sources/access_point)
- **navitia-proto** is updated the req and resp https://github.com/CanalTP/navitia-proto/pull/174
- the jormun part to expose the entry point -> https://github.com/CanalTP/navitia/compare/dev...benoit-bst:access_points_entry_point?expand=1#diff-8e8d1a1627b9a84ca31f69d538e5c5ea45941cf946eb927f4111979fef4cdbe8R44
- **Kraken** and **Jormun** parts are tested and a lot of tests by hand was done

_TODO_ : Had a piece of documentation into slate

At the end of this work the API can expose Access Point :tulip: 

ap_gare_de_lyon

### DIFF
--- a/source/access_point/access_point_api.cpp
+++ b/source/access_point/access_point_api.cpp
@@ -63,8 +63,8 @@ void access_points(PbCreator& pb_creator,
         const auto stop_points = data.get_data<type::StopPoint>(stop_point_indexes);
 
         // make Acess Point unique in to the response
-        for (auto& _sp : stop_points) {
-            for (auto& ap : _sp->access_points) {
+        for (const auto& _sp : stop_points) {
+            for (const auto& ap : _sp->access_points) {
                 auto it = ap_list.find(ap.name);
                 if (it == ap_list.end()) {
                     ap_list[ap.name] = ap;
--- a/source/access_point/access_point_api.cpp
+++ b/source/access_point/access_point_api.cpp
@@ -61,6 +61,8 @@ void access_points(PbCreator& pb_creator,
         const type::Indexes stop_point_indexes =
             ptref::make_query(type::Type_e::StopPoint, filter, forbidden_uris, data);
         const auto stop_points = data.get_data<type::StopPoint>(stop_point_indexes);
+
+        // make Acess Point unique in to the response
         for (auto& _sp : stop_points) {
             for (auto& ap : _sp->access_points) {
                 auto it = ap_list.find(ap.name);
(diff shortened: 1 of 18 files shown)


pbench (approved):

kadhikari (approved):

source/access_point/access_point_api.cpp
```diff
@@ -0,0 +1,90 @@
...
+        const type::Indexes stop_point_indexes =
+            ptref::make_query(type::Type_e::StopPoint, filter, forbidden_uris, data);
+        const auto stop_points = data.get_data<type::StopPoint>(stop_point_indexes);
+
+        // make Acess Point unique in to the response
+        for (auto& _sp : stop_points) {
+            for (auto& ap : _sp->access_points) {
+                auto it = ap_list.find(ap.name);
+                if (it == ap_list.end()) {
+                    ap_list[ap.name] = ap;
```
xlqian, line 70: Hmmm, my question would be functional... 

since the access_point is stored at the level of stop_point, as a consequence, an access_point may be stored multiple times in different stop_point, yet, every access_point actually stores the information regarding the stop_point in which it's attached to(ex.  length, traversal_time).

If I understand your code well, you are gonna display the first access_point you find in the list along with that information... could that be misleading/confusing to users?

source/access_point/access_point_api.cpp
```diff
@@ -0,0 +1,90 @@
...
+                   const ForbiddenUris& forbidden_uris) {
+    const type::Data& data = *pb_creator.data;
+    AccessPointList ap_list;
+    try {
+        const type::Indexes stop_point_indexes =
+            ptref::make_query(type::Type_e::StopPoint, filter, forbidden_uris, data);
+        const auto stop_points = data.get_data<type::StopPoint>(stop_point_indexes);
+
+        // make Acess Point unique in to the response
+        for (auto& _sp : stop_points) {
```
xlqian:
```suggestion
        for (const auto& _sp : stop_points) {
```

source/access_point/access_point_api.cpp
```diff
@@ -0,0 +1,90 @@
...
+            for (auto& ap : _sp->access_points) {
```
xlqian:
```suggestion
            for (const auto& ap : _sp->access_points) {
```

xlqian (approved):

source/access_point/access_point_api.cpp
benoit-bst: yes an **Access Point** can be stored in several SP. This is the reason why, when I want to expose the AP list, I need to check if the AP already exist. 
So, I used an _unordered map_ and check if key already exist :screwdriver:

PR #3652, opened 2022-01-27
Manage database resilience if at least one table is not accessible

kadhikari:
This feature is related to the failure while deploying navitia in PROD last Monday (24/01/2022).
* Navitia was not accessible during 2 hours as the table billing_plan was locked.
* This correction add more verification and use the already existing mechanism **can_connect_to_database()**.
* The database component (SQLAlchemy) is configured to throw a statement_timeout after 1000 ms (with a default configuration).
* Need to wait for maximum of 2 timeouts (one per navitia query) for each cache period.
* After the first time_out catch,  **can_connect_to_database()** mechanism is used and user is authorized without verification to continue query navitia.

Doc: https://navitia.atlassian.net/wiki/spaces/NAV/pages/11942920554/20220122+-+Incident+de+prod+pendant+la+livraison+de+navitia+v14

**Note: Difficult to add test. I request at least someone to test in his/her machine.**

### DIFF
--- a/source/jormungandr/jormungandr/authentication.py
+++ b/source/jormungandr/jormungandr/authentication.py
@@ -116,7 +116,7 @@ def get_token():
 
 def can_read_user():
     try:
-        User.can_read()
+        User.query.first()
     except Exception as e:
         logging.getLogger(__name__).error('No access to table User (error: {})'.format(e))
         g.can_connect_to_database = False
@@ -135,10 +135,8 @@ def has_access(region, api, abort, user):
     Warning: Please this function is cached therefore it should not be
     dependent of the request context, so keep it as a pure function.
     """
-    if current_app.config.get('PUBLIC', False):
-        # if jormungandr is on public mode we skip the authentification process
-        return True
-    if not can_connect_to_database():
+    # if jormungandr is on public mode or database is not accessible, we skip the authentication process
+    if current_app.config.get('PUBLIC', False) or (not can_connect_to_database()):
         return True
 
     if not user:
@@ -147,8 +145,7 @@ def has_access(region, api, abort, user):
         # Can connect to database but at least one table/attribute is not accessible due to transaction problem
         if can_read_user():
             abort_request(user=user)
-
-        if not can_connect_to_database():
+        else:
             return True
     try:
         model_instance = Instance.get_by_name(region)
@@ -233,8 +230,7 @@ def get_all_available_instances(user):
         # Can connect to database but at least one table/attribute is not accessible due to transaction problem
         if can_read_user():
             abort_request(user=user)
-
-        if not can_connect_to_database():
+        else:
             return None
 
     if not user.have_access_to_free_instances:
--- a/source/jormungandr/jormungandr/authentication.py
+++ b/source/jormungandr/jormungandr/authentication.py
@@ -114,6 +114,16 @@ def get_token():
         return auth
 
 
+def can_read_user():
+    try:
+        User.can_read()
+    except Exception as e:
+        logging.getLogger(__name__).error('No access to table User (error: {})'.format(e))
+        g.can_connect_to_database = False
+        return False
+    return True
+
+
 @memory_cache.memoize(
     current_app.config[str('MEMORY_CACHE_CONFIGURATION')].get(str('TIMEOUT_AUTHENTICATION'), 30)
 )
@@ -133,9 +143,19 @@ def has_access(region, api, abort, user):
 
     if not user:
         # no user --> no need to continue, we can abort, a user is mandatory even for free region
-        abort_request(user=user)
+        # To manage database error of the following type we should fetch one more time from database
+        # Can connect to database but at least one table/attribute is not accessible due to transaction problem
+        if can_read_user():
+            abort_request(user=user)
 
-    model_instance = Instance.get_by_name(region)
+        if not can_connect_to_database():
+            return True
+    try:
+        model_instance = Instance.get_by_name(region)
+    except Exception as e:
+        logging.getLogger(__name__).error('No access to table Instance (error: {})'.format(e))
+        g.can_connect_to_database = False
+        return True
 
     if not model_instance:
         if abort:
@@ -164,7 +184,14 @@ def cache_get_user(token):
     """
     if not can_connect_to_database():
         return None
-    return User.get_from_token(token, datetime.datetime.now())
+    try:
+        user = User.get_from_token(token, datetime.datetime.now())
+    except Exception as e:
+        logging.getLogger(__name__).error('No access to table User (error: {})'.format(e))
+        g.can_connect_to_database = False
+        return None
+
+    return user
 
 
 @memory_cache.memoize(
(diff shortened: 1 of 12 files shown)


source/jormungandr/jormungandr/interfaces/v1/Places.py
```diff
@@ -199,6 +199,11 @@ def get(self, region=None, lon=None, lat=None):
             response = i_manager.dispatch(args, "places", instance_name=self.region)
         else:
             available_instances = get_all_available_instances(user)
+
+            # If no instance available most probably due to database error
```
benoit-bst, line 203: Comment is not synch

source/jormungandr/jormungandr/authentication.py
```diff
@@ -133,9 +143,19 @@ def has_access(region, api, abort, user):
...
     if not user:
         # no user --> no need to continue, we can abort, a user is mandatory even for free region
-        abort_request(user=user)
+        # To manage database error of the following type we should fetch one more time from database
+        # Can connect to database but at least one table/attribute is not accessible due to transaction problem
+        if can_read_user():
+            abort_request(user=user)
 
-    model_instance = Instance.get_by_name(region)
+        if not can_connect_to_database():
```
benoit-bst: can_connect_to_database if a g var not a function ?

benoit-bst (commented):
I see fix for autocompletion. It is normal ?

source/jormungandr/jormungandr/authentication.py
kadhikari: It's a function also: https://github.com/CanalTP/navitia/blob/4ac0c4e7d19acfcd700b4dde2ac35d7dce02310b/source/jormungandr/jormungandr/utils.py#L679

source/jormungandr/jormungandr/interfaces/v1/Places.py
kadhikari: What is your proposal ?

xlqian:
`Need to wait for maximum of 2 timeouts (one per navitia query) for each cache period.` -> can you please elaborate on why `2 timeouts`?

source/jormungandr/jormungandr/authentication.py
```diff
@@ -114,6 +114,16 @@ def get_token():
...
 
 
+def can_read_user():
+    try:
+        User.can_read()
+    except Exception as e:
+        logging.getLogger(__name__).error('No access to table User (error: {})'.format(e))
+        g.can_connect_to_database = False
+        return False
+    return True
```
xlqian:
```suggestion
    try:
        User.query.first()
        # Why not true here?
        g.can_connect_to_database = True
        return True
    except Exception as e:
        logging.getLogger(__name__).error('No access to table User (error: {})'.format(e))
        g.can_connect_to_database = False
        return False
```

source/jormungandr/jormungandr/authentication.py
```diff
@@ -164,7 +184,14 @@ def cache_get_user(token):
     """
     if not can_connect_to_database():
         return None
-    return User.get_from_token(token, datetime.datetime.now())
+    try:
+        user = User.get_from_token(token, datetime.datetime.now())
+    except Exception as e:
+        logging.getLogger(__name__).error('No access to table User (error: {})'.format(e))
+        g.can_connect_to_database = False
+        return None
```
xlqian, line 189: Can you return a tuple (User, DbState)? So later, you don't have to test db's accessibility again.. https://github.com/CanalTP/navitia/pull/3652/files#diff-a7edc0c6cea6718d3c64b2d8a71e2848ec2ef57653932be2bbbee3e4179d8464R148

source/jormungandr/jormungandr/authentication.py
```diff
@@ -196,7 +229,14 @@ def get_all_available_instances(user):
...
     if not user:
         # for not-public navitia a user is mandatory
-        abort_request(user=user)
+        # To manage database error of the following type we should fetch one more time from database
+        # Can connect to database but at least one table/attribute is not accessible due to transaction problem
+        if can_read_user():
+            abort_request(user=user)
+
+        if not can_connect_to_database():
+            return None
```
xlqian: this piece of code is repeated a lot of times... Actually,  if you think about it, connectivity has become a preliminary condition for certain functions... you can make a decorator encapsulating that logic...

source/jormungandr/jormungandr/authentication.py
kadhikari: As discussed, it will be very difficult to reuse db state for the following access to database.

source/jormungandr/jormungandr/authentication.py
kadhikari: User.query.first() -> Done
g.can_connect_to_database = True -> There is no use of this positive value since the value will be valid for only one query to navitia..

source/jormungandr/jormungandr/authentication.py
kadhikari: A bit simplified

source/jormungandr/jormungandr/authentication.py
kadhikari: As discussed, it will be difficult to reuse the db state.

source/jormungandr/jormungandr/default_settings.py
```diff
@@ -5,6 +5,8 @@
 import json
 from flask_restful.inputs import boolean
 
+DEFAULT_SQLALCHEMY_ENGINE_OPTIONS = {'connect_args': {"options": "-c statement_timeout=1000"}}  # 1000ms
+
```
pbench:
```suggestion
# sql queries will return an exception if the query did not succeed before `statement_timeout`
DEFAULT_SQLALCHEMY_ENGINE_OPTIONS = {'connect_args': {"options": "-c statement_timeout=1000"}}  # 1000ms

```

source/jormungandr/jormungandr/external_services/external_service_provider_manager.py
```diff
@@ -118,11 +118,12 @@ def update_config(self):
...
         self._last_update = datetime.datetime.utcnow()
 
-        services = []
         try:
             services = self._external_service_getter()
-        except Exception:
-            self.logger.exception('Failure to retrieve external service configuration')
+        except Exception as e:
+            self.logger.error('No access to table external_service (error: {})'.format(e))
+            return
```
pbench:
```suggestion
            self.logger.error('No access to table external_service (error: {})'.format(e))
            # database is not accessible, so let's use the values already present in self._external_service_getter
            return
```

source/jormungandr/jormungandr/default_settings.py
kadhikari: :+1:

source/jormungandr/jormungandr/external_services/external_service_provider_manager.py
kadhikari: :+1:

pbench (approved):

azime (approved):

azime (approved):

PR #3655, opened 2022-01-28
[Building Packages] Skip tests when building packages

xlqian:
Skip all tests when building packages, except for main_routing_test and departure_board_test


source/CMakeLists.txt
```diff
@@ -83,9 +83,18 @@ add_subdirectory(equipment)
 add_subdirectory(position)
 add_subdirectory(access_point)
 
-# Add tests
 if(NOT SKIP_TESTS)
+    # Add tests
     add_subdirectory(tests)
+else ()
```
benoit-bst, line 89: you can do the same at a lower level but nevermind

benoit-bst (approved):

PR #3663, opened 2022-02-09
Early filtering of vj for stop point disruption

pbench:
First check if the vj goes through the deleted stop_point, before doing anything with application periods.

### DIFF
--- a/source/kraken/apply_disruption.cpp
+++ b/source/kraken/apply_disruption.cpp
@@ -552,9 +552,13 @@ struct add_impacts_visitor : public apply_impacts_visitor {
 
             LOG4CPLUS_TRACE(log, "VJ: " << vj->uri << " may be impacted");
 
-            bool stop_point_found = boost::range::find_if(vj->stop_time_list, [](const auto& stop_time) {
-                return stop_time.stop_point->uri == stop_point->uri; 
-            } );
+            bool stop_point_found = false;
+            for (const nt::StopTime& stop_time : vj->stop_time_list) {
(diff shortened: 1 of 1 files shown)


source/kraken/apply_disruption.cpp
```diff
@@ -552,6 +552,17 @@ struct add_impacts_visitor : public apply_impacts_visitor {
...
+            for (const nt::StopTime& stop_time : vj->stop_time_list) {
+                if (stop_time.stop_point->uri == stop_point->uri) {
+                    stop_point_found = true;
+                    break;
+                }
+            }
+            if (!stop_point_found) {
+                continue;
+            }
+
```
xlqian, line 565:
```suggestion
            bool stop_point_found = boost::range::find_if(vj->stop_time_list, [](const auto& stop_time) {
                return stop_time.stop_point->uri == stop_point->uri; 
            } );
            if (!stop_point_found) {
                continue;
            }

```

xlqian (approved):

kadhikari:
As i have a chaos database backup from PROD, i will check if there is any difference of disruptions in navitia between two versions.

azime (approved):

benoit-bst (approved):

kadhikari (approved):
I verified with @SGrenet realtime stop_times in route_schedules for metro 4 and 14.
I have the same result between idfm in prod and my post.
For me fine to merge.

**Loading chaos database duration for idfm prod reduced from 00:09:07 to 00:01:43**

PR #3668, opened 2022-02-21
[Jormungandr] Add access_points in db

xlqian:
Add access points in db

### DIFF
--- a/source/tyr/tests/integration/instance_test.py
+++ b/source/tyr/tests/integration/instance_test.py
@@ -160,6 +160,7 @@ def test_update_instances(create_instance):
         "stop_points_nearby_duration": 888,
         "places_proximity_radius": 10000,
         "transfer_path": True,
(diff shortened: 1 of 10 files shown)


pbench (approved):

source/tyr/tests/integration/instance_test.py
```diff
@@ -193,6 +194,7 @@ def test_update_instances(create_instance):
     assert resp['max_car_no_park_direct_path_distance'] == 16000
     assert resp['places_proximity_radius'] == 10000
     assert resp['transfer_path'] is True
+    assert resp['access_points'] is True
```
kadhikari, line 199: Minor: Another test with     assert resp['access_points'] is False will be preferable

kadhikari (approved):

benoit-bst (approved):

kadhikari (approved):

PR #3679, opened 2022-03-16
Here :  base duration for base datetime

benoit-bst:
### targets :dart: 

- Add the base date time feature with Here connector
- Fix the bug on base date time drift (I think it was an oblivion)
- clockwise is OKAY (departure/arrival)

Now the response is like this :

base_date

This is the first section.
for a duration - base duration = 7 min 23 secs. So, we can start the journey 7 min 23 secs later for a _"Base Journey"_

### DIFF
--- a/source/jormungandr/jormungandr/scenarios/helper_classes/helper_utils.py
+++ b/source/jormungandr/jormungandr/scenarios/helper_classes/helper_utils.py
@@ -120,6 +120,9 @@ def _align_fallback_direct_path_datetime(fallback_direct_path, fallback_extremit
         for s in journey.sections:
             s.begin_date_time += delta
             s.end_date_time += delta
+            if s.base_begin_date_time != None and s.base_end_date_time:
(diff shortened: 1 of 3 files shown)


stifoon (approved):

woshilapin (approved):

source/jormungandr/jormungandr/scenarios/helper_classes/helper_utils.py
```diff
@@ -120,6 +120,9 @@ def _align_fallback_direct_path_datetime(fallback_direct_path, fallback_extremit
         for s in journey.sections:
             s.begin_date_time += delta
             s.end_date_time += delta
+            if s.base_begin_date_time != None and s.base_end_date_time:
```
benoit-bst, line 123: The culprit for the diff, during the test.
Base date time needs to be drifted like classic date time
This fix a global bug

pbench (approved):

PR #3682, opened 2022-03-21
Robust tag

pbench:
Tag a journey as "robust" if all the following are true : 
  - fallbacks are either absent, walk or bike
  - the journey has at least one public transport section
  - all public transport sections use a robust physical mode 

robust physical modes are :
```
    "physical_mode:RapidTransit",
    "physical_mode:Metro",
    "physical_mode:Train",
    "physical_mode:RailShuttle",
    "physical_mode:LocalTrain",
    "physical_mode:LongDistanceTrain",
```

### DIFF
--- a/source/jormungandr/jormungandr/scenarios/new_default.py
+++ b/source/jormungandr/jormungandr/scenarios/new_default.py
@@ -753,18 +753,17 @@ robust_fallback_modes = [response_pb2.Bike, response_pb2.Walking]
 #  - all public transport sections use robust physical modes
 #  - start and end fallbacks use robust fallback mode
 def is_robust_journey(journey):
-    found_a_robust_mode = False
-    found_a_non_robust_fallback = False
+    found_a_pt_section_with_robust_mode = False
     for section in journey.sections:
 
-        # check if this section uses a non robust fallback mode
+        # if this section uses a non robust fallback mode
+        # the journey is not robust
         if section.HasField("street_network"):
             street_network = section.street_network
             if street_network.HasField("mode"):
                 mode = street_network.mode
                 if mode not in robust_fallback_modes:
-                    found_a_non_robust_fallback = True
-                    break
+                    return False
 
         if section.type != response_pb2.PUBLIC_TRANSPORT:
             continue
@@ -777,7 +776,7 @@ def is_robust_journey(journey):
             found_a_robust_mode = True
         else:
             return False
-    return found_a_robust_mode and not found_a_non_robust_fallback
+    return found_a_pt_section_with_robust_mode
 
 
 def type_journeys(resp, req):
--- a/source/jormungandr/jormungandr/scenarios/new_default.py
+++ b/source/jormungandr/jormungandr/scenarios/new_default.py
@@ -732,6 +732,32 @@ def _tag_by_mode(responses):
             _tag_journey_by_mode(j)
 
 
+def tag_robust_journeys(responses):
+    for r in responses:
+        for j in r.journeys:
+            if is_robust_journey(j):
+                j.tags.append("robust")
+
+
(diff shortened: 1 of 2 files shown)


source/jormungandr/jormungandr/scenarios/new_default.py
```diff
@@ -732,6 +732,54 @@ def _tag_by_mode(responses):
             _tag_journey_by_mode(j)
 
 
+def tag_robust_journeys(responses):
```
woshilapin, line 735: Should this private like the other tagging function?
```suggestion
def _tag_robust_journeys(responses):
```

That said, I see that `is_robust_journey` probably can't be private because of the tests... :thinking:

AzHicham (approved):

source/jormungandr/jormungandr/scenarios/new_default.py
```diff
@@ -732,6 +732,54 @@ def _tag_by_mode(responses):
...
+    for section in journey.sections:
+
+        # check if this section uses a non robust fallback mode
+        if section.HasField("street_network"):
+            street_network = section.street_network
+            if street_network.HasField("mode"):
+                mode = street_network.mode
+                if mode not in robust_fallback_modes:
+                    found_a_non_robust_fallback = True
+                    break
```
woshilapin: How about doing `return False` from here and not use the `found_a_non_robust_fallback` at all?

source/jormungandr/jormungandr/scenarios/new_default.py
```diff
@@ -732,6 +732,54 @@ def _tag_by_mode(responses):
...
+        if not section.HasField("uris"):
+            continue
+        uris = section.uris
+        if not uris.HasField("physical_mode"):
+            continue
+        if uris.physical_mode in robust_physical_modes:
+            found_a_robust_mode = True
+        else:
+            return False
+    return found_a_robust_mode and not found_a_non_robust_fallback
```
woshilapin: It feels like all of this could be written with early return in mind: as soon as you find a non-robust section, `return False`. Then if all the sections have been checked for a non-robust condition, then you can `return True` (see also the above comment about returning early for fallback modes).

```suggestion
        if uris.physical_mode not in robust_physical_modes:
            return False
    return True
```

source/jormungandr/jormungandr/scenarios/tests/qualifier_tests.py
```diff
@@ -242,3 +243,57 @@ def qualifier_nontransport_duration_with_tc_test():
...
+    journey.sections[0].street_network.mode = response_pb2.Bike
+
+    # no public transport in the journey, so it should not be tagged as robust
+    assert is_robust_journey(journey) == False
+
+    journey.sections.add()
+    journey.sections[1].type = response_pb2.PUBLIC_TRANSPORT
+    journey.sections[1].uris.physical_mode = "physical_mode:Bus"
+
+    assert is_robust_journey(journey) == False
```
woshilapin, line 286: Is this test useful considering you already test physical modes in the previous test `robust_journey_test`?

source/jormungandr/jormungandr/scenarios/new_default.py
pbench: because then a direct path would be "robust" and we want a journey to be "robust" only if it contains a public transport section

pbench: hum no, you're right, I can early return here :+1:

source/jormungandr/jormungandr/scenarios/new_default.py
woshilapin: As discussed, it's not enough since you also need to detect there is at least one section with a robust public transport... so you cannot do it without `found_a_robust_mode`. I'm resolving this comment.

woshilapin (approved):

PR #3686, opened 2022-03-22
[Distributed] feat: collage from_coord and to_coord to geojson

xlqian:
In this PR, the coordinates of `from` and `to` are add to the beginning and the end of the geojson when computing fallback and direct_path. So users can identify `from` and `to` in geojson more easily, plus it makes the geojson smoother when displaying

before:
image

after:
image

### DIFF
--- a/source/jormungandr/jormungandr/scenarios/helper_classes/streetnetwork_path.py
+++ b/source/jormungandr/jormungandr/scenarios/helper_classes/streetnetwork_path.py
@@ -96,7 +96,7 @@ class StreetNetworkPath:
                 logging.getLogger(__name__).exception('')
                 return None
 
-    def _replace_first_and_last_coord(self, dp):
+    def _add_first_and_last_coord(self, dp):
         if not dp or not dp.journeys or not dp.journeys[0].sections:
             return
 
@@ -132,7 +132,7 @@ class StreetNetworkPath:
 
         dp = self._direct_path_with_fp(self._streetnetwork_service)
 
-        self._replace_first_and_last_coord(dp)
+        self._add_first_and_last_coord(dp)
 
         if getattr(dp, "journeys", None):
             dp.journeys[0].internal_id = str(utils.generate_id())
--- a/source/jormungandr/jormungandr/scenarios/helper_classes/streetnetwork_path.py
+++ b/source/jormungandr/jormungandr/scenarios/helper_classes/streetnetwork_path.py
@@ -104,7 +104,7 @@ class StreetNetworkPath:
         if len(dp_coords) < 2:
             return
 
-        # we are insert the coord of the origin at the beginning of the geojson
+        # we are inserting the coord of the origin at the beginning of the geojson
         coord = utils.get_pt_object_coord(self._orig_obj)
         if coord != dp_coords[0]:
             dp_coords.add(lon=coord.lon, lat=coord.lat)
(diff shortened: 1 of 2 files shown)


source/jormungandr/jormungandr/scenarios/helper_classes/streetnetwork_path.py
```diff
@@ -95,6 +96,31 @@ def _direct_path_with_fp(self):
...
 
+    def _add_first_and_last_coord(self, dp):
+        if not dp or not dp.journeys or not dp.journeys[0].sections:
+            return
+
+        dp_coords = dp.journeys[0].sections[0].street_network.coordinates
+        if len(dp_coords) < 2:
+            return
+
+        # we are insert the coord of the origin at the beginning of the geojson
```
xlqian:
```suggestion
        # we are inserting the coord of the origin at the beginning of the geojson
```

source/jormungandr/jormungandr/scenarios/helper_classes/streetnetwork_path.py
```diff
@@ -95,6 +96,31 @@ def _direct_path_with_fp(self):
...
+            dp_coords.add(lon=coord.lon, lat=coord.lat)
+            # we cannot insert an element at the beginning of a list :(
+            # a little algo to move the last element to the beginning
+            tmp = type_pb2.GeographicalCoord()
+            for i in range(len(dp_coords)):
+                tmp.CopyFrom(dp_coords[i])
+                dp_coords[i].CopyFrom(dp_coords[-1])
+                dp_coords[-1].CopyFrom(tmp)
+
+        # we are append the coord of the destination at the end of the geojson
```
xlqian:
```suggestion
        # we are appending the coord of the destination at the end of the geojson
```

pbench (approved):

source/jormungandr/jormungandr/scenarios/helper_classes/streetnetwork_path.py
```diff
@@ -95,6 +96,31 @@ def _direct_path_with_fp(self):
...
+        coord = utils.get_pt_object_coord(self._orig_obj)
+        if coord != dp_coords[0]:
+            dp_coords.add(lon=coord.lon, lat=coord.lat)
+            # we cannot insert an element at the beginning of a list :(
+            # a little algo to move the last element to the beginning
+            tmp = type_pb2.GeographicalCoord()
+            for i in range(len(dp_coords)):
+                tmp.CopyFrom(dp_coords[i])
+                dp_coords[i].CopyFrom(dp_coords[-1])
+                dp_coords[-1].CopyFrom(tmp)
```
woshilapin, line 117: How about something like this instead?

```python
dp_coords = [coord].extend(dp_coords)
```

I don't know how performant it should be though because your solution probably doesn't reallocate a list whereas mine probably does.

woshilapin (approved):

source/jormungandr/jormungandr/scenarios/helper_classes/streetnetwork_path.py
xlqian: unfortunately, `dp_coords` is a proto object, which means the objects can only be handled through its own API. An assignment that you suggest is not allowed and insertion at the begging is not allowed either :(

PR #3696, opened 2022-03-29
[Jormungandr] fix: street informations

xlqian:
https://navitia.atlassian.net/browse/NAV-982

### DIFF
--- a/source/jormungandr/jormungandr/interfaces/v1/serializer/journey.py
+++ b/source/jormungandr/jormungandr/interfaces/v1/serializer/journey.py
@@ -65,6 +65,7 @@ from navitiacommon.response_pb2 import (
     SectionType,
     CyclePathType,
 )
+import navitiacommon.response_pb2
 from navitiacommon.type_pb2 import RTLevel
 from jormungandr.interfaces.v1.make_links import create_internal_link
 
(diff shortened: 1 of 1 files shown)


benoit-bst (approved):

source/jormungandr/jormungandr/interfaces/v1/serializer/journey.py
```diff
@@ -205,13 +206,21 @@ def get_via_uri(self, obj):
         return obj.via_uri if obj.HasField(str('via_uri')) else None
 
 
+CYCLEPATHTYPE_TO_STR = {
+    navitiacommon.response_pb2.NoCycleLane: "no_cycle_lane",
+    navitiacommon.response_pb2.SharedCycleWay: "shared_cycle_way",
+    navitiacommon.response_pb2.DedicatedCycleWay: "dedicated_cycle_way",
+    navitiacommon.response_pb2.SeparatedCycleWay: "separated_cycle_way",
+}
```
woshilapin, line 214: This means any evolution of the protobuf with a new variant of the enum, we will also need an evolution of code. Isn't there a way to do something different like `CyclePathType.Name(obj.cycle_path_type).to_snake_case()` instead? It seems some libraries can do just that like [`stringcase`](https://pypi.org/project/stringcase/) ou [`inflection`](https://pypi.org/project/inflection/) (and maybe others?).

xlqian: In our case, the enum's names just happen to be the same as what's to be displayed,  IMO, the enum is an object on the technic level, which means the enum's names are not necessarily the texts to be displayed... if we have to add/change the text to be displayed, so be it. :)

woshilapin (approved):

SGrenet (approved):

PR #3697, opened 2022-03-29
[Jormungandr] fix: access point in journey

xlqian:
https://navitia.atlassian.net/wiki/spaces/NAV/pages/7100538811/Les+Entr+es+Sorties

image

image

### DIFF
--- a/source/jormungandr/jormungandr/interfaces/v1/serializer/pt.py
+++ b/source/jormungandr/jormungandr/interfaces/v1/serializer/pt.py
@@ -405,18 +405,24 @@ class EquipmentDetailsSerializer(PbNestedSerializer):
 
 class AccessPointSerializer(PbGenericSerializer):
     coord = CoordSerializer(required=False)
+    access_point_code = jsonschema.MethodField(schema_type=str, display_none=False)
+
+    def get_access_point_code(self, obj):
+        return get_proto_attr_or_default(obj, 'stop_code')
+
+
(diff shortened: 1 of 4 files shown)


woshilapin (approved):

kadhikari (approved):

source/jormungandr/jormungandr/interfaces/v1/serializer/pt.py
```diff
@@ -826,3 +779,17 @@ class StopDateTimeSerializer(PbNestedSerializer):
...
+
+class PathWaySerializer(PbNestedSerializer):
+    access_point = jsonschema.MethodField(schema_type=AccessPointSerializer(), display_none=False)
+    is_entrance = BoolField()
+    is_exit = BoolField()
+    length = IntField()
+    traversal_time = IntField()
+    pathway_mode = IntField()
+    name = StrField()
+    id = jsonschema.Field(schema_type=str, display_none=True, attr='uri')
```
azime: To remove, it's filled in `PbGenericSerializer` class

xlqian: I agree with you, but in the spec we want that id...  I just respect the spec just no brainer... :man_shrugging:

benoit-bst (approved):
So know, where is stair_count option and others ?

pbench (approved):

PR #3698, opened 2022-03-30
[Distributed] feat: ability to use loki as crowfly service

xlqian:
`get_crowfly` in distributed is based on `kraken` exclusively. With the PR, `get_crowfly` can be switched to `loki`. 

Note that, `get_crowfly` based on `loki` can only do `stop_points`(with `access_points` which depends on the params).

### DIFF
--- a/source/jormungandr/jormungandr/pt_planners/common.py
+++ b/source/jormungandr/jormungandr/pt_planners/common.py
@@ -42,7 +42,7 @@ from abc import abstractmethod, ABCMeta
 
 from jormungandr import app
 from jormungandr.exceptions import DeadSocketException
-from navitiacommon import response_pb2
+from navitiacommon import response_pb2, request_pb2, type_pb2
 
 
 class ZmqSocket(six.with_metaclass(ABCMeta, object)):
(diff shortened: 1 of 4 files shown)


source/jormungandr/jormungandr/pt_planners/common.py
```diff
@@ -115,3 +115,57 @@ def send_and_receive(self, *args, **kwargs):
...
+    req.places_nearby.count = max_nb_crowfly
+    req.places_nearby.start_page = 0
+    req.disable_feedpublisher = True
+    req.places_nearby.types.append(object_type)
+
+    allowed_id_filter = ''
+    if allowed_id is not None:
+        allowed_id_count = len(allowed_id)
+        if allowed_id_count > 0:
+            poi_ids = ('poi.id={}'.format(uri) for uri in allowed_id)
```
pbench, line 152: these filter will no work with loki, right ?

xlqian: yep, loki cannot handle that kind of request

source/jormungandr/jormungandr/pt_planners/loki.py
```diff
@@ -59,3 +60,34 @@ def graphical_isochrones(
...
+        object_type=type_pb2.STOP_POINT,
+        filter=None,
+        stop_points_nearby_duration=300,
+        request_id=None,
+        depth=2,
+        forbidden_uris=[],
+        allowed_id=[],
+        **kwargs
+    ):
+        return get_crow_fly(
```
xlqian, line 89: add exceptions

pbench (approved):

PR #3708, opened 2022-04-06
[Jormun] Load zmq_socket config from database

AzHicham:


xlqian:
Be very cautious before merging the PR, with the current modification, you're gonna load configuration from DB even though local configuration files are present and **configurations from files will be overwritten**. 

Given that the default value of zmq_socket `after the migration` is "tcp://localhost:30000", we're gonna make every kraken in production listen to that port! Let's talk about it and find the best solution before merging the PR

PR #3719, opened 2022-04-11
[CI] Workflow/Artemis optimization

AzHicham:


woshilapin (approved):
I'm wondering if [this](https://docs.github.com/en/actions/using-workflows/reusing-workflows#creating-a-reusable-workflow) could be used to avoid duplication of code?

AzHicham:
> I'm wondering if [this](https://docs.github.com/en/actions/using-workflows/reusing-workflows#creating-a-reusable-workflow) could be used to avoid duplication of code?

For sure we can, but as I wanted to move fast to test Artemis with Debian 10. lets schedule it for another time ? ^^

PR #3720, opened 2022-04-11
docs: update some comment and documentation to Hove

woshilapin:
Some more small corrections about moving to `hove-io`. Only docs and comments in this PR.


source/ed/connectors/fusio_parser.h
```diff
@@ -32,7 +32,7 @@ www.navitia.io
 #include "gtfs_parser.h"
 
 /**
- * Read CanalTP custom transportation files
+ * Read Hove custom transportation files
```
stifoon, line 35: Maybe adding a link to https://github.com/hove-io/ntfs-specification would be pertinent :)

stifoon (approved):

source/ed/connectors/fusio_parser.h
woshilapin: Good idea. Done in 2c6b89133.

PR #3735, opened 2022-04-26
[CI] Change docker tag for release

AzHicham:
tag latest -> latest on release branch
tag dev -> latest on dev branch


woshilapin (approved):
But does it mean we won't have a `release` tag anymore? Was it used?

AzHicham:
We won't have it anymore & Nope it was not used :)

PR #3752, opened 2022-05-20
[Jormun] add alternatives asgard

xlqian:
https://navitia.atlassian.net/wiki/spaces/NAV/pages/11963105672/Valhalla+-+Alternatives+v+lo

Handle alternatives in Asgard. This feature is only available for bike mode and only activated when `direct_path=only_with_alternatives` is set.

The idea is to send a list of parameters instead of just one. Asgard is in charge of giving back different journeys regarding the received list of parameters from Jormungandr.

In order to tag the returned journeys correctly, every parameter to be sent is associated with a user tag (ex comfort, shortest, balanced... )  that the parameters are adjusted to represent.  
 
Asgard should be able to handle this feature once https://github.com/hove-io/asgard/pull/197/files is merged.

### DIFF
--- a/source/jormungandr/jormungandr/street_network/asgard.py
+++ b/source/jormungandr/jormungandr/street_network/asgard.py
@@ -85,22 +85,22 @@ class Languages(Enum):
 
 
 class DirectPathProfile(object):
-
-    def __init__(self,
-                 bike_use_roads,
-                 bike_use_hills,
-                 bike_use_ferry,
-                 bike_avoid_bad_surfaces,
-                 bike_shortest,
-                 bicycle_type,
-                 bike_use_living_streets,
-                 bike_maneuver_penalty,
-                 bike_service_penalty,
-                 bike_service_factor,
-                 bike_country_crossing_cost,
-                 bike_country_crossing_penalty,
-                 tag,
-                ):
+    def __init__(
+        self,
+        bike_use_roads=0.5,
+        bike_use_hills=0.5,
+        bike_use_ferry=0.5,
+        bike_avoid_bad_surfaces=0.25,
+        bike_shortest=False,
+        bicycle_type='hybrid',
+        bike_use_living_streets=0.5,
+        bike_maneuver_penalty=5,
+        bike_service_penalty=0,
+        bike_service_factor=1,
+        bike_country_crossing_cost=600,
+        bike_country_crossing_penalty=0,
+        tag=None,
+    ):
         self.bike_use_roads = bike_use_roads
         self.bike_use_hills = bike_use_hills
         self.bike_use_ferry = bike_use_ferry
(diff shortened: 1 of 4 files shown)


source/jormungandr/jormungandr/street_network/asgard.py
```diff
@@ -227,38 +273,101 @@ def handle_car_no_park_modes(mode):
...
-    def _create_direct_path_request(
-        self, mode, pt_object_origin, pt_object_destination, fallback_extremity, request, language="en-US"
-    ):
-        # asgard and kraken have a part of request in common
-        req = create_kraken_direct_path_request(
-            self, mode, pt_object_origin, pt_object_destination, fallback_extremity, request, language
+    @staticmethod
+    def is_alternatives_request(mode, direct_path_type, request):
+        return (
+            mode == 'bike'
```
woshilapin, line 279: Do we need to test for bikes too? From what I understand, it's sure is for GeoVelo only at the moment but who knows, we might have other mode with the same functionality. If you think it's better to keep it, then maybe rename the function into `is_bike_alternatives_request` to make it explicit about "it's only bike"?

source/jormungandr/tests/direct_path_asgard_integration_tests.py
```diff
@@ -201,43 +207,103 @@ def bss_route_response(request):
...
 def valid_response(request):
     if request.requested_api == type_pb2.direct_path:
-        if request.direct_path.streetnetwork_params.origin_mode == "bss":
+        if request.direct_path.profiles_params[0].origin_mode == "bss":
             return bss_route_response(request)
-        return route_response(request.direct_path.streetnetwork_params.origin_mode)
+        if len(request.direct_path.profiles_params) == 1:
+            return route_response(request.direct_path.profiles_params[0].origin_mode)
+        if len(request.direct_path.profiles_params) == len(DIRECT_PATH_ALTERNATIVES_PROFILES):
+            return alternative_route_response(request.direct_path.profiles_params)
```
woshilapin, line 227: Before, it was always returning `route_response()` but now, it will only return when there is `1` or `len(DIRECT_PATH_ALTERNATIVES_PROFILES)`. What about cases when it's other values?

pbench (approved):

source/jormungandr/jormungandr/street_network/asgard.py
xlqian: You understood it well. the alternative journeys are only available for bike mode for the moment, but not only bike for sure...

source/jormungandr/tests/direct_path_asgard_integration_tests.py
xlqian: `None` will be returned

PR #3755, opened 2022-05-31
[Jormun] Fix autocomplete without coverage and bragi version in configuration

kadhikari:
For a request on endpoint `/places` without coverage, navitia includes all data_sets `is_free=true` as well as data_sets related to `user`. Among the data_sets (instances), if at least one index on instance of type `munin_stop_<instance_name>` is absent, bragi reply with an error and hence navitia replies with `Internal Server Error`
Corrections:

- In the function get_all_available_instances we exclude 'kraken' so that navitia exclude all instances with autocopmlete_backend='kraken' while calling bragi.
- When autocomplete is used to get entrypoint_detail for a journey, if autocomplete is not in parameter, we use 'bragi'. This value will configurable with 'bragi' as default value.

Call example : https://navitia2-ws.ctp.customer.canaltp.fr/v1/places?q=st%20ouen&_autocomplete=bragi7&


source/navitiacommon/navitiacommon/models/__init__.py
```diff
@@ -216,22 +216,48 @@ def has_shape(self):
...
+                .filter(Authorization.user_id == self.id)
+                .filter(Instance.autocomplete_backend != exclude_backend)
+                .all()
+            )
+        else:
+            return (
+                Instance.query_existing()
+                .join(Authorization)
+                .filter(Authorization.user_id == self.id)
+                .filter(Instance.autocomplete_backend != exclude_backend)
```
woshilapin: I don't think you want this line here :thinking: because both branches of the `if/else` are identical...

woshilapin (approved):

pbench (approved):

source/navitiacommon/navitiacommon/models/__init__.py
kadhikari: You are right

woshilapin (approved):

SGrenet (approved):

source/navitiacommon/navitiacommon/models/__init__.py
```diff
@@ -216,22 +216,42 @@ def has_shape(self):
 
     # Note: the private methods below exist to be mocked in the integration tests
 
-    def _get_all_explicitly_authorized_instances(self):
-        return Instance.query_existing().join(Authorization).filter(Authorization.user_id == self.id).all()
-
-    def _get_all_free_instances(self):
-        return Instance.query_existing().filter_by(is_free=True).all()
+    def _get_all_explicitly_authorized_instances(self, exclude_backend=None):
+        if exclude_backend:
```
xlqian, line 220:
```suggestion
        if exclude_backend is not None:
```

source/navitiacommon/navitiacommon/models/__init__.py
```diff
@@ -216,22 +216,42 @@ def has_shape(self):
...
+                .filter(Instance.autocomplete_backend != exclude_backend)
+                .all()
+            )
+        else:
+            return Instance.query_existing().filter_by(is_free=True).all()
 
-    def _get_all_instances(self):
-        return Instance.query_existing().all()
+    def _get_all_instances(self, exclude_backend=None):
+        if exclude_backend:
```
xlqian, line 243:
```suggestion
        if exclude_backend is not None:
```

xlqian (approved):

source/jormungandr/jormungandr/default_settings.py
```diff
@@ -214,6 +214,7 @@
 )
 
 MAX_JOURNEYS_CALLS = int(os.getenv('JORMUNGANDR_MAX_JOURNEYS_CALLS', 20))
+BRAGI_VERSION = os.getenv('JORMUNGANDR_BRAGI_VERSION', 'bragi')
```
azime: DEFAULT_AUTOCOMPLETE_BACKEND

woshilapin (approved):

pbench (approved):

PR #3757, opened 2022-06-01
[tyr][loki] deposit ntfs in shared nfs for loki

pbench:
Tyr can now be configured to deposit the full ntfs in a shared directory, so that loki can read it from a local directory


woshilapin (approved):
Seems ok, but you'll probably need a pass of `black` formatter on this.

xlqian (approved):

PR #3727, opened 2022-04-20
[CI] Upgrade ci tools

AzHicham:
Upgraded tools: 
pre-commit: 1.18.3 -> 2.9.2
black: 18.9b0 -> 22.3.0
clang-format: 6.0 -> 10.0 (@hove-io/coreteam please run apt install clang-format to install clang-format-10 on ubuntu-20.04)
sonar-cli: 4.6.0.2311 -> 4.7.0.2747 
mypy: v0.670 -> v0.800

black, mypy, have been deleted from requierement-pre-commit.txt because they must be installed with the following command: pre-commit install --install-hooks

PR #3761, opened 2022-06-02
[Docs] minor fixes on formatting 

pbougue:
~After a `pre-commit run --all`.
Not sure if I used an up-to-date version of tools though, rejecting this PR is possible (hoping that CI does it if needed).~

I didn't see https://github.com/hove-io/navitia/pull/3727, so:

* Switch doc to clang-format 10 (above seems to work fine too, 11 for me, 14 on ubuntu-22.04).
* ~Pin `ubuntu-20.04` to avoid blind update of `ubuntu-latest` on github-actions
  cf. https://github.com/actions/virtual-environments#available-environments~
  After discussions, it's not really necessary.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -102,12 +102,13 @@ But bare in mind, it requires python 3.6+ to run.
 
 ### C++ formatting
 
-Our pre-commit hooks are running [Clang-format](https://releases.llvm.org/6.0.0/tools/clang/docs/ClangFormat.html) to format our c++ codebase. You'll need version 6.0 in order to pass our CI.
+Our pre-commit hooks are running [Clang-format](https://releases.llvm.org/10.0.0/tools/clang/docs/ClangFormat.html) to format our c++ codebase.
+You'll need version 10 (or above) in order to pass our CI.
 ```sh
-sudo apt install clang-format-6.0
+sudo apt install clang-format-10  # available on Ubuntu 20.04
 ```
 
 In case you might want to run Clang-format on the entire codebase, you can do:
 ```sh
-pre-commit run clang-format-6.0 --all
+pre-commit run clang-format --all
 ```


pbougue:
I didn't see https://github.com/hove-io/navitia/pull/3727
So https://github.com/hove-io/navitia/blob/dev/CONTRIBUTING.md#c-formatting should be updated, instead.
I'll try to take a look at this later.

You may close this if you prefer avoiding trailing PRs.

pbougue:
After check of #3727 :
* Switch doc to clang-format 10 ( >10 seems to work fine too).
* Pin `ubuntu-20.04` to avoid blind update of `ubuntu-latest` on github-actions
  cf. https://github.com/actions/virtual-environments#available-environments

Pinning to 20.04 seems smarter IMO (considering versions of packages installed), but you may disagree.

pbench (approved):

woshilapin (commented):
I see 2 problems here.

First, pinning mean there is a risk we drift with time with proper up-to-date tools.

Second, why such an old version of Ubuntu. 20.04 is 2 years old already.

pbougue:
Your arguments are right, but considering all options have downsides, I don't know what's better.

We may accept the risk of upgrading blindly, then we would just fix things as they move (but it may be a bit of a headache to debug when it happens).

Third option: We may also push directly to 22.04, it has to be tested.
But it's a bit young for ubuntu, and in beta on github-actions :shrug: 

My vote shyly stays for `ubuntu-20.04`.

Nota: I'm perfectly OK if one wants to update this PR while I'm away (no veto for me on that subject).

woshilapin:
I'd vote for `ubuntu-latest` so we fix things as they break because we're usually not good at regularly updating stuff.

pbougue:
I amended my commit (back to `ubuntu-latest`), so finally just a doc update.

woshilapin (approved):

PR #3771, opened 2022-06-14
[Jormungandr] no db check if db is unused

xlqian:


source/jormungandr/jormungandr/utils.py
```diff
@@ -677,6 +677,10 @@ def journeys_absent(objects):
 
 
 def can_connect_to_database():
+    # Why testing the db connection when the db is disabled?
+    if app.config['DISABLE_DATABASE']:
+        return True
```
pbench: When the database is disabled I would expect that `can_connect_to_database()`
```suggestion
        return False
```
But maybe you have a good reason to `return True `?

pbench (approved):

PR #3783, opened 2022-06-29
[TYR]: add parameter cosmogony_file while indexing for ES7

kadhikari:
Relates to https://github.com/hove-io/mimirsbrunn/pull/771
Include a new parameter `cosmogony_file`  while indexing data for ES7
- bano2mimir
- openaddresses2mimir
- osm2mimir
- stops2mimir
- ntfs2mimir
- poi2mimir

The added parameter will look like 
`./ntfs2mimir -s "elasticsearch.url='http://....canaltp.prod:9200'" -s "container.dataset='sandbox'"  -c /etc/mimir -i /srv/data/elasticsearch/prod/sandbox_ntfs_final_1y.zip -m production -s "cosmogony_file='/srv/ed/autocomplete/back../cosmogony_europe.jsonl.gz'" run` 

For more details: https://navitia.atlassian.net/browse/NAV-1339

### DIFF
--- a/source/tyr/tyr/binarisation.py
+++ b/source/tyr/tyr/binarisation.py
@@ -699,7 +699,7 @@ def get_bano2mimir_params(working_directory, autocomplete_instance, autocomplete
         current_app.config['MIMIR_PLATFORM_TAG'],
     ]
     if cosmogony_file is not None:
-        params.extend(['-s', 'cosmogony_file=\'{}\''.format(cosmogony_file)])
+        params.extend(['-s', 'admins.cosmogony_file=\'{}\''.format(cosmogony_file)])
     params.extend(['run'])
     return params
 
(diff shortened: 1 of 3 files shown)


source/tyr/tyr/binarisation.py
```diff
@@ -697,8 +697,10 @@ def get_bano2mimir_params(working_directory, autocomplete_instance, autocomplete
         working_directory,
         '-m',
         current_app.config['MIMIR_PLATFORM_TAG'],
-        'run',
     ]
+    if cosmogony_file is not None:
+        params.extend(['-s', 'cosmogony_file=\'{}\''.format(cosmogony_file)])
+    return params.extend(['run'])
```
woshilapin: I'm not sure this is going to work because `extend()` function doesn't return anything, right? You probably need to do it in two lines.
```suggestion
    params.extend(['run'])
    return params
```

The same should be fixed below.

pbench (commented):
Apart from the comment https://github.com/hove-io/navitia/pull/3783#discussion_r909593753
everything looks good to me !

source/tyr/tyr/binarisation.py
kadhikari: You are absolutely right. thanks

pbench (approved):

woshilapin (approved):

PR #3785, opened 2022-07-04
bump chaos proto

pbench:
Following 
https://github.com/hove-io/chaos-proto/pull/37

Do not merge until the corresponding update is deployed on the chaos side


xlqian (approved):

pbougue:
Superseded by #4055 
Staying on old name with typo as Chaos also stayed on that.

This may be reconsidered in the future, just chose to avoid exploring this path currently as it's not necessary to switch.

PR #3789, opened 2022-07-21
[Config][AWS]: Force create repositories for tyr web

azime:


source/tyr/tyr/helper.py
```diff
@@ -229,7 +250,9 @@ def load_instance_config(instance_name):
     instance.pg_dbname = config['database']['dbname']
     instance.pg_username = config['database']['username']
     instance.pg_password = config['database']['password']
-    instance.pg_port = config['database']['port']
+    instance.pg_port = int(config['database']['port'])
+    if force_create_paths:
```
pbench: why not always creating the paths if they do not exist ?

azime: this setting is there only for testing i can make a mock for `os.makedirs`
I make the change

woshilapin (approved):

source/tyr/tyr/helper.py
azime: Done

pbench (approved):

kadhikari (approved):

PR #3791, opened 2022-07-22
Fix RailSection bug

pbench:
When a RailSection impacts a vehicle journey already modified by a previous disruption, it sometimes fails.
The culprit seems to be the combination : 
```
section  = vj.get_sections_ranks(..)
blocked_sa_sequence_matching(..,vj,  section)
```
The first calls returns a list of stop_time ranks of the **base vj**,
while the function `blocked_sa_sequence_matching` interprets these ranks on the vj itself, not on the base vj.

So if vj already have some stop_times removed by a previous disruptions, the stop_time ranks of vj and base are different.

I really don't see the usefulness of `blocked_sa_sequence_matching` so I removed it :crossed_fingers:


pbench:
continued in https://github.com/hove-io/navitia/pull/3793

PR #3792, opened 2022-07-22
Fix rail section bug v2

pbench:


pbench:
continued in https://github.com/hove-io/navitia/pull/3793

PR #3793, opened 2022-07-26
[Kraken][RailSection]: Revamp rail section

pbench:
When a RailSection impacts a vehicle journey already modified by a previous disruption, it sometimes fails.
This is an attempt to fix this problem.

I revamped the RailSection object, to better understand what is going on.

### DIFF
--- a/source/type/message.cpp
+++ b/source/type/message.cpp
@@ -217,8 +217,8 @@ std::vector<ImpactedVJ> get_impacted_vehicle_journeys(const RailSection& rs,
     type::ValidityPattern impact_vp = impact.get_impact_vp(production_period);
 
     LOG4CPLUS_DEBUG(log, "validity pattern " << impact_vp.str());
-    LOG4CPLUS_DEBUG(log, "start stop area " << rs.start_point->label);
-    LOG4CPLUS_DEBUG(log, "end stop area " << rs.end_point->label);
+    LOG4CPLUS_DEBUG(log, "start stop area " << rs.start->label);
+    LOG4CPLUS_DEBUG(log, "end stop area " << rs.end->label);
     LOG4CPLUS_DEBUG(log, "impact application periods length " << impact.application_periods.size());
 
     auto apply_impacts_on_vj = [&](const nt::VehicleJourney& vj) {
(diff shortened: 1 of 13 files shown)


pbench:
This fixes https://navitia.atlassian.net/browse/NAV-1205

source/type/message.cpp
```diff
@@ -697,59 +620,237 @@ std::set<StopPoint*> LineSection::get_stop_points_section() const {
...
+    // let's first check if we can find the first impacted_stop_area in the stop_time list
+    const StopArea* first_stop_area = impacted_stop_areas.front();
+    auto find_first_stop_area = std::find_if(stop_times.begin(), stop_times.end(), [&](StopTime stop_time) {
+        return stop_time.is_in_stop_area(first_stop_area);
+    });
+    bool first_stop_area_found_in_stop_times = find_first_stop_area != stop_times.end();
+
+    // the first impacted stop_area appears in the stop_time list,
+    // let's keep iterating the stop_times to see if it contains all impacted_stop_areas consecutively
+    if (first_stop_area_found_in_stop_times) {
```
xlqian, line 796: I don't think this if-condition is necessary...

xlqian (approved):

source/type/message.cpp
```diff
@@ -697,59 +620,237 @@ std::set<StopPoint*> LineSection::get_stop_points_section() const {
...
+                                               << *line_uri << "' but also a route with no line: " << route->uri);
+                    return boost::none;
+                }
+                if (route->line->idx != line->idx) {
+                    LOG4CPLUS_WARN(logger, "Rail section has line: '"
+                                               << *line_uri << "' but also a route: '" << route->uri
+                                               << "' that belongs to a different line: " << route->line->uri);
+                    return boost::none;
+                }
+            }
```
woshilapin, line 718: If I understand correctly, if only one route doesn't belong to the line, we reject the entire disruption? Would it be a problem to have a less strict behavior and only drop the routes that do not belong to the line and keep working with the other routes?

source/type/message.cpp
```diff
@@ -697,59 +620,237 @@ std::set<StopPoint*> LineSection::get_stop_points_section() const {
...
+        }
+    }
+
+    // fill impacted stop_areas from start, blockeds and end
+    std::vector<StopArea*> impacted_stop_areas{};
+    if (blockeds.empty()) {
+        impacted_stop_areas.push_back(start);
+        if (end->idx != start->idx) {
+            impacted_stop_areas.push_back(end);
+        }
```
woshilapin: If I understand correctly, this corresponds to the following scenario.

```
A -> B -> C -> D -> E
     ^         ^
   start      end
```
and none of the stops are indicated as `blocking_areas`, right?
Is it expected that `C` is not impacted in that kind of case?

source/type/message.h
```diff
@@ -214,21 +215,59 @@ struct LineSection {
...
+          impacted_stop_areas(impacted_stop_areas_),
+          line(line_),
+          routes(routes_){};
+
+    // never null
+    StopArea* start;
+    // never null
+    StopArea* end;
+
+    std::vector<StopArea*> blockeds;
```
woshilapin, line 237: ANy reason why you changed the name from `blocked_stop_areas` to `blocked`, since you kept the suffix `_stop_areas` for `impacted_stop_areas`?

source/type/message.cpp
```diff
@@ -697,59 +620,237 @@ std::set<StopPoint*> LineSection::get_stop_points_section() const {
...
+    if (stop_times.empty()) {
+        return false;
+    }
+
+    // the RailSection impacts the vj if :
+    // the vj contains all impacted_stop_areas in their exact order
+
+    // let's first check if we can find the first impacted_stop_area in the stop_time list
+    const StopArea* first_stop_area = impacted_stop_areas.front();
+    auto find_first_stop_area = std::find_if(stop_times.begin(), stop_times.end(), [&](const StopTime& stop_time) {
```
woshilapin: Isn't this a `StopTime`? Because the name `find_first_stop_area` is misleading.

woshilapin (approved):

source/type/message.cpp
pbench: I just checked with @stifoon and he thinks it's okay this way

kadhikari (approved):

PR #3802, opened 2022-08-31
[aws migration] [Tyr]: Create target path and remove unused path (synonyms and aliases)

azime:


source/tyr/tyr/helper.py
```diff
@@ -251,7 +246,9 @@ def load_instance_config(instance_name):
     instance.pg_username = config['database']['username']
     instance.pg_password = config['database']['password']
     instance.pg_port = int(config['database']['port'])
-    create_repositories(instance)
+
+    create_repositories([instance.source_directory, instance.backup_directory], instance.name)
+    create_repositories([os.path.dirname(instance.target_file)], instance.name)
```
pbougue, line 247: all minors:
1. The `os.path.dirname()` is useless with the use of `os.makedirs`, no ?

2. You could group :
```suggestion
    create_repositories([instance.source_directory, instance.backup_directory, instance.target_file], instance.name)
```

3. My understanding is that it's a "silent" break if source_dir, target_dir or backup_dir are missing (None) in instance: is that OK ?

source/tyr/tests/integration/conftest.py
```diff
@@ -169,12 +169,14 @@ def create_repositories_instance_env_variables():
     backup_directory = "{path}/ed/backup".format(path=tmp_path)
     aliases_directory = "{path}/ed/aliases".format(path=tmp_path)
     synonyms_directory = "{path}/ed/synonyms_file".format(path=tmp_path)
```
pbougue: Shouldn't you remove those from tests as much as possible, to check it's safe?

pbougue (approved):
Non-blocking remarks, reply only may be enough.

source/tyr/tyr/helper.py
azime: 1- No, taget_file is file /path/data.nav.lz4
2- I can't because of a taget_file
3- Yes

source/tyr/tests/integration/conftest.py
```diff
@@ -167,14 +164,13 @@ def create_repositories_instance_env_variables():
...
+    target_path = "{path}/ed/target_file".format(path=tmp_path)
 
     os.environ["TYR_INSTANCE_auv"] = (
         '{"instance":{"name":"auv","source-directory":"' + source_directory + '",'
-        '"backup-directory":"' + backup_directory + '","aliases_file":"' + aliases_directory + '",'
-        '"synonyms_file":"' + synonyms_directory + '","target-file":"ed/target_file",'
-        '"tmp_file":"/ed/tmp_file","is-free":true,"exchange":"exchange"},'
+        '"backup-directory":"' + backup_directory + '","target-file":"ed/target_file",'
+        '"tmp_file":"/ed/tmp_file","is-free":true,"exchange":"exchange",'
+        '"target-file":"' + target_path + '/data.nav.lz4"},'
```
woshilapin: Seems to me that `target-file` is defined twice in this block.

source/tyr/tyr/helper.py
```diff
@@ -241,8 +236,8 @@ def load_instance_config(instance_name):
     instance.backup_directory = config['instance']['backup-directory']
     instance.target_file = config['instance']['target-file']
     instance.exchange = config['instance']['exchange']
-    instance.synonyms_file = config['instance']['synonyms_file']
-    instance.aliases_file = config['instance']['aliases_file']
+    instance.synonyms_file = config.get('instance').get('synonyms_file')
+    instance.aliases_file = config.get('instance').get('aliases_file')
```
woshilapin: Aren't you removing these 2 unused path (they've been removed from the test files above)?

woshilapin (approved):

source/tyr/tests/integration/conftest.py
azime: Done

source/tyr/tests/integration/conftest.py
azime: :+1:

source/tyr/tyr/helper.py
azime: Done

woshilapin (approved):

PR #3803, opened 2022-09-01
feat: add 'amqps' support for kraken

woshilapin:


pbougue (approved):
LGTM (looking a bit into fabric and conf), go test it :stuck_out_tongue:

source/kraken/maintenance_worker.cpp
```diff
@@ -355,13 +359,28 @@ void MaintenanceWorker::init_rabbitmq() {
...
-    if (broker_uri) {
-        LOG4CPLUS_INFO(logger, boost::format("connection to rabbitmq with uri: %s") % *broker_uri);
-        channel = AmqpClient::Channel::CreateFromUri(*broker_uri);
-    } else {
-        LOG4CPLUS_INFO(logger, boost::format("connection to rabbitmq: %s@%s:%s/%s") % username % host % port % vhost);
-        channel = AmqpClient::Channel::Create(host, port, username, password, vhost);
+    if (protocol == "amqp") {
+        // connection through URI or classic method
+        if (broker_uri) {
+            LOG4CPLUS_INFO(logger, boost::format("connection to rabbitmq with uri: %s") % *broker_uri);
```
azime: is it possible to change the log level (debug) or delete the mpd

azime (approved):

source/kraken/maintenance_worker.cpp
```diff
@@ -355,13 +359,28 @@ void MaintenanceWorker::init_rabbitmq() {
...
+        } else {
+            LOG4CPLUS_DEBUG(logger, boost::format("connection to rabbitmq: %s://%s@%s:%s/%s") % protocol % username
+                                       % host % port % vhost);
+            channel = AmqpClient::Channel::Create(host, port, username, password, vhost);
+        }
+    } else if (protocol == "amqps") {
+        // connection through URI or classic method
+        if (broker_uri) {
+            LOG4CPLUS_DEBUG(logger, boost::format("connection to rabbitmq with uri: %s") % *broker_uri);
+            channel = AmqpClient::Channel::CreateSecureFromUri(*broker_uri, path_to_ca_cert, path_to_client_key,
```
pbougue: To test: How about passing `verify_hostname=false` ?

source/kraken/maintenance_worker.cpp
woshilapin: Not relevant anymore, I used the new API from `SimpleAmqpClient` and am able to log everything always, except the password. I believe this should be OK now.

source/kraken/maintenance_worker.cpp
woshilapin: `verify_hostname=true` is fine... but `verify_peer=false` was the key!

pbougue:
Uses https://github.com/hove-io/SimpleAmqpClient/pull/4 and bumped librabbitmq-c

PR #3805, opened 2022-09-05
[aws migration]: Configure asgard zmq socket

azime:


pbougue (approved):
You probably separate work on purpose, but you're gonna have the same issue for kraken, right ?
(I didn't spot it already in config)

kadhikari (approved):

azime:
For kraken the config is in the env variable, it is not in the database

PR #3808, opened 2022-09-07
[aws migration]: Create autocomplete paths

azime:


kadhikari (approved):

source/tyr/tyr/resources.py
```diff
@@ -68,11 +68,12 @@
 )
 from tyr.helper import (
     get_instance_logger,
-    save_in_tmp,
```
pbougue, line 71: You can also remove the fn from helper, as it's not used anymore (and it will help review the PR to see what was in it)

source/tyr/tyr/resources.py
```diff
@@ -2416,6 +2423,9 @@ def post(self):
         f = args['file']
         file_name = f.filename
         file_path = str(os.path.join(os.path.abspath(current_app.config['CITIES_OSM_FILE_PATH']), file_name))
```
pbougue, line 2425: Not sure at all: Shouldn't you use tmp_dir or source_dir here, then?
(I don't get why you create directories as you do not seem to use them)

source/tyr/tyr/tasks.py
```diff
@@ -701,18 +702,7 @@ def heartbeat():
 def create_autocomplete_depot(name):
     autocomplete_dir = current_app.config['AUTOCOMPLETE_DIR']
```
pbougue: This line can be removed

pbougue (commented):
Apart from cleaning some things, I don't see issue, but I'm not self-confident in this tyr-autocomplete part of the code.
And sonar is a bit too pushy about the security-hotspot in tests, it's no problem IMO.

source/tyr/tyr/tasks.py
azime: :+1:

source/tyr/tyr/resources.py
azime: no it does not need tmp

woshilapin (approved):

source/tyr/tyr/resources.py
pbougue: OK, so you need only the "root" directory, but it is required to create it.
And the whole tree is too much, but it doesn't hurt.

Cleaning the way it works would be a bit more work and we do not want to handle it now.

source/tyr/tyr/resources.py
azime: Done

pbougue (approved):

PR #3814, opened 2022-09-13
Readyness endpoint

pbench:
Enable a readyness check for kraken by  moving the initial data load from the main thread into the "maintenance" thread. Consequently, the zmq thread is started and alive even before the initial data load completed, and kraken can answer zmq status requests immediately.

Adds three endpoints in the app "monitor_kraken" : 
- `/health `that returns a 200 when it obtains a response from kraken to a zmq status request, disregarding the content of the response
- `/ready` that returns a 200 when it it obtains a status response from kraken in which the field "loaded" is true
- `/status` that returns the content of the status response obtained from kraken

Note that these 3 endpoints are configured to call a **single** kraken, which is called through the zmq socket defined in the environment variable `KRAKEN_GENERAL_zmq_socket`.
In contrast, the "old" endpoint is used to monitor **several** kraken instance, and is configured through the python config `app.config['KRAKEN_DIR']`

Also note the "hack" to ensure backward compatibility with the current deployment mechanism https://github.com/hove-io/navitia/blob/b0c78408300724adb8cfd6b7f951daa01d781df7/source/monitor/monitor_kraken/app.py#L104-L108

### DIFF
--- a/source/monitor/monitor_kraken/app.py
+++ b/source/monitor/monitor_kraken/app.py
@@ -38,9 +38,15 @@ from monitor_kraken import response_pb2
 from monitor_kraken import type_pb2
 
 app = Flask(__name__)
+# config used in monitor() for several kraken
 app.config.from_object('monitor_kraken.default_settings')
 app.config.from_envvar('MONITOR_CONFIG_FILE', silent=True)
 context = zmq.Context()
+
+# config used in status() health() and ready()
+# for a single kraken
+single_kraken_zmq_socket = os.environ.get("KRAKEN_GENERAL_zmq_socket")
+single_kraken_zmq_timeout_in_ms = os.environ.get("MONITOR_KRAKEN_ZMQ_TIMEOUT", 10000)
 
 
 def get_kraken_uri(instance_name):
@@ -128,11 +134,9 @@ def monitor():
 # and dumps the response into json
 @app.route('/status')
 def status():
-    zmq_socket = os.environ.get("KRAKEN_GENERAL_zmq_socket")
-    if not zmq_socket:
+    if not single_kraken_zmq_socket:
         return json.dumps({'error': ['no zmq socket configured']}), 500
-    zmq_timeout_in_ms = os.environ.get("MONITOR_KRAKEN_ZMQ_TIMEOUT", 10000)
-    proto_response = request_kraken_zmq_status(zmq_socket, zmq_timeout_in_ms)
+    proto_response = request_kraken_zmq_status(single_kraken_zmq_socket, single_kraken_zmq_timeout_in_ms)
     if proto_response == None:
         return json.dumps({'error': 'zmq timed out'}), 503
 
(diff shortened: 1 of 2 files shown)


source/monitor/monitor_kraken/app.py
```diff
@@ -101,6 +101,14 @@ def monitor():
         response['kraken_version'] = resp.status.navitia_version
         response['data_version'] = resp.status.data_version
 
+        # Kraken used to not respond on zmq status requests during the initial
+        # data load. So the zmq requests timed-out and here we responded with a "timeout" response.
+        #  Now it does answer zmq requests even during the initial load.
```
woshilapin:
```suggestion
        # Now it does answer zmq requests even during the initial load.
```

woshilapin (approved):

pbougue (commented):
Should you add test? (I'm thinking of something like jormun integration tests but for monitor).

Also, does it affect Jormun? I'm thinking of coverage status being OK while data are not loaded. (There may be other possible issues I don't think of for jormun).

(didn't review code already)

EDIT: :heavy_check_mark: replied and considered below

azime (approved):

source/monitor/monitor_kraken/app.py
```diff
@@ -101,6 +101,14 @@ def monitor():
         response['kraken_version'] = resp.status.navitia_version
         response['data_version'] = resp.status.data_version
 
+        # Kraken used to not respond on zmq status requests during the initial
+        # data load. So the zmq requests timed-out and here we responded with a "timeout" response.
+        # Now it does answer zmq requests even during the initial load.
+        # So to be backward compatible, we send a "timeout" response when the data
+        # is not yet loaded by kraken.
+        if resp.status.loaded == False:
```
pbougue: minor: This could be moved up a bit before building the `response` dict

source/monitor/monitor_kraken/app.py
```diff
@@ -44,9 +44,6 @@
 
 
 def get_kraken_uri(instance_name):
-    uri = os.environ.get("KRAKEN_GENERAL_zmq_socket")
```
pbougue, line 47: Shouldn't we fallback to this in case config doesn't provide info? (which I guess is gonna be more common in the cloud?)
We may fallback before (no instance) also.

Not sure about all this, but this looks quite misleading (it is also probably not super important). Maybe just add a "deprecated" warning in code and in response?

pbougue: :heavy_check_mark: As discussed: doing nothing on that

Code comment does mention it's for multi-kraken monitoring, so it should be clear enough.

And fallback is not a big win to have a fallback on that... We may reconsider if it misleads somebody in the future, but it's unlikely.

source/monitor/monitor_kraken/app.py
```diff
@@ -114,5 +122,107 @@ def monitor():
...
+# Status of a single kraken through http://MONITOR_HOST/status
+# The kraken is accessed through the zmq socket defined in "KRAKEN_GENERAL_zmq_socket"
+# Performs a zmq status request to kraken
+# and dumps the response into json
+@app.route('/status')
+def status():
+    zmq_socket = os.environ.get("KRAKEN_GENERAL_zmq_socket")
+    if not zmq_socket:
+        return json.dumps({'error': ['no zmq socket configured']}), 500
+    zmq_timeout_in_ms = os.environ.get("MONITOR_KRAKEN_ZMQ_TIMEOUT", 10000)
```
pbougue: Should you add it to some kind of shared default conf (maybe just mutualize that value)?
Also 10s looks a bit long for timeout on a status, maybe devops would have an opinion on that.

pbougue (approved):
After discussion, replying to my previous PR-comment:
* tests are not cheap for the gain, so we will skip them for now
* it does affect jormungandr in case kraken is used before `/ready` is ok: used to timeout, will now lead to a bit inconsistent responses (impossible to find a start-point for journey, for example). Inconsistencies should not be major, and more important, kraken is not planned for use before being /ready. So this is only debt (on jormun's side, when interpreting kraken's response), but not an important issue.

pbougue (approved):
:ok_hand:

PR #3817, opened 2022-09-16
fix the datetime set in backtracking correction

pbench:
There was a confusion between the time-in-day (`st.arrival_time` and `st.departure_time`) and time-since-beginning-of-dataset (`section.get_out_dt` and `section.get_in_dt`).

So when a section was shorten, its modified endpoint was always set in the first day of the calendar.
We now compute the `base_dt` (that is midnight for the day on which the vj is used) from the section data, and then use this to set the correct datetime in the modified section.

Fix for https://navitia.atlassian.net/browse/NAV-1473

### DIFF
--- a/source/routing/raptor_api.cpp
+++ b/source/routing/raptor_api.cpp
@@ -1407,22 +1407,23 @@ bool shorten_section(navitia::routing::Journey::Section& section,
         if (vj == vj_to_skip) {
             continue;
         }
+
+        if (section.get_in_dt < section.get_in_st->boarding_time) {
+            LOG4CPLUS_ERROR(logger, "Error : section.get_in_dt < section.get_in_st->boarding_time");
+            return false;
+        }
+        // determine midnigth of the day at which the vj is used
+        navitia::DateTime base_dt = section.get_in_dt - section.get_in_st->boarding_time;
         for (const auto& st :
              boost::make_iterator_range(vj->stop_time_list.begin() + order.val, vj->stop_time_list.end())) {
             if (st.stop_point->stop_area->uri == stop_area_uri
                 && can_shorten_at(departures, destinations, st, clockwise)) {
-                // determine midnigth of the day at which the vj is used
-                // with protection from underflow
-                navitia::DateTime base_dt = 0;
-                if (section.get_in_dt > section.get_in_st->departure_time) {
-                    base_dt = section.get_in_dt - section.get_in_st->departure_time;
-                }
                 if (clockwise) {
                     section.get_out_st = &st;
-                    section.get_out_dt = base_dt + st.arrival_time;
+                    section.get_out_dt = base_dt + st.alighting_time;
                 } else {
                     section.get_in_st = &st;
-                    section.get_in_dt = base_dt + st.departure_time;
+                    section.get_in_dt = base_dt + st.boarding_time;
                 }
 
                 return true;
--- a/source/routing/raptor_api.cpp
+++ b/source/routing/raptor_api.cpp
@@ -1397,10 +1397,6 @@ bool shorten_section_clockwise(navitia::routing::Journey::Section& section,
         navitia::DateTime base_dt = section.get_in_dt - section.get_in_st->boarding_time;
         for (const auto& st :
              boost::make_iterator_range(vj->stop_time_list.begin() + order.val, vj->stop_time_list.end())) {
-            // if the end of original section is reached: no shortening on that section
-            if (&st == section.get_out_st) {
-                return false;
-            }
             // check stop_area is the same and it's possible to use stop_point to get out of PT
             if (st.stop_point->stop_area->uri == last_stop_area_uri
                 && navitia::contains(fallbacks, routing::SpIdx{st.stop_point->idx}) && st.drop_off_allowed()) {
@@ -1408,6 +1404,10 @@ bool shorten_section_clockwise(navitia::routing::Journey::Section& section,
                 section.get_out_dt = base_dt + st.alighting_time;
 
                 return true;
+            }
+            // if the end of original section is reached: interrupt search for shortening on that section
+            if (&st == section.get_out_st) {
+                return false;
             }
         }
     }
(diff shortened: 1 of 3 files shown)


pbench:
I'll try to add some tests

woshilapin (dismissed):

source/routing/raptor_api.cpp
```diff
@@ -1408,8 +1408,19 @@ bool shorten_section(navitia::routing::Journey::Section& section,
...
+                navitia::DateTime base_dt = 0;
+                if (section.get_in_dt > section.get_in_st->departure_time) {
+                    base_dt = section.get_in_dt - section.get_in_st->departure_time;
+                }
+                if (clockwise) {
+                    section.get_out_st = &st;
+                    section.get_out_dt = base_dt + st.arrival_time;
+                } else {
+                    section.get_in_st = &st;
+                    section.get_in_dt = base_dt + st.departure_time;
```
pbougue: Aren't you supposed to use `st.boarding_time` here instead? (alighting above)

source/routing/raptor_api.cpp
```diff
@@ -1408,8 +1408,19 @@ bool shorten_section(navitia::routing::Journey::Section& section,
              boost::make_iterator_range(vj->stop_time_list.begin() + order.val, vj->stop_time_list.end())) {
             if (can_shorten_at(departures, destinations, st, clockwise)
                 && st.stop_point->stop_area->uri == stop_area_uri && vj != vj_to_skip) {
```
pbougue: For clarity and perf, It looks like:
1. `vj != vj_to_skip` can be moved in the above loop (the one on vj), with a `continue`
2. the `stop_area` check could be done before, as it is gonna be false more often than `can_shorten_at()`

source/routing/raptor_api.cpp
```diff
@@ -1408,8 +1408,19 @@ bool shorten_section(navitia::routing::Journey::Section& section,
...
-                (clockwise ? section.get_out_st : section.get_in_st) = &st;
-                (clockwise ? section.get_out_dt : section.get_in_dt) = clockwise ? st.arrival_time : st.departure_time;
+                // determine midnigth of the day at which the vj is used
+                // with protection from underflow
+                navitia::DateTime base_dt = 0;
+                if (section.get_in_dt > section.get_in_st->departure_time) {
+                    base_dt = section.get_in_dt - section.get_in_st->departure_time;
```
pbougue: As discussed:
That `base_dt` processing should always be OK (except maybe for pass-midnight stay-in).
So underflow case is the sign of a bug somewhere else, and we should probably log+default to no-shortening (instead of defaulting to shortening on first day of dataset).

To handle pass-midnight stay-in (not sure it's possible from the rest of the code, but better to be safe) : this `base_dt` would probably be better computed outside of vj's loop.

Test suggestion (clockwise and anti-clockwise): past-midnight during first VJ, past-midnight during second VJ, past-midnight during stay-in. Maybe also a combination of them?
Those tests may be heavy, so feel free to ignore (part of) this suggestion.

stifoon: Good idea, but as we prioritize the migration from Kraken to Loki, let's keep it simple, even if the solution is limited. And let's assume that some cases won't be handled until we switch to Loki !

pbougue (approved):
Mainly self-review on that one :angel:

source/routing/raptor_api.cpp
```diff
@@ -1369,97 +1370,87 @@ void filter_late_journeys(RAPTOR::Journeys& journeys, const NightBusFilter::Para
...
 
                 return true;
             }
-            if (section.get_out_st == &st) {
-                return false;
-            }
         }
     }
-    LOG4CPLUS_WARN(logger, "Error occurred when modifying backtracking journeys");
+    LOG4CPLUS_ERROR(logger, "Error occurred when modifying backtracking journeys clockwise");
```
pbench, line 1414: This is not an error anymore to arrive here. You will reach the end of the function on every section (apart the last one) that cannot be shortened

source/routing/raptor_api.cpp
```diff
@@ -1369,97 +1370,87 @@ void filter_late_journeys(RAPTOR::Journeys& journeys, const NightBusFilter::Para
...
+                && navitia::contains(fallbacks, routing::SpIdx{st.stop_point->idx}) && st.pick_up_allowed()) {
+                section.get_in_st = &st;
+                section.get_in_dt = base_dt + st.boarding_time;
+
+                return true;
+            }
         }
     }
+
+    LOG4CPLUS_ERROR(logger, "Error occurred when modifying backtracking journeys anticlockwise");
```
pbench, line 1461: Same as above, this is not an error anymore to reach this line.

source/routing/raptor_api.cpp
```diff
@@ -1469,15 +1460,39 @@ void modify_backtracking_journeys(RAPTOR::Journeys& journeys,
...
         if (clockwise) {
-            auto last_section = journey.sections.rbegin();
+            const auto& last_section = journey.sections.rbegin();
             const auto& last_stop_area_uri = last_section->get_out_st->stop_point->stop_area->uri;
             const auto* last_vj = last_section->get_out_st->vehicle_journey;
-            shorten_journey(journey, last_stop_area_uri, last_vj, departures, destinations, clockwise);
+
+            size_t section_idx = 0;
+            // browse sections and modify any section that allows shortening
+            for (; section_idx < journey.sections.size(); section_idx++) {
```
pbench, line 1469: For clockwise, you should iterate sections backward and stops at the first section that can be shorten.

source/routing/raptor_api.cpp
pbougue: As discussed: Well-seen, only the last section of the VJ uses last_vj, so this case would happen regularly.
But then on intermediate sections we could search too far and create bugs.

So chose to keep it an error message and instead add a test that we reached section's end stop-time: https://github.com/hove-io/navitia/pull/3817/commits/4c3e911487e0d5193e2421e0c3c4546b6638e01c

source/routing/raptor_api.cpp
pbougue: See response above :wink:

source/routing/raptor_api.cpp
pbougue: :sweat_smile: Actually, stop shortening search AFTER checking the end stop-time: https://github.com/hove-io/navitia/pull/3817/commits/3bb9c1374a4309b540560449de066d2b9b7d0a1d

pbench:
Ok for me :white_check_mark:

PR #3818, opened 2022-09-16
Read rt topics env var

azime:
For env-var names, fix the way `_` is replaced to avoid replacing it in "final" parameters.

Also interpretation of `rt_topics` value to expand it to a vector (separator `;`) if it's a unique value (especially to allow list in env-var).

Result:
`$ KRAKEN_BROKER_rt_topics=rt1;rt2;rt3 ./kraken` should work as expected

### DIFF
--- a/source/kraken/configuration.cpp
+++ b/source/kraken/configuration.cpp
@@ -29,9 +29,6 @@ www.navitia.io
 */
 
 #include "configuration.h"
-
-#include "utils/exception.h"
-
 #include <boost/algorithm/string/predicate.hpp>
 #include <boost/algorithm/string/replace.hpp>
 #include <boost/optional.hpp>
--- a/source/kraken/configuration.cpp
+++ b/source/kraken/configuration.cpp
@@ -35,6 +35,7 @@ www.navitia.io
 
 #include <fstream>
 #include <iostream>
+#include "utils/functions.h"
 
 namespace po = boost::program_options;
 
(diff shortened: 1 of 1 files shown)


source/kraken/configuration.cpp
```diff
@@ -251,7 +251,11 @@ std::vector<std::string> Configuration::rt_topics() const {
     if (!this->vm.count("BROKER.rt_topics")) {
         return std::vector<std::string>();
     }
-    return this->vm["BROKER.rt_topics"].as<std::vector<std::string>>();
+    std::vector<std::string> result = this->vm["BROKER.rt_topics"].as<std::vector<std::string>>();
+    if (result.size() == 1) {
+        return split_string(result.at(0), ";");
```
pbench, line 254: If you have only one rt_topic with no ";" separator, split_string will return a vector containing only the input string, right ?

pbougue: Yes, from this PR and the signature it looks alright : https://github.com/hove-io/utils/pull/110/files#diff-bbd0a43b7217c0fb46210b9722b7d893f3092ebf7733a494ae48c10a4a122270R62

pbougue (approved):

woshilapin (approved):

source/kraken/configuration.cpp
azime: I just added these [tests](https://github.com/hove-io/utils/pull/111)

pbench (approved):

pbougue (approved):

PR #3823, opened 2022-09-19
Fix kraken monitor

azime:
In some kraken configurations (AWS), we find tcp://*:port.
We replace * by localhost =>  tcp://localhost:port

### DIFF
--- a/source/monitor/monitor_kraken/app.py
+++ b/source/monitor/monitor_kraken/app.py
@@ -209,7 +209,7 @@ def request_kraken_zmq_status(zmq_socket, zmq_timeout_in_ms):
     # discard messages when socket closed
     sock.setsockopt(zmq.LINGER, 0)
     try:
-        sock.connect(zmq_socket)
+        sock.connect(zmq_socket.replace('*', 'localhost'))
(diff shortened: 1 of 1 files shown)


source/monitor/monitor_kraken/app.py
```diff
@@ -209,7 +209,7 @@ def request_kraken_zmq_status(zmq_socket, zmq_timeout_in_ms):
     # discard messages when socket closed
     sock.setsockopt(zmq.LINGER, 0)
     try:
-        sock.connect(zmq_socket)
+        sock.connect(zmq_socket.replace('*', 'localhost'))
```
woshilapin, line 212: Are we sure we never going to have any `*` anywhere else in the URL than for the hostname?

woshilapin (approved):

pbench (approved):

source/monitor/monitor_kraken/app.py
azime: yes, the kraken uri has the form tcp://*:port

pbougue (approved):

PR #3824, opened 2022-09-20
Do launch maintenance thread

pbench:
In https://github.com/hove-io/navitia/pull/3814 the modification did launch the creation of the MaintenanceWorker, but did not call the operator::(), where multiple attempts to connect to rabbitmq are performed. 
Consequently, if the connection to rabbitmq failed during the creation of MaintenanceWorker, kraken did not attempt to reconnect afterwards.

This fixes the mistake.

### DIFF
--- a/source/kraken/kraken_zmq.cpp
+++ b/source/kraken/kraken_zmq.cpp
@@ -129,7 +129,7 @@ int main(int argn, char** argv) {
 
     threads.create_thread([&data_manager, conf, &metrics] {
         navitia::MaintenanceWorker maintenance_worker = navitia::MaintenanceWorker(data_manager, conf, metrics);
-        return maintenance_worker.run();
+        return maintenance_worker();
     });
     //
     // Data have been loaded, we can now accept connections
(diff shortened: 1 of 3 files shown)


kadhikari (approved):

pbougue (approved):
:exploding_head:

source/kraken/kraken_zmq.cpp
```diff
@@ -127,8 +127,10 @@ int main(int argn, char** argv) {
 
     const navitia::Metrics metrics(conf.metrics_binding(), conf.instance_name());
 
-    threads.create_thread(
-        [&data_manager, conf, &metrics] { return navitia::MaintenanceWorker(data_manager, conf, metrics); });
+    threads.create_thread([&data_manager, conf, &metrics] {
+        navitia::MaintenanceWorker maintenance_worker = navitia::MaintenanceWorker(data_manager, conf, metrics);
+        return maintenance_worker();
```
pbougue, line 132: minor (comment for later understanding): The `return` here is useless, the functor works forever.
And this functor call is what boost::thread used to do, but was missing in previous change.

PR #3827, opened 2022-09-26
[Monitor] always return the 'loaded' field in response

pbench:
Deployments failed because the monitor response did not contain the 'loaded' field, and the deployment script [use this field](https://github.com/hove-io/fabric_navitia/blob/3dac2039a02db87593c7368977681185d621e98e/fabfile/component/kraken.py#L371) to determine if the data is loaded, or if it should wait and ask kraken monitor again later


kadhikari (approved):

kadhikari:
We merge this pr to test the deploy on dev.
We can revert if it dosen't work

PR #3828, opened 2022-09-26
Tyr should repost data to secondary tyr

kadhikari:
During migration transition period of navitia, TYR on-premise should post data toward secondary TYR (aws) if configured.

Note: 

- Autocomplete global data (`cosmogony`, `france-latest.osm.pbf` and `bano.csv.gz`) will not be re-posted.
- Difficult to add test but well tested between TYR dev onpremise and TYR sandbox. coverages: ca-qc-sherbrooke and fr-auv

Parameter added in fab_navitia : https://github.com/hove-io/fabric_navitia/pull/370
To activate in dev : https://github.com/hove-io/navitia_deployment_conf/pull/1051/

Do not forget to set the environment variable REQUESTS_CA_BUNDLE
export REQUESTS_CA_BUNDLE=/etc/ssl/certs

### DIFF
--- a/source/tyr/tyr/default_settings.py
+++ b/source/tyr/tyr/default_settings.py
@@ -160,4 +160,6 @@ USE_SERPY = os.getenv('TYR_USE_SERPY', False)
 # deprecated and slow
 SQLALCHEMY_TRACK_MODIFICATIONS = os.getenv('TYR_SQLALCHEMY_TRACK_MODIFICATIONS', False)
 
-POST_DATA_TO_TYR = os.getenv('TYR_POST_DATA_TO_TYR', None)
+# Url of a secondary Tyr. The data posted to an instance of this Tyr via the '/jobs'
+# endpoint will be reposted to this url
+POST_DATA_TO_TYR = os.getenv('TYR_REPOST_DATA_TO_TYR', None)
--- a/source/tyr/tyr/default_settings.py
+++ b/source/tyr/tyr/default_settings.py
@@ -162,4 +162,4 @@ SQLALCHEMY_TRACK_MODIFICATIONS = os.getenv('TYR_SQLALCHEMY_TRACK_MODIFICATIONS',
 
 # Url of a secondary Tyr. The data posted to an instance of this Tyr via the '/jobs'
 # endpoint will be reposted to this url
-POST_DATA_TO_TYR = os.getenv('TYR_REPOST_DATA_TO_TYR', None)
+POST_DATA_TO_TYR = os.getenv('TYR_POST_DATA_TO_TYR', None)
--- a/source/tyr/tyr/default_settings.py
+++ b/source/tyr/tyr/default_settings.py
@@ -159,3 +159,5 @@ USE_SERPY = os.getenv('TYR_USE_SERPY', False)
 # https://flask-sqlalchemy.palletsprojects.com/en/2.x/signals/
 # deprecated and slow
 SQLALCHEMY_TRACK_MODIFICATIONS = os.getenv('TYR_SQLALCHEMY_TRACK_MODIFICATIONS', False)
+
+POST_DATA_TO_TYR = os.getenv('TYR_POST_DATA_TO_TYR', None)
--- a/source/tyr/tyr/helper.py
+++ b/source/tyr/tyr/helper.py
@@ -356,7 +356,6 @@ def repost_to_another_url(logger, url, content, instance_name):
     content.seek(0)
     file_to_post = {"file": (content.filename, content.stream, content.mimetype)}
     secondary_url = '{}/jobs/{}'.format(url, instance_name)
-    secondary_url = '{}/jobs/ca-qc-sherbrooke'.format(url)
     nb_try = 0
     while nb_try < MAX_TRY_FOR_REPOST_TO_SECONDARY_TYR:
         try:
--- a/source/tyr/tyr/helper.py
+++ b/source/tyr/tyr/helper.py
@@ -359,7 +359,7 @@ def repost_to_another_url(logger, url, content, instance_name):
     nb_try = 0
     while nb_try < MAX_TRY_FOR_REPOST_TO_SECONDARY_TYR:
         try:
-            resp = requests.post(secondary_url, files=file_to_post, verify=False)
+            resp = requests.post(secondary_url, files=file_to_post)
             logging.info('Info on posting data: {}'.format(resp.text))
             if resp.status_code == 200:
                 return True
--- a/source/tyr/tyr/helper.py
+++ b/source/tyr/tyr/helper.py
@@ -43,9 +43,11 @@ import sys
 from flask import json
 from jsonschema import validate, ValidationError
 from tyr.formats import instance_config_format
+import requests
 
 END_POINT_NOT_EXIST_MSG = 'end_point doesn\'t exist'
 BILLING_PLAN_NOT_EXIST_MSG = 'billing plan doesn\'t exist'
+MAX_NB_CALLS = 3
 
 
 def get_message(key, email, args):
@@ -348,3 +350,21 @@ def hide_domain(email):
     if "@" not in email:
         return email
     return "{}******".format(email.split('@')[0])
+
+
+def repost_to_another_tyr(logger, url, content, instance_name):
+    content.seek(0)
+    file_to_post = {"file": (content.filename, content.stream, content.mimetype)}
+    url_secondary_tyr = '{}/jobs/{}'.format(url, instance_name)
+    nb_try = 0
+    while nb_try < MAX_NB_CALLS:
+        try:
+            resp = requests.post(url_secondary_tyr, files=file_to_post, verify=False)
+            logging.info('Info on posting data: {}'.format(resp.text))
+            if resp.status_code == 200:
+                return True
+            nb_try = nb_try + 1
+        except Exception as e:
+            logger.error("Error while posting data to secondary tyr: {i}: {e}".format(i=url_secondary_tyr, e=e))
+            return False
+    return False
--- a/source/tyr/tyr/helper.py
+++ b/source/tyr/tyr/helper.py
@@ -47,7 +47,7 @@ import requests
 
 END_POINT_NOT_EXIST_MSG = 'end_point doesn\'t exist'
 BILLING_PLAN_NOT_EXIST_MSG = 'billing plan doesn\'t exist'
-MAX_NB_CALLS = 3
+MAX_TRY_FOR_REPOST_TO_SECONDARY_TYR = 3
 
 
 def get_message(key, email, args):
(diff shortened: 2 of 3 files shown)


source/tyr/tyr/resources.py
```diff
@@ -74,7 +74,10 @@
     hide_domain,
     create_autocomplete_instance_paths,
     create_repositories,
+    repost_to_another_tyr,
 )
+
+max_nb_calls = 3
```
woshilapin: I don't understand this line, it doesn't seem to be used in the rest of the file.

source/tyr/tyr/helper.py
```diff
@@ -348,3 +350,21 @@ def hide_domain(email):
     if "@" not in email:
         return email
     return "{}******".format(email.split('@')[0])
+
+
+def repost_to_another_tyr(logger, url, content, instance_name):
+    content.seek(0)
```
woshilapin, line 356: Why not using the following instead of passing around a `logger`?

```python
logger = get_instance_logger(instance)
```

source/tyr/tyr/helper.py
```diff
@@ -348,3 +350,21 @@ def hide_domain(email):
...
+    file_to_post = {"file": (content.filename, content.stream, content.mimetype)}
+    url_secondary_tyr = '{}/jobs/{}'.format(url, instance_name)
+    nb_try = 0
+    while nb_try < MAX_NB_CALLS:
+        try:
+            resp = requests.post(url_secondary_tyr, files=file_to_post, verify=False)
```
woshilapin: Now that you validated that it works, should we remove this `verify=False` and try to avoid bypassing security rules?

source/tyr/tyr/default_settings.py
```diff
@@ -159,3 +159,5 @@
 # https://flask-sqlalchemy.palletsprojects.com/en/2.x/signals/
 # deprecated and slow
 SQLALCHEMY_TRACK_MODIFICATIONS = os.getenv('TYR_SQLALCHEMY_TRACK_MODIFICATIONS', False)
+
+POST_DATA_TO_TYR = os.getenv('TYR_POST_DATA_TO_TYR', None)
```
pbench, line 165:
```suggestion
# Url of a secondary Tyr. The data posted to an instance of this Tyr via the '/jobs' endpoint will be reposted to this url
POST_DATA_TO_TYR = os.getenv('TYR_POST_DATA_TO_TYR', None)
```

source/tyr/tyr/resources.py
pbench:
```suggestion
```
It seems you don't use this variable

source/tyr/tyr/helper.py
```diff
@@ -43,9 +43,11 @@
 from flask import json
 from jsonschema import validate, ValidationError
 from tyr.formats import instance_config_format
+import requests
 
 END_POINT_NOT_EXIST_MSG = 'end_point doesn\'t exist'
 BILLING_PLAN_NOT_EXIST_MSG = 'billing plan doesn\'t exist'
+MAX_NB_CALLS = 3
```
pbench: Maybe a more specific name ?
```suggestion
MAX_TRY_FOR_REPOST_TO_SECONDARY_TYR= 3
```

source/tyr/tyr/default_settings.py
pbougue: I would also be closer to what it really does (it could be something else than tyr):
```suggestion
REPOST_DATA_TO_URL = os.getenv('TYR_REPOST_DATA_TO_URL', None)
```

source/tyr/tyr/helper.py
```diff
@@ -348,3 +350,21 @@ def hide_domain(email):
...
+    while nb_try < MAX_NB_CALLS:
+        try:
+            resp = requests.post(url_secondary_tyr, files=file_to_post, verify=False)
+            logging.info('Info on posting data: {}'.format(resp.text))
+            if resp.status_code == 200:
+                return True
+            nb_try = nb_try + 1
+        except Exception as e:
+            logger.error("Error while posting data to secondary tyr: {i}: {e}".format(i=url_secondary_tyr, e=e))
+            return False
```
woshilapin: Do we want to log such an error on each iteration of the loop? Maybe we want to catch the exception around the loop and inside the loop, only mention the number of the try, something like the following (not tested)?

```python
    try:
        while nb_try < MAX_NB_CALLS:
            resp = requests.post(url_secondary_tyr, files=file_to_post, verify=False)
            logging.info('Info on posting data: {}'.format(resp.text))
            if resp.status_code == 200:
                return True
            else:
                logger.warn("Failed to post data to secondary tyr [try {}]".format(nb_try)) 
            nb_try = nb_try + 1
    except Exception as e:
        logger.error("Error while posting data to secondary tyr: {i}: {e}".format(i=url_secondary_tyr, e=e))
        return False
```

source/tyr/tyr/helper.py
pbougue: Same, I'd push to:
```suggestion
MAX_TRY_FOR_REPOST_TO_URL = 3
```
(maybe `EXTERNAL_URL`,  `ANOTHER_URL`, `SECONDARY_URL` if `URL` is not clear enough)

source/tyr/tyr/helper.py
```diff
@@ -348,3 +350,21 @@ def hide_domain(email):
     if "@" not in email:
         return email
     return "{}******".format(email.split('@')[0])
+
+
+def repost_to_another_tyr(logger, url, content, instance_name):
```
pbougue:
```suggestion
def repost_to_another_url(logger, url, content, instance_name):
```

source/tyr/tyr/helper.py
```diff
@@ -348,3 +350,21 @@ def hide_domain(email):
...
+    nb_try = 0
+    while nb_try < MAX_NB_CALLS:
+        try:
+            resp = requests.post(url_secondary_tyr, files=file_to_post, verify=False)
+            logging.info('Info on posting data: {}'.format(resp.text))
+            if resp.status_code == 200:
+                return True
+            nb_try = nb_try + 1
+        except Exception as e:
+            logger.error("Error while posting data to secondary tyr: {i}: {e}".format(i=url_secondary_tyr, e=e))
```
pbougue:
```suggestion
            logger.error("Error while posting data to secondary url: {i}: {e}".format(i=secondary_url, e=e))
```

pbougue (approved):

source/tyr/tyr/default_settings.py
kadhikari: I would prefer : REPOST_DATA_TO_TYR

source/tyr/tyr/resources.py
kadhikari: :+1:

kadhikari: :+1:

source/tyr/tyr/helper.py
kadhikari: :+1:

woshilapin: I just realized my proposition is not equivalent: if the request actually raise an exception (not an HTTP error), then we will never retry 🤔

woshilapin: Maybe we need a two level exception catching.

```python
    try:
        while True:
            try:
                resp = requests.post(url_secondary_tyr, files=file_to_post, verify=False)
                logging.info('Info on posting data: {}'.format(resp.text))
                if resp.status_code == 200:
                    return True
            except Exception as e:
                if nb_try >= MAX_NB_CALLS:
                    raise e
            logger.warn("Failed to post data to secondary tyr [try {}]".format(nb_try))
            nb_try = nb_try + 1
    except Exception as e:
        logger.error("Error while posting data to secondary tyr: {i}: {e}".format(i=url_secondary_tyr, e=e))
        return False
```

source/tyr/tyr/helper.py
kadhikari: After installing certificate, it didn't work without this dirty fix this function will live for few weeks.
I am ready to test in dev without this fix.

source/tyr/tyr/helper.py
kadhikari: Because we need to pass instance instead of instance_name

source/tyr/tyr/helper.py
```diff
@@ -348,3 +350,22 @@ def hide_domain(email):
...
+    secondary_url = '{}/jobs/{}'.format(url, instance_name)
+    secondary_url = '{}/jobs/ca-qc-sherbrooke'.format(url)
+    nb_try = 0
+    while nb_try < MAX_TRY_FOR_REPOST_TO_SECONDARY_TYR:
+        try:
+            resp = requests.post(secondary_url, files=file_to_post, verify=False)
+            logging.info('Info on posting data: {}'.format(resp.text))
+            if resp.status_code == 200:
+                return True
+            nb_try = nb_try + 1
```
pbench: if an exception keep occuring inside `requests.post(...)` you will never increase the `nb_try` and you will get stuck forever in the loop. So I think it would be better to increase the counter at the beginning
```suggestion
            nb_try = nb_try + 1
            resp = requests.post(secondary_url, files=file_to_post, verify=False)
            logging.info('Info on posting data: {}'.format(resp.text))
            if resp.status_code == 200:
                return True
            
```

pbench: my bad, you will return from the function in case of exception :+1:

pbench (approved):

kadhikari:
Do not forget to set the environment variable REQUESTS_CA_BUNDLE
`export REQUESTS_CA_BUNDLE=/etc/ssl/certs`

woshilapin (approved):

source/tyr/tyr/helper.py
```diff
@@ -348,3 +350,21 @@ def hide_domain(email):
...
         return email
     return "{}******".format(email.split('@')[0])
+
+
+def repost_to_another_url(logger, url, content, instance_name):
+    content.seek(0)
+    file_to_post = {"file": (content.filename, content.stream, content.mimetype)}
+    secondary_url = '{}/jobs/{}'.format(url, instance_name)
+    nb_try = 0
+    while nb_try < MAX_TRY_FOR_REPOST_TO_SECONDARY_TYR:
```
xlqian, line 360: no sleep?

source/tyr/tyr/helper.py
```diff
@@ -348,3 +350,21 @@ def hide_domain(email):
...
+        try:
+            resp = requests.post(secondary_url, files=file_to_post)
+            logging.info('Info on posting data: {}'.format(resp.text))
+            if resp.status_code == 200:
+                return True
+            nb_try = nb_try + 1
+        except Exception as e:
+            logger.error("Error while posting data to secondary tyr: {i}: {e}".format(i=secondary_url, e=e))
+            return False
+    return False
```
xlqian, line 370:
```suggestion
    logger.error("Posting data failed after {} retries".format(nb_try))
    return False
```

xlqian (approved):

PR #3830, opened 2022-10-03
fix: tag without quotes for CI AWS images build

prhod:
Error seems to be caused by quotes in tag value

### DIFF
--- a/.github/workflows/publish_hove_images_aws.yml
+++ b/.github/workflows/publish_hove_images_aws.yml
@@ -40,7 +40,7 @@ jobs:
               echo "::set-output name=tag::dev"
             else :
               echo "Realease branch, collecting last version"
-              VERSION=$(curl https://api.github.com/repos/hove-io/navitia/tags | jq '.[0].name')
+              VERSION=$(curl https://api.github.com/repos/hove-io/navitia/tags | jq -r '.[0].name')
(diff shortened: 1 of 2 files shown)


woshilapin (approved):

.github/workflows/publish_hove_images_aws.yml
```diff
@@ -40,7 +40,7 @@ jobs:
               echo "::set-output name=tag::dev"
             else :
               echo "Realease branch, collecting last version"
-              VERSION=$(curl https://api.github.com/repos/hove-io/navitia/tags | jq '.[0].name')
+              VERSION=$(curl https://api.github.com/repos/hove-io/navitia/tags | jq -r '.[0].name')
```
woshilapin: How about the long version :wink: ?

```suggestion
              VERSION=$(curl https://api.github.com/repos/hove-io/navitia/tags | jq --raw-output '.[0].name')
```

prhod: I prefer the short one. Do you want me to change ?

woshilapin: The short one usually force the developer who reads the code to go to the documentation, the long version might avoid that :shrug: I'm fine either way.

azime (approved):

PR #3831, opened 2022-10-03
[RabbitMq] Do not create realtime queue before data is loaded + add queue time to live

pbench:
- Create the realtime queue only when the data has been successfully loaded (otherwise the queue just grows for nothing)
- add a time to live to the created queues, thanks to the `x-expires` [flag](https://www.rabbitmq.com/ttl.html#queue-ttl). It defaults to 2h, and can be configure with `BROKER.queue_expire`(in seconds)
- add a new parameter `BROKER.reconnect_wait` that specify (in seconds) how long kraken will wait between two attempts to connect do rabbitmq. Defaults to 1 second.

### DIFF
--- a/source/kraken/configuration.cpp
+++ b/source/kraken/configuration.cpp
@@ -94,6 +94,7 @@ po::options_description get_options_description(const boost::optional<std::strin
         ("BROKER.rt_topics", po::value<std::vector<std::string>>(), "list of realtime topic for this instance")
         ("BROKER.timeout", po::value<int>()->default_value(100), "timeout for maintenance worker in millisecond")
         ("BROKER.sleeptime", po::value<int>()->default_value(1), "sleeptime for maintenance worker in second")
+        ("BROKER.reconnect_wait", po::value<int>()->default_value(1), "Wait duration between connection attempts to rabbitmq, in seconds")
         ("BROKER.queue", po::value<std::string>(), "rabbitmq's queue name to be bound")
         ("BROKER.queue_auto_delete", po::value<bool>()->default_value(false), "auto delete rabbitmq's queue when unbind")
         ("BROKER.queue_expire", po::value<int>()->default_value(7200), "Rabbitmq queues created by kraken will be deleted by rabbitmq after this duration , in seconds")
@@ -233,6 +234,10 @@ int Configuration::broker_timeout() const {
 
 int Configuration::broker_sleeptime() const {
     return vm["BROKER.sleeptime"].as<int>();
+}
+
+int Configuration::broker_reconnect_wait() const {
+    return vm["BROKER.reconnect_wait"].as<int>();
 }
 
 std::string Configuration::broker_queue(const std::string& default_queue) const {
--- a/source/kraken/configuration.cpp
+++ b/source/kraken/configuration.cpp
@@ -96,6 +96,7 @@ po::options_description get_options_description(const boost::optional<std::strin
         ("BROKER.sleeptime", po::value<int>()->default_value(1), "sleeptime for maintenance worker in second")
         ("BROKER.queue", po::value<std::string>(), "rabbitmq's queue name to be bound")
         ("BROKER.queue_auto_delete", po::value<bool>()->default_value(false), "auto delete rabbitmq's queue when unbind")
+        ("BROKER.queue_ttl", po::value<int>()->default_value(7200), "Time To Live for rabbitmq queues created by kraken, in seconds")
 
         ("CHAOS.database", po::value<std::string>(), "Chaos database connection string")
         ("CHAOS.batch_size", po::value<int>()->default_value(1000000), "Chaos database row batch size");
@@ -243,6 +244,10 @@ std::string Configuration::broker_queue(const std::string& default_queue) const 
 
 bool Configuration::broker_queue_auto_delete() const {
     return vm["BROKER.queue_auto_delete"].as<bool>();
+}
+
+int Configuration::broker_queue_ttl() const {
+    return vm["BROKER.queue_ttl"].as<int>();
 }
 
 std::vector<std::string> Configuration::rt_topics() const {
--- a/source/kraken/configuration.cpp
+++ b/source/kraken/configuration.cpp
@@ -96,7 +96,7 @@ po::options_description get_options_description(const boost::optional<std::strin
         ("BROKER.sleeptime", po::value<int>()->default_value(1), "sleeptime for maintenance worker in second")
         ("BROKER.queue", po::value<std::string>(), "rabbitmq's queue name to be bound")
         ("BROKER.queue_auto_delete", po::value<bool>()->default_value(false), "auto delete rabbitmq's queue when unbind")
-        ("BROKER.queue_ttl", po::value<int>()->default_value(7200), "Time To Live for rabbitmq queues created by kraken, in seconds")
+        ("BROKER.queue_expire", po::value<int>()->default_value(7200), "Rabbitmq queues created by kraken will be deleted by rabbitmq after this duration , in seconds")
 
         ("CHAOS.database", po::value<std::string>(), "Chaos database connection string")
         ("CHAOS.batch_size", po::value<int>()->default_value(1000000), "Chaos database row batch size");
@@ -246,8 +246,8 @@ bool Configuration::broker_queue_auto_delete() const {
     return vm["BROKER.queue_auto_delete"].as<bool>();
 }
 
-int Configuration::broker_queue_ttl() const {
-    return vm["BROKER.queue_ttl"].as<int>();
+int Configuration::broker_queue_expire() const {
+    return vm["BROKER.queue_expire"].as<int>();
 }
 
 std::vector<std::string> Configuration::rt_topics() const {
--- a/source/kraken/maintenance_worker.cpp
+++ b/source/kraken/maintenance_worker.cpp
@@ -62,7 +62,7 @@ void MaintenanceWorker::run() {
 
     if (!is_data_loaded()) {
         // The data is not loaded, so we wait for a reload message on the task queue
-        // that results in a successfull data load.
+        // that results in a successful data load.
         // We may be disconnected from rabbitmq during this wait,
         // and reconnection to rabbitmq is handled by this while(true) loop
         while (true) {
@@ -70,7 +70,7 @@ void MaintenanceWorker::run() {
                 open_channel_to_rabbitmq();
                 create_task_queue();
                 // will return when a reload message on the task queue
-                // arrived and resulted in a successfull data load
+                // arrived and resulted in a successful data load
                 listen_to_task_queue_until_data_loaded();
                 // data is loaded, let's break from the while(true)  loop
                 break;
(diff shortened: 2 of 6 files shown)


source/kraken/maintenance_worker.cpp
```diff
@@ -56,14 +56,130 @@ namespace bg = boost::gregorian;
...
+void MaintenanceWorker::run() {
+    // try to load data from disk
+    load_data();
+
+    if (!is_data_loaded()) {
+        // The data is not loaded, so we wait for a reload message on the task queue
+        // that results in a successfull data load.
+        // We may be disconnected from rabbitmq during this wait,
+        // and reconnection to rabbitmq is handled by this while(true) loop
+        while (true) {
```
xlqian, line 68: is this loop still necessary since we use `libkeepalive` now?

xlqian: it seems wether the data is loaded or not, you need to open channel to rabbitmq, can you do that once at the beginning?

source/kraken/maintenance_worker.cpp
```diff
@@ -73,6 +189,106 @@ void MaintenanceWorker::load_data() {
     LOG4CPLUS_INFO(logger, "Loading database duration: " << duration);
 }
 
+void MaintenanceWorker::create_task_queue() {
+    if (!task_queue_created) {
```
xlqian: we can inverse the condition to avoid excessive indentations

source/kraken/maintenance_worker.cpp
pbench: - the loop is necessary to handle disconnection on rabbitmq side, like when the rabbitmq service died for some reason
- when rabbitmq dies, we need to reopen the channel, that's why it is in the loop

source/kraken/maintenance_worker.cpp
pbench: good idea :+1:

source/kraken/maintenance_worker.cpp
```diff
@@ -56,14 +56,130 @@ namespace bg = boost::gregorian;
 
 namespace navitia {
 
+void MaintenanceWorker::run() {
+    // try to load data from disk
+    load_data();
+
+    if (!is_data_loaded()) {
+        // The data is not loaded, so we wait for a reload message on the task queue
+        // that results in a successfull data load.
```
woshilapin:
```suggestion
        // that results in a successful data load.
```

source/kraken/maintenance_worker.cpp
```diff
@@ -56,14 +56,130 @@ namespace bg = boost::gregorian;
...
+        // We may be disconnected from rabbitmq during this wait,
+        // and reconnection to rabbitmq is handled by this while(true) loop
+        while (true) {
+            try {
+                open_channel_to_rabbitmq();
+                create_task_queue();
+                // will return when a reload message on the task queue
+                // arrived and resulted in a successfull data load
```
woshilapin:
```suggestion
                // arrived and resulted in a successful data load
```

source/kraken/maintenance_worker.cpp
```diff
@@ -73,6 +189,110 @@ void MaintenanceWorker::load_data() {
...
+        return;
+    }
+
+    std::string instance_name = conf.instance_name();
+    std::string hostname = get_hostname();
+    std::string default_queue_name = "kraken_" + hostname + "_" + instance_name;
+    std::string queue_name = conf.broker_queue(default_queue_name);
+    queue_name_task = (boost::format("%s_task") % queue_name).str();
+
+    // first we have to delete the queues, binding can change between two run, and it's don't seem possible
```
woshilapin:
```suggestion
    // first we have to delete the queues, binding can change between two run, and it doesn't seem possible
```

source/kraken/maintenance_worker.cpp
```diff
@@ -73,6 +189,110 @@ void MaintenanceWorker::load_data() {
...
+
+    std::string exchange_name = conf.broker_exchange();
+
+    std::string instance_name = conf.instance_name();
+    std::string hostname = get_hostname();
+    std::string default_queue_name = "kraken_" + hostname + "_" + instance_name;
+    std::string queue_name = conf.broker_queue(default_queue_name);
+    queue_name_rt = (boost::format("%s_rt") % queue_name).str();
+
+    // first we have to delete the queues, binding can change between two run, and it's don't seem possible
```
woshilapin:
```suggestion
    // first we have to delete the queues, binding can change between two run, and it doesn't seem possible
```

source/kraken/maintenance_worker.cpp
```diff
@@ -73,6 +189,110 @@ void MaintenanceWorker::load_data() {
...
+    bool durable = true;
+    bool exclusive = false;
+    bool auto_delete_queue = conf.broker_queue_auto_delete();
+
+    AmqpClient::Table args;
+    args.insert(std::make_pair("x-expires", conf.broker_queue_ttl() * 1000));
+
+    channel->DeclareQueue(this->queue_name_rt, passive, durable, exclusive, auto_delete_queue, args);
+    LOG4CPLUS_INFO(logger, "queue for disruptions: " << this->queue_name_rt);
+    // binding the queue to the exchange for all task for this instance
```
woshilapin:
```suggestion
    // binding the queue to the exchange for all tasks for this instance
```

source/kraken/maintenance_worker.cpp
```diff
@@ -73,6 +189,110 @@ void MaintenanceWorker::load_data() {
...
+    this->channel->DeclareExchange(exchange_name, "topic", false, true, false);
+
+    // creation of task queue for this kraken
+    bool passive = false;
+    bool durable = true;
+    bool exclusive = false;
+    bool auto_delete_queue = conf.broker_queue_auto_delete();
+
+    AmqpClient::Table args;
+    args.insert(std::make_pair("x-expires", conf.broker_queue_ttl() * 1000));
```
xlqian: so the queue is gonna die every 7200 seconds? what if the queue is not empty?

source/kraken/maintenance_worker.cpp
```diff
@@ -56,14 +56,130 @@ namespace bg = boost::gregorian;
...
+                create_task_queue();
+                // will return when a reload message on the task queue
+                // arrived and resulted in a successfull data load
+                listen_to_task_queue_until_data_loaded();
+                // data is loaded, let's break from the while(true)  loop
+                break;
+            } catch (const std::runtime_error& ex) {
+                LOG4CPLUS_ERROR(logger, "Connection to rabbitmq failed: " << ex.what());
+                data_manager.get_data()->is_connected_to_rabbitmq = false;
+                sleep(10);
```
xlqian:
```suggestion
                std::this_thread::sleep_for(10);
```

source/kraken/maintenance_worker.cpp
```diff
@@ -56,14 +56,130 @@ namespace bg = boost::gregorian;
...
+    while (true) {
+        try {
+            open_channel_to_rabbitmq();
+            create_task_queue();
+            create_realtime_queue();
+            listen_rabbitmq();
+        } catch (const std::runtime_error& ex) {
+            LOG4CPLUS_ERROR(logger, "Connection to rabbitmq failed: " << ex.what());
+            data_manager.get_data()->is_connected_to_rabbitmq = false;
+            sleep(10);
```
xlqian: (sorry to be a pest, I know that was not your code, it seems to be mine :see_no_evil: ) :
```suggestion
            std::this_thread::sleep_for(10);
```

source/kraken/maintenance_worker.cpp
pbougue: The doc is here: https://www.rabbitmq.com/ttl.html#queue-ttl
There are things that keep the queue "alive": consumer and re-declaration.
New messages don't, and that's what we want here: remove unconsumed queues that pile up messages and block all the Rabbit.

I'm not sure for "task" but for "rt", we can remove the queue as soon as the kraken linked to it is restarted (it will recreate the "rt" queue and ask for previously known realtime).
The queue should only be kept if kraken is still alive and can later reconnect (network issues).

source/kraken/configuration.cpp
```diff
@@ -96,6 +96,7 @@ po::options_description get_options_description(const boost::optional<std::strin
         ("BROKER.sleeptime", po::value<int>()->default_value(1), "sleeptime for maintenance worker in second")
         ("BROKER.queue", po::value<std::string>(), "rabbitmq's queue name to be bound")
         ("BROKER.queue_auto_delete", po::value<bool>()->default_value(false), "auto delete rabbitmq's queue when unbind")
+        ("BROKER.queue_ttl", po::value<int>()->default_value(7200), "Time To Live for rabbitmq queues created by kraken, in seconds")
```
pbougue: Not sure: as there is a `ttl` param in rabbit that can be applied to queues (affecting TTL to messages of the queue, not to the queue itself), maybe we should rename the param to `queue_expire`.
But as the paragraph of documentation of rabbit is "Queue TTL", I don't know :shrug:
My vote goes to `queue_expire`.

woshilapin: Note that we might also apply a `expire` parameter to the default `policy` too :shrug:

woshilapin (approved):

source/kraken/configuration.cpp
pbench: done

source/kraken/maintenance_worker.cpp
```diff
@@ -56,14 +56,116 @@ namespace bg = boost::gregorian;
...
+                // will return when a reload message on the task queue
+                // arrived and resulted in a successful data load
+                listen_to_task_queue_until_data_loaded();
+                // data is loaded, let's break from the while(true)  loop
+                break;
+            } catch (const std::runtime_error& ex) {
+                LOG4CPLUS_ERROR(logger, "Connection to rabbitmq failed: " << ex.what());
+                data_manager.get_data()->is_connected_to_rabbitmq = false;
+                channel_opened = false;
+                std::this_thread::sleep_for(std::chrono::seconds(1));
```
woshilapin: Just a suggestion...

```suggestion
                std::this_thread::sleep_for(std::chrono::seconds(conf.broker_sleeptime()));
```

source/kraken/maintenance_worker.cpp
```diff
@@ -56,14 +56,116 @@ namespace bg = boost::gregorian;
...
+            // will do nothing if the queue already exists
+            create_task_queue();
+
+            create_realtime_queue();
+            listen_rabbitmq();
+        } catch (const std::runtime_error& ex) {
+            LOG4CPLUS_ERROR(logger, "Connection to rabbitmq failed: " << ex.what());
+            data_manager.get_data()->is_connected_to_rabbitmq = false;
+            channel_opened = false;
+            std::this_thread::sleep_for(std::chrono::seconds(1));
```
woshilapin:
```suggestion
            std::this_thread::sleep_for(std::chrono::seconds(conf.broker_sleeptime()));
```

source/kraken/maintenance_worker.cpp
pbench: Now that I look carefully at where `conf.broker_sleeptime()` is used, its meaning is : "when I finished handling queues (i.e. checking if there is messages, and applying them if any) how long do I wait before looking at queues again".

Here, the duration means : "when connection to rabbitmq failed, how long do I wait before trying again to connect".

So I think I will introduce another parameter, rather than reuse `broker_sleeptime`

pbench: I used `BROKER.reconnect_wait` as the new parameter

source/kraken/maintenance_worker.cpp
pbench: I used `BROKER.reconnect_wait` as the new parameter

xlqian (approved):

azime (approved):

PR #3840, opened 2022-10-14
remove tcmalloc

pbench:


source/cmake_modules/CompilationFlags.cmake
```diff
@@ -76,7 +76,7 @@ if(STRIP_SYMBOLS)
     set(CMAKE_CXX_FLAGS_RELEASE "${CMAKE_CXX_FLAGS_RELEASE} -s")
 endif(STRIP_SYMBOLS)
 
-set(NAVITIA_ALLOCATOR "tcmalloc")
+set(NAVITIA_ALLOCATOR "")
```
pbougue, line 79: If that's made to last, shouldn't we make it configurable, or dependent on platform?

pbench:
this is not needed after all

PR #3846, opened 2022-11-04
IAM Credentials for S3 bucket 

pbench:
Try to adapt what is done in https://github.com/hove-io/asgard/pull/216/files


source/tyr/tyr/minio.py
```diff
@@ -1,42 +0,0 @@
-# coding: utf-8
```
woshilapin, line 1: This file was not used at all? :open_mouth:

woshilapin (approved):
We assume this new version of `tyr` is never going to be deployed on-premise? Note that I see the default value to AWS a bit problematic as an open-source repository: maybe we can default to `None` instead?

source/tyr/tyr/minio.py
pbench: well it was used only for reading some config variables, which are used only in one place. So i figured it was really not worth it to have a file for that

pbench:
Work continued in https://github.com/hove-io/navitia/pull/3849

PR #3847, opened 2022-11-07
Pr co2 emission physical mode

vpassama:
Hello,

This PR discussed with @pbench is about calculating missing CO2 emission value for car section provided by another backend than kraken (like asgard) in journey service.

The check about missing co2 is done in new_default scenario which will call kraken in order to do the calculation with a distance parameter (the street_network length of the car section).

Thank you.

### DIFF
--- a/source/jormungandr/jormungandr/scenarios/new_default.py
+++ b/source/jormungandr/jormungandr/scenarios/new_default.py
@@ -255,6 +255,29 @@ def sort_journeys(resp, journey_order, clockwise):
         resp.journeys.sort(key=cmp_to_key(journey_sorter[journey_order](clockwise=clockwise)))
 
 
+def fill_missing_co2_emission(pb_resp, instance, request_id):
+    if not pb_resp.journeys:
+        return
+
+    is_car_section_without_co2 = (
+        lambda s: s.type == response_pb2.STREET_NETWORK
+        and s.street_network.mode == response_pb2.Car
+        and not s.co2_emission.value
+    )
+
+    for j in pb_resp.journeys:
+        if 'car' not in j.tags:
+            continue
+
+        car_sections = (s for s in j.sections if is_car_section_without_co2(s))
+
+        for s in car_sections:
+            co2_emission = instance.georef.get_car_co2_emission(
+                s.street_network.length, request_id
+            )
+            s.co2_emission.CopyFrom(co2_emission)
+
+
 def compute_car_co2_emission(pb_resp, api_request, instance, request_id):
     if not pb_resp.journeys:
         return
@@ -1210,6 +1233,8 @@ class Scenario(simple.Scenario):
         type_journeys(pb_resp, api_request)
         culling_journeys(pb_resp, api_request)
 
+        # We fill empty co2_emission with estimations for car sections
+        fill_missing_co2_emission(pb_resp, instance, "{}_missing_car_co2".format(request_id))
         # We have to update total duration as some sections could be updated in distributed.
         update_durations(pb_resp)
         # We have to update total co2_emission as some sections could be updated in distributed.
--- a/source/kraken/worker.cpp
+++ b/source/kraken/worker.cpp
@@ -929,31 +929,37 @@ void Worker::heat_map(const pbnavitia::HeatMapRequest& request) {
                                     end_speed, end_mode, request.resolution(), center_and_stop_points.second);
 }
 
-void Worker::car_co2_emission_on_crow_fly(const pbnavitia::CarCO2EmissionRequest& request) {
+void Worker::car_co2_emission(const pbnavitia::CarCO2EmissionRequest& request) {
     const auto* data = this->pb_creator.data;
-    auto get_geographical_coord = [&](const pbnavitia::LocationContext& location) {
-        auto origin_type = data->get_type_of_id(location.place());
-        auto origin = type::EntryPoint{origin_type, location.place(), location.access_duration()};
-        auto coordinates = coord_of_entry_point(origin, *data);
-        return navitia::type::GeographicalCoord{coordinates.lon(), coordinates.lat()};
-    };
-    navitia::type::GeographicalCoord origin;
-    navitia::type::GeographicalCoord destin;
-    try {
-        origin = get_geographical_coord(request.origin());
-        destin = get_geographical_coord(request.destination());
-    } catch (const navitia::coord_conversion_exception& e) {
-        this->pb_creator.fill_pb_error(pbnavitia::Error::bad_format, e.what());
-        return;
-    }
+    auto distance = request.distance();
+    auto co2_estimation_coeff = 1;
 
-    auto distance = origin.distance_to(destin);
+    if (!request.has_distance()) {
+        auto get_geographical_coord = [&](const pbnavitia::LocationContext& location) {
+            auto origin_type = data->get_type_of_id(location.place());
+            auto origin = type::EntryPoint{origin_type, location.place(), location.access_duration()};
+            auto coordinates = coord_of_entry_point(origin, *data);
+            return navitia::type::GeographicalCoord{coordinates.lon(), coordinates.lat()};
+        };
+        navitia::type::GeographicalCoord origin;
+        navitia::type::GeographicalCoord destin;
+        try {
+            origin = get_geographical_coord(request.origin());
+            destin = get_geographical_coord(request.destination());
+        } catch (const navitia::coord_conversion_exception& e) {
+            this->pb_creator.fill_pb_error(pbnavitia::Error::bad_format, e.what());
+            return;
+        }
+
+        co2_estimation_coeff = CO2_ESTIMATION_COEFF;
+        distance = origin.distance_to(destin);
+    }
     auto car_mode = data->pt_data->physical_modes_map.find("physical_mode:Car");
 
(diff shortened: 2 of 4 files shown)


pbench:
Hi @vpassama !
Thanks for the PR :heart: 
For the update of navitia-proto : can you open a PR on the navitia-proto repo ? We will first merge the PR on navitia-proto and then this one.
Also on your changes on the protobuf : I don't think it is a good idea to reuse the `Co2Emission` object. This object is already used to represent a total of Co2 emitted along a section. You now add something which represents a rate : the CO2 emitted per km. Could you instead add a new object Co2EmissionRate so we don't get confused ?

source/jormungandr/jormungandr/scenarios/new_default.py
```diff
@@ -255,6 +255,40 @@ def sort_journeys(resp, journey_order, clockwise):
...
+                        car_mode = instance.send_and_receive(req, request_id=request_id)
+
+                    if not (car_mode.physical_modes
+                            and car_mode.physical_modes[0].HasField('co2_emission')
+                            and car_mode.physical_modes[0].co2_emission.value):
+                        return
+
+                    distance = 0
+                    for p in s.street_network.path_items:
+                        distance += p.length
```
pbench: Don't we already have the total distance somewhere ? Like in `s.street_network.length` ? 
Since path_items may be a big list, I am a bit worried about time it will take to make this computation

vpassama:
Thank you for your review @pbench, I'm working on the details right now. As you speak about a co2 rate, is the unit "gEC" a valid one in my case or should I use something else like 'g CO2/km' ? I'm not really sure I know what gEC means even if I tried to google it, I found some units there : [Émissions de CO2 des transports ferroviaires en France](https://fr.wikipedia.org/wiki/%C3%89missions_de_CO2_des_transports_ferroviaires_en_France) and they speak about "g.eqC/km" which is close but not really  the same. Thank you for your help.

edit : I looked at the [NTFS specifications](https://github.com/hove-io/ntfs-specification/blob/master/ntfs_fr.md#physical_modestxt-requis) and it seems you're expecting values in g CO2/voy/km.
I'm a bit confused about the value gEC as it seems it is not an equivalent of gCO2 (seen in the wiki) :  "un kgeqC correspond à 3,67 kgCO2" but maybe this is not the same thing.

source/jormungandr/jormungandr/scenarios/new_default.py
vpassama: Yes my bad I looked for this value but I missed it.

vpassama:
https://github.com/hove-io/navitia-proto/pull/186

xlqian:
> so WTH is gEC?

I did dig into the code to figure out the real meaning of it and here is what I found: https://github.com/hove-io/navitia/blob/7c7dcfd67f35a19f47d6d1208dd7351078d5147e/source/type/pb_converter.cpp#L1844
As I understand, it simply means 'g CO2/Km'

But what confused me more is this:
image

~I think the unit here is supposed to be 'g CO2/voy' instead of 'gEC'~

:exploding_head: 

Here is what I found in our french doc:
image

Mystery solved!

source/jormungandr/jormungandr/scenarios/new_default.py
```diff
@@ -255,6 +255,39 @@ def sort_journeys(resp, journey_order, clockwise):
         resp.journeys.sort(key=cmp_to_key(journey_sorter[journey_order](clockwise=clockwise)))
 
 
+def fill_missing_co2_emission(pb_resp, api_request, instance, request_id):
```
xlqian: you can simplify this function by using this request: https://github.com/hove-io/navitia-proto/blob/6022bbc485cbef5f03ac09c7ac74f719bf80fee1/request.proto#L195

Indeed, this is a request for estimation the co2 emission by computing the crow fly between two given coordinates, but you can add another optional field `distance` and modify the code here: https://github.com/hove-io/navitia/blob/9ad498d4edec7618e9677b9b0e23f4a2b33ed439/source/kraken/worker.cpp#L932 (you can rename the function to `car_co2_emission`).
 If the `distance` is provided in the request just return the `distance * co2_emission_rate` (https://github.com/hove-io/navitia/blob/9ad498d4edec7618e9677b9b0e23f4a2b33ed439/source/kraken/worker.cpp#L956), thus it can simplify your codes by a lot

source/type/pb_converter.cpp
```diff
@@ -750,6 +750,12 @@ void PbCreator::Filler::fill_pb_object(const nt::Network* n, pbnavitia::Network*
 void PbCreator::Filler::fill_pb_object(const nt::PhysicalMode* m, pbnavitia::PhysicalMode* physical_mode) {
     physical_mode->set_name(m->name);
     physical_mode->set_uri(m->uri);
+
+    if (depth >= 1 && m->co2_emission) {
+        pbnavitia::Co2EmissionRate* pb_co2_emission_rate = physical_mode->mutable_co2_emission_rate();
+        pb_co2_emission_rate->set_unit("gEC");
```
```suggestion
        pb_co2_emission_rate->set_unit("gEC/km");
```
If I understood the [ntfs doc](https://github.com/hove-io/ntfs-specification/blob/master/ntfs_fr.md#physical_modestxt-requis) correctly

source/jormungandr/jormungandr/scenarios/new_default.py
```diff
@@ -255,6 +255,39 @@ def sort_journeys(resp, journey_order, clockwise):
...
+                    if not (car_mode.physical_modes
+                            and car_mode.physical_modes[0].HasField('co2_emission_rate')
+                            and car_mode.physical_modes[0].co2_emission_rate.value):
+                        return
+
+                    s.co2_emission.unit = car_mode.physical_modes[0].co2_emission_rate.unit
+                    s.co2_emission.value = (
+                        s.street_network.length /
+                        1000 * car_mode.physical_modes[0].co2_emission_rate.value
+                    )
```
```suggestion
    is_car_section = lambda s: s.type == response_pb2.STREET_NETWORK and s.street_network.mode == response_pb2.Car and not s.co2_emission.value
    
    car_sections = (s for s in j.sections if is_car_section(s)  for j in pb_resp.journeys if helpers.is_car_direct_path(j))
    
    for s in car_sections:
        # function to be implemented in georef
        co2_emission = instance.georef.get_car_co2_emission(s.street_network.length, request_id)
        # not sure
        s.co2_emission.CopyFrom(co2_emission) 
        
```

vpassama:
Ok, so to summarize, @pbench @xlqian,

gEC is not geqC which would be 3,67 * gCO2 but is just an other term for gCO2 ?

Is your documentation reliable on this point or may it be a confusion made long times ago ?
I ask because I really didn't find another source related to this subject mentionning gEC as an existing unit.

xlqian:
Can you please rebase your PR to include the fix so that CI could run?

vpassama:
> Can you please rebase your PR to include the fix so that CI could run?

I didn't finish compiling and testing the changes. I'll do it when it'll be ok.
I wanted to update my navitia-proto PR but it has already be merged so I will open a new one.

You didn't reply on my last question about gCO2 unit, maybe you missed it.

xlqian:
> gEC is not geqC which would be 3,67 * gCO2 but is just an other term for gCO2 ?

Yes I'm quite sure it's just gCO2. @wassimbenaissa  Can you confirm it plz?

source/kraken/worker.cpp
```diff
@@ -929,25 +929,29 @@ void Worker::heat_map(const pbnavitia::HeatMapRequest& request) {
...
-    try {
-        origin = get_geographical_coord(request.origin());
-        destin = get_geographical_coord(request.destination());
-    } catch (const navitia::coord_conversion_exception& e) {
-        this->pb_creator.fill_pb_error(pbnavitia::Error::bad_format, e.what());
-        return;
-    }
+    auto distance = request.distance();
+
+    if (!distance) {
```
xlqian:
```suggestion
    if (!request.has_distance()) {
```

xlqian (approved):

xlqian:
Hello, we could do a tiny optimization: instead of calling kraken for EACH car journey, we can call kraken only once to retrieve co2_emission_rate, then cache it and reuse it...  (which is precisely what you have done... I'm sorry to be a pest :sweat: ) thus, we will need https://github.com/hove-io/navitia-proto/pull/186

vpassama:
> Hello, we could do a tiny optimization: instead of calling kraken for EACH car journey, we can call kraken only once to retrieve co2_emission_rate, then cache it and reuse it... (which is precisely what you have done... I'm sorry to be a pest 😓 ) thus, we will need [hove-io/navitia-proto#186](https://github.com/hove-io/navitia-proto/pull/186)

You want me to recode the first thing I did ? I lost all the code even it wasn't much but I just erased all the part with co2 emission rate in physical_mode api. So I revert everything and I just use a kraken call and then calculate distance * co2 in python ?

pbench:
@vpassama I'll merge the PR as is, so you can have the feature. 
@xlqian will do the optimization he is talking about in another PR. It will indeed prevent many calls to kraken.

pbench:
I'll reopen this PR from a branch of https://github.com/hove-io/navitia so that the CI can run (with PR from a fork, the Github Actions don't have access to secrets :cry: )

pbench:
Reopened in https://github.com/hove-io/navitia/pull/3851

PR #3849, opened 2022-11-08
Second try for IAM Credentials for S3 bucket

pbench:
The version 7.1 of the minio python package does handle IAM credentials out of the box, but it seems to not be compatible with python 2.7 https://github.com/hove-io/navitia/pull/3846

The last minio version which seems to be compatible with python 2.7 is minio 6.0. 
Unfortunately, it seems  minio 6.0 is not up to date with IAM credentials handling for ecs tasks.
So here I try to backport to minio 6.0 the way IAM credentials are handled for ecs tasks in minio 7.1 

See : 
https://stackoverflow.com/questions/57065458/cannot-access-instance-metadata-from-within-a-fargate-task
https://docs.aws.amazon.com/AmazonECS/latest/developerguide/task-iam-roles.html
https://github.com/minio/minio-py/blob/d45ba0cc1fb7b2cbcd8dfd2648c7851100929ccc/minio/credentials/providers.py#L343

### DIFF
--- a/source/tyr/tyr/binarisation.py
+++ b/source/tyr/tyr/binarisation.py
@@ -40,7 +40,6 @@ import shutil
 from functools import wraps
 
 from flask import current_app
-from minio import Minio
 from shapely.geometry import MultiPolygon
 from shapely import wkt
 from zipfile import BadZipfile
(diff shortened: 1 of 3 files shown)


woshilapin (approved):
This Python2.7 problem starts to be really annoying.

Backporting stuff from a recent version of minio in an old one, I was expecting the project to depend on a fork of minio, modified for our needs, not copy/paste code. If you're not forking, can you at least add comments about where the code come from.

xlqian (approved):

PR #3850, opened 2022-11-09
[AWS] Readyness Api added

azime:
**/v1/readyness**: new API added


xlqian (approved):

source/jormungandr/jormungandr/modules/v1_routing/resources/Readyness.py
```diff
@@ -0,0 +1,39 @@
...
+# https://groups.google.com/d/forum/navitia
+# www.navitia.io
+from __future__ import absolute_import, print_function, unicode_literals, division
+from jormungandr.module_resource import ModuleResource
+from jormungandr import i_manager
+
+
+class Readyness(ModuleResource):
+    def get(self):
+        instances_is_initialized = [instance.is_initialized for instance in i_manager.instances.values()]
```
xlqian: use a generator would be better, since `all` will stop iterating at the first `false` it will meet
```suggestion
        instances_is_initialized = (instance.is_initialized for instance in i_manager.instances.values())
```

pbench:
If I understand correctly,` instance.is_initialized `is set to `True` when a call to the kraken instance succeeded.
So this means if **one** kraken does not answer, jormun will be considered "not ready" and not used within aws.
So if I am not mistaken it means that with one kraken down *all* jormun becomes down.
Isn't it a bit risky ?

source/jormungandr/jormungandr/modules/v1_routing/resources/Readyness.py
azime: :+1:

xlqian (approved):

pbench (approved):

PR #3852, opened 2022-11-14
[Jormungandr] co2 optim: cache the physical_mode; fix unit problem

xlqian:
- Cache physical_mode 
- fix unit problem

### DIFF
--- a/source/jormungandr/jormungandr/georef.py
+++ b/source/jormungandr/jormungandr/georef.py
@@ -30,7 +30,7 @@
 from __future__ import absolute_import, print_function, unicode_literals, division
 from navitiacommon import request_pb2, response_pb2, type_pb2
 import logging
-from jormungandr import utils
+from jormungandr import utils, cache, memory_cache, app as current_app
 
 
 class Kraken(object):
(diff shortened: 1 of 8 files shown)


source/type/pb_converter.cpp
```diff
@@ -750,6 +750,16 @@ void PbCreator::Filler::fill_pb_object(const nt::Network* n, pbnavitia::Network*
...
     physical_mode->set_uri(m->uri);
+
+    if (depth == 0) {
+        return;
+    }
+
+    if (m->co2_emission) {
+        auto* emission_rate = physical_mode->mutable_co2_emission_rate();
+        emission_rate->set_value(*m->co2_emission);
+        emission_rate->set_unit("gCO2/Km");
```
```suggestion
        emission_rate->set_unit("gEC/Km");
```
Since the unit on CO2 emission on a section are gEC, this one should be gEC/Km, no ?

source/type/pb_converter.cpp
```diff
@@ -1,4 +1,4 @@
-/* Copyright © 2001-2022, Hove and/or its affiliates. All rights reserved.
+/* Copyright �� 2001-2022, Hove and/or its affiliates. All rights reserved.
```
pbench: You have some encoding problem here

source/jormungandr/jormungandr/georef.py
```diff
@@ -73,18 +73,32 @@ def get_car_co2_emission_on_crow_fly(self, origin, destination, request_id):
             return None
         return response.car_co2_emission
 
+    @memory_cache.memoize(
+        current_app.config[str('MEMORY_CACHE_CONFIGURATION')].get(str('TIMEOUT_AUTHENTICATION'), 30)
+    )
+    @cache.memoize(current_app.config[str('CACHE_CONFIGURATION')].get(str('TIMEOUT_AUTHENTICATION'), 300))
```
pbench, line 79: Out of curiosity : why do you have two different cache decorator ?

xlqian: there are two levels of Cache, one cache is at the `memory level` and another is at the redis level. This kind of setting is for the sake of saving network load towards redis.

pbench (approved):

PR #3855, opened 2022-11-15
[AWS]ENABLE_USER_EVENT parameter added

azime:


source/tyr/tyr/default_settings.py
```diff
@@ -165,3 +165,5 @@
 POST_DATA_TO_TYR = os.getenv('TYR_POST_DATA_TO_TYR', None)
 
 USE_LOCAL_SYS_LOG = os.getenv('TYR_USE_LOCAL_SYS_LOG', 'true').lower() in ['1', 'true', 'yes']
+
+ENABLE_USER_EVENT = os.getenv('TYR_ENABLE_USER_EVENT', True)
```
pbench:
```suggestion
ENABLE_USER_EVENT = os.getenv('TYR_ENABLE_USER_EVENT', 'true').lower() in ['1', 'true', 'yes']
```
since `os.getenv` returns a string see https://docs.python.org/3/library/os.html#os.getenv

pbench (approved):

source/tyr/tyr/default_settings.py
azime: Fixed

woshilapin (approved):

PR #3867, opened 2022-12-06
[jormungandr] Use transfer pool

xlqian:
centralize all logics of transfer-computing into  class `TransferPool` in order to ease the implementation of ACCESS_POINT_TRANSFER

### DIFF
--- a/source/jormungandr/jormungandr/scenarios/helper_classes/helper_utils.py
+++ b/source/jormungandr/jormungandr/scenarios/helper_classes/helper_utils.py
@@ -49,13 +49,6 @@ from contextlib import contextmanager
 import time
 
 CAR_PARK_DURATION = 300  # secs
-allowed_physical_mode_for_transfert_path = (
-    'physical_mode:Bus',
-    'physical_mode:Tramway',
-    'physical_mode:BusRapidTransit',
-    'physical_mode:Coach',
-    'physical_mode:Shuttle',
-)
 
 
 def _create_crowfly(pt_journey, crowfly_origin, crowfly_destination, begin, end, mode):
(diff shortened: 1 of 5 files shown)


source/jormungandr/jormungandr/scenarios/helper_classes/helper_utils.py
```diff
@@ -834,15 +762,7 @@ def complete_transfer(pt_journey, transfer_path_pool, request, transfer_sections
...
-            s.origin, s.destination, real_mode, fallback_extremity, direct_path_type, request
-        )
-
-        if transfer_journeys and transfer_journeys.journeys:
-            new_section = transfer_journeys.journeys[0].sections
-            s.street_network.CopyFrom(new_section[0].street_network)
+    for section in pt_journey.sections:
+        if section.type != response_pb2.TRANSFER:
+            continue
+        transfer_pool.wait_and_complete(section)
```
woshilapin, line 768: How about this instead?
```suggestion
        if section.type == response_pb2.TRANSFER:
            transfer_pool.wait_and_complete(section)
```

woshilapin (approved):
Do we have tests around this functionality? It'd be nice to see that there is no regression after this refactorization.

xlqian:
@woshilapin yep, there are a bunch of functional tests around on this feature

source/jormungandr/jormungandr/scenarios/helper_classes/helper_utils.py
xlqian: it's more like a personal choice :smile_cat: I did so coz I hate indentations in codes...

woshilapin: It's one more line and the indentation is there anyway :shrug: Let's stop doing Python if indentation is a problem :stuck_out_tongue:

source/jormungandr/jormungandr/scenarios/helper_classes/transfer.py
```diff
@@ -0,0 +1,143 @@
...
+
+ACCESS_POINTS_PHYSICAL_MODES = ("physical_mode:RapidTransit",)
+
+NO_ACCESS_POINTS_TRANSFER = set(
+    itertools.product(NO_ACCESS_POINTS_PHYSICAL_MODES, NO_ACCESS_POINTS_PHYSICAL_MODES)
+)
+
+ACCESS_POINTS_TRANSFER = set(
+    itertools.product(ACCESS_POINTS_PHYSICAL_MODES, NO_ACCESS_POINTS_PHYSICAL_MODES)
+) | set(itertools.product(NO_ACCESS_POINTS_PHYSICAL_MODES, ACCESS_POINTS_PHYSICAL_MODES))
```
pbench, line 58: Is `set(A) | set(B)` the union of `set(A)` and `set(B)` ?

source/jormungandr/jormungandr/scenarios/helper_classes/transfer.py
```diff
@@ -0,0 +1,143 @@
...
+    'physical_mode:Shuttle',
+)
+
+ACCESS_POINTS_PHYSICAL_MODES = ("physical_mode:RapidTransit",)
+
+NO_ACCESS_POINTS_TRANSFER = set(
+    itertools.product(NO_ACCESS_POINTS_PHYSICAL_MODES, NO_ACCESS_POINTS_PHYSICAL_MODES)
+)
+
+ACCESS_POINTS_TRANSFER = set(
```
pbench:
```suggestion
# if `(physical_mode:A, physical_mode:B) in NO_ACCESS_POINTS_TRANSFER` then it means that a transfer 
# where we get out of a vehicle of  `physical_mode:A` and then get in a vehicle of `physical_mode:B` 
# **will not** go through an access point
NO_ACCESS_POINTS_TRANSFER = set(
    itertools.product(NO_ACCESS_POINTS_PHYSICAL_MODES, NO_ACCESS_POINTS_PHYSICAL_MODES)
)
# if `(physical_mode:A, physical_mode:B) in ACCESS_POINTS_TRANSFER` then it means that a transfer 
# where we get out of a vehicle of  `physical_mode:A` and then get in a vehicle of `physical_mode:B` 
# **may use** an access point
ACCESS_POINTS_TRANSFER = set(
```
I had some trouble understanding what these sets are. Are my comments right ?

xlqian: You are doing a better job than  `OpenAI`!!!!

pbench (approved):

PR #3870, opened 2022-12-07
healthcheck returns 503 when not ready

pbench:
The /ready route of the monitor now returns a 503 instead of a 400 when kraken is not ready


woshilapin (approved):

pbench:
already done in https://github.com/hove-io/navitia/pull/3869

PR #3872, opened 2022-12-08
[jormungandr] implement access_points in transfer path

xlqian:
With this PR. transfer paths are allowed between surface transport vehicles (`tramway`, `bus` etc...) and underground transport vehicles (`metro` etc...) **_via_** access points. 

Note that : 
* no transfer path will be calculated if access points are not available for the underlying underground transport vehicles. 
* no transfer path will be calculated if both stop points of the transfer are underground. (this case may be handled later)

How the access point (if multiple access points are present for the underlying stop point) is picked in a nutshell:

Let's say we need to calculate the path **from**  `stop_point:Tramway` **to** `stop_point: Metro`, given that  `stop_point: Metro` has 3 access points around it. The first step is to choose the eligible access points by requesting the `georef` all the access points of `stop_point:Metro` and filtering out those are `is_entrance == False`. Then we compute a matrix of `1xN` (reciprocally `Nx1` when transferring from `stop_point:Metro` to `stop_point:Tramway`). By adding up the `traversal_time` of each access points to the street network duration, we pick the most rapid access point to get to  `stop_point:Metro`. Then we launch the real street network computation between `stop_point:Tramway` and the chosen access point.

In the following example, the blue line (underground transport) has multiple access points,
image

but we finally chose the best/most rapid access point to get to the platform
image

### DIFF
--- a/source/jormungandr/jormungandr/scenarios/helper_classes/helper_utils.py
+++ b/source/jormungandr/jormungandr/scenarios/helper_classes/helper_utils.py
@@ -779,20 +779,27 @@ def complete_transfer(pt_journey, transfer_pool):
         transfer_pool.wait_and_complete(section)
 
 
-def prepend_first_and_append_last_coord(dp, origin, destination):
-    """
-    Insert/append origin and destination's coordinates to the geojson
-    """
+def is_valid_direct_path_streetwork(dp):
     if not dp or not dp.journeys or not dp.journeys[0].sections:
-        return
+        return False
 
+    # a valid journey's must comprise at least two coordinates
     nb_coords = sum((len(sec.street_network.coordinates) for sec in dp.journeys[0].sections))
     if nb_coords < 2:
+        return False
+    return True
+
+
+def prepend_first_coord(dp, pt_obj):
+    """
+    prepend pt_object's coord to journeys' coordinates (geojson)
+    """
+    if not is_valid_direct_path_streetwork(dp):
         return
 
     starting_coords = dp.journeys[0].sections[0].street_network.coordinates
     # we are inserting the coord of the origin at the beginning of the geojson
-    coord = get_pt_object_coord(origin)
+    coord = get_pt_object_coord(pt_obj)
     if starting_coords and coord != starting_coords[0]:
         starting_coords.add(lon=coord.lon, lat=coord.lat)
         # we cannot insert an element at the beginning of a list :(
@@ -803,8 +810,16 @@ def prepend_first_and_append_last_coord(dp, origin, destination):
             starting_coords[i].CopyFrom(starting_coords[-1])
             starting_coords[-1].CopyFrom(tmp)
 
+
+def append_last_coord(dp, pt_obj):
+    """
+    append pt_object's coord to journeys' coordinates (geojson)
+    """
+    if not is_valid_direct_path_streetwork(dp):
+        return
+
     ending_coords = dp.journeys[0].sections[-1].street_network.coordinates
     # we are appending the coord of the destination at the end of the geojson
-    coord = get_pt_object_coord(destination)
+    coord = get_pt_object_coord(pt_obj)
     if ending_coords and coord != ending_coords[-1]:
         ending_coords.add(lon=coord.lon, lat=coord.lat)
--- a/source/jormungandr/jormungandr/scenarios/helper_classes/helper_utils.py
+++ b/source/jormungandr/jormungandr/scenarios/helper_classes/helper_utils.py
@@ -35,7 +35,6 @@ from jormungandr.utils import (
     SectionSorter,
     get_pt_object_coord,
     generate_id,
-    get_overriding_mode,
 )
 from jormungandr.street_network.utils import crowfly_distance_between
 from jormungandr.fallback_modes import FallbackModes, all_fallback_modes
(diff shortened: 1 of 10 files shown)


source/ptreferential/tests/ptref_test.cpp
```diff
@@ -154,7 +154,7 @@ BOOST_AUTO_TEST_CASE(get_indexes_test) {
         b.connection("stop3", "stop2", 10 * 60);
     });
 
-    // On cherche à retrouver la ligne 1, en passant le stoparea en filtre
+    // On cherche �� retrouver la ligne 1, en passant le stoparea en filtre
```
woshilapin: Mmmm, probably revert the entire file below this line?

source/routing/tests/routing_api_test_data.h
```diff
@@ -282,7 +282,7 @@ struct routing_api_data {
         b.data->geo_ref->ways.push_back(way);
 
         way = new navitia::georef::Way();
-        way->name = "rue ts Entrée/Sortie";  // T->S
+        way->name = "rue ts Entr��e/Sortie";  // T->S
```
woshilapin: Probably revert this.

source/ptreferential/tests/ptref_test.cpp
xlqian: :man_facepalming:  My editor loves formatting everything to utf-8

source/jormungandr/jormungandr/scenarios/helper_classes/transfer.py
```diff
@@ -131,19 +150,216 @@ def _do_no_access_point_transfer(self, section):
...
 
+        if direct_path and direct_path.journeys:
+            return TransferResult(direct_path, section.origin, section.destination)
+
+        return None
+
     def _aysnc_no_access_point_transfer(self, section):
         return self._future_manager.create_future(self._do_no_access_point_transfer, section)
 
+    def _get_access_points(self, stop_point_uri, predicator=lambda x: x):
```
pbench:
```suggestion
    def _get_access_points(self, stop_point_uri, access_point_filter=lambda x: x):
```

source/jormungandr/jormungandr/scenarios/helper_classes/transfer.py
```diff
@@ -131,19 +150,216 @@ def _do_no_access_point_transfer(self, section):
...
+        """
+        if prev_section_mode in ACCESS_POINTS_PHYSICAL_MODES:
+            return self._get_access_points(
+                section.origin.uri, predicator=lambda access_point: access_point.is_exit
+            )
+        elif next_section_mode in ACCESS_POINTS_PHYSICAL_MODES:
+            return self._get_access_points(
+                section.destination.uri, predicator=lambda access_point: access_point.is_entrance
+            )
+
```
pbench, line 189:
```suggestion
            )
     else:
           return None

```

source/jormungandr/jormungandr/scenarios/helper_classes/transfer.py
```diff
@@ -131,19 +150,216 @@ def _do_no_access_point_transfer(self, section):
...
+            FallbackModes.walking.name,
+            section.duration * 2,
+            self._request,
+            sub_request_id,
+        )
+
+        # now it's time to find the best combo
+        # (stop_point -> access_points or access_points -> stop_point)
+        best_access_point = self.determinate_the_best_access_point(routing_matrix, access_points)
+
```
pbench, line 287:
```suggestion
if best_access_point is None:
    return None
```

source/jormungandr/jormungandr/scenarios/helper_classes/helper_utils.py
```diff
@@ -766,3 +777,34 @@ def complete_transfer(pt_journey, transfer_pool):
         if section.type != response_pb2.TRANSFER:
             continue
         transfer_pool.wait_and_complete(section)
+
+
+def prepend_first_and_append_last_coord(dp, origin, destination):
```
pbench: Maybe split this function in two `prepend_first_coord()` and `append_last_coord()` ?

source/jormungandr/jormungandr/scenarios/helper_classes/helper_utils.py
```diff
@@ -766,3 +777,34 @@ def complete_transfer(pt_journey, transfer_pool):
...
+    """
+    Insert/append origin and destination's coordinates to the geojson
+    """
+    if not dp or not dp.journeys or not dp.journeys[0].sections:
+        return
+
+    nb_coords = sum((len(sec.street_network.coordinates) for sec in dp.journeys[0].sections))
```
pbench, line 789: I don't really understand this check. Maybe add some comments ?

source/jormungandr/jormungandr/scenarios/helper_classes/helper_utils.py
```diff
@@ -766,3 +777,34 @@ def complete_transfer(pt_journey, transfer_pool):
...
+    if nb_coords < 2:
+        return
```
pbench: I don't really understand this check. Maybe add some comments ?

xlqian: it just checks the length of the `geojson` of the street network, there should be at least 2 points in a `geojson` logically

source/jormungandr/jormungandr/scenarios/helper_classes/helper_utils.py
pbench: And don't you already have the exact same function [here](https://github.com/hove-io/navitia/blob/fca4622c715418b16686de7a106d836ff1fd4179/source/jormungandr/jormungandr/scenarios/helper_classes/streetnetwork_path.py#L104) ?

pbench (approved):

PR #3874, opened 2022-12-09
[jormungandr] fix zmq socket lifetime

xlqian:
the lifetime of the zmq socket was based on the time of the last successful request, which extends its lifetime in practice. 

In this PR, zmq socket lifetime will be based on its spawn time, which means once the TTL is out, zmq socket will be closed and removed from the socket pool.

### DIFF
--- a/source/jormungandr/jormungandr/instance.py
+++ b/source/jormungandr/jormungandr/instance.py
@@ -745,13 +745,12 @@ class Instance(object):
 
     @contextmanager
     def socket(self, context):
-
-        socket = None
         sockets = self._sockets
         socket_path = self.socket_path
 
+        t = None
         try:
-            socket, _ = sockets.pop()
+            socket, t = sockets.pop()
         except IndexError:  # there is no socket available: lets create one
             socket = context.socket(zmq.REQ)
             socket.connect(socket_path)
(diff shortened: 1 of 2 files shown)


azime (approved):

source/jormungandr/jormungandr/instance.py
```diff
@@ -745,21 +745,21 @@ def _reap_sockets(connector):
...
-            socket, _ = sockets.pop()
+            socket, t = sockets.pop()
         except IndexError:  # there is no socket available: lets create one
             socket = context.socket(zmq.REQ)
             socket.connect(socket_path)
+
         try:
             yield socket
         finally:
             if not socket.closed:
```
pbench, line 760: Maybe we should also test here if the TTL is expired, and close the socket if this is the case ?
Otherwise we may put back in the pool a socket with an expired TTL, which will be used before the reaper close it

xlqian: I agree with you, that's what has been done in asgard... I looked into the codes, there will be some work to do.. and I don't want to complicate this PR too much...

pbench: That is not just adding here something like 
```
if time.now() > t:
   socket.close()
```
?

pbench (approved):

PR #3883, opened 2022-12-20
[Jormungandr] Use transient socket for instance, pt planner and asgard, AND remove socket reaper

xlqian:
Use transient socket for all and remove socket reaper


source/jormungandr/jormungandr/__init__.py
```diff
@@ -128,5 +125,4 @@ def setup_package():
 
 from jormungandr import transient_socket
 
-transient_socket.TransientSocket.init_socket_reaper(app.config)
```
xlqian, line 131: :tada:

source/jormungandr/jormungandr/transient_socket.py
```diff
@@ -61,107 +60,89 @@ class TransientSocket(object):
...
+        self._zmq_socket = zmq_socket
         self.ttl = socket_ttl
+        self.semaphore = BoundedSemaphore(1)
+        self._sockets = SortedList([], key=lambda s: -s.t)
+
+    def make_new_socket(self):
+        start = time.time()
+        socket = self._zmq_context.socket(zmq.REQ)
+        socket.connect(self._zmq_socket)
+        self._logger.info(
```
pbench:
```suggestion
        self._logger.debug(
```
info is a bit too verbose for this kind of logs no ?

pbench (approved):
great cleaning !

PR #3885, opened 2022-12-21
[ci] ci: allow to release through Github Actions

woshilapin:
Let's try to make our life easier by allowing to make a release from Github Actions.

Everything should be triggered by the creation of a **tag** (which will be created when creating a Github Release).

This is an opinionated modification, since we won't use anymore the script for release. The script was essentially doing 2 things:
1. creates branches `release_x.y.z` and `release` and merge them into each other in a gigantic plate of spaghetti of branch merges
2. help for generating the changelog

For 1., I believe this is useless, let's tag directly the commit we estimate correct without merging forward and backward some branches (usually, the said commit is the last one `dev`).
For 2., Github releases can automatically generate the changelog :tada: note however that the changelog will be generated in Github, not anymore in `debian/changelog` (but since we also talk about not generating Debian package anymore... :shrug: ).

- [ ] need to merge https://github.com/hove-io/navitia-docker-compose/pull/151

Note: I removed the constraint `branches: [release, dev]` on some `workflow_run`, but they shouldn't break anything since they both depend on 2 jobs that already are triggered by either one of these branches (but now tags too).

### DIFF
--- a/.github/workflows/build_navitia_packages_for_release.yml
+++ b/.github/workflows/build_navitia_packages_for_release.yml
@@ -82,7 +82,7 @@ jobs:
         path: ./builder_from_package
     - name: build, create and publish images for branch release
       working-directory: builder_from_package
-      run: ./build.sh -o ${{secrets.access_token_github}} -d debian8 -n -r -e push -t latest -b release -u ${{secrets.docker_user}} -p ${{secrets.docker_password}}
+      run: ./build.sh -o ${{secrets.access_token_github}} -d debian8 -n -r -e push -t latest -b ${GITHUB_REF_NAME} -u ${{secrets.docker_user}} -p ${{secrets.docker_password}}
     - name: slack notification (the job has failed)
       if: failure()
       run: |
--- a/.github/workflows/build_navitia_packages_for_release.yml
+++ b/.github/workflows/build_navitia_packages_for_release.yml
@@ -2,6 +2,8 @@ name: Build Navitia Packages For Release
 
 on:
   push:
+    tags:
+      - '*'
     branches:
       - release
 
--- a/.github/workflows/publish_hove_images_aws.yml
+++ b/.github/workflows/publish_hove_images_aws.yml
@@ -18,7 +18,6 @@ on:
 
   workflow_run:
     workflows: ["Build Navitia Packages For Release", "Build Navitia Packages For Dev Multi Distributions"]
-    branches: [release, dev]
     types:
       - completed
 
--- a/.github/workflows/publish_hove_images_aws.yml
+++ b/.github/workflows/publish_hove_images_aws.yml
@@ -34,13 +34,13 @@ jobs:
           then
             echo "Workflow triggered due to finished upstream workflow"
             echo "::set-output name=branch::${{ github.event.workflow_run.head_branch }}"
-            if [[ '${{ github.event.workflow_run.head_branch }}' == 'dev' ]]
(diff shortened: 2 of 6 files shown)


.github/workflows/build_navitia_packages_for_release.yml
```diff
@@ -2,6 +2,8 @@ name: Build Navitia Packages For Release
 
 on:
   push:
+    tags:
+      - '*'
     branches:
       - release
```
pbench, line 8:
```suggestion
      - '*'
    branches:
      - dev
```
If we use github to make releases, it means that the `release` branch will become useless, right ? If yes, then I think we  need to run this workflow when new tags are pushed to the `dev` branch

woshilapin: Yes, totally right. But I intentionally keep everything in place before testing and validating this new workflow. Once done and tested, I'll prepare another PR to clean up all of that stuff.

pbougue:
The [questions about release-type](https://github.com/hove-io/navitia/blob/a24c5a97e484fd24da1e257b1e7c4ed58987dcea/release/script_release.py#L407-L417) were a little bugged (especially in case of hotfix, that can be backward compatible, or require a bina as well), but they were quite important, and we probably need to keep them.
But I don't know how bina is handled on delivery now that we don't use fabric (if it's always triggered, it would be less important).

woshilapin:
@pbench Thanks for the warnings. I've made this 9a8fa3fbb57278a5ec640bd17c4f1e84bc672e28 and the PR https://github.com/hove-io/navitia-docker-compose/pull/151 that should help (and also, but optional, https://github.com/hove-io/core_team_ci_tools/pull/10).

pbougue (approved):
It looks alright but I'm not sharp at all on github workflows here, so my review is not worth much here.

(reminder on trying to change filtering done on "not dev branch" to "version tag")

And then next job should cleanup and (re)write doc on how to release: questions on data version, web api retro-compat, how to pick semver, how to editorialize changelog...

azime (approved):

woshilapin:
@vpassama We're planning to change how releases are done. This PR is only changing how the CI works, but the result should be similar (Debian packages, Docker images). Later, we might completely drop the Debian packages entirely. If these are useful for you, this is the moment to shout out!

woshilapin:
For information, I've added 3538a797d for making `release` a specific case and the rest the default (`dev`).

.github/workflows/publish_hove_images_aws.yml
```diff
@@ -35,13 +34,13 @@ jobs:
...
-            if [[ '${{ github.event.workflow_run.head_branch }}' == 'dev' ]]
+            if [[ '${{ github.event.workflow_run.head_branch }}' == 'release' || '${{ github.event.workflow_run.head_branch }}' =~ v[0-9]+.[0-9]+.[0-9]+ ]]
             then
-              echo "::set-output name=tag::dev"
-            else :
-              echo "Realease branch, collecting last version"
+              echo "Release branch, collecting last version"
               VERSION=$(curl https://api.github.com/repos/hove-io/navitia/tags | jq --raw-output '.[0].name')
               echo "::set-output name=tag::$VERSION"
+            else :
```
pbougue, line 42: Should you also test that you're on `dev` here?
To avoid pushing something else, not sure.

woshilapin: I would say that's a feature more than a risk. Someone with a custom branch that would be able to trigger this part of code, might actually want this to be published in order to test... :shrug: Let's give it a chance :smile:

pbench (approved):

PR #3893, opened 2023-01-05
Jormun: Feat avoid pg authentication for bad token

kadhikari:
Some details of modifications:
- When no token provided (even with some empty spaces), no need to verify can_connect_to_database()
- Since we cache token - Object User, for a bad token (token not in database) we get a default user with type=no_access and cache it. There will be no access to the database during cache for this token.
- When database in not accessible, all the requests with a token (bad or good) will have access to all the apis when cache for each token is renewed.

For more details: https://navitia.atlassian.net/browse/NAV-1642

### DIFF
--- a/source/jormungandr/jormungandr/api.py
+++ b/source/jormungandr/jormungandr/api.py
@@ -90,19 +90,21 @@ def add_info_newrelic(response, *args, **kwargs):
         record_custom_parameter('navitia-request-id', request.id)
 
         token = get_token()
+        # No log will be added in newrelic if token is absent in the request
         if token:
             user = get_user(token=token, abort_if_no_token=False)
             if user:
                 record_custom_parameter('user_id', str(user.id))
-            # This method verifies database connection and gets object Key only once when cache expires.
-            app_name = get_app_name(token)
-            if app_name:
-                record_custom_parameter('token_name', app_name)
+                # This method verifies database connection and gets object Key only once when cache expires.
+                app_name = get_app_name(token)
+                if app_name:
+                    record_custom_parameter('token_name', app_name)
 
-        record_custom_parameter('version', __version__)
-        coverages = get_used_coverages()
-        if coverages:
-            record_custom_parameter('coverage', coverages[0])
+            record_custom_parameter('version', __version__)
+            # No access to database required
+            coverages = get_used_coverages()
+            if coverages:
+                record_custom_parameter('coverage', coverages[0])
     except:
         logger = logging.getLogger(__name__)
         logger.exception('error while reporting to newrelic:')
--- a/source/jormungandr/jormungandr/api.py
+++ b/source/jormungandr/jormungandr/api.py
@@ -90,13 +90,14 @@ def add_info_newrelic(response, *args, **kwargs):
     try:
         record_custom_parameter('navitia-request-id', request.id)
 
-        if can_connect_to_database():
-            token = get_token()
+        token = get_token()
+        if token:
             user = get_user(token=token, abort_if_no_token=False)
-            app_name = get_app_name(token)
             if user:
                 record_custom_parameter('user_id', str(user.id))
-            record_custom_parameter('token_name', app_name)
+            app_name = get_app_name(token)
+            if app_name:
+                record_custom_parameter('token_name', app_name)
 
         record_custom_parameter('version', __version__)
         coverages = get_used_coverages()
--- a/source/jormungandr/jormungandr/api.py
+++ b/source/jormungandr/jormungandr/api.py
@@ -42,7 +42,6 @@ import logging
 from jormungandr.new_relic import record_custom_parameter
 from jormungandr.authentication import get_user, get_token, get_app_name, get_used_coverages
 from jormungandr._version import __version__
-from jormungandr.utils import can_connect_to_database
 import six
 
 
@@ -95,6 +94,7 @@ def add_info_newrelic(response, *args, **kwargs):
             user = get_user(token=token, abort_if_no_token=False)
             if user:
                 record_custom_parameter('user_id', str(user.id))
+            # This method connects and gets object Key only once when cache expires.
             app_name = get_app_name(token)
             if app_name:
                 record_custom_parameter('token_name', app_name)
--- a/source/jormungandr/jormungandr/api.py
+++ b/source/jormungandr/jormungandr/api.py
@@ -94,7 +94,7 @@ def add_info_newrelic(response, *args, **kwargs):
             user = get_user(token=token, abort_if_no_token=False)
             if user:
                 record_custom_parameter('user_id', str(user.id))
-            # This method connects and gets object Key only once when cache expires.
+            # This method verifies database connection and gets object Key only once when cache expires.
             app_name = get_app_name(token)
             if app_name:
                 record_custom_parameter('token_name', app_name)
--- a/source/jormungandr/jormungandr/authentication.py
+++ b/source/jormungandr/jormungandr/authentication.py
@@ -141,6 +141,7 @@ def has_access(region, api, abort, user):
     # if jormungandr is on public mode or database is not accessible, we skip the authentication process
     logging.getLogger(__name__).debug('User "has_access" to region/api not cached')
 
+    # Connection to database verified only once when cache expires.
     if current_app.config.get('PUBLIC', False) or (not can_connect_to_database()):
         return True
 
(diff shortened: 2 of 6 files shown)


source/tyr/migrations/versions/04f9b89e3ef5_add_no_access_in_user_type.py
```diff
@@ -0,0 +1,47 @@
...
+    op.execute('ALTER TABLE "user" ALTER COLUMN type DROP DEFAULT')
+    op.execute('ALTER TABLE "user" ALTER COLUMN type TYPE _user_type USING type::text::_user_type')
+    old_type.drop(op.get_bind(), checkfirst=False)
+
+    new_type.create(op.get_bind(), checkfirst=False)
+    op.execute('ALTER TABLE "user" ALTER COLUMN type TYPE user_type USING type::text::user_type')
+    op.execute('ALTER TABLE "user" ALTER COLUMN type SET DEFAULT \'with_free_instances\'')
+    tmp_type.drop(op.get_bind(), checkfirst=False)
+    op.execute(
+        'INSERT INTO "user"(login, email, type) values(\'user_without_access\', \'user_without_access@noreply.com\', \'no_access\')'
```
woshilapin, line 26: Is it possible that this email is used to send information? I'm asking because we've seen abnormal level of incorrect email sent causing problem to the SMTP service.

source/tyr/migrations/versions/04f9b89e3ef5_add_no_access_in_user_type.py
```diff
@@ -0,0 +1,47 @@
...
+new_options = ('with_free_instances', 'without_free_instances', 'super_user', 'no_access')
+old_type = sa.Enum('with_free_instances', 'without_free_instances', 'super_user', name='user_type')
+new_type = sa.Enum(*new_options, name='user_type')
+tmp_type = sa.Enum(*new_options, name='_user_type')
+
+
+def upgrade():
+    tmp_type.create(op.get_bind(), checkfirst=False)
+    op.execute('ALTER TABLE "user" ALTER COLUMN type DROP DEFAULT')
+    op.execute('ALTER TABLE "user" ALTER COLUMN type TYPE _user_type USING type::text::_user_type')
```
woshilapin: What is the purpose of this line (and line 29) ?

source/navitiacommon/navitiacommon/models/__init__.py
```diff
@@ -194,6 +194,11 @@ def get_from_key(cls, token, valid_until):
         )
         return query.first()
 
+    @classmethod
+    def get_without_access(cls):
+        query = cls.query.filter(cls.type == 'no_access')
+        return query.first()
```
woshilapin: I'm not sure: does this means that we have only one and only one line `no_access` in the DB? If yes, can we enforce this? If we cannot enforce, we might want to document it somewhere at least?

source/jormungandr/jormungandr/authentication.py
```diff
@@ -211,6 +213,7 @@ def uncached_get_user(token):
 )
 @cache.memoize(current_app.config[str('CACHE_CONFIGURATION')].get(str('TIMEOUT_AUTHENTICATION'), 300))
 def cache_get_key(token):
+    # This verification is done only once when cache expires.
```
woshilapin, line 217: :heart:

woshilapin (approved):

source/tyr/migrations/versions/04f9b89e3ef5_add_no_access_in_user_type.py
```diff
@@ -0,0 +1,47 @@
...
+
+new_options = ('with_free_instances', 'without_free_instances', 'super_user', 'no_access')
+old_type = sa.Enum('with_free_instances', 'without_free_instances', 'super_user', name='user_type')
+new_type = sa.Enum(*new_options, name='user_type')
+tmp_type = sa.Enum(*new_options, name='_user_type')
+
+
+def upgrade():
+    tmp_type.create(op.get_bind(), checkfirst=False)
+    op.execute('ALTER TABLE "user" ALTER COLUMN type DROP DEFAULT')
```
pbougue: 1. On upgrade, you could more safely use `ALTER TYPE user_type ADD VALUE IF NOT EXISTS no_access`.
2. And if you keep it that way (at least for downgrade, below), I'm not sure why you need to DROP DEFAULT.

source/tyr/tyr/resources.py
```diff
@@ -1239,9 +1239,9 @@ def post(self):
             type=str,
             required=False,
             default='with_free_instances',
-            help='type of user: [with_free_instances, without_free_instances, super_user]',
+            help='type of user: [with_free_instances, without_free_instances, super_user, no_access]',
             location=('json', 'values'),
-            choices=['with_free_instances', 'without_free_instances', 'super_user'],
+            choices=['with_free_instances', 'without_free_instances', 'super_user', 'no_access'],
```
pbougue: Do you really want to allow the creation of such a user manually ?
(to be removed above, too)

source/navitiacommon/navitiacommon/models/__init__.py
pbougue: It doesn't seem to match the fact that no_access user is allowed in tyr, indeed.
Maybe having a hardcoded no-access user would be safer/cleaner (not sure, with ORM).

source/jormungandr/jormungandr/api.py
```diff
@@ -90,13 +89,15 @@ def add_info_newrelic(response, *args, **kwargs):
...
+        if token:
             user = get_user(token=token, abort_if_no_token=False)
-            app_name = get_app_name(token)
             if user:
                 record_custom_parameter('user_id', str(user.id))
-            record_custom_parameter('token_name', app_name)
+            # This method verifies database connection and gets object Key only once when cache expires.
+            app_name = get_app_name(token)
+            if app_name:
+                record_custom_parameter('token_name', app_name)
```
pbougue: :heart: no dedicated check of DB for logs

pbougue (commented):
I don't know how authent works (what's carried by coverage, what's carried by user), so I'm not sharp on that.
I'd say that `can_connect_to_database` is probably used too much indeed, but I don't know for sure where it's important and performant to keep it.

source/tyr/migrations/versions/04f9b89e3ef5_add_no_access_in_user_type.py
kadhikari: There will be no stat for this user so no treatment for quota-manager and no send email.

source/tyr/migrations/versions/04f9b89e3ef5_add_no_access_in_user_type.py
kadhikari: Temporary user_type while adding or deleting an element in the existing user_type.

source/navitiacommon/navitiacommon/models/__init__.py
kadhikari: We will never add an user by tyr of this type with token and endpoints.
The user is added during migration and will be unique.
Any other solution ?

pbougue: Maybe just not allow this in tyr, then (cf. https://github.com/hove-io/navitia/pull/3893#discussion_r1062597105) ?

source/tyr/migrations/versions/04f9b89e3ef5_add_no_access_in_user_type.py
kadhikari: :+1:

source/tyr/tyr/resources.py
kadhikari: Non, it should be deleted.

source/navitiacommon/navitiacommon/models/__init__.py
kadhikari: add a filter on login = 'user_without_access'

source/jormungandr/jormungandr/api.py
```diff
@@ -90,18 +89,22 @@ def add_info_newrelic(response, *args, **kwargs):
...
+                # This method verifies database connection and gets object Key only once when cache expires.
+                app_name = get_app_name(token)
+                if app_name:
+                    record_custom_parameter('token_name', app_name)
+
+            record_custom_parameter('version', __version__)
+            # No access to database required
+            coverages = get_used_coverages()
+            if coverages:
+                record_custom_parameter('coverage', coverages[0])
```
pbougue, line 107: IMO, we should still track calls without token, as this may cost a little, and we wouldn't see that something is happening without a NR log, no?

pbougue: This is just extra-attribute to the transaction log, so it's gonna be logged anyway, we can forget this comment :+1:

source/tyr/tyr/resources.py
kadhikari: Done

pbougue (approved):

azime (approved):

PR #3894, opened 2023-01-05
Jormun: Fix journey's sections sorter

azime:
Jira: [NAV-1678](https://navitia.atlassian.net/browse/NAV-1678)
The sort of sections is random when two sections have the same departure_date_time and arrival_date_time

**Before correction:**
image

[NAV-1678]: https://navitia.atlassian.net/browse/NAV-1678?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/source/jormungandr/jormungandr/utils.py
+++ b/source/jormungandr/jormungandr/utils.py
@@ -458,6 +458,8 @@ PeriodExtremity = namedtuple('PeriodExtremity', ['datetime', 'represents_start']
 
 class SectionSorter(object):
     def __call__(self, a, b):
+        if a.begin_date_time == b.begin_date_time and a.end_date_time == b.end_date_time:
+            return -1 if a.destination.uri == b.origin.uri else 1
         if a.begin_date_time != b.begin_date_time:
             return -1 if a.begin_date_time < b.begin_date_time else 1
         else:
--- a/source/jormungandr/jormungandr/utils.py
+++ b/source/jormungandr/jormungandr/utils.py
@@ -458,12 +458,16 @@ PeriodExtremity = namedtuple('PeriodExtremity', ['datetime', 'represents_start']
 
 class SectionSorter(object):
     def __call__(self, a, b):
-        if a.begin_date_time == b.begin_date_time and a.end_date_time == b.end_date_time:
-            return -1 if a.destination.uri == b.origin.uri else 1
         if a.begin_date_time != b.begin_date_time:
             return -1 if a.begin_date_time < b.begin_date_time else 1
-        else:
+        elif a.end_date_time != b.end_date_time:
             return -1 if a.end_date_time < b.end_date_time else 1
+        elif a.destination.uri == b.origin.uri:
+            return -1
+        elif a.origin.uri == b.destination.uri:
+            return 1
+        else:
+            return 0
 
 
 def make_namedtuple(typename, *fields, **fields_with_default):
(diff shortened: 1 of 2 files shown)


source/jormungandr/jormungandr/utils.py
```diff
@@ -458,6 +458,8 @@ def decode_polyline(encoded, precision=6):
 
 class SectionSorter(object):
     def __call__(self, a, b):
+        if a.begin_date_time == b.begin_date_time and a.end_date_time == b.end_date_time:
+            return -1 if a.destination.uri == b.origin.uri else 1
         if a.begin_date_time != b.begin_date_time:
             return -1 if a.begin_date_time < b.begin_date_time else 1
         else:
```
pbougue: Smart to try to join sections :bulb: 
But how about 3 sections having the same values, or 2 "distant" sections? Maybe raising would be better :shrug: 

Anyway, more efficient (and clearer IMO):
```py
        if a.begin_date_time != b.begin_date_time:
            return -1 if a.begin_date_time < b.begin_date_time else 1
        elif a.end_date_time != b.end_date_time:
           return -1 if a.end_date_time < b.end_date_time else 1
        elif a.destination.uri == b.origin.uri:
           return -1
        elif a.origin.uri == b.destination.uri:
           return 1
        else:  # still not decided, but default to something stable. Or is there a return 0 possible ?
           return -1 if a.origin.uri < b.origin.uri else 1  # or should we explicitly raise?
```

woshilapin: I like your proposition better, it's clearer.

However, about you last comment, if I'm not wrong, I wouldn't recommend to return `0` (even if it's probably possible). We want a stable sorting and depending on the sorting algorithm, the original order might or might not be preserved.

By the way, this class is called `SectionSorter`, but I don't think the naming is correct, it doesn't sort, it only compares. How about renaming it `SectionComparator`?

pbench: Now I am wondering why we need to sort sections :thinking:

azime: @pbench In the protobuf you cannot add a section at the beginning

pbench (approved):

source/jormungandr/jormungandr/utils.py
pbougue: @woshilapin returning a raw value in the else case (no matter the value) is not stable, as it will depend on which is left or right member. Sorting on uri is stable but not relevant... I guess raising has my preference, but it's not strong.
:heavy_check_mark: Anyway, it's quite clear and way better now, so it's fine by me.

Naming proposition by @woshilapin is better but not mandatory IMO.
@pbench avoid that sorting by a stitching mechanism of section-groups would be more correct, but this is more work, so I'm OK to keep this.

pbougue (approved):

xlqian (approved):

source/jormungandr/jormungandr/utils.py
```diff
@@ -460,8 +460,14 @@ class SectionSorter(object):
...
         if a.begin_date_time != b.begin_date_time:
             return -1 if a.begin_date_time < b.begin_date_time else 1
-        else:
+        elif a.end_date_time != b.end_date_time:
             return -1 if a.end_date_time < b.end_date_time else 1
+        elif a.destination.uri == b.origin.uri:
+            return -1
+        elif a.origin.uri == b.destination.uri:
+            return 1
+        else:
```
woshilapin, line 469: I wonder if we should put a log in this `else` case since it's not supposed to happen. Might help to debug in case it actually happens.

woshilapin (approved):

source/jormungandr/jormungandr/utils.py
pbougue: Indeed, logging (warning or error) here would be nice :+1:

PR #3909, opened 2023-01-13
[Kraken] Fix aging problem when both kirin & chaos are active

xlqian:
https://navitia.atlassian.net/browse/NAV-1650

In this PR, several issues are tackled. 

1. when both kirin and chaos are active, we observed a slow memory leak and the disruption's integration speed is slowly exacerbated.  This is due to the infinitely growing uri of the `vehicle_journey` impacted by `RailSecion` and `LineSection`. During my test, I spotted  one of these gargantuan uri attained a length of 33998 characters.

~~2.   A trip may be impacted several times by the disruptions/trip_update from kirin, whereas only the last trip_update will be taken into account. In order to avoid applying trip_updates redundantly, aka faster disruption integration ratio, we just apply the newest trip_update by reversing the buffer and ignoring the older ones which have the same id.~~ To be tackled in another PR

3. Kraken didn't poll all messages from rabbitmq, even though the buffer was not yet full. (5000 by default)
image
This is due to the low timeout(100ms in prod) in `BasicConsumeMessage`. A fast way to fix that is to augment that timeout. The risk is that if disruptions are arriving every (timeout -1ms), in the worst case, we have to wait `buffer_size*(timeout -1ms)` to fill the buffer (ex: 5000*(500-1)ms ~= 25s) before kraken starts to handle them. The solution is to add another timeout to limit that waiting time.

### DIFF
--- a/source/kraken/apply_disruption.cpp
+++ b/source/kraken/apply_disruption.cpp
@@ -50,6 +50,9 @@ www.navitia.io
 #include <boost/range/algorithm_ext/erase.hpp>
 #include <boost/variant/apply_visitor.hpp>
 #include <boost/variant/static_visitor.hpp>
+#include <boost/uuid/uuid.hpp>
+#include <boost/uuid/uuid_generators.hpp>
+#include <boost/uuid/uuid_io.hpp>
 
 #include <algorithm>
 #include <utility>
@@ -98,12 +101,10 @@ nt::VehicleJourney* create_vj_from_old_vj(nt::MetaVehicleJourney* mvj,
     return new_vj;
 }
 
-std::string make_new_vj_uri(const std::string& mvj_uri,
-                            nt::RTLevel rt_level,
-                            int nb_rt_vj,
-                            const std::string& disruption_uri) {
-    return "vehicle_journey:" + mvj_uri + ":" + type::get_string_from_rt_level(rt_level) + ":"
-           + std::to_string(nb_rt_vj) + ":" + disruption_uri;
+std::string make_new_vj_uri(const nt::MetaVehicleJourney* mvj, nt::RTLevel rt_level) {
+    boost::uuids::random_generator gen;
+    return "vehicle_journey:" + mvj->uri + ":" + type::get_string_from_rt_level(rt_level) + ":"
+           + boost::uuids::to_string(gen());
 }
 
 struct apply_impacts_visitor : public boost::static_visitor<> {
@@ -326,7 +327,7 @@ struct add_impacts_visitor : public apply_impacts_visitor {
             }
 
             auto nb_rt_vj = mvj->get_rt_vj().size();
-            auto new_vj_uri = make_new_vj_uri(mvj->uri, rt_level, nb_rt_vj, impact->disruption->uri);
+            auto new_vj_uri = make_new_vj_uri(mvj, rt_level);
 
             std::vector<type::StopTime> stoptimes;  // we copy all the stoptimes
             for (const auto& stu : impact->aux_info.stop_times) {
@@ -499,7 +500,7 @@ struct add_impacts_visitor : public apply_impacts_visitor {
                 continue;
             }
             auto nb_rt_vj = mvj->get_vjs_at(rt_level).size();
-            auto new_vj_uri = make_new_vj_uri(mvj->uri, rt_level, nb_rt_vj, impact->disruption->uri);
+            auto new_vj_uri = make_new_vj_uri(mvj, rt_level);
 
             new_vp.days = new_vp.days & (vj->validity_patterns[rt_level]->days >> vj->shift);
 
@@ -567,7 +568,7 @@ struct add_impacts_visitor : public apply_impacts_visitor {
                 continue;
             }
             auto nb_rt_vj = mvj->get_vjs_at(rt_level).size();
-            auto new_vj_uri = make_new_vj_uri(mvj->uri, rt_level, nb_rt_vj, impact->disruption->uri);
+            auto new_vj_uri = make_new_vj_uri(mvj, rt_level);
 
             new_vp.days = new_vp.days & (vj->validity_patterns[rt_level]->days >> vj->shift);
 
@@ -722,7 +723,7 @@ struct add_impacts_visitor : public apply_impacts_visitor {
             mvj->push_unique_impact(impact);
 
             auto nb_rt_vj = mvj->get_vjs_at(rt_level).size();
-            auto new_vj_uri = make_new_vj_uri(mvj->uri, rt_level, nb_rt_vj, impact->disruption->uri);
+            auto new_vj_uri = make_new_vj_uri(mvj, rt_level);
 
             new_vp.days = new_vp.days & (vj->validity_patterns[rt_level]->days >> vj->shift);
 
--- a/source/kraken/apply_disruption.cpp
+++ b/source/kraken/apply_disruption.cpp
@@ -189,18 +189,6 @@ type::ValidityPattern compute_base_disrupted_vp(const std::vector<boost::posix_t
     return vp;
 }
 
-std::string concatenate_impact_uris(const nt::MetaVehicleJourney& mvj) {
-    std::stringstream impacts_uris;
-    for (auto& mvj_impacts : mvj.modified_by) {
-        if (auto i = mvj_impacts.lock()) {
-            if (impacts_uris.str().find(i->disruption->uri) == std::string::npos) {
-                impacts_uris << ":" << i->disruption->uri;
-            }
-        }
-    }
-    return impacts_uris.str();
-}
-
 nt::Route* get_or_create_route(const nt::disruption::Impact& impact, nt::PT_Data& pt_data) {
     nt::Network* network = pt_data.get_or_create_network("network:additional_service", "additional service");
     nt::CommercialMode* comm_mode =
--- a/source/kraken/apply_disruption.cpp
+++ b/source/kraken/apply_disruption.cpp
@@ -101,6 +101,18 @@ nt::VehicleJourney* create_vj_from_old_vj(nt::MetaVehicleJourney* mvj,
     return new_vj;
 }
 
+std::string concatenate_impact_uris(const nt::MetaVehicleJourney& mvj) {
+    std::stringstream impacts_uris;
+    for (auto& mvj_impacts : mvj.modified_by) {
+        if (auto i = mvj_impacts.lock()) {
+            if (impacts_uris.str().find(i->disruption->uri) == std::string::npos) {
+                impacts_uris << ":" << i->disruption->uri;
+            }
+        }
+    }
+    return impacts_uris.str();
+}
+
 std::string make_new_vj_uri(const nt::MetaVehicleJourney* mvj, nt::RTLevel rt_level) {
     boost::uuids::random_generator gen;
     return "vehicle_journey:" + mvj->uri + ":" + type::get_string_from_rt_level(rt_level) + ":"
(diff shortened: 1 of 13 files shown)


source/kraken/apply_disruption.cpp
```diff
@@ -152,7 +160,7 @@ struct apply_impacts_visitor : public boost::static_visitor<> {
         // we cannot ensure that all VJ of a MetaVJ are on the same route,
         // and since we want all actions to operate on MetaVJ, we collect all MetaVJ of the route
         // (but we'll change only the route's vj)
-        std::set<nt::MetaVehicleJourney*> mvjs;
+        std::unordered_set<nt::MetaVehicleJourney*> mvjs;
```
pbougue, line 164: This is just a minor perf improvement, right?
(avoid useless sorting, that's all)

source/kraken/apply_disruption.cpp
```diff
@@ -98,6 +98,14 @@ nt::VehicleJourney* create_vj_from_old_vj(nt::MetaVehicleJourney* mvj,
     return new_vj;
 }
 
+std::string make_new_vj_uri(const std::string& mvj_uri,
```
pbougue: Minor: it looks like this function can be a method of the visitor to mutualize the way params are processed.
Maybe only pass the meta-VJ and let the function extract uri and VJ-number?

source/kraken/apply_disruption.cpp
```diff
@@ -716,8 +722,7 @@ struct add_impacts_visitor : public apply_impacts_visitor {
             mvj->push_unique_impact(impact);
 
             auto nb_rt_vj = mvj->get_vjs_at(rt_level).size();
-            std::string new_vj_uri = "vehicle_journey:" + mvj->uri + ":" + type::get_string_from_rt_level(rt_level)
-                                     + ":" + std::to_string(nb_rt_vj) + concatenate_impact_uris(*mvj);
```
pbougue, line 720: This `concatenate_impact_uris()` should be removed, it's unused after that.

I guess the uri leak was on that one, right?

source/kraken/apply_disruption.cpp
```diff
@@ -716,8 +722,7 @@ struct add_impacts_visitor : public apply_impacts_visitor {
...
+            auto new_vj_uri = make_new_vj_uri(mvj->uri, rt_level, nb_rt_vj, impact->disruption->uri);
```
pbougue: Not sure at all: Aren't you afraid of a case where multiple impacts of the same disruption are applied to the same VJ?
I think of something like below.

1. First impact:
uri = vj:mvj_uri:adapted:`1`:disruption_uri

2. Second impact (if vj cleanup is done after):
uri = vj:mvj_uri:adapted:`2`:disruption_uri
cleanup vj:mvj_uri:adapted:`1`:disruption_uri

3. Third impact: 
uri = vj:mvj_uri:adapted:`2`:disruption_uri -> conflict

From the way VJ are created, it may be silently replacing the previous with the same uri, and it may work... Or not.
Should you at least log if there is an uri conflict?

Maybe I'm just being paranoïd (about RT in kraken, i am :exploding_head:)

source/kraken/configuration.cpp
```diff
@@ -92,7 +92,9 @@ po::options_description get_options_description(const boost::optional<std::strin
         ("BROKER.vhost", po::value<std::string>()->default_value("/"), "vhost for rabbitmq")
         ("BROKER.exchange", po::value<std::string>()->default_value("navitia"), "exchange used in rabbitmq")
         ("BROKER.rt_topics", po::value<std::vector<std::string>>(), "list of realtime topic for this instance")
-        ("BROKER.timeout", po::value<int>()->default_value(100), "timeout for maintenance worker in millisecond")
+        ("BROKER.timeout", po::value<int>()->default_value(200), "timeout for maintenance worker in millisecond")
+        ("BROKER.max_batch_nb", po::value<int>()->default_value(5000), "max size of the realtime message")
+        ("BROKER.retrieving_timeout", po::value<int>()->default_value(10000), "max duration the worker is going to spend when retrieving messages")
```
pbougue: Trying to clarify the naming a bit here.

Also, would probably lower the value, as it's most of the time OK already with a 100ms timeout.
But maybe your tests showed opposite conclusion, in this case sorry :pray: 

```suggestion
        ("BROKER.retrieving_total_timeout", po::value<int>()->default_value(1000), "max total duration the worker is going to spend when retrieving messages")
```

pbougue (commented):
:clap: A bit shocked that uri itself is the major part of the issue

:information_source: minor=feel free to disregard.

source/kraken/configuration.cpp
```diff
@@ -92,7 +92,9 @@ po::options_description get_options_description(const boost::optional<std::strin
         ("BROKER.vhost", po::value<std::string>()->default_value("/"), "vhost for rabbitmq")
         ("BROKER.exchange", po::value<std::string>()->default_value("navitia"), "exchange used in rabbitmq")
         ("BROKER.rt_topics", po::value<std::vector<std::string>>(), "list of realtime topic for this instance")
-        ("BROKER.timeout", po::value<int>()->default_value(100), "timeout for maintenance worker in millisecond")
+        ("BROKER.timeout", po::value<int>()->default_value(200), "timeout for maintenance worker in millisecond")
+        ("BROKER.max_batch_nb", po::value<int>()->default_value(5000), "max size of the realtime message")
```
pbougue:
```suggestion
        ("BROKER.max_batch_nb", po::value<int>()->default_value(5000), "max number of realtime messages retrieved in a batch")
```

source/kraken/configuration.cpp
xlqian: 1s is just too few given the size of messages from kirin, a message may take several ms to be polled from rabbitmq.

source/kraken/apply_disruption.cpp
xlqian: oh, my bad, it doesn't mess up the performance at all, I just came across this code and thought `unordered_set` was better...

source/kraken/apply_disruption.cpp
xlqian: You're right. 

During the very first test, I was even more aggressive: I removed disruption_uri from the new vj's URI. Kraken did crash.

Then I realized the very same scenario that you described beyond had happened.  The reason why it works with only disruption_uri is that in the same disruption, there **should not be more than one impact** that impacts the same object.  So the `disruption_id`s in step3 and step2 are different. 

Now your comment gave me a second thoughts. May be we should keep `concatenate_impact_uris(*mvj)` in new vj's uri.

xlqian: or for the sake of speed, we can just use an uuid...

pbougue (approved):

source/kraken/apply_disruption.cpp
```diff
@@ -98,6 +101,24 @@ nt::VehicleJourney* create_vj_from_old_vj(nt::MetaVehicleJourney* mvj,
     return new_vj;
 }
 
+std::string concatenate_impact_uris(const nt::MetaVehicleJourney& mvj) {
```
pbench: this function is not used anymore right ? Maybe remove it then ?

source/kraken/configuration.cpp
```diff
@@ -92,7 +92,9 @@ po::options_description get_options_description(const boost::optional<std::strin
         ("BROKER.vhost", po::value<std::string>()->default_value("/"), "vhost for rabbitmq")
         ("BROKER.exchange", po::value<std::string>()->default_value("navitia"), "exchange used in rabbitmq")
         ("BROKER.rt_topics", po::value<std::vector<std::string>>(), "list of realtime topic for this instance")
-        ("BROKER.timeout", po::value<int>()->default_value(100), "timeout for maintenance worker in millisecond")
+        ("BROKER.timeout", po::value<int>()->default_value(200), "timeout for maintenance worker in millisecond")
+        ("BROKER.max_batch_nb", po::value<int>()->default_value(5000), "max number of realtime messages retrieved in a batch")
+        ("BROKER.total_retrieving_timeout", po::value<int>()->default_value(10000), "max total duration the worker is going to spend when retrieving messages")
```
pbench: is this timeout in miliseconds ? seconds ?

source/kraken/maintenance_worker.cpp
```diff
@@ -375,6 +378,10 @@ void MaintenanceWorker::handle_rt_in_batch(const std::vector<AmqpClient::Envelop
         }
         LOG4CPLUS_TRACE(logger, "received entity: " << feed_message.DebugString());
         for (const auto& entity : feed_message.entity()) {
+            auto res = applied_visited_id.insert(entity.id());
+            if (!res.second) {
+                continue;
+            }
```
pbench:
```suggestion
            auto res = applied_visited_id.insert(entity.id());
            // an newer disruption with the same id has already been seen, so we can ignore this one
            if (!res.second) {
                continue;
            }
```

I am a bit worried about this  "ignore disruptions whose id has already been seen".

- do we receive kirin disruptions with the same id ? When this happens are we sure we only need to take into account the last one ? Can we have two disruptions with the same id but that affect different vj ? poke @pbougue 
- what about chaos disruptions ? In particular, when we cancel a disruption, don't we get the same id twice ?

pbougue: This may deserve a code-comment, as the 3 of us asked ourselves the question :smiley:

* For kirin this is guaranteed that entity ID is the same only when the VJ is the same, and that taking only the last in the queue is valid.
~Details: actually it's not so true with parallelism on kirin, but the guarantee is that it's gonna be as good as it was (nothing else can be used to decide which message is the last one currently).~ Actually, even with parallelism it's guaranteed (no concurrent processing on the same VJ in Kirin).
:bulb: This may deserve a comment in https://github.com/hove-io/chaos-proto/blob/master/kirin_proto_doc.rs ? (I can have a shot if you agree)

* For chaos, this was discussed with chaos team and seemed OK, but I don't know about that particular case, I let @xlqian reply if he knows.

pbench: Allright if this has been checked, it sounds good to me.
I do agree that some comments (here and in the proto doc) explaining this may be useful.

> For kirin this is guaranteed that entity ID is the same only when the VJ is the same

same VJ *and* same date ?

source/kraken/apply_disruption.cpp
xlqian: I thought I had removed it... :open_mouth:

source/kraken/maintenance_worker.cpp
pbougue: Yes, same VJ and same date :+1: 

:white_check_mark: I'll try to add a little something into kirin_proto_doc.rs

xlqian: @pbench Now I'm having a second thought with your comment and I think I made a mistake after thinking it through... 
@pbougue I'm going to remove this trick(reversing the vector) in this PR and open up another PR to tackle this problem so that in case I messed it up, we don't have to revert the whole thing

pbench (approved):

source/kraken/apply_disruption.cpp
pbougue: As discussed, I just add some blabla here (really tracked in JIRA: https://navitia.atlassian.net/browse/NAV-1800): this UUID generation is preventing from requesting a given RT-VJ from its id on another kraken of the same coverage.

source/kraken/maintenance_worker.cpp
pbougue: We also have to reverse the order of the entities in the message to be safe that we still have the same result (although it would be weird for someone to send multiple time the same entity in the same message).

pbougue: After discussion: this was reverted from this PR because of uncertainty around chaos.
Tracked in JIRA https://navitia.atlassian.net/browse/NAV-1878

PR #3910, opened 2023-01-13
[Kraken] fix erase vj from dataset

xlqian:
fix "impossible to find the vj" when erasing vj from dataset

### DIFF
--- a/source/kraken/apply_disruption.cpp
+++ b/source/kraken/apply_disruption.cpp
@@ -38,6 +38,7 @@ www.navitia.io
 #include "type/meta_data.h"
 #include "type/meta_vehicle_journey.h"
 #include "type/pt_data.h"
+#include "type/dataset.h"
 #include "utils/logger.h"
 #include "utils/map_find.h"
 #include "utils/functions.h"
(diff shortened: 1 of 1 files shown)


source/kraken/apply_disruption.cpp
```diff
@@ -401,6 +401,7 @@ struct add_impacts_visitor : public apply_impacts_visitor {
                 vj->name = mvj->get_base_vj().at(0)->name;
                 vj->headsign = mvj->get_base_vj().at(0)->headsign;
                 vj->dataset = mvj->get_base_vj().at(0)->dataset;
+                vj->dataset->vehiclejourney_list.insert(vj);
             } else {
```
pbougue, line 405: You should also do this in the "else" part, no?

And maybe also for the other functors, not only metaVJ/route, but RailSection, LineSection?

xlqian: for other functors, the `vj->dataset` is not even set... if it works for them until now, I'm not gonna change it

pbougue: But in the "else" part, dataset is provided, so adding it would probably improve erase (replace) of full trip-add (no base VJ)?
Not sure, though.

xlqian: when handling other kinds of impact, new vj's dataset is left `nullptr`, 
according to this code: https://github.com/hove-io/navitia/blob/dev/source/type/meta_vehicle_journey.cpp#L94, `erase` is not even performed...

source/kraken/apply_disruption.cpp
```diff
@@ -407,7 +412,12 @@ struct add_impacts_visitor : public apply_impacts_visitor {
 
                 // for protection, use the datasets[0]
                 // TODO : Create default data set
-                vj->dataset = pt_data.datasets[0];
+                if (!pt_data.datasets.empty()) {
+                    if (pt_data.datasets[0]) {
```
pbougue, line 416: Looks like you're checking almost twice the same thing, but if you want to keep double check it's probably more expressive like:
```suggestion
                if (!pt_data.datasets.empty() && pt_data.datasets[0] != nullptr) {
```

pbougue (approved):

source/kraken/apply_disruption.cpp
pbougue: There was a little misunderstanding on this comment that had 2 point:
* one on the "else" part of this code, which is handled with your new commit
* the other on other functors, where erase is not performed as you mentioned

So I resolve this.

PR #3912, opened 2023-01-19
The ghost word matchings should be case insensitive

kadhikari:
While deleting ghost words from query string it should be case insensitive.

For more details: https://navitia.atlassian.net/browse/NAV-1629

### DIFF
--- a/source/jormungandr/tests/places_tests.py
+++ b/source/jormungandr/tests/places_tests.py
@@ -499,12 +499,11 @@ class TestPlaces(AbstractTestFixture):
         assert status == 200
         assert "places" not in response
 
-        # Search with 'Gare de'
+        # Upper or lower case shouldn't impact result
         response, status = self.query_region("places?q=Gare du stop", check=False)
         assert status == 200
         assert len(response['places']) == 3
 
-        # Search with 'gare DES'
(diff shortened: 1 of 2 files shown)


source/jormungandr/tests/places_tests.py
```diff
@@ -499,6 +499,16 @@ def test_places_stop_areas(self):
         assert status == 200
         assert "places" not in response
 
+        # Search with 'Gare de'
```
pbougue: (Same than https://github.com/hove-io/navitia/pull/3912/files#diff-4b14f66a73ac88395fe8b96e60ad37d4b9c9cb059ec4271e165be3bbb41da66aR507): comment outdated (de/du)
```suggestion
        # Upper or lower case shouldn't impact result
```

pbougue (approved):

xlqian (approved):

source/jormungandr/tests/places_tests.py
kadhikari: :+1:

pbougue (approved):

wassimbenaissa:
Super nice, thanks @kadhikari :)

PR #3914, opened 2023-01-20
Revert "[Jormungandr] Use transient socket for instance, pt planner and asgard, AND remove socket reaper"

pbougue:
Reverts hove-io/navitia#3883

Just in case.


xlqian (approved):

pbougue:
Rebased onto v15.26.0 to avoid embedding any other PR merged since.

xlqian:
~Rebased onto v15.26.3, pre-release created v15.26.5.~

xlqian:
Release v15.26.4

pbougue:
After some tests and use without reverting, this revert should not be necessary, ever.

PR #3918, opened 2023-01-24
NR: fix send_and_receive

azime:
Jormun error:
image


pbench (approved):

pbougue (approved):
Side info: If we ever want to hide it in NR, we might have to provide to manually to NR's web-interface.
You can ask me or @stifoon, we did it for Kirin.

PR #3926, opened 2023-01-31
Replace _get_authorized_instances

azime:
Recover instances authorized for a user with a single database call, instead of one database call for each configured instance

### DIFF
--- a/source/jormungandr/jormungandr/instance_manager.py
+++ b/source/jormungandr/jormungandr/instance_manager.py
@@ -39,7 +39,7 @@ import glob
 import logging
 from jormungandr.protobuf_to_dict import protobuf_to_dict
 from jormungandr.exceptions import ApiNotFound, RegionNotFound, DeadSocketException, InvalidArguments
-from jormungandr.authentication import get_all_available_instances
+from jormungandr.authentication import abort_request, can_read_user
 from jormungandr import authentication, cache, app
 from jormungandr.instance import Instance
 import gevent
@@ -298,7 +298,7 @@ class InstanceManager(object):
         else:
             # Requests without any coverage
             # fetch all the authorized instances (free + private) using cached function has_access()
-            authorized_instances = get_all_available_instances(user)  # self._get_authorized_instances(user, api)
+            authorized_instances = self.get_all_available_instances(user)
             if not authorized_instances:
                 # user doesn't have access to any of the instances
                 context = 'User has no access to any instance'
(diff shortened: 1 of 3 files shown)


pbench (approved):
:crossed_fingers:

source/jormungandr/jormungandr/instance_manager.py
```diff
@@ -297,7 +298,7 @@ def get_instances(self, name=None, lon=None, lat=None, object_id=None, api='ALL'
         else:
             # Requests without any coverage
             # fetch all the authorized instances (free + private) using cached function has_access()
-            authorized_instances = self._get_authorized_instances(user, api)
+            authorized_instances = get_all_available_instances(user)  # self._get_authorized_instances(user, api)
```
woshilapin: I believe you can remove the comment. `git` history is enough to remember.

woshilapin (approved):

source/jormungandr/jormungandr/instance_manager.py
azime: :+1:

source/jormungandr/jormungandr/instance_manager.py
```diff
@@ -341,3 +342,24 @@ def regions(self, region=None, lon=None, lat=None, request_id=None):
             resp_dict['region_id'] = key_region
             response['regions'].append(resp_dict)
         return response
+
+    def get_all_available_instances(self, user):
```
kadhikari, line 348: This function should be cached.
As discussed with @pbench if we can remove the condition as described in https://github.com/hove-io/navitia/blob/17d38284168f39d3f888fb77d19f2a9ec9083281/source/jormungandr/jormungandr/authentication.py#L255
we can re-user the function authentification.get_all_available_instances or add a simplified and cached function only to be used for this purpose.

kadhikari (changes requested):

source/jormungandr/jormungandr/instance_manager.py
pbench: Thanks ! I added a cache

kadhikari (approved):

source/jormungandr/jormungandr/instance_manager.py
```diff
@@ -341,3 +342,26 @@ def regions(self, region=None, lon=None, lat=None, request_id=None):
...
+            logging.getLogger(__name__).warning('get all available instances no user')
+            # for not-public navitia a user is mandatory
+            # To manage database error of the following type we should fetch one more time from database
+            # Can connect to database but at least one table/attribute is not accessible due to transaction problem
+            if can_read_user():
+                abort_request(user=user)
+            else:
+                return result
+
+        bdd_instances = user.get_all_available_instances()
```
xlqian, line 363:
```suggestion
        instances_from_db = user.get_all_available_instances()
```

xlqian (approved):

PR #3929, opened 2023-02-02
Remove get_without_access func

azime:
This correction prevents access to the database for an unknown token:
- Before: One access to database to verify + another access to the database to get user of type no_access during de cache period.
- After: One access to database to verify but no more access during de cache period.

**Note: This correction is more useful when someone attacks navitia with a script using a new token each time.** 

https://navitia.atlassian.net/browse/NAV-1779

### DIFF
--- a/source/jormungandr/jormungandr/authentication.py
+++ b/source/jormungandr/jormungandr/authentication.py
@@ -197,7 +197,7 @@ def uncached_get_user(token):
     try:
         user = User.get_from_token(token, datetime.datetime.now())
 
-        # if user doesn't exist for a token, get default token with user_type = no_access
+        # if user doesn't exist for a token, create default user with user.login = unknown_user
         if not user:
             user = get_unkown_user()
             logging.getLogger(__name__).warning('Invalid token : {}'.format(token[0:10]))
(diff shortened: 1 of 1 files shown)


kadhikari (approved):

source/jormungandr/jormungandr/authentication.py
```diff
@@ -199,7 +199,7 @@ def uncached_get_user(token):
 
         # if user doesn't exist for a token, get default token with user_type = no_access
```
woshilapin: Maybe this comment is not appropriate anymore.

source/jormungandr/jormungandr/authentication.py
```diff
@@ -290,6 +289,12 @@ def get_user(token, abort_if_no_token=True):
         return g.user
 
 
+def get_unkown_user():
+    user = User(login="unknown_user")
+    user.id = 0
```
woshilapin, line 294: Isn't it possible to do the following?

```suggestion
    user = User(id=0, login="unknown_user")
```

woshilapin (approved):

source/jormungandr/jormungandr/authentication.py
azime: The user [constructor](https://github.com/hove-io/navitia/blob/dev/source/navitiacommon/navitiacommon/models/__init__.py#L147) without id

source/jormungandr/jormungandr/authentication.py
azime: Done

kadhikari (approved):

woshilapin (approved):

PR #3941, opened 2023-02-20
Avoid calling the database in case of exception for a period configured

kadhikari:
If There is an exception while sending a query to the database, we should update last_update time to NOW so that no query is send to the data base for another update_interval as configured.

This correction avoids access to the data for each service as street_network, external_service as well as bss_provide for a period of update_interval as configured.

There are some minor modifications to reduce possible error. 

Relates to: https://navitia.atlassian.net/browse/NAV-1819

### DIFF
--- a/source/jormungandr/jormungandr/authentication.py
+++ b/source/jormungandr/jormungandr/authentication.py
@@ -120,7 +120,7 @@ def can_read_user():
     try:
         User.query.options(noload('*')).first()
     except Exception as e:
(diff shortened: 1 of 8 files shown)


pbench (approved):

pbougue (approved):
I know it's a sensitive word recently, but... Should we put a circuit breaker?

xlqian (approved):

PR #3943, opened 2023-02-22
[Jormungandr] Fix failure of caching instances list

xlqian:
### **Branch based on v15.26.7**

Flask-caching was not able to pickle a list a ORM obj T_T, https://github.com/hove-io/navitia/pull/3926/files#diff-cc347701ac573ba458d0543b788c67d7e83be259f7ae5fe0c095aed939880090R366

so instead of returning a list of ORM, we return a list of name...

### DIFF
--- a/source/jormungandr/jormungandr/instance_manager.py
+++ b/source/jormungandr/jormungandr/instance_manager.py
@@ -298,7 +298,7 @@ class InstanceManager(object):
         else:
             # Requests without any coverage
             # fetch all the authorized instances (free + private) using cached function has_access()
-            authorized_instances_name = self.get_all_available_instances(user)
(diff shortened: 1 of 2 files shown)


xlqian:
A new hotfix  **v15.26.8** will be made from this branch

azime (approved):

pbench (approved):

pbench:
This is strange that the ORM objects cannot be pickled.
We already do this in https://github.com/hove-io/navitia/blob/b08863f334b7f1e04f6dc698a9139745ba019a01/source/jormungandr/jormungandr/authentication.py#L233 :thinking:

xlqian:
@pbench 

hey, you are right... I confirm that this function has the same problem indeed

kadhikari (approved):

PR #3950, opened 2023-03-08
Jormun: Optim log readyness and others

kadhikari:
* Log on /readyness only with log_level=DEBUG
* Duplicate log deleted

For details: https://navitia.atlassian.net/browse/NAV-1858

### DIFF
--- a/source/jormungandr/jormungandr/api.py
+++ b/source/jormungandr/jormungandr/api.py
@@ -74,7 +74,11 @@ def access_log(response, *args, **kwargs):
         'status': response.status_code,
         'external_request_id': request.headers.get('x-request-id'),
     }
-    logger.info(u'"%(method)s %(path)s?%(query_string)s" %(status)s', d, extra=d)
+    if 'readyness' not in request.path:
+        logger.info(u'"%(method)s %(path)s?%(query_string)s" %(status)s', d, extra=d)
+    else:
+        logging.getLogger(__name__).debug(u'"%(method)s %(path)s?%(query_string)s" %(status)s', d, extra=d)
+
     return response
 
 
--- a/source/jormungandr/jormungandr/equipments/equipment_provider_manager.py
+++ b/source/jormungandr/jormungandr/equipments/equipment_provider_manager.py
@@ -75,9 +75,6 @@ class EquipmentProviderManager(object):
         :return: instance of provider
         """
         try:
-            if '.' not in cls:
-                self.logger.warning('impossible to build, wrongly formated class: {}'.format(cls))
-
             module_path, name = cls.rsplit('.', 1)
             module = import_module(module_path)
             attr = getattr(module, name)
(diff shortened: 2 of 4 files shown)


source/jormungandr/jormungandr/api.py
```diff
@@ -74,7 +74,11 @@ def access_log(response, *args, **kwargs):
         'status': response.status_code,
         'external_request_id': request.headers.get('x-request-id'),
     }
-    logger.info(u'"%(method)s %(path)s?%(query_string)s" %(status)s', d, extra=d)
+    if 'readyness' not in request.path:
+        logger.info(u'"%(method)s %(path)s?%(query_string)s" %(status)s', d, extra=d)
+    else:
+        logging.getLogger(__name__).debug(u'"%(method)s %(path)s?%(query_string)s" %(status)s', d, extra=d)
```
pbench, line 80: is there a reason for not using `logger` here as above ?
```suggestion
        logger.debug(u'"%(method)s %(path)s?%(query_string)s" %(status)s', d, extra=d)
```

source/jormungandr/jormungandr/equipments/equipment_provider_manager.py
```diff
@@ -75,9 +75,6 @@ def _init_class(self, cls, arguments):
         :return: instance of provider
         """
         try:
-            if '.' not in cls:
-                self.logger.warning('impossible to build, wrongly formated class: {}'.format(cls))
```
pbougue, line 79: I don't get what's affected by that change, even though it looks harmless.

pbougue (commented):
Waiting for a reply on comments (from @pbench and me) but it looks OK

source/jormungandr/jormungandr/equipments/equipment_provider_manager.py
kadhikari: It is most probable that there will be an exception and hence will write another log four lines after.
I simply deleted duplicate log writing.

source/jormungandr/jormungandr/api.py
kadhikari: Will test this suggestion.

pbench (approved):

source/jormungandr/jormungandr/api.py
kadhikari: This doesn't work as the logger is initialized with level = INFO.

source/jormungandr/jormungandr/equipments/equipment_provider_manager.py
pbougue: I think the change is that if we use a "root" class it won't raise anymore, but before it did... I guess it's not important.

source/jormungandr/jormungandr/api.py
pbougue: But if you change the logger's conf of `jormungandr.access` to `DEBUG` then it should work, no?
https://github.com/hove-io/navitia/blob/dev/source/jormungandr/jormungandr/default_settings.py#L96
And it's what we want I think (only log this when `jormungandr.access` conf is in DEBUG).

pbougue: After discussion: it would be cleaner to separate and route/activate at will.
But it's a bit more work and the use is not clear, so not spending the time for that and see in the future if it's needed.

pbougue (approved):

azime (approved):

PR #3955, opened 2023-03-24
Kraken: Add metrics around RT handling

pbougue:
:mag: please review by commit **with message**

Added metrics:
* retrieval duration of RT messages from Rabbit
* number of RT messages retrieved from Rabbit
* number of RT entities applied
* RT messages' ages (min/average/max)

Also a small fix.

:heavy_check_mark: tested locally that it works on NewRelic metrics.
:x: Sonar fails only on new code coverage, which is OK on metrics.

JIRA: https://navitia.atlassian.net/browse/NAV-1892

### DIFF
--- a/source/kraken/maintenance_worker.cpp
+++ b/source/kraken/maintenance_worker.cpp
@@ -366,11 +366,11 @@ void MaintenanceWorker::handle_rt_in_batch(const std::vector<AmqpClient::Envelop
     bool autocomplete_rebuilding_activated = false;
     auto rt_action = RTAction::chaos;
 
-    size_t applied_entity_number = 0u;
+    size_t applied_entity_count = 0u;
     boost::optional<pt::ptime> oldest_message_time;
     boost::optional<pt::ptime> youngest_message_time;
     uint64_t sum_message_age_until_begin_microseconds = 0u;
-    size_t dated_message_number = 0u;
+    size_t dated_message_count = 0u;
     for (auto& envelope : envelopes) {
         const auto routing_key = envelope->RoutingKey();
         LOG4CPLUS_DEBUG(logger, "realtime info received from " << routing_key);
@@ -388,7 +388,7 @@ void MaintenanceWorker::handle_rt_in_batch(const std::vector<AmqpClient::Envelop
             if (!youngest_message_time || youngest_message_time < message_time) {
                 youngest_message_time = message_time;
             }
-            ++dated_message_number;
+            ++dated_message_count;
             sum_message_age_until_begin_microseconds += pt::time_duration(begin - message_time).total_microseconds();
         }
         LOG4CPLUS_TRACE(logger, "received entity: " << feed_message.DebugString());
@@ -400,7 +400,7 @@ void MaintenanceWorker::handle_rt_in_batch(const std::vector<AmqpClient::Envelop
                 this->metrics.observe_data_cloning(duration.total_seconds());
                 LOG4CPLUS_INFO(logger, "data copied (cloned) in " << duration);
             }
-            ++applied_entity_number;
+            ++applied_entity_count;
             if (entity.is_deleted()) {
                 LOG4CPLUS_DEBUG(logger, "deletion of disruption " << entity.id());
                 rt_action = RTAction::deletion;
@@ -418,14 +418,14 @@ void MaintenanceWorker::handle_rt_in_batch(const std::vector<AmqpClient::Envelop
                 autocomplete_rebuilding_activated = autocomplete_rebuilding_needed(entity);
             } else {
                 LOG4CPLUS_WARN(logger, "unsupported gtfs rt feed");
-                --applied_entity_number;
+                --applied_entity_count;
             }
         }
     }
-    if (envelopes.size() > 0) {
-        // messages may contain multiple entities, but some may be skipped
-        this->metrics.observe_applied_rt_entity_number(applied_entity_number);
-        LOG4CPLUS_DEBUG(logger, "Number of RT entity really applied in this message batch: " << applied_entity_number);
+    if (!envelopes.empty()) {
+        // messages may contain multiple entities, and some may be skipped
+        this->metrics.observe_applied_rt_entity_count(applied_entity_count);
+        LOG4CPLUS_DEBUG(logger, "Number of RT entity really applied in this message batch: " << applied_entity_count);
     }
     if (data) {
         LOG4CPLUS_INFO(logger, "rebuilding relations");
@@ -459,15 +459,15 @@ void MaintenanceWorker::handle_rt_in_batch(const std::vector<AmqpClient::Envelop
             LOG4CPLUS_INFO(logger, "Data updated with realtime from kirin, "
                                        << envelopes.size() << " disruption(s) applied in " << duration);
         }
-        if (dated_message_number > 0) {
+        if (dated_message_count > 0) {
             auto min_age = end - *youngest_message_time;
             auto max_age = end - *oldest_message_time;
             auto sum_message_age_until_end_microseconds =
-                sum_message_age_until_begin_microseconds + (dated_message_number * (end - begin).total_microseconds());
-            auto average_age_microseconds = sum_message_age_until_end_microseconds / dated_message_number;
-            this->metrics.observe_rt_message_age_min(min_age.total_milliseconds() / 1000.0);
-            this->metrics.observe_rt_message_age_max(max_age.total_milliseconds() / 1000.0);
-            this->metrics.observe_rt_message_age_average(average_age_microseconds / 1000000.0);
+                sum_message_age_until_begin_microseconds + (dated_message_count * (end - begin).total_microseconds());
+            auto average_age_microseconds = sum_message_age_until_end_microseconds / dated_message_count;
+            this->metrics.observe_rt_message_age_min(double(min_age.total_milliseconds()) / 1000.0);
+            this->metrics.observe_rt_message_age_max(double(max_age.total_milliseconds()) / 1000.0);
+            this->metrics.observe_rt_message_age_average(double(average_age_microseconds) / 1000000.0);
             LOG4CPLUS_DEBUG(logger, "Known ages of RT message(s) in batch: min="
                                         << min_age << ", average=" << pt::microseconds(average_age_microseconds)
                                         << ", max=" << max_age);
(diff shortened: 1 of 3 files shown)


source/kraken/metrics.cpp
```diff
@@ -151,6 +151,48 @@ Metrics::Metrics(const boost::optional<std::string>& endpoint, const std::string
...
+                                              .Add({}, create_exponential_buckets(0.5, 2, 10));
+
+    this->rt_message_age_average_histogram = &prometheus::BuildHistogram()
+                                                  .Name("kraken_rt_message_age_average_seconds")
+                                                  .Help("Average age of RT message from a batch")
+                                                  .Labels({{"coverage", coverage}})
+                                                  .Register(*registry)
+                                                  .Add({}, create_exponential_buckets(0.5, 2, 10));
+
+    this->rt_message_age_max_histogram = &prometheus::BuildHistogram()
```
pbench, line 190: why not use a single `message_age_histogram` ? I seems to me you can recover the max and average from that (not use about the min)

pbench: and you could also recover the number of messages from this single histogram

source/kraken/maintenance_worker.cpp
```diff
@@ -508,7 +547,13 @@ void MaintenanceWorker::listen_rabbitmq() {
         size_t max_batch_nb = conf.broker_max_batch_nb();
 
         try {
+            auto begin_rt_retrieval = pt::microsec_clock::universal_time();
             auto rt_envelopes = consume_in_batch(rt_tag, max_batch_nb, timeout_ms, no_ack);
+            auto duration_rt_retieval = pt::microsec_clock::universal_time() - begin_rt_retrieval;
```
woshilapin:
```suggestion
            auto duration_rt_retrieval = pt::microsec_clock::universal_time() - begin_rt_retrieval;
```

source/kraken/metrics.cpp
```diff
@@ -151,6 +151,48 @@ Metrics::Metrics(const boost::optional<std::string>& endpoint, const std::string
...
+                                                        .Add({}, create_exponential_buckets(0.5, 2, 10));
+
+    this->retrieved_rt_message_number_histogram = &prometheus::BuildHistogram()
+                                                       .Name("kraken_retrieve_rt_message_number")
+                                                       .Help("number of RT messages retrieved from RabbitMQ")
+                                                       .Labels({{"coverage", coverage}})
+                                                       .Register(*registry)
+                                                       .Add({}, create_exponential_buckets(0.5, 2, 10));
+
+    this->applied_rt_entity_number_histogram = &prometheus::BuildHistogram()
```
pbench: I think a [Counter](https://prometheus.io/docs/concepts/metric_types/#counter) is more appropriate than an histogram here

source/kraken/maintenance_worker.cpp
```diff
@@ -366,6 +366,7 @@ void MaintenanceWorker::handle_rt_in_batch(const std::vector<AmqpClient::Envelop
     bool autocomplete_rebuilding_activated = false;
     auto rt_action = RTAction::chaos;
 
+    size_t applied_entity_number = 0u;
```
woshilapin: I'm often disturbed by names ending with `number`. Does it mean the identifier (the number) of an applied entity, or is it the counting of applied entities? In this case, I believe it's the count so I would suggest either `applied_entity_count` (or `number_of_applied_entities`?).

woshilapin (approved):

source/kraken/metrics.cpp
pbougue: I admit that I didn't dig on this and cut short to have it available on time: the min and max values are not precise from what I saw (maybe it's related to the config of histogram and its buckets).
One other advantage is to avoid tracking that many values, but I'm not sure as it's probably regrouped.
I'll try to dig a bit before making a choice.

EDIT: here is what's exposed by `/metrics` for `average`
```
# HELP kraken_rt_message_age_average_seconds Average age of RT message from a batch
# TYPE kraken_rt_message_age_average_seconds histogram
kraken_rt_message_age_average_seconds_count{coverage="default"} 4
kraken_rt_message_age_average_seconds_sum{coverage="default"} 20.293984
kraken_rt_message_age_average_seconds_bucket{coverage="default",le="0.500000"} 0
kraken_rt_message_age_average_seconds_bucket{coverage="default",le="1.000000"} 0
kraken_rt_message_age_average_seconds_bucket{coverage="default",le="2.000000"} 0
kraken_rt_message_age_average_seconds_bucket{coverage="default",le="4.000000"} 0
kraken_rt_message_age_average_seconds_bucket{coverage="default",le="8.000000"} 4
kraken_rt_message_age_average_seconds_bucket{coverage="default",le="16.000000"} 4
kraken_rt_message_age_average_seconds_bucket{coverage="default",le="32.000000"} 4
kraken_rt_message_age_average_seconds_bucket{coverage="default",le="64.000000"} 4
kraken_rt_message_age_average_seconds_bucket{coverage="default",le="128.000000"} 4
kraken_rt_message_age_average_seconds_bucket{coverage="default",le="256.000000"} 4
kraken_rt_message_age_average_seconds_bucket{coverage="default",le="+Inf"} 4
```
So min/max are dependant on the number of buckets and their size, we may use something like [quantiles](https://prometheus.io/docs/practices/histograms/#quantiles) but I think it will be too short for this PR.

source/kraken/metrics.cpp
pbougue: Let's discuss that, but I'm not sure that it's appropriate here: we want to track the number of applied entities in a batch, so it's not monotonic IMO.

pbench: :+1:

source/kraken/maintenance_worker.cpp
pbougue: Done in https://github.com/hove-io/navitia/pull/3955/commits/e7122af1bdb0aaec2c912e7f5dde1bb3f60b9c68

source/kraken/maintenance_worker.cpp
pbougue: Switched all to `count` in https://github.com/hove-io/navitia/pull/3955/commits/e7122af1bdb0aaec2c912e7f5dde1bb3f60b9c68

source/kraken/metrics.cpp
pbougue: As for the number of messages, the `count` from this histogram measures the number of messages retrieved that have a date, not all of them (not sure the ones from chaos have some), and not all the entities applied.
So it can be interesting to have it, but it's not what I aimed for when I listed the needs.

pbench (approved):

source/kraken/maintenance_worker.cpp
```diff
@@ -366,14 +366,30 @@ void MaintenanceWorker::handle_rt_in_batch(const std::vector<AmqpClient::Envelop
     bool autocomplete_rebuilding_activated = false;
     auto rt_action = RTAction::chaos;
 
+    size_t applied_entity_count = 0u;
+    boost::optional<pt::ptime> oldest_message_time;
```
xlqian:
```suggestion
        pt::ptime oldest_message_time{max_date_time};
        pt::ptime oldest_message_time{min_date_time};
```

source/kraken/maintenance_worker.cpp
```diff
@@ -366,14 +366,30 @@ void MaintenanceWorker::handle_rt_in_batch(const std::vector<AmqpClient::Envelop
...
+            continue;
+        }
+        if (feed_message.header().has_timestamp()) {
+            auto message_time = navitia::from_posix_timestamp(feed_message.header().timestamp());
+            if (!oldest_message_time || oldest_message_time > message_time) {
+                oldest_message_time = message_time;
+            }
+            if (!youngest_message_time || youngest_message_time < message_time) {
+                youngest_message_time = message_time;
+            }
```
xlqian:
```suggestion
            oldest_message_time = std::min(oldest_message_time, message_time);
            youngest_message_time = std::max(youngest_message_time, message_time);
```

woshilapin (approved):

source/kraken/metrics.cpp
xlqian: my question is why do you need to know min/max precisely?

pbougue: @xlqian : This metric gives the latency between the reception of RT info from provider to the moment it is available for the travelers (kirin's time is monitored too, to complete it).
So it is quite important for some clients to track it, and `max` or `average` is important and tracked to some clients.
`min` is less important, but it's interesting because it measures minimal latency unrelated to RT-messages themselves.

This is actually the end-goal of improving the RT handling time, the metric we are going to track.

To work on it, especially on max, steps of 128s (~2 min), 256s (~4 min) or "above" is too large (as we want to be able to show a precise max to client - and track it for our work).

pbench: you can specify more precise buckets if it fits your needs

source/kraken/maintenance_worker.cpp
pbougue: Done in https://github.com/hove-io/navitia/pull/3955/commits/64b8191eb3883bad091529a6451c56901e860c15

source/kraken/maintenance_worker.cpp
pbougue: Done in https://github.com/hove-io/navitia/pull/3955/commits/64b8191eb3883bad091529a6451c56901e860c15

source/kraken/metrics.cpp
pbench: Something like : 
```
 this->rt_message_average_histogram = &prometheus::BuildHistogram()
                                                  .Name("kraken_rt_message_age_average_seconds")
                                                  .Help(" Age of RT messages")
                                                  .Labels({{"coverage", coverage}})
                                                  .Register(*registry)
                                                  .Add({}, 
prometheus::Histogram::BucketBoundaries{0.1, 0.5, 1,   2,   5,   10, 30, 60, 90, 120, 180});
```

xlqian (approved):

source/kraken/metrics.cpp
pbougue: I tried adding a unique/more precise histogram (see https://github.com/hove-io/navitia/commit/fd891b3b481f2ad4de3c9571c4866328dca523f2 on branch `metrics_only_one_histogram_rt_message_age`) and the result in NewRelic on max is not good on one test.
As discussed, we will probably move on as-is and decide to dig after the release.

PR opened and closed immediately for memory: https://github.com/hove-io/navitia/pull/3971

PR #3957, opened 2023-03-30
CI: Fix generate tyr-worker

azime:
See: [Job](https://github.com/hove-io/navitia/actions/runs/4540173254/jobs/8033078777) GithubAction

image

### DIFF
--- a/docker/debian8/Dockerfile-tyr-worker
+++ b/docker/debian8/Dockerfile-tyr-worker
@@ -8,27 +8,27 @@ COPY mimirsbrunn7_*.deb ./
 COPY cosmogony2cities_*.deb ./
 
 # install navitia-common package
-RUN dpkg -i ./navitia-common_*.deb || exit 0
+RUN dpkg -i ./navitia-common_*.deb
 RUN apt-get install -f -y --force-yes
 
 # install navitia-tyr package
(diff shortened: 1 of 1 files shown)


woshilapin (approved):

pbench (approved):

docker/debian8/Dockerfile-tyr-worker
```diff
@@ -8,27 +8,27 @@ COPY mimirsbrunn7_*.deb ./
...
 RUN apt-get install -f -y --force-yes
 
 # install navitia-cities package
-RUN dpkg -i ./navitia-cities_*.deb || exit 0
+RUN dpkg -i ./navitia-cities_*.deb
 RUN apt-get install -f -y --force-yes
 
 # install navitia-mimirsbrunn package
-RUN dpkg -i ./mimirsbrunn7_jessie-*.deb || exit 0
+RUN dpkg -i ./mimirsbrunn7_jessie-*.deb
```
pbougue, line 27: I don't get why it's OK to build an image without mimir (or any other).

pbougue: Ah, after discussion, missed that `exit 0` is "OK", so it's indeed better as it will fail :+1:

pbougue (approved):

PR #3958, opened 2023-03-30
CI: rollback tyr-worker image generation

azime:
See [Job](https://github.com/hove-io/navitia/actions/runs/4563304837/jobs/8051627347) GithubAction


pbougue (approved):
It's a partial rolback, so better to have that than nothing.

But if the problem is requirements, we may switch the order for the to gain even more, and maybe have all "secured".

woshilapin (approved):

PR #3965, opened 2023-04-03
Jormungandr: Cache Co2 emission value instead of pb object

kadhikari:
Two problems occurred while caching pb object Co2Emission
-  Caching for key = uri, request_id: as request_id is unique for a request global to navitia, caching on uri and request_id is useless.
- Caching pb object raises error (see ticket for more details)

A little adjustment is made to pass request_id without including it's value as key  while caching.

For more details: https://navitia.atlassian.net/browse/NAV-1667

### DIFF
--- a/source/jormungandr/jormungandr/georef.py
+++ b/source/jormungandr/jormungandr/georef.py
@@ -73,10 +73,6 @@ class Kraken(object):
             return None
         return response.car_co2_emission
 
-    @memory_cache.memoize(
-        current_app.config[str('MEMORY_CACHE_CONFIGURATION')].get(str('TIMEOUT_AUTHENTICATION'), 30)
-    )
-    @cache.memoize(current_app.config[str('CACHE_CONFIGURATION')].get(str('TIMEOUT_AUTHENTICATION'), 300))
     def get_physical_mode(self, uri, request_id):
         req = request_pb2.Request()
         req.requested_api = type_pb2.PTREFERENTIAL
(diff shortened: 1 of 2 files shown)


azime (approved):

pbench (approved):

source/jormungandr/jormungandr/georef.py
```diff
@@ -90,15 +86,21 @@ def get_physical_mode(self, uri, request_id):
...
-        if car_mode is None or not car_mode.HasField('co2_emission_rate'):
-            logger.warning("Cannot compute car co2 emission with the distance {}".format(distance))
-            return response_pb2.Co2Emission(value=0, unit='gEC')
-
-        return response_pb2.Co2Emission(value=car_mode.co2_emission_rate.value * distance / 1000.0, unit='gEC')
+        @memory_cache.memoize(
+            current_app.config[str('MEMORY_CACHE_CONFIGURATION')].get(str('TIMEOUT_AUTHENTICATION'), 30)
+        )
+        @cache.memoize(current_app.config[str('CACHE_CONFIGURATION')].get(str('TIMEOUT_AUTHENTICATION'), 300))
+        def inner(mode_str):
```
woshilapin, line 96: Instead of using this convoluted solution of an inner function, to ignore the `request_id` from the cache key, `@cache.memoize` has an option to ignore some function parameters with `args_to_ignore` (see [documentation](https://flask-caching.readthedocs.io/en/latest/api.html#flask_caching.Cache.memoize)).

So the PR could only be

```
        @memory_cache.memoize(
            current_app.config[str('MEMORY_CACHE_CONFIGURATION')].get(str('TIMEOUT_AUTHENTICATION'), 30),
            args_to_ignore=["request_id"],
        )
        @cache.memoize(
            current_app.config[str('CACHE_CONFIGURATION')].get(str('TIMEOUT_AUTHENTICATION'), 300), 
            args_to_ignore=["request_id"],
        )
```

woshilapin (commented):
I believe this can be fixed differently in a much simpler way.

source/jormungandr/jormungandr/georef.py
```diff
@@ -90,15 +86,21 @@ def get_physical_mode(self, uri, request_id):
             return res.physical_modes[0]
         return None
 
-    def get_car_co2_emission(self, distance, request_id):
```
SGrenet, line 93: distance deletion is strange, is it useless?

source/jormungandr/jormungandr/scenarios/new_default.py
```diff
@@ -281,8 +281,10 @@ def fill_missing_co2_emission(pb_resp, instance, request_id):
         car_sections = (s for s in j.sections if is_car_section_without_co2(s))
 
         for s in car_sections:
-            co2_emission = instance.georef.get_car_co2_emission(s.street_network.length, request_id)
-            s.co2_emission.CopyFrom(co2_emission)
+            co2_emission = instance.georef.get_car_co2_emission(request_id)
+            pb_co2_emission = s.co2_emission
+            pb_co2_emission.value = (co2_emission * s.street_network.length) / 1000.0 if co2_emission > 0 else 0
```
woshilapin, line 286: I'm not sure that the `if co2_emission > 0 else 0` is useful. In case `co2_emission == 0`, the calculation will already be 0. Can `co2_emission` be negative?

kadhikari: This will avoid negative value

source/jormungandr/jormungandr/georef.py
kadhikari: It's useless as it's done in new_default

PR #3968, opened 2023-04-06
Kraken: Count number of unprefetched messages in queue + increase prefetch

pbougue:
Increasing prefetched messages (default to `100`) should improve retrieval speed.
Consume messages until this unprefetched+prefetched count (or max - increased to `10000` by default) is reached.
This allows for larger timeouts when there are unprefetched messages, as Kraken will stop at that number, and avoid continuing as long as messages arrive (which is constantly the case for some contributors). If there are only prefetched messages, rushing is possible.

Also:
* Add short "prefetch" timeouts config - `100ms and 10s` (in additional to "classic" ones - increased to `1s and 2min`)
* Common queue declaration (to avoid bad-redeclaration)
* Check (only for log currently) that consumer is unique
* minor cleanups of useless variables

Notes:
* `DeclareQueue()` uses `DeclareQueueWithCounts()` under the cover in SimpleAmpqClient, which allows for the common queue declaration refactor.
* `BasicConsume()` declares a consumer that prefetches 1 message by default. `DeclareQueueWithCounts` doesn't include prefetched messages. Hence tricky `max_message_to_retrieve`.

JIRA: https://navitia.atlassian.net/browse/NAV-1879

### DIFF
--- a/source/kraken/maintenance_worker.cpp
+++ b/source/kraken/maintenance_worker.cpp
@@ -175,6 +175,10 @@ void MaintenanceWorker::load_data() {
     LOG4CPLUS_INFO(logger, "Loading database duration: " << duration);
 }
 
+/**
+ * This function can be called multiple times without creating new queues. The goal is to take advantage
+ * of DeclareQueueWithCounts to get the number of messages remaining (just ready, not unacked) in the queue currently.
+ */
 std::pair<uint32_t, uint32_t> MaintenanceWorker::declare_queue_with_counts(const std::string& queue_name) {
     bool passive = false;
     bool durable = true;
@@ -494,6 +498,7 @@ std::vector<AmqpClient::Envelope::ptr_t> MaintenanceWorker::consume_in_batch(con
     std::vector<AmqpClient::Envelope::ptr_t> envelopes;
 
     // count ready messages (prefetched messages are excluded from this count)
+    // declare_queue_with_counts() is called but no queue is actually created
     auto count_messages_consumers = declare_queue_with_counts(queue_name);
     LOG4CPLUS_DEBUG(logger, "Nb of message(s) ready in queue " << queue_name << ": " << count_messages_consumers.first);
     LOG4CPLUS_DEBUG(logger, "Nb of consumer(s) on queue" << queue_name << ": " << count_messages_consumers.second);
--- a/source/kraken/maintenance_worker.cpp
+++ b/source/kraken/maintenance_worker.cpp
@@ -113,7 +113,6 @@ void MaintenanceWorker::open_channel_to_rabbitmq() {
     if (channel_opened) {
         return;
     }
-    std::string instance_name = conf.instance_name();
     // connection through URI, if URI is provided, it will be used in the first place as other other connection options
     // are neglected
     boost::optional<std::string> broker_uri = conf.broker_uri();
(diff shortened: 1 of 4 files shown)


source/kraken/maintenance_worker.cpp
```diff
@@ -485,29 +485,45 @@ void MaintenanceWorker::handle_rt_in_batch(const std::vector<AmqpClient::Envelop
...
-         * BasicConsumeMessage() timeout.
-         * */
-        bool queue_is_empty = !channel->BasicConsumeMessage(consume_tag, envelope, timeout_ms);
-        if (queue_is_empty) {
+        bool queue_is_empty_or_unreachable = !channel->BasicConsumeMessage(consume_tag, envelope, single_timeout);
+        if (queue_is_empty_or_unreachable) {
+            LOG4CPLUS_ERROR(logger,
+                            "Could not retrieve all the messages counted for RabbitMQ's RT queue " << queue_name);
+            // going on with what was retrieved so far
             break;
```
pbougue, line 532: An idea by @pbench (correct me if needed) to simplify code and params would be:
1. put a `continue` here instead of a `break`
2. only one total_timeout (i.e. 30s), single timeout may be hardcoded
3. remove count and just go with what was retrieved during those 30s

This would simplify code, the inconvenient is to have a strict latency even if there is only one message (and tests would need a little parameter adaptation for that reason).

woshilapin (approved):

source/kraken/maintenance_worker.cpp
```diff
@@ -176,6 +175,23 @@ void MaintenanceWorker::load_data() {
     LOG4CPLUS_INFO(logger, "Loading database duration: " << duration);
 }
 
+std::pair<uint32_t, uint32_t> MaintenanceWorker::declare_queue_with_counts(const std::string& queue_name) {
```
xlqian, line 182:
```suggestion
// This function could be called multiple times without creating new queues. The goal is to take advantage 
// of DeclareQueueWithCounts to get the number of messages remaining in the queue currently  
std::pair<uint32_t, uint32_t> MaintenanceWorker::declare_queue_with_counts(const std::string& queue_name) {
```

source/kraken/maintenance_worker.cpp
```diff
@@ -485,29 +485,45 @@ void MaintenanceWorker::handle_rt_in_batch(const std::vector<AmqpClient::Envelop
...
                                                                              bool no_ack) {
     assert(consume_tag != "");
     assert(max_nb);
-
     std::vector<AmqpClient::Envelope::ptr_t> envelopes;
-    envelopes.reserve(max_nb);
-    size_t consumed_nb = 0;
+
+    // count ready messages (prefetched messages are excluded from this count)
+    auto count_messages_consumers = declare_queue_with_counts(queue_name);
```
xlqian, line 502:
```suggestion
    // declare_queue_with_counts is called but no queues are actually created
    auto count_messages_consumers = declare_queue_with_counts(queue_name);
```

xlqian (approved):
:+1:  GG

source/kraken/maintenance_worker.cpp
pbougue: Done in https://github.com/hove-io/navitia/pull/3968/commits/868fc7b3fe6b5012a39fd6e4ae3da3d0e85a022f

source/kraken/maintenance_worker.cpp
pbougue: Done in https://github.com/hove-io/navitia/pull/3968/commits/868fc7b3fe6b5012a39fd6e4ae3da3d0e85a022f

PR #3969, opened 2023-04-06
[Jormungandr] Fix LEZ when fallbacks are car

xlqian:
In this PR, two cases are fixed:

- at least one of the fallbacks traverses LEZ, we set `on_path` to true for the whole journey :
image

- none of the fallbacks traverses LEZ, we set `on_path` to false for the whole journey:
image

### DIFF
--- a/source/jormungandr/jormungandr/scenarios/helper_classes/complete_pt_journey.py
+++ b/source/jormungandr/jormungandr/scenarios/helper_classes/complete_pt_journey.py
@@ -113,6 +113,23 @@ def wait_and_build_crowflies(
     return res
 
 
+def tag_LEZ(journey):
+    tmp_list = [
+        not section.low_emission_zone.on_path
+        for section in journey.sections
+        if section.HasField('low_emission_zone')
+    ]
+    # all([]) == True it's not what we want :(
+    if tmp_list and all(tmp_list):
+        journey.low_emission_zone.on_path = False
+    elif any(
+        section.low_emission_zone.on_path
+        for section in journey.sections
+        if section.HasField('low_emission_zone')
+    ):
+        journey.low_emission_zone.on_path = True
+
+
 def get_journeys_to_complete(responses, context, is_debug):
     """
     Prepare a list of journeys from the response that will be use to compute street-network as a fallback.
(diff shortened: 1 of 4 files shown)


source/jormungandr/jormungandr/scenarios/helper_classes/complete_pt_journey.py
```diff
@@ -190,6 +207,8 @@ def wait_and_complete_pt_journey(
                 dest_fallback_durations_pool=dest_fallback_durations_pool,
                 request=request,
             )
+            if {pt_element.dep_mode, pt_element.arr_mode} & {'car', 'car_no_park'}:
```
pbench, line 210: `&` means 'subset of ' ?

xlqian: it's intersection

source/jormungandr/jormungandr/scenarios/helper_classes/complete_pt_journey.py
```diff
@@ -113,6 +113,23 @@ def wait_and_build_crowflies(
...
+    if tmp_list and all(tmp_list):
+        journey.low_emission_zone.on_path = False
+    elif any(
+        section.low_emission_zone.on_path
+        for section in journey.sections
+        if section.HasField('low_emission_zone')
+    ):
+        journey.low_emission_zone.on_path = True
+
+
```
pbench, line 132:
```suggestion
     for section in journey.sections:
         if section.HasField('low_emission_zone'):
             if section.low_emission_zone.on_path:
                 journey.low_emission_zone.on_path = True
                 break
```

xlqian: it won't work if none of the fallbacks has LEZ...

xlqian: and for the sake of compatibility, we don't display `low_emission_zone.on_path` if the information of LEZ is unavailable...

pbench (approved):

azime (approved):

PR #3970, opened 2023-04-06
Refactoring PlaceUri in distributed

azime:


source/jormungandr/tests/isochrone_tests.py
```diff
@@ -44,7 +44,7 @@ def test_from_isochrone_coord(self):
         # not to use the jormungandr database
         query = "v1/coverage/basic_routing_test/journeys?from={}&datetime={}"
         query = query.format(s_coord, "20120614T080000")
-        self.query(query)
+        response = self.query(query)
```
woshilapin: I do not understand setting up a new variable `response` since this is the last line of the function.

source/jormungandr/jormungandr/utils.py
```diff
@@ -376,6 +378,57 @@ def generate_id():
...
+        logging.getLogger(__name__).error('Invalid dict_pt_object')
+        raise InvalidArguments('Invalid dict_pt_object')
+    map_pt_object_type = {
+        "stop_point": type_pb2.STOP_POINT,
+        "stop_area": type_pb2.STOP_AREA,
+        "address": type_pb2.ADDRESS,
+        "administrative_region": type_pb2.ADMINISTRATIVE_REGION,
+        "poi": type_pb2.POI,
+        "access_point": type_pb2.ACCESS_POINT,
+    }
```
woshilapin: This doesn't need to be instantiated on each call of the function, it can be a global `const`, isn't it?

source/jormungandr/jormungandr/utils.py
```diff
@@ -376,6 +378,57 @@ def generate_id():
...
+    pt_object.uri = dict_pt_object["id"]
+    pt_object.name = dict_pt_object.get("name", "")
+    map_pt_object = {
+        type_pb2.STOP_POINT: pt_object.stop_point,
+        type_pb2.STOP_AREA: pt_object.stop_area,
+        type_pb2.ADDRESS: pt_object.address,
+        type_pb2.ADMINISTRATIVE_REGION: pt_object.administrative_region,
+        type_pb2.POI: pt_object.poi,
+        type_pb2.ACCESS_POINT: pt_object.access_point,
+    }
```
woshilapin: I'm not sure to understand completely, but doesn't this create a `dict` with multiple entries, but only one will actually have a useful value? If this is the case, I believe we should be able to write the code in a more straightforward way.

source/jormungandr/jormungandr/street_network/kraken.py
```diff
@@ -155,6 +155,12 @@ def _direct_path(
         )
 
         response = instance.send_and_receive(req, request_id=request_id)
+        if response and response.journeys and response.journeys[0].sections:
+            if pt_object_origin.embedded_type == type_pb2.POI:
+                response.journeys[0].sections[0].origin.CopyFrom(pt_object_origin)
+            if pt_object_destination.embedded_type == type_pb2.POI:
+                response.journeys[0].sections[-1].destination.CopyFrom(pt_object_destination)
+
```
woshilapin: I'm not sure that I understand this piece of code. The reason is: I don't understand why we do have a piece of code to fill up `origin` and `destination` that is specific to POI. Wouldn't we have something similar for addresses or stops too?

source/jormungandr/jormungandr/scenarios/new_default.py
```diff
@@ -1196,12 +1200,33 @@ def fill_journeys(self, request_type, api_request, instance):
...
         destination_detail = self.get_entrypoint_detail(
             api_request.get('destination'), instance, request_id="{}_dest_detail".format(request_id)
         )
+
         # we store the origin/destination detail in g to be able to use them after the marshall
         g.origin_detail = origin_detail
         g.destination_detail = destination_detail
+        if not origin_detail:
+            origin_detail = json_address_from_uri(api_request.get('origin'))
+            if not origin_detail:
```
woshilapin: Instead of doing this 2 steps `if`, can't we do the same thing directly on line 1200 with the following code?

```python
        origin_detail = self.get_entrypoint_detail(
            api_request.get('origin'), instance, request_id="{}_origin_detail".format(request_id)
        ) or json_address_from_uri(api_request.get('origin'))
```

Same for `destination_detail`.

source/jormungandr/jormungandr/scenarios/new_default.py
```diff
@@ -1408,9 +1445,47 @@ def isochrone(self, request, instance):
...
+
+        pt_object_origin = None
+        pt_object_destination = None
+        if request.get('origin'):
+            origin_detail = self.get_entrypoint_detail(
+                request.get('origin'), instance, request_id="{}_origin_detail".format(request_id)
+            )
+            if not origin_detail:
+                origin_detail = json_address_from_uri(request.get('origin'))
+                if not origin_detail:
```
woshilapin: Can't we do this instead?
```suggestion
        if request.get('origin'):
            origin_detail = self.get_entrypoint_detail(
                request.get('origin'), instance, request_id="{}_origin_detail".format(request_id)
            ) or origin_detail = json_address_from_uri(request.get('origin'))
            if not origin_detail:
```

source/jormungandr/jormungandr/scenarios/new_default.py
```diff
@@ -1408,9 +1445,47 @@ def isochrone(self, request, instance):
         # Initialize a context for distributed
         distributed_context = self.get_context()
         request_id = request.get("request_id", None)
+
+        pt_object_origin = None
+        pt_object_destination = None
+        if request.get('origin'):
```
woshilapin: `request.get("origin")` is called 4 times in the following lines. Maybe putting it in a variable would be interesting.

source/jormungandr/jormungandr/scenarios/new_default.py
```diff
@@ -1408,9 +1445,47 @@ def isochrone(self, request, instance):
...
+            origin_detail = self.get_entrypoint_detail(
```
woshilapin: The logic of using `get_entry_detail`, then using `json_address_from_uri` as plan B is already done line [1212](https://github.com/hove-io/navitia/pull/3970/files#diff-6f8087c75aa4667ea11ebca99cccdd3f0d228658d83a1d6d13bf58b375a0e21dR1212). Maybe there is an opportunity for factorizing code here.

source/jormungandr/jormungandr/scenarios/distributed.py
```diff
@@ -529,10 +550,42 @@ def graphical_isochrones(self, request, instance):
             ] = instance.additional_time_after_first_section_taxi
 
         krakens_call = set({(request["origin_mode"][0], request["destination_mode"][0], "indifferent")})
+        pt_object_origin = None
+        pt_object_destination = None
+        request_id = request.get("request_id", None)
+        if request.get('origin'):
```
woshilapin: `request.get('origin')` is called 3 times in the following lines, it can probably be extracted in a variable.

source/jormungandr/jormungandr/scenarios/distributed.py
```diff
@@ -529,10 +550,42 @@ def graphical_isochrones(self, request, instance):
...
+            origin_detail = self.get_entrypoint_detail(
+                request.get('origin'), instance, request_id="{}_origin_detail".format(request_id)
+            )
```
woshilapin, line 568: Is it intentional that in `distributed` (compared to `new_default`), we don't try the following plan B?

```python
if not origin_detail:
    origin_detail = json_address_from_uri(api_request.get('origin'))
```

If this is intentional, I'm curious of why (probably an opportunity for me to learn something). If this is not intentional, then maybe there is a possibility to factorize functionality between `distributed` and `new_default`.

source/jormungandr/jormungandr/street_network/kraken.py
azime: No need to do it for other objects

source/jormungandr/jormungandr/scenarios/new_default.py
azime: No, _**self.get_entrypoint_detail()**_  can be None and we want in **_g.origin_detail_** None but not in origin_detail of line 1212 if it is an address

woshilapin: Ok, then maybe, instead this nested `if not origin_detail`, you can write the following to have only one `if` condition.

```python
origin_detail = origin_detail or json_address_from_uri(api_request.get('origin'))
if not origin_detail:
  ...
```

source/jormungandr/jormungandr/scenarios/distributed.py
xlqian: @woshilapin 
Given
https://github.com/hove-io/navitia/pull/3970/files#diff-ce8029605f769998ff99fc1ac7cf0ba49f3dbd60564c109fdd1242d92324039eR476

https://github.com/hove-io/navitia/pull/3970/files#diff-ce8029605f769998ff99fc1ac7cf0ba49f3dbd60564c109fdd1242d92324039eR436

I'm not even sure graphical isochrone is handled in `distributed`... 

but I take your points

source/jormungandr/jormungandr/scenarios/distributed.py
```diff
@@ -462,16 +472,27 @@ def call_kraken(self, request_type, request, instance, krakens_call, request_id,
         Note that the cleaning process depends on the implementation of futures.
         """
         try:
-            with FutureManager(self.greenlet_pool_size) as future_manager, timed_logger(
-                logger, 'call_kraken', request_id
-            ):
+            with FutureManager(1) as future_manager, timed_logger(logger, 'call_kraken', request_id):
```
xlqian:
```suggestion
            with FutureManager(self.greenlet_pool_size) as future_manager, timed_logger(logger, 'call_kraken', request_id):
```

azime: :+1:

source/jormungandr/tests/isochrone_tests.py
azime: Done

source/jormungandr/jormungandr/utils.py
azime: Done

source/jormungandr/jormungandr/scenarios/new_default.py
azime: Done

source/jormungandr/jormungandr/scenarios/new_default.py
azime: Done

source/jormungandr/jormungandr/scenarios/distributed.py
azime: Done

source/jormungandr/jormungandr/scenarios/distributed.py
azime: Done

source/jormungandr/jormungandr/scenarios/new_default.py
azime: Done

source/jormungandr/jormungandr/scenarios/new_default.py
azime: Done

source/jormungandr/jormungandr/scenarios/distributed.py
```diff
@@ -527,6 +529,12 @@ def finalise_journeys(self, request, responses, context, instance, is_debug, req
             final_e = FinaliseException(e)
             return [final_e.get()]
 
+    def get_detail_pt_object(self, instance, arg_pt_object, request_id):
+        if not arg_pt_object:
+            return None
+        detail = self.get_entrypoint_detail(arg_pt_object, instance, request_id="{}".format(request_id))
```
woshilapin:
```suggestion
        detail = self.get_entrypoint_detail(arg_pt_object, instance, request_id=request_id)
```

source/jormungandr/jormungandr/utils.py
woshilapin: We discussed it together and couldn't find a way to make it simpler. An alternative writing of the logic would imply multiple `if/else` which would cause a lot more lines. Either way is fine.

source/jormungandr/jormungandr/street_network/kraken.py
woshilapin: How about adding a comment, something like this?
```python
# In case of a POI, it might be an access point, but we want the parent POI to be shown on.
```

woshilapin (commented):
Sonar show 2 code smells that probably can be addressed.
https://sonarcloud.io/project/issues?id=Hove_navitia&pullRequest=3970&resolved=false&types=CODE_SMELL

source/jormungandr/jormungandr/street_network/kraken.py
azime: :+1:

source/jormungandr/jormungandr/scenarios/distributed.py
azime: Done

woshilapin (approved):

xlqian (approved):

PR #3980, opened 2023-04-20
fix portable min

pbench:
Default keyword arg must be provided i[n python 3.9](https://docs.python.org/3.9/library/functions.html?highlight=min#min), otherwise a ValueError is raised, and that [happen](https://navitia.atlassian.net/browse/NAV-1971)

JIRA: https://navitia.atlassian.net/browse/NAV-1971

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -8,7 +8,7 @@ repos:
     rev: 3.9.2
     hooks:
     - id: flake8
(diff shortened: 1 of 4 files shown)


pbougue (approved):
:exploding_head:

azime (approved):

pbougue (approved):
:+1: for tests

PR #3982, opened 2023-04-20
poi_access_points added for instance

azime:


pbench (approved):

xlqian (approved):

woshilapin (approved):
Do we need to do something on `navitia-configurator` for this new parameter?

Also, Sonar is complaining about coverage. I'm not sure if that's possible to improve it for such a PR, I guess you know better than me.

xlqian:
@woshilapin Indeed, we need to add that parameter in `navitia-configurator` to expose it, but nothing will be broken. in terms of sonar, by and large, codes in `tyr` are covered poorly... we do test the migration (both upgrade and downgrade), but the HTTP API is barely tested...

PR #3983, opened 2023-04-24
build jormun on arm64

xlqian:


pbench (approved):

azime (approved):

.github/workflows/build_dockers_debian11.yml
```diff
@@ -102,3 +102,72 @@ jobs:
...
+  build_navitia_arm64:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+        with:
+          submodules : recursive
+          fetch-depth: 0
+
+      - name: Set up QEMU
```
woshilapin, line 115: @prhod Using QEMU (emulation layer) to build ARM64 images takes a very long time. Do you think it would be feasible to deploy Github runners on ARM64 so the compilation would be straightforward?

docker/build_navitia.sh
```diff
@@ -1,5 +1,5 @@
 #!/bin/bash
 cd /navitia/navitia/
-mkdir -p build && cd build
+mkdir -p docker_build && cd docker_build
 cmake -DCMAKE_BUILD_TYPE=Release ../source
-make -j$(($(grep -c '^processor' /proc/cpuinfo)+1)) kraken
+make -j$(($(grep -c '^processor' /proc/cpuinfo)+1)) protobuf_files
```
woshilapin, line 5: I'm not sure that I understand why you can make that change. 2 different `make` targets. Won't it break something?

source/benchmark/bench.py
```diff
@@ -62,7 +62,7 @@ def _call_jormun(url):
     import time
 
     time1 = time.time()
-    r = requests.get(url, headers={'Authorization': TOKEN})
+    r = requests.get(url, headers={'Authorization': TOKEN}, verify=True)
```
woshilapin: Isn't it already the default behavior? Do we need to be explicit about it?

docker/debian11_arm64/Dockerfile-jormungandr
```diff
@@ -0,0 +1,61 @@
+FROM arm64v8/python:3.9-slim-bullseye as build_arm64
```
woshilapin: :shrug: 
```suggestion
FROM --platform=linux/arm64/v8 python:3.9-slim-bullseye as build_arm64
```

docker/build_navitia.sh
xlqian: no... the `protobuf_files` is needed by jormungandr and kraken... since we build only jormungandr, `protobuf_files` is enough

docker/debian11_arm64/Dockerfile-jormungandr
xlqian: let's try it

xlqian: done! it has worked

PR #3996, opened 2023-05-05
[Jormungandr] [Ridesharing]: Add two params while calling OuestGo

kadhikari:
Two more parameters added while calling OuestGo
- p[outward][mindate] = day of query
- p[outward][maxdate] = day of query + 1 day

Departure origin and pickup point is the same.
Drop off and arrival point is the same.

For more details:  https://navitia.atlassian.net/browse/NAV-2006

### DIFF
--- a/source/jormungandr/jormungandr/scenarios/ridesharing/ouestgo.py
+++ b/source/jormungandr/jormungandr/scenarios/ridesharing/ouestgo.py
@@ -39,8 +39,8 @@ from jormungandr.scenarios.ridesharing.ridesharing_service import (
     RsFeedPublisher,
     RidesharingServiceError,
 )
-from jormungandr.utils import Coords, get_weekday, make_timestamp_from_str
-from jormungandr.street_network.utils import crowfly_distance_between, get_manhattan_duration
+from jormungandr.utils import get_weekday, make_timestamp_from_str, timestamp_to_date_str, ONE_DAY
+from jormungandr.timezone import get_default_timezone
 
 DEFAULT_OUESTGO_FEED_PUBLISHER = {
     'id': 'OUESTGO',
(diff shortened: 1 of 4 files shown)


source/jormungandr/jormungandr/utils.py
```diff
@@ -62,6 +62,8 @@
 WEEK_DAYS_MAPPING = ("monday", "tuesday", "wednesday", "thursday", "friday", "saturday", "sunday")
 COVERAGE_ANY_BETA = "any-beta"
 BEST_BOARDING_POSITION_KEY = "{}-{}"
+ONE_DAY = 86400
```
pbougue, line 65: More pythonic: You can probably use `datetime.timedelta(days=1).total_seconds()` instead of `86400`.

source/jormungandr/jormungandr/scenarios/ridesharing/ouestgo.py
```diff
@@ -211,6 +188,12 @@ def _request_journeys(self, from_coord, to_coord, period_extremity, instance_par
             'p[to][longitude]': arr_lon,
             'signature': 'toto',
             'timestamp': period_extremity.datetime,
+            'p[outward][mindate]': timestamp_to_date_str(
+                period_extremity.datetime, timezone, _format="%Y-%m-%d"
```
pbougue: Minor: looks better to use the already defined one (or let the default one?)
```suggestion
                period_extremity.datetime, timezone, _format=DATE_FORMAT
```

source/jormungandr/jormungandr/timezone.py
```diff
@@ -81,3 +81,10 @@ def get_timezone():
     if not hasattr(g, 'timezone'):
         raise TechnicalError("No timezone set for this API")  # the set_request_timezone has to be called at init
     return g.timezone
+
+
+def get_default_timezone():
```
pbougue: I think of a clearer name IMO:
```suggestion
def get_timezone_or_paris():
```

source/jormungandr/jormungandr/scenarios/ridesharing/ouestgo.py
```diff
@@ -200,7 +177,7 @@ def _request_journeys(self, from_coord, to_coord, period_extremity, instance_par
         """
         dep_lat, dep_lon = from_coord.split(',')
         arr_lat, arr_lon = to_coord.split(',')
-
+        timezone = get_default_timezone()
```
pbougue: It deserves a comment that TZ for mindate/maxdate is the coverage's one, or Paris IMO:
```suggestion
        timezone = get_default_timezone()  # using coverage's TZ (or Paris) for mindate and maxdate
```

pbougue (commented):
It's not clearly specified anywhere that OuestGo garanties that `pickup=from` and `dropoff=to`.
Same goes for the TZ of `mindate` and `maxdate`.

If @wassimbenaissa (or @stifoon) is OK with those, code looks OK to me (just comments/naming).

source/jormungandr/jormungandr/scenarios/ridesharing/ouestgo.py
```diff
@@ -123,42 +123,19 @@ def _make_response(self, raw_json, request_datetime, from_coord, to_coord, insta
...
-                pickup_coord = Coords(lat=pickup_lat, lon=pickup_lon)
-
-                res.pickup_place = rsj.Place(addr='', lat=pickup_lat, lon=pickup_lon)
-
+                res.pickup_place = rsj.Place(addr='', lat=float(lat), lon=float(lon))
                 res.origin_pickup_shape = None  # Not specified
-                res.origin_pickup_distance = int(crowfly_distance_between(departure_coord, pickup_coord))
-                # we choose to calculate with speed=1.12 the average speed for a walker
-                res.origin_pickup_duration = get_manhattan_duration(
-                    res.origin_pickup_distance, instance_params.walking_speed
```
pbougue, line 141: Seen in sonar: You may remove `instance_params` from the fn.

source/jormungandr/jormungandr/timezone.py
```diff
@@ -81,3 +81,10 @@ def get_timezone():
...
+    try:
+        return get_timezone()
+    except:
```
pbougue: Seen in sonar: you may prefer (to avoid catch of BaseException and for a "cleaner" code)
```suggestion
    except Exception as _:
```

kadhikari: :+1:

source/jormungandr/jormungandr/scenarios/ridesharing/ouestgo.py
kadhikari: :+1:

source/jormungandr/jormungandr/scenarios/ridesharing/ouestgo.py
kadhikari: :+1:

source/jormungandr/jormungandr/scenarios/ridesharing/ouestgo.py
kadhikari: :+1:

wassimbenaissa (approved):

pbougue (approved):

PR #3998, opened 2023-05-10
Stop attractivity criteria

pbench:
Adds two new criteria for loki

### DIFF
--- a/source/jormungandr/jormungandr/interfaces/v1/journey_common.py
+++ b/source/jormungandr/jormungandr/interfaces/v1/journey_common.py
@@ -513,7 +513,9 @@ class JourneyCommon(ResourceUri, ResourceUtc):
 
         parser_get.add_argument(
             "criteria",
-            type=OptionValue(['classic', 'robustness', 'occupancy']),
+            type=OptionValue(
+                ['classic', 'robustness', 'occupancy', 'arrival_stop_capacity', 'departure_stop_capacity']
+            ),
             help="choose the criteria used to compute pt journeys ",
         )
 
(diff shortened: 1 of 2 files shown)


source/jormungandr/jormungandr/interfaces/v1/journey_common.py
```diff
@@ -513,7 +513,15 @@ def __init__(self, output_type_serializer):
...
-            type=OptionValue(['classic', 'robustness', 'occupancy']),
+            type=OptionValue(
+                [
+                    'classic',
+                    'robustness',
+                    'occupancy',
+                    'arrival_stop_attractivity',
+                    'departure_stop_attractivity',
+                ]
+            ),
```
pbougue, line 524: Shouldn't we hide the param for now and make it beta?

pbougue (approved):

source/jormungandr/jormungandr/interfaces/v1/journey_common.py
pbench: The "criteria" param already exists, and I think it needs to be in the navitia api schema to for our frontends.
I don't know if we can hide some values of a param ?

pbougue: I don't think we can hide only values.
We may just document that they are beta below?
I don't know any front that uses the param itself, but you must know better, I trust you if you think it's not a good idea to hide the whole param.

xlqian (approved):

source/jormungandr/jormungandr/interfaces/v1/journey_common.py
pbench: I documented that the `criteria` feature is in beta

pbougue (approved):

azime (approved):

PR #4000, opened 2023-05-11
[Kraken] remove/merge RT handling duration metrics

pbougue:
The output was not relevant as messages are handled in batch and a batch can contain different types: merging all into one metric and keep only logs if need arises for more details.


pbench (approved):
It was indeed strange to use the last applied entity to choose in which histogram to store whole duration of the batch :exploding_head: 

However, it might be useful to have distinct histograms (chaos/kirin/cancel chaos) for the duration of the application of a single entity

pbougue:
I agree about that, this PR cleans things but if the need for differentiated metrics on single entities ever arises, it would be legitimate IMO.

kadhikari (approved):

PR #4001, opened 2023-05-12
Keep olympics journeys

pbench:
Add a hidden request parameter `_jo`. When set to `true`, any journey tagged `olympics` will not be deleted by jormun filtering.
Indeed, sometimes an `olympics` journey is filtered by jormun because it uses the same line as another journey, but arrives later & walks more (we want to keep the olympics journeys because it ends at a specific stops).
The tag `olympics` is [set in Loki ](https://github.com/hove-io/loki/pull/435) for now

### DIFF
--- a/source/jormungandr/jormungandr/interfaces/v1/journey_common.py
+++ b/source/jormungandr/jormungandr/interfaces/v1/journey_common.py
@@ -525,6 +525,14 @@ class JourneyCommon(ResourceUri, ResourceUtc):
             help="choose the criteria used to compute pt journeys, feature in beta ",
         )
 
+        parser_get.add_argument(
+            "_jo",
+            type=BooleanType(),
+            hidden=True,
+            default=False,
+            help="do not delete journeys tagged for olympics",
+        )
+
         # Advanced parameters for valhalla bike
         parser_get.add_argument(
             "bike_use_roads",
--- a/source/jormungandr/jormungandr/interfaces/v1/journey_common.py
+++ b/source/jormungandr/jormungandr/interfaces/v1/journey_common.py
@@ -526,7 +526,7 @@ class JourneyCommon(ResourceUri, ResourceUtc):
         )
 
         parser_get.add_argument(
-            "_jo",
+            "_keep_olympics_journeys",
             type=BooleanType(),
             hidden=True,
             default=False,
--- a/source/jormungandr/jormungandr/scenarios/journey_filter.py
+++ b/source/jormungandr/jormungandr/scenarios/journey_filter.py
@@ -45,17 +45,16 @@ def delete_journeys(responses, request):
     if request.get('debug', False):
         return
 
-    keep_jo = request.get("_jo", False)
-
-    def to_delete(journey):
-        if keep_jo:
-            return to_be_deleted(journey) and "olympics" not in journey.tags
-        else:
-            return to_be_deleted(journey)
+    keep_olympics_journeys = request.get("_keep_olympics_journeys", False)
 
     nb_deleted = 0
     for r in responses:
-        nb_deleted += pb_del_if(r.journeys, lambda j: to_delete(j))
+        if keep_olympics_journeys:
+            for journey in r.journeys:
+                if 'to_delete' in journey.tags and 'olympics' in journey.tags:
+                    journey.tags.remove("to_delete")
+
+        nb_deleted += pb_del_if(r.journeys, lambda j: to_be_deleted(j))
 
     if nb_deleted:
         logging.getLogger(__name__).info('filtering {} journeys'.format(nb_deleted))
--- a/source/jormungandr/jormungandr/scenarios/journey_filter.py
+++ b/source/jormungandr/jormungandr/scenarios/journey_filter.py
@@ -45,9 +45,17 @@ def delete_journeys(responses, request):
     if request.get('debug', False):
         return
 
+    keep_jo = request.get("_jo", False)
+
+    def to_delete(journey):
+        if keep_jo:
+            return to_be_deleted(journey) and "olympics" not in journey.tags
+        else:
+            return to_be_deleted(journey)
+
     nb_deleted = 0
     for r in responses:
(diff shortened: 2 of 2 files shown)


source/jormungandr/jormungandr/interfaces/v1/journey_common.py
```diff
@@ -525,6 +525,14 @@ def __init__(self, output_type_serializer):
             help="choose the criteria used to compute pt journeys, feature in beta ",
         )
 
+        parser_get.add_argument(
+            "_jo",
+            type=BooleanType(),
+            hidden=True,
+            default=False,
+            help="do not delete journeys tagged for olympics",
```
pbougue: IMO: :fr: naming should be avoided and the olympics is also too specific.
```suggestion
            "_protect",
            type=BooleanType(),
            hidden=True,
            default=False,
            help="Enable protection (from deletion) for journeys protected because of \"subjective\" extra criteria",
```

source/jormungandr/jormungandr/scenarios/journey_filter.py
```diff
@@ -45,9 +45,17 @@ def delete_journeys(responses, request):
     if request.get('debug', False):
         return
 
+    keep_jo = request.get("_jo", False)
+
+    def to_delete(journey):
+        if keep_jo:
+            return to_be_deleted(journey) and "olympics" not in journey.tags
```
pbougue: Same, I'd avoid being too specific:
```suggestion
            return to_be_deleted(journey) and "protected" not in journey.tags
```

source/jormungandr/jormungandr/scenarios/journey_filter.py
```diff
@@ -45,9 +45,17 @@ def delete_journeys(responses, request):
     if request.get('debug', False):
         return
 
+    keep_jo = request.get("_jo", False)
+
+    def to_delete(journey):
```
xlqian: I recommend you take https://github.com/hove-io/navitia/blob/fe7c69a69dc4cb6331d98adf920f94da3d72da96/source/jormungandr/jormungandr/scenarios/journey_filter.py#L164 as an example how filters are handled... it seems to be more tedious, in practice,  all filters are gathered in only one place... and it would be easier for the user to debug...

source/jormungandr/jormungandr/interfaces/v1/journey_common.py
pbench: Well, the `olympics` tag will be included in the navitia response, and is named as such to be understandable by the client/frontend. 
I think being specific here is needed, and replacing the `olympics` tag  by `protected` will only make things more obscure.

On the name of the parameter, maybe replace `_jo` by `_olympics` ? It may be useful for testing other features related to olympics. 

poke @wassimbenaissa what do you think ?

source/jormungandr/jormungandr/scenarios/journey_filter.py
pbench: Maybe I don't get your comment, but I want some journeys **not** to be filtered by any filter. 

For example, on this [request](https://playground.navitia.io/play.html?request=https%3A%2F%2Fhormungandr.corefront.sbx.aws.private%2Fv1%2Fcoverage%2Fidfm%2Fjourneys%3Ffrom%3Dstop_area%253AIDFM%253A71607%26to%3Dpoi%253A14945%26datetime%3D20230512T121903%26_pt_planner%3Dloki%26debug%3Dtrue%26criteria%3Darrival_stop_attractivity%26_walking_transfer_penalty%3D360%26) the second journey is filtered because it is similar to the first, but I want to keep it because it ends at a stop that have a high capacity.
Note that we also want to keep the first journey.

source/jormungandr/jormungandr/interfaces/v1/journey_common.py
woshilapin: I'd agree with @pbench at this point, it's already specific, let's not make something generic. Since we don't even know what the future holds (planning a feature for an unknown future often ends up being a feature for only one specific case).

But I do agree :gb: `_olympics` would be better than :fr: `_jo`. Even better maybe, `_keep_olympics` which would be even more explicit about what it does.

source/jormungandr/jormungandr/scenarios/journey_filter.py
woshilapin: `to_be_deleted` seems to work on the presence (or absence) of the tag `to_delete`. I'd say it'd be more logical to reuse the same mechanism everywhere in the code instead of creating some other logic to handle journeys that must be kept. So I'd suggest to do something along these lines instead.

```python
def to_delete(journey):
  if keep_jo:
    journeys.tags.remove('to_delete')
  to_be_deleted(journey)
```

source/jormungandr/jormungandr/interfaces/v1/journey_common.py
pbougue: OK, you've understood my point about being generic (which also forces us to be able to explain what's done), the final decision is not a blocker either way for me.

Another underlying point (not blocking either IMO, especially as it is a hidden param) is, do we want:
1. `_olympics` > "large" parameter that is only protecting some journeys now but is probably going to do more (in that case maybe document it a bit differently?).
2. `_keep_olympics` > narrower parameter that describes better just what's done (protect journeys tagged "olympics").

pbench: ok :+1: , let's go with the more specific one '_keep_olympics_journeys'

source/jormungandr/jormungandr/scenarios/journey_filter.py
pbench: good idea ! I remove the 'to_delete' tag in https://github.com/hove-io/navitia/pull/4001/commits/dc2d0668118587199c12bd776a907dc38e1e5ffd

pbougue (approved):

woshilapin (approved):

PR #4004, opened 2023-05-15
Doc: Parameter "key" no longer available for authentication

stifoon:
Parameter "key" no longer available


azime (approved):

pbench (approved):
Should we also remove [the code](https://github.com/hove-io/navitia/blob/e4c24597de8e05328ed330d0538f13e72fef28dc/source/jormungandr/jormungandr/authentication.py#L94-L95) accepting the token in the `key` parameter ?

kadhikari (approved):

kadhikari:
I suggest to add:
"will be deprecated and hence will be deleted asap" just above
https://github.com/hove-io/navitia/blob/e4c24597de8e05328ed330d0538f13e72fef28dc/source/jormungandr/jormungandr/authentication.py#L94-L95

woshilapin (approved):

kadhikari (approved):

PR #4021, opened 2023-05-25
Kraken: Smarter sorting of next_stop_time

pbench:
The function `NextStopTimeData::TimesStopTimes<Getter>::init` takes a list of `stop_times`, and sort them.
However, when two `stop_times` are equal, the function compares the earliest stop_time of the corresponding vehicle_journeys. And to determine the earliest `stop_time` of a `vehicle_journey`, we need to scan all its `stop_times`.

So if we need to sort N `stop_times` that belongs to `vehicle_journeys` with at most K `stop_times`, this means that in the worst case each call to the comparison operator is O(K). Sorting may perform O(N log N) calls to the comparison operator, resulting in a worst-case behavior of O(K N log N).

Instead of determining the earliest `stop_time` of a `vehicle_journey` inside the comparison operator, this PR moves this operation before the sort.

- We first scan all the `stop_times` to be sorted, and compute the earliest `stop_time` of their corresponding `vehicle_journey`. We store the `stop_time`, along its corresponding earliest `stop_time` inside a temporary vector. Overall an O(K N) operation
- We then perform the sort on the temporary vector. Since the earliest `stop_times` are already computed, the comparison operator is O(1)
- Then we copy the sorted `stop_times` in the `self.stop_times` vector that is used afterward for journeys computation.

This results in a worst case behavior of O(K N + N Log N) improving on the previous O(K N log N).

In practice, the time to build the `dataRaptor` was reduced from 2mn to 30s on `fr-se-lyon`, and did not change on `idfm`.

https://navitia.atlassian.net/browse/NAV-1806

### DIFF
--- a/source/routing/next_stop_time.cpp
+++ b/source/routing/next_stop_time.cpp
@@ -61,31 +61,43 @@ bool NextStopTimeData::Arrival::is_valid(const type::StopTime& st) const {
 
 template <typename Getter>
 void NextStopTimeData::TimesStopTimes<Getter>::init(const JourneyPattern& jp, const JourneyPatternPoint& jpp) {
+    using StopTimePair = std::pair<const type::StopTime*, const type::StopTime*>;
     // collect the stop times at the given jpp
     const auto jpp_order = jpp.order;
-    stop_times.reserve(jp.discrete_vjs.size());
+    std::vector<StopTimePair> stop_times_and_earliest_stop_time;
+    stop_times_and_earliest_stop_time.reserve(jp.discrete_vjs.size());
     for (const auto& vj : jp.discrete_vjs) {
-        const auto& st = get_corresponding_stop_time(*vj, jpp_order);
+        const type::StopTime& st = get_corresponding_stop_time(*vj, jpp_order);
         if (!getter.is_valid(st)) {
             continue;
         }
-        stop_times.push_back(&st);
+        const type::StopTime& earliest_stop_time = navitia::earliest_stop_time(st.vehicle_journey->stop_time_list);
+        StopTimePair pair = StopTimePair(&st, &earliest_stop_time);
+        stop_times_and_earliest_stop_time.push_back(pair);
     }
 
     // sort the stop times in ascending order
-    boost::sort(stop_times, [&](const type::StopTime* st1, const type::StopTime* st2) {
+    boost::sort(stop_times_and_earliest_stop_time, [&](StopTimePair pair1, StopTimePair pair2) {
+        const type::StopTime* st1 = pair1.first;
+        const type::StopTime* st2 = pair2.first;
         const auto time1 = DateTimeUtils::hour(getter.get_time(*st1));
         const auto time2 = DateTimeUtils::hour(getter.get_time(*st2));
         if (time1 != time2) {
             return time1 < time2;
         }
-        const auto& st1_first = navitia::earliest_stop_time(st1->vehicle_journey->stop_time_list);
-        const auto& st2_first = navitia::earliest_stop_time(st2->vehicle_journey->stop_time_list);
-        if (getter.get_time(st1_first) != getter.get_time(st2_first)) {
-            return getter.get_time(st1_first) < getter.get_time(st2_first);
+        const type::StopTime* st1_first = pair1.second;
+        const type::StopTime* st2_first = pair2.second;
+        auto st1_first_time = getter.get_time(*st1_first);
+        auto st2_first_time = getter.get_time(*st2_first);
+        if (st1_first_time != st2_first_time) {
+            return st1_first_time < st2_first_time;
         }
-        return st1_first.vehicle_journey->idx < st2_first.vehicle_journey->idx;
+        return st1_first->vehicle_journey->idx < st2_first->vehicle_journey->idx;
     });
+    stop_times.reserve(jp.discrete_vjs.size());
+    for (const auto& pair : stop_times_and_earliest_stop_time) {
+        stop_times.push_back(pair.first);
+    }
 
     // collect the corresponding times
     times.reserve(stop_times.size());
--- a/source/routing/next_stop_time.cpp
+++ b/source/routing/next_stop_time.cpp
@@ -72,8 +72,7 @@ void NextStopTimeData::TimesStopTimes<Getter>::init(const JourneyPattern& jp, co
             continue;
         }
         const type::StopTime& earliest_stop_time = navitia::earliest_stop_time(st.vehicle_journey->stop_time_list);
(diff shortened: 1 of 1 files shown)


xlqian (approved):

source/routing/next_stop_time.cpp
```diff
@@ -61,31 +61,43 @@ bool NextStopTimeData::Arrival::is_valid(const type::StopTime& st) const {
...
     for (const auto& vj : jp.discrete_vjs) {
-        const auto& st = get_corresponding_stop_time(*vj, jpp_order);
+        const type::StopTime& st = get_corresponding_stop_time(*vj, jpp_order);
         if (!getter.is_valid(st)) {
             continue;
         }
-        stop_times.push_back(&st);
+        const type::StopTime& earliest_stop_time = navitia::earliest_stop_time(st.vehicle_journey->stop_time_list);
+        StopTimePair pair = StopTimePair(&st, &earliest_stop_time);
+        stop_times_and_earliest_stop_time.push_back(pair);
```
xlqian:
```suggestion
        stop_times_and_earliest_stop_time.emplace_back(&st, &earliest_stop_time);
```

source/routing/next_stop_time.cpp
```diff
@@ -61,31 +61,42 @@ bool NextStopTimeData::Arrival::is_valid(const type::StopTime& st) const {
...
+    boost::sort(stop_times_and_earliest_stop_time, [&](StopTimePair pair1, StopTimePair pair2) {
+        const type::StopTime* st1 = pair1.first;
+        const type::StopTime* st2 = pair2.first;
         const auto time1 = DateTimeUtils::hour(getter.get_time(*st1));
         const auto time2 = DateTimeUtils::hour(getter.get_time(*st2));
         if (time1 != time2) {
             return time1 < time2;
         }
-        const auto& st1_first = navitia::earliest_stop_time(st1->vehicle_journey->stop_time_list);
-        const auto& st2_first = navitia::earliest_stop_time(st2->vehicle_journey->stop_time_list);
```
pbougue, line 83: How about infinite caching (hash_map) of this instead? (searching for one stop_time is caching for all its friends of the same vj).
Theoretical complexity and principle would be the same, but it should save a lot of useless `earliest_stop_time()` calls, no?
And also the final reallocation, but that's probably negligible.

source/routing/next_stop_time.cpp
```diff
@@ -61,31 +61,42 @@ bool NextStopTimeData::Arrival::is_valid(const type::StopTime& st) const {
...
+    std::vector<StopTimePair> stop_times_and_earliest_stop_time;
+    stop_times_and_earliest_stop_time.reserve(jp.discrete_vjs.size());
     for (const auto& vj : jp.discrete_vjs) {
-        const auto& st = get_corresponding_stop_time(*vj, jpp_order);
+        const type::StopTime& st = get_corresponding_stop_time(*vj, jpp_order);
         if (!getter.is_valid(st)) {
             continue;
         }
-        stop_times.push_back(&st);
+        const type::StopTime& earliest_stop_time = navitia::earliest_stop_time(st.vehicle_journey->stop_time_list);
```
pbougue, line 74: This is a bigger work, but it looks like we should (and probably already do) guarantee that `vj.stop_time_list` is sorted (both in time and order).

pbougue (approved):
:heart_eyes: Nice!

azime (approved):

source/routing/next_stop_time.cpp
pbench: We could cache this, but I don't know where to put the cache. 
The cache would make sense only if it lives outside of the `init()` function : on a single call to `init()`, I don't expect to have many stop_times belonging to the same vj.

source/routing/next_stop_time.cpp
pbench: Yep, it would make more sense to have `vj.stop_time_list ` sorted. Maybe it already is ? I did not check

pbougue: I tried to have a look, and it's hard to have a guarantee, probably also the boarding/alighting times can mess things up.
But sharing the work upfront would certainly help... Although it's a bigger job.

source/routing/next_stop_time.cpp
pbougue: Side-effects are not nice, but that's the only way I can think of for a shared cache here.

Another lead is to lower the cache and cache only the given stop-time, which would make the first comparison O(K) and let the other comparisons O(1) resulting in the same theoretical complexity than what you did, but saving all the cases where getting first-stop is useless, I think.

pbougue: Maybe onboard general cache structure in NextStopTimeData, and clean it up at a smart time.

source/routing/next_stop_time.cpp
```diff
@@ -61,31 +61,42 @@ bool NextStopTimeData::Arrival::is_valid(const type::StopTime& st) const {
...
         if (time1 != time2) {
             return time1 < time2;
         }
-        const auto& st1_first = navitia::earliest_stop_time(st1->vehicle_journey->stop_time_list);
-        const auto& st2_first = navitia::earliest_stop_time(st2->vehicle_journey->stop_time_list);
-        if (getter.get_time(st1_first) != getter.get_time(st2_first)) {
-            return getter.get_time(st1_first) < getter.get_time(st2_first);
+        const type::StopTime* st1_first = pair1.second;
+        const type::StopTime* st2_first = pair2.second;
+        auto st1_first_time = getter.get_time(*st1_first);
```
woshilapin, line 89: I'm curious, why not using `DateTimeUtils::hour(getter.get_time(*st1))` like the above comparison?

woshilapin (approved):

source/routing/next_stop_time.cpp
pbench: To be honest I don't known ! I just kept the comparison that was made before.
Certainly something about after-midnight stop_times :thinking:

source/routing/next_stop_time.cpp
pbench: > Another lead is to lower the cache and cache only the given stop-time, which would make the first comparison O(K) and let the other comparisons O(1) resulting in the same theoretical complexity than what you did, but saving all the cases where getting first-stop is useless, I think.

Yes, it may be slightly better, but I am not sure either: using a hashmap is slightly more expensive than a plain vector.
Anyway, I don't think the potential gains area worth the time spent trying.

However, I do think it is worth keeping in mind for future work the sorting of `vj.stop_time_list`, and removing the need to compute/cache the earliest stop_time

PR #4028, opened 2023-06-05
Fix parsing error on allowed_id for same_journey_schedules in Python3

kadhikari:
Finally we use to_dict(flat=False) on request.args. Thanks to @pbougue 
For details: https://navitia.atlassian.net/browse/NAV-1998

### DIFF
--- a/source/jormungandr/jormungandr/interfaces/v1/Journeys.py
+++ b/source/jormungandr/jormungandr/interfaces/v1/Journeys.py
@@ -159,7 +159,7 @@ class add_journey_href(object):
                     args['direct_path'] = 'only' if 'non_pt' in journey['tags'] else 'none'
                     args['min_nb_journeys'] = 5
                     args['is_journey_schedules'] = True
-                    param_values = vars(allowed_id_args).get('allowed_id[]', [])
+                    param_values = allowed_id_args.get('allowed_id[]') or []
                     allowed_ids.update(param_values)
                     args['allowed_id[]'] = list(allowed_ids)
                     args['_type'] = 'journeys'
--- a/source/jormungandr/jormungandr/interfaces/v1/Journeys.py
+++ b/source/jormungandr/jormungandr/interfaces/v1/Journeys.py
@@ -120,9 +120,9 @@ class add_journey_href(object):
             if has_invalid_reponse_code(objects) or journeys_absent(objects):
                 return objects
 
-            # This is a temporary hack to resolve Error introduced during migration to Python3
+            # TODO: This is a temporary hack to resolve Error introduced during migration to Python3
             # Instead of using wrongly parsed request.args we use original args to get existing allowed_id[]
-            allowed_id_args = args[0].parsers['get'].parse_args()
+            raw_args = args[0].parsers['get'].parse_args()
             for journey in objects[0]['journeys']:
                 args = dict(request.args)
                 allowed_ids = {
@@ -159,7 +159,7 @@ class add_journey_href(object):
                     args['direct_path'] = 'only' if 'non_pt' in journey['tags'] else 'none'
                     args['min_nb_journeys'] = 5
                     args['is_journey_schedules'] = True
-                    param_values = allowed_id_args.get('allowed_id[]') or []
+                    param_values = raw_args.get('allowed_id[]') or []
                     allowed_ids.update(param_values)
                     args['allowed_id[]'] = list(allowed_ids)
                     args['_type'] = 'journeys'
--- a/source/jormungandr/jormungandr/interfaces/v1/Journeys.py
+++ b/source/jormungandr/jormungandr/interfaces/v1/Journeys.py
@@ -120,11 +120,10 @@ class add_journey_href(object):
             if has_invalid_reponse_code(objects) or journeys_absent(objects):
                 return objects
 
-            # TODO: This is a temporary hack to resolve Error introduced during migration to Python3
-            # Instead of using wrongly parsed request.args we use original args to get existing allowed_id[]
-            raw_args = args[0].parsers['get'].parse_args()
             for journey in objects[0]['journeys']:
-                args = dict(request.args)
+                # Note: request.args is a MultiDict, we want to flatten it by having list as value when needed
+                # From Python3.6 onwards dict(request.args) != request.args.to_dict(flat=False)
+                args = request.args.to_dict(flat=False)
                 allowed_ids = {
                     o['stop_point']['id']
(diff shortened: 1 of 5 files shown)


source/jormungandr/jormungandr/interfaces/v1/Journeys.py
```diff
@@ -120,6 +120,9 @@ def wrapper(*args, **kwargs):
             if has_invalid_reponse_code(objects) or journeys_absent(objects):
                 return objects
 
+            # This is a temporary hack to resolve Error introduced during migration to Python3
```
woshilapin: If this is a temporary hack, what will be the long term solution?

woshilapin (commented):
Do we understand what is causing the issue? Can't we fix that instead of making a hack around it?

source/jormungandr/jormungandr/interfaces/v1/Journeys.py
kadhikari: As this is the very first error found on args parsing in Python3, I suppose we can wait for other errors to be detected and hence do more effort to find/make better solution :smile:

xlqian: `request.args` is a global variable from slack used to store arguments parsed previously which is buggy in python3, very likely due to the (extremely) old version of Flask that we are using...

source/jormungandr/jormungandr/interfaces/v1/Journeys.py
```diff
@@ -156,7 +159,8 @@ def wrapper(*args, **kwargs):
                     args['direct_path'] = 'only' if 'non_pt' in journey['tags'] else 'none'
                     args['min_nb_journeys'] = 5
                     args['is_journey_schedules'] = True
-                    allowed_ids.update(args.get('allowed_id[]', []))
+                    param_values = vars(allowed_id_args).get('allowed_id[]', [])
```
xlqian:
```python
allowed_id_args.get('allowed_id[]', []) or []
```

kadhikari: :+1:

xlqian (dismissed):

source/jormungandr/jormungandr/interfaces/v1/Journeys.py
```diff
@@ -120,6 +120,9 @@ def wrapper(*args, **kwargs):
             if has_invalid_reponse_code(objects) or journeys_absent(objects):
                 return objects
 
+            # This is a temporary hack to resolve Error introduced during migration to Python3
+            # Instead of using wrongly parsed request.args we use original args to get existing allowed_id[]
+            allowed_id_args = args[0].parsers['get'].parse_args()
```
pbougue:
```suggestion
            raw_args = args[0].parsers['get'].parse_args()
```

source/jormungandr/jormungandr/interfaces/v1/Journeys.py
```diff
@@ -156,7 +159,8 @@ def wrapper(*args, **kwargs):
                     args['direct_path'] = 'only' if 'non_pt' in journey['tags'] else 'none'
                     args['min_nb_journeys'] = 5
                     args['is_journey_schedules'] = True
-                    allowed_ids.update(args.get('allowed_id[]', []))
+                    param_values = allowed_id_args.get('allowed_id[]') or []
+                    allowed_ids.update(param_values)
```
pbougue: 1. Is it possible to add a unit test on that ? :pray:

2. I suspect the final `[]` might disturb flask, and I think we should add tests for similar params on all endpoints (maybe only manual if automated tests are not possible).

source/jormungandr/jormungandr/interfaces/v1/Journeys.py
pbougue: Maybe we can clarify in comment (TODO?) that this is supposed to be removed once flask is upgraded, then?
Or that we might also just wait for jormungandr to be replaced.

source/jormungandr/jormungandr/interfaces/v1/Journeys.py
kadhikari: :+1:

source/jormungandr/jormungandr/interfaces/v1/Journeys.py
kadhikari: :+1:

kadhikari: Since raw_args is a NameSpace but not a dict, 
- raw_args.get('allowed_id[]', []) gives **None** if parameter **allowed_id[]** is absent.
- raw_args.get('allowed_id[]') or [] gives **[]** if parameter **allowed_id[]** is absent.

azime (approved):

source/jormungandr/jormungandr/interfaces/v1/Journeys.py
pbougue: After digging more, a stable/general correction is now proposed: https://github.com/hove-io/navitia/pull/4028/files#diff-9f144f0945ab2aaeaed8f7771af1bfe0e9a14c86d0317959d37734ff6b0eecedR124-R126

pbougue (approved):

source/jormungandr/jormungandr/interfaces/v1/Journeys.py
pbougue: Thanks for the tests !

source/jormungandr/jormungandr/interfaces/v1/Schedules.py
```diff
@@ -330,7 +330,9 @@ def wrapper(*args, **kwargs):
                     max_dt = dt
             if "links" not in response:
                 response["links"] = []
-            kwargs_links = dict(deepcopy(request.args))
+            # request.args is a MultiDict, we want to flatten it by having list as value when needed
+            # From Python3.6 onwards dict(request.args) != request.args.to_dict(flat=False)
+            kwargs_links = deepcopy(request.args).to_dict(flat=False)
```
xlqian, line 335: why do you need a deepcopy?

xlqian (approved):

source/jormungandr/jormungandr/interfaces/v1/Schedules.py
kadhikari: I think there is a valid reason to deepcopy before correction, so I simply transformed the code.

PR #4038, opened 2023-06-14
All: fix ubuntu 22.04 dev setup

pbougue:
Please review with care, especially changes in `source/routing/labels.cpp` (perfs?)

### DIFF
--- a/source/routing/labels.cpp
+++ b/source/routing/labels.cpp
@@ -40,7 +40,7 @@ Labels::Labels(const std::vector<type::StopPoint*> stop_points) : labels(stop_po
 Labels::Labels(const Map& dt_pts, const Map& dt_transfers, const Map& walkings, const Map& walking_transfers) {
     const auto& zip = boost::combine(dt_pts, dt_transfers.values(), walkings.values(), walking_transfers.values());
     labels.resize(zip.size());
-    for (const auto z : zip) {
+    for (const auto& z : zip) {
         const auto sp_idx = z.get<0>().first;
         labels[sp_idx] = Label{z.get<0>().second, z.get<1>(), z.get<2>(), z.get<3>()};
     }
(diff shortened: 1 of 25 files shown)


source/routing/labels.cpp
```diff
@@ -52,7 +52,7 @@ std::array<Labels::Map, 4> Labels::inrow_labels() {
     for (auto& row_label : row_labels)
         row_label.resize(labels.size());
 
-    for (const auto& l : labels) {
+    for (const auto l : labels) {
```
pbench: I really don't know what are the consequences of this change, the meaning of `auto`  vs `auto&` is not clear for me.
Do you really need this change to enable compilation on ubuntu 22.04 ?

pbougue: This is just a clang warning, it's compilable without it.

As warned by clang, "this (and `z` above) binds to a temporary value produced by a range".
After reading a [bit about it](https://stackoverflow.com/questions/50066139/what-is-clangs-range-loop-analysis-diagnostic-about) this warning asks for explicitness on the fact that actual object (proxy) **may** be produced/copied.
So I guess the point is more that we may dig for more perf... Not the point of this PR (will let the clang warning, though).

So I will revert that change.

pbougue: Done in https://github.com/hove-io/navitia/pull/4038/commits/7a94b050db2aa68af012421dc69d95277fc6f2e8

pbougue:
:sob: On deb8, CMake is too old for prometheus

xlqian (approved):

pbougue:
CMake for debian 8 is fixed (see the [last 2 commits](https://github.com/hove-io/navitia/pull/4038/files/66a6a8264580945f4c81986388544c0764a66fa3..f5ac29b11df07169380e9ca3e541663387d66f7a) for that) :tada:
Docker build is [tested](https://github.com/hove-io/navitia/actions/runs/5392149557/jobs/9790081312) in the [penultimate commit](https://github.com/hove-io/navitia/pull/4038/commits/6f65c02eac9cfd1e7df6a94fe56ca8f727bdd9a0) (last commit just reverts the trigger).
So we should be good to go on that PR.

xlqian (approved):

azime (approved):

PR #4042, opened 2023-06-19
Modification step_1

kadhikari:


.github/workflows/publish_navitia_documentation.yml
```diff
@@ -6,9 +6,12 @@ on:
       - release
   workflow_dispatch:
 
+env:
+  REGION: eu-west-1
+
```
woshilapin, line 11: Do you need that here?

woshilapin (approved):

azime (approved):

.github/workflows/publish_navitia_documentation.yml
kadhikari: Yes
Without doesn't work

PR #4048, opened 2023-06-22
[kraken] disruptions: fix SQL request by adding missing integer cast

vpassama:
Hello,

We found a bug in fill_disruptions_from_database sql request as we use now a newer postgresql version which default behavior has changed on a specific value select (epoch from).
The problem occurs when time_slots are used in CHAOS as there is a missing integer cast in the request.

closes #4066


pbougue (approved):

kadhikari (approved):

pbougue:
Thanks ! :pray:

PR #4057, opened 2023-06-28
clean workspace

azime:


.github/workflows/sonarcloud.yml
```diff
@@ -117,3 +117,11 @@ jobs:
           sonar-scanner-$SONAR_SCANNER_VERSION-linux/bin/sonar-scanner \
                 -Dsonar.login=$SONAR_TOKEN \
                 -Dsonar.cfamily.threads=$(nproc)
+
+    - name: Clean build directory
+      working-directory: ./build
+      run: |
+        make clean
```
pbougue: (same below) I didn't test it, but I think `git clean -xdf` will actually remove the whole `build` directory (that's created inside local sources).
Also `make clean` keeps `cmake` artefacts and cache, so if necessary, `rm -rf ./build` is probably more complete.

.github/workflows/sonarcloud.yml
```diff
@@ -117,3 +117,11 @@ jobs:
...
+
+    - name: clean up workspace
+      run: git clean -xdf
```
pbougue: After sonar, you will need a sudo
```suggestion
      run: sudo git clean -xdf
```

azime:
Fixed [Here](https://github.com/hove-io/navitia/pull/4097)

PR #4063, opened 2023-07-03
Kraken: additional attributes for realtime trip creation

pbougue:
:mag: Please review by commit with message (cleaning and preparatory work is separated).

 :warning: Major bump of version (bina required)

Handle new attributes (when filled) from NTFSRT-TripUpdate when creating a realtime VJ (instead of using default values).
Requirement : same behavior than before when values are empty (existing tests should not change).

TODO:
- [x] Tests will be done in separate commits (hopefully not changing the current code :crossed_fingers:)

Supersedes #4058 
JIRA: https://navitia.atlassian.net/browse/NAV-1926

### DIFF
--- a/source/kraken/apply_disruption.cpp
+++ b/source/kraken/apply_disruption.cpp
@@ -424,9 +424,8 @@ struct add_impacts_visitor : public apply_impacts_visitor {
             } else {
                 // Affect the headsign to vj if present in gtfs-rt
                 if (!impact->headsign.empty()) {
-                    vj->headsign = impact->headsign;
                     vj->name = impact->headsign;
-                    pt_data.headsign_handler.change_name_and_register_as_headsign(*vj, impact->headsign);
+                    pt_data.headsign_handler.change_vj_headsign_and_register(*vj, impact->headsign);
                 }
 
                 // for protection, use the datasets[0]
--- a/source/kraken/apply_disruption.cpp
+++ b/source/kraken/apply_disruption.cpp
@@ -200,21 +200,66 @@ type::ValidityPattern compute_base_disrupted_vp(const std::vector<boost::posix_t
 }
 
 nt::Route* get_or_create_route(const nt::disruption::Impact& impact, nt::PT_Data& pt_data) {
-    nt::Network* network = pt_data.get_or_create_network("network:additional_service", "additional service");
-    nt::CommercialMode* comm_mode =
-        pt_data.get_or_create_commercial_mode("commercial_mode:additional_service", "additional service");
+    log4cplus::Logger logger = log4cplus::Logger::getInstance("log");
+    // If ptobject id is present in the impact (line, network, dataset, commercial_mode), it should exist
+    // otherwise disruption is rejected
+    // whereas if ptobject id is absent, create a default object
+    nt::Network* network = nullptr;
+    if (!impact.network_id.empty()) {
+        network = pt_data.get_network(impact.network_id);
+        if (network == nullptr) {
+            LOG4CPLUS_INFO(logger, "Rejecting impact (" << impact.uri << "): network_id (" << impact.network_id
+                                                        << ") provided is absent");
+            return nullptr;
+        }
+    } else {
+        network = pt_data.get_or_create_network("network:additional_service", "additional service");
+    }
+
+    nt::CommercialMode* comm_mode = nullptr;
+    if (!impact.commercial_mode_id.empty()) {
+        comm_mode = pt_data.get_commercial_mode(impact.commercial_mode_id);
+        if (comm_mode == nullptr) {
+            LOG4CPLUS_INFO(logger, "Rejecting impact (" << impact.uri << "): commercial_mode_id ("
+                                                        << impact.commercial_mode_id << ") provided is absent");
+            return nullptr;
+        }
+    } else {
+        comm_mode = pt_data.get_or_create_commercial_mode("commercial_mode:additional_service", "additional service");
+    }
 
     // We get the first and last stop_area to create route and line
     const auto& st_depart = impact.aux_info.stop_times.front();
     const auto& sa_depart = st_depart.stop_time.stop_point->stop_area;
     const auto& st_arrival = impact.aux_info.stop_times.back();
     const auto& sa_arrival = st_arrival.stop_time.stop_point->stop_area;
-    std::string line_uri = "line:" + sa_depart->uri + "_" + sa_arrival->uri;
-    std::string line_name = sa_depart->name + " - " + sa_arrival->name;
-    std::string route_uri = "route:" + sa_depart->uri + "_" + sa_arrival->uri;
-    std::string route_name = sa_depart->name + " - " + sa_arrival->name;
-    nt::Line* line = pt_data.get_or_create_line(line_uri, line_name, network, comm_mode);
-    nt::Route* route = pt_data.get_or_create_route(route_uri, route_name, line, sa_arrival, "forward");
+    nt::Line* line = nullptr;
+    if (!impact.line_id.empty()) {
+        line = pt_data.get_line(impact.line_id);
+        if (line == nullptr) {
+            LOG4CPLUS_INFO(logger, "Rejecting impact (" << impact.uri << "): line_id (" << impact.line_id
+                                                        << ") provided is absent");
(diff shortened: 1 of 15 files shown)


source/type/message.cpp
```diff
@@ -118,7 +118,8 @@ SERIALIZABLE(ApplicationPattern)
 template <class Archive>
 void Impact::serialize(Archive& ar, const unsigned int /*unused*/) {
     ar& uri& company_id& physical_mode_id& headsign& created_at& updated_at& application_periods& severity&
-        _informed_entities& messages& disruption& aux_info& application_patterns;
+        _informed_entities& messages& disruption& aux_info& application_patterns& trip_short_name& dataset_id&
+            network_id& commercial_mode_id& line_id& route_id;
```
pbougue, line 122: Required for data cloning (RT or base-schedule reload) : https://github.com/hove-io/navitia/blame/714f409886302e6b38a195c4ee4e9db66eb0c199/source/type/data.cpp#L874-L883

source/kraken/apply_disruption.cpp
```diff
@@ -200,21 +200,66 @@ type::ValidityPattern compute_base_disrupted_vp(const std::vector<boost::posix_t
...
+
+    std::string route_uri = "";
+    std::string route_name = "";
+    std::string direction_type = "";
+    if (!impact.route_id.empty()) {
+        route_uri = impact.route_id;
+        route_name = "Additional service";
+        direction_type = "outbound";
+    } else {
+        route_uri = "route:" + sa_depart->uri + "_" + sa_arrival->uri;
```
kadhikari, line 258:
```suggestion
        route_uri = "route:" + line->uri
```

source/kraken/apply_disruption.cpp
```diff
@@ -200,21 +200,66 @@ type::ValidityPattern compute_base_disrupted_vp(const std::vector<boost::posix_t
...
+        route_name = sa_depart->name + " - " + sa_arrival->name;
+        direction_type = "forward";
```
kadhikari, line 260:
```suggestion
        direction_type = "outbound";
```

kadhikari: Be sure of name, direction type .. as explain in the doc

source/kraken/apply_disruption.cpp
```diff
@@ -424,21 +472,35 @@ struct add_impacts_visitor : public apply_impacts_visitor {
...
-                    }
+                if (dataset == nullptr && !pt_data.datasets.empty() && pt_data.datasets[0]) {
+                    dataset = pt_data.datasets[0];
+                    LOG4CPLUS_INFO(log, "impact ("
+                                            << impact->uri
+                                            << "): associate random dataset to new VJ because base VJ doesn't exist");
+                }
+                if (dataset != nullptr) {
+                    vj->dataset = dataset;
+                    vj->dataset->vehiclejourney_list.insert(vj);
```
kadhikari, line 502: Minor: Can this block be moved above and be include in 
if (dataset == nullptr && !pt_data.datasets.empty() && pt_data.datasets[0]) {

source/kraken/apply_disruption.cpp
kadhikari: OK for me as i found: 
Requirement : same behavior than before when values are empty (existing tests should not change).

source/kraken/apply_disruption.cpp
kadhikari: OK for me as i found:
Requirement : same behavior than before when values are empty (existing tests should not change).

source/kraken/apply_disruption.cpp
pbougue: Letting it here allows checking that dataset is not `nullptr` after previous block (in case `pt_data.datasets[0]`is null).
In this case, this line would crash.
So really unlikely, but protecting a crash was the goal here.

kadhikari: :+1:

source/kraken/tests/realtime_test.cpp
```diff
@@ -3405,6 +3604,165 @@ BOOST_FIXTURE_TEST_CASE(company_id_doesnt_exist_in_new_trip, AddTripDataset) {
...
+        navitia::type::EntryPoint origin(origin_type, from);
+        navitia::type::EntryPoint destination(destination_type, to);
+
+        navitia::PbCreator pb_creator(b.data.get(), "20190101T073000"_dt, null_time_period);
+        make_response(pb_creator, raptor, origin, destination, {ntest::to_posix_timestamp(datetime)}, true,
+                      navitia::type::AccessibiliteParams(), {}, {}, sn_worker, nt::RTLevel::RealTime, 2_min);
+        return pb_creator.get_response();
+    };
+
+    // If the company id doesn't exist inside the data with ADDED type, we reject the new trip
```
kadhikari:
```suggestion
    // If the line id doesn't exist inside the data with ADDED type, we reject the new trip
```

source/kraken/tests/realtime_test.cpp
```diff
@@ -3405,6 +3604,165 @@ BOOST_FIXTURE_TEST_CASE(company_id_doesnt_exist_in_new_trip, AddTripDataset) {
...
```
kadhikari:
```suggestion
    // If the dataset id doesn't exist inside the data with ADDED type, we reject the new trip
```

source/kraken/tests/realtime_test.cpp
```diff
@@ -3405,6 +3604,165 @@ BOOST_FIXTURE_TEST_CASE(company_id_doesnt_exist_in_new_trip, AddTripDataset) {
...
```
kadhikari:
```suggestion
    // If the network id doesn't exist inside the data with ADDED type, we reject the new trip
```

source/kraken/tests/realtime_test.cpp
```diff
@@ -3405,6 +3604,165 @@ BOOST_FIXTURE_TEST_CASE(company_id_doesnt_exist_in_new_trip, AddTripDataset) {
...
```
kadhikari:
```suggestion
    // If the commercial_mode id doesn't exist inside the data with ADDED type, we reject the new trip
```

kadhikari (approved):

source/type/message.cpp
xlqian: it's probably possible to load the old data without re-bina.... have you tried to load the `data.nav` with this PR?

source/kraken/realtime.cpp
```diff
@@ -320,23 +320,61 @@ static bool is_handleable(const transit_realtime::TripUpdate& trip_update,
...
-        if (is_added_trip(trip_update) && trip_update.vehicle().HasExtension(kirin::physical_mode_id)
-            && pt_data.physical_modes_map.find(trip_update.vehicle().GetExtension(kirin::physical_mode_id))
-                   == pt_data.physical_modes_map.end()) {
-            LOG4CPLUS_DEBUG(log, "Trip physical mode id " << trip_update.vehicle().GetExtension(kirin::physical_mode_id)
-                                                          << " doesn't exist: ignoring trip update id "
-                                                          << trip_update.trip().trip_id());
-            return false;
+        if (is_added_trip(trip_update)) {
+            // check if company id exists
+            if (trip_update.trip().HasExtension(kirin::company_id)
```
xlqian, line 325: minor: it'll be much more readable you can put all these codes in functions... like `check_company_id()` `check_physical_mode_id()`

besides, these blocs of codes look quite similar...

xlqian (approved):

source/kraken/realtime.cpp
pbougue: I'm not willing to dig more on templating.
`kirin::company_id`'s type is `google::protobuf::internal::ExtensionIdentifier< ::transit_realtime::TripDescriptor, ::google::protobuf::internal::StringTypeTraits, 9, false>` :sweat_smile: and 4 others have the same type, but not physical_mode_id.
But I agree on readability for that, some intermediate steps may be done, I won't dig, but it would be legitimate IMO.

source/type/message.cpp
pbougue: After discussion, let's downgrade back to old data_version.

PR #4064, opened 2023-07-04
[Jormungandr] Make olympics sites attractivities configurable

xlqian:
- [x] Need to merge https://github.com/hove-io/navitia-proto/pull/202
- [x] Update submodule reference to `hove-io/navitia-proto#master`


woshilapin (approved):
Do you think it would make sense to ensure that the value is between 0-255 in `jormungandr`? In case it's bigger, should we cap to 255 or just ignore the attractivity? (I would go for the latter).

xlqian:
@woshilapin  if a value bigger than 255 is filled in the parameter, the request will be rejected (400) with a warning

xlqian:
we can merge this PR before the PR in Loki. it's totally harmless

PR #4065, opened 2023-07-05
Add some tests on new trip with additional attributes

kadhikari:


source/jormungandr/tests/kirin_realtime_tests.py
```diff
@@ -2076,6 +2076,299 @@ def test_add_new_trip(self):
...
+        response = self.query_region('lines?')
+        assert len(response['lines']) == 7
+        assert len(response['lines'][0]['routes']) == 2
+        line_id = response['lines'][0]['id']
+        assert line_id == 'A'
+
+        response = self.query_region('commercial_modes?')
+        assert len(response['commercial_modes']) == 8
+        commercial_mode_id = response['commercial_modes'][0]['id']
+        assert commercial_mode_id == '0x0'
```
pbougue:
```suggestion
        assert commercial_mode_id == '0x0'  # Tramway
```

source/jormungandr/tests/kirin_realtime_tests.py
```diff
@@ -2076,6 +2076,299 @@ def test_add_new_trip(self):
...
+                ),
+            ],
+            disruption_id='new_trip',
+            effect='additional_service',
+            physical_mode_id='physical_mode:Bus',  # this physical mode exists in kraken
+            headsign='trip_headsign',
+            trip_short_name='trip_short_name',
+            network_id=network_id,
+            commercial_mode_id=commercial_mode_id,
+            line_id=line_id,
```
pbougue, line 2184: To really test the use-case, you should provide a route_id, from spec it will be:
`route_id="route:{}:additional_service".format(line_id),`
(then some "output" attributes of route are expected to change) .

source/jormungandr/tests/kirin_realtime_tests.py
```diff
@@ -2076,6 +2076,299 @@ def test_add_new_trip(self):
...
+                    departure_delay=0,
+                    is_added=True,
+                    arrival=tstamp("20120614T080102"),
+                    departure=tstamp("20120614T080102"),
+                ),
+            ],
+            disruption_id='new_trip',
+            effect='additional_service',
+            physical_mode_id='physical_mode:Bus',  # this physical mode exists in kraken
+            headsign='trip_headsign',
```
pbougue, line 2179: Can't you add dataset, too ?
`default:dataset` should be the uri of the one in dataset I think (to be tested before disruption).
Then to test that, the only way may be to request `vehicle_journeys/<vj_id>/datasets`.

source/jormungandr/tests/kirin_realtime_tests.py
kadhikari: :+1:

source/jormungandr/tests/kirin_realtime_tests.py
kadhikari: :+1:

source/jormungandr/tests/kirin_realtime_tests.py
kadhikari: :+1:

pbougue (approved):

PR #4071, opened 2023-07-17
[Jormungandr] Manage address within zones in journeys

azime:
Jira: [NAV-2158](https://navitia.atlassian.net/browse/NAV-2158)

From = coord in Poi shape (Jardin du Luxembourg (Paris))
image

From=POI (Jardin du Luxembourg (Paris))
image

[NAV-2158]: https://navitia.atlassian.net/browse/NAV-2158?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/source/jormungandr/jormungandr/utils.py
+++ b/source/jormungandr/jormungandr/utils.py
@@ -494,7 +494,7 @@ def get_pt_object_from_json(dict_pt_object, instance):
     return pt_object
 
 
-def transform_entrypoint(dict_pt_object):
+def transform_entrypoint(dict_pt_object, uri):
     if not dict_pt_object:
         return None
     if MAP_STRING_PTOBJECT_TYPE.get(dict_pt_object.get("embedded_type")) != type_pb2.ADDRESS:
@@ -510,8 +510,11 @@ def transform_entrypoint(dict_pt_object):
     )
     if not first_within_zone:
         return dict_pt_object
-
-    first_within_zone["poi"]["coord"] = dict_pt_object["address"]["coord"]
+    lon, lat = get_lon_lat(uri)
+    if lon is None and lat is None:
+        return dict_pt_object
+    first_within_zone["poi"]["coord"]["lon"] = lon
+    first_within_zone["poi"]["coord"]["lat"] = lat
     return first_within_zone
 
 
--- a/source/jormungandr/jormungandr/utils.py
+++ b/source/jormungandr/jormungandr/utils.py
@@ -510,11 +510,11 @@ def transform_entrypoint(dict_pt_object, uri):
     )
     if not first_within_zone:
         return dict_pt_object
-    lon, lat = get_lon_lat(uri)
-    if lon is None and lat is None:
+    if not is_coord(uri):
         return dict_pt_object
-    first_within_zone["poi"]["coord"]["lon"] = lon
-    first_within_zone["poi"]["coord"]["lat"] = lat
+    lon, lat = get_lon_lat(uri)
+    first_within_zone["poi"]["coord"]["lon"] = "{}".format(lon)
+    first_within_zone["poi"]["coord"]["lat"] = "{}".format(lat)
     return first_within_zone
 
 
--- a/source/jormungandr/jormungandr/utils.py
+++ b/source/jormungandr/jormungandr/utils.py
@@ -494,12 +494,8 @@ def get_pt_object_from_json(dict_pt_object, instance):
     return pt_object
 
 
-def transform_entrypoint(dict_pt_object, uri):
-    if not dict_pt_object:
-        return None
-    if MAP_STRING_PTOBJECT_TYPE.get(dict_pt_object.get("embedded_type")) != type_pb2.ADDRESS:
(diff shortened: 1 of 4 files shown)


source/jormungandr/jormungandr/utils.py
```diff
@@ -494,6 +494,30 @@ def get_pt_object_from_json(dict_pt_object, instance):
     return pt_object
 
 
+def transform_entrypoint(dict_pt_object, uri):
+    if not dict_pt_object:
+        return None
+    if MAP_STRING_PTOBJECT_TYPE.get(dict_pt_object.get("embedded_type")) != type_pb2.ADDRESS:
+        return dict_pt_object
```
woshilapin: I'm wondering if you should have a second sub-function here. The goal of a sub-function is for documentation. Since the sub-function also has a name, it helps to understand a sub-part of the code.
```suggestion
    if MAP_STRING_PTOBJECT_TYPE.get(dict_pt_object.get("embedded_type")) == type_pb2.ADDRESS:
        replace_address_with_custom_poi(dict_pt_object, uri)
    else:
        return dict_pt_object
```

All the code below would go in the new function.

source/jormungandr/jormungandr/utils.py
```diff
@@ -494,6 +494,30 @@ def get_pt_object_from_json(dict_pt_object, instance):
...
+    first_within_zone = next(
```
woshilapin: You could rename this `new_poi` to express better the fact that we're actually creating a new object (based on an existing POI, but with custom coordinates).

source/jormungandr/jormungandr/utils.py
```diff
@@ -494,6 +494,30 @@ def get_pt_object_from_json(dict_pt_object, instance):
...
+        ),
+        None,
+    )
+    if not first_within_zone:
+        return dict_pt_object
+    if not is_coord(uri):
+        return dict_pt_object
+    lon, lat = get_lon_lat(uri)
+    first_within_zone["poi"]["coord"]["lon"] = "{}".format(lon)
+    first_within_zone["poi"]["coord"]["lat"] = "{}".format(lat)
```
woshilapin: I would put a comment here to express why we're changing the coordinate. Here is a suggestion.

```
# We're putting back the coordinate of the end-user, who is in a POI area (with entrypoints).
# If we don't, barycenter of the POI area will be displayed which means nothing for the end user.
```

source/jormungandr/jormungandr/utils.py
```diff
@@ -494,6 +494,30 @@ def get_pt_object_from_json(dict_pt_object, instance):
     return pt_object
 
 
+def transform_entrypoint(dict_pt_object, uri):
```
woshilapin: I can't find a good name for this function... but I'd say we should try to find another name. Here are suggestions for brainstorming, maybe you will find other ideas?

Building the name of the function with this shape.
```<verb>_<name>```

`<verbs>`:
- improve
- customize
- replace
- rectify

`<names>`:
- entrypoint_detail
- address (`replace_address` for example)

other ideas:
- `rectify_entrypoint_with_uri`
- `entrypoint_uri_refocus` (I kind of like this one, we're changing back the entrypoint to refocus on the end-user information: the `from` or the `to`)

None of them are really good, but I'm hoping to help find ideas.

**Review by woshilapin — APPROVED** [review]
As discussed together, the code is fine. All of my comments are only suggestions to try to make the code more maintainable in the long term.

File: source/jormungandr/jormungandr/utils.py (hunk 1 above)
Comment (azime) [reply source/jormungandr/jormungandr/utils.py]: Done

File: source/jormungandr/jormungandr/utils.py (hunk 2 above)
Comment (azime) [reply source/jormungandr/jormungandr/utils.py]: Done

File: source/jormungandr/jormungandr/utils.py (hunk 3 above)
Comment (azime) [reply source/jormungandr/jormungandr/utils.py]: :+1: Done

File: source/jormungandr/jormungandr/utils.py (hunk 4 above)
Comment (azime) [reply source/jormungandr/jormungandr/utils.py]: Done

File: source/jormungandr/jormungandr/utils.py [hunk 5]
```diff
@@ -494,6 +494,36 @@ def get_pt_object_from_json(dict_pt_object, instance):
     return pt_object
 
 
+def replace_address_with_custom_poi(dict_pt_object, uri):
+    new_poi = next(
```
xlqian, line 498: my question may be pedantic:

could the address be included in more than one shape?  if so, how could you be sure that the within_zones are sorted in a deterministic order?

xlqian (approved):

source/jormungandr/jormungandr/utils.py
woshilapin: On `bragi`, the `within_zones` are order by distance to the coordinate. See https://github.com/hove-io/mimirsbrunn/pull/849/files.

woshilapin (approved):

PR #4072, opened 2023-07-19
[Jormungandr] Read default stop points attract from file

xlqian:
https://navitia.atlassian.net/browse/NAV-2159
https://navitia.atlassian.net/browse/NAV-2133


source/jormungandr/jormungandr/utils.py
```diff
@@ -1107,6 +1107,34 @@ def read_origin_destination_data(file_path):
...
+            csv_reader = csv.DictReader(f, fieldnames)
+            # skip the header
+            next(csv_reader)
+
+            for line in csv_reader:
+                stop_point_id = line['stop_point_id']
+                attractivity = line['attractivity']
+                stop_points_attractivities[stop_point_id] = int(attractivity)
+
+        return stop_points_attractivities
```
woshilapin: I believe you can avoid the initialization of `stop_points_attractivities` line 1118 and just do this.

```suggestion
            return { line['stop_point_id']: int(line['attractivity'] for line in csv_reader }
```

You might need an additional `return None` somewhere to ensure we `return` in every branch of this function.

woshilapin (approved):

PR #4073, opened 2023-07-20
Kraken: Fix crash on rail-section without line

pbougue:
* TDD rail-section without line crashes before
* Tests: tinker chaos-disruption mocker
* Tests: fix "encoding" that was breaking in case of non-200
* fix and secure code

:mag: easier to review by commit and without blank-changes :wink: 

JIRA: https://navitia.atlassian.net/browse/NAV-2146

### DIFF
--- a/source/jormungandr/tests/chaos_disruptions_tests.py
+++ b/source/jormungandr/tests/chaos_disruptions_tests.py
@@ -1310,7 +1310,7 @@ def make_mock_chaos_item(
         tag.id = t['id']
         tag.name = t['name']
 
-    if not impacted_obj or not impacted_obj_type:
+    if impacted_obj_type is None or (impacted_obj_type not in ["line_section", "rail_section"] and impacted_obj is None):
         return feed_message.SerializeToString()
 
     # Impacts
@@ -1349,7 +1349,7 @@ def make_mock_chaos_item(
     }
 
     ptobject = impact.informed_entities.add()
-    ptobject.uri = impacted_obj
+    ptobject.uri = impacted_obj or "None"
     ptobject.pt_object_type = type_col.get(impacted_obj_type, chaos_pb2.PtObject.unkown_type)
     if ptobject.pt_object_type in [chaos_pb2.PtObject.line_section, chaos_pb2.PtObject.rail_section]:
         disrupted_section = None
@@ -1362,8 +1362,9 @@ def make_mock_chaos_item(
                 pb_sa.uri = sa
                 pb_sa.order = idx
 
-        disrupted_section.line.uri = impacted_obj
-        disrupted_section.line.pt_object_type = chaos_pb2.PtObject.line
+        if impacted_obj is not None:
+            disrupted_section.line.uri = impacted_obj or "None"
+            disrupted_section.line.pt_object_type = chaos_pb2.PtObject.line
         pb_start = disrupted_section.start_point
         pb_start.uri = start
         pb_start.pt_object_type = chaos_pb2.PtObject.stop_area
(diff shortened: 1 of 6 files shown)


source/type/message.cpp
```diff
@@ -707,20 +710,23 @@ boost::optional<RailSection> try_make_rail_section(
...
+                LOG4CPLUS_WARN(logger, "Rail section has line: '"
+                                            << *line_uri << "' but also a route with no line: " << route->uri);
+                return boost::none;
+            } else if (line == nullptr) {
+                line = route->line;
+            }
+            if (route->line->idx != line->idx) {
+                LOG4CPLUS_WARN(logger, "Rail section has line (or a route from line): '"
+                                            << *line_uri << "' but also a route: '" << route->uri
+                                            << "' that belongs to a different line: " << route->line->uri);
```
woshilapin: Nice to log that something weird is going on. I wonder if we should also inform on what will happen (drop the rail section?)

source/jormungandr/tests/chaos_disruptions_tests.py
```diff
@@ -1349,7 +1349,7 @@ def make_mock_chaos_item(
     }
 
     ptobject = impact.informed_entities.add()
-    ptobject.uri = impacted_obj
+    ptobject.uri = impacted_obj or "None"
```
woshilapin, line 1354: Is it a typo?

```suggestion
    ptobject.uri = impacted_obj or None
```

woshilapin (approved):

source/jormungandr/tests/chaos_disruptions_tests.py
pbougue: It's not, uri is a pbf object, it's typed.
Not especially proud of what I did on that mock (didn't want to refactor it, but it would be nice).

source/type/message.cpp
pbougue: Yep, it's dropped, I changed the log to mention that : https://github.com/hove-io/navitia/pull/4073/commits/1b57194213f70cece8be69cddd43f746ac3f4317

woshilapin (approved):

source/type/message.cpp
```diff
@@ -707,20 +710,23 @@ boost::optional<RailSection> try_make_rail_section(
...
+                return boost::none;
+            } else if (line == nullptr) {
+                line = route->line;
+            }
+            if (route->line->idx != line->idx) {
+                LOG4CPLUS_WARN(logger, "Rejected rail section has line (or a route from line): '"
+                                           << *line_uri << "' but also a route: '" << route->uri
+                                           << "' that belongs to a different line: " << route->line->uri);
+                return boost::none;
             }
```
xlqian:
```suggestion
            if (route->line == nullptr) {
                LOG4CPLUS_WARN(logger, "Rejected rail section has line: '"
                                           << *line_uri << "' but also a route with no line: " << route->uri);
                return boost::none;
            }
            if (line == nullptr) {
                line = route->line;
            }else if (route->line->idx != line->idx) {
                LOG4CPLUS_WARN(logger, "Rejected rail section has line (or a route from line): '"
                                           << *line_uri << "' but also a route: '" << route->uri
                                           << "' that belongs to a different line: " << route->line->uri);
                return boost::none;
            }
```

xlqian (approved):

source/type/message.cpp
pbougue: Done in https://github.com/hove-io/navitia/pull/4073/commits/b808e034ba00a17bdd6afcad1ba559d0bdae81a0

PR #4078, opened 2023-07-26
[Jormungandr] Fix transfer path when no access points is reachable

xlqian:


pbench (approved):

source/jormungandr/jormungandr/scenarios/helper_classes/transfer.py
```diff
@@ -282,11 +282,17 @@ def _do_access_point_transfer(self, section, prev_section_mode, next_section_mod
             origins,
             destinations,
             FallbackModes.walking.name,
-            section.duration * 2,
+            section.duration * 3,
```
woshilapin, line 285: This seems pretty random for a change. Can you explain a bit why we do this change?

xlqian: yes, it's an empirical value.

PR #4080, opened 2023-07-27
[Workflow] Attempt to fix workflow with chaos-proto

xlqian:


pbench (approved):

.github/workflows/workflow.yml
```diff
@@ -93,6 +93,7 @@ jobs:
           LANG: C.UTF-8
         run: |
           git config --global --add safe.directory /__w/navitia/navitia
+          git config --global --add safe.directory /__w/navitia/navitia/source/chaos-proto
```
woshilapin, line 96: Do we need this for any of the other repository we depend on? Is it specific because the repo has been privatize?

woshilapin (approved):

.github/workflows/workflow.yml
xlqian: I'm not sure if we should put it everywhere, that's why it's called an "attempt"... 

> Is it specific because the repo has been privatize?

yes, apparently it's in connection with the fact that runners are re-used by different github actions

PR #4081, opened 2023-07-28
[Jormungandr] Enhance selection of best olympics

xlqian:


source/jormungandr/jormungandr/scenarios/journey_filter.py
```diff
@@ -746,26 +746,37 @@ def apply_final_journey_filters_post_finalize(response_list, request):
...
 
+def compute_journey_virtual_duration(journey, criteria, virtual_fallbacks):
+    extremity_pt_section = get_journey_extremity_pt_section(journey, criteria)
+    extremity = get_journey_pt_extremity(journey, criteria)
+    virtual_fallback = virtual_fallbacks.get(extremity.uri) or 0
+
+    if criteria == "arrival_stop_attractivity":
+        return extremity_pt_section.end_date_time - journey.departure_date_time + virtual_fallback
+
+    return extremity_pt_section.begin_date_time - journey.arrival_date_time + virtual_fallback
```
pbench: If I am not mistaken : 
```suggestion
    return journey.arrival_date_time - extremity_pt_section.begin_date_time  + virtual_fallback
```

xlqian: oops, indeed

source/jormungandr/jormungandr/scenarios/journey_filter.py
```diff
@@ -746,26 +746,37 @@ def apply_final_journey_filters_post_finalize(response_list, request):
...
 
     assert extemity_pt_section
     if criteria == "arrival_stop_attractivity":
         return extemity_pt_section.destination
     return extemity_pt_section.origin
 
 
+def compute_journey_virtual_duration(journey, criteria, virtual_fallbacks):
+    extremity_pt_section = get_journey_extremity_pt_section(journey, criteria)
+    extremity = get_journey_pt_extremity(journey, criteria)
```
woshilapin: We're calling twice `get_journey_extermity_pt_section` in these 2 lines. Can we try to reorganize the code to avoid that? I guess we could make `get_journey_pt_extremity` take an `extremity_pt_section` has an input argument?

source/jormungandr/jormungandr/scenarios/journey_filter.py
```diff
@@ -846,22 +857,21 @@ def filter_olympics_journeys_v1(responses, request):
...
+            virtual_duration = compute_journey_virtual_duration(
+                j, request.get('criteria'), virtual_fallback_durations
+            )
+
+            if virtual_duration < best[1]:
+                best = (j, virtual_duration)
 
-    best[0].tags.append('best_olympics')
+    if best[0] is not None:
+        best[0].tags.append('best_olympics')
```
woshilapin, line 873: Should we log something in an `else` branch. It might help us debug in case we don't see a journey tagged with `best_olympics`?

woshilapin (approved):

source/jormungandr/jormungandr/scenarios/journey_filter.py
xlqian: ah... a crustacean...

woshilapin (approved):

source/jormungandr/jormungandr/scenarios/journey_filter.py
xlqian: good catch... I have tried, but the code rapidly became a bowl of spaghetti, since `get_journey_extermity_pt_section` is also used elsewhere... since you bring it up, I have thought up another idea...

woshilapin (approved):

source/jormungandr/jormungandr/scenarios/journey_filter.py
woshilapin: Ah, I thought about something like this but was worried it would be awful. It's actually quite better than what I was worried about. Good for me :+1:

pbench (approved):

PR #4083, opened 2023-08-08
[Kraken] Add pt journey fare in kraken

xlqian:
In this PR, I proceed to separate the fare computation from Raptor and render it into an independent API, to which Jormungandr will be able to inquire about the fare for a public transport journey, whichever the calculator is.

The core part of this PR is the conversion from a public transport journey to a simplified version which contains all the information needed by the fare computation. 

Note: The fare computation per se remains as it is.

### DIFF
--- a/source/fare/fare.h
+++ b/source/fare/fare.h
@@ -1,4 +1,4 @@
-/* Copyright �� 2001-2022, Hove and/or its affiliates. All rights reserved.
+/* Copyright © 2001-2022, Hove and/or its affiliates. All rights reserved.
 
 This file is part of Navitia,
     the software to build cool stuff with public transport.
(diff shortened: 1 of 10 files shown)


pbench (approved):

source/fare/fare.h
```diff
@@ -275,19 +276,25 @@ std::ostream& operator<<(std::ostream& ss, const Label& l);
 
 /// Contient les données retournées par navitia
```
pbougue, line 277: :see_no_evil: :fr:

pbougue (approved):
Should you "double" all the existing fare tests, translating them to the new API?

Not sure it's easy: should we test it from jorm integration tests too?
Not sure it's worth the effort here as it's probably only the initial request dispatch that's not tested.

PR #4089, opened 2023-08-11
[Jormungandr] Implement pt journey fare

xlqian:


source/jormungandr/jormungandr/pt_journey_fare/pt_journey_fare.py
```diff
@@ -0,0 +1,40 @@
...
+
+# Using abc.ABCMeta in a way it is compatible both with Python 2.7 and Python 3.x
+# http://stackoverflow.com/a/38668373/1614576
+ABC = abc.ABCMeta(str("ABC"), (object,), {})
+
+
+class AbstractPtJourneyFare(ABC):  # type: ignore
+    @abc.abstractmethod
+    def get_pt_journeys_fare(self, pt_journeys, request_id):
+        pass
```
pbougue: Optional:
```suggestion
        raise NotImplementedError("Please implement")
```

source/jormungandr/jormungandr/instance_manager.py
```diff
@@ -121,6 +121,7 @@ def register_instance(self, config):
             self._streetnetwork_backend_manager,
             config.get('external_services_providers', []),
             config.get('pt_planners', {}),
+            config.get('pt_journey_fare', {}),
```
pbougue: Not sure, for consistency, as the surrounding are plural:
```suggestion
            config.get('pt_journey_fares', {}),
```

source/jormungandr/jormungandr/interfaces/v1/Journeys.py
```diff
@@ -683,6 +683,20 @@ def __init__(self):
             help="used to adjust the search range in Asgard when computing matrix",
         )
 
+        parser_get.add_argument(
+            "_pt_journey_fare",
+            type=OptionValue(['kraken']),
+            hidden=True,
+            help="choose which pt engine to compute the pt journey's fare",
```
pbougue: Optional:
```suggestion
            help="choose which PT-fare engine to compute the journey's PT-fares",
```

source/jormungandr/jormungandr/interfaces/v1/Journeys.py
```diff
@@ -683,6 +683,20 @@ def __init__(self):
...
+        )
+
+        parser_get.add_argument(
+            "_compute_pt_journey_fare",
+            type=BooleanType(),
+            hidden=True,
+            help="whether use external engine to compute pt journey fare",
```
pbougue: Is this argument currently used to "force" (re-)computing of fares (even if using `kraken` for journey computing)? 
I'm not sure about what confuses me, but if it's also clearer to you, you may change to: 
```suggestion
            "_must_compute_pt_journey_fare_externally",
            type=BooleanType(),
            hidden=True,
            help="whether to use external engine to compute pt journey fare",
```

source/jormungandr/jormungandr/pt_journey_fare/kraken.py
```diff
@@ -0,0 +1,67 @@
...
+
+class Kraken(AbstractPtJourneyFare):
+    def __init__(self, instance):
+        self.instance = instance
+
+    @staticmethod
+    def _pt_sections(journey):
+        return [s for s in journey.sections if s.type == response_pb2.PUBLIC_TRANSPORT]
+
+    def create_fare_reqeust(self, pt_journeys):
```
pbougue:
```suggestion
    def create_fare_request(self, pt_journeys):
```

source/jormungandr/jormungandr/pt_journey_fare/pt_journey_fare.py
```diff
@@ -0,0 +1,40 @@
...
+
+
+# Using abc.ABCMeta in a way it is compatible both with Python 2.7 and Python 3.x
+# http://stackoverflow.com/a/38668373/1614576
+ABC = abc.ABCMeta(str("ABC"), (object,), {})
+
+
+class AbstractPtJourneyFare(ABC):  # type: ignore
+    @abc.abstractmethod
+    def get_pt_journeys_fare(self, pt_journeys, request_id):
```
pbougue: Not sure about consistency with other apis, but as you get multiple elements:
```suggestion
    def get_pt_journeys_fares(self, pt_journeys, request_id):
```

source/jormungandr/jormungandr/interfaces/v1/Journeys.py
```diff
@@ -683,6 +683,20 @@ def __init__(self):
...
+        parser_get.add_argument(
+            "_pt_journey_fare",
+            type=OptionValue(['kraken']),
+            hidden=True,
+            help="choose which pt engine to compute the pt journey's fare",
+        )
+
+        parser_get.add_argument(
+            "_compute_pt_journey_fare",
+            type=BooleanType(),
```
pbougue, line 697: Should we add a default value? Considering that pt-planner is kraken, I'd say default to `False`?

source/jormungandr/jormungandr/interfaces/v1/Journeys.py
```diff
@@ -683,6 +683,20 @@ def __init__(self):
             help="used to adjust the search range in Asgard when computing matrix",
         )
 
+        parser_get.add_argument(
+            "_pt_journey_fare",
+            type=OptionValue(['kraken']),
```
pbougue, line 688: Same, should we add a default value here?

pbougue (approved):
After discussion, the assumed behavior for this feature is gonna be:
"trigger (or not) a fare-computation, no matter if it was already computed".

Which means that the trigger-fare and pt-journey planner params will have to be maintained/used at the same time.

So it may lead to (**bold** are "weird" but assumed):
* double fares in result (case of **Kraken journey + trigger**)
* single fares in result (case of Loki journey + trigger, or **Kraken + no-trigger**)
* no fares in result (case of Loki journey + no-trigger).

This explains:
* https://github.com/hove-io/navitia/pull/4089/files#diff-4dd2365a9c6c8a1e3ebdbdc1ae539bc28c2fb811aaade8052d76ed284b4e6300R578-R579

PR #4105, opened 2023-09-11
Jormungandr: Remove clear_cache function

azime:


source/jormungandr/jormungandr/instance_manager.py
```diff
@@ -256,15 +235,16 @@ def _find_coverage_by_object_id_in_instances(self, instances, object_id):
...
+            if self._exists_id_in_instance(instance.name, instance.publication_date, object_id):
                 valid_instances.append(instance)
         if not valid_instances:
             raise RegionNotFound(object_id=object_id)
 
         return valid_instances
 
     @cache.memoize(app.config[str('CACHE_CONFIGURATION')].get(str('TIMEOUT_PTOBJECTS'), None))
-    def _exists_id_in_instance(self, instance, object_id):
+    def _exists_id_in_instance(self, instance_name, instance_publication_date, object_id):
```
pbougue, line 246: Please add a comment on why the params are like that, it's confusing for an outsider :exploding_head: 
(not the time for it, but I can't get out of my head that something more idiomatic must be doable)

source/jormungandr/jormungandr/instance_manager.py
```diff
@@ -256,15 +235,16 @@ def _find_coverage_by_object_id_in_instances(self, instances, object_id):
...
+        instance = self.instances[instance_name]
         return instance.has_id(object_id)
```
pbougue, line 254: We're pretty safe given how it's called currently, but I think we should protect from the None case.

source/jormungandr/jormungandr/instance_manager.py
azime: Done

source/jormungandr/jormungandr/instance_manager.py
```diff
@@ -256,15 +235,19 @@ def _find_coverage_by_object_id_in_instances(self, instances, object_id):
...
         if not valid_instances:
             raise RegionNotFound(object_id=object_id)
 
         return valid_instances
 
     @cache.memoize(app.config[str('CACHE_CONFIGURATION')].get(str('TIMEOUT_PTOBJECTS'), None))
-    def _exists_id_in_instance(self, instance, object_id):
+    def _exists_id_in_instance(self, instance_name, instance_publication_date, object_id):
+        """
+        published_date is used to invalidate the cache when updating the ntfs, poi
```
pbougue: You've got an extra `, poi` in you comment anyway, I think.

Also I thought of a "clearer" naming idea:
```suggestion
    def _exists_id_in_instance(self, instance_name, extra_cache_key, object_id):
        """
        instance's published_date is usually provided as extra_cache_key to invalidate the cache when updating the ntfs
```

pbougue (approved):
Very elegant solution :heart_eyes: 
:+1: This way :
* we stop cache invalidation for all coverages, and do it one by one.
* we keep the "old" cache which avoids rebounds between old and new cache when krakens for a given coverages are not all updated.

The 2 down-points (:-1:) left I see are (not mandatory to do or decide now IMO):
* still an important coverage cache creation at the time of data-update (but the only way I see is to just let the TTL do the job)
* dependant on other "real kraken requests" to update the `instance.publication_date` in jormun: it would be better to have a mutualized ping for that (but redis-cache is required to avoid flooding of krakens by jormuns).

PR #4109, opened 2023-09-19
[jormungandr] delete origin destination allowed ids management

kadhikari:
This modification will revert most of the code from PR:
-  https://github.com/hove-io/navitia/pull/4023
- https://github.com/hove-io/navitia/pull/4051

For more details: https://navitia.atlassian.net/browse/NAV-2227


xlqian (approved):

azime (approved):

pbougue (approved):
OK.
Although I'm not a fan of letting unused code `additional_parameters` (as discussed indeed, it would be done in multiple deployments to remove it, as the db must be unused before the column is removed).

PR #4111, opened 2023-09-21
CI - adding cleanup step in merge workflows

prhod:


.github/workflows/publish_navitia_documentation.yml
```diff
@@ -40,3 +42,9 @@ jobs:
       if: failure()
       run: |
           echo '{"text":":warning: Navitia Github Actions: publish_navitia_documentation failed ! (https://github.com/hove-io/navitia/actions?query=workflow%3A%22Publish+Navitia+Documentation%22)"}' | http --json POST ${{secrets.SLACK_NAVITIA_TEAM_URL}}
+
+    - name: clean up workspace
+      if: ${{ always() }}
+      run: |
+        rm -rf ./*
+        rm -rf ./.??*
```
woshilapin, line 50: Just curious, why do we need this second line?

woshilapin (approved):

.github/workflows/publish_navitia_documentation.yml
prhod: `rm -rf ./*` does not remove dot files. The second line is to force the removal of `.git` (and other existing dot things)

woshilapin: OK, but why the `?` (doubled). I thought `./.*` would have been enough :thinking: Is it to avoid matching `.` (current directory) and `..` (parent directory)?

PR #4115, opened 2023-09-27
[jormungandr] Add log in transient socket exception

kadhikari:
* Some key words added in log.
* Will help to find exception logs on sockets.
* As discussed with @pbougue and @xlqian it is quite risky to modify the function used every where in jormungandr to call kraken.

### DIFF
--- a/source/jormungandr/jormungandr/transient_socket.py
+++ b/source/jormungandr/jormungandr/transient_socket.py
@@ -178,4 +178,8 @@ class TransientSocket(object):
             socket.setsockopt(zmq.LINGER, 0)
             socket.close()
         except:
-            self._logger.exception("Error while closing transient socket")
+            self._logger.exception(
+                'Error while closing transient socket with coverage: %s, zmq_socket: %s',
+                self.name,
+                self._zmq_socket,
+            )
--- a/source/jormungandr/jormungandr/transient_socket.py
+++ b/source/jormungandr/jormungandr/transient_socket.py
@@ -157,7 +157,7 @@ class TransientSocket(object):
             raise e
         except:
             self.close_socket(timed_socket.socket)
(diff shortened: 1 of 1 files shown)


source/jormungandr/jormungandr/transient_socket.py
```diff
@@ -157,7 +157,7 @@ def call(self, content, timeout, debug_cb=lambda: "", quiet=False):
             raise e
         except:
             self.close_socket(timed_socket.socket)
-            self._logger.exception("")
+            self._logger.exception("Unexpected transient socket exception")
```
pbench: maybe also log `self.name`, `self._zmq_socket` and `debug_cb()` ?

kadhikari: As discussed, all the information are logged and these words added will help to find those logged messages.

xlqian: @pbench's comment is pertinent.

```suggestion
            self._logger.exception("Unexpected transient socket exception with coverage: %s,  zmq_socket: %s, debug_info: %s", self.name, self._zmq_socket, debug_cb())
```

xlqian (approved):

source/jormungandr/jormungandr/transient_socket.py
pbougue: OK, it may help indeed, didn't think of that.
This is also worth doing for the other change I think.

pbougue: ~As I think more: it may help even more around DeadSocketExceptions, no?
(I can try to have a look at how to have this, if you agree)~
:hand_over_mouth: Its already there

pbougue (approved):

azime (approved):

PR #4116, opened 2023-09-27
[docker] jormungandr-processes param removed

azime:
Follow-up on https://github.com/hove-io/navitia/pull/4112

I leave this variable among the environment variables only to avoid params overlap

### DIFF
--- a/docker/run_jormungandr.sh
+++ b/docker/run_jormungandr.sh
@@ -7,7 +7,7 @@ monitor_cache2="name=monitor,items=100"
 
 function show_help() {
     cat << EOF
-Usage: ${0##*/} -m monitor-processes -r max-requests
+Usage: ${0##*/} -m monitor-process -r max-requests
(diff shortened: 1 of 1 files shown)


docker/run_jormungandr.sh
```diff
@@ -7,17 +7,14 @@ monitor_cache2="name=monitor,items=100"
 
 function show_help() {
     cat << EOF
-Usage: ${0##*/} -a jormungandr-processes -m monitor-processes -r max-requests
-    -a      app-processes number
+Usage: ${0##*/} -m monitor-processes -r max-requests
```
pbougue: very minor:
```suggestion
Usage: ${0##*/} -m monitor-process -r max-requests
```

pbougue (approved):
Why do you remove this ?
Is it because it duplicates another param and can be misleading ?
> OK, edited main comment :heavy_check_mark:

xlqian (approved):

PR #4126, opened 2023-10-04
[jormungandr] fix a breaking python2.7 version during import

vpassama:
Hi,

IDK if python2.7 is still supported in jormungandr but I found a breaking compatibility issue.
It seems a module import was made on a python3+ lib version (urllib), I didn't test my change but it works in our env.

Thank you.


vgeof-hove:
Thanks for your PR. Python2.7 support has indeed been removed some months ago.

PR #4127, opened 2023-10-04
[jormungandr] NewRelic added for stat_manager

azime:
Jira: [NAV-2346](https://navitia.atlassian.net/browse/NAV-2346)

[NAV-2346]: https://navitia.atlassian.net/browse/NAV-2346?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/source/jormungandr/jormungandr/new_relic.py
+++ b/source/jormungandr/jormungandr/new_relic.py
@@ -213,11 +213,12 @@ def statManagerEvent(call_name, group_name):
             except Exception as e:
                 event_params["status"] = "failed"
                 event_params.update({"reason": str(e)})
-            duration = timeit.default_timer() - start_time
-            event_params.update({"duration": duration})
+            finally:
+                duration = timeit.default_timer() - start_time
+                event_params.update({"duration": duration})
 
-            # Send the custom event to newrelic !
-            record_custom_event("stat_manager", event_params)
+                # Send the custom event to newrelic !
+                record_custom_event("stat_manager", event_params)
 
             return result
 
--- a/source/jormungandr/jormungandr/new_relic.py
+++ b/source/jormungandr/jormungandr/new_relic.py
@@ -121,6 +121,15 @@ def background_task(name, group):  # type: (Text, Text) -> Callable
     return wrap
 
 
+def get_common_event_params(service_name, call_name, status="ok"):
+    return {
+        "service": service_name,
+        "call": call_name,
+        "status": status,
+        "az": app.config.get("DEPLOYMENT_AZ", "unknown"),
+    }
+
+
 def distributedEvent(call_name, group_name):
     """
     Custom event that we publish to New Relic for distributed scenario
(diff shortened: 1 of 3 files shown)


source/jormungandr/jormungandr/new_relic.py
```diff
@@ -188,3 +188,39 @@ def record_streetnetwork_call(call_name, connector_name, mode, coverage_name):
...
+            except Exception as e:
+                event_params["status"] = "failed"
+                event_params.update({"reason": str(e)})
+            duration = timeit.default_timer() - start_time
+            event_params.update({"duration": duration})
+
+            # Send the custom event to newrelic !
+            record_custom_event("stat_manager", event_params)
+
+            return result
```
pbougue: 2 remarks :
1. Maybe you should also re-raise in the except, not sure.
2. More idiomatic:
```suggestion
                return func(obj, *args, **kwargs)
            except Exception as e:
                event_params["status"] = "failed"
                event_params.update({"reason": str(e)})
            finally:
                duration = timeit.default_timer() - start_time
                event_params.update({"duration": duration})

                # Send the custom event to newrelic !
                record_custom_event("stat_manager", event_params)
```

source/jormungandr/jormungandr/stat_manager.py
```diff
@@ -50,6 +50,10 @@
 f_datetime = "%Y%m%dT%H%M%S"
 
 
+class StatManagerError(RuntimeError):
+    pass
```
pbougue: This should be moved in exceptions.py, no?

source/jormungandr/jormungandr/stat_manager.py
```diff
@@ -556,7 +564,11 @@ def __call__(self, f):
         def wrapper(*args, **kwargs):
             start_time = time.time()
             call_result = f(*args, **kwargs)
-            self.manager.manage_stat(start_time, call_result)
+            try:
+                self.manager.manage_stat(self, start_time, call_result)
+            except (StatManagerError, Exception):
+                # if stat are not working we don't want jormungandr to stop.
+                pass
```
pbougue: StatManagerError is an Exception, so you may just:
```suggestion
            except Exception:
                # if stat are not working we don't want jormungandr to stop.
                pass
```

source/jormungandr/jormungandr/stat_manager.py
azime: Done

source/jormungandr/jormungandr/stat_manager.py
azime: Done

source/jormungandr/jormungandr/new_relic.py
azime: Done

pbougue (approved):

PR #4128, opened 2023-10-04
[kraken] Fix: Recursive display of line and routes

kadhikari:
Before correction:
* Lines are displayed with routes recursively depending on depth: for depth=3 `line.routes[].line.routes[]`
* Routes are displayed with line recursively depending on depth: for depth=3 `routes[].line.routes[].line`
* with depth = 0 line doesn't have Network.

After correction
* Lines are displayed with routes not depending on depth: for depth=1, 2 and 3 `line.routes[]`
* Routes are displayed with line not depending on depth: for depth=1, 2 and 3 `routes[].line`
* Line will always have Network even with depth = 0.

For more details: 
* https://navitia.atlassian.net/browse/NAV-2348
* https://navitia.atlassian.net/browse/NAV-2348?focusedCommentId=54235

Comparison of response time for some request on my PC before and after the correction:
* url_navitia/lines/rerc/lines?
-> depth=1 : 12.5 ms -> 13.0 ms   / 57.5 ko -> 57.5 ko 
-> depth=2 : 13.6 ms -> 12.7 ms   / 80.4 ko -> 57.6 ko
-> depth=3 : 70 ms -> 12.8 ms      / 1.5 Mo -> 57.6 ko 

* url_navitia/lines/rerc/routes?
-> depth=1 : 15 ms -> 14.6 ms   / 69.6 ko -> 75.8 ko (Network added in route.line) 
-> depth=2 : 76.5 ms -> 18 ms   / 1.4 Mo -> 94.8 ko 
-> depth=3 : 393 ms -> 340 ms  / 5.9 Mo -> 4.1 Mo (No change is made on route.stop_point and its sub-objects)

### DIFF
--- a/source/type/pb_converter.cpp
+++ b/source/type/pb_converter.cpp
@@ -793,14 +793,14 @@ void PbCreator::Filler::fill_pb_object(const nt::Line* l, pbnavitia::Line* line)
     }
     fill(l->physical_mode_list, line->mutable_physical_modes());
     fill(l->commercial_mode, line);
+    fill(l->network, line);
 
     if (depth > 0) {
         if (!this->pb_creator.disable_geojson) {
             fill(&l->shape, line);
         }
 
-        fill(l->route_list, line->mutable_routes());
-        fill(l->network, line);
+        copy(0, dump_message_options).fill(l->route_list, line->mutable_routes());
 
         fill(l->line_group_list, line->mutable_line_groups());
     }
@@ -871,7 +871,10 @@ void PbCreator::Filler::fill_pb_object(const nt::Route* r, pbnavitia::Route* rou
         return;
     }
 
-    fill(r->line, route);
+    copy(0, dump_message_options).fill(r->line, route);
+    if ((depth > 0) && (!this->pb_creator.disable_geojson)) {
+        fill(&r->line->shape, route->mutable_line());
+    }
 
     if (!this->pb_creator.disable_geojson) {
         fill(&r->shape, route);
--- a/source/type/pb_converter.cpp
+++ b/source/type/pb_converter.cpp
@@ -872,7 +872,7 @@ void PbCreator::Filler::fill_pb_object(const nt::Route* r, pbnavitia::Route* rou
     }
 
     copy(0, dump_message_options).fill(r->line, route);
-    if ((depth > 0) && (!this->pb_creator.disable_geojson)) {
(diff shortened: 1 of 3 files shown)


source/type/pb_converter.cpp
```diff
@@ -871,7 +871,10 @@ void PbCreator::Filler::fill_pb_object(const nt::Route* r, pbnavitia::Route* rou
         return;
     }
 
-    fill(r->line, route);
+    copy(0, dump_message_options).fill(r->line, route);
+    if ((depth > 0) && (!this->pb_creator.disable_geojson)) {
```
pbougue: Do we still write the shape in the sub-object here?
My take is we shouldn't (and tests will change).
If we want to keep the shape, then, it must be with a `depth > 1`.

source/type/pb_converter.cpp
```diff
@@ -871,7 +871,10 @@ void PbCreator::Filler::fill_pb_object(const nt::Route* r, pbnavitia::Route* rou
...
+        fill(&r->line->shape, route->mutable_line());
```
pbougue, line 876: :heavy_check_mark: In case someone wants to change that in the **future** it's possible (but not done in this PR).


The code here works because shape doesn't have sub-object, but it would be more consistent to do:
```suggestion
        copy(0, dump_message_options).fill(&r->line->shape, route->mutable_line());
```

source/type/pb_converter.cpp
```diff
@@ -793,14 +793,14 @@ void PbCreator::Filler::fill_pb_object(const nt::Line* l, pbnavitia::Line* line)
     }
     fill(l->physical_mode_list, line->mutable_physical_modes());
     fill(l->commercial_mode, line);
+    fill(l->network, line);
```
pbougue, line 796: :heavy_check_mark: OK as it doesn't have sub-object, it won't loop forever :+1:

source/type/pb_converter.cpp
kadhikari: :+1:

source/type/pb_converter.cpp
kadhikari: :+1:

source/type/pb_converter.cpp
kadhikari: :+1:

SGrenet (approved):
Ok, perf test are useful

pbougue (approved):

PR #4134, opened 2023-10-12
Fix: Cache coords object instead of type_pb2.GeographicalCoord()

kadhikari:
Coordinate object of type type_pb2.GeographicalCoord() cannot be cached. 
This object should be transformed to jormun object utils.Coords(lat, lon)

Error introduced in https://github.com/hove-io/navitia/pull/4119

Ticket: https://navitia.atlassian.net/browse/NAV-2397

### DIFF
--- a/source/jormungandr/jormungandr/instance_manager.py
+++ b/source/jormungandr/jormungandr/instance_manager.py
@@ -42,6 +42,7 @@ from jormungandr.exceptions import ApiNotFound, RegionNotFound, DeadSocketExcept
 from jormungandr.authentication import abort_request, can_read_user
 from jormungandr import authentication, cache, memory_cache, app
 from jormungandr.instance import Instance
+from jormungandr.utils import Coords
 import gevent
 import os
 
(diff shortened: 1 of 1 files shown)


woshilapin (approved):

source/jormungandr/jormungandr/instance_manager.py
```diff
@@ -258,24 +259,27 @@ def _get_first_object_coord_in_instances_by_id(self, instances, object_id):
...
                 return coord
-        return coord
+        return None
 
     @memory_cache.memoize(app.config[str('MEMORY_CACHE_CONFIGURATION')].get(str('TIMEOUT_PTOBJECTS'), 30))
     @cache.memoize(app.config[str('CACHE_CONFIGURATION')].get(str('TIMEOUT_PTOBJECTS'), 300))
     def _get_object_coord_in_instance_by_id(self, instance_name, instance_publication_date, object_id):
         """
         instance's published_date is usually provided as extra_cache_key to invalidate the cache when updating the ntfs
+        As type_pb2.GeographicalCoord() cannot be cached, we should transform it to utils.Coord
```
xlqian:
```suggestion
        As type_pb2.GeographicalCoord() cannot be cached, due to the fact that objects from protobuf are not 'picklable', we should transform it to utils.Coord
```

xlqian (approved):

source/jormungandr/jormungandr/instance_manager.py
kadhikari: :+1:

azime (approved):

PR #4137, opened 2023-10-16
[tyr] Loki: enrich ntfs with addresses/admins

datanel:
https://navitia.atlassian.net/browse/NAV-2217?page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-53940

Just before putting the NTFS to S3 for Loki, it is enriched with addresses and administrative regions calling bragi.
For that an intermediate folder `ntfs_with_addreses` is created

### DIFF
--- a/source/tyr/tyr/binarisation.py
+++ b/source/tyr/tyr/binarisation.py
@@ -1139,12 +1139,18 @@ def enrich_ntfs_with_addresses(filename, job_id, dataset_uid):
     logger = get_instance_logger(instance, task_id=job_id)
     filename = zip_if_needed(filename)
 
+    file_dir = os.path.dirname(filename)
+    file_basename = os.path.basename(filename)
+    output_dir = file_dir + "/enriched_with_addresses"
+    os.makedirs(output_dir, 0o755)
+    output = output_dir + "/" + file_basename
+
     try:
         params = [
             "--input",
             filename,
             "--output",
-            filename,
+            output,
             "--bragi-url",
             current_app.config['BRAGI_URL'],
         ]
@@ -1163,7 +1169,7 @@ def enrich_ntfs_with_addresses(filename, job_id, dataset_uid):
     finally:
         models.db.session.commit()
 
-    return filename
+    return output
 
 
 @celery.task(bind=True)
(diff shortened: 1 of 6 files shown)


azime (approved):

source/tyr/tyr/binarisation.py
```diff
@@ -1125,9 +1125,47 @@ def poi2mimir(self, instance_name, input, autocomplete_version, job_id=None, dat
...
+
+    logger = get_instance_logger(instance, task_id=job_id)
+    filename = zip_if_needed(filename)
+
+    try:
+        params = [
+            "--input",
+            filename,
+            "--output",
+            filename,
```
pbench: I am not sure if this is a good idea to replace the input ntfs with the enriched one.
Maybe write to a different location ?

datanel: it works and it should not be a problem imho.
if there is a problem, filename should not be atered by the output.
if the job succeed, the ntfs is just replaced.

datanel: ok I see, the ntfs is sent to fusio2ed after, we should keep the original; ok

pbench: Yeah I don't know exactly what happen with the ntfs (nor when), but I know that *some* things can happen (sent to fusio2ed, moved to a backup directory, ... ?). 
So better be safe than sorry :)

datanel: fixed in the last commit.
How can I increase the code coverage ? Sonar is warning

source/tyr/tyr/binarisation.py
```diff
@@ -1125,9 +1125,53 @@ def poi2mimir(self, instance_name, input, autocomplete_version, job_id=None, dat
...
+            "--bragi-url",
+            current_app.config['BRAGI_URL'],
+        ]
+
+        res = None
+        with collect_metric("enrich-ntfs-with-addresses", job, dataset_uid):
+            res = launch_exec("enrich-ntfs-with-addresses", params, logger)
+        if res != 0:
+            raise ValueError("enrich-ntfs-with-addresses failed")
+        dataset.state = "done"
```
pbench: Not sure when we should set the `dataset.state` to `done` ?

pbench (approved):

PR #4160, opened 2023-11-07
Replace make_response with jsonify

codelion:
Since the intention here is to output JSON it will be better to use jsonify instead of make_reponse.


xlqian:
Hi, thank you for the contribution. SInce the credentials are accessible solely from `hove-io/navitia`, tests/checks can't run successfully over forked repo, can I open a new PR by cherry-picking your commits? 

cheers

codelion:
Yes, please go ahead @xlqian

PR #4177, opened 2023-11-30
activate aggressive memory decommit

xlqian:
It seems a bug has been declared in tcmalloc that it takes ages to drop big object, here is an attempt of fixing 
https://github.com/StarRocks/starrocks/issues/183


pbench (approved):
I guess this needs to be tested. I don't really understand what this setting do (and I didn't find any documentation besides the issue you linked)

PR #4182, opened 2023-12-05
upgrade packages for python 3.11

datanel:
minimum upgrade for running jormungandr on python 3.11.

All the tests don't pass
I've deactivated mypy


source/jormungandr/requirements.txt
```diff
@@ -1,26 +1,26 @@
 configparser==4.0.2 ; python_version == "2.7"
 setuptools==44.1.0 ; python_version == "2.7"
-Flask==1.1.1
-flask_restful==0.3.7
+Flask==2.0.0
```
xlqian: risky risky.... :(

datanel: no choice

PR #4212, opened 2024-02-07
[Tyr Worker] Process poi and push to asgard s3

xlqian:


source/tyr/tyr/binarisation.py
```diff
@@ -1233,6 +1236,46 @@ def gtfs2s3(self, instance_config, filename, job_id, dataset_uid):
...
+    poi_to_excluded_zones(filename, excluded_zone_dir, instance.name)
+
+    try:
+        with collect_metric("poi2Asgard", job, dataset_uid):
+            asgard_bucket = current_app.config.get('MINIO_ASGARD_BUCKET_NAME', None)
+            if not asgard_bucket:
+                raise Exception("Asgard Bucket is None")
+
+            bash_command = (
+                "env REQUESTS_CA_BUNDLE=/etc/ssl/certs/ca-certificates.crt "
```
xlqian, line 1261: even though it's corrected here: https://github.com/hove-io/navitia/pull/4212/files#diff-9677996dffe09856de00ff00f281b66e799a3187047cbb83bf62efe46d434186R30,  I still have to put the env var here and I have absolutely no idea why... otherwise the `aws sync` wouldn't work...

source/tyr/tyr/poi_to_excluded_zones.py
```diff
@@ -0,0 +1,68 @@
...
+        if not geometry_id:
+            logger.error("{} could not be found in poi.txt".format(row[0]))
+        shape = geometries_shapes.get(geometry_id)
+        if not shape:
+            logger.error("{} could not be found in geometries.txt".format(geometry_id))
+
+        for i, zone in enumerate(zones):
+            output_id = "{}_{}_{}".format(poi_id, i, instance_name)
+            output = {'id': output_id}
+            output.update(zone)
```
pbench, line 56: you can have a poi with several zones but all with the same shape ? :thinking:

xlqian: yeahhhhh, I admit that this was not well thought through... 

the need is for a given shape, we wish it could be excluded based on a list of modes and a list of ranges of date, so the ideal format in my opinion is

```json
{
  "name": "poi:toto",
  "shape": "POLYGON ((...))",
  "excluded_on": [
    {
      "modes": [
        "walking",
        "bike"
      ],
      "periods": [
        {
          "from": "20240708",
          "to": "20240709"
        },
        {
          "from": "202407010",
          "to": "20240711"
        }
      ]
    },
    {
      "modes": [
        "car"
      ],
      "periods": [
        {
          "from": "202407010",
          "to": "202407011"
        }
      ]
    }
  ]
}
```

pbench: ah ok ! you have the same shape but for different modes, or different activation periods, did I get it right ?

pbench (approved):

azime (approved):

PR #4213, opened 2024-02-08
Adding cache for transfer_path

azime:
Jira: [NAV-2719](https://navitia.atlassian.net/browse/NAV-2719)

[NAV-2719]: https://navitia.atlassian.net/browse/NAV-2719?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/source/jormungandr/jormungandr/scenarios/helper_classes/transfer.py
+++ b/source/jormungandr/jormungandr/scenarios/helper_classes/transfer.py
@@ -308,24 +308,29 @@ class TransferPool(object):
             )
         return None, None, None
 
+    def add_attribute(self, pp_object, json_data, attributes):
+        for attribute in attributes:
+            if attribute in json_data:
+                setattr(pp_object, attribute, json_data.get(attribute))
+
     def pb_object_from_json(self, json_object):
         pb_object = type_pb2.PtObject()
         str_embedded_type = json_object.get("embedded_type")
         pb_embedded_type = MAP_STRING_PTOBJECT_TYPE.get(str_embedded_type)
         str_embedded_type = str_embedded_type.lower()
         pb_object.embedded_type = pb_embedded_type
-        for attribute in ["uri", "name"]:
-            setattr(pb_object, attribute, json_object.get(attribute))
+        self.add_attribute(pb_object, json_object, ["uri", "name"])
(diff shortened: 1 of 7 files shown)


pbench (approved):

pbench (approved):

source/jormungandr/jormungandr/scenarios/helper_classes/transfer.py
```diff
@@ -167,9 +192,16 @@ def _do_no_access_point_transfer(self, section):
     def _aysnc_no_access_point_transfer(self, section):
         return self._future_manager.create_future(self._do_no_access_point_transfer, section)
 
-    def _get_access_points(self, stop_point_uri, access_point_filter=lambda x: x):
-        sub_request_id = "{}_transfer_start_{}".format(self._request_id, stop_point_uri)
-        stop_points = self._instance.georef.get_stop_points_from_uri(stop_point_uri, sub_request_id, depth=2)
+    def _get_access_points(self, pt_object, access_point_filter=lambda x: x):
```
xlqian: there should be a more proper way to do it, you can create an abstract function in PtPlanner that returns access points, and the implementation should be "hidden" in the concrete class base on the type of PtPlanner,

source/jormungandr/jormungandr/scenarios/helper_classes/transfer.py
```diff
@@ -255,36 +288,48 @@ def _get_transfer_result(self, section, origin, destination):
             sub_request_id,
         )
         if direct_path and direct_path.journeys:
-            return TransferResult(direct_path, origin, destination)
-        return None
-
-    def _do_access_point_transfer(self, section, prev_section_mode, next_section_mode):
-        access_points = self.get_underlying_access_points(section, prev_section_mode, next_section_mode)
+            return (
```
xlqian, line 281: We just need to be cautious/aware of the huge amount of string that could be generated by this thus impacting the performance cache/redis, (it's not particularly worrying since Asgard has been doing so...) On top of that, serialize/deserialize could be time-consuming as well

xlqian (approved):

PR #4214, opened 2024-02-09
[Tyr-web]: fix uwsgi version

azime:
Fix [CI](https://github.com/hove-io/navitia/actions/runs/7835256179/job/21398073685)


pascal-vs (approved):
that means we will stick to an obsolete version, let's think about it later

pbench (approved):

PR #4235, opened 2024-03-07
[FEAT]: street_network backend configuration by user

kadhikari:
Street network backends are stored in the table streetnetwork_backend.
They are used in the following attributes of the table instance:
- street_network_bss
- street_network_ridesharing
- street_network_taxi
- street_network_car_no_park
- street_network_car
- street_network_walking
- street_network_bike
So a street_network backend is configured for an instance (coverage) + mode.

This feature will allow to configure a street_network backend for a user + mode.
- To have minimum time delay while fetching street_network backend for a user + mode, an attribute has_sn_backend is used.
- The new table sn_backend_authorization is excluded from automatic join while getting user for each request to navitia.
- Attribute user.has_sn_backend is manipulated by /post and /delete on api /sn_backend_authorizations
- In the api /users, sn_backend_authorizations will be present.

For more details: https://navitia.atlassian.net/browse/NAV-2800


source/tyr/tests/integration/users_test.py
```diff
@@ -141,8 +142,10 @@ def teardown():
         with app.app_context():
             end_point = models.EndPoint.query.get(d['end_point'])
             billing_plan = models.BillingPlan.query.get(d['billing_plan'])
+            # sn_backend_authorization = models.SnBackendAuthorization.query.get(d['sn_backend_authorization'])
```
xlqian: still need it?

xlqian (approved):

pbench (approved):

PR #4238, opened 2024-03-12
[kraken]: Load disruptions for chaos db

kadhikari:
This fix will only solve errors occurred while loading disruptions from chaos database.

- chaos-proto object are created from the chaos database using a SQL query.
- While loading ptobject of type line_section and rail_section, the same parent ptobject was re used fetching already created ptobject by ptobject_uri.
- All the children ptobject except the first one were neglected.
- Use of ptobject_id instead of ptobject_uri should work for all ptobject types.

Ticket:  https://navitia.atlassian.net/browse/NAV-2821

### DIFF
--- a/source/kraken/fill_disruption_from_database.h
+++ b/source/kraken/fill_disruption_from_database.h
@@ -40,6 +40,7 @@ www.navitia.io
 #include <string>
 #include <memory>
 #include <utility>
+#include <map>
 
 namespace navitia {
 #define FILL_NULLABLE_(var_name, arg_name, col_name, type_name) \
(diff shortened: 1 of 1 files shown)


azime (approved):

pbench (commented):
there is no need to order by pt_object id in the sql request ?

kadhikari:
> there is no need to order by pt_object id in the sql request ?
No need as it may break message, channel and channel types cycles.
We fetch well the concerned ptobject from de map

pbench (approved):
:+1:

PR #4247, opened 2024-03-20
[CI] Fix docker builders: download release asset from cosmogony2cities

xlqian:


.github/workflows/dockers_builder_dev.yml
```diff
@@ -1,6 +1,7 @@
 name: Build Navitia Dockers (Dev)
 
 on:
+  pull_request:
```
xlqian: for testing

azime (approved):

PR #4251, opened 2024-03-26
[Distributed] Timeout when direct path takes too long

xlqian:
In Distributed, the direct path is the very first step performed by the algorithm and the process will NOT carry on until this step finishes. In cases where a VERY long direct path is required, serval hundreds of ms may elapse before Distributed could keep on. 

The idea here is not to wait too long in the sections where the direct path is dispensable.  

The chosen default timeout(100ms) is wholly empirical.


source/jormungandr/jormungandr/scenarios/helper_classes/streetnetwork_path.py
```diff
@@ -223,15 +224,27 @@ def _async_request(self):
...
+        # timeout=None -> wait forever...
+        timer = gevent.timeout.Timeout(timeout, exception=False)
+
+        best_res = None
+        with timer:
+            best_res = min(
+                (
+                    future.wait_and_get()
+                    for future in self._futures
+                    if future.wait_and_get().response is not None
```
pbench, line 238: If I understand correctly, the timeout will be for the completion of all futures in self._futures, and not on each individual future. 
Is this what you want to do ?

xlqian: yes! I cannot determine the best journey until all requests are completed

pbench: Ok, I understand that you need the completion of all futures to determine the best journey.
But is it a good idea to put the timer on the completion of all futures, and not on each individual future ?
If you have a lot of futures, maybe 100ms will not be enough to compute all the futures ?

pbench (approved):

azime (approved):

PR #4273, opened 2024-05-24
[Jormun]: Do not fill co2 as will be filled in distributed

kadhikari:
In the response information on co2 is updated if absent as followings (feature developed after the connector here):

- in each section of type CAR / CAR_NO_PARC
- in context. car_direct_path

Do not fill any information on co2 in the connector Here as they will be filled by Jormun

Ticket: https://navitia.atlassian.net/browse/NAV-2795


wassimbenaissa:
All good!

pbench (approved):

PR #4341, opened 2025-01-15
[jormun]:Intiaize _access_points = false for isochrone

kadhikari:
Ticket: https://navitia.atlassian.net/browse/NAV-3634


adrien3d (approved):

xlqian:
can you test a classic isochrone, with only either `from` or `to` in the request? 

you should also handle this case :)

kadhikari:
> can you test a classic isochrone, with only either `from` or `to` in the request?
> 
> you should also handle this case :)

* api /journeys with from or to works well in customer as well as production

PR #4401, opened 2025-05-06
feat: add mpm event config file to update max request workers

AmAdili:


AmAdili:
https://navitia.atlassian.net/browse/POPS-128

PR #4418, opened 2025-06-18
feat: authorize wider ip range

AmAdili:


docker/apache/status.conf
```diff
@@ -6,8 +6,7 @@
         <Location /server-status>
                 SetHandler server-status
                 Require local
-                Require ip 172.18.0.0/16
-                #Require ip 192.0.2.0/24
+                Require ip 172.0.0.0/8
```
dcorreia: why 172.0.0.0/8 and not 172.16.0.0/12 ?

AmAdili: to avoid future changes, but 172.16.0.0/12 is better indeed

dcorreia (approved):

PR #4422, opened 2025-07-22
Improve logs

vgeof-hove:


nonifier:
It tried locally, and it seems that all logs are still working fine 👌🏼

nonifier (approved):
Less is more !

PR #4494, opened 2025-12-30
Revert "[NAV-4486] Removing authentication with the key in the reques…

vgeof-hove:
…t params"

This reverts commit 5679b23157aeabb756f6228d2de74f13ea22c597.


vgeof-hove:
Temporarily revert the deletion of the `keys` parameter for authentication, as customers still use it.

azime (approved):

PR #4579, opened 2026-07-28
Fix: nav-5803: Use loki as default pt_planner

kadhikari:
https://navitia.atlassian.net/browse/NAV-5803


remi-lucas-hove (approved):

kadhikari:
As kraken is used for all the tests, we can't simply replace 'kraken' by 'loki'. I close this PR.

